PR #1, opened 2016-09-30
Adds initial project files for Gradle plugin

felipecsl:
This is the skeleton project structure for the Apollo Gradle plugin.
The structure is pretty standard and mostly just boilerplate.
I added an empty plugin class in Kotlin.
Heavily inspired by the square/sqldelight setup 😃


gradle.properties
```diff
@@ -0,0 +1,14 @@
+GROUP=com.apollostack.android
+VERSION_NAME=0.0.1-SNAPSHOT
+
+POM_URL=https://github.com/apollostack/apollo-android/
+POM_SCM_URL=https://github.com/square/apollo-android/
```
martijnwalraven: Should this be 'apollostack' instead of 'square' in the GitHub URL? (also the two lines down)

apollo-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'kotlin'
+apply plugin: 'idea'
```
VerachadW: Do we really need this one? Are we going to have some UI components in IDE?

gradle.properties
felipecsl: whoops, my bad. good catch!

apollo-gradle-plugin/build.gradle
felipecsl: This plugin adds tasks for generating the `.idea` files needed by Android Studio/IntelliJ. It has nothing to do with UI AFAIK.
I personally haven't used it much since these files are generated when you open the project in the IDEA, but they seem pretty standard in most java projects.
If you don't like it I can remove it.

felipecsl: Docs page: https://docs.gradle.org/current/userguide/idea_plugin.html

VerachadW: I see. This should be my misunderstanding. Please leave it be then.

build.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+  dep = [
+      androidPlugin: 'com.android.tools.build:gradle:2.2.0',
+      supportAnnotations: 'com.android.support:support-annotations:24.2.1',
+      javaPoet: 'com.squareup:javapoet:1.7.0',
+      kotlinStdLib: "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion",
+      kotlinReflect: "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion",
+      kotlinGradlePlugin: "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion",
+      junit: 'junit:junit:4.12',
+      retrolambda: 'me.tatarka:gradle-retrolambda:3.3.0'
```
jpshelley: Curious what we see retroLambda being used for? Just want to make sure its worth it since we're including kotlin as well with nice lambda support.

felipecsl: Agree, we might not need retrolambda. Will remove 👍

felipecsl:
Comments addressed. If there isn't anything else, this should be ready to go

jpshelley:
First time working with plugins to be honest, but it looks like a good start. I'm assuming we'll be writing the plugin in kotlin rather then groovy or java? I know Gradle is highly supporting Kotlin moving forward so it sounds like a good idea. But the Gradle docs still show Groovy as their main language.

VerachadW:
@jpshelley In my opinion, It's fine to write a plugin with pure Kotlin since it had been proved by Jetbrains themselves.

felipecsl:
Alright if there are no other concerns, I'll merge this by EOD (PST Timezone 😄 )

martijnwalraven:
Sounds good to me!

PR #2, opened 2016-10-09
Adds empty Apollo gradle task and compiler

felipecsl:
More boilerplate code 😄 
Having a separate compiler project allows us to potentially share the implementation with other project types (not just Android, but also pure Java, server apps, etc).


felipecsl:
ping!

martijnwalraven:
@felipecsl: This looks good to me, but I don't really know anything about Gradle builds, so I don't feel I have much to add here. Maybe @jpshelley can chime in?

jpshelley:
To be honest don't know that much about it either at this point. I'd say merge and if we need to edit it later we can do so.

rafaelwkerr:
Hi folks, I'm interested in working on this OSS, which is the prerequisite?

PR #4, opened 2016-12-03
Initial work on converting IR to Java interfaces

felipecsl:
This is still over simplified and not quite looking like the Swift code yet, but serves as a good starting point.

## Tasks

- [x] Parse IR JSON
- [x] Use IR as input for JavaPoet codegen
- [x] Write interfaces to build outputs directory
- [x] Write unit test for happy path
- [x] Dedupe the interface name for types that share the same name but have different set of fields
- [x] ~~Add more tests for non trivial IR files~~ - will do in a follow up PR

## Questions/TODOs

* ~~What to do about external types (declared in the `typesUsed` array)?~~ #13
* Handle multiple operations
* Handle zero operations
* ~~Handle other primitive types (for now it's just String)~~ #11
* ~~Handle type nullability with `@Nullable` annotations~~ #10
* ~~Handle list types~~ #9
* ~~Recurse into type definitions~~ #5
* ~~Use JSON file relative directory for the generated class package, like SQLDelight~~ #7


apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -1,6 +1,51 @@
...
+      TypeSpec.interfaceBuilder(it.type)
+          .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+          .addMethods(it.fields.map { it.toMethodSpec() })
+          .build()
+    }
+    val typeSpec = TypeSpec.interfaceBuilder(operation.operationName)
+        .addMethods(methods)
+        .addTypes(customTypes)
+        .build()
+    JavaFile.builder("test", typeSpec).build()
```
sav007: Have you considered that package will be defined by relative path of input json, for example in case  test/data/HeroName.json the package will be "test.data". It will make easier for user to organize the project structure. All generated packages will reflect the folder structure where json files are placed. The same way how SQLDelight does: https://github.com/square/sqldelight#example

felipecsl: @sav007 good point, forgot to keep track of that TODO. That was exactly what I had in mind.
Let me add an item for that

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/Field.kt
```diff
@@ -0,0 +1,20 @@
+package com.apollostack.compiler.ir
+
+class Field(val responseName: String, val fieldName: String, val type: String,
```
sav007: Should we make this class as data class?
And not sure if the `equals` method implemented correctly, as it takes into account only `type` but there can be the case when 2 fields have the same `type` but different let say `fieldName`, so are they equal?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -1,6 +1,51 @@
...
+
+  fun write(relativePath: String): Status {
+    val irAdapter = moshi.adapter(QueryIntermediateRepresentation::class.java)
+    val ir = irAdapter.fromJson(File(relativePath).readText())
+    // TODO: Handle multiple or no operations
+    val operation = ir.operations.first()
+    val methods = operation.fields.map { it.toMethodSpec() }
+    val customTypes = operation.fields.filter { it.fields.any() }.distinct().map {
+      // TODO: Also need to recurse into inner types
+      TypeSpec.interfaceBuilder(it.type)
```
sav007: So the name of interface is based on the type of field, but there can be next case:

```
query CollectionsByHandle {
  shop {
    snowboards:collectionByHandle(handle: "snowboards") {
      id
      title
    }
    
    goggles:collectionByHandle(handle: "goggles") {
      id
      title
      bodyHtml
      updatedAt
    }
  }
}
```

that will produce next IR

```
...
"fields": [
						{
							"responseName": "snowboards",
							"fieldName": "collectionByHandle",
							"type": "Collection",
							"fields": [
								{
									"responseName": "id",
									"fieldName": "id",
									"type": "ID!"
								},
								{
									"responseName": "title",
									"fieldName": "title",
									"type": "String!"
								}
							],
							"fragmentSpreads": [],
							"inlineFragments": []
						},
						{
							"responseName": "goggles",
							"fieldName": "collectionByHandle",
							"type": "Collection",
							"fields": [
								{
									"responseName": "id",
									"fieldName": "id",
									"type": "ID!"
								},
								{
									"responseName": "title",
									"fieldName": "title",
									"type": "String!"
								},
								{
									"responseName": "bodyHtml",
									"fieldName": "bodyHtml",
									"type": "String"
								},
								{
									"responseName": "updatedAt",
									"fieldName": "updatedAt",
									"type": "DateTime!"
								}
							],
							"fragmentSpreads": [],
							"inlineFragments": []
						}
...
```

So both fields `snowboards` and `goggles` has the same type `Collection` but with different fields.
I guess `type` is not enough as the name of generated interface class, if there is alias it should be taken into account. So as example there should be generated 2 interfaces with names `SnowboardsCollection` and `GogglesCollection`

felipecsl: Just implemented this!

PR #15, opened 2016-12-05
Add DESIGN.md with discussion of the mapping design

martijnwalraven:


felipecsl:
@martijnwalraven this was extremely helpful. Thanks for the detailed writeup!
I suggest we use this document as our source of truth and send PRs to it as we change our assumptions/decisions.

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+}
+
+interface HeroDetails {
+  String name();
+  List<Friend> friends();
+  interface Friend {
+    String name();
+  }
+}
+```
```
sav007: So to access the Luke's friend name we have to call this `luke.fragments(). heroDetails().friends()[0].name()` but at the same time to access friend id: `luke.friends()[0].id()`. So the `Character` object info for friends is spread across 2 places. From user perspective I guess it will be a little confusing, as I user I would like to work only with `luke.friends()`. Otherwise I need to put some logic to math users from 2 sources and merge them into one.

DESIGN.md
```diff
@@ -0,0 +1,370 @@
+# Apollo Android design discussion
+
+These are some preliminary ideas for mapping GraphQL query results to Java, based on what we currently do for Swift in Apollo iOS.
+
+## Query classes
+
+I would expect us to generate classes for queries (and other operations). Instances of these classes represent a particular query, fully configured with the required variables.
```
felipecsl: Generating interfaces gives users more flexibility for plugging in their own implementation for the models (eg.: AutoValue, Immutables or even manually extend).
Also, following the composition over inheritance pattern, we could have a separate class that takes the interfaces and configures them for execution.

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+  // etc.
+}
+
+enum Episode {
+    NEWHOPE,
+    EMPIRE,
+    JEDI
+}
+
+class HeroAndFriendsNameQuery implements GraphQLQuery<HeroAndFriendsNameQuery.Data> {
```
felipecsl: if you keep `HeroAndFriendsNameQuery` as an `interface`, then its structure is much simpler and you don't need the inner `Data` `interface` at all. Just `interface HeroAndFriendsNameQuery extends GraphQLQuery` and that's it!

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+
+            interface Friend {
+                String name();
+            }
+        }
+    }
+}
+```
+
+Nested interfaces are named based on the response name, not the field type or field name. Response names are guaranteed to be unique within a selection set, and it also seems to make more semantic sense. (A field `author` would refer to `Author` for example, and not to `User`.)
```
felipecsl: this is very important: `Response names are guaranteed to be unique within a selection set`

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+        }
+        Luke luke();
+        interface Luke {
+            String name();
+        }
+    }
+}
+```
+
+I don't think we ever want `r2` and `luke` to refer to the same type here. So I don't think we need the dedupe behavior from [PR#4](https://github.com/apollostack/apollo-android/pull/4). Even if they select the same fields now, that is accidental from the perspective of the type system, so we still want the types to be separate to avoid relying on these accidental shared fields.
```
felipecsl: Nice 👍  lets make that change

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+interface HeroDetails {
+  String name();
+  List<Friend> friends();
+  interface Friend {
+    String name();
+  }
+}
+```
+
+In Swift, this ran into limitations of the type system because types of properties in protocols (the equivalent of interfaces) are invariant. So the `List<Friend> friends()` declarations in subinterfaces would be incompatible. Even if this isn't necessarily a problem in Java, a mapping like this quickly gets complicated, and there may be simpler alternatives.
```
felipecsl: If I understand the problem correctly, `List<? extends Friend> friends()` should do it

DESIGN.md
felipecsl: Agreed this is pretty awkward.

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+`hero().fragments().humanDetails()` (can be null)
+
+## Unique nested type names
+
+An unexpected complication in Java seems to be that nested types can never have the same simple name as an enclosing type. I'm not sure I understand the reasoning here (couldn't scoping rules or qualified names still differentiate between them?), but it is part of the language spec:
+
+> It is a compile-time error if a class has the same simple name as any of its enclosing classes or interfaces.
+http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1
+
+So unless we can find a way around this, we may have to give up on nesting and generate prefixed interface names ourselves.
```
felipecsl: Wait @martijnwalraven, based on your statement `"Response names are guaranteed to be unique within a selection set"` I thought this wasn't a problem!?

DESIGN.md
sav007: Agree, another example if someone uses Realm Db and decides to use generated classes as a schema for db. Realm implies for schema models to be inherited from RealmObject. So with interfaces approach they can manually extend from RealmObject and implement generated interfaces. It's not quite good example from architecture, but still there is such possibility.

DESIGN.md
martijnwalraven: The way to look at this is that those different parts are consumed by different UI components. It requires some getting used to this way of thinking, but instead of having one model for a type, you define 'view models' that only contain the data needed by the component they belong to.

This is what [Relay](https://facebook.github.io/relay/docs/thinking-in-relay.html#content) refers to as colocation and what we also do in Apollo iOS, where you edit your UI code and corresponding `.graphql` file side by side.

If you have a need to access `id` from `HeroDetails`, you would simply add it to the fragment as well. The server won't send any duplicate data, but this makes it clear where data needs come from and when data is no longer needed.

DESIGN.md
martijnwalraven: Well, I still think we want to differentiate between a fully configured query (with values for variables) and the *result* of a query. The idea is that these query instances can be used to parse a result from a server response, read from the cache, etc. So the first result could come from the server for instance, but then after another query changes the same objects, the first query can be reloaded from the (updated) cache.

DESIGN.md
martijnwalraven: I think we need to know more about how we want to use these query instances before we decide on an architecture, but I agree we want flexibility.

DESIGN.md
martijnwalraven: Yes, that is what Swift currently doesn't support.

DESIGN.md
martijnwalraven: See above, response names are only unique within a single level of a selection set.

DESIGN.md
felipecsl: Ok that makes more sense to me now. Thanks for clarifying!

DESIGN.md
felipecsl: Indeed, sounds like we should stick to Interfaces for now and revisit that decision later down the road

DESIGN.md
felipecsl: Oh, I was under the impression that the selection set was referring to the entire thing. Well in that case we'll still need the deduping then

DESIGN.md
```diff
@@ -0,0 +1,370 @@
...
+
+interface Luke extends HeroDetails {
+  String name();
+  List<Friend> friends();
+  interface Friend extends HeroDetails.Friend {
+    String id();
+  }
+}
+
+interface HeroDetails {
```
sav007: will codegen fail if I define 2 fragments with the same name in different query files?

martijnwalraven: Yes, all `.graphql` files are processed together, and GraphQL validation rules will make sure a fragment has only been defined once (and generate a validation error otherwise).

sav007: we need to make sure that we do not regenerate all java classes in gradle plugin if only one query file has been added or changed. Or maybe it's not so bad from performance perspective for generating all of them, just to not slow down user gradle build drastically

felipecsl: Definitely something to keep in mind once we get to that part

DESIGN.md
sav007: So if we go with prefixes it will look like this:
```
interface Hero
interface HeroFriends
interface HeroFriendsFriends
```

felipecsl: I'd rather only add a prefix if we really have to (for deduping) in order to avoid clutter.

DESIGN.md
```diff
@@ -368,3 +368,232 @@ http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1
...
+
+```
+
+So we need runtime knowledge of the original query structure in order to perform normalized caching. One option would be to keep a separate representation of the query structure, as JSON for example. But that means we'd have to read in and traverse that structure when we parse a response, and we already have a representation of the query structure in the form of the generated model code.
+
+### Interleaving parsing and caching
+
+We'll have to discuss how this translates to what we're trying to do in Apollo Android, but in Apollo iOS the solution we've converged on is to interleave parsing and caching.
+
+Every model is able to instantiate itself from a `GraphQLResultReader` that can be used to read individual field values from either a network response or the normalized cache. When reading from a network response, it is also responsible for keeping track of the normalized records.
```
sav007: @felipecsl is there where we are going to divert from iOS approach, and instead of models to be instantiated by reading from `GraphQLResultReader` but have retrofit converter that will parse response to models?

DESIGN.md
```diff
@@ -368,3 +368,232 @@ http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1
...
+The only way to know whether to cache `height` as `height(unit:FOOT)` or `height(unit:METER)` is to know the `__typename` of the hero (either `Human` or `Droid`), and you won't know this until runtime.
+
+But because we're interleaving parsing and caching, we'll do the right thing by default without any additional effort.
+
+## Programming model
+
+The API you use to fetch queries should allow you to take advantage of the cache transparently. A `fetchQuery` method could allow you to specify a caching policy for example. By default, you probably want to read a query result from the cache if all data is there, but sometimes you want to make sure you always get the most recent data or have other specific policies you'd like to enforce.
+
+But while fetching a result is a one time operation, you often want to subscribe to a query. This means you're interested not just in the current result, but in any updates to that result as well. You could expose a `watchQuery` method for example, where the callback can be invoked multiple times. The client will then make sure that every time it learns a piece of data in your result changes, you get the updated result delivered.
+
```
martijnwalraven: I don't know how popular RxJava is on Android, but on iOS we didn't want to bring in a dependency on RxSwift. And we could always add a reactive interface on top of a callback style.

DESIGN.md
martijnwalraven: I don't know enough about the way serialization libraries work on Android, but it would be nice if we could still have the ability to parse models from the normalized cache (without materializing a full JSON result structure) and to keep track of normalized records while parsing a response.

digitalbuddha:
Is there something still blocking this?

digitalbuddha:
closing due to not needing a merge

PR #18, opened 2016-12-05
Use JSON file relative directory for the generated class package

### LINKED ISSUE #7: [Code Generation] Use JSON file relative directory for the generated class package
Like Sqldelight does

sav007:
Use JSON file relative directory for the generated class package.
Refactor junit tests

Resolves #7 


apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -9,18 +9,27 @@ import java.io.File
...
+    val packageName = queryFile.absolutePath.formatPackageName()
     val irAdapter = moshi.adapter(QueryIntermediateRepresentation::class.java)
-    val ir = irAdapter.fromJson(File(relativePath).readText())
+    val ir = irAdapter.fromJson(queryFile.readText())
     // TODO: Handle multiple or no operations
     val operation = ir.operations.first()
     val typeSpecBuilder = OperationTypeSpecBuilder(operation.operationName, operation.fields)
-    JavaFile.builder("test", typeSpecBuilder.build()).build()
-        .writeTo(OUTPUT_DIRECTORY.fold(File("build"), ::File))
+    JavaFile.builder(packageName, typeSpecBuilder.build()).build().writeTo(OUTPUT_DIRECTORY.fold(File("build"), ::File))
```
felipecsl: nit: break this in multiple lines for better readability

apollo-compiler/src/test/kotlin/com/apollostack/compiler/GraphqlCompilerTest.kt
```diff
@@ -11,167 +11,70 @@ class GraphqlCompilerTest {
...
-    assertThat(outputFile.readText()).isEqualTo(
-      """package test;
-
-import java.lang.Integer;
-import java.lang.String;
-import java.util.List;
-
-public interface HeroDetails {
-  Character hero();
+    val queryFile = File("src/test/graphql/com/apollostack/compiler/query/HeroDetails.json")
```
felipecsl: nit: I think I'd move the test files to `src/test/graphql/com/example/` instead, so it's clear that those are test files. 
I think using `com/apollostack/compiler` makes a bit confusing since those are not actual real files, just test data

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -9,18 +9,27 @@ import java.io.File
...
   companion object {
     const val FILE_EXTENSION = "graphql"
     val OUTPUT_DIRECTORY = listOf("generated", "source", "apollo")
   }
+
+  private fun String.formatPackageName(): String {
+    val parts = split(File.separatorChar)
+    (2..parts.size)
+      .filter { parts[it - 2] == "src" && parts[it] == "graphql" }
+      .forEach { return parts.subList(it + 1, parts.size).dropLast(1).joinToString(".") }
```
felipecsl: I don't quite get what's going on in this method. @sav007 do you mind elaborating?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -9,18 +9,32 @@ import java.io.File
...
+  private fun String.formatPackageName(): String {
+    val parts = split(File.separatorChar)
+    val srcFolderIndex = parts.indexOfFirst { it == "src" }
+    val graphqlFolderIndex = parts.indexOfFirst { it == "graphql" }
+    if (graphqlFolderIndex - srcFolderIndex != 2) {
+      throw IllegalArgumentException("Files must be organized like src/main/graphql/...")
+    }
+
+    return parts.subList(graphqlFolderIndex + 1, parts.size).dropLast(1).joinToString(".")
+  }
```
sav007: This method validates that IR json placed under correct `src/[main|test|etc]/graphql/...` folder and converts the absolutePath to package name by striping first parts of the path. So as example in our tests  `src/test/graphql/com/example` will result `com.example`

felipecsl: Ok this is much more readable now, thanks for refactoring!

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -9,18 +9,32 @@ import java.io.File
...
 
   companion object {
     const val FILE_EXTENSION = "graphql"
     val OUTPUT_DIRECTORY = listOf("generated", "source", "apollo")
   }
+
+  private fun String.formatPackageName(): String {
+    val parts = split(File.separatorChar)
+    val srcFolderIndex = parts.indexOfFirst { it == "src" }
+    val graphqlFolderIndex = parts.indexOfFirst { it == "graphql" }
```
felipecsl: I feel like this could break if the File's absolute path contains two directories called `src` and `graphql` outside of the project structure that pass the `if` check below.
Can we use the relative path instead?

sav007: Well, I guess the chances to have such case is very low. 
Sorry but by relative path what do you mean? Roll back to have `relativePath:String` as param instead of File?

felipecsl: Never mind, I see that sqldelight is doing the same thing, so that's probably fine 😄

Issue #17, opened 2016-12-05
[Code Generation] Generate nested typeSpecs instead of in the root level


felipecsl:
As discussed on #15 
Need to use the field `responseName` as the interface name and handle duplicated names by making them unique

felipecsl:
Fixed via #20

PR #20, opened 2016-12-06
Handle nullability and nesting

### LINKED ISSUE #10: [Code Generation] Handle type nullability
Add with `@Nullable` annotations to methods

felipecsl:
this PR was supposed to handle just the Nullability part, but turned out
to be much bigger since once I started to implement that, I started
bumping into roadblocks with the current implementation that wouldn't
let me properly propragate the nullability information without
refactoring a lot of code.
It ended up being much easier to also make the change to generate
nested types with the correct names based on the responseName field.
The code turned out to be **at lot** simpler and much more similar to
the Swift counterparts. This still doesnt handle deduping of duplicate
types but that shouldnt be too hard to implement now. We just need to
decide on how to actualy do it (maybe add a suffix to the interface
name? Maybe a number, or a letter, or a symbol?) I'm open to ideas here.

Fixes #10 and #17

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollostack/compiler/OperationTypeSpecBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollostack/compiler/OperationTypeSpecBuilder.kt
@@ -1,58 +1,26 @@
 package com.apollostack.compiler
 
 import com.apollostack.compiler.ir.Field
+import com.cesarferreira.pluralize.singularize
 import com.squareup.javapoet.TypeSpec
-import java.util.*
 import javax.lang.model.element.Modifier
 
 class OperationTypeSpecBuilder(
-  val operationName: String,
-  val fields: List<Field>
-) {
+    val operationName: String,
+    val fields: List<Field>) {
 
   fun build(): TypeSpec {
-    val typeSpecs = buildTypeSpecs(operationName, fields, "")
-    return typeSpecs
-      .single { it.name == operationName }
-      .toBuilder()
-      .addTypes(typeSpecs.filter { it.name != operationName })
-      .build()
+    return buildTypeSpecs(operationName, fields)
   }
 
-  private fun buildTypeSpecs(currentTypeName: String, fields: List<Field>, typeNamePrefix: String): List<TypeSpec> {
-    val fieldToTypeMap = HashMap<Field, String>()
-    val typeToFieldMap = HashMap<String, Field>()
-    val nonScalarFields = fields.filter { it.fields?.any() ?: false }
-    nonScalarFields.forEach {
-      val fieldType = "$typeNamePrefix${it.toTypeName()}"
-      val existingMapping = typeToFieldMap[fieldType]
-      if (existingMapping != null && it.fields != existingMapping.fields) {
-        // If we already saw this type and it has a different set of fields, it means
-        // we have the same type used multiple times with different fields. In this case,
-        // we'll use Field.toTypeName() to disambiguate the name since the type cannot be
-        // used more than once. We also need to use the explicit naming for the previously
-        // seen type, so we need to also remove it and re-add with the explicit naming.
-
-        val fieldNewType = "$typeNamePrefix${it.toTypeName(it.responseName)}"
-        typeToFieldMap.put(fieldNewType, it)
-        fieldToTypeMap.put(it, fieldNewType)
-
-        val existingFieldType = "$typeNamePrefix${it.toTypeName()}"
-        val existingFieldNewType = "$typeNamePrefix${existingMapping.toTypeName(existingMapping.responseName)}"
-        typeToFieldMap.put(existingFieldNewType, existingMapping)
-        typeToFieldMap.remove(existingFieldType)
-        fieldToTypeMap.put(existingMapping, existingFieldNewType)
-      } else {
-        typeToFieldMap.put(fieldType, it)
-        fieldToTypeMap.put(it, fieldType)
-      }
-    }
-
-    val typeSpec = TypeSpec.interfaceBuilder(currentTypeName)
-      .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
-      .addMethods(fields.map { it.toMethodSpec(fieldToTypeMap[it]) })
-      .build()
-
-    return listOf(typeSpec) + typeToFieldMap.flatMap { buildTypeSpecs(it.key, it.value.fields!!, it.key) }
+  private fun buildTypeSpecs(currentTypeName: String, fields: List<Field>): TypeSpec {
+    val innerTypes = fields
+        .filter(Field::isNonScalar)
+        .map { buildTypeSpecs(it.responseName.capitalize().singularize(), it.fields!!) }
+    return TypeSpec.interfaceBuilder(currentTypeName)
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .addMethods(fields.map(Field::toMethodSpec))
+        .addTypes(innerTypes)
+        .build()
   }
 }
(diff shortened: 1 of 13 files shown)


martijnwalraven:
Sounds great! I agree with [your comment](https://github.com/apollostack/apollo-android/pull/15#discussion_r90968893) to only deduplicate when needed.

Something like `interface HeroFriendsFriends` makes most semantic sense, but feels weird when nested because that means the qualified name already has that same information. So maybe just a simple number suffix, like `Friends2`?

Is there any precedence for these types of workarounds in Java / Android? (I still don't understand the restriction, because it seems the type being nested and having a different fully qualified name should be enough to disambiguate.)

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
```diff
@@ -20,30 +22,47 @@ sealed class GraphQlType(val nullable: Boolean) {
...
-    }
-  }
+  fun toJavaTypeName() = graphQlTypeToJavaTypeName(this, !nullable, nullable)
 
   companion object {
-
-    private fun String.normalizeTypeName() = removeSuffix("!").removeSurrounding(prefix = "[", suffix = "]").removeSuffix("!")
+    private val NULLABLE_ANNOTATION = AnnotationSpec.builder(Nullable::class.java).build()
+    private val LIST_TYPE = ClassName.get(List::class.java)
+    private val GRAPHQLTYPE_TO_JAVA_TIME = mapOf(
```
sav007: TIME you probably meant TYPE?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/Field.kt
```diff
@@ -1,31 +1,46 @@
...
+        .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+        .build()
   }
 
-  private fun String.toTypeName(): TypeName = GraphQlType.resolveByName(this).toJavaTypeName()
+  private fun toTypeName(responseType: String): TypeName =
+      GraphQlType.resolveByName(responseType).toJavaTypeName()
+
+  private fun methodResponsType(): String {
+    // TODO: Do we also need to handle nested lists?
```
sav007: I thought our GraphQlType class supports nested lists, I remember I even added test case for that, or I missing smth?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -1,58 +1,26 @@
...
+  private fun buildTypeSpecs(currentTypeName: String, fields: List<Field>): TypeSpec {
+    val innerTypes = fields
+        .filter(Field::isNonScalar)
+        .map { buildTypeSpecs(it.responseName.capitalize().singularize(), it.fields!!) }
+    return TypeSpec.interfaceBuilder(currentTypeName)
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .addMethods(fields.map(Field::toMethodSpec))
+        .addTypes(innerTypes)
+        .build()
   }
```
sav007: wow, nice so simpler :)

sav007:
As for dup nested declaration I guess we need to add prefix|suffix to the inner interface only, so as eaxample to fix this issue:
```
  static interface Hero {

    static interface Friend {

      static interface Node {

        static interface Friend {

        }
      }
    }
  }
```

we need to turn it into: 
```
static interface Hero {

    static interface Friend {

      static interface Node {

        static interface NodeFriend {

        }
      }
    }
  }
```

And that is all, we don't need to prefix|suffix them all. And the algorithm should be simple, we just track if we have already defined outer interface with the same name, we just add prefix - the name of surrounding. Or even simpler just add "_" :)

sav007:
Examples:

```
static interface Hero {

    static interface Friend {

      static interface Node {

        static interface NodeFriend {
          
          static interface NodeFriendHero {
            
            static interface NodeFriendHeroFriend {
              
            } 
          }

        }
      }
    }
  }
```
or
```
static interface Hero {

    static interface Friend {

      static interface Node {

        static interface Friend1 {

          static interface Hero1 {

            static interface Friend2 {

            }
          }

        }
      }
    }
  }
```

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
```diff
@@ -20,30 +22,47 @@ sealed class GraphQlType(val nullable: Boolean) {
...
-      is GraphQLList -> ParameterizedTypeName.get(ClassName.get(List::class.java), listType.toJavaTypeName())
-      is GraphQlUnknown -> ClassName.get("", typeName)
-    }
-  }
+  fun toJavaTypeName() = graphQlTypeToJavaTypeName(this, !nullable, nullable)
 
   companion object {
-
-    private fun String.normalizeTypeName() = removeSuffix("!").removeSurrounding(prefix = "[", suffix = "]").removeSuffix("!")
+    private val NULLABLE_ANNOTATION = AnnotationSpec.builder(Nullable::class.java).build()
```
marwanad: Any preference for this over `android.support.annotation.Nullable`?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/Field.kt
felipecsl: Yeah, actually the scalarTypes test uses a nested list and it seems to be working fine. I'll remove the TODO

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
```diff
@@ -1,11 +1,13 @@
 package com.apollostack.compiler.ir
 
+import com.apollostack.compiler.normalizeTypeName
+import com.squareup.javapoet.AnnotationSpec
 import com.squareup.javapoet.ClassName
 import com.squareup.javapoet.ParameterizedTypeName
 import com.squareup.javapoet.TypeName
+import javax.annotation.Nullable
```
felipecsl: Either way works fine. I guess android support annotations is better. I'll change it

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
felipecsl: will switch to android support

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
felipecsl: good catch :)

felipecsl:
As for the naming pattern for deduping, I propose we suffix it with `$`.
This is a fairly common practice, especially on annotation processors you see that a lot. This one is a slightly different scenario but I think it wouldn't sound that weird.
My second option would be to suffix with `_`. 
What do you guys think?

martijnwalraven:
So if there are more than two duplicates this would become `$$` etc?

felipecsl:
@martijnwalraven yeah, not pretty but gets the job done...

sav007:
so lets stop on using '$' for now, to put this issue to bed :)

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
felipecsl: android support annotations is neither on jcenter or maven central, so we're gonna have to stick to jsr305

felipecsl:
Anyone has other comments on this? Otherwise merging on green build 🍏

sav007:
LGTM

PR #24, opened 2016-12-07
Add support of enum types

### LINKED ISSUE #22: [Code Generation]  Add support of enum types
Latest `apollo-codegen:0.9.5` includes type information as part of `typesUsed`.
See https://github.com/apollostack/apollo-codegen/commit/a90cf59a4afb4f4ec372e568c3e40f63a1d5fcd1

sav007:
Add support of generation enum defined in IR `typesUsed` section
Closes #22 

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/TypeDeclaration.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/TypeDeclaration.kt
@@ -10,13 +10,23 @@ data class TypeDeclaration(
 
   fun toTypeSpec(): TypeSpec {
     if (kind == "EnumType") {
-      val builder = TypeSpec.enumBuilder(name).addModifiers(Modifier.PUBLIC)
-      values?.forEach {
-        builder.addEnumConstant(it.name.toUpperCase())
-      }
-      return builder.build()
+      return enumTypeToTypeSpec()
     } else {
       throw UnsupportedOperationException("unsupported $kind type declaration")
     }
+  }
+
+  private fun enumTypeToTypeSpec():TypeSpec {
+    fun TypeSpec.Builder.addTypeDeclarationValue(value: TypeDeclarationValue) {
+      val enumConstBuilder = TypeSpec.anonymousClassBuilder("")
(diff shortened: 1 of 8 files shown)


apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/TypeDeclaration.kt
```diff
@@ -0,0 +1,22 @@
...
+
+data class TypeDeclaration(
+    val kind: String,
+    val name: String,
+    val values: List<TypeDeclarationValue>?) {
+
+  fun toTypeSpec(): TypeSpec {
+    if (kind == "EnumType") {
+      val builder = TypeSpec.enumBuilder(name).addModifiers(Modifier.PUBLIC)
+      values?.forEach {
```
felipecsl: if you have null `values`, you'll end up getting an empty enum, which probably won't compile

apollo-compiler/src/test/graphql/com/example/EpisodeExpected.java
```diff
@@ -0,0 +1,9 @@
+package com.example;
+
+public enum Episode {
```
felipecsl: should we make the enum also an inner class to the query? @martijnwalraven what do you think?
Also would be nice to use the `description` field from the IR to add javadocs to the enum constants

sav007: As I understood enums are global types, and they are the same across all queries

sav007: As for description sure will add

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/TypeDeclaration.kt
sav007: actually empty enum compiles, agree it doesn't make sense to have such type but otherwise are we going to throw exception if enum is empty and fail generation?

felipecsl: lets leave like this for now, not a big deal

PR #37, opened 2016-12-09
Adds empty sample app module + code formatting

felipecsl:


apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -28,12 +28,9 @@ open class GraphqlCompiler {
 
   private fun String.formatPackageName(): String {
     val parts = split(File.separatorChar)
-
-    for (i in 2..parts.size) {
-      if (parts[i - 2] == "src" && parts[i] == "graphql") {
-        return parts.subList(i + 1, parts.size).dropLast(1).joinToString(".")
-      }
-    }
+    (2..parts.size)
```
felipecsl: @sav007 turns out this is the code you had originally haha, Android Studio suggested the auto refactoring

sav007: Yeah :)

sav007:
Hey could you please extract your code style formatter and put it under git, that we all will be at the same page. And can we increase margin from 100 -> 120 please

felipecsl:
@sav007 i did it already, it's the file `.idea/codeStyleSettings.xml` in this PR
I'll increase the margin to 120

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
marwanad: What do you think about `parts.size - 1 downTo 2` instead? 
It doesn't really matter but that should handle all the cases now 😀 (including the `src/foo/grahpql/src/bar/graphql/Query.json`)

felipecsl: sounds good to me

felipecsl:
@sav007 updated, PTAL

Issue #41, opened 2016-12-10
[Code Generation] Generate query wrapper class around query text source and variables


sav007:
Create generic GraphQLQuery interface like iOS does
Generate query class implemtentation for each query by implementing GraphQLQuery.
Generated query class implemtentations should allow set arguments.

See https://github.com/apollostack/apollo-android/blob/design-doc/DESIGN.md#query-classes
Swift examples can be found here: https://github.com/apollostack/apollo-ios/blob/master/Tests/ApolloTests/StarWars/API.swift

sav007:
AS example of generated class on swift:
```
public final class HeroAndFriendsNamesQuery: GraphQLQuery {
  public static let operationDefinition =
    "query HeroAndFriendsNames($episode: Episode) {" +
    "  hero(episode: $episode) {" +
    "    __typename" +
    "    name" +
    "    friends {" +
    "      __typename" +
    "      name" +
    "    }" +
    "  }" +
    "}"

  public let episode: Episode?

  public init(episode: Episode? = nil) {
    self.episode = episode
  }

  public var variables: GraphQLMap? {
    return ["episode": episode]
  }
```

felipecsl:
Fixed via #45

PR #45, opened 2016-12-12
Add generation of query wrapper with operation/fragments definitions

sav007:
Add new `apollo-api` module that will share api classes between compiler and runtime modules, as we can't just add `apollo-runtime` (android lib module) as dependency to `apollo-compiler` (java module).

Add generation of wrapper query class that holds operation/fragments definitions as well as nested interface class that represents response.

@felipecsl @marwanad @martijnwalraven pls take a look

Connected to #41

### DIFF
--- a/apollo-api/build.gradle
+++ b/apollo-api/build.gradle
@@ -1,0 +1,4 @@
+apply plugin: 'java'
+
+targetCompatibility = JavaVersion.VERSION_1_6
+sourceCompatibility = JavaVersion.VERSION_1_6
--- a/apollo-api/build.gradle
+++ b/apollo-api/build.gradle
@@ -1,4 +1,4 @@
 apply plugin: 'java'
 
-targetCompatibility = JavaVersion.VERSION_1_6
-sourceCompatibility = JavaVersion.VERSION_1_6
+targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_7
--- a/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
+++ b/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
@@ -4,7 +4,7 @@ import java.util.List;
 
 public interface GraphQLQuery {
 
-  String getOperationDefinition();
+  String operationDefinition();
 
-  List<String> getFragmentDefinitions();
+  List<String> fragmentDefinitions();
 }
--- a/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
+++ b/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
@@ -2,7 +2,7 @@ package com.apollostack.api;
 
 import java.util.List;
 
-public interface GraphQLQuery<T> {
+public interface GraphQLQuery {
 
   String getOperationDefinition();
 
--- a/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
+++ b/apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
@@ -1,0 +1,10 @@
+package com.apollostack.api;
+
+import java.util.List;
+
+public interface GraphQLQuery<T> {
+
+  String getOperationDefinition();
+
+  List<String> getFragmentDefinitions();
+}
--- a/apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
@@ -26,7 +26,7 @@ class QueryTypeSpecBuilder(
 
   private fun TypeSpec.Builder.addOperationSourceDefinition(operation: Operation): TypeSpec.Builder {
     addField(FieldSpec.builder(ClassName.get(String::class.java), OPERATION_SOURCE_FIELD_NAME)
-        .addModifiers(Modifier.PRIVATE, Modifier.FINAL)
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
         .initializer("\$S", operation.source)
         .build()
     )
@@ -68,7 +68,7 @@ class QueryTypeSpecBuilder(
           FieldSpec.builder(
               ParameterizedTypeName.get(ClassName.get(List::class.java), ClassName.get(String::class.java)),
               FRAGMENT_SOURCES_FIELD_NAME)
-              .addModifiers(Modifier.PRIVATE, Modifier.FINAL)
+              .addModifiers(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
               .initializer(fragments.toSourceDefinitionCode())
               .build()
       )
--- a/apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
@@ -1,0 +1,90 @@
+package com.apollostack.compiler
+
+import com.apollostack.api.GraphQLQuery
+import com.apollostack.compiler.ir.Fragment
+import com.apollostack.compiler.ir.Operation
+import com.squareup.javapoet.*
+import java.util.*
+import javax.lang.model.element.Modifier
+
+class QueryTypeSpecBuilder(
+    val operation: Operation,
+    val fragments: List<Fragment>
+) {
+  fun build(): TypeSpec {
+    val queryClassName = operation.operationName.capitalize()
+    val queryInterfaceTypeName = ParameterizedTypeName.get(ClassName.get(GraphQLQuery::class.java),
+        ClassName.get("", "$queryClassName.${Operation.INTERFACE_TYPE_SPEC_NAME}"))
+    return TypeSpec.classBuilder(queryClassName)
+        .addSuperinterface(queryInterfaceTypeName)
+        .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
+        .addOperationSourceDefinition(operation)
+        .addFragmentSourceDefinitions(fragments)
+        .addType(operation.toTypeSpec())
+        .build()
+  }
+
+  private fun TypeSpec.Builder.addOperationSourceDefinition(operation: Operation): TypeSpec.Builder {
+    addField(FieldSpec.builder(ClassName.get(String::class.java), OPERATION_SOURCE_FIELD_NAME)
+        .addModifiers(Modifier.PRIVATE, Modifier.FINAL)
+        .initializer("\$S", operation.source)
+        .build()
+    )
+    addMethod(MethodSpec.methodBuilder("getOperationDefinition")
+        .addAnnotation(Override::class.java)
+        .addModifiers(Modifier.PUBLIC)
+        .returns(ClassName.get(String::class.java))
+        .addStatement("return $OPERATION_SOURCE_FIELD_NAME")
+        .build()
+    )
+    return this
+  }
+
+  private fun List<Fragment>.toSourceDefinitionCode(): CodeBlock {
+    val codeBlockBuilder = CodeBlock.builder()
+    codeBlockBuilder.add("\$T.unmodifiableList(", ClassName.get(Collections::class.java))
+    codeBlockBuilder.add("\$T.asList(\n", ClassName.get(Arrays::class.java))
+    codeBlockBuilder.indent()
+    forEachIndexed { i, fragment ->
+      codeBlockBuilder.add("\$S\$L", fragment.source, if (i < this.size - 1) "," else "")
+    }
+    codeBlockBuilder.unindent()
+    codeBlockBuilder.add("\n)")
+    codeBlockBuilder.add(")")
+    return codeBlockBuilder.build()
+  }
+
+  private fun TypeSpec.Builder.addFragmentSourceDefinitions(fragments: List<Fragment>): TypeSpec.Builder {
+    if (fragments.isEmpty()) {
+      addMethod(MethodSpec.methodBuilder("getFragmentDefinitions")
+          .addAnnotation(Override::class.java)
+          .addModifiers(Modifier.PUBLIC)
+          .returns(ParameterizedTypeName.get(ClassName.get(List::class.java), ClassName.get(String::class.java)))
+          .addStatement("return \$T.emptyList()", ClassName.get(Collections::class.java))
+          .build()
+      )
+    } else {
+      addField(
+          FieldSpec.builder(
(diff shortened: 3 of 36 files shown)


sav007:
As example of generated query class: https://github.com/apollostack/apollo-android/blob/6f7592bf0552ec00b77e820dc4778e0b10cf6edb/apollo-compiler/src/test/graphql/com/example/simple_fragment/SimpleFragmentExpected.java

apollo-api/build.gradle
```diff
@@ -0,0 +1,4 @@
+apply plugin: 'java'
+
+targetCompatibility = JavaVersion.VERSION_1_6
+sourceCompatibility = JavaVersion.VERSION_1_6
```
felipecsl: why not java 7?

apollo-compiler/src/test/graphql/com/example/directives/HeroNameDirectiveExpected.java
```diff
@@ -1,12 +1,35 @@
...
+import java.lang.Override;
 import java.lang.String;
+import java.util.Collections;
+import java.util.List;
 import javax.annotation.Nullable;
 
-public interface HeroNameDirective {
-  @Nullable Hero hero();
+public final class HeroNameDirective implements GraphQLQuery<HeroNameDirective.Data> {
+  private final String OPERATION_DEFINITION = "query HeroNameDirective {\n"
```
felipecsl: static

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
```diff
@@ -1,15 +1,56 @@
...
+  private final String OPERATION_DEFINITION = "query FragmentFriendsConnection {\n"
+      + "  hero {\n"
+      + "    __typename\n"
+      + "    ...HeroDetails\n"
+      + "  }\n"
+      + "}";
 
-  interface Hero {
-    Fragments fragments();
+  private final List<String> FRAGMENT_DEFINITIONS = Collections.unmodifiableList(Arrays.asList(
```
felipecsl: why a list? also static

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
```diff
@@ -1,15 +1,56 @@
...
+  }
+
+  public interface Data {
+    @Nullable Hero hero();
+
+    interface Hero {
+      Fragments fragments();
+
+      interface Fragments {
+        HeroDetails heroDetails();
```
felipecsl: shouldnt this be annotated with `@Nonnull` now?

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
```diff
@@ -1,15 +1,56 @@
...
+import java.lang.Override;
+import java.lang.String;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import javax.annotation.Nullable;
 
-public interface FragmentFriendsConnection {
-  @Nullable Hero hero();
+public final class FragmentFriendsConnection implements GraphQLQuery<FragmentFriendsConnection.Data> {
```
felipecsl: why change it to a class?

sav007: First: want to override `com.apollostack.api.GraphQLQuery#getOperationDefinition` and `com.apollostack.api.GraphQLQuery#getFragmentDefinitions`, as well as in future it will hold argumetns. So `Apollo` client will deal only with `GraphQLQuery` to get everything required to run query.
Second there is no need for user to create impl of this query, only for response.

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
sav007: yeah forgot to add this modifier, will add

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
sav007: Strange should be, will check

sav007: Ok checked, it wasn't implemented yet. Will create ticket for that

sav007: #46 created

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
felipecsl: hmm I guess that makes sense since all the interface definition has moved to the `Data` interface 👍

apollo-compiler/src/test/graphql/com/example/fragment_friends_connection/FragmentFriendsConnectionExpected.java
```diff
@@ -1,15 +1,56 @@
...
+        + "        name\n"
+        + "      }\n"
+        + "    }\n"
+        + "  }\n"
+        + "}"
+  ));
 
-    interface Fragments {
-      HeroDetails heroDetails();
+  @Override
```
felipecsl: method annotations in the same line, like fields

sav007: oh, wait override belongs to method not to return type. So JavaPoet allows only to add annotation to method that will lead to what we have now:
```
@Override
  public String getOperationDefinition() {
    return OPERATION_DEFINITION;
  }
```
We can annotate return type but it will generate:
```
public @Override String getOperationDefinition() {
    return OPERATION_DEFINITION;
  }
```
That is not right as override in this case belongs to type but not to method itself

felipecsl: yeah I know, I was just referring to keeping it in the same line as the method like below:

```java
@Override public String getOperationDefinition() {
  return OPERATION_DEFINITION;
}
```

apollo-api/src/main/java/com/apollostack/api/GraphQLQuery.java
```diff
@@ -0,0 +1,10 @@
+package com.apollostack.api;
+
+import java.util.List;
+
+public interface GraphQLQuery<T> {
```
felipecsl: It doesn't make sense to add a generic type to this interface if none of its methods are bound to `<T>`. Looks like we can remove it for now

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+import java.util.*
+import javax.lang.model.element.Modifier
+
+class QueryTypeSpecBuilder(
+    val operation: Operation,
+    val fragments: List<Fragment>
+) {
+  fun build(): TypeSpec {
+    val queryClassName = operation.operationName.capitalize()
+    val queryInterfaceTypeName = ParameterizedTypeName.get(ClassName.get(GraphQLQuery::class.java),
```
felipecsl: sounds like we should create a constant for `ClassName.get(GraphQLQuery::class.java)`. If we have lots of queries, this will be done N times and we can avoid that, just like we're doing for the nullable/nonnull annotations

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+  private fun TypeSpec.Builder.addOperationSourceDefinition(operation: Operation): TypeSpec.Builder {
+    addField(FieldSpec.builder(ClassName.get(String::class.java), OPERATION_SOURCE_FIELD_NAME)
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
+        .initializer("\$S", operation.source)
+        .build()
+    )
+    addMethod(MethodSpec.methodBuilder("getOperationDefinition")
+        .addAnnotation(JavaPoetUtils.OVERRIDE_ANNOTATION)
+        .addModifiers(Modifier.PUBLIC)
+        .returns(ClassName.get(String::class.java))
```
felipecsl: also can constantize `ClassName.get(String::class.java)`

apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/QueryExpected.java
```diff
@@ -1,29 +1,70 @@
...
+      + "  }\n"
+      + "  luke: hero {\n"
+      + "    __typename\n"
+      + "    ...HumanDetails\n"
+      + "    ...DroidDetails\n"
+      + "  }\n"
+      + "}";
 
-  @Nullable Luke luke();
+  public static final List<String> FRAGMENT_DEFINITIONS = Collections.unmodifiableList(Arrays.asList(
```
felipecsl: why is this a list?

apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/QueryExpected.java
```diff
@@ -1,29 +1,70 @@
...
+        + "}","fragment DroidDetails on Droid {\n"
+        + "  name\n"
+        + "  primaryFunction\n"
+        + "}"
+  ));
 
-  interface R2 {
-    Fragments fragments();
+  @Override
+  public String getOperationDefinition() {
```
felipecsl: omit the `get` prefixes

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+          .addAnnotation(JavaPoetUtils.OVERRIDE_ANNOTATION)
+          .addModifiers(Modifier.PUBLIC)
+          .returns(ParameterizedTypeName.get(ClassName.get(List::class.java), ClassName.get(String::class.java)))
+          .addStatement("return \$T.emptyList()", ClassName.get(Collections::class.java))
+          .build()
+      )
+    } else {
+      addField(
+          FieldSpec.builder(
+              ParameterizedTypeName.get(ClassName.get(List::class.java), ClassName.get(String::class.java)),
```
felipecsl: we should constantize all these ClassName.get calls

apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/QueryExpected.java
sav007: we can have several fragments definitions, we can append all fragment sources to one but not sure do we need it?

apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/QueryExpected.java
sav007: even for generated java classes? I thought it's only applicable for Kotlin

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+    codeBlockBuilder.unindent()
+    codeBlockBuilder.add("\n)")
+    codeBlockBuilder.add(")")
+    return codeBlockBuilder.build()
+  }
+
+  private fun TypeSpec.Builder.addFragmentSourceDefinitions(fragments: List<Fragment>): TypeSpec.Builder {
+    if (fragments.isEmpty()) {
+      addMethod(MethodSpec.methodBuilder("getFragmentDefinitions")
```
felipecsl: omit the get: `fragmentDefinitions`

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      addField(
+          FieldSpec.builder(
+              ParameterizedTypeName.get(JavaPoetUtils.LIST_CLASS_NAME, JavaPoetUtils.STRING_CLASS_NAME),
+              FRAGMENT_SOURCES_FIELD_NAME)
+              .addModifiers(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
+              .initializer(fragments.toSourceDefinitionCode())
+              .build()
+      )
+      addMethod(MethodSpec.methodBuilder("getFragmentDefinitions")
```
felipecsl: same here, also extract a constant to avoid repetition

apollo-compiler/src/main/kotlin/com/apollostack/compiler/QueryTypeSpecBuilder.kt
```diff
@@ -0,0 +1,88 @@
...
+    return codeBlockBuilder.build()
+  }
+
+  private fun TypeSpec.Builder.addFragmentSourceDefinitions(fragments: List<Fragment>): TypeSpec.Builder {
+    if (fragments.isEmpty()) {
+      addMethod(MethodSpec.methodBuilder("getFragmentDefinitions")
+          .addAnnotation(JavaPoetUtils.OVERRIDE_ANNOTATION)
+          .addModifiers(Modifier.PUBLIC)
+          .returns(ParameterizedTypeName.get(JavaPoetUtils.LIST_CLASS_NAME, JavaPoetUtils.STRING_CLASS_NAME))
+          .addStatement("return \$T.emptyList()", ClassName.get(Collections::class.java))
```
felipecsl: could constantize `ClassName.get(Collections::class.java)` too

apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/QueryExpected.java
felipecsl: doesn't sound like we need this. `fragmentDefinitions()` is not being called anywhere yet, right?

sav007:
@felipecsl For now, but we will require it to send as string to server, as we have to define with query all fragments being used. So I assume the query executor will aggregate all fragment sources to one string and send with query string.

felipecsl:
@sav007 alright this seems to be good to go, wanna merge this? I'll need it for #47

PR #47, opened 2016-12-12
[Runtime] First stab at Apollo runtime API

### LINKED ISSUE #44: [Runtime] Create simple Apollo object
It should have a simple builder interface that takes at least a retrofit object and is able to fire GraphQL requests using the generated query interfaces and optionally arguments.

felipecsl:
After #45 is merged, we can modify this to use `GraphQLQuery` as the `Observable` type in the service interface.
I'm still not 100% sure on what the right API for `Apollo` is, but this sounds like a sane starting point.
We can keep iterating on it until it feels right
Fixes #44


apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
```diff
@@ -0,0 +1,14 @@
+package com.apollostack.android;
+
+import android.database.Observable;
```
sav007: database? or you meant observable from Rx?

build.gradle
```diff
@@ -10,9 +10,11 @@ ext {
       compiletesting: 'com.google.testing.compile:compile-testing:0.10',
       javaPoet: 'com.squareup:javapoet:1.8.0',
       moshi: 'com.squareup.moshi:moshi:1.3.1',
+      rxjava: 'io.reactivex:rxjava:1.2.3',
```
sav007: So we force user to use RxJava? I guess @martijnwalraven had some concerns about this.
I see cases where regular callbacks can be easy for user without Rx knowledge, but not sure

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+/** A {@link retrofit2.Retrofit} service for executing GraphQL requests. */
+public interface GraphqlService {
```
sav007: So how API will look like from user perspective, do I need to extend/implement this interface?

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
```diff
@@ -0,0 +1,14 @@
...
+  // TODO: Replace the wildcard with an interface once apollo-api module is merged.
```
sav007: So this call will return `Observable<GraphQLQuery>`? But why query, this method should return the `DATA`? 
So maybe it should be smth like this:
```
<DATA> Observable<DATA> execute(GraphQLQuery<DATA> query)
```
`GraphQLQuery` will have method `marshal`, and then we don't need `arguments ` param

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
felipecsl: oops, wrong import

build.gradle
felipecsl: yeah I agree RxJava shouldn't be required.

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
felipecsl: Yeah my idea was that you'd extend this interface, but now that we're talking about it, it doens't make a lot of sense. Maybe this should be generated by the compiler as well, I'm not sure. Or maybe the user should implement it themselves and we don't provide it at all.
What do you think?

sav007: I would rather generate this interface or maybe even it won't be required to be dynamic (generated) but rather compile time (will ship with runtime module) as I see it's used only internally. So from user perspective the API should be very simple: `Apollo.execute(query, callback)` or with Rx `Apollo.execute(query).subscribe()` and he shouldn't care how this execution was implemented (using or not retrofit, etc. ). Yes if user decides to set-up own converter we should allow that but that will expose only settings of `Retrofit.Builder` not service.

felipecsl: Oh, I just looked into the Retrofit documentation. My concern was that we needed the API Endpoint for the GraphQL requests, but that can be set as part of the Retrofit [`baseUrl`](http://square.github.io/retrofit/2.x/retrofit/retrofit2/Retrofit.Builder.html#baseUrl-okhttp3.HttpUrl-), so the API interface methods don't need to set the path themselves.
Alright, we're fine then. You're right that this should be only used internally.
I think we can have this be generated by the compiler as well, I'll make the changes.
One thing though is that we'll need some kind of plugin setting where the user will tell us if  they want to use RxJava or not, since that affects the generated code. Something like:
```groovy
apollo {
  useRxJava true
}
```
We could also add other plugin settings this way in the future.

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
felipecsl: Right, I'll add a common interface to all the `Data` objects, so we can use that here

apollo-runtime/src/main/java/com/apollostack/android/GraphqlService.java
sav007: Would be nice if exclude this settings at all, as it increases friction of adoption this library, we already have big one: node.js. I guess we don't want to user to have a big doc how to set-up all these stuff

felipecsl:
@sav007 so based on our discussion, just to confirm: I'll generate an extra interface from `GraphQLCompiler.write()` that will have one method for each of the operations used, so that can be consumed by Retrofit`

felipecsl:
Actually this PR is getting too big, already, so let me do this in a separate PR.
@sav007 PTAL on this one, there's a few refactorings and renames. Most importantly, all the Data interfaces now extend `Query.Data`

apollo-api/src/main/java/com/apollostack/api/Query.java
```diff
@@ -2,9 +2,14 @@
 
 import java.util.List;
 
-public interface GraphQLQuery {
-
+/** TODO */
```
sav007: What this TODO is for?

apollo-runtime/src/main/java/com/apollostack/android/Apollo.java
```diff
@@ -0,0 +1,36 @@
+package com.apollostack.android;
+
+import retrofit2.Retrofit;
+
+import static com.apollostack.android.Utils.checkNotNull;
+
+/** TODO */
+public final class Apollo {
```
sav007: So you are going to refactor this class in next PR?

sav007:
So except `GraphQLQuery` and `Apollo` classes, that I assume are going to be refactored in next PR, looks great to me

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/GraphQlType.kt
```diff
@@ -1,55 +1,56 @@
...
+import com.apollostack.compiler.ClassNames
+import com.apollostack.compiler.Annotations
 import com.apollostack.compiler.normalizeTypeName
 import com.squareup.javapoet.ClassName
 import com.squareup.javapoet.ParameterizedTypeName
 import com.squareup.javapoet.TypeName
 
-sealed class GraphQlType(val isOptional: Boolean) {
-  class GraphQlString(isOptional: Boolean) : GraphQlType(isOptional)
+sealed class GraphQLType(val isOptional: Boolean) {
```
sav007: File name should be renamed as well

apollo-api/src/main/java/com/apollostack/api/Query.java
felipecsl: for adding the javadocs

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphqlCompiler.kt
```diff
@@ -5,22 +5,18 @@ import com.squareup.javapoet.JavaFile
 import com.squareup.moshi.Moshi
 import java.io.File
 
-open class GraphqlCompiler {
+open class GraphQLCompiler {
```
felipecsl: good catch

PR #52, opened 2016-12-15
add apollo-codgen gradle tasks

### LINKED ISSUE #43: [Gradle Plugin] Create gradle task to fire Apollo compiler
The gradle task should read the `.graphql` files and use `apollo-codegen` to translate them into IR files.
Once we have the IR, we can use `GraphQLCompiler` to generate the interfaces.

marwanad:
@martijnwalraven @felipecsl @sav007 please take a look. 

It's a lot of text but I just wanted to keep those points for the record. Also please check the comments on the diff.

ApolloPlugin now performs the following actions:
- Adds a dependency on `gradle-node`. This handles installing node nicely under `.gradle` and npm dependencies.

- The Apollo plugin starts off by adding the node plugin. This creates 5 gradle tasks (3 npm, 2 yarn) under the node group. Those won't have to be used.

- The Apollo plugin then creates an `ApolloCodegenInstall` task. This task basically executes `npm install apollo-codgen` and depends on `nodeSetup` task.

- Adds an `ApolloExtension` to every sourceSet in the project. This essentially allows to have configurable options through the Gradle DSL. For example, let the user decide whether they want to have an Rx Observable returned, etc. Here, it's used to allow the user to define custom sourceSet directories. 

- For every variant in the android project, register an ApolloIRGenTask. This would execute `apollo-codgen generate [...]`. Variants can have multiple sourceSets so that's the reason for having a List of ApolloExtensions in each ApolloIRGenTask. 

It also creates a top-level apolloIRGenTask which depends on the variant ones. Thus, running this will basically run the ones on all registered variants. 

Thus for `apollo-sample` project, you'll end up with the following tasks:

```
Apollo tasks
------------
generateApolloClasses
generateApolloIR
generateDebugApolloClasses - Generate Android classes for Debug GraphQL queries
generateDebugApolloIR - Generate IR files using apollo-codgen for Debug GraphQL queries
generateReleaseApolloClasses - Generate Android classes for Release GraphQL queries
generateReleaseApolloIR - Generate IR files using apollo-codgen for Release GraphQL queries
installApolloCodegen - Runs npm install for apollo-codegen

Node tasks
----------
nodeSetup - Download and install a local node/npm version.
npmInstall - Install node packages from package.json.
npmSetup - Setup a specific version of npm to be used by the build.
yarn - Install node packages using Yarn.
yarnSetup - Setup a specific version of Yarn to be used by the build.
```

### Running the snapshot jar.

For now, to run on the `apollo-sample` project:

1. Add the following under the apollo-sample gradle build script:
```
buildscript {
    // needed for apollo-compiler dependencies
    mavenCentral()
    jcenter()
    maven { url "https://jitpack.io" }
  repositories { flatDir name: 'libs', dirs: "../apollo-gradle-plugin/build/libs" }
}
  dependencies {
    classpath project(':apollo-compiler')
    classpath 'test-plugin:apollo-gradle-plugin:0.0.1-SNAPSHOT'
    classpath dep.gradleNodePlugin
}
apply plugin: 'com.apollostack.android'
```
2. Create a `src/main/graphql/` folder with a schema.json and some .graphql query files. You can find some [here](https://github.com/apollostack/apollo-codegen/tree/master/test/starwars). (Ensure that no naming conflicts exist in the files you're using from apollo-codgen)

3. Run `./gradlew apollo-sample:generateApolloIR`. This has a dependency on the node install task so it should install all dependencies if they don't exist.

Discussion Points:

### ApolloClassgenTask

I think we should add some sort of a Validator class that ships as part of the compiler that validates the JSON IR before attempting to generate the classes and handles exceptions nicely. This should be relatively simple since it would only validate JSON files. Open to discussions on that point.
### Testing
We might need some integration tests once all the components are ready. 

Closes #43


apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -0,0 +1,34 @@
...
+class ApolloExtension {
+  private final Project project
+  static String NAME = "apollo"
+  String graphqlPath
+  // Maps GraphQL query files' relative paths to a sourceSet directory
+  private Multimap<String, String> cachedFiles
+
+  ApolloExtension(Project project, String sourceSet) {
+    this.project = project
+    this.graphqlPath = "src/$sourceSet/graphql"
```
marwanad: For example, `src/main/graphql`, `src/debug/graphql`, etc

This sets the default but we can allow the user to have a custom one through extensions in the Gradle DSL.
This should be valid for all cases since the compiler enforces the `src/x/graphql` structure anyways.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,60 @@
...
+import com.google.common.collect.Sets
+import com.moowork.gradle.node.task.NodeTask
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.util.PatternSet
+
+public class ApolloIRgenTask extends NodeTask {
+  private static final String APOLLO_CODEGEN = "node_modules/apollo-codegen/lib/cli.js"
+  static final String NAME = "generate%sApolloIR"
+  List<ApolloExtension> config
```
marwanad: The ApolloIRgenTask is registered per variant and a variant can have multiple sourceSet directories.

For example, the debug variant on android has src/main and src/debug in its sourceSets. So the List of extensions was needed here.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,60 @@
...
+  String variant
+
+  @InputFiles
+  Collection<File> getInputFiles() {
```
marwanad: The extension stores the file paths as a multimap with keys corresponding to the relative path from the project root, and values are all `grapghql` files found.
This builds a Set containing those files. 
(We can go further and have it as a Set of absolute paths since that's the only use case here).

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,60 @@
...
+        entry.value.each { file ->
+          inputFiles.add(new File(entry.key, file))
+        }
+      }
+    }
+    return inputFiles
+  }
+
+  @Override
+  public void exec() {
```
marwanad: The following is equivalent running the following under `src/x/graphql` to:

`apollo-codgen **/**.graphql --schema schema.json --ouput VariantNameApi.json --target json`

A bunch of things to note here:

- This generates one IR for all the graphql files found under the directory. @martijnwalraven brought a good point about being able to reference fragments defined in other .graphql files.

- If we want to generate one per graphql, we'd have to basically run a shell script in a loop because gradle tasks custom can't instantiate other tasks. The solution would be a bit ugly since the class would no longer extend `NodeTask`.

I think a good solution to keep the process and maintain the package structure would be to have the relative path as a field in the IR. 

- As a consequence of the variants and sourceSet structure; this would actually do the following:
For every variant, grab all graphql files in its sourceSets. This means that for the debug variant, it would grab all graphql files under `src/main/graphql` and `src/debug/graphql`. 
Two things follow from this approach:

1. If duplicate operation names exist, apollo-codgen would fail. IMHO, it's safe to assume that won't happen in most of the cases since it defeats the purpose of having variants/flavors.

2. Where do we pick our schema file from? The flow right now would loop through the sourceSets directories and returns the first schema.json file it finds. IMO, it's also valid to assume that if you have multiple variants with different APIs, you're less likely to place your schema under the `main` sourceSet which is shared between variants.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloCodegenInstallTask.groovy
```diff
@@ -0,0 +1,20 @@
+package com.apollostack.android.gradle
+
+import com.moowork.gradle.node.npm.NpmTask
+
+class ApolloCodegenInstallTask extends NpmTask {
+  static final String NAME = "installApolloCodegen"
+  private static final String INSTALL_DIR = "node_modules/apollo-codegen"
+
+  public ApolloCodegenInstallTask() {
+    super()
```
felipecsl: you don't need this, do you?

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -0,0 +1,34 @@
+package com.apollostack.android.gradle
+
+import com.google.common.collect.ArrayListMultimap
+import com.google.common.collect.Multimap
+import org.gradle.api.Project
+import org.gradle.api.tasks.util.PatternSet
+
+class ApolloExtension {
+  private final Project project
+  static String NAME = "apollo"
```
felipecsl: final

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -0,0 +1,34 @@
...
+    this.project = project
+    this.graphqlPath = "src/$sourceSet/graphql"
+  }
+
+  Multimap<String, String> getFiles() {
+    if (cachedFiles != null) {
+      return cachedFiles
+    }
+    PatternSet patternSet = new PatternSet().include("**/*.graphql")
+    ArrayListMultimap<String, String> files = ArrayListMultimap.create()
```
felipecsl: nice use of multimap :)

felipecsl: Also just assign this to `cachedFiles` directly?

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -0,0 +1,34 @@
...
+    project.files(graphqlPath).getAsFileTree().matching(patternSet).visit { el ->
+      if (!el.directory) {
```
felipecsl: i think you can probably `filter { !it.directory }` to make it cleaner

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
felipecsl: Thanks for the thorough explanation @marwanad.
Bullet point 1 -> I think that's exactly what we want (one IR for all the graphql files)

```
I think a good solution to keep the flow neat and maintain the package structure would be to 
have the relative path as a field in the IR.
```
^ You mean because you need to be able to generate the java interfaces in the same directory hierarchy as the original .graphql files?

Bullet point 3, item i -> 👍 , do we need to check for that though, or just let apollo-codegen fail for us?
Bullet point 3, item ii -> Sounds like we should follow the same lookup pattern that android uses. First in the variant specific directory (`src/buildTypeFlavorName/graphql`), then build type (`src/buildType/graphql`) and finally main (`src/main/graphql`)

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,60 @@
...
+    File apolloScript = project.file(APOLLO_CODEGEN)
+    if (!apolloScript.isFile()) {
+      throw new GradleException("Apollo-codegen was not found in node_modules. Please run 'gradle ${ApolloCodegenInstallTask.NAME}")
```
felipecsl: nit: break this line, it's too long

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
marwanad: > you need to be able to generate the java interfaces in the same directory hierarchy as the original .graphql files?

yup, this.

> do we need to check for that though, or just let apollo-codegen fail for us?

I'd say let apollo-codgen do it. The error thrown is pretty descriptive.

```
There can only be one operation named "HeroName".
There can only be one operation named "HeroName".
There can only be one operation named "TwoHeroes".
There can only be one operation named "TwoHeroes".
Validation of GraphQL query document failed
```

For 3, the ordering we get is 
`[src/main/graphql, src/demo/graphql, src/debug/graphql, src/demoDebug/graphql]`

so reversing the lookup should work in that case.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloCodegenInstallTask.groovy
marwanad: nope not needed

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
marwanad: I need the call to `visit` because `relativePath` exists only on a `FileVisitDetails` object. A filter would imply a `File` type.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
martijnwalraven: Yes, I would definitely rely on `apollo-codegen` for validation. It can also give you file names and line numbers, which we use to show validation errors inline in Xcode. We should do something similar for Android Studio.

martijnwalraven:
I don't think we need to validate the IR, but we may want to think about compatibility between `apollo-codegen` versions. On iOS we've added a check to make sure the runtime Apollo framework is compatible with the version of `apollo-codegen` installed. For now, we just rely on the major and minor version to be the same (so all patch versions are considered to be compatible).

We probably won't have to introduce backwards incompatible changes on Android as often, because we don't do the code generation from `apollo-codegen`, but there could still be cases where we make changes to the tool interface or IR.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,64 @@
...
+    File apolloScript = project.file(APOLLO_CODEGEN)
+    if (!apolloScript.isFile()) {
+      throw new GradleException("Apollo-codegen was not found in node_modules. Please run 'gradle " +
+          "${ApolloCodegenInstallTask.NAME}")
+    }
+
+    def schemaPath = getSchemaPath()
+    setScript(apolloScript)
+    List<String> apolloArgs = ["generate"]
+    apolloArgs.addAll(getInputFiles().collect { project.file(it).absolutePath })
```
sav007: Do we run generation every time when project is compiled?
Shouldn't we check if any file has been changed and only then run regeneration of schema?

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
sav007: @martijnwalraven Is it possible to update the `apollo-codegen` to generate the one IR json per `*.graphql` file?

build.gradle
```diff
@@ -18,7 +18,9 @@ ext {
       kotlinStdLib: "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion",
       kotlinReflect: "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion",
       kotlinGradlePlugin: "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion",
+      gradleNodePlugin: "com.moowork.gradle:gradle-node-plugin:1.0.0",
       junit: 'junit:junit:4.12',
+      spock: 'org.spockframework:spock-core:0.7-groovy-2.0',
```
sav007: Nice, we can try to use the similiar `Speck` test framework for our unit tests in the compiler module

marwanad: yeah Spek is awesome

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,83 @@
...
+
+public class ApolloIRgenTask extends NodeTask {
+  private static final String APOLLO_CODEGEN = "node_modules/apollo-codegen/lib/cli.js"
+  private static final String DEFAULT_OUTPUT_DIR = "src/main/graphql"
+
+  static final String NAME = "generate%sApolloIR"
+  List<ApolloExtension> config
+  String variant
+
+  @OutputDirectory
```
marwanad: @sav007 updated the task to be incremental. Gradle would check the input files and output directory. If none of the inputs or outputs have changed, the task is skipped.

```
Executing external task 'generateDebugApolloIR'...
Configuration on demand is an incubating feature.
Incremental java compilation is an incubating feature.
:apollo-sample:nodeSetup UP-TO-DATE
:apollo-sample:installApolloCodegen UP-TO-DATE
:apollo-sample:generateDebugApolloIR UP-TO-DATE
```

The `ApolloClassgenTask` will be made incremental too.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
sav007: Oh I see it's independent tasks , but we are going to use `apt` later, I guess. So it will be part of compilation. Right @felipecsl ?

marwanad: The way I see the flow would be:

ApolloCodgenTask would be dependent on ApolloIRgen Task. Then we register ApolloClassgenTask as a java generating task. 

```
This will make the generate[Variant]Sources task depend on this task and add the new source folders as compilation inputs. The new source folders are also added to the model.
```
So, yes it will be part of compilation definitely.

marwanad:
@martijnwalraven Interesting. Since we'd be adding `apollo-runtime` as a dependency in `apollo-gradle`, we have more control over compatibility. 
We can modify the NpmInstall task to install a particular version of `apollo-codgen`. That way, everything moves together.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloCodeGenInstallTask.groovy
```diff
@@ -0,0 +1,37 @@
...
+    group = ApolloPlugin.TASK_GROUP
+    description = "Runs npm install for apollo-codegen"
+    File installDir = this.project.file(INSTALL_DIR)
+
+    setArgs(["install", "apollo-codegen@$APOLLOCODEGEN_VERSION"])
+
+    if (!installDir.exists()) {
+      installDir.mkdirs()
+    } else {
+      if (!apolloVersion().equals(APOLLOCODEGEN_VERSION)) {
```
marwanad: The way Gradle decides whether to skip a task or not is by looking at input and output dirs to see if they've changed.

Doing something like `npm install apollocodegen@VERSION --force` will not work if we want to update the npm package in later releases. 
So, this deletes the `node_modules` and reinstalls if `user version != APOLLOCODEGEN_VERSION`

apollo-compiler/src/test/kotlin/com/apollostack/compiler/GraphQLCompilerTest.kt
```diff
@@ -23,7 +24,7 @@ class GraphQLCompilerTest {
     val actual = actualFileFor("hero_name", "TestQuery")
     val expected = expectedFileFor("hero_name", "TestQuery")
 
-    compiler.write(irFileFor("hero_name", "TestQuery"))
+    compiler.write(irFileFor("hero_name", "TestQuery"), outputDir)
```
sav007: How about to extract this as util function:

```
private fun compile(pkg: String, queryFileName: String) {
  compiler.write(irFileFor("hero_name", "TestQuery"), outputDir)
}
```

So in tests it will be much shorter:

```
compile("hero_name", "TestQuery")
```

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+    return inputFiles
+  }
+
+  public void init(String variantName, List<ApolloExtension> extensionsConfig) {
+    variant = variantName
+    config = extensionsConfig
+    group = ApolloPlugin.TASK_GROUP
+    description = "Generate IR files using apollo-codgen for ${variant.capitalize()} GraphQL queries"
```
sav007: nit: `IR files -> IR file`, right? We generate one big IR file

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -0,0 +1,83 @@
...
+    apolloArgs.addAll(["--schema", "$outputDir.parent/schema.json",
+                       "--output", "$outputDir.absolutePath/${variant.capitalize()}API.json",
+                       "--target", "json"]);
+    setArgs(apolloArgs)
+    super.exec()
+  }
+
+  private String getSchemaPath() {
+    String schemaPath = ""
+    PatternSet patternSet = new PatternSet().include("**/schema.json")
```
sav007: Shouldn't it be configured? Maybe user should specify the path to schema file?

sav007:
@marwanad did you try to test it with apt? Maybe we need to update our sample project to use this plugin with apt, to see how it works?

marwanad:
@sav007 I haven't tried it with apt. I was leaving this part to the integration tests. But yeah I'll try it with apt and update the sample project or start writing integration tests as part of the `gradle-plugin` module.

sav007:
@marwanad Maybe it better to make it as separate PR.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
sav007: I think we can provide both options, if user doesn't specify path to schema file we will try to search. Or he can explicitly provide the path as config option.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloCodeGenInstallTask.groovy
```diff
@@ -0,0 +1,37 @@
...
+    getOutputs().dir(installDir)
+  }
+
+  String apolloVersion() {
+    String version = ""
+    File packageFile = project.file("${INSTALL_DIR}/package.json")
+    if (packageFile.isFile()) {
+      def input = new JsonSlurper().parseText(packageFile.text)
+      version = input.version
+    }
```
felipecsl: `else throw`? Otherwise we'll return an empty string, which is invalid

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -0,0 +1,35 @@
...
+  ApolloExtension(Project project, String sourceSet) {
+    this.project = project
+    this.graphqlPath = "src/$sourceSet/graphql"
+  }
+
+  Multimap<String, String> getFiles() {
+    if (cachedFiles != null) {
+      return cachedFiles
+    }
+    PatternSet patternSet = new PatternSet().include("**/*.${GraphQLCompiler.FILE_EXTENSION}")
```
felipecsl: this can be a constant too

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
felipecsl: Agreed, but not required for this PR, could be in a future one

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloPlugin.groovy
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private static ApolloIRGenTask createApolloIRGenTask(Project project, String name, List<ApolloExtension> config) {
+    String taskName = String.format(ApolloIRGenTask.NAME, name.capitalize())
+    ApolloIRGenTask task = project.tasks.create(taskName, ApolloIRGenTask)
+    task.init(name, config)
+    return task
+  }
+
+  private static ApolloClassGenTask createApolloClassGenTask(Project project, String name, List<ApolloExtension> conf) {
```
felipecsl: nit: break this line, too long

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloCodeGenInstallTask.groovy
marwanad: If `package.json` doesn't exist, this means that the user has deleted the `package.json` file so we still want to force an install, without throwing a `GradleException`.

We can maybe throw an Exception here and catch in the constructor block above, if that's what you mean.

felipecsl: Oh, I see, that's actually a valid use case. I think I'd rather return `null` then instead of empty string. No need to throw an exception then

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloExtension.groovy
```diff
@@ -7,29 +7,39 @@ import org.gradle.api.Project
...
+  static final String NAME = "apollo"
+
   private final Project project
-  static String NAME = "apollo"
-  String graphqlPath
+  private final String sourceSet
   // Maps GraphQL query files' relative paths to a sourceSet directory
   private Multimap<String, String> cachedFiles
 
+  // Configurable extension params
```
marwanad: Those extensions are per sourceSet, can be configured as follows:

```
  sourceSets.main {
    apollo {
      graphQLPath = 'src/somethingElse'
      schemaFile = 'src/something/aSchemaFile.json'
    }
  }
```
The paths are assumed to be relative to the project directory.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -10,20 +10,22 @@ import org.gradle.api.tasks.util.PatternSet
...
-  /** Output directory for the generated IR, default to src/main/graphql **/
+  /** Output directory for the generated IR, defaults to src/main/graphql **/
   File outputDir
 
   @InputFiles
   Collection<File> getInputFiles() {
     Set<File> inputFiles = Sets.newHashSet()
     config.each { ext ->
-      ext.getFiles().asMap().entrySet().each { entry ->
+      ext.getFiles(ext.graphQLPath ?: "src/${ext.getSourceSet()}/graphql").asMap().entrySet().each { entry ->
```
marwanad: If a user specified a `graphQLPath` for a sourceSet, then this would take precedence over `src/sourceSetName/graphql`.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
...
-    outputDir = new File("${getSchemaPath()}/generatedIR")
+
+    possibleGraphQLPaths = buildPossibleGraphQLPaths()
+    outputDir = project.file("src/$variant/graphql/generatedIR")
   }
 
   @Override
   public void exec() {
     File apolloScript = project.file(APOLLO_CODEGEN)
+    File schemaFile = userProvidedSchemaFile() ?: searchForSchemaFile()
```
marwanad: `userProvidedSchemaFile` also takes precedence here

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
     variant = variantName
     config = extensionsConfig
     group = ApolloPlugin.TASK_GROUP
-    description = "Generate IR files using apollo-codgen for ${variant.capitalize()} GraphQL queries"
+    description = "Generate an IR file using apollo-codgen for ${variant.capitalize()} GraphQL queries"
     dependsOn(ApolloCodeGenInstallTask.NAME)
-    outputDir = new File("${getSchemaPath()}/generatedIR")
+
+    possibleGraphQLPaths = buildPossibleGraphQLPaths()
+    outputDir = project.file("src/$variant/graphql/generatedIR")
```
marwanad: We can no longer depend on user provided options to decide the output directory. Because, we have the `src/main/graphql` check in the compiler, so compilation would fail if we generate the IR in a violating directory.

The IR will always be generated under `src/VariantName/graphql/generatedIR`. 

**The variant name is the build variant source set which is the product flavor plus build type**

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
...
-    if (!new File("$outputDir.parent/schema.json").isFile()) {
+
+    if (!schemaFile.isFile()) {
       throw new GradleException("Couldn't find a schema file. Please ensure a valid schema.json files exists in the " +
           "sourceSet directory")
     }
 
     setScript(apolloScript)
     List<String> apolloArgs = ["generate"]
+    println "sending request with schema path ${schemaFile.absolutePath}"
```
marwanad: remove

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
...
                        "--target", "json"]);
     setArgs(apolloArgs)
     super.exec()
   }
 
-  private String getSchemaPath() {
-    String schemaPath = ""
-    PatternSet patternSet = new PatternSet().include("**/schema.json")
-    for (String path : config.reverse().graphqlPath) {
+  private File userProvidedSchemaFile() {
```
marwanad: So this is interesting. The IRGen task is per build type variant. But extensions are per sourceSet. A built type variant will have multiple configurable sourceSet extensions.
Thus, a user can provide multiple schemaFiles. We do a check for that and throw an Exception.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
...
+        return schemaFile
+      }
+    }
+    return project.file("$DEFAULT_OUTPUT_DIR/schema.json")
+  }
+
+  private List<String> buildPossibleGraphQLPaths() {
+    List<String> graphQLPaths = new ArrayList<>();
+    config.each { ext ->
+      graphQLPaths.add("src/${ext.getSourceSet()}/graphql")
```
marwanad: Those are the default sourceSets returned in the reverse lookup order.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRgenTask.groovy
```diff
@@ -36,48 +38,79 @@ public class ApolloIRGenTask extends NodeTask {
...
+      if (schemaFiles.size() > 1) {
+        throw new IllegalArgumentException("More than two schema files were specified for the build variant $variant." +
+            " Please ensure that only one schema field is specified for $variant's source sets")
+      }
+      schemaFile = project.file(schemaFile.get(0))
+    }
+    return schemaFile
+  }
+
+  private File searchForSchemaFile() {
```
marwanad: This is where you end up if a `schemaFile` wasn't provided in the sourceSet extension. 

This searches the `possibleGrapQLPaths` list which contains the default sourceSets paths: `src/sourceSet/graphql` plus any user provided `graphQLPath` extensions

marwanad:
@sav007 @felipecsl addressed the comments above and allowed for a configurable `schemaPath` which led to some changes in how the lookup for files works, etc.

I left the comments [here](https://github.com/apollostack/apollo-android/pull/52/commits/e0d938bada80945a7e8f2395a5ddf4bbc028eedb) and [this](https://github.com/apollostack/apollo-android/pull/52/commits/e0d938bada80945a7e8f2395a5ddf4bbc028eedb#r93240126) basically sums it up.

I feel that this should be good enough for the scope of this PR, we can add the integration tests and sample-app changes in another PR.

sav007:
@marwanad sorry my PR was the first :)

marwanad:
@felipecsl any other comments/concerns?

PR #53, opened 2016-12-15
Update query code generation

### LINKED ISSUE #51: [Code Generation] Change Query.operationDefinition to return the string with joined query source, fragment sources and variables

sav007:
Refactor `com.apollostack.api.Query`
Update how query been generated to comply with `Query` api

Example of generated code with arguments: https://github.com/apollostack/apollo-android/blob/f926d22c72904d2962ba68e7880c74a3d0ffe03c/apollo-compiler/src/test/graphql/com/example/simple_arguments/TestQueryExpected.java

Example of generated code with fragments: https://github.com/apollostack/apollo-android/blob/f926d22c72904d2962ba68e7880c74a3d0ffe03c/apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/TestQueryExpected.java

API usage example:
```
final Query query = new TestQuery(
        new TestQuery.Variables.Builder()
            .episode(Episode.JEDI)
            .includeName(true)
            .build()
    );
```

To get access to operation definition source and variables map for serialization purpose:
```
String definition = query.operationDefinition();
Map<String, Object> varialbles = query.variables().toMap();
```

Closes #51 

### DIFF
--- a/apollo-api/src/main/java/com/apollostack/api/Query.java
+++ b/apollo-api/src/main/java/com/apollostack/api/Query.java
@@ -6,7 +6,7 @@ import java.util.Map;
 /** TODO */
 public interface Query<V extends Query.Variables> {
   /** TODO */
-  String operationDefinition();
+  String queryDocument();
 
   /** TODO */
   V variables();
--- a/apollo-api/src/main/java/com/apollostack/api/Query.java
+++ b/apollo-api/src/main/java/com/apollostack/api/Query.java
@@ -1,17 +1,34 @@
 package com.apollostack.api;
 
-import java.util.List;
+import java.util.Collections;
 import java.util.Map;
 
 /** TODO */
-public interface Query {
+public interface Query<V extends Query.Variables> {
   /** TODO */
   String operationDefinition();
+
   /** TODO */
-  List<String> fragmentDefinitions();
-  /** TODO */
-  Map<String, Object> variableDefinitions();
+  V variables();
+
   /** TODO */
   interface Data {
   }
+
+  /** TODO */
+  abstract class Variables {
+    protected final Map<String, Object> data;
+
+    protected Variables(Map<String, Object> data) {
+      this.data = Collections.unmodifiableMap(data);
+    }
+
+    /** TODO */
+    public Map<String, Object> toMap() {
+      return data;
+    }
+  }
+
+  Variables EMPTY_VARIABLES = new Variables(Collections.<String, Object>emptyMap()) {
+  };
 }
--- a/apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/TestQueryExpected.java
+++ b/apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/TestQueryExpected.java
@@ -1,17 +1,16 @@
 package com.example.fragment_with_inline_fragment;
 
 import com.apollostack.api.Query;
-import java.lang.Object;
 import java.lang.Override;
 import java.lang.String;
+import java.lang.StringBuilder;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-public final class TestQuery implements Query {
+public final class TestQuery implements Query<Query.Variables> {
   public static final String OPERATION_DEFINITION = "query TestQuery {\n"
       + "  hero {\n"
       + "    __typename\n"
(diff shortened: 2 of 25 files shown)


martijnwalraven:
Great work! A few comments about the generated code:

- I think you don't need to include the fragment source in `FRAGMENT_DEFINITIONS`, because you can put this in the fragment itself and share it between operations. This is what we do in Swift:

```swift
  public static let operationDefinition =
    "query HeroDetailsWithFragment($episode: Episode) {" +
    "  hero(episode: $episode) {" +
    "    __typename" +
    "    ...HeroDetails" +
    "  }" +
    "}"
  public static let queryDocument = operationDefinition.appending(HeroDetails.fragmentDefinition)
```

- I think having an instance method called `operationDefinition` is a bit confusing, because this returns the complete query document (so the operation definition + the fragment definitions). Also, this does not depend on the variables, so you don't need to build the string in the constructor, and can compute it statically. If we can't make `queryDocument` a static method (because interfaces < Java 8 don't support this), maybe the instance method could return the static field?

- Is the reason for having a separate `Variables` class and builder that Java doesn't have keyword arguments? So passing variables individually to the constructor could be confusing? In Swift we'd do something like: 

```swift
HeroAndFriendsNamesQuery(episode: .jedi)
CreateReviewForEpisodeMutation(episode: .jedi, review: ReviewInput(stars: 5, commentary: "This is a great movie!"))
```

sav007:
@martijnwalraven 

1. >I think you don't need to include the fragment source in FRAGMENT_DEFINITIONS, because you can put this in the fragment itself and share it between operations

Oh yeah you are right. Thanks, will change that

2. >Also, this does not depend on the variables, so you don't need to build the string in the constructor, and can compute it statically

we can't make it static as during serialization to json we will work with generic `Query` interface only, and I believe Android still doesn't supporting Java8 interface default methods so we have to have instance method. So you want to have `queryDocument` that will return joined query and fragment definition, right?

3. > Is the reason for having a separate Variables class and builder that Java doesn't have keyword arguments?

Java doesn't support named arguments, and if we have more than 3 args then it will be messy, especially when they are all boolean.

felipecsl:
I don't understand why you're building the `query` field in the constructor with `StringBuilder` if it could be a `static final` field. Everything is constant, why build it dynamically?

martijnwalraven:
And then would it work to add an instance method `queryDocument` that returns the static field as I suggested (to still be able access it through the `Query`interface under Java < 8)?

sav007:
@felipecsl yes it could be but then how in converter (that takes only `Query` interface as parameter) you get access to this static field if it defined only in the generated subclass? Converter does know anything about the generated models, it works only with `Query` interface.

@martijnwalraven yes `queryDocument` will return the static field, that will be join of query and fragments definitions

martijnwalraven:
@sav007: That's why I suggested returning the static field through the generated instance method

sav007:
@martijnwalraven @felipecsl updated

Example of generated code with arguments: https://github.com/apollostack/apollo-android/blob/f926d22c72904d2962ba68e7880c74a3d0ffe03c/apollo-compiler/src/test/graphql/com/example/simple_arguments/TestQueryExpected.java

Example of generated code with fragments: https://github.com/apollostack/apollo-android/blob/f926d22c72904d2962ba68e7880c74a3d0ffe03c/apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/TestQueryExpected.java

apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/TestQueryExpected.java
```diff
@@ -21,39 +17,23 @@
...
-        + "    }\n"
-        + "  }\n"
-        + "  ... on Droid {\n"
-        + "    name\n"
-        + "    primaryFunction\n"
-        + "  }\n"
-        + "}"
-  ));
+  public static final String QUERY_DOCUMENT = OPERATION_DEFINITION + "\n"
+   + HeroDetails.FRAGMENT_DEFINITION;
```
felipecsl: much better now 👍

apollo-api/src/main/java/com/apollostack/api/Query.java
```diff
@@ -1,17 +1,34 @@
 package com.apollostack.api;
 
-import java.util.List;
+import java.util.Collections;
 import java.util.Map;
 
 /** TODO */
-public interface Query {
+public interface Query<V extends Query.Variables> {
```
felipecsl: I still think we don't need this `V` generic type on `Query`. 
First because you don't always have variable, so a lot of times it's just `Query<Query.Variables>` which is not super helpful. 
Second, I think we can just roll with `Query.Variables` all the time, right? Why does the `variables()` method need need to be generic?

sav007: @felipecsl 
1. I will argue, especially if query paginated and you will have to pass cursor, limit and as example of Shopify schema I would say variables will be used 80% of times
2. `variables()` is not generic it returns `V extends Query.Variables`, so in case if query has variable you will get statically typed variables, yes internally we don't care but from user perspective would be nice to have, if user decides to get variables for query

felipecsl: ok fair enough, I haven't used GraphQL in production yet to know how often you use variables, but sounds like it's something quite useful

sav007:
@martijnwalraven @felipecsl any objections to merge?

PR #59, opened 2016-12-21
[Runtime] Adds Moshi converter module

felipecsl:
~~Just as a reminder, the converter implementation should be pluggable. We shouldn't have a direct dependency on Moshi (or RxJava, for example), but this should work for now.~~ 

**Update:** I've created a `apollo-converters` directory with a `converter-moshi` module inside, in the same fashion as Retrofit does.
The implementation is now self contained and opt-in.

~~We'll need to change the `Query.Data` interfaces to classes as well, since Moshi isn't able to deserialize the JSON into interfaces.
If everyone is okay with that, I'll go ahead and make the change to the compiler as part of this PR.
Right now, I've manually modified `HeroDetails.java` in the runtime test code to be able to make the test pass.~~

**Update**: I actually don't think this is a good idea aymore. Let's table this for now and I'll just implement the interface in the tests directly.

## Pending taks:

- [x] Assert on the request format in the MockWebServer tests to make sure the body is being sent correctly
- [x] Handle variables
- [x] Handle error scenarios in `ApolloResponseBodyConverter`


apollo-runtime/src/test/java/com/apollostack/android/HeroDetails.java
```diff
@@ -0,0 +1,47 @@
...
+    return variables;
+  }
+
+  public static class Data implements GraphQLQuery.Data {
+    @Nullable AllPeople allPeople;
+
+    static class AllPeople {
+      @Nullable List<People> people;
+
+      static class People {
```
felipecsl: I've manually modified these types to be static classes instead of interfaces, so the test would pass (as described in the PR summary).
I can go ahead and change the compiler code to do this if you guys are 👍

apollo-runtime/src/main/java/com/apollostack/android/ApolloResponseBodyConverter.java
```diff
@@ -0,0 +1,32 @@
...
+
+  ApolloResponseBodyConverter(JsonAdapter<?> adapter) {
+    this.adapter = adapter;
+  }
+
+  @Override public T convert(ResponseBody value) throws IOException {
+    try {
+      JsonReader reader = JsonReader.of(value.source());
+      reader.beginObject();
+      if (!reader.nextName().equals("data")) {
```
marwanad: Might be outside this PR's scope but worth mentioning: In some cases, the data field may [not](http://facebook.github.io/graphql/#sec-Data) exist.  Also, an errors array may be returned.

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
```diff
@@ -0,0 +1,75 @@
...
+        .build();
+    service = retrofit.create(Service.class);
+  }
+
+  @Test public void simpleQuery() throws IOException {
+    server.enqueue(mockResponse("src/test/graphql/simpleQueryResponse.json"));
+    Call<HeroDetails.Data> call = service.heroDetails(new PostBody(HeroDetails.OPERATION_DEFINITION));
+    Response<HeroDetails.Data> response = call.execute();
+    HeroDetails.Data body = response.body();
+    List<String> actual = FluentIterable.from(body.allPeople.people)
```
marwanad: TIL`FluentIterable`

apollo-runtime/src/main/java/com/apollostack/android/ApolloResponseBodyConverter.java
felipecsl: great catch, I'll add a TODO for it

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
felipecsl: Guava has all kinds of hidden gems :)

apollo-runtime/src/main/java/com/apollostack/android/ApolloConverterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public ApolloConverterFactory(Moshi moshi) {
+    this.moshi = moshi;
+    moshiConverterFactory = MoshiConverterFactory.create(moshi);
+  }
+
+  @Override public Converter<ResponseBody, ?> responseBodyConverter(Type type, Annotation[] annotations,
+      Retrofit retrofit) {
+    JsonAdapter<?> adapter = moshi.adapter(type);
+    return new ApolloResponseBodyConverter<>(adapter);
```
sav007: I guess we need to check if type doesn't subclass  of `GraphQLOperation.Data` then return null. 
So our adapter will be limited to parse only responses from graphql.

From java doc of retrofit: 
>/**
     * Returns a {@link Converter} for converting an HTTP response body to {@code type}, or null if
     * {@code type} cannot be handled by this factory. This is used to create converters for
     * response types such as {@code SimpleResponse} from a {@code Call<SimpleResponse>}
     * declaration.
     */

apollo-runtime/src/main/java/com/apollostack/android/ApolloConverterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override public Converter<?, RequestBody> requestBodyConverter(Type type, Annotation[] parameterAnnotations,
+      Annotation[] methodAnnotations, Retrofit retrofit) {
+    return moshiConverterFactory.requestBodyConverter(type, parameterAnnotations, methodAnnotations, retrofit);
```
sav007: The same as above

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
```diff
@@ -0,0 +1,75 @@
...
+        "Wilhuff Tarkin", "Chewbacca", "Han Solo", "Greedo", "Jabba Desilijic Tiure", "Wedge Antilles",
+        "Jek Tono Porkins", "Yoda", "Palpatine", "Boba Fett", "IG-88", "Bossk", "Lando Calrissian", "Lobot", "Ackbar",
+        "Mon Mothma", "Arvel Crynyd", "Wicket Systri Warrick", "Nien Nunb", "Qui-Gon Jinn", "Nute Gunray",
+        "Finis Valorum", "Padmé Amidala", "Jar Jar Binks", "Roos Tarpals", "Rugor Nass", "Ric Olié", "Watto", "Sebulba",
+        "Quarsh Panaka", "Shmi Skywalker", "Darth Maul", "Bib Fortuna", "Ayla Secura", "Ratts Tyerel", "Dud Bolt",
+        "Gasgano", "Ben Quadinaros", "Mace Windu", "Ki-Adi-Mundi", "Kit Fisto", "Eeth Koth", "Adi Gallia",
+        "Saesee Tiin", "Yarael Poof", "Plo Koon", "Mas Amedda", "Gregar Typho", "Cordé", "Cliegg Lars",
+        "Poggle the Lesser", "Luminara Unduli", "Barriss Offee", "Dormé", "Dooku", "Bail Prestor Organa", "Jango Fett",
+        "Zam Wesell", "Dexter Jettster", "Lama Su", "Taun We", "Jocasta Nu", "R4-P17", "Wat Tambor", "San Hill",
+        "Shaak Ti", "Grievous", "Tarfful", "Raymus Antilles", "Sly Moore", "Tion Medon"));
```
sav007: Should we limit this just to couple (~5) values :)

apollo-runtime/src/main/java/com/apollostack/android/ApolloResponseBodyConverter.java
```diff
@@ -0,0 +1,37 @@
...
+      reader.beginObject();
+      /* TODO handle other possible situations as described in the spec: http://facebook.github.io/graphql/#sec-Data
+       * If an error was encountered before execution begins, the data entry should not be present in the result.
+       * If an error was encountered during the execution that prevented a valid response, the data entry in the
+       * response should be null.
+       */
+      if (!reader.nextName().equals("data")) {
+        throw new IllegalStateException("Malformed input JSON. Expected 'data' object");
+      }
+      T responseBody = (T) adapter.fromJson(reader);
```
sav007: I think we need to add todo here, as parsing fragments are not supported yet

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
felipecsl: Yeah this is too much 😛

apollo-runtime/src/main/java/com/apollostack/android/ApolloResponseBodyConverter.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override public T convert(ResponseBody value) throws IOException {
+    try {
+      JsonReader reader = JsonReader.of(value.source());
+      reader.beginObject();
+      /* TODO handle other possible situations as described in the spec: http://facebook.github.io/graphql/#sec-Data
+       * If an error was encountered before execution begins, the data entry should not be present in the result.
+       * If an error was encountered during the execution that prevented a valid response, the data entry in the
+       * response should be null.
+       */
+      if (!reader.nextName().equals("data")) {
```
sav007: Should we add in this PR basic parsing of error response?

Example:
```
{
  "errors": [
    {
      "message": "Field 'pageInfo1' doesn't exist on type 'CollectionConnection'",
      "locations" 
....
```

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
sav007: I hope it is not transitive dependency, as it will bring if I don't mistake +1400 methods
We should be careful with libs we include with runtime module, we don't want to inflate users dex file

apollo-runtime/src/main/java/com/apollostack/android/ApolloResponseBodyConverter.java
felipecsl: sounds good, will do

apollo-runtime/src/test/java/com/apollostack/android/ConverterTest.java
felipecsl: Guava is a transitive dependency of `truth`, which is test only. We're fine ;)

felipecsl:
Summary updated

apollo-converters/moshi/src/test/java/com/apollostack/android/ConverterTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Nullable @Override public AllPeoplePOJO allPeople() {
+      return allPeople;
+    }
+  }
+
+  @Rule public final MockWebServer server = new MockWebServer();
+
+  @Before public void setUp() {
+    Moshi moshi = new Moshi.Builder()
+        .add(new PostBodyJsonAdapterFactory())
```
sav007: Can we do this in ApolloConverterFactory, to be transparent for user about our internal jsonAdapter?

apollo-converters/moshi/src/main/java/com/apollostack/android/PostBodyJsonAdapterFactory.java
```diff
@@ -0,0 +1,20 @@
...
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+
+import static com.squareup.moshi.Types.getRawType;
+
+public class PostBodyJsonAdapterFactory implements JsonAdapter.Factory {
```
sav007: I would like to hide this class from user knowledge

apollo-converters/moshi/src/main/java/com/apollostack/android/PostBodyJsonAdapter.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.apollostack.api.GraphQLOperation;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+
+import java.io.IOException;
+
+class PostBodyJsonAdapter<T extends GraphQLOperation.Variables> extends JsonAdapter<PostBody<T>> {
```
sav007: Just curious why do we need to create custom moshi adapter, can't we just use the standard moshi?

apollo-converters/moshi/src/test/java/com/apollostack/android/ConverterTest.java
felipecsl: hmm yeah I think that can be done. Good point.

apollo-converters/moshi/src/main/java/com/apollostack/android/ApolloRequestBodyConverter.java
```diff
@@ -0,0 +1,25 @@
...
+
+class ApolloRequestBodyConverter implements Converter<PostBody, RequestBody> {
+  private static final MediaType MEDIA_TYPE = MediaType.parse("application/json; charset=UTF-8");
+  private final JsonAdapter<PostBody> adapter;
+
+  ApolloRequestBodyConverter(JsonAdapter<PostBody> adapter) {
+    this.adapter = adapter;
+  }
+
+  @Override public RequestBody convert(PostBody value) throws IOException {
```
sav007: Here can we user GraphQLOperation instead PostBody as a value parameter?

apollo-api/src/main/java/com/apollostack/android/PostBody.java
```diff
@@ -0,0 +1,24 @@
+package com.apollostack.android;
+
+import com.apollostack.api.GraphQLOperation;
+import com.apollostack.api.GraphQLQuery;
+
+import javax.annotation.Nullable;
+
+public class PostBody<T extends GraphQLOperation.Variables> {
```
sav007: So we will hide this class from user

felipecsl: Unfortunately Moshi can't handle it the `variables` serialization. I think it's because of this issue: https://github.com/square/moshi/issues/133

apollo-converters/moshi/src/main/java/com/apollostack/android/PostBodyJsonAdapter.java
felipecsl: Explained above

apollo-converters/moshi/src/main/java/com/apollostack/android/ApolloRequestBodyConverter.java
felipecsl: good idea, I think we can. Let me try it out

apollo-api/src/main/java/com/apollostack/android/PostBody.java
sav007: I see, ok can we then make this class package protected and create it in ApolloRequestBodyConverter and pass it to JsonAdapter?

felipecsl:
@sav007 aaand `PostBody` is gone 🌮

apollo-converters/moshi/src/test/java/com/apollostack/android/HeroDetailsDataPOJO.java
```diff
@@ -0,0 +1,41 @@
+package com.apollostack.android;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+class HeroDetailsDataPOJO implements HeroDetails.Data {
```
sav007: Did you feel the pain to implement the Data classes manually ? :)

sav007:
Looks awesome

apollo-converters/moshi/src/test/java/com/apollostack/android/HeroDetailsDataPOJO.java
sav007: I'm more and more confident to make them as classes

felipecsl: haha yeah it's pretty mechanical and boring, but I think this should be opt-in. That is, we still generate the interfaces as usual and if you want, we can also generate POJOs or otherwise you can use your own stuff like AutoValue for example

felipecsl:
Alright this is ready to go, PTAL 👀

apollo-converters/moshi/src/test/java/com/apollostack/android/HeroDetailsWithArgument.java
```diff
@@ -0,0 +1,73 @@
+package com.apollostack.android;
+
+import com.apollostack.api.GraphQLOperation;
+import com.apollostack.api.GraphQLQuery;
+
+import javax.annotation.Nullable;
+
+public final class HeroDetailsWithArgument implements GraphQLQuery<HeroDetailsWithArgument.Variables> {
```
sav007: Shouldn't this class be generated by our gradle plugin?

felipecsl: I actually haven't tried the plugin yet, but in theory, yes. I'll give it a try

felipecsl:
@sav007 I don't want to add the plugin as a dependency to apollo-runtime so I'll park this for now, lets revisit later

PR #65, opened 2016-12-21
refactor the compiler tests into parameterized tests

### LINKED ISSUE #57: [Refactoring] Remove code duplication from GraphQLCompilerTest
All Junit tests in GraphQLCompilerTest follow the same pattern:
1. check actual generated file with expected
2. compile

A lot of code duplication. We can do better, define only test fixtures (probably map of actual / expected file paths) and run tests against them

marwanad:
Converts the codegeneration tests to a Parameterized test class. 
To add new tests, simply drop the IR and expected files under `src/test/graphql/com/example/some_package` (This includes fragments and enums too).

Closes #57


sav007:
Jeez, this is cool !!!

PR #70, opened 2016-12-22
Apply Apollo gradle plugin to sample app

felipecsl:
Still a few details to iron out but it worked pretty well!

### DIFF
--- a/apollo-sample/build.gradle
+++ b/apollo-sample/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     // TODO: Temporary workaround while we don't have the plugin snapshot on Sonatype.
     // This is a shadow (fat) jar with all the transitive dependencies.
     // In order to generate a new one for development, run:
-    // ./gradlew apollo-gradle-plugin:assemble
+    // ./gradlew apollo-gradle-plugin:shadowJar
     // copy the JAR from apollo-gradle-plugin/build/libs to apollo-sample/libs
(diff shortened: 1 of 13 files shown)


apollo-sample/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
felipecsl: Yeah I think so, would save an extra step

apollo-sample/build.gradle
```diff
@@ -1,6 +1,17 @@
...
+    classpath dep.gradleNodePlugin
     classpath dep.androidPlugin
+    classpath dep.javaPoet
+    classpath dep.kotlinStdLib
+    classpath dep.moshi
+    classpath dep.pluralizer
+    classpath dep.jsr305
+    classpath files(project(':apollo-api').jar.outputs.files.getFiles())
+    classpath files(project(':apollo-compiler').jar.outputs.files.getFiles())
+    classpath files(project(':apollo-gradle-plugin').jar.outputs.files.getFiles())
```
felipecsl: this doens't work in CI and causes weird problems. I'll just build a fat jar for now and use that instead until we get artifacts on sonatype snapshots

apollo-sample/package.json
felipecsl: Created #71

felipecsl:
@marwanad this is ready to go now

PR #73, opened 2016-12-23
[Sample] Make the sample display the response

felipecsl:
It actually works 😄 

## GraphiQL:

image

## Android emulator: 

image

I had to make one change to `GraphQLID` to treat it as String instead of `long`. According to the spec:

```
The ID scalar type represents a unique identifier, often used to refetch an object or
 as the key for a cache. The ID type is serialized in the same way as a String; however, 
it is not intended to be human‐readable. While it is often numeric, it should always 
serialize as a String.
```

Annotating the interfaces with `@AutoValue` was actually not a lot of work and I think this is a totally acceptable use case. 
Going back to the debate about generating classes vs. interfaces, I think we should keep generating the interfaces as we are right now and generate POJOs as an opt-in, so we don't ruin the AutoValue use case.


sav007:
It's alive!!! Great job 👍

sav007:
@felipecsl but if change the code generation from interfaces to classes, this sample will be broken

apollo-sample/src/main/java/com/example/apollostack/sample/DroidDetailsData.java
```diff
@@ -0,0 +1,26 @@
...
+@AutoValue
+public abstract class DroidDetailsData implements DroidDetails.Data {
+  @Override @Nullable public abstract Species species();
+
+  public static JsonAdapter<DroidDetailsData> jsonAdapter(Moshi moshi) {
+    return new AutoValue_DroidDetailsData.MoshiJsonAdapter(moshi);
+  }
+
+  @AutoValue
+  public abstract static class Species implements Specy {
```
felipecsl: the pluralization lib erroneously singularized `species` to `specy`. I'm not sure but I think this word doesn't have a singular version

felipecsl:
@sav007 how hard do you think it would be to have the compiler optionally generate POJOs that implement the interfaces we have today. Maybe add a boolean argument to the `GraphQLCompiler#write()` method that tells it whether to also generate POJOs for the Data classes, so we can at least for now roll with what we have

marwanad:
🎉 🎉 🎉

apollo-sample/src/main/java/com/example/apollostack/sample/DroidDetailsData.java
marwanad: Seems odd that they have it here too: https://github.com/cesarferreira/kotlin-pluralizer/blob/322fdd6d0e3e3da1bf59657a4e64bda8acc79ad5/library/src/main/kotlin/com/cesarferreira/pluralize/Pluralize.kt#L87

Oh, just glancing at this line, https://github.com/cesarferreira/kotlin-pluralizer/blob/322fdd6d0e3e3da1bf59657a4e64bda8acc79ad5/library/src/main/kotlin/com/cesarferreira/pluralize/Pluralize.kt#L48

They use `contains` which case sensitive.

felipecsl: Ah dammit, good catch @marwanad, I'll create an issue for this

felipecsl: Created #77

PR #79, opened 2016-12-24
Introduce GraphQLReader to be used in pojo construction

sav007:
Introduce GraphQLReader to be used in pojo construction.


apollo-sample/src/main/java/com/example/apollostack/sample/PeopleFragment.java
```diff
@@ -0,0 +1,51 @@
...
+      + "  species {\n"
+      + "    name\n"
+      + "  }\n"
+      + "}";
+
+  private @Nullable String name;
+
+  private @Nullable Specy species;
+
+  public PeopleFragment(BufferedResponseStreamReader streamReader) {
```
sav007: Fragments we always instantiate from BufferedResponseStreamReader

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryResponse.java
```diff
@@ -0,0 +1,182 @@
...
+
+import com.apollostack.api.graphql.ResponseStreamReader;
+
+import java.io.IOException;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class TestQueryResponse {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponse.java
```diff
@@ -0,0 +1,198 @@
...
+import com.apollostack.api.graphql.Operation;
+import com.apollostack.api.graphql.ResponseStreamReader;
+
+import java.io.IOException;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class TestQueryWithFragmentResponse {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponse.java
```diff
@@ -0,0 +1,198 @@
...
+
+        public Edge(ResponseStreamReader streamReader) throws IOException {
+          while (streamReader.hasNext()) {
+            String nextName = streamReader.nextName();
+            if ("cursor".equals(nextName)) {
+              this.cursor = streamReader.nextString();
+            } else if ("node".equals(nextName)) {
+              this.node = streamReader.nextOptionalObject(new ResponseStreamReader.NestedReader<Node>() {
+                @Override public Node read(ResponseStreamReader reader) throws IOException {
+                  return new Node(new BufferedResponseStreamReader(reader));
```
sav007: For objects that define fragments we switch from stream reader to buffered reader

apollo-sample/src/main/java/com/example/apollostack/sample/PeopleFragment.java
```diff
@@ -0,0 +1,51 @@
+package com.example.apollostack.sample;
+
+import com.apollostack.api.graphql.BufferedResponseStreamReader;
+
+import java.lang.String;
+
+import javax.annotation.Nullable;
+
+public class PeopleFragment {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/PeopleFragment.java
```diff
@@ -0,0 +1,49 @@
+package com.example.apollostack.sample;
+
+import com.apollostack.api.graphql.BufferedResponseReader;
+
+import javax.annotation.Nullable;
+
+public class PeopleFragment {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/PeopleFragment.java
```diff
@@ -0,0 +1,49 @@
...
+      + "  species {\n"
+      + "    name\n"
+      + "  }\n"
+      + "}";
+
+  private @Nullable String name;
+
+  private @Nullable Specy species;
+
+  public PeopleFragment(BufferedResponseReader reader) {
```
sav007: For fragments we always must read from `BufferedResponseReader `

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponse.java
```diff
@@ -0,0 +1,198 @@
...
+
+        public Edge(ResponseStreamReader streamReader) throws IOException {
+          while (streamReader.hasNext()) {
+            String nextName = streamReader.nextName();
+            if ("cursor".equals(nextName)) {
+              this.cursor = streamReader.nextString();
+            } else if ("node".equals(nextName)) {
+              this.node = streamReader.nextOptionalObject(new ResponseStreamReader.NestedReader<Node>() {
+                @Override public Node read(ResponseStreamReader reader) throws IOException {
+                  return new Node(reader.toBufferedReader());
```
sav007: If we see that object defines fragments (in our case `Node` has fragment) we switch from stream parsing to buffered

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/ApolloResponseBodyConverter.java
```diff
@@ -0,0 +1,107 @@
...
+
+    return new Response<>(data, errors);
+  }
+
+  private Operation.Data readResponseData(ResponseStreamReader reader) throws IOException {
+    return reader.nextObject(new ResponseStreamReader.NestedReader<Operation.Data>() {
+      @Override public Operation.Data read(ResponseStreamReader reader) throws IOException {
+        //noinspection TryWithIdenticalCatches
+        try {
+          return (Operation.Data) ((Class<?>) type).getConstructor(ResponseStreamReader.class).newInstance(reader);
```
sav007: This reflection is needed as we need somehow to get access to constructor of Root response class. This code will be called only once per response.

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/BufferedSourceJsonReader.java
```diff
@@ -0,0 +1,1067 @@
...
+
+import java.io.EOFException;
+import java.io.IOException;
+
+import okio.Buffer;
+import okio.BufferedSource;
+import okio.ByteString;
+
+/** TODO add Modifications copyright **/
+final class BufferedSourceJsonReader extends JsonReader {
```
sav007: Was copied from moshi, wrapper around `okio.BufferedSource`.

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/JsonDataException.java
```diff
@@ -0,0 +1,47 @@
...
+ * the call to {@link JsonReader#nextBoolean} is made, a {@code JsonDataException} is thrown.
+ *
+ * <p>Exceptions of this type should be fixed by either changing the application code to accept
+ * the unexpected JSON, or by changing the JSON to conform to the application's expectations.
+ *
+ * <p>This exception may also be triggered if a document's nesting exceeds 31 levels. This depth is
+ * sufficient for all practical applications, but shallow enough to avoid uglier failures like
+ * {@link StackOverflowError}.
+ */
+public final class JsonDataException extends RuntimeException {
```
sav007: Was copied from moshi, required by `BufferedSourceJsonReader` wrapper around `okio.BufferedSource`.

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/JsonEncodingException.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.apollostack.converter.pojo;
+
+import java.io.IOException;
+
+/** Thrown when the data being parsed is not encoded as valid JSON. */
+public final class JsonEncodingException extends IOException {
```
sav007: Was copied from moshi, required by `BufferedSourceJsonReader` wrapper around `okio.BufferedSource`.

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/JsonReader.java
```diff
@@ -0,0 +1,404 @@
...
+ * JavaScript's only numeric type and very large values like {@code
+ * 9007199254740993} cannot be represented exactly on that platform. To minimize
+ * precision loss, extremely large values should be written and read as strings
+ * in JSON.
+ *
+ * <p>Each {@code JsonReader} may be used to read a single JSON stream. Instances
+ * of this class are not thread safe.
+ */
+/** TODO add Modifications copyright **/
+abstract class JsonReader implements Closeable {
```
sav007: Was copied from moshi, required by `BufferedSourceJsonReader` wrapper around `okio.BufferedSource`.

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/JsonScope.java
```diff
@@ -0,0 +1,79 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.apollostack.converter.pojo;
+
+/** Lexical scoping elements within a JSON reader or writer. */
+final class JsonScope {
```
sav007: Was copied from moshi, required by `BufferedSourceJsonReader` wrapper around `okio.BufferedSource`.

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryResponseData.java
```diff
@@ -0,0 +1,160 @@
...
+import com.apollostack.api.graphql.Operation;
+import com.apollostack.api.graphql.ResponseStreamReader;
+
+import java.io.IOException;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class TestQueryResponseData implements Operation.Data {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponseData.java
```diff
@@ -0,0 +1,174 @@
...
+import com.apollostack.api.graphql.Operation;
+import com.apollostack.api.graphql.ResponseStreamReader;
+
+import java.io.IOException;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class TestQueryWithFragmentResponseData implements Operation.Data {
```
sav007: Will be generated

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponseData.java
```diff
@@ -0,0 +1,174 @@
...
+
+      public Edge(ResponseStreamReader streamReader) throws IOException {
+        while (streamReader.hasNext()) {
+          String nextName = streamReader.nextName();
+          if ("cursor".equals(nextName)) {
+            this.cursor = streamReader.nextString();
+          } else if ("node".equals(nextName)) {
+            this.node = streamReader.nextOptionalObject(new ResponseStreamReader.NestedReader<Node>() {
+              @Override public Node read(ResponseStreamReader reader) throws IOException {
+                return new Node(reader.toBufferedReader());
```
sav007: For objects that define fragments we switch from stream reader to buffered reader

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/ApolloResponseBodyConverter.java
digitalbuddha: I have no problem with that. Mostly the cost is high when you need to reflect over dozens of field names.  One is in the ballpark of 30ms on slowest of devices.

sav007:
@martijnwalraven 
`ResponseStreamReader` will be used only for network responses and only if no fragment support.
`ResponseReader`  used for network response if it has fragments and for normalized cache.
Generated POJO code will have 2 constructors: one takes `ResponseStreamReader` another `ResponseReader`. So POJO can be constructed from network or cache, we need only call appropriate constructor.
`ResponseStreamReader` introduces efficient way of parsing network stream only, without having buffer so it different from `ResponseReader` API due to stream parsing specific.

martijnwalraven:
@sav007: Ah, I didn't realize you were planning on implementing both for every POJO. Having two different constructors would solve the first issue (being able to construct objects from either the network or the cache), but it doesn't solve the second one (interleaving parsing and caching).

So even if we want to go with supporting two different `Reader` interfaces, we would still need to add field information to the `ResponseStreamReader` API to be able to normalize responses.

This could definitely work, but it feels we would make things simpler if we could unify the interfaces. And it would also be more performant because we don't necessarily have to rely on buffering (if we make the object responsible for reading all fields, including the ones defined in fragments). What do you think of my proposals?

Maybe I'm complicating matters prematurely and we should just get something working first. But I'd prefer to get the design right before we start implementing it, especially because any future changes to the runtime API would also result in corresponding changes to the code generation logic.

sav007:
@martijnwalraven I see your point, so you mean `ResponseStreamReader` will be responsible to write response to the normalized cache, right?

If so, yes we will require to add field info. I will update PR and try to unify, make `ResponseStreamReader` extends `ResponseReader`, so they will share `read*` api calls but for streaming we still need `nextName`, `skipNext` apis.

martijnwalraven:
@sav007: Yep, the `ResponseStreamReader` should keep track of records as it parses the response.

If we're not unifying the semantics, it might make more sense to keep `ResponseStreamReader` and `ResponseStreamReader` separate for now. I feel `next...` better signals the fact that the order we call these methods in matters.

apollo-api/src/main/java/com/apollostack/api/graphql/ResponseReader.java
```diff
@@ -0,0 +1,42 @@
+package com.apollostack.api.graphql;
+
+import java.util.List;
+
+/** TODO java doc **/
+public interface ResponseReader {
```
felipecsl: Does this have to be public since all the methods are package-private?

felipecsl: Also it would be nice if you could break this PR down into multiple smaller ones if possible

apollo-api/src/main/java/com/apollostack/api/graphql/ResponseStreamReader.java
```diff
@@ -0,0 +1,57 @@
+package com.apollostack.api.graphql;
+
+import java.io.IOException;
+import java.util.List;
+
+/** TODO java doc **/
+public interface ResponseStreamReader {
```
felipecsl: same question

apollo-sample/build.gradle
```diff
@@ -31,8 +31,12 @@ android {
 
 dependencies {
   compile dep.appcompat
+  compile dep.jsr305
+  compile dep.retrofitMoshiConverter
```
felipecsl: you shouln't need this

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponseData.java
felipecsl: why is that? Sorry, I'm trying to catch up!

apollo-converters/pojo/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'java'
+
+targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_7
+
+dependencies {
+  compile dep.jsr305
+  compile dep.retrofitMoshiConverter
```
felipecsl: not sure why this is needed too

sav007: Somehow but intellij idea doesn't recognize transient dependencies and marks for sample module `SampleApplication` MoshiConverterFactory as red. Will investigate and remove these

apollo-api/src/main/java/com/apollostack/api/graphql/ResponseStreamReader.java
sav007: it should be as it used in converter module in com.apollostack.converter.pojo.ApolloResponseBodyConverter

sav007:
@felipecsl this PR will be smaller when I remove examples of generated classes. And these classes just have been copied from moshi lib as `BufferedSourceJsonReader` package protected but we need it as it provides support of `okio.BufferedSource`:
`com.apollostack.converter.pojo.BufferedSourceJsonReader`
`com.apollostack.converter.pojo.JsonDataException`
`com.apollostack.converter.pojo.JsonEncodingException`
`com.apollostack.converter.pojo.JsonReader`
`com.apollostack.converter.pojo.JsonScope`

apollo-sample/src/main/java/com/example/apollostack/sample/TestQueryWithFragmentResponseData.java
sav007: Because we can't use stream to read for example `name`, `species ` fields for both `Node` and `PeopleFragment`

felipecsl: Got it 👍  maybe worth adding a comment to clarify

felipecsl: d'oh! of course! sorry, slow start after holidays 😄

sav007:
@martijnwalraven @felipecsl I've updated PR, pls take a look

martijnwalraven:
@sav007: Using the `ValueHandler` interface and passing in fields seems like a nice way of making the readers share the same semantics! I find it hard to see all the implications from the code, but I think we could get this to work. So maybe we should just go ahead and get something running, and then we can always fine tune it.

sav007:
@felipecsl @martijnwalraven will merge this PR on green

PR #83, opened 2017-01-10
Handle operation packages

### LINKED ISSUE #69: Generated classes are created under generatedIR package
The gradle plugin doesn't seem to be respecting the original package name of each graphql file and just adding them all into the `generatedIR` package (eg.: `build/generated/source/apollo/generatedIR/DroidDetails.java`)

The generated package name should match the original file path.

marwanad:
This adds support to generating and packaging the classes based on a `path` field under the operation object and the irFile relative path.

Operations are generated under a namespace matching the folder structure after `src/main/graphql` (restricted to a `src/x/graphql/....` structure). 

Fragments and Type Declarations are generated by appending `.fragment` or `.type` to the namespace that matches the folder structure starting from where the IR file was found.

For example, if the IR path was the following
```
src/main/graphql/com/example/api/API.json
```
Fragments will be generated under `com.example.api.fragment` and types under `come.example.api.type`. This makes handling and resolving the fields referenced in fragments and types much easier.

@martijnwalraven @felipecsl @sav007 please take a look. If no complaints, I'll push the apollo-codegen and Gradle plugin changes.

Closes #69


apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/CodeGenerator.kt
```diff
@@ -3,5 +3,5 @@ package com.apollostack.compiler.ir
 import com.squareup.javapoet.TypeSpec
 
 interface CodeGenerator {
-  fun toTypeSpec(): TypeSpec
-}
\ No newline at end of file
+  fun toTypeSpec(irPkgName: String): TypeSpec
```
felipecsl: I see `pkg` here and `package` in other places. Do mind using `package` everywhere for consistency? Eg.: `irPackageName`

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/TypeDeclaration.kt
```diff
@@ -38,5 +38,6 @@ data class TypeDeclaration(
     return builder.build()
   }
 
-  private fun inputObjectToTypeSpec() = InputObjectTypeSpecBuilder(name, fields ?: emptyList()).build()
-}
\ No newline at end of file
+  private fun inputObjectToTypeSpec(irPkgName: String) = InputObjectTypeSpecBuilder(name, fields ?: emptyList(),
```
felipecsl: break the method body into a new line

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/graphql/Type.kt
```diff
@@ -21,7 +22,8 @@ sealed class Type(val isOptional: kotlin.Boolean) {
 
   class Unknown(isOptional: kotlin.Boolean, val typeName: kotlin.String) : Type(isOptional)
 
-  fun toJavaTypeName() = graphQlTypeToJavaTypeName(this, !isOptional, isOptional)
+  fun toJavaTypeName(packageName: kotlin.String = "") = graphQlTypeToJavaTypeName(this, !isOptional, isOptional,
```
felipecsl: new line for method body

felipecsl:
Looks like this requires a change to the IR format to include the path field. Is @martijnwalraven 👍 with that?

marwanad:
@felipecsl @martijnwalraven Another alternative would be for the plugin to handle the `.graphql` files lookup and instead of running `apollo-codegen generate **/*.graphql` pass it the file paths.

Then, `GraphQlCompiler` would take a Map of `.graphql` files and their paths and does the resolution there.

It seems better to just have it in the IR with `apollo-codegen generate --paths true` or something.

I'm open to other approaches.

sav007:
>Fragments will be generated under com.example.api.fragment and types under come.example.api.type. This makes handling and resolving the fields referenced in fragments and types much easier.

Not sure if it will be easy as then we need to resolve package names inside code generation

sav007:
@marwanad @felipecsl sorry maybe I missing smth but why we can't keep package where we found *.graphql files

so for instance if we found *.graphql files in 2 places:
`src/main/graphql/com/example/customer/*.grapqhl`
`src/main/graphql/com/example/merchant/*.graphql`

Then we will generate 2 packages and place all code generated to these packages:
`com.example.customer.*.classes`
`com.example.merchant.*.classes`

marwanad:
> Not sure if it will be easy as then we need to resolve package names inside code generation

 It's already [handled](https://github.com/apollostack/apollo-android/pull/83/files#diff-db681cd13cab195b3efd0f10e4154002) as part of the PR.

All the builders that handle code generation take a `irPackageName` parameter and so are able to resolve fragment and input type packages by appending `.fragment` or `.type` to that package.

Using javapoet's `ClassName.get("$irPackageName.fragment", fragmentName)` handles the imports.


> sorry maybe I missing smth but why we can't keep package where we found *.graphql files

We're doing the generation based on the IR file, not .graphql files. We decided we're going to generate one big IR file for all the `.graphql` files using apollo-codegen. The IR file doesn't contain any information on the location of those `.graphql` files. Does that answer your question?


I hope this example clears it up:

```
src/main/graphql/com/example/api/schema.json
src/main/graphql/com/example/api/customer/Customer.graphql
src/main/graphql/com/example/api/merchant/Merchant.graphql
```

1. Gradle plugin will find the `schema.json` file, run `apollo-codegen generate **/*.graphql -- schema src/main/graphql/api/schema.json`. This will pick up all `.graphql` files under src/main and generate one big IR file.

2. The apollo-codegen script generates an APIDebug.json file under `source/apollo/generated/com/example/api/APIDebug.json`. The `com.example.api` is what I refer to as `irPackageName` above.

3. All fragments and types under  `Customer.graphql`, `Merchant.graphql` will be generated under `com.example.api.fragment` and `com.example.api.type` respectively.

4. The `CustomerQuery` and `MerchantQuery` files will be generated under `com.example.api.customer` and `com.example.api.merchant` respectively and they're aware of where the referenced fragments are.


Note that it's not even required for the query files to be under the `api` package because they're aware of where the IR file is and thus are also aware of where to reference the types and fragments. So, for instance they could live under `src/main/graphql/com/example/model/` and everything would work still.

(PS. I haven't pushed the gradle plugin change yet)

@sav007

sav007:
>  We're doing the generation based on the IR file, not .graphql files. We decided we're going to generate one big IR file for all the .graphql files using apollo-codegen.

Right but can we break it to smaller ones per package?

If user wants to share fragments then he needs to put all *.graphql files in one folder that will define the package name for all generated classes

sav007:
> The CustomerQuery and MerchantQuery files will be generated under com.example.api.customer and com.example.api.merchant respectively and they're aware of where the referenced fragments and types are.

Agh I see, hm yes makes sense. Thx for explanation!

apollo-compiler/src/main/kotlin/com/apollostack/compiler/SchemaTypeSpecBuilder.kt
```diff
@@ -65,15 +65,15 @@ class SchemaTypeSpecBuilder {
...
 
     /** Returns a generic `Fragments` interface with methods for each of the provided fragments */
-    private fun newFragmentInterfaceSpec(fragments: List<String>): TypeSpec =
+    private fun newFragmentInterfaceSpec(fragments: List<String>, irPkgName: String): TypeSpec =
         TypeSpec.interfaceBuilder(FRAGMENTS_INTERFACE_NAME)
             .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
             .addMethods(fragments.map {
               MethodSpec.methodBuilder(it.decapitalize())
-                  .returns(ClassName.get("", it.capitalize()))
+                  .returns(ClassName.get("$irPkgName.${GraphQLCompiler.FRAGMENT_PACKAGE_PREFIX}", it.capitalize()))
```
sav007: Should it be passed as val to constructor, to break relation of this class to `GraphQLCompiler`

apollo-compiler/src/main/kotlin/com/apollostack/compiler/Util.kt
```diff
@@ -1,8 +1,17 @@
 package com.apollostack.compiler
 
 import com.squareup.javapoet.*
+import java.io.File
 import javax.lang.model.element.Modifier
 
+fun String.formatPackageName(): String {
```
sav007: If I'm right but this func is used only in GraphQLCompiler should it be there?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/FragmentTypeSpecBuilder.kt
```diff
@@ -1,14 +1,21 @@
...
-  override fun toTypeSpec(): TypeSpec = fragment.toTypeSpec().let {
+    val generateClasses: Boolean,
+    val fragmentPkgName: String,
+    val irPkgName: String
+) : TypeSpecBuilder {
+  override fun packageName(): String {
+    return fragmentPkgName
+  }
+
+  override fun toTypeSpec(): TypeSpec = fragment.toTypeSpec(irPkgName).let {
```
sav007: Always have a question for that, should it be on the next line when you use `=` or not ? :)

apollo-compiler/src/main/kotlin/com/apollostack/compiler/TypeSpecBuilder.kt
```diff
@@ -0,0 +1,8 @@
+package com.apollostack.compiler
+
+import com.squareup.javapoet.TypeSpec
+
+interface TypeSpecBuilder {
```
sav007: Not sure do we need this interface, as we can still use `CodeGenerator` and `packageName` used only in `GraphQLCompiler` where it even set

apollo-compiler/src/main/kotlin/com/apollostack/compiler/Util.kt
marwanad: yup only used by the compiler, will move.

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/FragmentTypeSpecBuilder.kt
marwanad: I couldn't find a Kotlin style guide.
But, if we're going to follow the Scala one which has very similar constructs.

```
When the method body is a single expression longer than 30 (or so) characters 
but still shorter than 70 (or so) characters, it should be given on the following line,
indented two spaces
```

apollo-compiler/src/main/kotlin/com/apollostack/compiler/TypeSpecBuilder.kt
marwanad: CodeGenerator is used everywhere by multiple code generator. I meant to have this interface to be used with the top-level spec builders, `OperationTypeSpecBuilder`, `FragmentTypeSpecBuilder`, `TypeDeclarationTypeSpecBuilder` on which the `write` method gets called.

marwanad:
> Right but can we break it to smaller ones per package?
If user wants to share fragments then he needs to put all *.graphql files in one folder that will define the package name for all generated classes

We definitely can but another reason we decided to go with one IR for all is because doing it per package is going to be hackier (basically would be executing a `apollo-codegen` shell command in a loop for every folder with query files but it can definitely be done). 

I also think that one for all is less error prone because `apollo-codegen` restricts operation names and fragment names to be unique so we can avoid generating duplicates in some cases. If the user has multiple operations and fragments with the same name, he'll get an error. 

> PS this PR will be pain in my a... for my opened PR for pojo generation :)

LOL yeah 😁  . I took a look, there are quite a few changes to the code generation builders and fragment logic.

apollo-compiler/src/main/kotlin/com/apollostack/compiler/TypeSpecBuilder.kt
marwanad: Also, this would force them to implemented a `packageName` method, which isn't really relevant to their functionality.

apollo-compiler/src/main/kotlin/com/apollostack/compiler/SchemaTypeSpecBuilder.kt
marwanad: yeah that would make more sense.
Maybe those builders should be receive a `fragmentPackage` and `typePackage` params instead of one `irPkgName`

apollo-compiler/src/main/kotlin/com/apollostack/compiler/TypeSpecBuilder.kt
marwanad: Oh nvm, I see what you're saying.
We definitely don't need that.

sav007:
@marwanad we can't merge this PR without changes to apollo-codegen, right?

marwanad:
@sav007 yup, I was waiting for @martijnwalraven's approval before pushing the change.

apollo-compiler/src/main/kotlin/com/apollostack/compiler/SchemaTypeSpecBuilder.kt
felipecsl: agreed

felipecsl:
I think going for the change on `apollo-codegen` is still the cleanest solution. I'll ping Martijn about it during out next hangouts

apollo-compiler/src/main/kotlin/com/apollostack/compiler/TypeDeclarationTypeSpecBuilder.kt
```diff
@@ -0,0 +1,11 @@
+package com.apollostack.compiler
+
+import com.apollostack.compiler.ir.CodeGenerator
+import com.apollostack.compiler.ir.TypeDeclaration
+import com.squareup.javapoet.TypeSpec
+
+class TypeDeclarationTypeSpecBuilder(
+    val type: TypeDeclaration
+) : CodeGenerator {
+  override fun toTypeSpec(fragmentsPkgName: String, typesPkgName: String): TypeSpec = type.toTypeSpec(fragmentsPkgName, typesPkgName)
```
marwanad: new line

felipecsl:
Created https://github.com/apollostack/apollo-codegen/issues/38

felipecsl:
@marwanad this should have all you need for this PR: https://github.com/apollostack/apollo-codegen/issues/38#issuecomment-273919387
I think the key is `filePath` instead of just `path`

marwanad:
@felipecsl Awesome! I updated the plugin and ran the tasks on the sample project and it seems to work 😄 

PS. I removed the `DebugAPI.json` and `ReleaseAPI.json`, since those will be generated by the plugin anyways.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRGenTask.groovy
```diff
@@ -40,7 +42,8 @@ public class ApolloIRGenTask extends NodeTask {
     dependsOn(ApolloCodeGenInstallTask.NAME)
 
     possibleGraphQLPaths = buildPossibleGraphQLPaths()
-    outputDir = project.file("src/$variant/graphql/generatedIR")
+    schemaFile = userProvidedSchemaFile() ?: searchForSchemaFile()
+    outputDir = schemaFile.getParentFile()
```
marwanad: Should we be generating the IR under the `build` folder? For now, it's generated under the same folder as the schema file.

sav007:
@marwanad sorry :( let me know if you need help of resolving conflicts

marwanad:
@sav007 updated. I'm not sure if any specific handling would be needed for the pojo case.

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRGenTask.groovy
marwanad: A consequence of this is that IR Files for all variants will be under `src/main/graphql/...wherever_the_schema_file_was`.

A better alternative imo would be doing the same check we do in `GraphQlCompiler` for a structure matching `src/x/graphql/**` and generating the schema file in the directory that results after replacing x with the variant name.

sav007: So the implications of having IR file under `src` folder is that it will be under source control. I don't think this is good idea as it will be a big file that probably won't be possible to review before committing to source control. So I would move it to `build` folder

sav007:
@marwanad is this PR still blocked by changes required to `apollo-codegen` or we are good to go?

apollo-gradle-plugin/src/main/groovy/com/apollostack/android/gradle/ApolloIRGenTask.groovy
marwanad: yup, that's a great point.

marwanad:
@sav007 it's good to go now

apollo-compiler/src/main/kotlin/com/apollostack/compiler/BuilderTypeSpecBuilder.kt
```diff
@@ -7,7 +7,8 @@ import javax.lang.model.element.Modifier
 class BuilderTypeSpecBuilder(
     val targetObjectClassName: ClassName,
     val fields: List<Pair<String, Type>>,
-    val fieldDefaultValues: Map<String, Any?>
+    val fieldDefaultValues: Map<String, Any?>,
+    val typesPkgName: String
```
sav007: nit: to be consistent `typesPkgName` -> `typesPakageName `

apollo-compiler/src/main/kotlin/com/apollostack/compiler/GraphQLCompiler.kt
```diff
@@ -1,20 +1,34 @@
...
+
     val operationTypeBuilders = ir.operations.map { OperationTypeSpecBuilder(it, ir.fragments) }
     (operationTypeBuilders + ir.fragments + ir.typesUsed).forEach {
-      JavaFile.builder(packageName, it.toTypeSpec(!generateClasses, emptyList(), ir.typesUsed)).build().writeTo(outputDir)
+      val javaFilePackageName = when (it) {
+        is OperationTypeSpecBuilder -> it.operation.filePath.formatPackageName()
+        is Fragment -> fragmentsPackage
+        is TypeDeclaration -> typesPackage
+        else -> irPackageName
+      }
```
sav007: nit: I would extract line 22 - 27 as function `formatPackageName(:CodeGenerator)` that will return package name

apollo-compiler/src/main/kotlin/com/apollostack/compiler/InputObjectTypeSpecBuilder.kt
```diff
@@ -7,7 +7,8 @@ import javax.lang.model.element.Modifier
 
 class InputObjectTypeSpecBuilder(
     val name: String,
-    val fields: List<TypeDeclarationField>
+    val fields: List<TypeDeclarationField>,
+    val typesPkgName: String
```
sav007: nit: `typesPkgName` -> `typesPackageName`

marwanad:
Updated. Will merge by tomorrow if no more comments.

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/CodeGenerator.kt
```diff
@@ -3,5 +3,6 @@ package com.apollostack.compiler.ir
 import com.squareup.javapoet.TypeSpec
 
 interface CodeGenerator {
-  fun toTypeSpec(abstractClass: Boolean, reservedTypeNames: List<String>, typeDeclarations: List<TypeDeclaration>): TypeSpec
-}
\ No newline at end of file
+  fun toTypeSpec(abstractClass: Boolean, reservedTypeNames: List<String>, typeDeclarations: List<TypeDeclaration>,
+      fragmentsPackageName: String = "", typesPackageName: String = ""): TypeSpec
```
sav007: This method becomes messy, probably needs to be refactored in future

marwanad: Created #92

PR #90, opened 2017-01-20
Pojo converter integration test

### LINKED ISSUE #81: Cover pojo converter module with unit tests

sav007:
Closes #81 

Don't be scary about the size of this PR, 3 classes were copied and will be removed from the source to be generated as soon as we start publishing snapshots.

Another large file represents mocked json response. 

@digitalbuddha


apollo-converters/pojo/src/test/java/com/apollostack/converter/pojo/AllPlanets.java
```diff
@@ -0,0 +1,295 @@
...
+import com.apollostack.api.graphql.ResponseReader;
+import java.io.IOException;
+import java.lang.Integer;
+import java.lang.Object;
+import java.lang.Override;
+import java.lang.String;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public final class AllPlanets implements Query<Operation.Variables> {
```
sav007: Generated class

apollo-converters/pojo/src/test/java/com/apollostack/converter/pojo/FilmFragment.java
```diff
@@ -0,0 +1,66 @@
...
+import com.apollostack.api.graphql.ResponseFieldMapper;
+import com.apollostack.api.graphql.ResponseReader;
+import java.io.IOException;
+import java.lang.Object;
+import java.lang.Override;
+import java.lang.String;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class FilmFragment {
```
sav007: Generated class

apollo-converters/pojo/src/test/java/com/apollostack/converter/pojo/PlanetFargment.java
```diff
@@ -0,0 +1,79 @@
...
+import com.apollostack.api.graphql.ResponseReader;
+import java.io.IOException;
+import java.lang.Double;
+import java.lang.Object;
+import java.lang.Override;
+import java.lang.String;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class PlanetFargment {
```
sav007: Generated class

apollo-converters/pojo/src/main/java/com/apollostack/converter/pojo/ApolloConverterFactory.java
```diff
@@ -23,4 +24,5 @@
     }
     return null;
   }
+
```
marwanad: not needed

apollo-compiler/src/main/kotlin/com/apollostack/compiler/ir/CodeGenerator.kt
```diff
@@ -3,5 +3,6 @@ package com.apollostack.compiler.ir
 import com.squareup.javapoet.TypeSpec
 
 interface CodeGenerator {
-  fun toTypeSpec(abstractClass: Boolean, reservedTypeNames: List<String>, typeDeclarations: List<TypeDeclaration>): TypeSpec
+  fun toTypeSpec(abstractClass: Boolean, reservedTypeNames: List<String>,
```
marwanad: can keep on same line

sav007:
@marwanad 
Not sure that this is good idea to move them under `apollo-gradle-plugin` as these test specific to pojo converter only, not to the gradle plugin, otherwise `apollo-gradle-plugin` will have to have dependency to all submodules: `apollo-converters/moshi`, `apollo-converters/pojo`, `apollo-converters/*`. Unfortunately we can't run gradle annotation processor from submodule, only from maven repo, so that's why we have to add gradle dep to snapshot `apollo-gradle-plugin` artifact.

Not sure even if we move test to `apollo-gradle-plugin` module how we will run groovy `ApolloClassGenTask` before running java unit tests?

marwanad:
@sav007 
The dependencies won't be added to `apollo-gradle-plugin` but rather to the [folder](https://github.com/apollostack/apollo-android/pull/80/commits/480b3102172f8f06d001fe81efcb388ce23ccbef#diff-4e220ce7392211083cf8966972d0a3e8)`apollo-gradle-plugin/src/test/testProject/build.gradle` 
But, yeah I completely agree that having them under the Gradle plugin is a bit odd.

How about having a separate module for integration tests? That's another option.


> how we will run groovy ApolloClassGenTask before running java unit tests?

I guess we'd have to define the integration tests in a different sourceSet then, so `src/test` would be for unit tests and `src/integrationTest` would be for the integration ones and we can configure the ordering through Gradle tasks then.

sav007:
We will revisit this later, I guess with integration tests we want to make sure that changes to converter and the latest artifact of gradle plugin that uploaded to maven work fine. So for now let see how it goes with `com.apollographql.android.plugin:+` that will grab the latest published gradle plugin

PR #93, opened 2017-01-22
Add javax.annotation.Generated annotation to generated classes

sav007:
Annotate generated classes with `javax.annotation.Generated` to indicated that source is been generated by apollo

@digitalbuddha


apollo-compiler/src/main/kotlin/com/apollostack/compiler/Annotations.kt
```diff
@@ -1,11 +1,15 @@
...
+import javax.annotation.Generated
 import javax.annotation.Nonnull
 import javax.annotation.Nullable
 
 object Annotations {
   val NULLABLE: AnnotationSpec = AnnotationSpec.builder(Nullable::class.java).build()
   val NONNULL: AnnotationSpec = AnnotationSpec.builder(Nonnull::class.java).build()
   val OVERRIDE: AnnotationSpec = AnnotationSpec.builder(Override::class.java).build()
+  val GENERATED_BY_APOLLO: AnnotationSpec = AnnotationSpec.builder(Generated::class.java)
+      .addMember("value", CodeBlock.of("\$S", "Apollo GraphQl")).build()
```
digitalbuddha: @martijnwalraven should we do ApolloData as the string?

apollo-compiler/src/main/kotlin/com/apollostack/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -13,7 +13,9 @@ class OperationTypeSpecBuilder(
 
   override fun toTypeSpec(abstractClass: Boolean, reservedTypeNames: List<String>,
       typeDeclarations: List<TypeDeclaration>, fragmentsPackage: String, typesPackage: String): TypeSpec {
-    return TypeSpec.classBuilder(OPERATION_TYPE_NAME)
+    return TypeSpec
+        .classBuilder(OPERATION_TYPE_NAME)
```
digitalbuddha: Spacing is inconsistent. In last file   .classBuilder was on previous line

apollo-compiler/src/main/kotlin/com/apollostack/compiler/Annotations.kt
digitalbuddha: Of we keep GraphQL why is the Q capital  but not the L. Not sure if Nitpick

martijnwalraven: Yeah, I think Apollo GraphQL would be fine.

PR #97, opened 2017-01-22
Rename package apollostack to apollographql

### LINKED ISSUE #87: Rename packages from com.apollostack to com.apollodata

felipecsl:
Fixes #87 
Also:

* Update build tools to `25.0.2`
* Use uploaded version of the artifacts instead of shadow jars
* Remove shadow jar dependencies
* Upgrade gradle to `3.3`
* Upgrade Android gradle plugin to `2.3.0-beta2`


felipecsl:
dang @marwanad beat me to it 😢

marwanad:
GROUP_ID is `com.apollographql.android`, shouldn't we append the package identifiers to that? 
ie. `com.apollographql.android.api` instead of `com.apollographql.api`?

felipecsl:
@marwanad that information is in the `POM_ARTIFACT_ID`.
By the way I'm still getting 2 failures on `ApolloPluginSpec` because of:
```
Could not get unknown property 'GraphQLExtension' for source set android test of type com.android.build.gradle.internal.api.DefaultAndroidSourceSet
```

Any idea why? Other than that, this should be good to go

marwanad:
@felipecsl I think the import is missing [here](https://github.com/apollostack/apollo-android/pull/97/files#diff-48b8b90eca3e959c36b23a6acd7dabd0L7).

apollo-gradle-plugin/src/test/testProject/build.gradle
```diff
@@ -1,17 +1,13 @@
 buildscript {
   repositories {
-    mavenCentral()
     jcenter()
+    maven { url "https://jitpack.io" }
```
digitalbuddha: What is jitpack for?

digitalbuddha:
change build tools in travis.yml to `build-tools-25.0.2`

apollo-gradle-plugin/src/test/testProject/build.gradle
felipecsl: for the `kotlin-pluralizer` dependency

build.gradle
```diff
@@ -2,16 +2,17 @@ ext {
...
   dep = [
-      androidPlugin: 'com.android.tools.build:gradle:2.3.0-beta1',
+      androidPlugin: 'com.android.tools.build:gradle:2.3.0-beta2',
       supportAnnotations: 'com.android.support:support-annotations:25.1.0',
       compiletesting: 'com.google.testing.compile:compile-testing:0.10',
       javaPoet: 'com.squareup:javapoet:1.8.0',
       moshi: 'com.squareup.moshi:moshi:1.3.1',
       rxjava: 'io.reactivex:rxjava:1.2.3',
       jsr305: 'com.google.code.findbugs:jsr305:3.0.1',
+      jsr205: 'javax.annotation:jsr250-api:1.0',
```
marwanad: `jsr250` instead? Also, not sure why is this needed.

felipecsl: for the `@Generated` annotation. For some reason that couldn't be find anywhere else.
Do you know if that's in any of the other deps already?

felipecsl:
Man I hate groovy...

build.gradle
marwanad: Don't think it is. Dagger seems to include both so it should be fine.

felipecsl:
@marwanad hmm you're right, the naming is actually inconsistent. Do you think the Group ID should be just `com.apollographql`?

martijnwalraven:
@felipecsl: I think `com.apollographql.android`would be the safer choice, because we may want to release other Java tools in the future.

felipecsl:
@martijnwalraven you're right, I hadn't realized that the repo name is already restricted to android.
I was under the assumption that this could also be targeted at pure Java projects

marwanad:
If so, we should probably append `android` to the package names of the other modules then.
I assume the Gradle plugin would also be targeting android only?

felipecsl:
It sounds to me like we don't have to be that restrictive and actually try to make this as platform agnostic as possible, not making it restricted to android.

martijnwalraven:
Yeah, we had the same discussion with Apollo iOS. It actually runs on Mac and tvOS, and should even work on server-side Swift. But we wanted the name to reflect the primary focus and make it easy for people to find it.

Also, we probably want to be able to take advantage of Android-specific features and not commit ourselves to be platform agnostic. We can always extract parts later if there is a need to support other platforms.

felipecsl:
This is ready to go. Waiting for a 👍

digitalbuddha:
@felipecsl failing compilation, I think imports in sample are incorrect.

felipecsl:
@digitalbuddha I updated the package names and uploaded a new snapshot but it's still using the old one since the update doesn't reflect immediately. I think i'll just revert the last commit and send it as a separate PR

PR #99, opened 2017-01-22
Contribution Guidelines

digitalbuddha:
@martijnwalraven let me know if anything needs to be changed. This is what I use for my projects.


martijnwalraven:
Looks good to me! We can always expand on this as needed.

PR #108, opened 2017-01-23
fix leading dot in fragment and types package name

### LINKED ISSUE #107: Fragments generate incorrect "import" and "package" information
given the following modification to the android/apollo-sample project
```
query DroidDetails {
  species(id: "c3BlY2llczoy") {
    id
    name
    ...SpeciesInformation

  }
}

fragment SpeciesInformation on Species{
  classification
}
```

if generates an extra **dot** for the _import_ and _package_ statements

```
//DroidDetails.java
import .fragment.SpeciesInformation;
```

```
//fragments/SpeciesInformat.java
package .fragment;
```

marwanad:
Should fix #107.


digitalbuddha:
Would it be possible to create a test to validate fix?

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -46,7 +46,7 @@ class OperationTypeSpecBuilder(
 
   private fun TypeSpec.Builder.addQueryDocumentDefinition(fragments: List<Fragment>): TypeSpec.Builder {
     val initializeCodeBuilder = CodeBlock.builder().add(OPERATION_DEFINITION_FIELD_NAME)
-    fragments.forEach {
+    fragments.filter { operation.fragmentsReferenced.contains(it.fragmentName) }.forEach {
```
marwanad: @sav007 fyi, another bug was because we were attempting to append the fragment definition for an operation that doesn't reference that fragment. (See the file below)

Fragments are defined per schema file so here we need to filter out those that aren't referenced in the operation before appending their definition.

marwanad:
Dammit, we need to have a new SNAPSHOT before the added test passes. I guess I'll add it in a separate PR then.

Issue #88, opened 2017-01-20
[Code Generation] Add support of custom scalar types


sav007:
In most GraphQL service implementations, there is also a way to specify custom scalar types. For example, we could define a Date type

digitalbuddha:
Ideally we follow a pattern similar to sqldelight where user would be able to define custom types within  .graphql files and then passes in adapters for how to serialize.  Maybe we can add custom type Metadata to schema files?

sav007:
we can use comments :)
```
query Test($count:Int!) {
  shop {
    products(first:$count) {
      edges {
        node {
          title
          createdAt # as java.util.Date
        }
      }
    }
  }
}
```

marwanad:
I feel a top-level extension could be useful in that case. We're going to have one anyways for setting the POJO flag, so might as well let the user define the mapping there.

The plugin can then manipulate the `schema.json` file with that information before passing it to the compiler.

martijnwalraven:
Yeah, you don't want to specify the mapping on a per-field basis, so the `.graphql` file is not the right place I think.

So some kind of project wide configuration makes the most sense to me. You'd specify a mapping from `DateTime` to `java.util.Date` in there for instance, and something like a type adapter.

You wouldn't want to manipulate the `schema.json`, but the compiler should be aware of the mapping and generate fields of the right type.

sav007:
@martijnwalraven @digitalbuddha @felipecsl @marwanad so what is gonna be how user will define the mapping between GraphqQL custom types to java classes? should it be in gradle as config option for plugin or some file like meta or smth?

PR #111, opened 2017-01-23
Custom scalar type - abstract type generation

sav007:
This PR is part of #88, and addresses interface generation with provided mapping of custom schema scalar type to user defined java class name (see `GraphQLCompiler#write`).

This PR doesn't define how this mapping will be provided to `GraphQLCompiler` that is still open question and will be addressed in separate PR.

Pojo generation will be addressed in separate PR as well.

If custom schema scalar mapping provided, class `CustomScalarType` will be generated that represents enum of all defined scalar type mapping (see example https://github.com/sav007/apollo-android/blob/024c64a2c1b0013c8f0cd111785207b17436eaad/apollo-compiler/src/test/graphql/com/example/custom_scalar_type/CustomScalarTypeExpected.java)

This enum will be required later for pojo generation when user will provide scalar type adapters to our response converter, to make it type safe.

```
...
converter.withTypeAdapter(CustomScalarType.DATE, adapter)
...
```


apollo-compiler/src/test/graphql/com/example/custom_scalar_type/CustomScalarTypeExpected.java
```diff
@@ -0,0 +1,20 @@
...
+import com.apollographql.android.api.graphql.ScalarTypeMapping;
+import java.lang.Class;
+import java.lang.String;
+import java.util.Date;
+import javax.annotation.Generated;
+
+@Generated("Apollo GraphQL")
+public enum CustomScalarType implements ScalarTypeMapping {
+  DATE {
+    public String type() {
```
digitalbuddha: do we need type, can't we do `clazz.getSimpleCassName`

sav007: this `type` is type defined in GraphQl schema: like `Date`, `DateTime`, `SomeOtherType`.
`clazz` - is actually java mapping class for this `type`

PR #112, opened 2017-01-25
Pojo custom scalar type

sav007:
This PR is part of #88 that addresses pojo code generation with custom type support.
New type `CUSTOM` of `Field` has been added  with provided `TypeMapping` that will be used in `ResponseReader` to resolve type adapter to be used for response value conversion.

The changes to `ResponseReader` and pojo converter module to utilize user provided type adapters will be addressed in next PR


sav007:
@digitalbuddha this is on my list next

PR #113, opened 2017-01-25
add custom scalar types map param to the plugin write call

marwanad:
Groovy strikes again..

The compiler `write` signature was updated to take an extra map parameter.

I'll start working on a Java ~8~ port to avoid those errors.


digitalbuddha:
"I'll start working on a Java 8 port to avoid those errors." how will that work for those without java8 installed?

marwanad:
@digitalbuddha hmm, I'm not too sure. Wouldn't setting the sourceCompatibility in the build script work `sourceCompatibility = JavaVersion.VERSION_1_8`?

Also, fyi I commented out the integration tests, will uncomment when the new snapshot is out.

digitalbuddha:
It would but I'd be hesitant to have that as requirement. Even with that flag a person would need jdk8 installed which is not a guarantee among android devs. Maybe keep  java 7 as min java version for time being?

marwanad:
@digitalbuddha Ahh I see. That makes sense.

PR #115, opened 2017-01-25
add apollo-api dependency to gradle plugin

marwanad:
Seems that the `apollo-api` dependency should be included with the plugin because the
`compile project (":apollo-api")` block under `apollo-compiler` isn't transitive(?) so we need to have it explicitly under the plugin.

The Gradle plugin needs that because the generated classes reference the api classes and it seems there won't be a use-case where `apollo-api` won't be needed.

### DIFF
--- a/apollo-gradle-plugin/build.gradle
+++ b/apollo-gradle-plugin/build.gradle
@@ -10,6 +10,7 @@ dependencies {
   compileOnly gradleApi()
   compile localGroovy()
   compile project(':apollo-compiler')
+  compile project(':apollo-api')
   compile dep.gradleNodePlugin
   compile dep.androidPlugin
 
--- a/apollo-gradle-plugin/src/test/testProject/build.gradle
+++ b/apollo-gradle-plugin/src/test/testProject/build.gradle
@@ -41,6 +41,7 @@ dependencies {
 
   compile 'com.jakewharton.auto.value:auto-value-annotations:1.3'
   compile 'com.ryanharter.auto.value:auto-value-moshi-annotations:0.4.2'
+  //TODO: remove after new snapshot
   compile 'com.apollographql.android:api:0.0.1-SNAPSHOT'
   compile 'io.reactivex.rxjava2:rxjava:2.0.3'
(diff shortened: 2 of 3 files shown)


marwanad:
That's what I thought too.

I just tried it on `apollo-sample` with just the `com.apollographql.android:gradle-plugin:0.0.1-SNAPSHOT` dependency and it couldn't resolve the ` com.apollographql.android.api.graphql` classes.

felipecsl:
OK I think I found the mystery. The apollo gradle plugin is a classpath dependency, not a regular dependency. That means it will be included in the classpath for the build script (and all its transitive dependencies) but that has nothing to do with the `compile` configuration.
We could maybe have it dynamically add `apollo-api` as a compile dependency. In fact, we probably should do that.
For that reason, I think we should close this PR (it's not the right solution)

felipecsl:
BTW you can verify that by running `./gradlew apollo-sample:buildEnvironment`:

```
\--- com.apollographql.android:gradle-plugin:0.0.1-SNAPSHOT
     +--- com.apollographql.android:compiler:0.0.1-SNAPSHOT
     |    +--- com.squareup:javapoet:1.8.0
     |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.0.5 (*)
     |    +--- com.squareup.moshi:moshi:1.3.1
     |    |    \--- com.squareup.okio:okio:1.11.0
     |    +--- com.github.cesarferreira:kotlin-pluralizer:0.2.7
     |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.0.3 -> 1.0.5 (*)
     |    +--- com.google.code.findbugs:jsr305:3.0.1
     |    \--- com.apollographql.android:api:0.0.1-SNAPSHOT
     |         +--- com.google.code.findbugs:jsr305:3.0.1
     |         \--- javax.annotation:jsr250-api:1.0
     +--- com.moowork.gradle:gradle-node-plugin:1.0.0
     \--- com.android.tools.build:gradle:2.3.0-beta2 (*)
```

marwanad:
That makes perfect sense. Thanks for looking into it! 

I'll work on a proper fix and reopen the PR.

PR #118, opened 2017-01-26
Converter custom scalar type support

sav007:
This PR is another part of #88 that enables to add custom type adapter to POJO converters

The size is large due to the mocked response for integration test

Example of setting up converter with custom type adapter: https://github.com/apollostack/apollo-android/pull/118/files#diff-a55293c5fd086ec33edf055baafc672f


apollo-converters/pojo/src/main/java/com/apollographql/android/converter/pojo/ApolloConverterFactory.java
```diff
@@ -2,25 +2,38 @@
...
+
+import javax.annotation.Nonnull;
 
 import okhttp3.ResponseBody;
 import retrofit2.Converter;
 import retrofit2.Retrofit;
 
 /** TODO */
 public class ApolloConverterFactory extends Converter.Factory {
+  private final Map<TypeMapping, CustomTypeAdapter> customTypeAdapters = new HashMap<>();
```
digitalbuddha: nitpick: https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

apollo-converters/pojo/src/test/java/com/apollographql/android/converter/pojo/IntegrationTest.java
```diff
@@ -35,14 +39,35 @@
...
+
+      @Override public String encode(Date value) {
+        return null;
+      }
+    };
+
     Retrofit retrofit = new Retrofit.Builder()
         .baseUrl(server.url("/"))
-        .addConverterFactory(new ApolloConverterFactory())
+        .addConverterFactory(new ApolloConverterFactory()
```
digitalbuddha: I feel like the api would be nicer if I can do `.addConverterFactory(ApolloConverter.builder().withCustomTypeAdapter().build())`

PR #119, opened 2017-01-27
move plugin extensions and tasks to java

marwanad:
Only `ApolloPlugin` is left. This will be in a separate PR, because it's a bigger change.

I'll move the tests to use Spec for Kotlin, otherwise the test methods names would be messy. (also in a separate PR)


marwanad:
oh TIL. I'll update it then.

PR #125, opened 2017-01-28
add apollo-api dependency through the gradle plugin

marwanad:
Dynamically adds the `apollo-api` dependency when configuring the apollo-plugin since it's needed by the generated classes (follows from #115)

### DIFF
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
@@ -3,6 +3,7 @@ package com.apollographql.android.gradle
 import com.android.build.gradle.AppPlugin
 import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.api.BaseVariant
+import com.apollographql.android.VersionKt
 import com.google.common.collect.Lists
 import com.moowork.gradle.node.NodeExtension
 import com.moowork.gradle.node.NodePlugin
@@ -10,15 +11,11 @@ import org.gradle.api.DomainObjectCollection
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.artifacts.DependencyResolutionListener
+import org.gradle.api.artifacts.ResolvableDependencies
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.compile.JavaCompile
-
-import com.apollographql.android.gradle.GraphQLExtension
-import com.apollographql.android.gradle.ApolloClassGenTask
-import com.apollographql.android.gradle.ApolloIRGenTask
-import com.apollographql.android.gradle.ApolloExtension
-import com.apollographql.android.gradle.ApolloCodeGenInstallTask
 
 class ApolloPlugin implements Plugin<Project> {
   private static final String NODE_VERSION = "6.7.0"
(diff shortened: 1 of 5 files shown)


apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
```diff
@@ -40,7 +37,19 @@ class ApolloPlugin implements Plugin<Project> {
     setupNode()
     project.extensions.create(ApolloExtension.NAME, ApolloExtension)
     createSourceSetExtensions()
-    //TODO: add dependency on apollo-runtime once we have the jars on nexus
+
+    def compileDepSet = project.configurations.getByName("compile").dependencies
+    project.getGradle().addListener(new DependencyResolutionListener() {
+      @Override
+      void beforeResolve(ResolvableDependencies resolvableDependencies) {
+        //TODO: apollo-runtime dependency?
```
digitalbuddha: please make github issue point to this line rather than TODO

apollo-sample/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
   compile project(':apollo-converters:moshi')
   compile project(':apollo-converters:pojo')
   compile project(":apollo-runtime")
+  //TODO: remove dependency after new snapshot release of the plugin
```
digitalbuddha: I'd prefer this to be in an issue too, keeps code cleaner

martinbonnin: I would actually have loved the comment there :), would have saved me e bit of time

martinbonnin:
Sorry guys, I'm catching up on this. What's wrong with asking the devs that want to use apollo to explicitely add the dependency to their build.gradle file ? This dynamic solution seems a bit voodoo to me.

marwanad:
@martinbonnin The generated classes reference classes defined in `apollo-api`. So, the generated classes would be useless without the `apollo-api` dependency.

martinbonnin:
@marwanad yes, I understand that. 
The thing is I think I would prefer to explicitely add the dependency to my build.gradle file without having a plugin do it for me. I've had really bad experiences with fabric or play-services plugins and now I'm a bit paranoid about it. We've had a long conversation about versioning on slack and eventually both solutions will work. 
What is more worrying is that, as I'm speaking, the sample app does not compile because api is included twice.

marwanad:
@martinbonnin catching up on the slack convo.

For the sample, there's an issue #133 (we had to wait for a snapshot release to fix that).

martinbonnin:
@marwanad that's not related to the versioning scheme. You would have had to wait in all cases as 'sample' is pulling the plugin from sonatype. I'm wondering wheter this is something fixable. i.e can you compile a gradle plugin and use it automatically from an android app.

marwanad:
@martinbonnin afaik this isn't feasible, unless you build a `fat-jar` and add it as a `flatDir` dependency which we used to do before having the artifacts on maven.

marwanad:
@martinbonnin actually, it is [possible](https://docs.gradle.org/current/userguide/organizing_build_logic.html#sec:build_sources).

martinbonnin:
Would that work for us ? like putting the plugin in apollo-android/buildSrc and then having a dummy module just to deploy it to maven ?

marwanad:
So, what I was thinking of was `apollo-android/apollo-sample/buildSrc` that would reference the gradle plugin source sets and then have a subproject/module under `apollo-sample` which is the actual android sample that would pick up that plugin dependency.

The deploying to maven would still happen through the `apollo-gradle-plugin` deployJar task, and this setup is just to have `apollo-sample` reference the locally built gradle-plugin, hence catch errors faster.

(Let me know if I misunderstood your first question)

PR #127, opened 2017-01-28
allow to use custom HTTPS certificates in debug mode (useful to debug…

martinbonnin:
… with charles HTTPS proxy)


apollo-sample/src/main/AndroidManifest.xml
```diff
@@ -1,23 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.example.apollographql.sample">
+    package="com.example.apollographql.sample">
```
sav007: we have code style with 2 spaces instead 4 as tab indention

PR #130, opened 2017-01-29
replace gson usages in the plugin with moshi

marwanad:
This _should_ fix the following:

```
Caused by: org.gradle.api.tasks.TaskInstantiationException: 
Could not create task of type 'ApolloCodeGenInstallTask'.

Caused by: java.lang.NoClassDefFoundError: org/gradle/internal/impldep/com/google/gson/stream
/JsonReader
```
Unfortunately, only way to verify is through manual testing.


apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/ApolloPluginBasicAndroidTest.groovy
```diff
@@ -39,6 +39,19 @@ class ApolloPluginBasicAndroidTest extends Specification {
     assert new File(testProjectDir, "build/generated/source/apollo/fragment/SpeciesInformation.java").isFile()
   }
 
+  def "installApolloCodegenTask is up-to-date"() {
```
marwanad: sanity check, to make sure the task runs fine.

PR #131, opened 2017-01-29
Fix apollo codegen task

martinbonnin:
Hopefully fixes `com.squareup.moshi.JsonDataException: Expected a name but was BEGIN_ARRAY at path $._args`


apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
```diff
@@ -54,25 +59,15 @@ private String getApolloVersion() {
...
-      jsonReader.endObject();
-      jsonReader.close();
 
+    Moshi moshi = new Moshi.Builder().build();
+    JsonAdapter<PackageJson> adapter = moshi.adapter(PackageJson.class);
+    try {
+      PackageJson packageJson = adapter.fromJson(Okio.buffer(Okio.source(packageFile)));
+      return packageJson.version;
+    } catch (IOException e) {
+      return null;
```
digitalbuddha: maybe log during exception?

PR #134, opened 2017-01-30
update module dependencies

marwanad:
we now dynamically add the apollo-api dependency through the plugin.

This updates the module to deal with the change and avoid conflicts when different modules reference different versions.

### DIFF
--- a/apollo-converters/moshi/build.gradle
+++ b/apollo-converters/moshi/build.gradle
@@ -3,11 +3,18 @@ apply plugin: 'java'
 targetCompatibility = JavaVersion.VERSION_1_7
 sourceCompatibility = JavaVersion.VERSION_1_7
 
+configurations {
+  compileOnly
+}
+
+sourceSets.test.compileClasspath += configurations.compileOnly
+sourceSets.test.runtimeClasspath += configurations.compileOnly
+
 dependencies {
   compile dep.moshi
   compile dep.jsr305
   compile dep.retrofitMoshiConverter
-  compile project (":apollo-api")
+  compileOnly project(":apollo-api")
 
   testCompile dep.junit
   testCompile dep.truth
--- a/apollo-runtime/build.gradle
+++ b/apollo-runtime/build.gradle
@@ -34,7 +34,7 @@ dependencies {
   compile dep.jsr305
   compile dep.retrofit
   compile dep.supportAnnotations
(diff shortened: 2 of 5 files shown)


apollo-runtime/build.gradle
```diff
@@ -34,7 +34,7 @@ dependencies {
   compile dep.jsr305
   compile dep.retrofit
   compile dep.supportAnnotations
-  compile project (":apollo-api")
+  provided project (":apollo-api")
```
sav007: is it correct, shouldn't it be `compileOnly `

apollo-converters/moshi/build.gradle
```diff
@@ -3,11 +3,18 @@ apply plugin: 'java'
 targetCompatibility = JavaVersion.VERSION_1_7
 sourceCompatibility = JavaVersion.VERSION_1_7
 
+configurations {
```
sav007: I thought compileOnly is supported by gradle and we don't need custom configuration?

marwanad: `compileOnly` is part of the java plugin applied to a gradle buildscript. Not all projects support it. 

For android libraries, `compileOnly` isn't available since we're not applying the java plugin. We have to use `provided` for those.

For our other modules, we apply the java plugin so `compileOnly` is used.

```
compile-only dependencies can now be declared for all Java projects using the Java Plugin
```

This line creates a new `compileOnly` configuration. Its main task is to update the test sourceSet’s compileClasspath and runtimeClasspath to include to the `compileOnly` provided dependencies. Otherwise, you can't run the converters test.

marwanad: Quoting Gradle Release Notes:

```
You can now declare dependencies to be used only at compile time in conjunction with the Java plugin. Compile only dependencies are used only during source compilation and are not included in the runtime classpath or exposed to dependent projects. This behavior is similar to that of the 'provided' scope available in Maven based builds. However, unlike Maven provided dependencies, compile only dependencies in Gradle are not included on the test classpath.
```

marwanad: For android libraries:

https://developer.android.com/studio/build/dependencies.html#library_dependency_configurations

sav007: Thx for explanation

PR #140, opened 2017-01-31
Restrict overriding custom scalar types

### LINKED ISSUE #88: [Code Generation] Add support of custom scalar types

sav007:
This PR is the last part of #88 and addresses issue with custom type mapping.
User won't be able override any GraphQL type with own type, only scalar types that defined in `typesUsed` of IR.json (GraphQL custom scalar types) are eligible for mapping.

Small refactoring, renaming TypeMapping to ScalarType

Closes #88

### DIFF
--- a/apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/ApolloPluginPojoAndroidTest.groovy
+++ b/apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/ApolloPluginPojoAndroidTest.groovy
@@ -23,12 +23,12 @@ class ApolloPluginPojoAndroidTest extends Specification {
     def result = GradleRunner.create()
         .withProjectDir(testProjectDir)
         .withPluginClasspath()
-        .withArguments("build")
+        .withArguments("generateApolloClasses")
         .forwardStdError(new OutputStreamWriter(System.err))
         .build()
 
     then:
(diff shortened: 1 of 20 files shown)


apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/ApolloPluginPojoAndroidTest.groovy
```diff
@@ -23,12 +23,12 @@ class ApolloPluginPojoAndroidTest extends Specification {
...
         .withProjectDir(testProjectDir)
         .withPluginClasspath()
-        .withArguments("build")
+        .withArguments("generateApolloClasses")
         .forwardStdError(new OutputStreamWriter(System.err))
         .build()
 
     then:
-    result.task(":build").outcome == TaskOutcome.SUCCESS
+    result.task(":generateApolloClasses").outcome == TaskOutcome.SUCCESS
```
marwanad: I assume we'd want to use `:build` once we have the api snapshot on maven?

sav007: yes will change this next

PR #141, opened 2017-02-01
fix Fragment class imports in Operation code generation

kassim:
I was finding a few of my Fragment classes weren't being imported during code generation
**edit:** possibly due to Fragments being used within the other Fragments (only the highest level Fragment was being imported)

to solve this I've applied the appropriate arguments as described in JavaPoet's documentation for using classes that may not have yet been generated
https://github.com/square/javapoet#t-for-types


kassim:
Here. Sorry I should have provided that beforehand. I've updated the description stating this is possibly due to Fragments being used within the other Fragments - as only the highest level Fragment (SearchFields) was being imported


```
query Search ($query: String!, $country: String!, $after: String, $withEvents: Boolean!,$withLocations: Boolean!,$withCities: Boolean!) {
    events: search(query: $query, country: $country, first: 10, type: EVENT, after: $after)
    @include(if:$withEvents) {
        ...SearchFields
    }
    locations: search(query: $query, country: $country, first: 10, type: LOCATION, after: $after)
    @include(if:$withLocations) {
        ...SearchFields
    }
    cities: search(query: $query, country: $country, first: 10, type: CITY, after: $after)
    @include(if:$withCities) {
        ...SearchFields
    }
}
fragment SearchFields on SearchConnection {
    edges {
        cursor
        node {
            ...EventFields
            ...LocationFields
            ...CityFields
        }
    }
}
fragment CityFields on City {
    id
    name
    country
}
fragment EventFields  on Event {
    id
    name
    country
    location {
        ...LocationFields
    }
    imageUrl
    startDate
}
fragment LocationFields  on Location {
    id
    name
    city {
        ...CityFields

    }
}
```

marwanad:
Oh interesting, so it lead to something like that:

```
String QUERY_DOCUMENT = OPERATION_DEFINITION + "\n"
   + SearchFields.FRAGMENT_DEFINITION + 
CityFields.FRAGMENT_DEFINITION + 
EventFields.FRAGMENT_DEFINITION + 
LocationFields.FRAGMENT_DEFINITION;

  public interface Data extends Operation.Data {
      Fragments fragments();

      interface Fragments {
        SearchFields searchFields();
      }
    }
  }
```
And it was missing the imports for city, events and location, correct?

kassim:
correct

marwanad:
Do you mind throwing a `.graphql` test fixture [here](https://github.com/apollographql/apollo-android/tree/master/apollo-gradle-plugin/src/test/testProject/android/basic/src/main/graphql/com/example)?

I think this from: https://github.com/graphql/swapi-graphql/tree/master/doc/example_queries
should be good enough.

```
{
  allStarships(first: 7) {
    edges {
      node {
        ...starshipFragment
      }
    }
  }
}

fragment starshipFragment on Starship {
  id
  name
  model 
  costInCredits
  pilotConnection { edges { node { ...pilotFragment }}}
}
fragment pilotFragment on Person {
  name
  homeworld { name }
}
```

kassim:
added

and double checked, ApolloPluginBasicAndroidTest was failing before my changes with:
```
:compileDebugJavaWithJavac/Users/kassim/git/apollo-android/apollo-gradle-plugin/build/inegrationTests/basic/build/generated/source/apollo/com/example/AllStarships.java:26: error: cannot find symbol
   + PilotFragment.FRAGMENT_DEFINITION;
     ^
  symbol:   variable PilotFragment
  location: class AllStarships
1 error
 FAILED
```

but succeed after the changes are applied

sav007:
@kassim I would rather add test case to `apollo-compiler` than to plugin, as this fix related to compiler only

kassim:
okay so when I add that same query to the sample project it doesn't import PilotFragment

so my solution only seemed to solve for my issue which is a shame

any suggestions?

sav007:
@kassim looking into this

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -42,12 +42,14 @@ class OperationTypeSpecBuilder(
     )
   }
 
-  private fun TypeSpec.Builder.addQueryDocumentDefinition(fragments: List<Fragment>): TypeSpec.Builder {
+  private fun TypeSpec.Builder.addQueryDocumentDefinition(fragments: List<Fragment>, fragmentsPackage: String): TypeSpec.Builder {
     val initializeCodeBuilder = CodeBlock.builder().add(OPERATION_DEFINITION_FIELD_NAME)
     fragments.filter { operation.fragmentsReferenced.contains(it.fragmentName) }.forEach {
+      var className = ClassName.get(fragmentsPackage, it.interfaceTypeName());
```
sav007: Should be val

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -17,7 +17,7 @@ class OperationTypeSpecBuilder(
         .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
         .addQuerySuperInterface(operation.variables.isNotEmpty())
         .addOperationDefinition(operation)
-        .addQueryDocumentDefinition(fragments)
+        .addQueryDocumentDefinition(fragments, context.fragmentsPackage)
```
sav007: I guess better pass the whole context

sav007:
It's because sample uses maven dep, but your fix hasn't been published to maven yet.

sav007:
create `/apollo-android/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment` and put these files: https://github.com/sav007/apollo-android/tree/bug/fragment-imports/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment

marwanad:
@kassim for the compiler test fixtures we don't use `.graphql` files. You'd need to use the IR file generated from apollo-codegen,

`apollo-codegen generate AllStarships.graphql --schema schema.json --output TestQuery.json --target json`

Place this file in a folder under `apollo-compiler/src/test/graphql/com/example/non_referenced_child_fragments_imports` or something.

Then place, the expected generated files under that folder in the following format:
`TestQueryExpected.java`, `.....Expected.java`, `....Expected.java`.

(We should definitely be documenting that somewhere).

PR #147, opened 2017-02-03
Preparing for next release

felipecsl:
Just released and tagged Apollo Android 0.1.0

### Summary of changes

* Updated some outdated dependencies
* Updated the sample app to use the POJO converter
* Made the POJO converter turned on by default on the plugin
* Added some basic documentation to the readme


build.gradle
```diff
@@ -1,12 +1,12 @@
...
-  kotlinVersion = '1.0.5'
+  kotlinVersion = '1.0.6'
 
   compileSdkVersion = 25
   buildToolsVersion = '25.0.2'
 
   dep = [
-      androidPlugin: 'com.android.tools.build:gradle:2.3.0-beta2',
-      apolloPlugin: 'com.apollographql.android:gradle-plugin:0.0.1-SNAPSHOT',
+      androidPlugin: 'com.android.tools.build:gradle:2.3.0-beta3',
```
digitalbuddha: personally I'm not a fan of depending on a beta as they expire. This comes up if you ever want to be able to build a snapshot of an older tag.

felipecsl: Hmm good point. I could go back to 2.2.3 then

felipecsl: The sample actually wont build with 2.2.3 because of some weird manifest merger reason. I'll just keep it on the beta for now

Issue #77, opened 2016-12-23
String Pluralization library should be case insensitive


felipecsl:
For example, it's incorrectly singularizing `species` to `specy`

PR #149, opened 2017-02-03
bump pluralizer dep to fix uncountables pluralizing

marwanad:
I pushed a fix to the [pluralizer](https://github.com/cesarferreira/kotlin-pluralizer) repo to make comparisons case-insensitive.

Part of #77 (Part left would be updating the sample after the new snapshots are up)


felipecsl:
Wait shouldn't this have broken the current tests we have that contain the word "Specy"?

marwanad:
None of the tests refer to `Specy. Only the sample refers to that class for Autovalue.

screen shot 2017-02-03 at 1 13 22 am

felipecsl:
Solid, thanks, I was just too lazy to check 😄

PR #150, opened 2017-02-03
use proper singularized version of species in the sample

### LINKED ISSUE #77: String Pluralization library should be case insensitive

marwanad:
Should be 🍏  once it picks up the new plugin snapshot.

Closes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,9 @@ public final class DroidDetails implements Query<Operation.Variables> {
   public static class Data implements Operation.Data {
     private static final ResponseFieldMapper<Data> MAPPER = new ResponseFieldMapper<Data>() {
       private final Field[] FIELDS = {
-        Field.forObject("species", "species", null, true, new Field.ObjectReader<Specy>() {
-          @Override public Specy read(final ResponseReader reader) throws IOException {
-            return new Specy(reader);
+        Field.forObject("species", "species", null, true, new Field.ObjectReader<Species>() {
+          @Override public Species read(final ResponseReader reader) throws IOException {
+            return new Species(reader);
           }
         })
       };
@@ -83,7 +83,7 @@ public final class DroidDetails implements Query<Operation.Variables> {
           public void handle(final int fieldIndex, final Object value) throws IOException {
             switch (fieldIndex) {
               case 0: {
-                instance.species = (Specy) value;
+                instance.species = (Species) value;
                 break;
               }
             }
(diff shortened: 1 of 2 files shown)


marwanad:
Oh, sample now uses pojo now so no need for that.

I'll just verify that the generated `DroidDetails` has Species instead and update the PR with the README changes.

build.gradle
```diff
@@ -6,7 +6,7 @@ ext {
 
   dep = [
       androidPlugin: 'com.android.tools.build:gradle:2.3.0-beta3',
-      apolloPlugin: "com.apollographql.android:gradle-plugin:0.1.0",
+      apolloPlugin: "com.apollographql.android:gradle-plugin:0.2.0-SNAPSHOT",
```
marwanad: Is it worth having some kind of a `dev` branch that we build snapshots off instead and make that our target branch? and keep master for the releases?

README.md
```diff
@@ -134,16 +157,37 @@ public final class DroidDetails implements Query<Operation.Variables> {
...
       private @Nullable String name;
 
       private @Nullable String classification;
 
-      public Specy(ResponseReader reader) throws IOException {
+      public Species(ResponseReader reader) throws IOException {
         MAPPER.map(reader, this);
       }
 
+      public Species(@Nonnull String id, @Nullable String name, @Nullable String classification) {
```
marwanad: will get rid of this. I just copied off the generated class.

README.md
```diff
@@ -134,16 +157,37 @@ public final class DroidDetails implements Query<Operation.Variables> {
         }
       };
 
+      public static final Creator CREATOR = new Creator() {
```
marwanad: will get rid of this too.

build.gradle
felipecsl: I'm not a fan of having development outside of master, we can keep master always pointing to the SNAPSHOTS I think. This shouldn't be pointing to 0.1.0 in the first place

marwanad: Ahh, fair enough. That solves it.

sav007:
@marwanad but does compiler correctly generate interface names for Species ?
If not then example will mislead user as the outcome of generation is different

marwanad:
@sav007 It does for 0.2-SNAPSHOT, not for 0.1.0.

Shall we remove that/update the README to use the snapshot version?

Issue #142, opened 2017-02-01
Update ResponseFieldMapper generation to use Factory


sav007:
In order to support `AutoValue`, `Immutable` annotation based model generation, user should be able to register custom `Factory` that will return instances of his defined models. So the idea is to change `ResponseFieldMapper` to have `Factory` as a constructor parameter. Then during reading response mapper will call this factory to instantiate user defined models with parsed response fields. (The same as SQLDelight does https://github.com/square/sqldelight#autovalue).

To register these factories user must:

```
...
Retrofit retrofit = new Retrofit.Builder()
        ....
        .addConverterFactory(new ApolloConverterFactory.Builder()
            .withCustomTypeAdapter(CustomType.DATETIME, dateCustomTypeAdapter)
            .withMapper(AllPlanets.Data.class, new AllPlanets.Data.Mapper(AllPlanets.Data.FACTORY))
            .withMapper(ProductsWithDate.Data.class, new ProductsWithDate.Data.Mapper(ProductsWithDate.Data.FACTORY))
            .build())
...
```

where:
- `AllPlanets.Data` is the root of query data response
- `AllPlanets.Data.Mapper` will be generated by Apollo
- `AllPlanets.Data.FACTORY` user defined factory that creates `AllPlanets.Data` models

Issue #48, opened 2016-12-12
Add javadocs to public classes/interfaces


felipecsl:

VisheshVadhera:
Will be happy to work on this.

PR #157, opened 2017-02-04
Refactor API Field

sav007:
Part of #142

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
+++ b/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
@@ -4,72 +4,70 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 
-/** TODO */
-public final class Field {
+/** TODO add javadocs */
+public class Field {
   private final Type type;
   private final String responseName;
   private final String fieldName;
   private final Map<String, Object> arguments;
-  private final ObjectReader objectReader;
-  private final ListReader listReader;
   private final boolean optional;
-  private final ScalarType scalarType;
 
   public static Field forString(String responseName, String fieldName, Map<String, Object> arguments,
       boolean optional) {
-    return new Field(Type.STRING, responseName, fieldName, arguments, null, null, optional, null);
+    return new Field(Type.STRING, responseName, fieldName, arguments, optional);
   }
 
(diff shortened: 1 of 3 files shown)


apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
```diff
@@ -4,72 +4,70 @@
 import java.util.Collections;
 import java.util.Map;
 
-/** TODO */
-public final class Field {
+/** TODO add javadocs */
```
marwanad: we have #48, so we can get rid of that.

apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
```diff
@@ -4,72 +4,70 @@
...
   public static Field forBoolean(String responseName, String fieldName, Map<String, Object> arguments,
       boolean optional) {
-    return new Field(Type.BOOLEAN, responseName, fieldName, arguments, null, null, optional, null);
+    return new Field(Type.BOOLEAN, responseName, fieldName, arguments, optional);
   }
 
   public static <T> Field forObject(String responseName, String fieldName, Map<String, Object> arguments,
       boolean optional, ObjectReader<T> objectReader) {
-    return new Field(Type.OBJECT, responseName, fieldName, arguments, objectReader, null, optional, null);
+    return new ObjectField(responseName, fieldName, arguments, optional, objectReader);
```
marwanad: I kinda find it weird that a method in the parent class returns an instance of one of it subclasses. (unless there's a use-case outside this PR changes)

Wouldn't it be better just to have those subclasses `ObjectField, ScalarListField, CustomTypeField` defined outside and have those methods define there instead?

In both cases we'll have to do the cast [here](https://github.com/apollographql/apollo-android/pull/157/files#diff-fcfdd9c152da5f099c8263ad8783940cR52) anyways.

sav007: This is the one version of factory pattern (factory embedded into parent class) and we don't support any arbitrary field only that provided by this factory.
Plus `Field` constructor is private so you can't instantiate it.

The cast needs only for fields that has some additional information , like reader delegate etc.

PR #163, opened 2017-02-06
Update sample app withe the latest changes to compiler

sav007:


apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -31,10 +32,12 @@
         .baseUrl(BASE_URL)
         .client(okHttpClient)
         .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
-        .addConverterFactory(new ApolloConverterFactory.Builder().build())
+        .addConverterFactory(new ApolloConverterFactory.Builder()
+            .withResponseFieldMapper(FeedQuery.Data.class, new FeedQuery.Data.Mapper(FeedQuery.Data.FACTORY))
```
digitalbuddha: so sorry could you explain this line again.  Will we need to register every object or is this for a particular custom field

sav007: Yes user must provide Mapper instance for each query top level Data class. Converter then will use it to parse query response. So basically number of queries == number of mapper registrations

sav007: We will need to update readme in next release

digitalbuddha:
thanks lgtm!

Issue #161, opened 2017-02-05
Drop :apollo-converters:moshi support


sav007:
Apollo retrofit response converter based on Moshi library is no longer supported in favour of POJO converter. Drop module `:apollo-converters:moshi` and make POJO to be default one.
Rename module `:apollo-converters:pojo` -> `:apollo-converter`

kassim:
you mean request converter? the Moshi converter plugin doesn't have response conversion

I was about to post an issue about a namespace / class labelling issue (Moshi converter being `com.apollographql.android.ApolloConverterFactory` while POJO being `com.apollographql.android.converter.pojo.ApolloConverterFactory`) but equally support dropping of the Moshi converter

In my opinion it would be nice to have the default converter support the previous format of
`@POST("/") Observable<Response<FilmsData>> allFilms(@Body Films query);`
that the Moshi plugin provided, as this current format
`@POST("/") Observable<Response<Films.Data>> allFilms(@Body OperationRequest<Operation.Variables> query);` 
doesn't seem as presentable

sav007:
@kassim you can always create own class that extends `OperationRequest` like `FilmsRequest extends OperationRequest<Films.Variables>` and then pass  `allFilms(@Body FilmsRequest request)`

PR #164, opened 2017-02-07
Drop moshi converter

sav007:
part of #161 

POJO converter will be transformed to be default and only one response / request converter


felipecsl:
Did we decide to just not support interface-only code generation anymore?

sav007:
@felipecsl No, we still support interfaces, POJO converter can serve now both user defined models and POJO generated models

PR #165, opened 2017-02-07
Add support of handling custom scalar types to OperationRequest

### LINKED ISSUE #144: Add support of GraphQL custom scalar type adapters to InputObjectType
Gradle plugin provides option to specify mapping GraphQL custom scalar type to java type, plus converter allows to register user adapters to convert from string value of such type to java type and vice versa. But these adapters are used only in response parsing not for mutation requests (inside InputObjectType).

sav007:
This PR is part of #161 

Adds support of user defined scalar type adapters to OperationRequest serialization.
Internally converter uses moshi to serialize OperationRequest object (query text and variables).

Next final step of #161 is renaming :apollo-converters:pojo to -> :apollo-converter

Closes #144


apollo-converters/pojo/src/main/java/com/apollographql/android/converter/pojo/ApolloConverterFactory.java
```diff
@@ -21,11 +27,26 @@
 public final class ApolloConverterFactory extends Converter.Factory {
   private final Map<Type, ResponseFieldMapper> responseFieldMappers;
   private final Map<ScalarType, CustomTypeAdapter> customTypeAdapters;
+  private final Moshi moshi;
```
sav007: We use moshi to serialize query request that can have variables (with scalar or input object types).

digitalbuddha: I'm not crazy about having a dependency on moshi. I still feel like gson is the most popular converter on android. Having both including in classpath is extraneous methods.  Maybe we can make a convert factory type setup similar to what retrofit does?

sav007: Moshi lib is tiny: 457 methods only, we can eventually remove it and generate own serialization (the same as we did with ResponseReader), but for now to save time lets leave.

As for Gson, we could allow user to define own lib for serialization request but then user must  handle custom scalar type serialization in request by himself. So in this case he is going to have 2 places where define scalar type conversion: first for response (in `com.apollographql.android.converter.pojo.CustomTypeAdapter`), second for Gson.

In current implementation he defines these conversion in one place for both request / response: 
```
CustomTypeAdapter<Date> dateCustomTypeAdapter = new CustomTypeAdapter<Date>() {
      @Override public Date decode(String value) {
        try {
          return DATE_FORMAT.parse(value);
        } catch (ParseException e) {
          throw new RuntimeException(e);
        }
      }

      @Override public String encode(Date value) {
        return DATE_FORMAT.format(value);
      }
    };
....
new ApolloConverterFactory.Builder()
    .withCustomTypeAdapter(CustomType.DATETIME, dateCustomTypeAdapter)
```

digitalbuddha: I'm not saying to generate parsing. I'd like to be open to having a ConverterFacotory which can be a gson or moshi factory.  Apollo client should be agnostic to how the parsing is done but instead just delegate to a user defined parser factory.

PR #168, opened 2017-02-08
Move :apollo-converters:pojo to apollo-converter

sav007:
Almost there #161 

Renaming and moving POJO converter.


.travis.yml
```diff
@@ -14,7 +14,7 @@ jdk:
   - oraclejdk8
 
 script:
-  - ./gradlew clean check -x checkstyleTest --stacktrace
+  - ./gradlew clean check -x checkstyleTest -x :apollo-gradle-plugin:test --stacktrace
```
digitalbuddha: check should be build otherwise there is a chance that tests pass but the assembly of the aar will fail.

apollo-converter/gradle.properties
```diff
@@ -1,4 +1,4 @@
-POM_ARTIFACT_ID=converter-pojo
+POM_ARTIFACT_ID=converter
```
digitalbuddha: this will change the uploaded artifact name on maven central I believe

gradle.properties
```diff
@@ -1,5 +1,5 @@
 GROUP=com.apollographql.android
-VERSION_NAME=0.2.1-SNAPSHOT
+VERSION_NAME=0.2.2-SNAPSHOT
```
digitalbuddha: I dont think this is necessary if you do  ./gradlew build **--refresh dependencies**

apollo-converter/gradle.properties
sav007: yes this is intention

gradle.properties
sav007: then in sample app we should mention that, as it has dep to snapshot

.travis.yml
sav007: Sorry but what do you mean.
Because of we updated snapshot version, all plugin tests will be failed as they depends on published version of snapshot api to maven. This flag supposes to disable just running unit tests, but compilation is still enabled. So not sure why packaging should fail if the compilation succeed

digitalbuddha: Sorry I meant in general CI should be running Build task rather than Check test. If you only run check you leave possibility of the check passing but archiving failing due to packaging issue

Issue #169, opened 2017-02-08
Generation of nested structures with the same name results in compiler errors


nubstick27:
There appears to be an issue with nested structures which define the same name multiple times. 

Briefly, a structure such as:
```
mutation UpvotePost($postId: Int!) {
  upvotePost(postId: $postId) {
    ...PostDetails
  }
}
```

results in a generated class like:
```
public final class UpvotePost implements Mutation<UpvotePost.Variables> {
    public static class Data implements Operation.Data {
        public static class UpvotePost { ... }
    }
}
```

which has a compilation error.

sav007:
We are going to use the same logic for resolving dup names by adding `$` as suffix, so in this case we ate going to have `UpvotePost$`

PR #170, opened 2017-02-08
Introduce test breakage in input_object_type test

nubstick27:
There appears to be an issue where a nested structure that results in
multiple classes of the same name is not well-supported. This change to
the `input_object_type` test highlights the issue.

This breakage highlights issue #169


sav007:
@nubstick27 please provide the solution for your PR, otherwise master will be broken

nubstick27:
Ah! Sorry, I made this only to be referenced by #169. I was hoping for guidance and discussion before implementing a solution. I will close this for now.

Issue #166, opened 2017-02-07
Update sample app to be aligned with iOS


sav007:
See https://github.com/apollographql/frontpage-ios-app

nubstick27:
Hello, I would love to contribute for this Issue. I am thinking to port the `frontpage-ios-app` to the `apollo-sample` directory.

marwanad:
The sample iOS app uses the frontpage-server https://github.com/apollographql/frontpage-server which isn't hosted publicly. It would require running a local instance of that server.

Closing for now.

PR #171, opened 2017-02-09
Update sample app to be aligned with iOS

nubstick27:
Initial pass at porting the `frontend-ios-app` repository to the `apollo-sample` project.

I would love reviews and comments!

#166


apollo-sample/src/main/java/com/example/apollographql/sample/PostsService.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.apollographql.android.api.graphql.Response;
+import com.example.AllPosts;
+import com.example.Upvote;
+
+import io.reactivex.Observable;
+import retrofit2.http.Body;
+import retrofit2.http.POST;
+
+interface PostsService {
```
marwanad: nit: rename to `FrontPageService` since that's the schema we're using: https://github.com/apollographql/frontpage-server

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -15,10 +16,11 @@
 import retrofit2.converter.moshi.MoshiConverterFactory;
 
 public class SampleApplication extends Application {
-  private static final String BASE_URL = "https://githunt-api.herokuapp.com";
+  // Change localhost to your machine's local IP address when running from a device
+  private static final String BASE_URL = "http://10.6.1.232:8080";
```
martinbonnin: This feels complicated. Could we host a publicly available server somewhere ?

martinbonnin:
Love the direction this is taking. I would like it if the sample server had some custom scalar types like dates. It would help me and certainly others understand the work @sav007 is currently doing supporting them.

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
darrinps: Agree with Martin. This should be a well known server. Everything that can be done to make it easy on someone looking into this for the first time should be considered right?

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -23,39 +27,60 @@
...
+
+    final SampleApplication application = (SampleApplication) getApplication();
+
+    final Subject<Integer> upvoteObserver = PublishSubject.create();
+
+    final PostsAdapter adapter = new PostsAdapter(upvoteObserver);
+    responses.setAdapter(adapter);
+    responses.setLayoutManager(new LinearLayoutManager(this));
+
+    upvoteObserver.subscribe(new SimpleObserver<Integer>() {
```
sav007: I guess you should keep subscription and then unsubscribe it in `onDestroy` or even in `onStop`

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -23,39 +27,60 @@
...
+    });
+
+    application.postsService()
+        .allPosts(new AllPosts())
         .subscribeOn(Schedulers.io())
         .observeOn(AndroidSchedulers.mainThread())
-        .subscribe(new Observer<Response<FeedQuery.Data>>() {
-          @Override public void onSubscribe(Disposable d) {
-          }
+        .subscribe(new SimpleObserver<Response<AllPosts.Data>>() {
```
digitalbuddha: Need to unsubscribe in on destroy to prevent memory leak

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -23,39 +27,60 @@
...
           }
+        });
+  }
 
-          @Override public void onError(Throwable e) {
-            Log.e(TAG, "", e);
-            Toast.makeText(MainActivity.this, "onError(): " + e.getMessage(), Toast.LENGTH_LONG)
-                .show();
-          }
+  private abstract class SimpleObserver<T> implements Observer<T> {
```
digitalbuddha: I'd prefer to exclude this as it makes sample more complicated. I feel like some users might think that this part is a requirement.

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -23,39 +27,60 @@
...
+    application.postsService()
+        .allPosts(new AllPosts())
         .subscribeOn(Schedulers.io())
         .observeOn(AndroidSchedulers.mainThread())
-        .subscribe(new Observer<Response<FeedQuery.Data>>() {
-          @Override public void onSubscribe(Disposable d) {
-          }
+        .subscribe(new SimpleObserver<Response<AllPosts.Data>>() {
+          @Override public void onNext(Response<AllPosts.Data> response) {
+            txtResponse.setVisibility(View.GONE);
```
sav007: Be aware of potentially memory leaks here, as you have implicit ref to main activity, and pass this ref to background thread as observer, so main activity will be kept as long as this subscription is alive

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -23,39 +27,60 @@
...
-        .githuntFeed(new OperationRequest<>(
-            new FeedQuery(FeedQuery.Variables.builder()
-                .limit(10)
-                .type(FeedType.HOT)
-                .build())))
+    final RecyclerView responses = (RecyclerView) findViewById(R.id.responses);
+
+    final SampleApplication application = (SampleApplication) getApplication();
+
+    final Subject<Integer> upvoteObserver = PublishSubject.create();
```
sav007: Should it be called, publisher or observable? As it's not observer

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
sav007: Why we decided not not use Apollo example hosted on heroku ?

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
digitalbuddha: It feels unnecessary for the activity to create the subject. Might make sense to have adapter create subject and the activity can do adapter.getUpdater.subscribe

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -3,7 +3,8 @@
 import android.app.Application;
 
 import com.apollographql.android.converter.pojo.ApolloConverterFactory;
```
sav007: with latest PR merged this should be changed to:
com.apollographql.android.converter.ApolloConverterFactory

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -33,11 +35,12 @@
         .client(okHttpClient)
         .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
         .addConverterFactory(new ApolloConverterFactory.Builder()
-            .withResponseFieldMapper(FeedQuery.Data.class, new FeedQuery.Data.Mapper(FeedQuery.Data.FACTORY))
+            .withResponseFieldMapper(AllPosts.Data.class, new AllPosts.Data.Mapper(AllPosts.Data.FACTORY))
+            .withResponseFieldMapper(Upvote.Data.class, new Upvote.Data.Mapper(Upvote.Data.FACTORY))
             .build())
         .addConverterFactory(MoshiConverterFactory.create())
```
sav007: With latest changes this is not needed anymore

sav007:
@nubstick27 I fixed sample app, small changes

https://github.com/apollographql/apollo-android/pull/174

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
nubstick27: Ah, this wasn't an explicit decision so much as I copied the behavior from the `frontpage-ios-app`. the `10.x.x.x` IP is my mistake in checking in my local IP (the ios app has checked in `127.0.0.1`)

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
nubstick27: 👍

nubstick27:
I've tried to address the concerns. Thanks to everyone for feedback and review!

The server question is still up in the air; I would personally love to see a publicly hosted `frontpage-server`, but I don't know who should be responsible for that.

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -2,60 +2,129 @@
...
+    responses = (RecyclerView) findViewById(R.id.responses);
+
+    responses.setAdapter(postsAdapter);
+    responses.setLayoutManager(new LinearLayoutManager(this));
+  }
+
+  @Override protected void onResume() {
+    super.onResume();
+
+    final TextView txtResponse = MainActivity.this.txtResponse;
```
sav007: I guess you can drop `MainActivity.this`

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -2,60 +2,129 @@
...
+    if (txtResponse != null) {
```
sav007: Do we need to check for null ?

nubstick27: good point -- will remove in the onResume block at least.

sav007:
Great work, only one big concern about base url

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -14,10 +15,11 @@
 import retrofit2.Retrofit;
 
 public class SampleApplication extends Application {
-  private static final String BASE_URL = "https://githunt-api.herokuapp.com";
+  // Change localhost to your machine's local IP address when running from a device
+  private static final String BASE_URL = "http://127.0.0.1:8080";
```
nubstick27: It would be really nice. I agree that not everyone who wants to run android `apollo-sample` will want to install node + run `frontpage-server` themselves.

nubstick27:
@sav007 would you be ok with merging this as-is for now, until questions about public `frontpage-server` are answered?

sav007:
@nubstick27 we still deciding what to do, I guess we will merge it over weekends

nubstick27:
Thanks for the update!

digitalbuddha:
This pr seems stale, unless there are objections I am closing it in 24 hrs.

nubstick27:
Agreed, i think the concern over not requiring the developer to run a local server is valid. If there are no plans to host the frontpage server publicly, please close.

digitalbuddha:
Closing we can reopen later

PR #176, opened 2017-02-09
update the README.md to 0.2.1

martinbonnin:


sav007:
pls rebase

martinbonnin:
rebased

sav007:
Oh, I guess these changes will be out of date as soon as we merge this pr #171

martinbonnin:
Ideally, we would setup a frontpage public server and use that in the README.md too

README.md
```diff
@@ -62,140 +86,486 @@ public final class DroidDetails implements Query<Operation.Variables> {
...
-            .enqueue(new Callback<Response<DroidDetails.Data>>() {
-              @Override
-              public void onResponse(Call<Response<DroidDetails.Data>> call,
-                  retrofit2.Response<Response<DroidDetails.Data>> response) {
-                // use the graphql response which is now in Java Models.
-              }
+```
+    okHttpClient = new OkHttpClient.Builder()
+        .cache(new Cache(new File(getCacheDir(), "okhttp"), 10 * 1024))
+        .addNetworkInterceptor(loggingInterceptor)
```
digitalbuddha: not sure if this is needed. Ideally example would be as simple as possible.

sav007: and this is not make sense as OkHttp doesn't support caching for POST methods

martinbonnin: Definitely not needed, I don't know where this came from. I just removed it. Thanks for noticing it.

martinbonnin:
Something else is I'm not sure we should include the whole generated code. Maybe just a small sample to explain what the compiler is doing under the hood but having all these lines seems overkill.

digitalbuddha:
Maybe link to generated code instead

sav007:
Yeah, generated code is pretty much verbose, I guess we need to cut it.
We can't put a link as it's generated code and not under the git

digitalbuddha:
Hey so my last pr was more of a bandaid. Could you rebase this as you had some great changes too.

martinbonnin:
I just forced pushed a new version (not sure this is the correct way to do...). It adds the "custom scalar types" and "generateClasses" section. @digitalbuddha  I think they were the only things not in your pr already.

PR #177, opened 2017-02-09
Add fallback to Object in case if there is no custom scalar type mapping specified

### LINKED ISSUE #173: Default to `String` for custom types that do not have a mapping
Would be nice if `DateTime` for an example could by default be readable as a string for users that did not register any custom adapters.

sav007:
Closes #173 

If user didn't specify custom type mapping, Object will be used as java type for fields with GraphQL custom scalar.
As for the parsing response values for such fields it will be decided by json reader, that supports by default only 3 types: Boolean, String, BigDecimal.

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/pojo_custom_scalar_type/TestQueryExpected.java
+++ b/apollo-compiler/src/test/graphql/com/example/pojo_custom_scalar_type/TestQueryExpected.java
@@ -79,8 +79,9 @@ public final class TestQuery implements Query<Operation.Variables> {
       public static final Creator CREATOR = new Creator() {
         @Override
         public Hero create(@Nonnull String name, @Nonnull Date birthDate,
-            @Nonnull List<? extends Date> appearanceDates) {
-          return new Hero(name, birthDate, appearanceDates);
+            @Nonnull List<? extends Date> appearanceDates,
+            @Nonnull Object fieldWithUnsupportedType) {
+          return new Hero(name, birthDate, appearanceDates, fieldWithUnsupportedType);
         }
       };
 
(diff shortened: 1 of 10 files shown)


sav007:
+ @martinbonnin

martinbonnin:
Thanks :)

apollo-compiler/src/test/graphql/com/example/pojo_custom_scalar_type/TestQueryExpected.java
```diff
@@ -161,10 +171,14 @@ public void handle(final int fieldIndex, final Object value) throws IOException
                   contentValues.appearanceDates = (List<? extends Date>) value;
                   break;
                 }
+                case 3: {
+                  contentValues.fieldWithUnsupportedType = (Object) value;
+                  break;
+                }
```
brianPlummer: would be good to have a default label...nitpick

apollo-converter/src/test/java/com/apollographql/android/converter/ProductsWithUnsupportedCustomScalarTypes.java
```diff
@@ -0,0 +1,480 @@
...
+            return CREATOR;
+          }
+
+          @Override
+          public Edge.Factory edgeFactory() {
+            return Edge.FACTORY;
+          }
+        };
+
+        private @Nonnull List<? extends Edge> edges;
```
brianPlummer: can be final

apollo-converter/src/test/java/com/apollographql/android/converter/ProductsWithUnsupportedCustomScalarTypes.java
```diff
@@ -0,0 +1,480 @@
...
+          return CREATOR;
+        }
+
+        @Override
+        public Product.Factory productFactory() {
+          return Product.FACTORY;
+        }
+      };
+
+      private @Nonnull Product products;
```
brianPlummer: can be final

apollo-compiler/src/test/graphql/com/example/pojo_custom_scalar_type/TestQueryExpected.java
sav007: sorry what do you mean?

digitalbuddha: A default case to the switch

sav007: agh, well this code is generated, so we guarantee that there won't be default case

brianPlummer: gotcha! :)

PR #178, opened 2017-02-10
Enable Mapper class generation for abstract models

sav007:
Most of this PR is update test fixtures

@brianPlummer


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/SchemaTypeSpecBuilder.kt
```diff
@@ -40,10 +41,7 @@ class SchemaTypeSpecBuilder(
             it
                 .withValueInitConstructor()
                 .withCreatorImplementation()
-                .withFactoryImplementation()
-                .toBuilder()
-                .addType(mapper)
-                .build()
+                .withFactoryImplementation(exclude = listOf(mapper.name))
```
marwanad: making sure that was intended,  the `toBuilder().build()` calls are no longer needed?

sav007: yes, no needed withFactoryImplementation will handle this internally

PR #191, opened 2017-02-13
Fix generation of nested structures with the same name results in compiler errors

### LINKED ISSUE #169: Generation of nested structures with the same name results in compiler errors

sav007:
Closes #169

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/unique_type_name/HeroDetailQueryExpected.java
+++ b/apollo-compiler/src/test/graphql/com/example/unique_type_name/HeroDetailQueryExpected.java
@@ -18,9 +18,9 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 @Generated("Apollo GraphQL")
-public final class TestQuery implements Query<Operation.Variables> {
-  public static final String OPERATION_DEFINITION = "query TestQuery {\n"
-      + "  hero {\n"
+public final class HeroDetailQuery implements Query<Operation.Variables> {
+  public static final String OPERATION_DEFINITION = "query HeroDetailQuery {\n"
+      + "  heroDetailQuery {\n"
       + "    __typename\n"
       + "    name\n"
       + "    friends {\n"
@@ -46,7 +46,7 @@ public final class TestQuery implements Query<Operation.Variables> {
 
   private final Operation.Variables variables;
 
-  public TestQuery() {
+  public HeroDetailQuery() {
     this.variables = Operation.EMPTY_VARIABLES;
   }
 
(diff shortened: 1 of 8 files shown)


apollo-compiler/src/test/graphql/com/example/unique_type_name/HeroDetailQueryExpected.java
```diff
@@ -61,9 +61,9 @@ public String queryDocument() {
   }
 
   public interface Data extends Operation.Data {
-    @Nullable Hero hero();
+    @Nullable HeroDetailQuery$ heroDetailQuery();
```
digitalbuddha: what happens if nested twice?  tbh I'd prefer `HeroDetailQuery$$1`

sav007: It will be `HeroDetailQuery$$`

digitalbuddha: And just for clarity 5 would be Query$$$$$ so just as example imagine you see following in logs... 
NullPointerExxeption: Foo$$$$ is null 

Will you be able to tell without dragging your finger in screen whether that was 4 or 5 dollar signs. 
I don't feel strongly about it but can see it leading to confusing traces. As example rxjava does Action0  -  Action9

sav007: Not sure why we initially decided to go with $, but it's really easy to change in compiler. But the idea that you in most cases you won't have such issue with more than 2-3 nested classes with the same name. But I don't have any strong opinion for this as Hero4 and Hero$$$$ from my perspective is the same evil, and for sure I will try to avoid such queries with the same nested names (will use aliases)

sav007: @digitalbuddha so are we changing $ to number or leaving as is?

digitalbuddha: I prefer number but you make the call :-)

digitalbuddha:
Now you can blame me if people don't like it ☺️

digitalbuddha:
```
com.apollographql.android.compiler.CodeGenTest >  generateExpectedClasses[pojo_fragment_with_inline_fragment] FAILED
    org.junit.ComparisonFailure at CodegenTest.kt:31
```

PR #202, opened 2017-02-16
Generate map initialized with POJO Mappers

### LINKED ISSUE #198: Auto Generate ApolloConverterFactory with response mappers
The setup process seems a little unwieldy to have to register a new ResponseMapper for each new query.  It would be nice to leverage code gen to make the process easier.  Initial proposal is to generate `MyApolloConverterFactory` which will already have the response mappers registered. This way adding a new query does not mean that I need to make changes to my restadapter.

sav007:
Closes #198


martinbonnin:
Not sure if it's an error yet but my generated `ResponseFieldMappers.java` has no package declaration so I'm not sure how I am supposed to include it :-/

Issue #209, opened 2017-02-17
Make Code Gen work on windows


digitalbuddha:
Windows and npm are not playing well together.  What I think is happening:
Gradle plugin installs its own node without root privileges, when it tries to run the node task we get a sad permission error.  This is the command that is being run:
```"install" "apollo-codegen@0.10.5" "--save" "--save-exact"```

and the glorious errors that this results in :-)

```
10:14:27.360 [ERROR] [system.err] npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Users\\Home\\Desktop\\apollo-android-master\\apollo-android-master\\apollo-sample\\.gradle\\nodejs\\node-v6.7.0-win-x64\\node.exe" "C:\\Users\\Home\\Desktop\\apollo-android-master\\apollo-android-master\\apollo-sample\\.gradle\\nodejs\\node-v6.7.0-win-x64\\node_modules\\npm\\bin\\npm-cli.js" "install" "apollo-codegen@0.10.5" "--save" "--save-exact"
npm ERR! node v6.7.0
npm ERR! npm  v3.10.3
npm ERR! path C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\node_modules\apollo-codegen
npm ERR! code EPERM
npm ERR! errno -4048
npm ERR! syscall rename

npm ERR! Error: EPERM: operation not permitted, rename 'C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\node_modules\apollo-codegen' -> 'C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\node_modules\.apollo-codegen.DELETE'
npm ERR!     at moveAway (C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\.gradle\nodejs\node-v6.7.0-win-x64\node_modules\npm\lib\install\action\finalize.js:38:5)
npm ERR!     at destStatted (C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\.gradle\nodejs\node-v6.7.0-win-x64\node_modules\npm\lib\install\action\finalize.js:27:7)
npm ERR!     at FSReqWrap.oncomplete (fs.js:123:15)
npm ERR! 
[...]

PR #217, opened 2017-02-19
fix apollo-codegen install task up-to-date check

marwanad:
This is an attempt at fixing the errors on windows machines (Working towards #209)

As seen on several issues on GitHub, running `npm cache clean` before `npm install` can fix the errors in #209 stack-trace. 

Running that doesn't come at an extra cost since we're using a local fresh npm installation anyways and `ApolloCodegenInstallTask` runs only one and is up-to-date after.
(Unfortunately I can't verify that since I don't have a windows machine. If someone is able to verify, just checkout this branch and running `./gradlew apollo-gradle-plugin:test` should be sufficient)

### DIFF
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
@@ -42,7 +42,7 @@ class ApolloPlugin implements Plugin<Project> {
     project.getGradle().addListener(new DependencyResolutionListener() {
       @Override
       void beforeResolve(ResolvableDependencies resolvableDependencies) {
-        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:${VersionKt.VERSION}"))
+        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:0.2.1"))
         project.getGradle().removeListener(this)
       }
 
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
@@ -42,7 +42,7 @@ class ApolloPlugin implements Plugin<Project> {
     project.getGradle().addListener(new DependencyResolutionListener() {
       @Override
       void beforeResolve(ResolvableDependencies resolvableDependencies) {
-        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:0.2.1"))
+        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:${VersionKt.VERSION}"))
         project.getGradle().removeListener(this)
       }
 
--- a/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
+++ b/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
@@ -30,14 +30,14 @@ public class ApolloCodeGenInstallTask extends NpmTask {
     installDir = getProject().file(INSTALL_DIR);
 
     final File apolloPackageFile = getProject().file("package.json");
-    final boolean isDifferentCodegenVersion = isDifferentCodegenVersion(getApolloVersion());
+    final boolean isSameCodegenVersion = isSameApolloCodegenVersion(getApolloVersion());
 
-    if (isDifferentCodegenVersion) {
-      Utils.deleteDirectory(new File(INSTALL_DIR));
+    if (!isSameCodegenVersion) {
+      Utils.deleteDirectory(installDir);
     }
     getOutputs().upToDateWhen(new Spec<Task>() {
       public boolean isSatisfiedBy(Task element) {
-        return apolloPackageFile.isFile() && isDifferentCodegenVersion;
+        return apolloPackageFile.isFile() && isSameCodegenVersion;
       }
     });
 
@@ -71,7 +71,7 @@ public class ApolloCodeGenInstallTask extends NpmTask {
       return null;
     }
   }
-  private boolean isDifferentCodegenVersion(String packageVersion) {
+  private boolean isSameApolloCodegenVersion(String packageVersion) {
     return packageVersion != null && packageVersion.equals(APOLLOCODEGEN_VERSION);
   }
 
--- a/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
+++ b/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
@@ -1,18 +1,15 @@
 package com.apollographql.android.gradle;
 
+import com.google.common.collect.Lists;
+
 import java.io.File;
-import java.io.FileReader;
-import java.io.FileWriter;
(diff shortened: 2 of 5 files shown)


digitalbuddha:
Could we instead not use local install?

marwanad:
Yeah, we can check if node is installed globally first. If not then we use the local installation. If that's the case, we might as well do use the global `apollo-codegen` if one is installed.

Problem is we'd need to make sure that the global installed node version works with the apollo-codegen but it shouldn't be too bad.

digitalbuddha:
So the issue is that we install node but that node instance does have admin privileges as a result it can't clean some folder. On last  call we discussed adding instructions for user to install node properly and then use apollo code gen. If version is wrong we can throw an error with instructions link

marwanad:
One of the issues `https://github.com/npm/npm/issues/10826`  suggested the clean cache fix and quoting https://blogs.msdn.microsoft.com/matt-harrington/2012/02/23/how-to-fix-node-js-npm-permission-problems/ 
> It turns out that you don’t have to run the command again as Administrator, and doing so won’t fix the problem.  Try npm cache clean first. 


I'd agree that it kinda doesn't make sense but I doubt running the global one would make a difference since you're still not running it with admin privileges and it's more likely to be installed in some protected user directory under C:/.

IMO, I'd prefer if we can do all the setup ourselves (if we can guarantee the windows fix) since there's less friction involved and I feel it's less error-prone.

Plus, the flow of letting the user install node, apollo-codegen and then place the generated IR in a folder could be a pain to do on CI because the IR would have to go under VS and it kind of doesn't make sense for it to be there because it's an IR (unless you have a script to do the IR generation).

digitalbuddha:
@martijnwalraven @sav007 want to chime in?  I don't believe clean cache worked for me

digitalbuddha:
not sure what this means
```
:apollo-gradle-plugin:test
com.apollographql.android.gradle.integration.ApolloPluginBasicAndroidTest > builds successfully and generates expected outputs FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure at ApolloPluginBasicAndroidTest.groovy:27
22 tests completed, 1 failed
```

marwanad:
@digitalbuddha actually scratch that. I have another hypothesis to why that's happening.

So, I took another look, ran `./gradlew apollo-sample:build` and took a look at the `installApolloCodegen` in particular. The task never goes to up-to-date even after multiple runs which is a bug. 

 It would then attempt to overwrite the installed on every single time. I think this explains this part in the stacktrace:

`npm ERR! Error: EPERM: operation not permitted, rename 'C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\node_modules\apollo-codegen' -> 'C:\Users\Home\Desktop\apollo-android-master\apollo-android-master\apollo-sample\node_modules\.apollo-codegen.DELETE'`

It's attempting to delete the old one by renaming it and reinstalling again. I think this got introduced when we moved the task to java.

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
```diff
@@ -7,46 +7,39 @@
...
-import com.squareup.moshi.JsonReader;
+import com.moowork.gradle.node.npm.NpmInstallTask;
 import com.squareup.moshi.JsonWriter;
 import com.squareup.moshi.JsonAdapter;
 import com.squareup.moshi.Moshi;
 
 import okio.Okio;
 
-public class ApolloCodeGenInstallTask extends NpmTask {
+public class ApolloCodeGenInstallTask extends NpmInstallTask {
```
marwanad: extend NpmInstallTask directly and define the dependencies in the package.json file since we're creating one anyways.

This basically now runs `npm install` which would install packages under the dependencies object in the package.json file.

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
```diff
@@ -7,46 +7,39 @@
...
     setDescription("Runs npm install for apollo-codegen");
-    installDir = getProject().file(INSTALL_DIR);
 
     final File apolloPackageFile = getProject().file("package.json");
     final String apolloVersion = getApolloVersion();
 
     getOutputs().upToDateWhen(new Spec<Task>() {
       public boolean isSatisfiedBy(Task element) {
-        return !(apolloPackageFile.isFile() || (apolloVersion != null && apolloVersion.equals(APOLLOCODEGEN_VERSION)));
+        return apolloVersion != null && apolloVersion.equals(APOLLOCODEGEN_VERSION);
```
marwanad: This check was wrong before. So the task never got outdated.

npm attempts to overwrite to rename and delete the files every time which doesn't make windows happy.

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
```diff
@@ -7,46 +7,39 @@
...
     setDescription("Runs npm install for apollo-codegen");
-    installDir = getProject().file(INSTALL_DIR);
 
     final File apolloPackageFile = getProject().file("package.json");
     final String apolloVersion = getApolloVersion();
 
     getOutputs().upToDateWhen(new Spec<Task>() {
       public boolean isSatisfiedBy(Task element) {
-        return !(apolloPackageFile.isFile() || (apolloVersion != null && apolloVersion.equals(APOLLOCODEGEN_VERSION)));
+        return apolloPackageFile.isFile() && apolloVersion != null && apolloVersion.equals(APOLLOCODEGEN_VERSION);
```
marwanad: This check was wrong before (it shouldn't be negated). So the task never got outdated.

npm attempts to overwrite to rename and delete the files every time which doesn't make windows happy.

apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
```diff
@@ -42,7 +42,7 @@ class ApolloPlugin implements Plugin<Project> {
     project.getGradle().addListener(new DependencyResolutionListener() {
       @Override
       void beforeResolve(ResolvableDependencies resolvableDependencies) {
-        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:${VersionKt.VERSION}"))
+        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:0.2.1"))
```
digitalbuddha: did you mean to hardcode this?

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
digitalbuddha: this makes a lot of sense. Great find!

apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
marwanad: VersionKt.VERSION would resolve to 0.2.2-SNAPSHOT which we don't have yet so the build would fail.

We'd need another PR to change that after the snapshots are uploaded.

marwanad:
We should probably get this through (even if it doesn't fix the windows issue) because it still fixes another bug.

digitalbuddha:
👍

PR #218, opened 2017-02-19
Feature 216/move converter

### LINKED ISSUE #216: Move converter to Apollo client
Part of master ticket #214 

- [ ] Merge :apollo-converter into :apollo-runtime.
- [ ] Rename `Apollo` to `ApolloClient`
- [ ] Change `ApolloClient` initialization to accept user `OkHttpClient` instead `Retrofit`
- [ ] Add graphql server endpoint as a parameter to `ApolloClient` initialization
- [ ] Introduce API calls to perform query execution in `ApolloClient`:
  - `void fetch(Query, Callback<Operation.Data>)` - async (aligned with iOS API)
  - `void fetch(Query, Callback<Operation.Data>, Handler callbackHandler)` - async (aligned with iOS API)
  - `void perform(Mutation, Callback<Operation.Data>)` - async (aligned with iOS API)
  - `void perform(Mutation, Callback<Operation.Data>, Handler callbackHandler)` - async (aligned with iOS API)
  - `Operation.Data execute(Operation)` - sync (this call user can wrap with Rx)
- [ ] Merge logic from `ApolloRequestBodyConverter` and `ApolloResponseBodyConverter` into `ApolloClient`

sav007:
Closes #216 

First version of `ApolloClient`.

Api:
`<T extends Operation> ApolloCall newCall(@Nonnull T operation)`

```
public interface ApolloCall {

  void cancel();

  <T extends Operation.Data> Response<T> execute() throws IOException;

  <T extends Operation.Data> ApolloCall enqueue(@Nullable Callback<T> callback);

  ApolloCall clone();

  public interface Callback<T extends Operation.Data> {
    void onResponse(Response<T> response);

    void onFailure(Exception e);
  }
}
```

Usage (will be scheduled on okhttp3.Dispatcher#executorService):
```
apolloClient.newCall(new AllPlanets()).enqueue(new ApolloCall.Callback<AllPlanets.Data>() {
      @Override public void onResponse(Response<AllPlanets.Data> response) {
      ...
      }

      @Override public void onFailure(Exception e) {
      ...
      }
    });
```
Or (sync call)
```
Response<AllPlanets.Data> body = apolloClient.newCall(new AllPlanets()).execute();
```


apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java
```diff
@@ -0,0 +1,95 @@
...
+      return this;
+    }
+
+    public <T> Builder withCustomTypeAdapter(@Nonnull ScalarType scalarType,
+        @Nonnull final CustomTypeAdapter<T> customTypeAdapter) {
+      customTypeAdapters.put(scalarType, customTypeAdapter);
+      return this;
+    }
+
+    public <T> Builder withResponseFieldMapper(@Nonnull Class type,
```
sav007: Will be removed in next PR, as we drop support of interfaces

apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java
```diff
@@ -0,0 +1,95 @@
...
+        @Nonnull ResponseFieldMapper<T> responseFieldMapper) {
+      responseFieldMappers.put(type, responseFieldMapper);
+      return this;
+    }
+
+    public Builder withResponseFieldMappers(@Nonnull Map<Type, ResponseFieldMapper> mappers) {
```
sav007: Will be removed in next PR, as we drop support of interfaces

apollo-runtime/src/main/java/com/apollographql/android/HttpOperationRequest.java
```diff
@@ -0,0 +1,121 @@
...
+    } catch (Exception e) {
+      if (callback != null) {
+        callback.onFailure(e);
+      }
+      return this;
+    }
+
+    Call httpCall = callFactory.newCall(request);
+    if (callRef.compareAndSet(null, httpCall)) {
+      httpCall.enqueue(new okhttp3.Callback() {
```
sav007: So it will be scheduled to be executed on `okhttp3.Dispatcher` that has own `ExecutorService`

apollo-runtime/src/test/java/com/apollographql/android/converter/IntegrationTest.java
```diff
@@ -68,25 +54,22 @@
...
-        .addConverterFactory(new ApolloConverterFactory.Builder()
-            .withResponseFieldMapper(AllPlanets.Data.class, new AllPlanets.Data.Mapper(AllPlanets.Data.FACTORY))
-            .withResponseFieldMapper(ProductsWithDate.Data.class, new ProductsWithDate.Data.Mapper(ProductsWithDate
-                .Data.FACTORY))
-            .withResponseFieldMapper(ProductsWithUnsupportedCustomScalarTypes.Data.class,
-                new ProductsWithUnsupportedCustomScalarTypes.Data.Mapper(ProductsWithUnsupportedCustomScalarTypes.Data.FACTORY))
-            .withCustomTypeAdapter(CustomType.DATETIME, dateCustomTypeAdapter)
-            .build())
+        .okHttpClient(new OkHttpClient.Builder().build())
+        .withResponseFieldMapper(AllPlanets.class, new AllPlanets.Data.Mapper(AllPlanets.Data.FACTORY))
```
sav007: with next PR these mapper registrations won't be needed at all.

apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.annotation.Nonnull;
+
+import okhttp3.Call;
+import okhttp3.HttpUrl;
+import okhttp3.OkHttpClient;
+
+import static com.apollographql.android.Utils.checkNotNull;
+
+/** TODO */
+public final class ApolloClient {
```
digitalbuddha: Should implement something like 
```
interface Factory {
    ApolloCall newCall(Operation operation);
  }
```

This would keep client stateless. Instead `ApolloCall` is an interface similar to `Okhttp3.Call` which can actually do exection

apollo-runtime/src/main/java/com/apollographql/android/HttpOperationRequest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+import okhttp3.Call;
+import okhttp3.MediaType;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okio.Buffer;
+
+final class HttpOperationRequest<T extends Operation.Data> implements OperationRequest<T> {
```
digitalbuddha: Let's move all this to a `RealApolloCall` which implements `ApolloCall`, you can model that after 
```
public interface Call extends Cloneable {
  /** Returns the original request that initiated this call. */
  Request request();

  /**
   * Invokes the request immediately, and blocks until the response can be processed or is in
   * error.
   *
   * <p>The caller may read the response body with the response's {@link Response#body} method. To
   * avoid leaking resources callers must {@linkplain ResponseBody close the response body}.
   *
   * <p>Note that transport-layer success (receiving a HTTP response code, headers and body) does
   * not necessarily indicate application-layer success: {@code response} may still indicate an
   * unhappy HTTP response code like 404 or 500.
   *
   * @throws IOException if the request could not be executed due to cancellation, a connectivity
   * problem or timeout. Because networks can fail during an exchange, it is possible that the
   * remote server accepted the request before the failure.
   * @throws IllegalStateException when the call has already been executed.
   */
  Response execute() throws IOException;

  /**
   * Schedules the request to be executed at some point in the future.
   *
   * <p>The {@link OkHttpClient#dispatcher dispatcher} defines when the request will run: usually
   * immediately unless there are several other requests currently being executed.
   *
   * <p>This client will later call back {@code responseCallback} with either an HTTP response or a
   * failure exception.
   *
   * @throws IllegalStateException when the call has already been executed.
   */
  void enqueue(Callback responseCallback);

  /** Cancels the request, if possible. Requests that are already complete cannot be canceled. */
  void cancel();

  /**
   * Returns true if this call has been either {@linkplain #execute() executed} or {@linkplain
   * #enqueue(Callback) enqueued}. It is an error to execute a call more than once.
   */
  boolean isExecuted();

  boolean isCanceled();

  /**
   * Create a new, identical call to this one which can be enqueued or executed even if this call
   * has already been.
   */
  Call clone();

  interface Factory {
    Call newCall(Request request);
  }
}
```

apollo-runtime/src/main/java/com/apollographql/android/OperationRequest.java
```diff
@@ -0,0 +1,23 @@
+package com.apollographql.android;
+
+import com.apollographql.android.api.graphql.Operation;
+import com.apollographql.android.api.graphql.Response;
+
+import java.io.IOException;
+
+import javax.annotation.Nullable;
+
+public interface OperationRequest<T extends Operation.Data> {
```
digitalbuddha: `ApolloCall`

apollo-runtime/src/main/java/com/apollographql/android/HttpOperationRequest.java
```diff
@@ -0,0 +1,121 @@
...
+      });
+      return this;
+    } else {
+      throw new IllegalStateException("Already Executed");
+    }
+  }
+
+  private Response<T> parse(okhttp3.Response response) throws IOException {
+    int code = response.code();
+    if (code < 200 || code >= 300) {
```
brianPlummer: you probably could use:  https://square.github.io/okhttp/3.x/okhttp/okhttp3/Response.html#isSuccessful--

brianPlummer: purposefully excluding 300 and up...my bad...

apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java
sav007: Do we need to make it implement Factory, now?

apollo-api/src/main/java/com/apollographql/android/api/graphql/Response.java
```diff
@@ -18,6 +21,10 @@ public boolean isSuccessful() {
     return errors == null || errors.isEmpty();
   }
 
+  @Nonnull public Operation getOperation() {
```
digitalbuddha: nitpick - if there is no setOperation I don't think the `get` prefix is necessary

apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Override public <T extends Operation.Data> Response<T> execute() throws IOException {
+    RequestBody requestBody = requestBody(operation);
+    Request request = baseRequest.newBuilder()
+        .post(requestBody)
+        .build();
+
+    Call call = callFactory.newCall(request);
+    if (callRef.compareAndSet(null, call)) {
```
digitalbuddha: is this necessary with implementation of call.execute being 
```
 @Override public Response execute() throws IOException {
    synchronized (this) {
      if (executed) throw new IllegalStateException("Already Executed");
      executed = true;
    }
    captureCallStackTrace();
    try {
      client.dispatcher().executed(this);
      Response result = getResponseWithInterceptorChain();
      if (result == null) throw new IOException("Canceled");
      return result;
    } finally {
      client.dispatcher().finished(this);
    }
  }
````

File: apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java (hunk 5 above)
Comment (digitalbuddha) [inline apollo-runtime/src/main/java/com/apollographql/android/ApolloClient.java]: Don't need but would be nice to have interface that defines that ApolloClient is a factory that creates ApolloCall.  In case we ever want to migrate to another Client that can also produce ApolooCall

File: apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java (hunk 10 above)
Comment (sav007) [inline apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java]: yes we need this as RealApolloCall keeps ref on okhttp3.Call in order to cancel request.

File: apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy [hunk 11]
```diff
@@ -42,7 +42,7 @@ class ApolloPlugin implements Plugin<Project> {
     project.getGradle().addListener(new DependencyResolutionListener() {
       @Override
       void beforeResolve(ResolvableDependencies resolvableDependencies) {
-        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:${VersionKt.VERSION}"))
+        compileDepSet.add(project.dependencies.create("com.apollographql.android:api:0.2.1"))
```
sav007: Just to make plugin test happy, as now they fails

digitalbuddha: ok but can't merge this way or it will make snapshot like this

apollo-runtime/build.gradle
```diff
@@ -32,11 +32,19 @@ android {
...
+
   testCompile dep.junit
   testCompile dep.truth
+  testCompile dep.mockWebServer
 }
 
 apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
+
+tasks.withType(Checkstyle) {
+  exclude '**/BufferedSourceJsonReader.java'
```
digitalbuddha: why excluded?

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.IOException;
+
+import javax.annotation.Nullable;
+
+public interface ApolloCall {
+
+  void cancel();
+
+  <T extends Operation.Data> Response<T> execute() throws IOException;
```
digitalbuddha: Nullability annotation on return value

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -0,0 +1,25 @@
...
+
+  <T extends Operation.Data> ApolloCall enqueue(@Nullable Callback<T> callback);
+
+  ApolloCall clone();
```
digitalbuddha: @NonNull

apollo-runtime/build.gradle
sav007: Because checkstyle will fail and BufferedSourceJsonReader was copied from retrofit as is without modification (we need it for ResponseReader)

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public interface Callback<T extends Operation.Data> {
+    void onResponse(Response<T> response);
```
digitalbuddha: nullability on both methods

apollo-runtime/src/main/java/com/apollographql/android/ApolloCallFactory.java
```diff
@@ -0,0 +1,9 @@
+package com.apollographql.android;
+
+import com.apollographql.android.api.graphql.Operation;
+
+import javax.annotation.Nonnull;
+
+public interface ApolloCallFactory {
+  <T extends Operation> ApolloCall newCall(@Nonnull T operation);
```
digitalbuddha: nullability

apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Override public void cancel() {
+    Call call = httpCall;
+    if (call != null) {
+      call.cancel();
+    }
+  }
+
+  @Override public <T extends Operation.Data> Response<T> execute() throws IOException {
```
digitalbuddha: nullability (I'm really calling out only public api methods)

apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java
```diff
@@ -0,0 +1,136 @@
...
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already Executed");
+      executed = true;
+    }
+    httpCall = httpCallFactory.newCall(httpRequest(operation));
+    return parseHttpResponse(httpCall.execute());
+  }
+
+  @Override public <T extends Operation.Data> ApolloCall enqueue(final Callback<T> callback) {
```
digitalbuddha: same

digitalbuddha:
👍 @brianPlummer mind taking a 1 over again?

build.gradle
```diff
@@ -31,7 +29,8 @@ ext {
       spock: 'org.spockframework:spock-core:0.7-groovy-2.0',
       truth: 'com.google.truth:truth:0.30',
       mockWebServer: 'com.squareup.okhttp3:mockwebserver:3.5.0',
-      okhttpLoggingInterceptor: 'com.squareup.okhttp3:logging-interceptor:3.5.0'
+      okhttpLoggingInterceptor: 'com.squareup.okhttp3:logging-interceptor:3.5.0',
```
brianPlummer: this should be 3.6.0 as well....feel free to leave it if you want...i can make this an okhttp version var

brianPlummer:
+1 looks great! good work! :)

PR #223, opened 2017-02-21
Simplify mapper generation

sav007:
The purpose of this PR is to simplify of Mapper class generation as we drop interface generation support. We don't need to generate `Creators`, `Factories`.

Plus we don't need to register mapper in `ApolloClient.Builder`

This PR affects all test fixtures, that's why it so big.


brianPlummer:
Looks Good! great work!  qq.....regarding toLowerCase/toUpperCase().......is there any concern with going with the default Locale?  i realize this was not implemented with this pr....just curious if it is something that we should worry about going forward........

sav007:
@brianPlummer I'm sorry but what do you mean ` toLowerCase/toUpperCase()` exactly ?

brianPlummer:
@sav007 this isn't an issue for this pr exactly.  For the generated code:  https://github.com/apollographql/apollo-android/pull/223/files#diff-641ea2d2de2c5a892122a0978e65470eR81   we are using toLowerCase() without specifying a locale.....the system in this situation will go with Locale.getDefault() this could have unintended consequences if the device is in other locales potentially......does it make sense to output Locale.English?  i'm not really what the best course of action, or if it is really a concern...just something that occurred to me as a potential issue... :) i hope this helps :)

digitalbuddha:
Lgtm

sav007:
@brianPlummer good catch, updated

Issue #136, opened 2017-01-30
Pull schema files from a common folder


marwanad:
Pull the schema files from a folder when creating the test directory (instead of having to copy a full project over)

marwanad:
Closed by #226.

PR #226, opened 2017-02-22
Allow multiple schema files

### LINKED ISSUE #124: Gradle Plugin Resolution Changes
This is mainly to document the `schema` and `.graphql` files lookup and open discussions to fit use cases.

## SourceSet Extensions

Currently we support users providing their own schema and graphql paths and those would be respected as long as they don't conflict.

For instance,

```
android {
sourceSets {
  debug {
   graphql {
   graphQLPath = "x"
   schemaFile = "y"
  }
  }
 }
}

```

## Current  Behaviour

If a user provide those two config options for a SOURCESET **AND** they don't provide a conflicting for another sourceSet that is shared by a variant. Those two will be respected and only files under graphQLPath and schemaFile will be used to generate the IR. 

If, on the other hand they also provide those two config options for the `main` sourceSet for example, we throw an exception because it kinda doesn't make sense for that. For instance, that would be an example for a conflicting configuration
```
android {
sourceSets {
  debug {
   graphql {
   graphQLPath = "x"
   schemaFile = "y"
  }
  main {
   graphql {
   graphQLPath = "z"
   schemaFile = "y"
  }
  }
 }
}
```

Regarding flavours (cc: @digitalbuddha) and if no extension options were specified.

Consider the following
```
 project.android {
      compileSdkVersion 25
      buildToolsVersion "25.0.2"
      productFlavors {
        demo {
          applicationIdSuffix ".demo"
          versionNameSuffix "-demo"
        }
        full {
          applicationIdSuffix ".full"
          versionNameSuffix "-full"
        }
      }
    }
```

The tasks added by apollo plugin will be `generatedApolloIRDemoDebug`, `generatedApolloIRDemoRelease`, `generatedApolloIRFullDebug`, `generatedApolloIRFullRelease`

Consider the demo flavour, it will the following sourceSets
```
src/demoDebug/ (build variant source set)
src/debug/ (build type source set)
src/demo/ (product flavor source set)
src/main/ (main source set)
```
If **no config options** were specified, we do the resolution in a way that's similar to the Gradle lookup mechanism.
So, if we are running `generateApolloDemoDebug`, if a schema file exists under `src/demoDebug/graphql/`, it uses that file and passes it all `.graphql` files found under all the **sourceSets** (the four mentioned above). If not it keeps looking in the other sourceSet directories. We also pick the schemaFile in a style similar to the Gradle resolution. Ie. pick the one from `demoDebug` in that case regardless of the existence of others in the other variant sourceSets.

marwanad:
This allows multiple schema files to exist in a source set directory.

_The catch is that only one file should exist in the path from the source set directory root to that schema._ For example, `src/main/graphql/schema.json` and `src/main/graphql/apollo/schema.json` is invalid but `src/main/graphql/apollo/schema.json` and `src/main/graphql/githunt/schema.json` is valid.

**TL;DRr**: 

Essentially the graphql query files are picked from all child directories relative to where the schema files is found and this is done for all source sets belonging to that variant. 

Only the schema file from the highest priority source set directory is picked to be used with the selected graphql files.

_________

For a flavoured project:
The following tasks will be created when the plugin is applied:

- generateApolloIRDemoDebug which gets its input files from `src/main/graphql`, `src/demo/graphql`, `src/debug/graphql` and `src/demoDebug/graphql`.

- generateApolloIRDemoRelease

- generateApolloIRFullDebug

- generateApolloIRFullRelease

Suppose the following schema files exist:
```
src/demoDebug/graphql/githunt/schema.json
src/main/graphql/githunt/schema.json

src/demoDebug/graphql/apollo/schema.json
src/release/graphql/starwars/schema.json
```
And the following query files:

```
src/demoDebug/graphql/githunt/feed/query1.graphql
src/main/graphql/githunt/feed/query2.graphql

src/demoDebug/graphql/apollo/query3.graphql
src/main/graphql/apollo/query4.graphql
src/demoDebug/graphql/apollo/query5.graphql

src/release/graphql/starwars/query6.graphql
```

On running **generateApolloIRDemoDebug**, since the githunt schema was found under both the demoDebug and main source sets, the plugin will have to make a decision on which one to use. This is done in a way similar to build system resolution where it would pick the schema file under src/demoDebug. 

So it would run the following command:
`apollo-codegen generate query2.graphql query1.graphq -- schema demoDebug/githunt/schema.json`

Here it also picks the query files under the main source set, similar to how Gradle deals with flavors when building the apks.

For the second schema file found under demoDebug, `apollo/schema.json`. The following would be executed:

`apollo-codegen generate query3.graphql query4.graphql query5.graphql -- schema demoDebug/apollo/schema.json`.

Nothing will be done with the release schemas and queries.
___________________

The third commit is for adding tests with multiple schemas, hence the huge diff.

Closes #124 and #136 (done as part of refactoring the test fixtures)


.travis.yml
```diff
@@ -14,7 +14,7 @@ jdk:
   - oraclejdk8
 
 script:
-  - ./gradlew clean check -x checkstyleTest -x :apollo-gradle-plugin:test --stacktrace
+  - ./gradlew clean check -x checkstyleTest --stacktrace
```
digitalbuddha: I'd prefer to run build rather than Check. This way if we have dexing errors they will be caught in CI

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodegenArgs.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.File;
+import java.util.Set;
+
+
+public class ApolloCodegenArgs {
+  private File schemaFile;
+  private Set<String> queryFiles;
+
+  public ApolloCodegenArgs(File schema, Set<String> queries) {
```
digitalbuddha: Does this class need both a constructor and setters? Seems like a value object, could we make immutable

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodegenArgs.java
```diff
@@ -0,0 +1,31 @@
+package com.apollographql.android.gradle;
+
+import java.io.File;
+import java.util.Set;
+
+
+public class ApolloCodegenArgs {
```
digitalbuddha: Final pls

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloIRGenTask.java
```diff
@@ -1,172 +1,173 @@
...
-      throw new GradleException("Apollo-codegen was not found in node_modules. Please run 'gradle " +
-          "installApolloCodegen");
+      throw new GradleException("Apollo-codegen was not found in node_modules. Please run the installApolloCodegen task.");
     }
+    setScript(apolloScript);
 
-    if (!schemaFile.isFile()) {
-      throw new GradleException("Couldn't find a schema file. Please ensure a valid schema.json files exists in the " +
-          "sourceSet directory");
+    Map<String, ApolloCodegenArgs> schemaQueryMap = buildSchemaQueryMap(getInputs().getSourceFiles().getFiles());
```
digitalbuddha: Immutable pls

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloIRGenTask.java
```diff
@@ -1,172 +1,173 @@
...
+   * The schema file under the source set with the highest priority is used and all the graphql query files under the
+   * schema file's subdirectories from all source sets are used to generate the IR.
+   *
+   * If any of the schema file's ancestor directories contain a schema file, a GradleException is
+   * thrown. This is considered to be an ambiguous case.
+   *
+   * @param files - task input files which consist of .graphql query files and schema.json files
+   * @return - a map with schema files as a key and associated query files as a value
+   */
+  private Map<String, ApolloCodegenArgs> buildSchemaQueryMap(Set<File> files) {
```
digitalbuddha: There seems to quite a bit of logic in this method. Maybe it would make sense to break it up a bit and unit test some pieces?

apollo-gradle-plugin/src/test/testProject/android/buildScriptFixtures/flavored.gradle
```diff
@@ -0,0 +1,60 @@
...
+}
+
+plugins {
+  id 'com.android.application'
+  id 'com.apollographql.android'
+}
+
+android {
+  compileSdkVersion 25
+  buildToolsVersion '25.0.2'
```
digitalbuddha: These should be in a properties file. I'm assuming as we update version of sample we would want tests to update too.

digitalbuddha:
Thank you so much for this! Nothing major just some code style stuff.

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodegenArgs.java
marwanad: yeah, should be a value object.

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloIRGenTask.java
```diff
@@ -103,11 +93,11 @@ public void exec() {
                 " Please ensure no schema files exist on the path to another one");
     }
 
-    Map<String, ApolloCodegenArgs> schemaQueryMap = new HashMap<>();
+    ImmutableMap.Builder<String, ApolloCodegenArgs> schemaQueryMap = ImmutableMap.builder();
     for (final File f : schemaFiles) {
       final String normalizedSchemaFileName = getPathRelativeToSourceSet(f);
       // ensures that only the highest priority schema file is used
-      if (schemaQueryMap.containsKey(normalizedSchemaFileName)) {
+      if (schemaQueryMap.build().containsKey(normalizedSchemaFileName)) {
```
marwanad: I assume this step would be less efficient compared to using a mutable map.

apollo-gradle-plugin/src/test/testProject/android/buildScriptFixtures/flavored.gradle
sav007: Not sure that property will be picked up, as this is a test fixture and I guess gradle plugin test runs it in isolated env from our main gradle.properties file, but I can be wrong.

sav007:
awesome!

sav007:
Test `FlavoredMultiSchemaSpec` fails

apollo-gradle-plugin/src/test/testProject/android/buildScriptFixtures/flavored.gradle
marwanad: We can write those through a groovy util method. I suppose that would make updating those values less painful.

sav007:
@marwanad merge?

Issue #225, opened 2017-02-21
Automate updating IR test files for new codegen version


BenSchwab:
As apollo-codegen changes, we will need to update our IR test suite. We should automate this. We can use the form:
\-test/graphql/com/example/test_name
\-\-\- TestQuery.json
\-\-\- TestQuery.graphl
\-\-\- (schema.json)
\-\-\- TestQueryExpected.java

The codegen will run on a general schema file, unless a custom one is present in the test's folder.

This will only need to run occasionally, so I am leaning on requiring apollo-codegen to be to installed  and having gradle call a command line.

digitalbuddha:
The requiring code gen to be installed part... Will that work with Travis?

BenSchwab:
I am envisioning this being run very infrequently. So the TestQuery.json files would be checked-in.

We could automate this to occur on every build, and follow a similar approach to `ApolloIRGenTask`.

This has the upside that every time we update the codegen version, we won't forget to regenerate the test files. Downside is that it is another point of potential windows npm pain.

marwanad:
We currently do so with the gradle plugin integration tests (https://github.com/apollographql/apollo-android/tree/master/apollo-gradle-plugin/src/test/testProject/android/basic/src/main/graphql/com/example). That is having the schema files and .graphql files and then applying the plugin and running `gradle build`.

The tests under apollo-compiler don't have any logic associate with them apart from checking that the generated files compile correctly iirc.

Correct me if I'm wrong but I think if bumping the apollo-codegen introduces breaking changes, they will be caught by the integration tests because either `apollo-compiler` will fail on a malformed IR or javac will fail.

BenSchwab:
@marwanad The apollo-compiler implicitly depends on the results of apollo-codegen. Currently, we are checking in the IR files that we use to generate code, and then assert that it matches an expected generated file. 

Here's why I think this would be helpful -- I need to bump the apollo-codegen version so we have access to field arguments to use for caching. Ultimately, I will edit the compiler to add the field arguments in the generated code. If I were to update the TestQueryExpected.java, the tests will fail because the IR files are "frozen" from an old version with no arguments. So I would have to go by hand calling apollo-codgen and replacing the IR file.

In terms of bumping-codgen introducing breaking changes, I would argue that the tests in gradle plugin integration tests are not sufficient. There are lots of parts of the GraphQL spec that aren't exercised in those tests, but which are contained in the compiler tests. (type names, directives, complex arguments etc.)

[...]

PR #231, opened 2017-02-22
Bump Apollo Codgen to version 0.10.8. Simplify compiler test fixture …

BenSchwab:
…generation.

Apollo-codegen 0.10.8 supports field arguments, which are needed to generate
cache keys.

The process of updating the code-gen version was very manual as the
apollo code-gen tests run from pure IR (TestQuery.json) files.

When we bump the codegen, we will want to regenerate the TestQuery.json
files. This diff extracts the TestQuery.graphql for each of the compiler
tests and creates a simple gradle plugin to execute apollo-codegen
with the *Local* version of apollo-codegen.

While it would be nice to have this happen automatically, doing so
would require following a approach similar to the apollo-gradle-plugin
which installs apollo-codegen to a local repository. Sharing this code
would require a more significant refactor to avoid logic duplication.

Instead, whenever we update the codegen plugin, we can run
`./gradlew :apollo-compiler:updateTestIRFiles`.
This will record the local apollo codegen version used to `apollo-codegen.properties`
to force review acknowledgment of the version used.

This diff mostly unifies the compiler tests on one
schema (https://github.com/apollographql/starwars-server).  Custom logic needed for certain tests was injected into the schema. 

In cases where the test needs something completely custom, you can place a schema.json file in the test folder.

Closes [225](https://github.com/apollographql/apollo-android/issues/225)


apollo-compiler/src/test/graphql/com/example/directives/TestQuery.json
```diff
@@ -1,6 +1,7 @@
 {
   "operations": [
     {
+      "filePath": "/Users/ben_schwab/airapps/apollo-android/apollo-compiler/src/test/graphql/com/example/directives/TestQuery.graphql",
```
brianPlummer: "/Users/ben_schwab/" is this intentional?

BenSchwab: It seems that the apollo-codgen records this as an absolute path now. I could go and remove all these lines, but they don't seem to have an impact on the codegen.

brianPlummer: no worries then.  just double-checking :)

apollo-compiler/gradle/update-test-IR-files.gradle
```diff
@@ -0,0 +1,68 @@
...
+    }
+    def dir = new File("apollo-compiler/src/test/graphql/com/example")
+    dir.eachFile(FileType.DIRECTORIES) {
+      testDir -> generateIRFile(testDir)
+    }
+    logApolloVersion(apolloVersion)
+  }
+
+  private static def logApolloVersion(apolloVersion) {
+    File outputFile = new File('apollo-compiler/src/test/graphql/apollo-codegen.properties')
```
marwanad: does this file get used after?

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloCodeGenInstallTask.java
```diff
@@ -19,7 +19,7 @@
 public class ApolloCodeGenInstallTask extends NpmTask {
   static final String NAME = "installApolloCodegen";
   private static final String INSTALL_DIR = "node_modules/apollo-codegen";
-  private static final String APOLLOCODEGEN_VERSION = "0.10.5";
+  private static final String APOLLOCODEGEN_VERSION = "0.10.8";
```
marwanad: Would it make more sense to add it to the companion object in `GraphQLCompiler` and use that instead?

BenSchwab: Yeah, good call. Seems like a better fit as this is something the Compiler is "aware" of, albeit indirectly

PR #233, opened 2017-02-23
Feature/operation variables map

sav007:
Add generation of valueMap for operation variables

Required for caching


apollo-compiler/src/test/graphql/com/example/input_object_type/TestQuery.java
```diff
@@ -54,9 +57,13 @@ public String queryDocument() {
 
     private final @Nonnull ReviewInput review;
 
+    private final Map<String, Object> valueMap = new LinkedHashMap<>();
```
digitalbuddha: yay `LinkedHashMap` :-)

PR #235, opened 2017-02-23
don't add apollo-api dependency if running tests

marwanad:
Instead of running the integration tests with `gradle build`, we now set a JVM
flag and run the test through `gradle build -Dapollographql.skipApi=true`. This
flag is checked by `ApolloPlugin` and if set to true, it doesn't add the
apollo-api dependency.

Now, build fixtures add the apollo-api classes to its source sets and so the
tests will run against the project version of apollo-api.

The jar archives would still include the apollo-api dependency as before.

### DIFF
--- a/apollo-api/build.gradle
+++ b/apollo-api/build.gradle
@@ -5,7 +5,7 @@ sourceCompatibility = JavaVersion.VERSION_1_7
 
 dependencies {
   compile dep.jsr305
-  compile dep.jsr205
+  compile dep.jsr250
 }
 
 apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
--- a/apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/BasicAndroidSpec.groovy
+++ b/apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/BasicAndroidSpec.groovy
@@ -23,12 +23,12 @@ class BasicAndroidSpec extends Specification {
     def result = GradleRunner.create()
         .withProjectDir(testProjectDir)
         .withPluginClasspath()
-        .withArguments("generateApolloClasses")
+        .withArguments("build", "-Dapollographql.skipApi=true")
         .forwardStdError(new OutputStreamWriter(System.err))
         .build()
 
     then:
-    result.task(":generateApolloClasses").outcome == TaskOutcome.SUCCESS
+    result.task(":build").outcome == TaskOutcome.SUCCESS
     // IR Files generated successfully
     assert new File(testProjectDir,
         "build/generated/source/apollo/generatedIR/release/src/main/graphql/ReleaseAPI.json").isFile()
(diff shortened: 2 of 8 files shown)


apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/BasicAndroidSpec.groovy
```diff
@@ -23,12 +23,12 @@ class BasicAndroidSpec extends Specification {
     def result = GradleRunner.create()
         .withProjectDir(testProjectDir)
         .withPluginClasspath()
-        .withArguments("generateApolloClasses")
+        .withArguments("build", "-Dapollographql.skipApi=true")
```
marwanad: changed to build because this actually runs `javac` and checks if there are import errors, etc

apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/BasicAndroidSpec.groovy
```diff
@@ -38,6 +38,7 @@ class BasicAndroidSpec extends Specification {
     // Java classes generated successfully
     assert new File(testProjectDir, "build/generated/source/apollo/com/example/DroidDetails.java").isFile()
     assert new File(testProjectDir, "build/generated/source/apollo/com/example/Films.java").isFile()
+    assert new File(testProjectDir, "build/generated/source/apollo/type/CustomType.java").isFile()
```
marwanad: Checks for the custom scalar type.

apollo-gradle-plugin/src/test/testProject/android/buildScriptFixtures/flavored.gradle
```diff
@@ -48,6 +55,8 @@ repositories {
 
 dependencies {
   compile 'com.android.support:appcompat-v7:25.1.1'
+  compile 'com.google.code.findbugs:jsr305:3.0.1'
```
marwanad: The downside to this approach, is that we'd have to add the apollo-api dependencies manually.

I'm not sure if there's a way to avoid having to do that.

apollo-api/build.gradle
```diff
@@ -5,7 +5,7 @@ sourceCompatibility = JavaVersion.VERSION_1_7
 
 dependencies {
   compile dep.jsr305
-  compile dep.jsr205
+  compile dep.jsr250
```
digitalbuddha: intentional change from 205 to 250?

apollo-gradle-plugin/src/test/testProject/android/buildScriptFixtures/basic.gradle
```diff
@@ -37,6 +44,9 @@ repositories {
 
 dependencies {
   compile 'com.android.support:appcompat-v7:25.1.1'
+  compile 'com.google.code.findbugs:jsr305:3.0.1'
+  compile 'javax.annotation:jsr250-api:1.0'
+  compile 'com.fernandocejas:arrow:1.0.0'
```
digitalbuddha: I think we can use Date or any other Custom type to not carry this additional dependency....

apollo-api/build.gradle
marwanad: Apparently it's called `jsr250-api:1.0'`

digitalbuddha:
Sorry I wasn't clear with the Custom Types Mapping. I feel like we should still have it there. I just don't think the type should be one that warrants a library inclusion. For example `UUID` can be used as the "Custom Type"

marwanad:
@digitalbuddha we still have the `DateTime` custom scalar type in both the schema and tests. Unless, it would make a difference to have more than one.

digitalbuddha:
nope all set thanks!

PR #236, opened 2017-02-23
Feature/simplify mapper generation2

sav007:
Due to the fact we dropped support of stream reader we can simplify of mapper generation.


digitalbuddha:
Lgtm I only looked at the code gen fyi

PR #237, opened 2017-02-23
update sample app to reflect the latest changes

marwanad:
Sample app doesn't build if the latest snapshot is used. 

This brings it back to a usable state.


apollo-sample/build.gradle
```diff
@@ -36,7 +35,3 @@ dependencies {
 }
 
 apply plugin: 'com.apollographql.android'
-
-apollo {
```
marwanad: @brianPlummer removing this block should fix your problem

PR #238, opened 2017-02-24
move project dependencies to a separate script

marwanad:
This allows us to refer to those dependencies anywhere in the project without
having to depend on the project/subproject structures.

Test fixtures now have access to those dependencies and build tools versions 🎉 

(Obviously we could've done the same using `apply from: '../build.gradle'` but this file is project-specific and could have other tasks defined in the future so it's safer to have them in a separate script.)


dependencies.gradle
```diff
@@ -0,0 +1,42 @@
+def versions = [
```
digitalbuddha: Mind sticking this in /gradle folder?

Issue #206, opened 2017-02-17
Parse field arguments from IR and supply to generated mapper fields


BenSchwab:
We need field arguments to properly generate cache-keys for fields in a normalized cache. These field arguments need to have variable values resoled at run-time.

Simple Example of IR arguments:

```
	"source": "query TestQuery($episode: Episode, $includeName: Boolean!) {\n  hero(episode: $episode) {\n    __typename\n    name @include(if: $includeName)\n  }\n  human(id: \"1000\") {\n    __typename\n    name\n    height(unit: FOOT)\n  }\n}",
			"fields": [
				{
					"responseName": "hero",
					"fieldName": "hero",
					"type": "Character",
					"args": [
						{
							"name": "episode",
							"value": {
								"kind": "Variable",
								"variableName": "episode"
							}
						}
					],
					"fields": [
						{
							"responseName": "name",
							"fieldName": "name",
							"type": "String!",
							"isConditional": true
						}
					],
					"fragmentSpreads": [],
					"inlineFragments": []
				},
				{
					"responseName": "human",
					"fieldName": "human",
					"type": "Human",
					"args": [
						{
							"name": "id",
							"value": "1000"
						}
					],
					"fields": [
						{
							"responseName": "name",
							"fieldName": "name",
							"type": "String!"
						},
						{
							"responseName": "height",
							"fieldName": "height",
							"type": "Float",
							"args": [
								{
									"name": "unit",
									"value": "FOOT"
								}
							]
						}
					],
					"fragmentSpreads": [],
					"inlineFragments": []
				}
			],
			"fragmentsReferenced": []
		}
	],

```

However, per the graphql spec, arguments can nest. Therefore, a variable could be buried at the bottom of a long argument chain.

More Complex example:
```
"source": "query ComplexArguments($stars: Int!) {\n  author(input: {thoughts: \"yeah\", input: {stars: $stars}}) {\n    __typename\n    id\n  }\n}",
			"fields": [
				{
					"responseName": "author",
					"fieldName": "author",
					"type": "User",
					"args": [
						{
							"name": "input",
							"value": {
								"thoughts": "yeah",
								"input": {
									"stars": {
										"kind": "Variable",
										"variableName": "stars"
									}
								}
							}
						}
					],
					"fields": [
						{
							"responseName": "id",
							"fieldName": "id",
							"type": "Int!"
						}
					],
					"fragmentSpreads": [],
					"inlineFragments": []
				}
			],
[...]

PR #241, opened 2017-02-24
Add field arguments to generated code

BenSchwab:
Lots of files changes because of updated tests. For more concise diff view:

44dc826 Has the changes in the code generator to add field arguments
a51e2f3 Has the new tests, and updated tests

Some notes:

Added a helper class UnmodifiableMapBuilder to the API because creating maps inline in Java is hard. I was tempted to do new Hashmap() {{ put, put}}. But due to the nested nature of arguments, this would creates tons of anonymous inner classes. This class roughly follows the structure of Guava's ImmutableMapBuilder.

This PR allows us to run quick substitution routine in the reader in replace all objects of type {type: Variable...} with their runtime value.

Closes [206](https://github.com/apollographql/apollo-android/issues/206)


BenSchwab:
@digitalbuddha fyi updated PR to change the List<Map> into just a Map as explained in my comment.

digitalbuddha:
Looks great!

Issue #132, opened 2017-01-30
dynamically add apollo-runtime dependency in plugin


marwanad:

sav007:
Not sure that we need it for now, as Apollo runtime doesn't have anything useful. Even sample app doesn't use it.
Downgrading the priority to low

PR #242, opened 2017-02-24
fix sample to use apollo-runtime and okhttp

brianPlummer:
image


digitalbuddha:
closes #242

apollo-sample/build.gradle
```diff
@@ -29,8 +29,9 @@ dependencies {
   compile dep.appcompat
   compile dep.rxjava2
   compile dep.rxandroid
-  compile dep.retrofitRx2Adapter
   compile dep.okhttpLoggingInterceptor
+  compile project(path: ':apollo-runtime')
```
digitalbuddha: please depend on the maven artifact here. this way we will know if there was a problem with the published version

brianPlummer: done!

marwanad: might be outside the scope of this PR but should we also add this dependency through the Plugin?

It seems that apollo-runtime would have to be included for all use-cases.

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -18,25 +21,31 @@
...
     final TextView txtResponse = (TextView) findViewById(R.id.txt_response);
-    SampleApplication application = (SampleApplication) getApplication();
-    application.githuntApiService()
-        .githuntFeed(new FeedQuery(FeedQuery.Variables.builder()
-                .limit(10)
-                .type(FeedType.HOT)
-                .build()))
+    application = (SampleApplication) getApplication();
+
+    Observable<Response<FeedQuery.Data>> query = Observable.fromCallable(new Callable<Response<FeedQuery.Data>>() {
```
digitalbuddha: Wow this looks really ugly.  No fault of yours. Let's try this for now (until I get the Rxjava Call Adapter stuff in).  Move `executeFeedQuery` to `SampleApplication` have that method take a param of limit and return an observable of the Response.  This way the activity can just do application.executeFeedQuery(10).subscribe()

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
```diff
@@ -18,25 +21,31 @@
...
+      @Override public Response<FeedQuery.Data> call() throws Exception {
+        return executeFeedQuery();
+      }
+    });
+
+    query
         .subscribeOn(Schedulers.io())
         .observeOn(AndroidSchedulers.mainThread())
         .subscribe(new Observer<Response<FeedQuery.Data>>() {
```
digitalbuddha: please use a `DisposableObserver` which does not have the `OnSubscribe` callback

apollo-sample/src/main/java/com/example/apollographql/sample/MainActivity.java
digitalbuddha: Actually theres no need to even do rxjava anymore. there is an apolloclient method .enque that takes a callback and schedules to run in background... use that instead and drop rxjava from sample please

digitalbuddha:
Yes please make issue

apollo-sample/build.gradle
marwanad: #132 is useful now.

apollo-sample/src/main/java/com/example/apollographql/sample/SampleApplication.java
```diff
@@ -2,46 +2,35 @@
...
+  private static final String BASE_URL = "https://githunt-api.herokuapp.com/graphql";
+  private ApolloClient apolloClient;
 
   @Override public void onCreate() {
     super.onCreate();
     HttpLoggingInterceptor loggingInterceptor = new HttpLoggingInterceptor();
     loggingInterceptor.setLevel(HttpLoggingInterceptor.Level.BODY);
-    okHttpClient = new OkHttpClient.Builder()
+    OkHttpClient okHttpClient = new OkHttpClient.Builder()
         .cache(new Cache(new File(getCacheDir(), "okhttp"), 10 * 1024))
```
digitalbuddha: remove cache, it does nothing as all post requests.

Issue #243, opened 2017-02-24
Add RxJavaApollo wrappers


digitalbuddha:
Since we dropped Retrofit, Rxjava support is no longer free.  We need to create some sort of wrappers for apollo which would allow user to return RxJava observables.  Since there is a dependency on RxJava we will need to make call adapters. There will be an ApolloRx1Adapter and an ApolloRx2Adapter.

digitalbuddha:
First implementation is to only support `Observable` there will be follow up issues to support `Single`.

digitalbuddha:
After a bit of discussion on slack:

- We will be building adapters in the style of `CallAdapter<T>`.  
- We will have ApolloClient<T> register an adapter<T>, if non registered default will be used for `newCall`
- We will create override <T> newCall(operation,CallAdapter<T>) so that users can use many call adapters

digitalbuddha:
Adapters won't work do to use not being able to do R<T> in java. Here's the backup plan https://gist.github.com/digitalbuddha/3c1dbccf059f0279e46e540b2c5fd89e

marwanad:
Closed by #341

PR #245, opened 2017-02-24
Make ApolloClient generic, add CallAdapter

digitalbuddha:
Part of #243 

- Make Apollo Client Generic based on `CallAdapter<T>` return type
- Create ApolloCallAdapter that implements CallAdapter<ApolloCall>
- Pass in CallAdapter<T> to Apollo Client Builder
- Pass`NewCall` through `CallAdapter`
- Clean up directory structure by creating impl folder (runtime was getting out of control), I'm open to other structures as well. Initial split was going to be /converter and /adapter but then some classes would need to be made public.


apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
```diff
@@ -0,0 +1,111 @@
...
+          writer.value(customTypeAdapter.encode(value));
+        }
+      });
+      return this;
+    }
+
+    public ApolloClient<B> build() {
+      checkNotNull(okHttpClient, "okHttpClient is null");
+      checkNotNull(serverUrl, "serverUrl is null");
+      checkNotNull(callAdapter, "serverUrl is null");
```
sav007: checkNotNull(callAdapter, "callAdapter is null");

sav007:
Should we instead of `impl` package have `internal`?

digitalbuddha:
The problem with internal is that the classes would still be public right?

sav007:
No I meant to rename package `impl` to `internal` (like okHttp has)

apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
```diff
@@ -0,0 +1,111 @@
...
+
+    return adapter.adapt(call);
+  }
+
+  public static class Builder<B> {
+    private OkHttpClient okHttpClient;
+    private HttpUrl serverUrl;
+    private final Map<ScalarType, CustomTypeAdapter> customTypeAdapters = new LinkedHashMap<>();
+    private Moshi.Builder moshiBuilder = new Moshi.Builder();
+    private CallAdapter<B> callAdapter;
```
BenSchwab: Any reason to not default to the ApolloCallAdapter?

digitalbuddha: I can't for the life of me figure out how to do that while keeping apollo client typesafe. Like then someone can do builder<Foo> and it'll bomb at runtime

apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
```diff
@@ -0,0 +1,111 @@
...
+public final class ApolloClient<R> implements ApolloCall.Factory<R> {
+  public static <B> Builder<B> builder() {
+    return new Builder<>();
+  }
+
+  private final HttpUrl serverUrl;
+  private final Call.Factory httpCallFactory;
+  private final Map<ScalarType, CustomTypeAdapter> customTypeAdapters;
+  private final Moshi moshi;
+  private CallAdapter<R> adapter;
```
brianPlummer: should be final

Issue #195, opened 2017-02-14
Add ability to prefetch raw response. 


digitalbuddha:
To fulfill background update/syncing it would be nice to be able to defer parsing until data is read rather than when it is downloaded.  The idea is to have a mechanism to prefetch graph responses but not yet inflate into models.  Maybe this can be rolled into into the cache implementation.

Use Case:
NY Times wants to download all the days articles overnight and store them for offline use.  There is no reason to inflate the models prior to storing as a user may not see the data before it will become stale.  When a user opens a screen apollo client should be able to inflate the json for a particular query and then return it to the user.

This feature would ideally separate "apollo downloader" from "apollo parser"  allowing a user to download a query result but never inflate it.

martijnwalraven:
I think this feature could be implemented in terms of a response cache. If we give the network interface the ability to prefetch responses without parsing them, we could then store them on disk and have future requests read them from there.

A response cache is obviously less flexible than a normalized cache, because it is tied to a particular query and variables. But it seems like a good fit for use cases like this.

digitalbuddha:
Response cache sounds reasonable. I'd still want to be able to utilize a normalized cache as well if possible. This would be bytes caching while the normalized one can be pojo caching imo

martijnwalraven:
Yep, that's exactly what I had in mind. When you read a response from the response cache and parse it into model types it will be normalized and put into the cache as usual.

digitalbuddha:
Basic Idea:
Since we are using `OKHTTP` we should exploring making an ApolloCacheInterceptor for full request/response caching.    I see this working as:
We implement a Request/Response Cache on disk, for each request the `Cache` can return a `Record` which will contain a`RecordState` == {FRESH, STALE, MISSING}.  

**On new query request if cache is:**
- FRESH -> Return cached response
- STALE -> Hit Network, Update Cache, If network error return stale response (Network before Stale)
- MISSING -> Hit network update cache

When the Interceptor is instantiated Users can pass in an `ExpirationDuration` which will drive the Stale logic.

Additionally, we need a way to pass in a flag `FORCE_NETWORK` to bust through the cache on a per request basis.  Initial proposal is to have a flag when creating a query `new Query(TRUE|FALSE)` which will drive the response caching policy.  We would also need a way to clear the cache, say if a user logs out of an account.

Ideally we would also create a new ApolloClient public api
`void prefillCache(query)`
`Callback prefill(query)`

[...]

PR #246, opened 2017-02-24
Cache interceptor

sav007:
Part of #195 

This is the first version of http cache implementation.
This PR makes each call made with `ApolloClient` that has set `HttpCache` to be cached.

Cache control (or cache policy) with invalidation of cache will be implemented in next PR.


apollo-runtime/src/main/java/com/apollographql/android/HttpCacheInterceptor.java
```diff
@@ -0,0 +1,27 @@
+package com.apollographql.android;
+
+import java.io.IOException;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+final class HttpCacheInterceptor implements Interceptor {
+  static final String CACHE_CONTROL_HEADER = "APOLLO-CACHE-CONTROL";
```
sav007: We need specs for this

apollo-runtime/src/main/java/com/apollographql/android/RealApolloCall.java
```diff
@@ -126,6 +126,7 @@ private Request httpRequest(Operation operation) throws IOException {
         .post(requestBody)
         .header("Accept", ACCEPT_TYPE)
         .header("Content-Type", CONTENT_TYPE)
+        .header(HttpCacheInterceptor.CACHE_CONTROL_HEADER, "cache")
```
sav007: for now all request are cached, we need specs for our custom cache control header

brianPlummer:
looks good so far

apollo-runtime/src/main/java/com/apollographql/android/HttpCacheInterceptor.java
digitalbuddha: One option: we can return a `CachingApolloClient` with one of:
```
client.newNetworkCall(operation)
client.newCall(operation,FORCE_NETWORK)
```

@martijnwalraven how does iOS bust through cache?

apollo-runtime/src/test/java/com/apollographql/android/IntegrationTest.java
```diff
@@ -186,24 +211,42 @@
     assertThat(data.shop().products().edges().get(0).node().unsupportedCustomScalarTypeBool()).isEqualTo(Boolean.TRUE);
     assertThat(data.shop().products().edges().get(0).node().unsupportedCustomScalarTypeString()).isInstanceOf(String.class);
     assertThat(data.shop().products().edges().get(0).node().unsupportedCustomScalarTypeString()).isEqualTo("something");
+
+    okhttp3.Response cachedResponse = httpCache.read(((RealApolloCall) call).httpCall.request());
+    assertThat(cachedResponse).isNotNull();
+    assertThat(cachedResponse.body().source().readString(Charsets.UTF_8))
```
digitalbuddha: `.readUtf8`

digitalbuddha:
I noticed that you pulled the `CacheProxy` logic back into the `HTTPCache`, just an fyi thats how Square originally had it and then they moved it out.  I'm not crazy of it being in the Cache as it kind of breaks the single responsibility principle. Everything else looks good enough to continue

digitalbuddha:
Does it make sense to slap a license on top of `httpcache` since it did start as an okhttp class?

sav007:
`HttpCache` is not from okHttp, but `HeadersEntry` is copied, so I will add to it

apollo-runtime/src/main/java/com/apollographql/android/HttpCache.java
```diff
@@ -0,0 +1,456 @@
...
+
+public final class HttpCache {
+  private static final int VERSION = 99991;
+  private static final int ENTRY_METADATA = 0;
+  private static final int ENTRY_BODY = 1;
+  private static final int ENTRY_COUNT = 2;
+  private final DiskLruCache cache;
+
+  public HttpCache(File directory, long maxSize, FileSystem fileSystem) {
+    this.cache = DiskLruCache.create(fileSystem, directory, VERSION, ENTRY_COUNT, maxSize);
```
digitalbuddha: I don't think an Lru cache is going to work for my use case. Can we have another constructor where I can pass in something the implements a cache or file system interface?

sav007: Sure will provide interface then

digitalbuddha: Could you have a public method to get the key again from a request. Thank you kindly.

apollo-runtime/src/main/java/com/apollographql/android/cache/ResponseHeaderRecord.java
```diff
@@ -0,0 +1,314 @@
...
+import okhttp3.internal.platform.Platform;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
+import okio.ByteString;
+import okio.Okio;
+import okio.Source;
+
+/**
+ * Class was copied and modified from {@link okhttp3.Cache.Entry}
```
brianPlummer: thank you for this!! :)

Issue #203, opened 2017-02-16
Add orderIndependent cacheKey to Field


BenSchwab:
iOS implementaiton: https://github.com/apollographql/apollo-ios/blob/master/Sources/Field.swift#L22

PR #247, opened 2017-02-25
Add orderIndependentCacheKey for field

BenSchwab:
Closes [203](https://github.com/apollographql/apollo-android/issues/203)

Will return a cache key for a field, with variable arguments resolved. For example:

`hero(episode:JEDI,nested:[bar:2,foo:1])`.

This is [consistent with iOS](https://github.com/apollographql/apollo-ios/blob/master/Tests/ApolloTests/CacheKeyForFieldTests.swift):

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
+++ b/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
@@ -105,17 +105,8 @@ public class Field {
 
   private String orderIndependentKey(Map<String, Object> objectMap, Operation.Variables variables) {
     if (isArgumentValueVariableType(objectMap)) {
-      Object variable = objectMap.get(VARIABLE_NAME_KEY);
-      //noinspection SuspiciousMethodCalls
-      Object resolvedVariable = variables.valueMap().get(variable);
-      if (resolvedVariable instanceof Map) {
-        //noinspection unchecked
-        return orderIndependentKey((Map<String, Object>) resolvedVariable, variables);
-      } else {
-        return resolvedVariable.toString();
-      }
+      return orderIndependentKeyForVariableArgument(objectMap, variables);
     }
-
     List<Map.Entry<String, Object>> sortedArguments = new ArrayList<>(objectMap.entrySet());
     Collections.sort(sortedArguments, new Comparator<Map.Entry<String, Object>>() {
       @Override public int compare(Map.Entry<String, Object> argumentOne, Map.Entry<String, Object> argumentTwo) {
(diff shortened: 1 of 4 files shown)


apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
```diff
@@ -89,7 +96,64 @@ public boolean optional() {
...
 
-  public static enum Type {
+  public String cacheKey(Operation.Variables variables) {
+    if (arguments.isEmpty()) {
+      return fieldName();
+    }
+    return String.format("%s(%s)", fieldName(), orderIndependentKey(arguments, variables));
+  }
+
+  private String orderIndependentKey(Map<String, Object> objectMap, Operation.Variables variables) {
```
brianPlummer: consider breaking up this function into some small pieces...looks good otherwise!! :)

apollo-runtime/src/test/java/com/apollographql/android/converter/CacheKeyForFieldTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @Test
+  public void testFieldWithNoArgumentsWithAlias() {
+    Field field = Field.forString("r2", "hero", null, false);
+    Operation.Variables variables = new Operation.Variables() {
+      @Nonnull @Override public Map<String, Object> valueMap() {
+        return super.valueMap();
+      }
+    };
+    Assert.assertEquals("hero", field.cacheKey(variables));
```
brianPlummer: check out assertJ in some of the other tests.  it offers a nice fluent api:  assertThat(ield.cacheKey(variables)).isEqualTo("hero")   ....not a big deal, just a suggestion

apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
BenSchwab: 👍 I'll pull out the key for variable as it is conceptually different from the rest

apollo-runtime/src/test/java/com/apollographql/android/converter/CacheKeyForFieldTest.java
```diff
@@ -0,0 +1,176 @@
...
+import junit.framework.Assert;
+
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.annotation.Nonnull;
+
+public class CacheKeyForFieldTest {
```
sav007: Should this test be in :apollo-api module?

BenSchwab: Hmm, yeah your right. Looks like this only touches code in the api module

apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
```diff
@@ -89,7 +96,67 @@ public boolean optional() {
...
+    if (isArgumentValueVariableType(objectMap)) {
```
sav007: maybe I'm wrong, but does this check means that we can't mix variables and arguments?
Do we handle when arguments have constants, plus variables?

sav007: Oh, I see we have nested arguments.
Disregard previous comment

PR #248, opened 2017-02-25
Update sample app to use CallAdapter

BenSchwab:
Fix sample app so tests pass


brianPlummer:
thank you much! was my fault :(

digitalbuddha:
Actually was my fault, was waiting for snapshot to become available and then got lazy :-)

PR #249, opened 2017-02-25
Introduce cache control

sav007:
Part of #195 

Introduces simple http cache control.
Cache stale/expiration will be addressed in next PRs.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
@@ -10,11 +10,17 @@ import javax.annotation.Nullable;
 
 public interface ApolloCall {
 
-  void cancel();
-
   @Nonnull <T extends Operation.Data> Response<T> execute() throws IOException;
 
   @Nonnull <T extends Operation.Data> ApolloCall enqueue(@Nullable Callback<T> callback);
+
+  @Nonnull ApolloCall network();
+
+  @Nonnull ApolloCall cache();
+
+  @Nonnull ApolloCall networkBeforeStale();
+
+  void cancel();
 
   @Nonnull ApolloCall clone();
 
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
@@ -7,6 +7,11 @@ import okhttp3.Request;
 import okhttp3.Response;
 
 public final class HttpCacheInterceptor implements Interceptor {
+  public static final String CACHE_CONTROL_DEFAULT = "default";
+  public static final String CACHE_CONTROL_NETWORK_ONLY = "network";
+  public static final String CACHE_CONTROL_CACHE_ONLY = "cache";
+  public static final String CACHE_CONTROL_NETWORK_BEFORE_STALE = "networkBeforeStale";
+
   public static final String CACHE_KEY_HEADER = "APOLLO-CACHE-KEY";
   public static final String CACHE_CONTROL_HEADER = "APOLLO-CACHE-CONTROL";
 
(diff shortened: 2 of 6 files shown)


apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -10,12 +10,18 @@
 
 public interface ApolloCall {
 
-  void cancel();
-
   @Nonnull <T extends Operation.Data> Response<T> execute() throws IOException;
 
   @Nonnull <T extends Operation.Data> ApolloCall enqueue(@Nullable Callback<T> callback);
 
+  @Nonnull ApolloCall network();
```
digitalbuddha: I'm not sure if there is any gain from it but another way to set this up is that each of these methods return a subclass such as `NetworkCall`,`CacheCall` `NetworkBeforeStaleCall` this way its a more polymorphic solution of instanceOf checks rather than checking flags.

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
```diff
@@ -3,10 +3,17 @@
 import java.io.IOException;
 
 import okhttp3.Interceptor;
+import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.Response;
+import okhttp3.internal.Util;
 
 public final class HttpCacheInterceptor implements Interceptor {
+  public static final String CACHE_CONTROL_DEFAULT = "default";
```
digitalbuddha: enum pls

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
```diff
@@ -18,15 +25,37 @@ public HttpCacheInterceptor(HttpCache cache) {
 
   @Override public Response intercept(Chain chain) throws IOException {
     Request request = chain.request();
+
+    String cacheControl = request.header(CACHE_CONTROL_HEADER);
+    String cacheKey = request.header(CACHE_KEY_HEADER);
+    if (cacheControl == null
```
digitalbuddha: nitpick `if(skipCache())` will take some of the logic out of this method

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -162,4 +187,29 @@ static String cacheKey(RequestBody requestBody) throws IOException {
     requestBody.writeTo(hashBuffer);
     return hashBuffer.readByteString().md5().hex();
   }
+
+  private static String cacheControlHeader(CachePolicy cachePolicy) {
```
digitalbuddha: I feel like this can be rolled up into the enum as string values. So we don't need this mapping and can just do cachePolicy.toString() or .headerValue() smth like http://stackoverflow.com/questions/3978654/best-way-to-create-enum-of-strings

apollo-runtime/src/test/java/com/apollographql/android/impl/CacheTest.java
```diff
@@ -55,41 +79,67 @@
...
+    server.enqueue(mockResponse);
+
+    call = apolloClient.newCall(new ProductsWithUnsupportedCustomScalarTypes());
+    body = call.execute();
+    assertThat(body.isSuccessful()).isTrue();
+    checkCachedResponse(call, new File("src/test/graphql/productsWithUnsupportedCustomScalarTypes.json"));
   }
 
-  @Test public void noCache() throws Exception {
+  //TODO we need resolution for this as we get GraphQL error (not http) and in some cases there can be response with data and error
```
digitalbuddha: pls make issue

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
sav007: Not sure if we need this now, as all these subclasses are different from each other from cachePolicy flag only. I guess we can review this later if we really need to introduce them

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
```diff
@@ -18,15 +20,83 @@ public HttpCacheInterceptor(HttpCache cache) {
 
   @Override public Response intercept(Chain chain) throws IOException {
     Request request = chain.request();
+    if (isSkipCache(request)) {
```
digitalbuddha: Nitpick 'shouldSkipCache'

PR #252, opened 2017-02-26
Add cache eviction / stale logic

sav007:
Part of #195

Cache stale/eviction logic.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
@@ -50,7 +50,7 @@ public final class HttpCacheInterceptor implements Interceptor {
     }
 
     Response networkResponse = withServedDateHeader(chain.proceed(request));
-    return resolve(networkResponse, cacheResponse, cacheKey, cacheControl(request));
+    return resolveResponse(networkResponse, cacheResponse, cacheKey, cacheControl(request));
   }
 
   private boolean shouldSkipCache(Request request) {
@@ -104,7 +104,7 @@ public final class HttpCacheInterceptor implements Interceptor {
         : response;
   }
 
-  private Response resolve(Response networkResponse, Response cacheResponse, String cacheKey,
+  private Response resolveResponse(Response networkResponse, Response cacheResponse, String cacheKey,
       CacheControl cacheControl) throws IOException {
     if (networkResponse.isSuccessful()) {
       cacheResponse.close();
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
@@ -1,16 +1,19 @@
 package com.apollographql.android.cache;
 
 import java.io.IOException;
+import java.util.Date;
 
 import okhttp3.Interceptor;
 import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.internal.Util;
+import okhttp3.internal.http.HttpDate;
 
 public final class HttpCacheInterceptor implements Interceptor {
   public static final String CACHE_KEY_HEADER = "APOLLO-CACHE-KEY";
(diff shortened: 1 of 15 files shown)


apollo-runtime/src/main/java/com/apollographql/android/cache/TimeoutEvictionStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+    this.timeout = timeout;
+  }
+
+  @Override public boolean isStale(@Nonnull Response response) {
+    String servedDateStr = response.header(HttpCacheInterceptor.CACHE_SERVED_DATE_HEADER);
+    if (servedDateStr == null) {
+      return true;
+    }
+    Date servedDate = HttpDate.parse(servedDateStr);
+    return servedDate == null || System.currentTimeMillis() - servedDate.getTime() > timeout;
```
marwanad: The first check is redundant. We're already doing that on L#19.

Also, question shouldn't it be false in that case? Why are we assuming it being stale if the header is null.

sav007: it shouldn't happen at all, as we always attach date to the header, but I thought if it happened then treat as always stale.

sav007: as for the null `HttpDate.parse` can potentially return null (see okhttp3.internal.http.HttpDate#parse)

apollo-runtime/src/main/java/com/apollographql/android/cache/TimeoutEvictionStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+
+import javax.annotation.Nonnull;
+
+import okhttp3.Response;
+import okhttp3.internal.http.HttpDate;
+
+public final class TimeoutEvictionStrategy implements ResponseCacheStore.EvictionStrategy {
+  private final long timeout;
+
+  public TimeoutEvictionStrategy(long timeout) {
```
digitalbuddha: please use TimeUnit pattern here ie. long expirationDuration,
                                      @Nonnull TimeUnit expirationUnit

apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
```diff
@@ -35,13 +35,13 @@
   private final Moshi moshi;
   private CallAdapter<R> adapter;
 
-  private ApolloClient(Builder<R> builder) {
+  private ApolloClient(Builder builder, CallAdapter<R> callAdapter) {
```
digitalbuddha: so the call adapter stuff is going to go away after my pr #254 since we couldn't return a `R<T>` style thing

sav007: Sure, will then wait for your PR to go first

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCacheInterceptor.java
```diff
@@ -54,9 +68,15 @@ private boolean shouldSkipNetwork(Request request) {
 
   private Response cacheOnlyResponse(Request request) throws IOException {
     String cacheKey = request.header(CACHE_KEY_HEADER);
-    Response cachedResponse = cache.read(cacheKey);
-    if (cachedResponse != null) {
-      return cachedResponse;
+    Response cacheResponse = cache.read(cacheKey);
+    if (cacheResponse != null) {
+      if (cache.isStale(cacheResponse)) {
+        cache.remove(cacheKey);
```
BenSchwab: A little concerned about the CacheInterceptor including eviction logic. I feel the CacheInterceptor should only use expiration for *fetching preference*.

Staleness is two-fold indicator - one, it may be used to signal a change in "fetch-strategy", i.e. if the cache is stale, prefer the network. But, in that situation, you may want to still return cache if network fails. If a particular cache thinks keeping stale data around is harmful, it should return `null` on a `read` of stale data, or even run it's own eviction subroutine.

sav007: Agree, will remove this logic

PR #254, opened 2017-02-26
make client.newCall typesafe

digitalbuddha:
This was primary change
```public interface Operation<T extends Operation.Data, V extends Operation.Variables> ```

Now when you do client.newCall(Operation) it is typesafe to that operation.

Additionally I removed the additional parameter on `ApolloClient` for the `CallAdapter` it will not working since we cannot do a return type of a parameterized generic ie R<T> (because java).  My new plan to deal with the call adapter stuff is to make wrappers around the client.

### DIFF
--- a/apollo-runtime/src/test/java/com/apollographql/android/impl/IntegrationTest.java
+++ b/apollo-runtime/src/test/java/com/apollographql/android/impl/IntegrationTest.java
@@ -45,8 +45,10 @@ import static com.google.common.truth.Truth.assertThat;
 public class IntegrationTest {
   private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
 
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: apollo-runtime/src/test/java/com/apollographql/android/impl/IntegrationTest.java [hunk 1]
```diff
@@ -45,8 +45,10 @@
 public class IntegrationTest {
   private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
 
-  private ApolloClient<ApolloCall> apolloClient;
+  private ApolloClient apolloClient;
+  private HttpCache httpCache;
```
sav007: I remember i removed these, they moved to own unit tests

sav007:
just questions to `IntegrationTest` all cache logic has been moved to `CacheTest`, other then that lgtm except it has some compilation errors

PR #255, opened 2017-02-28
Handle different cases of cache failure

sav007:
Part of #195 

Handles different cases when caching can fail due to file system.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCache.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCache.java
@@ -23,7 +23,7 @@ public final class HttpCache {
   public void clear() {
     try {
       cacheStore.delete();
-    } catch (IOException ignore) {
+    } catch (Exception ignore) {
     }
   }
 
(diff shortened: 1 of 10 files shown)


apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCache.java
```diff
@@ -23,24 +23,32 @@ public HttpCache(@Nonnull ResponseCacheStore cacheStore, @Nonnull EvictionStrate
   public void clear() {
     try {
       cacheStore.delete();
-    } catch (IOException ignore) {
+    } catch (Exception ignore) {
```
marwanad: Is that change needed?

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCache.java
```diff
@@ -51,12 +59,43 @@ boolean isStale(@Nonnull Response response) {
...
+    }
+  }
+
+  private static void abortQuietly(ResponseCacheRecordEditor cacheRecordEditor) {
+    try {
+      if (cacheRecordEditor != null) {
+        cacheRecordEditor.abort();
+      }
+    } catch (RuntimeException rethrown) {
```
marwanad: Would it be better to modify the method signature to throw that instead since it's always thrown when abort is called?

apollo-runtime/src/main/java/com/apollographql/android/cache/HttpCache.java
```diff
@@ -51,12 +59,43 @@ boolean isStale(@Nonnull Response response) {
...
+      throw rethrown;
+    } catch (Exception ignore) {
```
marwanad: Do we need to log something here?

PR #257, opened 2017-02-28
Prefetch api

### LINKED ISSUE #195: Add ability to prefetch raw response.

sav007:
Ooook final piece, closes #195 

Adds Apollo prefetch api call.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/ApolloPrefetchCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/ApolloPrefetchCall.java
@@ -1,0 +1,24 @@
+package com.apollographql.android;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface ApolloPrefetchCall {
+
+  void execute() throws IOException;
+
+  @Nonnull ApolloPrefetchCall enqueue(@Nullable Callback callback);
+
+  ApolloPrefetchCall clone();
+
+  void cancel();
+
+  interface Callback {
+    void onSuccess();
+
+    void onFailure(@Nonnull Exception e);
+  }
+
+}
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/CacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/CacheInterceptor.java
@@ -92,12 +92,6 @@ final class CacheInterceptor implements Interceptor {
     return CacheControl.valueOfHttpHeader(request.header(HttpCache.CACHE_CONTROL_HEADER));
   }
 
-  private static Response strip(Response response) {
-    return response != null && response.body() != null
-        ? response.newBuilder().body(null).networkResponse(null).cacheResponse(null).build()
-        : response;
-  }
-
   private Response resolveResponse(Response networkResponse, Response cacheResponse, String cacheKey,
       CacheControl cacheControl) throws IOException {
     if (networkResponse.isSuccessful()) {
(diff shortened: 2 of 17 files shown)


apollo-runtime/src/main/java/com/apollographql/android/ApolloPrefetchCall.java
```diff
@@ -0,0 +1,24 @@
+package com.apollographql.android;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface ApolloPrefetchCall {
```
digitalbuddha: Id like to suggest rename `ApolloPrefetch` it would sound nicer to do client.apolloPrefetch().execute()

apollo-runtime/src/main/java/com/apollographql/android/cache/CacheInterceptor.java
```diff
@@ -121,6 +130,22 @@ private Response withServedDateHeader(Response response) throws IOException {
...
+    return cacheResponse
+        .newBuilder()
+        .networkResponse(strip(networkResponse))
+        .build();
+  }
+
+  private static boolean isPrefetchResponse(Request request) {
+    return Boolean.TRUE.toString().equals(request.header(HttpCache.CACHE_PREFETCH_HEADER));
+  }
+
```
digitalbuddha: General note that this class is getting a bit long. Would be nice to move some stuff to a utility or some delegate

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloPrefetchCall.java
```diff
@@ -0,0 +1,75 @@
...
+    super(operation, serverUrl, httpCallFactory, moshi);
+  }
+
+  @Override public void execute() throws IOException {
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already Executed");
+      executed = true;
+    }
+    httpCall = prepareHttpCall(HttpCache.CacheControl.DEFAULT, true);
+    httpCall.execute().code();
```
digitalbuddha: What is .code() for?

digitalbuddha:
One more thing to consider. It would be nice to have some sort of mechanism to delete the cache the first time it is retrieved. The thinking is that once we get the data into the normalized cache the response cache is redundant. So the client should have a flag like deleteOnFetchFromCache(true)

apollo-runtime/src/main/java/com/apollographql/android/ApolloPrefetchCall.java
sav007: Did you mean `client.prefetch().execute`? Without prefix `apollo`

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloPrefetchCall.java
sav007: My bad, just left over from test code I guess

apollo-runtime/src/main/java/com/apollographql/android/ApolloPrefetchCall.java
digitalbuddha: Sorry no apolloClient.prefetch().execute() sorry

Issue #258, opened 2017-03-01
Create `apollo-integration` module for tests that rely on generated code


BenSchwab:
An apollo-integration module will be a stand-a-lone project that applies the apollo-gradle plugin, and only has tests. This will help:

1) Write tests that depend on GraphQL queries
2) Less fragile (test auto update when the compiler changes)

PR #260, opened 2017-03-01
Add Normalized Cache structure and Response normalization

BenSchwab:
This PR will focus on the general structure of the normalized cache, and response normalization (writing to the cache). Reading from the cache is not done.

Some change summaries:

- Give `BufferedResponseReader` a `ResponseReaderShadow` to allow normalization while reading network response
- `GraphQLResponseNormalizer` contains the meat of the normalization construction process, which mostly matches iOS implementation
- Temporarily use unbounded in memory cache `InMemoryNormalizedCacheStore` (but may prove useful as an intermediate store)
- `NormalizedCacheStore` allows user to plug in how cached data is stored
- Set up rough structure of `NormalizedCache`

Closes [258](https://github.com/apollographql/apollo-android/issues/258)


BenSchwab:
@sav007 @digitalbuddha

Still working on writing tests, cleaning things up etc. But let me know any early thoughts/concerns.

This should roughly set the structure of the normalized cache (similar pattern to network cache), and I will be able to break out a lot of smaller tasks including:
- Preparing `ApolloRecord` for easy disk/serialized cache support
- Adding reading from NormalizedCache via interceptor
- Cache expiration, fetch strategy
- Query watching

apollo-integration/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,7 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.ben_schwab.topshelf.apollointegration">
+
+  <application android:allowBackup="true" android:label="@string/app_name" android:supportsRtl="true">
```
digitalbuddha: my understanding of `android:allowBackup` flag is that if any manifest that is being merged has that flag set then the client app will have it no matter what their main manifest says.  Our internal policy is to have libraries define android:allowBackup="false" so that client apps can still overwrite it.

apollo-integration/src/test/graphql/com/apollographql/android/impl/normalizer/HeroNameResponse.json
```diff
@@ -0,0 +1,7 @@
+{
```
digitalbuddha: seems identical to `EpisodeHeroNameResponse.json` confirming that both are needed

apollo-integration/src/test/java/com/apollographql/android/impl/ResponseNormalizationTest.java
```diff
@@ -0,0 +1,348 @@
...
+import okhttp3.mockwebserver.MockWebServer;
+
+import static com.apollographql.android.impl.normalizer.type.Episode.JEDI;
+import static com.google.common.truth.Truth.assertThat;
+
+public class ResponseNormalizationTest {
+
+  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
+
+  ApolloClient apolloClient;
```
digitalbuddha: any reason not to make fields private?

apollo-integration/src/test/java/com/apollographql/android/impl/ResponseNormalizationTest.java
```diff
@@ -0,0 +1,348 @@
...
+    };
+
+    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    cacheStore = new InMemoryNormalizedCacheStore();
+
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(okHttpClient)
+        .normalizedCache(cacheStore, new CacheKeyResolver() {
+          @Nullable @Override public String cacheKeyForJsonObject(Map<String, Object> jsonObject) {
```
digitalbuddha: nitpick: method should be named resolve unless there are other variants of how the resolver can work

apollo-integration/src/test/java/com/apollographql/android/impl/ResponseNormalizationTest.java
```diff
@@ -0,0 +1,348 @@
...
+  @Test public void testHeroName() throws IOException {
+    MockResponse mockResponse = mockResponse
+        ("src/test/graphql/com/apollographql/android/impl/normalizer/HeroNameResponse.json");
+    server.enqueue(mockResponse);
+
+    ApolloCall<HeroName.Data> call = apolloClient.newCall(new HeroName());
+    Response<HeroName.Data> body = call.execute();
+    assertThat(body.isSuccessful()).isTrue();
+
+    ApolloRecord record = cacheStore.read("QUERY_ROOT");
```
digitalbuddha: lets move the next few strings into private constants

apollo-runtime/src/main/java/com/apollographql/android/cache/ResponseReaderShadow.java
```diff
@@ -0,0 +1,30 @@
...
+
+  void didResolve(Field field, Operation.Variables variables);
+
+  void didParseScalar(Object value);
+
+  void willParseObject(Map<String, Object> objectMap);
+
+  void didParseObject(Map<String, Object> objectMap);
+
+  void didParseList(List array);
```
digitalbuddha: `willParseList` not necessary?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/ApolloRecord.java
```diff
@@ -0,0 +1,29 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.HashMap;
+import java.util.Map;
+
+//Todo: Enhance this class to better support generalized serialization ISSUE____
```
digitalbuddha: reminder to create issue :-)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/ApolloRecord.java
```diff
@@ -0,0 +1,29 @@
...
+public class ApolloRecord {
+
+  public final String key;
+
+  //Todo: provide serialized wrapper?
```
digitalbuddha: same, imo every TODO needs issue for tracking.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/CacheKeyResolver.java
```diff
@@ -0,0 +1,9 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+public interface CacheKeyResolver {
+  @Nullable String cacheKeyForJsonObject(Map<String, Object> jsonObject);
```
digitalbuddha: @Nonnull on param?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/InMemoryNormalizedCacheStore.java
```diff
@@ -0,0 +1,27 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class InMemoryNormalizedCacheStore extends NormalizedCacheStore {
+
+  private final Map<String, ApolloRecord> store = new HashMap<>();
```
digitalbuddha: nitpick and sorta opinion to use LinkedHashMap https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/  also do we need to worry about threading here?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/InMemoryNormalizedCacheStore.java
```diff
@@ -0,0 +1,27 @@
...
+  @Override public void mergeOrInsert(ApolloRecord apolloRecord) {
+    final ApolloRecord oldRecord = store.get(apolloRecord.key);
+    if (oldRecord == null) {
+      store.put(apolloRecord.key, apolloRecord);
+    } else {
+      oldRecord.mergeWith(apolloRecord);
+    }
+  }
+
+  public Collection<ApolloRecord> getAllRecords() {
```
digitalbuddha: unless there is a `setAllRecords` I don't think the `get` prefix is necessary (similar to how its store.values() not store.getValues())

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/ApolloRecord.java
```diff
@@ -0,0 +1,29 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.HashMap;
+import java.util.Map;
+
+//Todo: Enhance this class to better support generalized serialization ISSUE____
+public class ApolloRecord {
```
sav007: package protected, final ?

sav007:
General comments:
- should we drop prefix `normalized` and just use cache
- by default make all classes as closed as possible (make final, package protected etc)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Stack;
+
+import javax.annotation.Nullable;
+
+public class GraphQLResponseNormalizer implements ResponseReaderShadow {
+
+  private Stack<List<String>> pathStack;
```
sav007: Stack is based on Vector that is synchronized. Do we really need synchronization? Question about performance, should we replace with light weight impl of stack structure?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/NormalizedCache.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private final NormalizedCacheStore cacheStore;
+  private final CacheKeyResolver cacheKeyResolver;
+
+  public NormalizedCache(NormalizedCacheStore cacheStore, CacheKeyResolver cacheKeyResolver) {
+    this.cacheStore = cacheStore;
+    this.cacheKeyResolver = cacheKeyResolver;
+  }
+
+  public NormalizedCacheStore getCacheStore() {
```
sav007: nit: We are avoiding of `get` so `cacheStore()`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/NormalizedCache.java
```diff
@@ -0,0 +1,51 @@
...
+    return cacheStore;
+  }
+
+  public CacheKeyResolver getCacheKeyResolver() {
+    return cacheKeyResolver;
+  }
+
+  public GraphQLResponseNormalizer createResponseNormalizer() {
```
sav007: nit: `responseNormalizer()`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
```diff
@@ -0,0 +1,137 @@
...
+
+  @Override public void willResolveRootQuery(Operation operation) {
+    pathStack = new Stack<>();
+    recordStack = new Stack<>();
+    valueStack = new Stack<>();
+    dependentKeys = new HashSet<>();
+
+    path = new ArrayList<>();
+    currentRecord = new ApolloRecord(rootKeyForOperation(operation));
+    recordSet = new InMemoryNormalizedCacheStore();
```
sav007: Is this intentionally, use `InMemoryNormalizedCacheStore `  shouldn't it be configured in constructor what impl of `NormalizedCacheStore` will be used?

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
```diff
@@ -17,15 +19,20 @@
   private final Operation operation;
   private final ResponseFieldMapper responseFieldMapper;
   private final Map<ScalarType, CustomTypeAdapter> customTypeAdapters;
+  private final NormalizedCache normalizedCache;
+  private final GraphQLResponseNormalizer resultNormalizer;
```
sav007: nit: `responseNormalizer` ?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Stack;
+
+import javax.annotation.Nullable;
+
+public class GraphQLResponseNormalizer implements ResponseReaderShadow {
```
sav007: `GraphQLResponseNormalizer` -> `ResponseNormalizer` ?

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
```diff
@@ -53,6 +60,7 @@
         }
       }
       jsonReader.endObject();
+      this.normalizedCache.getCacheStore().mergeOrInsert(resultNormalizer.getRecords());
```
sav007: can we expect any `Exception` here? What if this operation failed?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/NormalizedCacheStore.java
```diff
@@ -0,0 +1,17 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.Collection;
+
+public abstract class NormalizedCacheStore {
```
sav007: Are all methods safe? I mean do we expect any exceptions while we reading/merging etc?

stubailo:
Hey - it looks like a lot of these comments are focused on things like variable naming and code style - are there any comments about the overall design or architecture?

digitalbuddha:
Nothing jumped out as an issue from an architectural standpoint.

apollo-integration/src/main/AndroidManifest.xml
BenSchwab: oops, yeah I have been burnt by this before

apollo-integration/src/test/graphql/com/apollographql/android/impl/normalizer/HeroNameResponse.json
BenSchwab: It is identical, just thought for organizational purposes 1 graphql query = 1 graphql response of the same name.

apollo-runtime/src/main/java/com/apollographql/android/cache/ResponseReaderShadow.java
BenSchwab: iOS only uses `willParseList` to allocate more space on their "stack" data structure. I'm not aware of any similar pattern in Java (besides setting the size when a collection is first initialized)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
BenSchwab: Don't need synchronization here. Any suggestions to a lighter stack impl?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
```diff
@@ -0,0 +1,137 @@
...
+  private Stack<List<String>> pathStack;
+  private Stack<ApolloRecord> recordStack;
+  private Stack<Object> valueStack;
+  private Set<String> dependentKeys;
+  private final CacheKeyResolver cacheKeyResolver;
+
+  private List<String> path;
+  private ApolloRecord currentRecord;
+
+  private InMemoryNormalizedCacheStore recordSet;
```
BenSchwab: I wasn't sure if I wanted to add `allRecords` to the generic interface, which is necessary for the `NormalizedCacheStore`. This is really more of a per-request store that will be thrown out once parsing is done (and the results added to the aggregate store the user supplies)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
sav007: I was thinking most of custom structure that behaves as stack, but we can use `ArrayDeque`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
BenSchwab: I'm thinking that using a intermediate store might be valuable as if there is a parsing error in the middle of normalization, we probably don't want to have some of the records from the response stored, and others not. 

Also, some store implementations may be more efficient with a batched call to `mergeOrInsert(Collection)` as opposed to serial calls to `mergeOrInsert(Record)`

sav007:
so far design looks reasonable, no any major concerns

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
BenSchwab: Open to discussion on what the right behavior is. I was thinking on an `Exception` we don't normalize anything.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/NormalizedCacheStore.java
BenSchwab: Curious to your architecture opinion on this. I tend to lean towards requiring implementations to catch their own `IO` related errors. 
Generally, an `exception` on read can be replaced with null.

I wouldn't be sure what to do with `exception` on a merge though..

sav007: I think that broken cache, should only fail in one case when cache only policy is set.
Other than that if cache is broken for some reason, we should always hit network.
Of course we need to log that as error for user to debug, plus user should understand if response is taken from cache or from network, to proper handle this.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
digitalbuddha: I feel like a more scalable solution is to use temporary names within the store. So you start writing to a file/record called temp, if everything succeeds you rename to the actual record filename/key. This way there's no transactions and no duplication of memory allocation.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/NormalizedCacheStore.java
digitalbuddha: It almost feels like we want to return an absent optional here. Or something similar. So when reading it can return an empty container or something.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
BenSchwab: Tried ArrayDeque, but it does not support nulls. So rolled our own "SimpleStack"

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/GraphQLResponseNormalizer.java
BenSchwab: Replaced with the concept of a `RecordSet` which is more inline with the iOS implementation

apollo-runtime/src/main/java/com/apollographql/android/impl/BufferedResponseReader.java
```diff
@@ -64,15 +67,28 @@
         value = readCustomType((Field.CustomTypeField) field);
         break;
       case CONDITIONAL:
-        value = readConditional((Field.ConditionalTypeField) field);
+        value = readConditional((Field.ConditionalTypeField) field, operation.variables());
```
BenSchwab: The parsing for conditional is slightly strange 

When we call readConditional, the field is actually `__typename` and not the conditional field. However, we will need to store the `_typename` for reading from the cache. Thus, read conditional actually stores two values at once, and can't follow as clean as a pattern as every other type of field.

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
```diff
@@ -53,6 +59,7 @@
         }
       }
       jsonReader.endObject();
+      this.cache.cacheStore().merge(responseNormalizer.records());
```
BenSchwab: Resurfacing some comments from earlier: 

I still feel like the right thing is to either store all, or none of the records from a response, and not burden cachestore implementations with the concept of transactions, or rollbacks.

I also don't love that the ResponseBodyConverter is saving the records at all. I feel like it's job should be to create the records as the response is parsed, and then return them with the Data.

BenSchwab:
@sav007 Looks like you moved the http code to `cache/http`. should this stuff go in `cache/normalized`

digitalbuddha:
Sounds reasonable. The other one should be in cache/http maybe?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/InMemoryCacheStore.java
```diff
@@ -0,0 +1,24 @@
+package com.apollographql.android.cache.normalized;
+
+import java.util.Collection;
+
+public final class InMemoryCacheStore extends CacheStore {
+
+  private final RecordSet recordSet;
```
sav007: Do we need to have `RecordSet` or we can merge it to this class?

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
sav007: > is to either store all, or none of the records from a response

Completely agree.  And that what I actually meant by transaction, if we failed to parse then don't store at all. But at the same time we need to make sure that `responseNormalizer` won't crash parsing network response

>I also don't love that the ResponseBodyConverter is saving the records at all

can we  pass the instance of normalizer in constructor and move logic of merging to `RealApolloCall`?

sav007:
One question do we need really `RecordSet` or we can work with `CacheStore` only?

apollo-runtime/src/main/java/com/apollographql/android/impl/ResponseBodyConverter.java
BenSchwab: Good call. The normalizer should be a single use object, so changed it to have `RealApolloCall` send it `ResponseBodyConverter` and then save the records.

Also, good point about `responseNormalizer` crashing. In theory, if it's built right, and we get  a well-structured graphql response, it won't throw an error. On a malformed response, it may throw something. Not sure who would error first -- the generated response reading, or the normalizer in that case.

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -133,7 +137,10 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
       throw new HttpException(response);
     } else {
       try {
-        return responseBodyConverter.convert(response.body());
+        ResponseNormalizer normalizer = cache.responseNormalizer();
+        Response<T> convertedResponse = responseBodyConverter.convert(response.body(), normalizer);
+        cache.cacheStore().merge(normalizer.records());
```
sav007: should we wrap this with own try/catch?

BenSchwab: So the question is if someone has a faulty `cacheStore`, is it better to fail the entire response, or catch the error?

My issue is if we catch the error, how to we communicate that the `cacheStore` is faulty?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/InMemoryCacheStore.java
BenSchwab: I feel that `RecordSet` better describes it's purpose (a in memory data structure to store records). 

I don't think we will keep `InMemoryCacheStore` around for very long. I think we will want to ship with some sort `CacheStore` implementation, but I think it will either be disk based or at least a LRU in-memory cache.

BenSchwab: Now that I think about it, let's not include this at all in the `runtime`. The only thing using it is the tests, so I will move it there. We can cross the bridge of what default to ship with later.

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
sav007: Yeah, we can log but I guess you are right, lets leave it and fail for now

digitalbuddha:
lgtm as well.

PR #261, opened 2017-03-01
Move http cache to appropriate package

sav007:


apollo-runtime/src/main/java/com/apollographql/android/cache/http/CacheInterceptor.java
```diff
@@ -25,23 +15,23 @@
 
   @Override public Response intercept(Chain chain) throws IOException {
     Request request = chain.request();
-    if (shouldSkipCache(request)) {
```
digitalbuddha: Curious why you no like static Imports?

PR #262, opened 2017-03-01
Feature 259/http cache expire after read

### LINKED ISSUE #259: Add ExpireAfterRead cache policy
We should add a ExpireAfterRead cache policy for http cache. The theory is that you can delete http cache first time you read response into normalized cache.

sav007:
Closes #259


digitalbuddha:
Something else I think this needs is in flight logic. Say when my app starts I want to async ask for config query from a dozen different places. Apollo client should be smart about same request is in progress and piggyback on single  response. So like an in flight map?

sav007:
It sounds like Js Relay, so to track active queries and manage bulk notifications of the response, but I guess this is totally different ticket not related to cache

digitalbuddha:
Sounds good I'll review this later tonight.

apollo-runtime/src/test/java/com/apollographql/android/impl/CacheTest.java
```diff
@@ -355,6 +355,20 @@
...
+    assertThat(apolloClient.newCall(new AllPlanets()).execute().isSuccessful()).isTrue();
+    checkCachedResponse("src/test/graphql/allPlanetsResponse.json");
+
+    assertThat(apolloClient.newCall(new AllPlanets()).expireAfterRead().execute().isSuccessful()).isTrue();
+    checkNoCachedResponse();
+    try {
+      apolloClient.newCall(new AllPlanets()).cache().execute();
+      fail("exception expected");
+    } catch (Exception expected) {
+    }
```
digitalbuddha: Let's add 1 more  new call and execute to assert you get a new network response after expiring the last one.

PR #273, opened 2017-03-04
Memoize mappers

### LINKED ISSUE #224: Memoize mappers
Currently we create a new instance of mappers anytime we make a request. Ideally we would Memoize the mappers within ApolloClient do that we can reuse them of same mapper is required

sav007:
Closes #224 

Changes:
- nested mappers instantiated in default constructor (example https://github.com/apollographql/apollo-android/pull/273/files#diff-e0f8aa3bbca9447766d25967d5c39f5fR541)
- top level mappers for operations are kept in hashmap: https://github.com/apollographql/apollo-android/pull/273/files#diff-07ac49222007995fd70e4d0b091b9557

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/SchemaTypeResponseMapperBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/SchemaTypeResponseMapperBuilder.kt
@@ -334,8 +334,7 @@ class SchemaTypeResponseMapperBuilder(
           .map { it.type.withoutAnnotations() }
           .map { it.let { if (it.isList()) it.listParamType() else it } }
           .filter { !it.isScalar() && !it.isCustomScalarType() }
-          .map { it.overrideTypeName(typeOverrideMap) }
-          .map { it as ClassName }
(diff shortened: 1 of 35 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/SchemaTypeResponseMapperBuilder.kt
```diff
@@ -375,19 +327,35 @@ class SchemaTypeResponseMapperBuilder(
...
 
+  private fun mapperFields() =
+      fields
+          .map { it.fieldSpec(customScalarTypeMap = context.customTypeMap) }
+          .plus(inlineFragments.map { it.fieldSpec() })
+          .map { it.type.withoutAnnotations() }
+          .map { it.let { if (it.isList()) it.listParamType() else it } }
+          .filter { !it.isScalar() && !it.isCustomScalarType() }
+          .map { it.overrideTypeName(typeOverrideMap) }
+          .map { it as ClassName }
```
marwanad: nit: combine this operation with the one above?

PR #276, opened 2017-03-06
Bump gradle version to 2.3.0

BenSchwab:


apollo-integration/src/test/java/com/apollographql/android/impl/ResponseNormalizationTest.java
```diff
@@ -37,6 +37,7 @@
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 
+import static android.R.attr.src;
```
digitalbuddha: extra import with no extra code?

Issue #277, opened 2017-03-06
NPE on null argument when building cache key


kassim:
I'm receiving a NPE when making a request with a null argument in my query
structured `search(query: $queryText, country:  first: 10, after: $after)` where the $after variable passed is `null`

stacktrace:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.Object.toString()' on a null object reference
    at com.apollographql.android.api.graphql.Field.orderIndependentKeyForVariableArgument(Field.java:155)
    at com.apollographql.android.api.graphql.Field.orderIndependentKey(Field.java:108)
    at com.apollographql.android.api.graphql.Field.orderIndependentKey(Field.java:127)
    at com.apollographql.android.api.graphql.Field.cacheKey(Field.java:103)
    at com.apollographql.android.impl.ResponseNormalizer.willResolve(ResponseNormalizer.java:57)
    at com.apollographql.android.impl.BufferedResponseReader.willResolve(BufferedResponseReader.java:82)
    at com.apollographql.android.impl.BufferedResponseReader.read(BufferedResponseReader.java:40)
    at com.kassim.Search$Data$Mapper.map(Search.java:653)
    at com.kassim.Search$Data$Mapper.map(Search.java:587)
    at com.apollographql.android.impl.ResponseBodyConverter$1.read(ResponseBodyConverter.java:48)
    at com.apollographql.android.impl.ResponseJsonStreamReader.nextObject(ResponseJsonStreamReader.java:90)
    at com.apollographql.android.impl.ResponseBodyConverter.convert(ResponseBodyConverter.java:43)
    at com.apollographql.android.impl.RealApolloCall.handleResponse(RealApolloCall.java:147)
    at com.apollographql.android.impl.RealApolloCall.execute(RealApolloCall.java:57)
```

possibly related to PR https://github.com/apollographql/apollo-android/pull/247

PR #278, opened 2017-03-06
fix: NPE on null argument when building cache key

kassim:
fixes https://github.com/apollographql/apollo-android/issues/277

NPE thrown when building cache key and one of the query args are null

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
+++ b/apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
@@ -148,7 +148,9 @@ public class Field {
     Object variable = objectMap.get(VARIABLE_NAME_KEY);
     //noinspection SuspiciousMethodCalls
     Object resolvedVariable = variables.valueMap().get(variable);
-    if (resolvedVariable instanceof Map) {
+    if (resolvedVariable == null) {
+      return null;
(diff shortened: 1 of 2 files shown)


apollo-api/src/main/java/com/apollographql/android/api/graphql/Field.java
```diff
@@ -148,7 +148,9 @@ private String orderIndependentKeyForVariableArgument(Map<String, Object> object
     Object variable = objectMap.get(VARIABLE_NAME_KEY);
     //noinspection SuspiciousMethodCalls
     Object resolvedVariable = variables.valueMap().get(variable);
-    if (resolvedVariable instanceof Map) {
```
sav007: maybe I'm missing smth but in case if `resolvedVariable ` is null this will always be false without NPE: `if (null instanceof Map)`

kassim: @sav007 the next line `null.toString()` gives NPE (the diff doesn't show that) 

An alternative could be to return `String.valueOf(resolvedVariable)` which would avoid the need for a null check if that's preferred

sav007: Oh my bad github collapsed next line so didn't see it, thx for explanation

PR #279, opened 2017-03-06
Read normalized cache

### LINKED ISSUE #266: Add read support for the normalized cache
This should be able to closely follow the pattern established by the HttpCache, but we won't be able to use OkHttp interceptors because we need to convert from request -> data. Not request -> responseBody.

In terms of the actual reading we can follow the [iOS approach](https://github.com/apollographql/apollo-ios/blob/master/Sources/GraphQLResultReader.swift#L1) and use `GraphQLResolver`, which is essentially a function field -> Object. For network this function would just access a buffered json map with the field name. For normalized cache we would look up the field and follow references.

Another option, would be to have a `CacheResponseReader` and leave `BufferedResponseReader` as is.

sav007:
Introduces reading from cache.

Closes #266

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
@@ -1,0 +1,210 @@
+package com.apollographql.android.impl;
+
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
+
+import com.apollographql.android.api.graphql.Response;
+import com.apollographql.android.cache.normalized.CacheKeyResolver;
+import com.apollographql.android.impl.normalizer.EpisodeHeroName;
+import com.apollographql.android.impl.normalizer.HeroAndFriendsNames;
+import com.apollographql.android.impl.normalizer.HeroAndFriendsNamesWithIDForParentOnly;
+import com.apollographql.android.impl.normalizer.HeroAndFriendsNamesWithIDs;
+import com.apollographql.android.impl.normalizer.HeroAppearsIn;
+import com.apollographql.android.impl.normalizer.HeroParentTypeDependentField;
+import com.apollographql.android.impl.normalizer.HeroTypeDependentAliasedField;
+import com.apollographql.android.impl.normalizer.SameHeroTwice;
+import com.apollographql.android.impl.normalizer.type.Episode;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+import okhttp3.OkHttpClient;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class NormalizedCacheTestCase {
+  private static final String NORMALIZER_TEST_PATH = "src/test/graphql/com/apollographql/android/impl/normalizer/";
+
+  private ApolloClient apolloClient;
+  private MockWebServer server;
+  private InMemoryCacheStore cacheStore;
+
+  @Before public void setUp() {
+    server = new MockWebServer();
+
+    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    cacheStore = new InMemoryCacheStore();
+
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(okHttpClient)
+        .normalizedCache(cacheStore, new CacheKeyResolver() {
+          @Nullable @Override public String resolve(Map<String, Object> jsonObject) {
+            return (String) jsonObject.get("id");
+          }
+        })
+        .build();
+  }
+
+  private static MockResponse mockResponse(String normalizerFileName) throws IOException {
+    return new MockResponse().setChunkedBody(Files.toString(new File(NORMALIZER_TEST_PATH + normalizerFileName),
+        Charsets.UTF_8), 32);
+  }
+
+  @Test public void episodeHeroName() throws IOException {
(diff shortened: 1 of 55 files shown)


apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
```diff
@@ -0,0 +1,210 @@
...
+    server = new MockWebServer();
+
+    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    cacheStore = new InMemoryCacheStore();
+
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(okHttpClient)
+        .normalizedCache(cacheStore, new CacheKeyResolver() {
+          @Nullable @Override public String resolve(Map<String, Object> jsonObject) {
```
digitalbuddha: this should not be nullable

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -24,32 +29,42 @@
...
   }
 
   @Nonnull @Override public Response<T> execute() throws IOException {
     synchronized (this) {
       if (executed) throw new IllegalStateException("Already Executed");
       executed = true;
     }
+
+    Response<T> cachedResponse = cachedResponse();
+    if (cachedResponse != null) {
```
digitalbuddha: so not part of this pr but it kinda makes me sad that we have null everywhere in api. It would be nice to do something like `Optional`, I wouldn't mind doing the refactor myself if no one has issues. Also open to other suggestions @BenSchwab @sav007 @brianPlummer

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -132,17 +157,53 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
...
+      return null;
+    }
+
+    T cachedData = cachedData();
+    if (cachedData != null) {
+      return new Response<>(operation, cachedData, null);
+    }
+
+    if (cacheControl == CacheControl.CACHE_ONLY) {
+      return new Response<>(operation, null, null);
```
digitalbuddha: pls use constants that represent what null means, its difficult to see if this is correct or not when there are 2 unnamed params of null null.  Also why did ew not write the whole thing in kotlin? :-) too late now I guess

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -132,17 +157,53 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
...
+    }
+
+    return null;
+  }
+
+  @SuppressWarnings("unchecked") private T cachedData() {
```
digitalbuddha: @Nullable pls

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
sav007: Kotlin I guess because it will bring runtime with it, here we use pure java. But I know this is would be nicer to have on Kotlin. Maybe we need to consider some `Optional` approach instead of null

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
sav007: yeah, agree we need to consider using `Optional`

apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
sav007: that how the we ported from iOS I guess. if null that means we are going to use path as a key.
We can change `CacheKeyResolver` to return not String but some class like `CacheKey`. And null replace with `CacheKey.EMPTY` smth like this

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
BenSchwab: +1 at the very least we can use the @Nullable so we can get lint warnings

apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
sav007: will create issue for this

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
digitalbuddha: https://github.com/apollographql/apollo-android/issues/290 Feel free to do it otherwise I'll do it after the sql stuff

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Cache.java
```diff
@@ -49,19 +59,15 @@ public ResponseNormalizer responseNormalizer() {
...
-
     }
-
   }, new CacheKeyResolver() {
     @Nullable @Override public String resolve(Map<String, Object> jsonObject) {
       return null;
     }
   });
 
   //Todo: add interceptor for reading (https://github.com/apollographql/apollo-android/issues/266)
```
BenSchwab: Remove this Todo?

apollo-integration/src/test/java/com/apollographql/android/impl/NormalizedCacheTestCase.java
sav007: https://github.com/apollographql/apollo-android/issues/291

PR #282, opened 2017-03-06
fix BufferedResopnseReader handling null on readCustomType

kassim:
Fixes:https://github.com/apollographql/apollo-android/issues/283

A call to `readerShadow.didParseNull()` was missing from `BufferedResopnseReader.readCustomType()` when handling a null value causing the `ResponseNormalizer`'s `valueStack` to be out of sync

This PR adds the missing call and adds tests for when a null may be handled


digitalbuddha:
thanks for contribution, you're going to be our npe fixer :-)

PR #289, opened 2017-03-07
check if apollo-api exists before adding it

### LINKED ISSUE #288: Allow the user to override the apollo-api dependency
`apollo-api` adds the JSR dependencies with a `compile` scope. 

If conflicts exist and user try to exclude the JSR dependencies through:

```
compile ('com.apollographql.android:api:0.2.2-SNAPSHOT') {
        exclude group: 'com.google.code.findbugs'
        exclude group: 'javax.annotation'
    }
```
those exclusions will not be applied because the plugin would override those.

We need a way to accept that override from the user.

marwanad:
Closes #288 

Users can provide their own api dependency and exclude the JSR dependency and it should work as expected.

(As long as the api version matches the plugin version)

```
compile ('com.apollographql.android:api:0.2.2-SNAPSHOT') {
        exclude group: 'com.google.code.findbugs'
        exclude group: 'javax.annotation'
    }
```

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
@@ -75,6 +75,6 @@ open class GraphQLCompiler {
   companion object {
     const val FILE_EXTENSION = "graphql"
     val OUTPUT_DIRECTORY = listOf("generated", "source", "apollo")
-    const val APOLLOCODEGEN_VERSION = "0.10.8"
+    const val APOLLOCODEGEN_VERSION = "0.10.9"
   }
 }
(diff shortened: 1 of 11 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
```diff
@@ -75,6 +75,6 @@ open class GraphQLCompiler {
   companion object {
     const val FILE_EXTENSION = "graphql"
     val OUTPUT_DIRECTORY = listOf("generated", "source", "apollo")
-    const val APOLLOCODEGEN_VERSION = "0.10.8"
+    const val APOLLOCODEGEN_VERSION = "0.10.9"
```
BenSchwab: Did you check if this changes any of the IR files? (you can run `./gradlew apollo-compiler:updateTestIRFiles`

marwanad: Missed that. It should change the IR files.

PR #293, opened 2017-03-07
Fragments Not Getting Populated, When Fragment Is On Interface

### LINKED ISSUE #222: Fragments Not Getting Populated, When Fragment Is On Interface
Attached are two sample projects: `StaticList` and `StaticFragment`. They are nearly identical, querying [my public test server](https://graphql-demo.commonsware.com/0.1/graphql) and showing a list of trips in a `RecyclerView`.

`StaticList` works just fine with `apollo-android` 0.2.1. It does not use a fragment.

`StaticFragment` changes the GraphQL document from:

```
query getAllTrips {
  allTrips {
    id
    title
    startTime
    priority
    duration
    creationTime
  }
}
```

to:

```
query getAllTrips {
  allTrips {
    ...commonFields
    startTime
    priority
    duration
    creationTime
  }
}

fragment commonFields on Plan {
  id
  title
}
```

(this fragment is pointless, but it's an illustration)

The code gets generated fine. However, the generated `commonFields()` method on the generated `Fragments` class returns `null`.

The problem appears to lie here in the generated code:

```java
          @Override
          public @Nonnull Fragments map(ResponseReader reader) throws IOException {
            CommonFields commonFields = null;
            if (conditionalType.equals(CommonFields.TYPE_CONDITION)) {
              commonFields = new CommonFields.Mapper(factory.commonFieldsFactory()).map(reader);
            }
            return factory.creator().create(commonFields);
          }
        }
```

`CommonFields.TYPE_CONDITION` is `Plan`, because that is what the fragment calls for. `allTrips` returns `[Trip!]!`, and `Trip` implements `Plan`. Hence, the `if` check here fails (`Trip` does not equal `Plan`), and so the `CommonFields` fragment does not get populated.

AFAICT, my GraphQL syntax is fine, and GraphiQL is happy as a clam with it.

[StaticList.zip](https://github.com/apollographql/apollo-android/files/788546/StaticList.zip)
[StaticFragment.zip](https://github.com/apollographql/apollo-android/files/788548/StaticFragment.zip)

sav007:
Closes #222


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Fragment.kt
```diff
@@ -3,14 +3,16 @@ package com.apollographql.android.compiler.ir
 import com.apollographql.android.compiler.Annotations
 import com.apollographql.android.compiler.ClassNames
 import com.apollographql.android.compiler.SchemaTypeSpecBuilder
+import com.squareup.javapoet.CodeBlock
 import com.squareup.javapoet.FieldSpec
 import com.squareup.javapoet.TypeSpec
+import java.util.*
```
marwanad: nit: no need for wildcard import?

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Fragment.kt
```diff
@@ -3,14 +3,16 @@ package com.apollographql.android.compiler.ir
...
 import javax.lang.model.element.Modifier
 
 data class Fragment(
     val fragmentName: String,
     val source: String,
-    val typeCondition: String,
+    val possibleTypes: List<String>,
```
marwanad: just making sure, `possibleTypes` would always exist in the fragment object in the IR (It's not going to be optional in any case)?

sav007: that is the idea, @martijnwalraven mentioned that on iOS they don't use `typeCondition ` at all, so we can say `possibleTypes` is mandatory

marwanad:
@sav007 might need to rebase since the 0.10.9 change is in master already.

PR #306, opened 2017-03-09
Separate cache control for http and normalized caches

### LINKED ISSUE #281: [Normalized Cache] CacheControl should be updated to serve both http and normalized cache
`com.apollographql.android.cache.http.HttpCache.CacheControl` is designed particularly for http cache. It should be updated to serve both http and normalized caches according to this doc:
https://quip.com/BxU7A5nRM4JW

sav007:
Closes #281


brianPlummer:
lgtm

apollo-runtime/src/main/java/com/apollographql/android/cache/http/CacheInterceptor.java
```diff
@@ -67,33 +51,51 @@ private Response cacheOnlyResponse(Request request) throws IOException {
...
-    cacheResponse.close();
     return networkResponse.newBuilder()
         .cacheResponse(Utils.strip(cacheResponse))
         .build();
   }
 
+  private Response cacheFirst(Request request, Chain chain, String cacheKey) throws IOException {
+    Response cacheResponse = cache.read(cacheKey, shouldExpireAfterRead(request));
+    if (cacheResponse == null || cache.isStale(cacheResponse)) {
+      Response networkResponse = withServedDateHeader(chain.proceed(request));
```
digitalbuddha: hey so something that I just thought of now. I sort of remember some logic in okhttp cache that would strip timestamps prior to finding the cache key. Did we do a similar?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -61,13 +64,14 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
       executed = true;
     }
 
-    Response<T> cachedResponse = cachedResponse();
-    if (cachedResponse != null) {
-      return cachedResponse;
+    if (cacheControl == CacheControl.CACHE_ONLY || cacheControl == CacheControl.CACHE_FIRST) {
+      Response<T> cacheResponse = cacheFirstResponse();
+      if (cacheResponse != null) {
```
digitalbuddha: nulls make me sad, I'll keep point it out :-)

apollo-runtime/src/main/java/com/apollographql/android/cache/http/CacheInterceptor.java
sav007: What do you mean?
`withServedDateHeader ` adds custom http header (`APOLLO-SERVED-DATE`) that represents the time of the request. This header will be used for checking if it stale or not

digitalbuddha: Say you have value object:
class RequestPojo{
String myParam
String timestamp=new Date().getTime();
}

That request will always have a cache miss when trying to figure out if it exists in cache.  A common practice in building caches is to null out all data prior to creating the "Cache Key" in our case it would be prior to the md5 calculation.

digitalbuddha: ignore me, only header dates are erased and you are doing that too.

Issue #305, opened 2017-03-09
apollo-integration isn't using the project version of the plugin


marwanad:
Made some changes to the compiler (and hence the plugin) but those weren't picked by the apollo-integration.

One fix would be to change the apollo-integration module to a plugin test source set instead of a project module. I was thinking of creating an `integrationTest` configuration and having that run after the plugin unit tests. (We can keep the folder structure the same (ie. move all integration tests under `apollo-android/apollo-integration`)

Initially, I tried having the plugin under `buildSrc` but that was a pain because the plugin depends on other submodules (apollo-compiler and indirectly apollo-api) so we can’t assume it’s a “separate build” and use `buildSrc` for that without introducing further complexities.

marwanad:
@BenSchwab Are you fine with those changes/have other suggestions?

PR #310, opened 2017-03-10
Integration test configuration

marwanad:
(Diff is big due to moving files around / adding expected java outputs to integration tests)

- This creates an `integrationTest` configuration and an integrationTest tasks. This ensures that the integrationTests always run after the plugin unit tests. 

- Moves the integration tests to a separate folder outside the plugin.

- Refactors the integration tests to be Paramterized and verify that the outputs are as expected instead of just checking if the output File exists.

The relevant added test files are under this [directory](https://github.com/apollographql/apollo-android/tree/integrationTest-configuration/apollo-integration-tests/kotlin/com/apollographql/android/gradle/integration).

And the test fixtures are [here](https://github.com/apollographql/apollo-android/tree/integrationTest-configuration/apollo-integration-tests/testFixtures).

This [folder](https://github.com/apollographql/apollo-android/tree/integrationTest-configuration/apollo-integration-tests/testFixtures/normalizer) was copied from the `apollo-integration` module and still contains the same tests under [src/test](https://github.com/apollographql/apollo-android/tree/integrationTest-configuration/apollo-integration-tests/testFixtures/normalizer/src/test). 
Closes of #305


marwanad:
For some reason, you can't runt those through AS. 

Only running through `gradle integrationTest` works. I'll close now and reopen later.

Issue #304, opened 2017-03-09
Add QueryWatching


BenSchwab:
Proposed changes:

- Add watch to `ApolloCall` : ` @Nonnull ApolloCall<T> watch(@Nullable Callback<T> callback);` (This callback can / will be called multiple times)
- `Watch` first calls `enqueue`, respecting caching preference, and then subscribes to `RealCache`'s changed keys with the queries `dependentKeys` (iff first `execute` is successful)
- Have `CacheStore` return changed keys on write, `RealCache` aggregates and dispatches to subscribers
- In subscribe, if dependent keys !disjoint to change keys, refetch watched query

Note, this will add more logic bloat to `RealApolloCall`

sav007:
`dependentKeys` does it holds all keys been used by query?

sav007:
Could you please elaborate more on `In subscribe, if dependent keys !disjoint to change keys, refetch watched query`. I thought if `ApolloCall` receives notification about dependentKeys changes that it depends on, then we refetch query

BenSchwab:
"In subscribe, if dependent keys !disjoint to change keys, refetch watched query"

Is worded poorly. `ApolloCall` or `ApolloWatcher` will refetch when a dependentKey changes. This occurs when dependKeys for one query is not disjoint from the set of changedKeys from another query.

PR #312, opened 2017-03-10
Add Query Watching

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/304

This is a WIP query watching PR. There will be many more tests, naming will be reconsidered, TODOs fixed etc...

But wanted to get early thoughts on general design.

 I played around with couple ways, but settled on have an the flow `ApolloCall.toWatcher()` -> `ApolloWatcher`

The reason is that in order to watch a query you need to execute the query at least once to generate the `dependentKeys`. So a user would first need to configure the "root query" with whatever cache control they want, and then transform it into a watcher.

Then they call `enqueueAndWatch` which performs the initial fetch. Upon successful fetch, the watcher then subscribes to change keys events on the cache.

Importantly, `refetch` cachePolicy is separate than the initial fetch policy.

One thing to take a look at: whenever a query is refetched, we actually unsubscribe from cache changed keys event, until the query is returned with new `dependentKeys`. This is because when there is an outgoing request, the current dependentKeys are "stale". When we move to more sophisticated threading model this will be important.


apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+  };
+
+  public RealApolloWatcher(RealApolloCall<T> originalCall, Cache cache) {
+    activeCall = originalCall;
+    this.cache = cache;
+  }
+
+  @Nonnull public WatcherSubscription enqueueAndWatch(@Nullable final ApolloCall.Callback<T> callback) {
+    this.callback = callback;
```
sav007: Should we add safe guard if we already watching? I mean we don't support multiple subscription right?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      @Override public void onFailure(@Nonnull Exception e) {
+        sourceCallback.onFailure(e);
+      }
+    };
+  }
+
+  private void fetch() {
+    activeCall.enqueue(callbackProxy(callback, activeCall));
```
sav007: Should be careful, active call can't be run multiple times, should it be `activeCall.clone()`?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Nonnull public RealApolloWatcher<T> refetchCacheControl(@Nonnull CacheControl cacheControl) {
+    Utils.checkNotNull(cacheControl, "cacheControl == null");
+    this.refetchCacheControl = cacheControl;
+    return this;
+  }
+
+  @Nonnull public RealApolloWatcher<T> refetch() {
+    activeCall.cancel(); //Todo: is this necessary / good? We don't want people to chain refetch().refetch()
+    activeCall = activeCall.clone().cacheControl(refetchCacheControl);
```
sav007: I guess this line should be moved to `fetch` function, it will be easier to understand that we actually doing activeCall.clone() before enqueue

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -74,7 +76,7 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
     return executeNetworkRequest();
   }
 
-  @Nonnull @Override public ApolloCall<T> enqueue(@Nullable final Callback<T> callback) {
+  @Nonnull @Override public void enqueue(@Nullable final Callback<T> callback) {
```
sav007: what the reason to remove return value?

sav007: And can we have like this: `apollocall.enqueue(callback).watch()` ?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java
```diff
@@ -9,22 +15,35 @@
 public final class RealCache implements Cache {
   private final CacheStore cacheStore;
   private final CacheKeyResolver cacheKeyResolver;
+  private Map<RecordChangeSubscriber, Set<String>> subscribers;
```
sav007: Should we make `RecordChangeSubscriber ` as weak ref? So WeakHashMap?
Then it will be user responsibility to keep around subscribtion?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
BenSchwab: `enqueue` should only be called once -- I can add some safeguards for that. But `enqueue` just calls the ApolloCall that this was transformed from, so that should never have been executed before.

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
BenSchwab: We need to guarantee that the a Call has been executed when we watch. Otherwise we have no dependentKeys to monitor. So I don't like relying on have the user chain `apollocall.enqueue.watch()`. Instead I have it has `apolloCall.toWatcher().enqueueAndWatch()`.

Is there a value in having `enqueue` return an ApolloCall? OkHttp has void return null. It just seems dangerous. Once you call `enqueue` many functions on that `ApolloCall` now explode, and returning it seems to imply that the object is still safe to use.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java
BenSchwab: 👍 Good Idea!

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
BenSchwab: 👍

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
sav007: > Is there a value in having enqueue return an ApolloCall

no just was curious, thx makes sense!!

sav007:
One more question do we have a vision how query watching will utilize RxJava?

digitalbuddha:
It's going to use rxjava fromEmitter/create api which can wrap any async call into rx. I have good idea how to get that piece in.

digitalbuddha:
Here was original idea for rx support. I don't see why it won't work for this too. https://gist.github.com/digitalbuddha/3c1dbccf059f0279e46e540b2c5fd89e

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
BenSchwab: So we actually aren't cloning before `enqueue` only from refetch. Per other comments, it is only possible to transform a `ApolloCall` to and `ApolloWatcher` if it has not been executed. However, calls to `refetch` will clone the `ApolloCall` because in this case it has been used. I haven't convinced myself that `refetch` as a lot of value -- but added it because it is part of the iOS watcher interface.

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,83 @@
...
+import com.apollographql.android.api.graphql.Operation;
+import com.apollographql.android.api.graphql.Response;
+import com.apollographql.android.cache.normalized.Cache;
+import com.apollographql.android.cache.normalized.CacheControl;
+import com.apollographql.android.impl.util.Utils;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+final class RealApolloWatcher<T extends Operation.Data> implements ApolloWatcher<T> {
```
sav007: Do we need to have `cancel` or `dispose` method, to cancel watching via `RealApolloWatcher` as well not only via `WatcherSubscription `

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,83 @@
...
+    return watcherSubscription;
+  }
+
+  @Nonnull public RealApolloWatcher<T> refetchCacheControl(@Nonnull CacheControl cacheControl) {
+    Utils.checkNotNull(cacheControl, "cacheControl == null");
+    this.refetchCacheControl = cacheControl;
+    return this;
+  }
+
+  @Nonnull public RealApolloWatcher<T> refetch() {
```
sav007: I guess we can merge `refetchCacheControl`with this method, so to have 2 API calls:
`public RealApolloWatcher<T> refetch(@Nonnull CacheControl cacheControl)`
`public RealApolloWatcher<T> refetch()`

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -110,10 +112,13 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
         httpCall.enqueue(new HttCallback(callback));
       }
     });
-    return this;
   }
 
-  @Nonnull @Override public ApolloCall<T> httpCacheControl(@Nonnull HttpCacheControl httpCacheControl) {
+  @Nonnull @Override public RealApolloWatcher<T> toWatcher() {
```
sav007: What happens if I call this multiple times for the same `RealApolloCall`

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -110,10 +112,13 @@ private RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory http
...
+    return new RealApolloWatcher<>(this, cache);
```
sav007: shouldn't we call `new RealApolloWatcher<>(clone(), cache)`?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java
```diff
@@ -12,33 +20,41 @@
   private final CacheStore cacheStore;
   private final CacheKeyResolver cacheKeyResolver;
   private final ReadWriteLock lock;
+  private Map<RecordChangeSubscriber, Set<String>> subscribers;
```
sav007: final

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
BenSchwab: Overloading seems fair -- but we need to keep `refetchCacheControl` because it plays a roll in `enqueueAndWatch` because when dependenKeys change `refetchCacheControl` is used to make the new request.

sav007: My intention was to make it as much immutable as possible, after we called enqueueAndWatch what if we change the cache control `refetchCacheControl`? Should we add safe guard to prevent user from changing cache control after he called  `enqueueAndWatch `?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
BenSchwab: With `clone`, this would be acceptable. I don't see a real harm in allowing it, but it does go against the explicit single-use immutable style of the other methods.

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,81 @@
...
+      refetch();
+    }
+  };
+
+  RealApolloWatcher(RealApolloCall<T> originalCall, Cache cache) {
+    activeCall = originalCall;
+    this.cache = cache;
+  }
+
+  public void enqueueAndWatch(@Nullable final ApolloCall.Callback<T> callback) {
```
sav007: nit: @Override

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloWatcher.java
```diff
@@ -0,0 +1,81 @@
...
+    synchronized (this) {
+      if (executed) throw new IllegalStateException("Already Executed.");
+      executed = true;
+    }
+    this.callback = callback;
+    activeCall.enqueue(callbackProxy(this.callback, activeCall));
+  }
+
+  @Nonnull public RealApolloWatcher<T> refetchCacheControl(@Nonnull CacheControl cacheControl) {
```
sav007: nit: @Override

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -14,7 +14,9 @@
 
   @Nonnull Response<T> execute() throws IOException;
 
-  @Nonnull ApolloCall<T> enqueue(@Nullable Callback<T> callback);
+  void enqueue(@Nullable Callback<T> callback);
+
+  @Nonnull ApolloWatcher<T> toWatcher();
```
sav007: nit: maybe simple `watcher()` ?

BenSchwab:
Thanks for the reviews!

PR #313, opened 2017-03-10
Introduce Apollo dispatcher for async calls

### LINKED ISSUE #280: [Normalized Cache] Reading from cache should be operated on own executor service
Reading from cache in case `com.apollographql.android.impl.RealApolloCall#enqueue` called should be done on designated `ExecutorService`.

sav007:
Closes #280 

Introduces dispatcher (`ExecutorService`) with unbound threads count and TTL 60 sec.
This dispatcher is going to be used when:
- `com.apollographql.android.impl.RealApolloCall#enqueue` called
- `com.apollographql.android.impl.RealApolloCall#handleResponse` saves to the cache.

`RealCache` gets `ReadWriteLock` for any cache operation (one thread can write to the cache, multiple can read).

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/android/impl/AsyncNormalizedCacheTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/android/impl/AsyncNormalizedCacheTestCase.java
@@ -1,5 +1,7 @@
 package com.apollographql.android.impl;
 
+import com.google.common.base.Strings;
+
 import android.support.annotation.NonNull;
 
 import com.apollographql.android.ApolloCall;
@@ -44,7 +46,7 @@ public class AsyncNormalizedCacheTestCase {
         .normalizedCache(cacheStore, new CacheKeyResolver() {
           @Nonnull @Override public CacheKey resolve(@NonNull Map<String, Object> jsonObject) {
             String id = (String) jsonObject.get("id");
-            if (id == null || id.isEmpty()) {
+            if (Strings.isNullOrEmpty(id)) {
               return CacheKey.NO_KEY;
             }
             return CacheKey.from(id);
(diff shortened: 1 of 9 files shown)


apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
```diff
@@ -147,6 +156,12 @@ public void toJson(JsonWriter writer, T value) throws IOException {
       return this;
     }
 
+    public Builder dispatcher(@Nonnull ExecutorService dispatcher) {
```
marwanad: I might be missing something here, but is there any reason for us to expose this API if the dispatcher would only be used for internal stuff like enqueuing and cache writes?

apollo-integration/src/test/java/com/apollographql/android/impl/AsyncNormalizedCacheTestCase.java
```diff
@@ -0,0 +1,89 @@
...
+    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    cacheStore = new InMemoryCacheStore();
+
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(okHttpClient)
+        .normalizedCache(cacheStore, new CacheKeyResolver() {
+          @Nonnull @Override public CacheKey resolve(@NonNull Map<String, Object> jsonObject) {
+            String id = (String) jsonObject.get("id");
+            if (id == null || id.isEmpty()) {
```
digitalbuddha: Strings.isNullOrEmpty()

apollo-runtime/src/main/java/com/apollographql/android/impl/ApolloClient.java
sav007: first it used in tests, maybe user wants to pass own executor to test as well

PR #315, opened 2017-03-10
Close stale cache responses and fix build

sav007:


NightlyNexus:
`okhttp3.internal` is not a public API, and `ResponseBody.close()` already closes quietly.

Issue #303, opened 2017-03-09
Efficiently handle batched reads/writes in SQLStore


BenSchwab:
We always will batch writes, and reads will usually be batched, so we should make them as efficient as possible in SQLStore. This may involve transactions, or writing new multi-target queries .

PR #319, opened 2017-03-13
Run batch insert in a transaction

BenSchwab:
- Only insert if fields actually change
- Run batch insert in transaciton
- Fix potential NPE on updating a field which was, and is, `null`

closes https://github.com/apollographql/apollo-android/issues/303

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/sql/SqlStore.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/sql/SqlStore.java
@@ -7,9 +7,11 @@ import android.database.sqlite.SQLiteStatement;
 import com.apollographql.android.cache.normalized.CacheStore;
 import com.apollographql.android.cache.normalized.Record;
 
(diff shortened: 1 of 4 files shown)


BenSchwab:
Looks like we still have spurious failures :(

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/sql/SqlStore.java
```diff
@@ -46,15 +48,28 @@ private SqlStore(ApolloSqlHelper dbHelper, FieldsAdapter parser) {
...
+    return changedKeys;
+  }
+
+  @Nonnull @Override public Set<String> merge(Collection<Record> recordSet) {
+    Set<String> changedKeys = Collections.emptySet();
+    try {
+      database.beginTransaction();
+      changedKeys = super.merge(recordSet);
+    } finally {
+      database.endTransaction();
```
sav007: I believe we are missing `setTransactionSuccessful()`?

PR #320, opened 2017-03-13
shaded Guava Optional

digitalbuddha:
Pulled out Guava Optional and Tests.  I removed anything that needed a `Function` or `Supplier`


apollo-runtime/src/test/java/com/apollographql/android/internal/OptionalTest.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
```
marwanad: are those classes available for us?

digitalbuddha: It auto imported, I think we have guava in tests already

digitalbuddha:
Also shading did not work well as it was an "all or nothing" and I did not want to bring in annotations as dependencies that we don't really need. Since it was only 3 files I moved them in.  (going offline for few hrs, feel free to merge otherwise I"ll fix tonight or am)

apollo-runtime/src/test/java/com/apollographql/android/internal/OptionalTest.java
marwanad: Oh right, truth or mockito has it

sav007:
Could you please move these to :apollo-api module as we are going to use it in (compiler) generated classes as well, so I guess it's more reasonable to be there instead of runtime

sav007:
And regarding this:
> I removed anything that needed a Function or Supplier

But that what user needs:
```
String productImageUrl = product
    .transform(product -> product.imageConnection)
    .transform(connection -> connection.edges)
    .transform(edges -> edges.get(0))
    .transform(edge -> edge.image())
    .transform(image -> image.src())
    .or("")
```

sav007:
How about this port: https://github.com/dmstocking/support-optional
It's the most close to Java8

digitalbuddha:
Not the craziest about taking a lib in that has 11 stars. Is the transform stuff the only thing missing from this PR, I can move use Func1 interface for that...

sav007:
Yeah, I guess we are missing only `transform `

digitalbuddha:
Cool got a few things to iron out at  and then will move this to other module. I hit a bit of snag since no guava there and no precondition utils in api test but nothing blocking ☺️

PR #321, opened 2017-03-14
check if guava exists in the list of resolved dependencies

marwanad:
This is needed for choosing which Optional class to go with when doing the
codegen.

List of resolved dependencies includes those with an explicit `compile 'x'` statement as well as whatever other transitive dependencies they bring.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
@@ -11,7 +11,7 @@ open class GraphQLCompiler {
   private val moshi = Moshi.Builder().build()
   private val irAdapter = moshi.adapter(CodeGenerationIR::class.java)
 
-  fun write(irFile: File, outputDir: File, customTypeMap: Map<String, String> = emptyMap()) {
+  fun write(irFile: File, outputDir: File, customTypeMap: Map<String, String> = emptyMap(), hasGuava: Boolean = false) {
     val ir = irAdapter.fromJson(irFile.readText())
     val irPackageName = irFile.absolutePath.formatPackageName()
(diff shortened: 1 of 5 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/GraphQLCompiler.kt
```diff
@@ -11,7 +11,7 @@ open class GraphQLCompiler {
   private val moshi = Moshi.Builder().build()
   private val irAdapter = moshi.adapter(CodeGenerationIR::class.java)
 
-  fun write(irFile: File, outputDir: File, customTypeMap: Map<String, String> = emptyMap()) {
+  fun write(irFile: File, outputDir: File, customTypeMap: Map<String, String> = emptyMap(), hasGuava: Boolean) {
```
marwanad: can move to a data class if this gets ugly

apollo-gradle-plugin/src/main/groovy/com/apollographql/android/gradle/ApolloPlugin.groovy
```diff
@@ -60,54 +61,61 @@ class ApolloPlugin implements Plugin<Project> {
...
         }
-        project.getGradle().removeListener(this)
       }
 
       @Override
-      void afterResolve(ResolvableDependencies resolvableDependencies) {}
+      void afterResolve(ResolvableDependencies resolvableDependencies) {
+        resolvableDependencies.getResolutionResult().allDependencies.each {
+          if (it.toString().contains("com.google.guava:guava")) {
+            hasGuava = true
```
sav007: can we simplify: `hasGuava |=  it.toString().contains("com.google.guava:guava")`?

marwanad:
I add the travis cache again. This should fix the flakiness, hopefully.

Issue #290, opened 2017-03-07
Introduce Optionals


digitalbuddha:
We have had 2 bugs and contributions due to NPE somewhere we did not expect. I'd like to add Optional to the project.  I have no preference to our own smaller `Optional` type or to bring in a library that does it.  I can't imagine this being more than 20 methods.

PR #322, opened 2017-03-15
Generate Optional instead of null

sav007:
Part of #290 

Variables generation remains the same due to the plan of refactoring to simplify builder. So will be done in next PR.

Input object generation remains the same as they being used in serialization of request body via moshi.


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ClassNames.kt
```diff
@@ -4,6 +4,7 @@ import com.apollographql.android.api.graphql.Mutation
 import com.apollographql.android.api.graphql.Operation
 import com.apollographql.android.api.graphql.Query
 import com.apollographql.android.api.graphql.ResponseReader
+import com.apollographql.android.api.graphql.internal.Optional
```
marwanad: outside the scope of the PR, but I feel Optional shouldn't be in the `internal` package since we're exposing that API to consumers.

sav007:
@marwanad could you pls review again I add new extension to our plugin to switch on/off Optional generation

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloExtension.java
```diff
@@ -8,6 +8,7 @@
 public class ApolloExtension {
   static final String NAME = "apollo";
   private Map<String, String> customTypeMapping = new LinkedHashMap<>();
+  private boolean optionalSupport = false;
```
marwanad: nit: rename to `optionalEnabled`?

apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloClassGenTask.java
```diff
@@ -37,7 +40,11 @@ void generateClasses(IncrementalTaskInputs inputs) {
     inputs.outOfDate(new Action<InputFileDetails>() {
       @Override
       public void execute(InputFileDetails inputFileDetails) {
-        new GraphQLCompiler().write(inputFileDetails.getFile(), outputDir, customTypeMapping, hasGuavaDep);
+        NullableValueType nullableValueType = NullableValueType.NONE;
```
marwanad: I think it would be better if we keep this logic part of the compiler.

Might as well create a data class `GraphQLCompilerArgs` to wrap all the arguments and doing these checks in `GraphQLCompiler.write()`?

Issue #327, opened 2017-03-16
Cache read locking needs to be done at the request level, not cache level


BenSchwab:
We read a request from the cache in multiple calls - we need to lock for the entire request read, and not on each record read, because another request could acquire the write lock in between reads.

For symmetry, we should move write locker up a level too -- even though writing is only called once for a request.

sav007:
>because another request could acquire the write lock in between reads 

Why this is an issue?

sav007:
> For symmetry, we should move write locker up a level too -- even though writing is only called once for a request.

Isn't it what we have right now: https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java#L44

BenSchwab:
We call read inside the field resolver:

```
private Record valueFor(Record record, Field.ObjectField field) {
    CacheReference cacheReference = fieldValue(record, field);
    return cacheReference != null ? cache.read(cacheReference.key()) : null;
  }
```

This measn read is called many times for one request -- this means we unlock/ lock many times for one request.

Thus, a write could interleave during the request read. This means that for one request half of it's response might be read from the cache before a write changed a bunch of values, and half after the write changed a bunch of values.

sav007:
Oh, wow... you are right, I thought we lock on request level

PR #329, opened 2017-03-16
Add read lock which spans entire read operation for a request

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/327

This changes the `Cache` interface to return `read` and `write` "transaction" which continue relevant methods for actually reading and writing.

This is necessitated by the fact that when reading a request from the cache, we call `read` many different times for a single request. We cannot have a `write` interleave during any of these `read`s, so we need a way to have a "read transaction".

To prevent a cache from being used improperly during a read transaction, the actual `read` and `write` methods are moved behind transaction facades. As a transaction will only be returned/started when the internal cache lock allows it, this guarantees that the cache will not be misused.

One peculiarity is that writes do not need the same multi-operation support as read (we call write just one for a response). In fact, fully supporting a `WriteTransaction` with a finish would make it less efficient to call query watching subscribers. So, instead I opted to have `WriteTransaction` have a `writeAndFinish` method. This allows `RealCache` to release the lock before notifying subscribers.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/http/ResponseBodyProxy.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/http/ResponseBodyProxy.java
@@ -72,7 +72,7 @@ final class ResponseBodyProxy extends ResponseBody {
         bytesRead = responseBodySource.read(sink, byteCount);
       } catch (IOException e) {
         if (!closed) {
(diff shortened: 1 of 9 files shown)


BenSchwab:
Actually, I don't like that transactions can call read / write again after calling close. Will refactor.

PR #330, opened 2017-03-16
And read/write transactions to cache interface

BenSchwab:
Closes https://github.com/apollographql/apollo-android/issues/327

- Adds read/write transactions. Each transaction is facade over RealCache, surfacing only the methods that are safe to call in each transaction

To better match the general apollo architecture: https://airbnb.quip.com/wsn6AS0cYH84
- changes key subscriber to have key filtering done in subscribers


apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java
```diff
@@ -1,77 +1,86 @@
...
-        subscriberEntry.getKey().onDependentKeysChanged();
-      }
+    for (RecordChangeSubscriber subscriber : iterableSubscribers) {
+      subscriber.onCacheKeysChanged(changedKeys);
     }
   }
 
-  @Override public Record read(@Nonnull String key) {
+  @Override public RealReadTransaction readTransaction() {
     lock.readLock().lock();
```
sav007: Not sure this is good idea to lock in one place but unlock in `RealReadTransaction` (via closeRead).
It would be better do it in one place `RealReadTransaction` so pass lock object to `RealReadTransaction`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealReadTransaction.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Nullable @Override public Record read(@Nonnull String key) {
+    if (closed) {
+      throw new IllegalStateException("Cannot call read on a transaction which has already closed.");
+    }
+    return cache.read(key);
+  }
+
+  @Override public Collection<Record> read(@Nonnull Collection<String> keys) {
```
sav007: Should it be more generic like:

```
<R> R execute(Transactional<Cache, R> action) {
  return action.call(cache);
}
```

Usage:
```
transaction.execute(new Transactional<>() {
  public Collection<Record> call(Cache cache) {
    return cache.read(keys);  
  }
})
```

Even maybe we can merge Read and Write transactions to one class

BenSchwab:
@sav007 Took a crack at the generic functional approach. Agree that keeping the locking logic in one spot is a lot cleaner!

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RealCache.java
```diff
@@ -1,77 +1,94 @@
...
   }
 
-  @Override public Record read(@Nonnull String key) {
-    lock.readLock().lock();
-    try {
-      return cacheStore.loadRecord(checkNotNull(key, "key == null"));
-    } finally {
-      lock.readLock().unlock();
-    }
+  @Override public <R> Transaction<ReadableCache, R> readTransaction() {
```
sav007: Not sure if it makes sense, but we can simplify API even more, so instead returning the transaction why don't pass Transactional as a parameter here:

```
@Override public <R> R executeWithReadTransaction(Transactional<ReadableCache, R> transactional) {
  try {
    lock.readLock().lock();
    return transactional.call(RealCache.this);
  } finally {
    lock.readLock().unlock();
  }
}
```

The same for write. In this case we can rid off one indirection `Transaction` class at all.

BenSchwab:
Yep - you are 100% right the two levels of indirection is unnecssary

PR #331, opened 2017-03-16
Make Response object return optional if such config selected

### LINKED ISSUE #290: Introduce Optionals

sav007:
Closes #290 

Address issue with `Response` data field doesn't return data wrapped with optional.

Take a look at https://github.com/apollographql/apollo-android/pull/331/files#diff-bd1f7dd25c2f308ad46f434f16893f11 and https://github.com/apollographql/apollo-android/pull/331/files#diff-1d4349ba6464dd54d277f125b3253316 as example how it will be used.


apollo-api/src/main/java/com/apollographql/android/api/graphql/Operation.java
```diff
@@ -6,14 +6,16 @@
...
   /** TODO */
   String queryDocument();
 
   /** TODO */
   V variables();
 
-  ResponseFieldMapper<? extends Operation.Data> responseFieldMapper();
+  ResponseFieldMapper<D> responseFieldMapper();
+
+  T wrapData(D data);
```
sav007: will be called by ResponseConverter

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.java
```diff
@@ -44,12 +44,17 @@ public String queryDocument() {
   }
 
   @Override
+  public Optional<TestQuery.Data> wrapData(TestQuery.Data data) {
+    return Optional.fromNullable(data);
```
sav007: Example how data will be wrapped with optional

apollo-runtime/src/main/java/com/apollographql/android/impl/HttpResponseBodyConverter.java
```diff
@@ -55,7 +55,7 @@
         }
       }
       jsonReader.endObject();
-      return new Response<>(operation, data, errors);
+      return new Response<>(operation, operation.wrapData(data), errors);
```
sav007: this call will wrap data

BenSchwab:
Do we have any tests that test what happens when `optional` is disabled?

marwanad:
@BenSchwab There aren't any for the plugin extension but there are for the [compiler generation](https://github.com/apollographql/apollo-android/pull/322/files#diff-be5ed569a931b8335d61a5097b179cccR77)

sav007:
Yes, that correct. We can have 2 Optional implementations: guava or our own shadowed version. So boolean field is not enough. Plus our Runtime lib knows nothing about guava. So only if user uses Guava we will wrap data (via method `wrapData`) to guava Optional, otherwise we will wrap it into our shadowed Optional

sav007:
@digitalbuddha @marwanad any comments?

digitalbuddha:
Not from first glance need to look tonight

Issue #328, opened 2017-03-16
Reading from cache needs a ReaderShadow to collect dependent keys


BenSchwab:
if a request is cache only, we will never collect the dependent keys necessary for query watching because we currently do not currently shadow cache reading. This results in query watching not working for when a cache-only request is used.

PR #333, opened 2017-03-17
Shadow cache read to generate dependent keys.

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/328

This enables query watching when the initial request was "CACHE_ONLY"

- Generifies `ResponseNormalizer` and `cacheKeyResolver` so we can resolve a key from a Record (just return the key)

- Moves dependentKeys inside of Response. This reduces mutable state stored in `ApolloCall`.

- temporarily disables integration tests (ran locally on mavenLocal() and they pass) because of API change. Will turn back on once snapshot is regenerated.

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/android/impl/ApolloWatcherTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/android/impl/ApolloWatcherTest.java
@@ -43,7 +43,7 @@ public class ApolloWatcherTest {
     apolloClient = ApolloClient.builder()
         .serverUrl(server.url("/"))
         .okHttpClient(okHttpClient)
-        .normalizedCache(cacheStore, new CacheKeyResolver() {
+        .normalizedCache(cacheStore, new CacheKeyResolver<Map<String, Object>>() {
           @Nonnull @Override public CacheKey resolve(@NonNull Map<String, Object> jsonObject) {
             String id = (String) jsonObject.get("id");
             if (id == null || id.isEmpty()) {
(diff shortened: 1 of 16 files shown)


apollo-integration/src/test/java/com/apollographql/android/impl/ApolloWatcherTest.java
```diff
@@ -234,6 +234,58 @@ public void rewatchCacheControl() throws IOException, InterruptedException {
...
+      }
+    });
+    cacheWarmUpLatch.await();
+
+    //Cache is now "warm" with response data
+
+    final CountDownLatch firstResponseLatch = new CountDownLatch(1);
+    final CountDownLatch secondResponseLatch = new CountDownLatch(2);
+
+    ApolloWatcher<EpisodeHeroName.Data> watcher = apolloClient.newCall(query)
```
digitalbuddha: Is only way to set cache policy per request? Is it possible to set default cache policy for all requests?

BenSchwab: Good point. Created an issue: https://github.com/apollographql/apollo-android/issues/335

sav007:
@BenSchwab not related to this PR, but I noticed that notifying subscribers in `com.apollographql.android.cache.normalized.RealCache#write` might produce 'ConcurrentModificationException' for  `Map<RecordChangeSubscriber, Set<String>> iterableSubscribers = new LinkedHashMap<>(subscribers);` we are iterating subscribers but at the same time from other thread new subscriber has been registered

BenSchwab:
@sav007  I believe https://github.com/apollographql/apollo-android/pull/330/files#diff-74663566d863b56d4de4c7bb71c3ec89R34 fixes that. 

We shallow copy the list before iterating in a `synchronized` block. afaik this should be sufficient for preventing concurrent modification. (Also, I believe that sets do not produce concurrent modification exceptions because of how their iterator is written). lmk if you have a cleaner pattern, but the array copy should be lightweight as I don't there will be a large number active subscribers at any one time.

 If a subscriber unsubscribes while iterating across subscribers, there is an internal check in apollo watcher to ensure that we don't refetch after `unsubscribe` is called

PR #336, opened 2017-03-20
Generate operation builder instead of variables builder

sav007:
Move variables builder generation upper to operation.
So creation of query with arguments will be next:
```
TestQuery.builder().episode(Episode.NEWHOPE).stars(5).build()
```

Or via public constructor:
```
new TestQuery(Episode.NEWHOPE, 5)
```

@kassim


sav007:
sure with next PR

PR #339, opened 2017-03-21
Add global configuration for cache control

### LINKED ISSUE #335: Set global default cache strategy on ApolloClient
It would be convenient to set the default cache strategy while constructing the apollo instance.
Something like:

```

 return ApolloClient.builder()
        .serverUrl(graphQLBaseUrl)
        **.defaultCacheStrategy(NETWORK_FIRST)**
        .okHttpClient(okHttpClient)
        .build();

```

sav007:
Closes #335


apollo-runtime/src/test/resources/productsWithDate.json
```diff
@@ -1,76 +0,0 @@
-{
```
BenSchwab: Unused?

sav007: yeah, left from previous PR that supposed to remove them

PR #340, opened 2017-03-21
Fix singularize for list only

### LINKED ISSUE #318: response class names are made singular when they shouldn't be
`query {events{...}}` will produce a class name `Event` for example, which we wouldn't want, because "events" is the connection name, not a list of nodes

sav007:
Closes #318 

@kassim


digitalbuddha:
lgtm

PR #341, opened 2017-03-22
RxJava1 bindings

digitalbuddha:
I wanted to gather thoughts on how the Rx stuff should work. Here's an attempt at doing wrappers for `ApolloWatcher` &`ApolloCall`. Similar to Retrofit we will cancel a request when the `Observable` is unsubscribed from.


BenSchwab:
This looks like a good approach to me 👍

apollo-runtime/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
   compile dep.supportAnnotations
   compile dep.okHttp
   compile dep.moshi
+  compile dep.rxjava
```
sav007: shouldn't it be in separate module?

apollo-runtime/src/main/java/com/apollographql/android/impl/RxApollo.java
```diff
@@ -0,0 +1,74 @@
+package com.apollographql.android.impl;
```
sav007: I guess this class must be in :apollo-rxsupport

apollo-runtime/src/main/java/com/apollographql/android/impl/RxApollo.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+          }
+        });
+
+      }
+    });
+  }
+
+
+  public static <T> Observable<T> from(final ApolloCall<T> call) {
```
sav007: Single?

apollo-runtime/build.gradle
digitalbuddha: yup! will move to module next, wanted to get feedback on api being ok

apollo-runtime/src/main/java/com/apollographql/android/impl/RxApollo.java
```diff
@@ -0,0 +1,74 @@
...
+          }
+        } catch (IOException e) {
+          Exceptions.throwIfFatal(e);
+          if (!subscriber.isUnsubscribed()) {
+            subscriber.onError(e);
+          }
+          return;
+        }
+        if (!subscriber.isUnsubscribed()) {
+          subscriber.onCompleted();
```
sav007: is onError not enough? I thought you must call `onError` or `onCompleted` but not both

digitalbuddha: I copied from retrofit, its returning from catch block so it will on complete if no exception

apollo-runtime/src/main/java/com/apollographql/android/impl/RxApollo.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+          }
+        });
+
+      }
+    });
+  }
+
+  public static <T> Single<T> from(final ApolloCall<T> call) {
+    return Single.create(new Single.OnSubscribe<T>() {
```
sav007: if you change it to `Single.fromCallable` then you don't need try/catch

sav007: Oh you need access to subscriber

apollo-runtime/src/main/java/com/apollographql/android/impl/RxApollo.java
```diff
@@ -0,0 +1,79 @@
...
+
+import rx.Observable;
+import rx.Single;
+import rx.SingleSubscriber;
+import rx.Subscriber;
+import rx.exceptions.Exceptions;
+import rx.functions.Action0;
+import rx.subscriptions.Subscriptions;
+
+public class RxApollo {
```
sav007: final, with private constructor?

digitalbuddha:
so fyi rx-support module needs to be android library since it needs `ApolloClient` which lives in android library runtime

digitalbuddha:
closes #243

apollo-runtime/build.gradle
sav007: should be removed

apollo-rxsupport/build.gradle
```diff
@@ -0,0 +1,61 @@
+buildscript {
+  dependencies {
+    classpath dep.androidPlugin
+  }
+}
+
+apply plugin: 'com.android.library'
```
sav007: you think it should be android lib?

apollo-rxsupport/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=rx
+POM_NAME=ApolloStack Runtime
```
sav007: RxSupport

apollo-rxsupport/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=rx
```
sav007: rxsupport?

apollo-rxsupport/build.gradle
digitalbuddha: it depends on Runtime which is currently an android lib.  If/when we move sql stuff out to its own module we can change this to java

apollo-runtime/src/main/java/com/apollographql/android/ApolloCall.java
```diff
@@ -10,7 +10,8 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-public interface ApolloCall<T> {
+
```
sav007: nit: remove line

apollo-rxsupport/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+}
+
+dependencies {
+  compile dep.jsr305
+  compile dep.supportAnnotations
+  compile dep.okHttp
+  compile dep.moshi
+  compile dep.rxjava
+  compile project(":apollo-runtime")
+  compile project(":apollo-api")
```
marwanad: both of those should be provided because rx-support should assume those would be available at runtime


(Also, `compile project(":apollo-runtime")` doesn't bring the transitive deps with it. That's why when having `compile project(":apollo-api")`  under runtime, still required you to bring that in.

apollo-integration/build.gradle
```diff
@@ -26,8 +26,9 @@ android {
 }
 
 dependencies {
-  provided project(':apollo-api')
+  compile project(':apollo-api')
```
marwanad: can be `provided` because otherwise it would cause dex errors when the plugin is added.

apollo-rxsupport/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+  defaultConfig {
+    minSdkVersion 11
+    testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
+  }
+}
+
+dependencies {
+  compile dep.jsr305
+  compile dep.supportAnnotations
+  compile dep.okHttp
```
marwanad: ditto all those apart from Rx should be provided.

marwanad: actually, I don't think you'd need to include any of those. apart from Rx.

Issue #265, opened 2017-03-02
Implement serialization standards for Record


BenSchwab:

BenSchwab:
duplicate of https://github.com/apollographql/apollo-android/issues/286

digitalbuddha:
I'm actually going to reopen and do this as prework for #286

digitalbuddha:
Thx! I'm getting there. Feels like I'm coding for a whiteboard interview ☺️

BenSchwab:
Had a feeling that this might be a tricky, hence the very descriptive "no description provided" 😛

PR #311, opened 2017-03-10
Add runtime dependency through plugin

marwanad:
adds the `apollo-runtime` dependency through the plugin (in a similar way to apollo-api)

Closes #132.

PR #342, opened 2017-03-22
Add record field serialization

BenSchwab:
While doing work on LRU cache I realized it would be nice to have the record serializer for record size estimation. So took a crack at closing https://github.com/apollographql/apollo-android/issues/265.

Our current network parsing set-up follows this pattern `BufferedSource` -> `Map<String, Object>` -> `T Data`.  Where `BufferedSource` represents a json, and the transformation to `Map<String, Object>`. This transformation is performed by `ResponseJsonStreamReader`. 

When reading from the cache, we want to go from `Cache` -> `Map<String, Object>` where this map must be in _nearly_ the same form as the network response. The difference is that this map replaces objects with `CacheReference`s.

It seemed bad to duplicate all the code in `ResponseJsonStreamReader` so I opened `ResponseJsonStreamReader` for extension, and write a very light `CacheJsonStreamReader` which strongly types `CacheReferences` instead of parsing them as strings. I don't love relying on inheritance (although it is very lightweight) so @sav007 if you have any other ideas of how to share the code here, please let me know.

Currently, when reading from the SQL store we get a string. This might be bad for the memory concerns that @digitalbuddha had. So I added a `fromBufferedSource` to the `FieldAdapter` (which we use under the hood in `fromJson`.) But we might need to expand the SQLStore to having some sort of streaming byte array for reading records.


apollo-runtime/build.gradle
```diff
@@ -36,7 +36,7 @@ dependencies {
   compile dep.supportAnnotations
   compile dep.okHttp
   compile dep.moshi
-  provided project(":apollo-api")
+  compile project(":apollo-api")
```
BenSchwab: I had to change this to compile for the test to work (`SQLStoreTest` now needs access to `Optional` in `:apollo-api` @marwanad any reason this needed to be `provided`?

BenSchwab: Changed it back to provided, but added compile to the test variants.

marwanad: iirc, at that point, the `apollo-api` dependency wasn't really needed at all and we weren't' decided on how runtime would look like yet, we only had an `Apollo` class that wraps `Retrofit` and a `Utils` class. At that time, we also changed the plugin to add `apollo-api` deps to the compile dependency set.

We're gonna have to change it to `compile` anyways (#311). I guess for now we can keep it this way and I'll revisit this when I re-open #311.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
```diff
@@ -5,12 +5,47 @@
 import java.util.Map;
 import java.util.Set;
 
-//Todo: Enhance this class to better support generalized serialization
-// https://github.com/apollographql/apollo-android/issues/265
 public final class Record {
 
   private final String key;
   private final Map<String, Object> fields;
+  private int sizeEstimateBytes = 0;
```
sav007: can we remove `memoizedSize` and use instead `sizeEstimateBytes = -1`

sav007: And I can't find the place where we set value to this field?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
```diff
@@ -51,4 +83,17 @@ public String key() {
   public Map<String, Object> fields() {
     return fields;
   }
+
+  public int sizeEstimateBytes() {
```
sav007: Who and when this will be called?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
```diff
@@ -51,4 +83,17 @@ public String key() {
...
+    if (memoizedSize) {
+      return sizeEstimateBytes;
+    }
+    return calculateSizeEstimate(FieldsAdapter.create());
+  }
+
+  private int calculateSizeEstimate(FieldsAdapter adapter) {
```
sav007: Can we inject this calculation to the place where we create Record, for instance:
in com.apollographql.android.cache.normalized.sql.SqlStore#cursorToRecord

there we have already `String jsonOfFields = cursor.getString(2);` so can we take size of `jsonOfFields` as size for record?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
BenSchwab: This will be used by a bounded memory LRU cache, (and possibly the sql cache -- though SQL might have a better way to track size). In the memory LRU cache, everytime a record is changed (via a merge) we would have to recalculate a size estimate. 

I don't know if I love using the `toJSON` to determine the size, as that isn't very performant, which will be a requirment for the memory cache.

 I'm going to remove the size stuff from this diff, as the serialization for `Record` is valuable in it's own right.

I think for the Memory cache we need very performant

BenSchwab:
Decided to remove the record sizing stuff from this PR. I don't think calculating the size of the json string is the "right" way to do this, especially for the primary use case which is the memory lru cache.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/CacheJsonStreamReader.java
```diff
@@ -0,0 +1,28 @@
...
+
+  CacheJsonStreamReader(JsonReader jsonReader) {
+    super(jsonReader);
+  }
+
+  @Override protected Object readScalar(ResponseJsonStreamReader streamReader) throws IOException {
+    Object scalar = super.readScalar(streamReader);
+    if (scalar instanceof String) {
+      String scalarString = (String) scalar;
+      if (CacheReference.isSerializedReference((String) scalar)) {
```
sav007: nit: `CacheReference.isSerializedReference(scalarString)`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/CacheReference.java
```diff
@@ -31,5 +34,24 @@ public String key() {
...
+  public static CacheReference deserialize(String serializedCacheReference) {
+    Matcher matcher = SERIALIZATION_REGEX_PATTERN.matcher(serializedCacheReference);
+    if (!matcher.find() || matcher.groupCount() != 1) {
+      throw new IllegalArgumentException("Not a cache reference: " + serializedCacheReference
+          + " Must be of the form:" + SERIALIZATION_TEMPLATE);
+    }
+    return new CacheReference(matcher.group(1));
+  }
+
+  public static boolean isSerializedReference(String value) {
```
sav007: nit: should it be called `canDeserialize` ?

apollo-runtime/src/androidTest/java/com/apollographql/android/cache/normalized/sql/SqlStoreTest.java
```diff
@@ -16,15 +17,15 @@
...
+  public static final String FIELDS = "{\"fieldKey\": \"value\"}";
   public static final String IN_MEMORY_DB = null; //null means db is memory only
   private SqlStore sqlStore;
 
   @Before
   public void setUp() {
     ApolloSqlHelper apolloSqlHelper = ApolloSqlHelper.create(InstrumentationRegistry.getTargetContext(),
         IN_MEMORY_DB);
-    sqlStore = SqlStore.create(apolloSqlHelper, FieldsAdapter.create(new Moshi.Builder().build()));
+    sqlStore = SqlStore.create(apolloSqlHelper, FieldsAdapter.create());
```
digitalbuddha: any reason for this change? I'd prefer to be able to use 1 moshi instance for whole app apollo or otherwise

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/FieldsAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public Map<String, Object> fromBufferedSource(BufferedSource bufferedSource) throws IOException {
+    CacheJsonStreamReader jsonStreamReader
+        = new CacheJsonStreamReader(new BufferedSourceJsonReader(bufferedSource));
+    return jsonStreamReader.buffer();
+  }
+
+  public Map<String, Object> fromJson(String fields) throws IOException {
+    final BufferedSource bufferSource = Okio.buffer(Okio.source(new ByteArrayInputStream(fields.getBytes())));
+    return fromBufferedSource(bufferSource);
```
digitalbuddha: nitpick, it will read better as from(bufferedSource)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/FieldsAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+    return new FieldsAdapter(moshi);
+  }
+
+  public String toJson(Map<String, Object> fields) {
+    return serializationAdapter.toJson(fields);
+  }
+
+  public Map<String, Object> fromBufferedSource(BufferedSource bufferedSource) throws IOException {
+    CacheJsonStreamReader jsonStreamReader
+        = new CacheJsonStreamReader(new BufferedSourceJsonReader(bufferedSource));
```
digitalbuddha: Personally I like APIs with static calls for object creation like 
`ApolloReader.cacheJsonStream(ApolloReader. bufferedSourceJson(bufferedSource)`
This way you can do static imports and become `cacheJsonStream(bufferedSourceToJson(source))`

apollo-runtime/src/androidTest/java/com/apollographql/android/cache/normalized/sql/SqlStoreTest.java
BenSchwab: I might be missing something of how Moshi works, but it seems dangerous to register our BigDecimal converter on a global Moshi instance, in case there is an existing converter. (Currently, it seems we do not allow a user to pass in a Moshi instance to Apollo anyway).

We could register these convertors on the Apollo's Moshi instance in ApolloClient. But that Moshi instance seems primarily concerned with serializing variable objects, and initializing cache serializers there would take some thought of how to do it properly without spaghetti dependencies in the ApolloClient builder.

digitalbuddha: My thinking was from a performance standup. Moshi caches adapters but it's on a per instance level. So say I have same date converter on app and apollo client, I'd prefer to not have to cache both. Currently our app works with 1 instance of gson. It doesn't have to be solved now. I can do a pr  if it becomes an issue

PR #343, opened 2017-03-23
Refactor RealApolloCall to utilize application interceptors

### LINKED ISSUE #314: Refactor RealApolloCall and introduce application interceptors
`RealApolloCall` class becomes unmaintainable and clumsy.
Per our discussion we want to break it into independent pieces and follow chain of responsibility pattern, the way how OkHttp does with http interceptors.

For now we are going to have 2 interceptors: normalized cache and http call.

sav007:
Closes #314 

Introduces 3 application call interceptors: 
- `CacheCallInterceptor` responsible for normalized caching
- `ParseCallInterceptor` responsible for parsing response
- `ServerCallInterceptor` responsible for making server http request


apollo-runtime/src/main/java/com/apollographql/android/impl/CacheCallInterceptor.java
```diff
@@ -0,0 +1,176 @@
...
+      InterceptorResponse response = resolveNetworkFirstCacheResponse(operation, networkResponse);
+      if (response != null) {
+        return response;
+      }
+      throw e;
+    }
+  }
+
+  @Override
+  public void interceptAsync(final Operation operation, final CallInterceptorChain chain,
```
digitalbuddha: looking at OKHTTP they dont seem to need 2 paths but instead make an `AsycCall` when someone calls enqueue which makes the binding between async and sync world https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RealCall.java#L106

sav007: The issue that we switching between threads, we start from our dispatcher that we are using for our cache and then jump to OkHttp dispatcher. Initially I thought about to make the same as in OkHttp but then it means we have to copy and modify `okhttp3.Dispatcher` that I don't want to go with

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
```diff
@@ -10,79 +10,64 @@
...
     this.httpCache = httpCache;
     this.httpCacheControl = httpCacheControl;
+    this.moshi = moshi;
     this.responseFieldMapper = responseFieldMapper;
     this.customTypeAdapters = customTypeAdapters;
     this.cache = cache;
     this.cacheControl = cacheControl;
     this.dispatcher = dispatcher;
+
+    interceptorChain = new RealCallInterceptorChain(operation)
```
sav007: How about this to change API from `com.apollographql.android.impl.RealCallInterceptorChain#chain` to `addIntercetor`

sav007:
@digitalbuddha regarding
>Just some style design feedback for discussion, I more question if I as a user want to add another interceptor, how would that work?

I intentionally closed interceptors for now, as we need to think do we really want to open them, for what purpose? And if do then we need to change API of ApolloClient as well

digitalbuddha:
1 interceptor I've added myself to okhttp instance  is a logginginterceptor

apollo-runtime/src/main/java/com/apollographql/android/impl/RealApolloCall.java
digitalbuddha: Sounds good!

apollo-runtime/src/main/java/com/apollographql/android/impl/CacheCallInterceptor.java
digitalbuddha: Makes sense. I always wondered why our internal code base had the dispatcher copied out, now I know why ☺️

sav007:
https://github.com/apollographql/apollo-android/issues/346 has been created

apollo-runtime/src/main/java/com/apollographql/android/impl/CallInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+    void onResponse(InterceptorResponse response);
+
+    void onFailure(Throwable t);
+  }
+
+  final class InterceptorResponse {
+    final okhttp3.Response httpResponse;
+    final Response parsedResponse;
+    final Collection<Record> cacheRecords;
+
```
BenSchwab: Should we make these wrapped in `Optional`?

apollo-runtime/src/main/java/com/apollographql/android/impl/RealCallInterceptorChain.java
```diff
@@ -0,0 +1,103 @@
...
+    });
+  }
+
+  private RealCallInterceptorChain(Operation operation, CallInterceptor currentInterceptor) {
+    this.operation = operation;
+    this.currentInterceptor = currentInterceptor;
+  }
+
+  @Override public CallInterceptorChain chain(CallInterceptor callInterceptor) {
+    RealCallInterceptorChain lasInterceptorChain = this;
```
BenSchwab: is `lasInterecprotChain` a typo?

apollo-runtime/src/main/java/com/apollographql/android/impl/ParseResponse.java
```diff
@@ -0,0 +1,16 @@
+package com.apollographql.android.impl;
+
+import com.apollographql.android.api.graphql.Response;
+import com.apollographql.android.cache.normalized.Record;
+
+import java.util.Collection;
+
+final class ParseResponse<T> {
```
BenSchwab: Unused?

Issue #324, opened 2017-03-15
Add LRU Normalized Cache implementation


BenSchwab:
We should ship a version of an in memory normalized cache. This could be used instead of SQL, or we could even ship a "double-layer" cache that would read from memory if possible.

Android has a LRU cache: https://developer.android.com/reference/android/util/LruCache.html
Guava has memory caches: https://github.com/google/guava/wiki/CachesExplained

The Guava stuff looks really cool - but I'm not sure we need all of it's goodies. But not having to depend on an Android library is always a plus, even if we already have other Android dependencies.

For time based expiration to be useful, we would have to allow a user to pass in a "TTL" type parameter. This is something we will probably want for the SQL store. https://github.com/apollographql/apollo-android/issues/326

For size based expiration, we have the ability to accurately approximate cache size, because we are storing records whose fields are ultimately either a primitive or a reference.

@digitalbuddha I believe you recommended Guava. Would the right approach be to shade the Gauva cache / bring its files?

digitalbuddha:
Already one step ahead of you. Here's nytimes guava cache lite. I painstakingly built it up by moving one file at a time https://github.com/NYTimes/Store/tree/develop/cache currently sits at about 400 methods.

BenSchwab:
Great! 
So we can just add the dependency like `compile com.nytimes.android:cache:CurrentVersion`?

@sav007 @marwanad @brianPlummer Any objections to this dependency?

BenSchwab:
There might actually be some value in pulling out all the cache implementations we ship with to a `apollo-cache` project. That way if someone doesn't want to use any caches, they won't get any of these dependencies.

digitalbuddha:
Yup thats the dependency. Since Guava at v 19 was locked in as last one for Android until 21 we don't expect much changes to pull in from Guava but watch it carefully.    As for the splitting project into multiple artifacts, I can do that once we are at a good point prior to release. Same thing was done with RxAndroid where it started as monolith and then was split. As long as we do separate packages for caches it should make refactoring easier later.

digitalbuddha:
@BenSchwab I don't think we need to pull out to a separate project. How do you feel about just keeping different modules and artifacts similar to how the rx module is going to work.  There is niceties in being 1 repo/project when it comes to refactoring or searching.

BenSchwab:
Yep, totally agree

PR #348, opened 2017-03-24
Add bounded memory LRU cache

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/324

- Adds dependency on https://github.com/NYTimes/Store/tree/develop/cache
- Uses size estimation that Picasso uses for default cache size https://github.com/square/picasso/blob/master/picasso/src/main/java/com/squareup/picasso/Utils.java#L242
- Dynamically tracks record size estimation by calculating a byte size for all record field types. This feels a little sketchy, as how the JVM actually sizes an object depends on a lot of factors. But, it roughly seems to be accurate. However, open to better ideas :)
- Size estimation is calculate on need-to-know basis. Currently this `LRUCache` is the only thing that needs to use the record size, so if we are not using the the `LRUCache` we won't waste time calculating record size


apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.annotation.Nullable;
+
+import static android.content.Context.ACTIVITY_SERVICE;
+import static android.content.pm.ApplicationInfo.FLAG_LARGE_HEAP;
+
+public final class LruCacheStore extends CacheStore {
+
+  private final Cache<String, Record> lruCache;
+
+  public static LruCacheStore create(Context context) {
```
digitalbuddha: any reason not to have a time expiration configuration as well?  So for us we want articles to expire every day so that someone doesn't see old news. :-)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public static LruCacheStore createUnbounded() {
+    return new LruCacheStore(Optional.<Long>absent(), Optional.<Long>absent());
+  }
+
+  /**
+   * Derived from https://github.com/square/picasso cache sizing.
+   */
+  private static long calculateMemoryCacheSize(Context context) {
+    ActivityManager activityManager = (ActivityManager) context.getSystemService(ACTIVITY_SERVICE);
```
digitalbuddha: is there a way to do this without dependency on android?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+    boolean largeHeap = (context.getApplicationInfo().flags & FLAG_LARGE_HEAP) != 0;
+    int memoryClass = largeHeap ? activityManager.getLargeMemoryClass() : activityManager.getMemoryClass();
+    // Target ~15% of the available heap.
+    return (int) (1024L * 1024L * memoryClass / 7);
+  }
+
+  private LruCacheStore(Optional<Long> maximumWeightInBytes, Optional<Long> maximumEntries) {
+    final CacheBuilder<Object, Object> lruCacheBuilder = CacheBuilder.<String, Record>newBuilder();
```
digitalbuddha: `CacheBuilder<Object, Object>`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
BenSchwab: Yeah, I'll try to implement something similar to https://github.com/NYTimes/Store/blob/develop/store/src/main/java/com/nytimes/android/external/store/base/impl/MemoryPolicy.java

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
BenSchwab: I don't think so -- but we depend on android for the SQL store, so this could be moved to the "Android cache" module, and leave the rest of the LRU stuff in a "core cache" module. Though, we are starting to get tons of modules :)

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -0,0 +1,87 @@
...
+   */
+  private static long calculateMemoryCacheSize(Context context) {
+    ActivityManager activityManager = (ActivityManager) context.getSystemService(ACTIVITY_SERVICE);
+    boolean largeHeap = (context.getApplicationInfo().flags & FLAG_LARGE_HEAP) != 0;
+    int memoryClass = largeHeap ? activityManager.getLargeMemoryClass() : activityManager.getMemoryClass();
+    // Target ~15% of the available heap.
+    return (int) (1024L * 1024L * memoryClass / 7);
+  }
+
+  private LruCacheStore(Optional<Long> maximumWeightInBytes, Optional<Long> maximumEntries) {
```
sav007: Should we call `maximumWeightInBytes ` just `maxSizeInBytes`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/RecordWeigher.java
```diff
@@ -0,0 +1,47 @@
+package com.apollographql.android.cache.normalized;
+
+import java.math.BigDecimal;
+import java.util.List;
+import java.util.Map;
+
+public final class RecordWeigher {
```
sav007: private constructor

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
```diff
@@ -81,4 +84,17 @@ public String key() {
     return fields;
   }
 
+  public int sizeEstimateBytes() {
```
sav007: just to make sure is it thread safe?

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
BenSchwab: Oddly, the correct code is ` final CacheBuilder<Object, Object> lruCacheBuilder = CacheBuilder.newBuilder();` The <String, Record> I put is ignored. 

Seems like the CacheBuilder keeps things untyped, until you add something typed, such as a weigher.

Then when you build, so long as your target type is a subtype of the weigher type it allows it.

digitalbuddha: TIL!

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
BenSchwab: I'll make `sizeInBytes` volatile

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/EvictionPolicy.java
```diff
@@ -0,0 +1,117 @@
...
+import android.content.Context;
+
+import com.apollographql.android.api.graphql.internal.Optional;
+
+import java.util.concurrent.TimeUnit;
+
+import static android.content.Context.ACTIVITY_SERVICE;
+import static android.content.pm.ApplicationInfo.FLAG_LARGE_HEAP;
+
+public final class EvictionPolicy {
```
sav007: Nice, we should use it for httpcache as well

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/Record.java
sav007: Should it be called `sizeInBytes ()`

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/EvictionPolicy.java
```diff
@@ -0,0 +1,117 @@
+package com.apollographql.android.cache.normalized.lru;
+
+import android.app.ActivityManager;
+import android.content.Context;
```
sav007: Only one concern if we want to keep runtime pure java, we shouldn't have android deps

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/EvictionPolicy.java
```diff
@@ -0,0 +1,117 @@
...
+      return new EvictionPolicy(maxSizeBytes, maxEntries, expireAfterAccess, expireAfterAccessTimeUnit,
+          expireAfterWrite, expireAfterWriteTimeUnit);
+    }
+
+  }
+
+  /**
+   * Derived from https://github.com/square/picasso cache sizing.
+   */
+  private static long calculateDeviceMemoryCacheSize(Context context) {
```
digitalbuddha: Would love to find a way to calculate this without context

marwanad: Not entirely sure but it seems that those values are just dalvik VM [properties](https://android.googlesource.com/platform/frameworks/base/+/742a67127366c376fdf188ff99ba30b27d3bf90c/core/java/android/app/ActivityManager.java#181). Are those accessible through a `System.getProperty()`?

The closest thing we can get to without context is `Runtime.getRuntime().maxMemory` which gives the number of bytes in heap instead of the recommended usage so maybe use that and reduce the target even more?

digitalbuddha:
We could also abstract and have user pass the value in with a static method
that can give default. Have the static method live within the android
module.

BenSchwab:
I'll remove the autosizing option, and create an issue to add it back when we have an android module.

Also, just stuck in `DualCache` on this diff. It was just a few lines of code.

digitalbuddha:
Cool! Quick glance dual cache looks nice. I'm curious if we can use the get(key, callable) interface of the cache to get the disk to just be a fallback of the memory get.

BenSchwab:
Yeah, I was thinking about using the `get(key, callable)`. But this approach seems more generic. You could, in theory, want a dual cache of two sql databases ¯\_(ツ)_/¯

digitalbuddha:
The only reason I'd suggest using the get callable is that it throttles multiple calls. So say 4 consumers all hit the cache from different threads, first one will get through and then share response with the other 3.  More of a nice to have then a need to have ☺️

BenSchwab:
That's a compelling reason. Additionally, it will handle loading into memory on a primary cache miss, which the `DualCache` I have up does not.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/EvictionPolicy.java
```diff
@@ -0,0 +1,97 @@
...
+  public static final EvictionPolicy NO_EVICTION = EvictionPolicy.builder().build();
+
+  private final Optional<Long> maxSizeBytes;
+  private final Optional<Long> maxEntries;
+  private final Optional<Long> expireAfterAccess;
+  private final Optional<TimeUnit> expireAfterAccessTimeUnit;
+  private final Optional<Long> expireAfterWrite;
+  private final Optional<TimeUnit> expireAfterWriteTimeUnit;
+
+  public Optional<Long> maxSizeBytes() {
```
marwanad: Should this return the value instead of `Optional`? I assume that would be part of the public runtime API.

BenSchwab: Good point, I think this is actually just for internal use, after building the policy and passing to the cache, you shouldn't need to access the settings. will change to package private.

PR #349, opened 2017-03-24
fix integration tests for list types

marwanad:
Fixes the tests after the changes in #340.

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/android/impl/IntegrationTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/android/impl/IntegrationTest.java
@@ -108,8 +108,8 @@ public class IntegrationTest {
     assertThat(data.allPlanets().planets().size()).isEqualTo(60);
 
     List<String> planets = FluentIterable.from(data.allPlanets().planets())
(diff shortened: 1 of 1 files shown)


apollo-integration/src/test/java/com/apollographql/android/impl/IntegrationTest.java
```diff
@@ -153,12 +153,9 @@
     assertThat(data.allFilms().films()).hasSize(6);
 
     List<String> dates = FluentIterable.from(data.allFilms().films())
-        .transform(new Function<AllFilms.Data.AllFilm.Film, String>() {
-          @Override public String apply(AllFilms.Data.AllFilm.Film film) {
+        .transform(new Function<AllFilms.Data.AllFilms1.Film, String>() {
+          @Override public String apply(AllFilms.Data.AllFilms1.Film film) {
             Date releaseDate = film.releaseDate();
-            if (releaseDate == null) {
```
marwanad: releaseDate is non-null so the check is redundant

sav007:
Thx, man I was sure that I fixed them, but I guess missed to commit

PR #350, opened 2017-03-24
update sample module to reflect plugin changes

marwanad:
Fixes some stuff after #311 and changed where the system property is set. For the `apollo-integration` module, the system property is set in the build script.

With this and #311, the plugin now adds `apollo-runtime` (which includes api in its dependency list).

(The old behaviour is still preserved, if the user provides their own `compile dep.apollo-runtime` this would be used instead, if it matches the plugin version number.)

### DIFF
--- a/apollo-integration/build.gradle
+++ b/apollo-integration/build.gradle
@@ -4,6 +4,7 @@ buildscript {
     classpath dep.apolloPlugin
   }
 }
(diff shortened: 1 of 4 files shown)


marwanad:
@sav007 feel free to merge to get your PR through

digitalbuddha:
Lgtm

PR #352, opened 2017-03-25
Plugin config option to generate accessor methods 

### LINKED ISSUE #332: Add support for `public final` model fields instead of accessor methods
As we have seen with some sample queries, a innocent looking `gql` can generate a lot of methods.

We should consider either allowing a user to specify `apollo { publicFinalFields = true }`, or just move to them altogether.

The main argument for accessor/getter functions is that it allows for the underlying data model to change it's implementation without needed downstream changes. However, afaik, the current graphql spec doesn't require any complicated getter logic.

This should also help with kotlin interop as getters without `get___` will not autoconvert to properties:  https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters

sav007:
Closes #332 

Adds gradle configuration option to generate accessor methods.

By default it set to `true`.
https://github.com/apollographql/apollo-android/pull/352/files#diff-b04637123f448c4d6ce43973170b8ab0

Fixes couple issues with missing `Nonnull` annotations
https://github.com/apollographql/apollo-android/pull/352/files#diff-0a57476f4fa471ab6621dbf28e9ff5cd

Makes `List` in models be immutable
https://github.com/apollographql/apollo-android/pull/352/files#diff-537e62e5f38dfdafae746f4f0cd114a1

Examples:
https://github.com/apollographql/apollo-android/pull/352/files#diff-9d649787d30729102b2e44b96102b22b
https://github.com/apollographql/apollo-android/pull/352/files#diff-bc1759fd8b2ecbc12c0c279ca4b91843

### DIFF
--- a/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloClassGenTask.java
+++ b/apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloClassGenTask.java
@@ -81,7 +81,7 @@ public class ApolloClassGenTask extends SourceTask {
     this.useOptional = useOptional;
   }
 
-  public boolean isGenerateAccessors() {
+  public boolean shouldGenerateAccessors() {
     return generateAccessors;
   }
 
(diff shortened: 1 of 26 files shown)


apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloClassGenTask.java
```diff
@@ -77,4 +80,12 @@ public boolean isUseOptional() {
   public void setUseOptional(boolean useOptional) {
     this.useOptional = useOptional;
   }
+
+  public boolean isGenerateAccessors() {
+    return generateAccessors;
```
digitalbuddha: Nitpick shouldGenerateAccessors?

PR #353, opened 2017-03-26
Small fixes

sav007:
- Add Jsr305 dep back to runtime
- Fix semantics of converting async calls to Rx

Couple sources for RxJav suggest avoid of using Single/Observer.create

### DIFF
--- a/apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
+++ b/apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
@@ -29,6 +29,12 @@ public final class RxApollo {
 
   @Nonnull
   public static <T> Observable<T> from(@Nonnull final ApolloWatcher<T> watcher) {
+    return from(watcher, Emitter.BackpressureMode.LATEST);
+  }
+
+  @Nonnull public static <T> Observable<T> from(@Nonnull final ApolloWatcher<T> watcher,
+      @Nonnull Emitter.BackpressureMode backpressureMode) {
+    checkNotNull(backpressureMode, "backpressureMode == null");
     checkNotNull(watcher, "watcher == null");
     return Observable.fromEmitter(new Action1<Emitter<T>>() {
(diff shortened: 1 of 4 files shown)


apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
```diff
@@ -26,65 +25,47 @@ private RxApollo() {
...
-            if (!subscriber.isUnsubscribed()) {
-              subscriber.onError(e);
-            }
+            emitter.onError(e);
           }
         });
-
       }
-    });
+    }, Emitter.BackpressureMode.LATEST);
```
marwanad: Should this be taken as a parameter? 

Would there be cases where the user would want to buffer unconsumed `onNext` calls instead?

digitalbuddha:
Emitted is a Deprecated api that disappeared in rx2 you sure you want to use it? I used the same logic that retrofit does with their call adapters fyi.

sav007:
@digitalbuddha 

from RxJava2 stand point of view the watcher must be wrapped in `Flowable.create(FlowableEmitter, BackpressureStrategy)`

As for the single yeah we can use create, as we don;t need backpressure strategy.

apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
sav007: Good point

digitalbuddha:
This also seems to not have the if subscribed logic prior to emitting. Is that not necessary with emitter api?

sav007:
Yes you don't need this logic, it suppose be handled internally by emitter:

see http://reactivex.io/RxJava/javadoc/rx/Observable.html fromEmitter.

Btw should we upgrade to RxJava 1.3?

digitalbuddha:
Observable.fromEmitter does not exist in 1.3

PR #354, opened 2017-03-27
Use Observable.fromEmitter instead of Observable.create

mandrizzle:
The `Observable.create(OnSubscribe)` method that is being used to create an `Observable` does not handle backpressure properly. As a matter of fact, since RxJava 1.2.7, the `Observable.create` method that is being used by `RxApollo` is deprecated and is now called `unsafeCreate` for this very reason. See [here](https://github.com/ReactiveX/RxJava/releases/tag/v1.2.7) 

To fix this without updating to the latest version of RxJava 1.x, it is recommended to use `fromEmitter`. This safely wraps async callbacks to return an `Observable` while handling backpressure for us.


sav007:
@mandrizzle sorry this is dup of https://github.com/apollographql/apollo-android/pull/353

PR #363, opened 2017-03-28
Handle missing record in SqlStore

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/364

SqlStore was throwing an sql exception when trying to load a missing record. Added check on the cursor. Also, fixed another bug using `Optional` when SqlStore returns `absent`.

I should have caught this in my last PR -- but this time tested on local fork in sandbox application, and the dual memcache sql cache is working

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
@@ -59,8 +59,7 @@ public final class LruCacheStore extends CacheStore {
             // get(key, callable) requires non-null. If null, an exception should be
             //thrown, which will be converted to null in the catch clause.
             if (record == null) {
-              throw new ExecutionException(
-                  new IOException(String.format("Record{id=%s} not present in secondary cache", key)));
+              throw new IOException(String.format("Record{id=%s} not present in secondary cache", key));
             }
             return record;
           }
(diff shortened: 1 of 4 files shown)


apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -54,7 +55,14 @@ public LruCacheStore(EvictionPolicy evictionPolicy, CacheStore secondaryCacheSto
       try {
         return lruCache.get(key, new Callable<Record>() {
           @Override public Record call() throws Exception {
-            return secondaryCacheStore.get().loadRecord(key);
+            Record record = secondaryCacheStore.get().loadRecord(key);
+            // get(key, callable) requires non-null. If null, an exception should be
+            //thrown, which will be converted to null in the catch clause.
+            if (record == null) {
+              throw new ExecutionException(
+                  new IOException(String.format("Record{id=%s} not present in secondary cache", key)));
```
sav007: Do we need to wrap IOException ? Maybe we need to make loadRecord to throw IOException?

BenSchwab: Hmm, yeah looks like the wrapping is unnecessary.

We could make `loadRecord` throw `IOException` if we feel that an `IOException` is better than returning `null` if the record is not present. I personally feel that `null` is fine. This would have to be on the `CacheStore` interface, so an exception seems a bit heavy handed to me.

apollo-runtime/src/main/java/com/apollographql/android/cache/normalized/lru/LruCacheStore.java
```diff
@@ -54,7 +55,13 @@ public LruCacheStore(EvictionPolicy evictionPolicy, CacheStore secondaryCacheSto
...
+            Record record = secondaryCacheStore.get().loadRecord(key);
+            // get(key, callable) requires non-null. If null, an exception should be
+            //thrown, which will be converted to null in the catch clause.
+            if (record == null) {
+              throw new IOException(String.format("Record{id=%s} not present in secondary cache", key));
+            }
+            return record;
           }
         });
       } catch (ExecutionException e) {
```
sav007: Just `Exception`?

PR #365, opened 2017-03-28
Response body empty when HttpException throw

### LINKED ISSUE #361: Response body empty when HttpException thrown
Our GraphQL endpoint requires OAuth authentication. When we use the apollo client to hit our endpoint and our endpoint returns a status code of `401`, the apollo client throws an `HttpException` with a `okhttp3.Response` object. The problem is, before the exception is thrown, the `Response` is closed. This means when we try to read the body of the `Response` when we try to handle the exception, it is blank.  

This is from `ApolloParseInterceptor#parse` which is where the premature closure of the response happens:
```java
  Util.closeQuietly(response);
  throw new HttpException(response);
```

A solution to this could be to not to add a `Response` to the exception. We can instead create an immutable object, say `NetworkResponse` that has a `responseBody` and `responseCode` that we can add to the exception.

sav007:
Closes #361 

@mandrizzle

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/android/impl/util/HttpException.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/impl/util/HttpException.java
@@ -7,7 +7,7 @@ import okhttp3.Response;
 public class HttpException extends RuntimeException {
   private final int code;
   private final String message;
-  private final Response rawResponse;
+  private final transient Response rawResponse;
 
   public HttpException(@Nullable okhttp3.Response rawResponse) {
     super(formatMessage(rawResponse));
@@ -24,7 +24,7 @@ public class HttpException extends RuntimeException {
     return message;
   }
 
-  public @Nullable Response rawResponse() {
+  @Nullable public Response rawResponse() {
     return rawResponse;
   }
 
--- a/apollo-runtime/src/main/java/com/apollographql/android/impl/util/HttpException.java
+++ b/apollo-runtime/src/main/java/com/apollographql/android/impl/util/HttpException.java
@@ -1,15 +1,37 @@
 package com.apollographql.android.impl.util;
 
+import javax.annotation.Nullable;
+
 import okhttp3.Response;
 
 public class HttpException extends RuntimeException {
-  private final okhttp3.Response rawResponse;
+  private final int code;
+  private final String message;
+  private final Response rawResponse;
 
-  public HttpException(okhttp3.Response rawResponse) {
+  public HttpException(@Nullable okhttp3.Response rawResponse) {
+    super(formatMessage(rawResponse));
+    this.code = rawResponse != null ? rawResponse.code() : 0;
+    this.message = rawResponse != null ? rawResponse.message() : "";
     this.rawResponse = rawResponse;
   }
 
-  public Response rawResponse() {
+  public int code() {
+    return code;
+  }
+
+  public String message() {
+    return message;
+  }
+
+  public @Nullable Response rawResponse() {
     return rawResponse;
+  }
+
+  private static String formatMessage(Response response) {
+    if (response == null) {
+      return "Empty HTTP response";
+    }
+    return "HTTP " + response.code() + " " + response.message();
   }
 }
--- a/apollo-runtime/src/test/java/com/apollographql/android/impl/InterceptorTest.java
+++ b/apollo-runtime/src/test/java/com/apollographql/android/impl/InterceptorTest.java
@@ -1,0 +1,65 @@
+package com.apollographql.android.impl;
+
+import com.apollographql.android.api.graphql.Query;
+import com.apollographql.android.api.graphql.ResponseFieldMapper;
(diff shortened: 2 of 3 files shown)


apollo-runtime/src/test/java/com/apollographql/android/impl/InterceptorTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        @Override public Object wrapData(Data data) {
+          return data;
+        }
+      }).execute();
+      fail("Expected HttpException");
+    } catch (HttpException httpException) {
+      assertThat(httpException.code()).isEqualTo(401);
+      assertThat(httpException.message()).isEqualTo("Client Error");
+      assertThat(httpException.rawResponse().body().string()).isEqualTo("Unauthorized request!");
```
mandrizzle: Is is a good idea to not close the body of the response? I am not sure of the implications of not doing so. Personally, I would have changed the `HttpException` class to have a static method that looks like:

```java
  public static HttpException from(Response response) throws IOException {
     try {
       String body = response.body() != null ? response.body().string() : "";
       return new HttpException(response, response.code(), response.message(), body);
     } finally {
       closeQuietly(response);
     }
   }
```

So that the body does get closed but before it does, we have read the contents of it. Also IIRC, the `ResponseBody` in the `Response` object is what gets closed when `close` is called. So if we read and copy the contents over, it is probably safe to close.

apollo-runtime/src/main/java/com/apollographql/android/impl/util/HttpException.java
```diff
@@ -1,15 +1,37 @@
...
 
+import javax.annotation.Nullable;
+
 import okhttp3.Response;
 
 public class HttpException extends RuntimeException {
-  private final okhttp3.Response rawResponse;
+  private final int code;
+  private final String message;
+  private final Response rawResponse;
```
mandrizzle: Since `Response` isn't `Serializable`, this is a good time to add `transient` to this since technically `Exceptions` should be `Serializable`

apollo-runtime/src/test/java/com/apollographql/android/impl/InterceptorTest.java
mandrizzle: Just to add, calling `string()` does close `ResponseBody` but if no one was to call it, it would not close.

mandrizzle:
Just throwing out an idea here. What if we reserve `onFailure` for `IOExceptions`aka network issues but repurpose `graphql.Response` to also return server errors?

I mention this because as of right now, I think the public api for the `ApolloClient` is easy to screw up. 

1. `onFailure` passes in a `Throwable` which means one needs to be aware to cast what is passed in back to an `HttpException` to be able to read server errors. 

2.  After thinking about it, I think we shouldn't close the `ResponseBody` and instead, leave it for the consumer to close it. Because the consumer might want to read in the `ResponseBody` from the error via `BufferedSource` rather than as a `String`. Now the problem I see is, since the `ResponseBody` that we need to close is part of the `Exception` that is thrown, it can easily be ignored or lost which means people might see [this](https://github.com/square/okhttp/issues/2311)

3. There are other implications having non 200 status codes being returned as exceptions. One that I can think of is how a server error can kill a `rx` stream from the `RxApollo` package when really calling `onError` in streams should be for fatal, unrecoverable, exceptions. See [here](http://blog.danlew.net/2015/12/08/error-handling-in-rxjava/) for more details on that.

Thoughts on this? I realize it might require a bigger, breaking, change but may be worth it.

sav007:
@mandrizzle Thanks, these are good points:

>What if we reserve onFailure for IOExceptionsaka network issues but repurpose graphql.Response to also return server errors

>1 onFailure passes in a Throwable which means one needs to be aware to cast what is passed in back to an HttpException to be able to read server errors.

Except IOException we can potentially have: `JsonDataException`, `JsonEncodingException`.
I suggest to not include any exceptions in `graphql.Response` as this object represents the real response from server (with errors returned by GraphQL). But I suggest to wrap all Exceptions into some thing like checked `ApolloException`. So `OnFailure` will always get the `ApolloException` with cause == original one

>2 After thinking about it, I think we shouldn't close the ResponseBody and instead, leave it for the consumer to close it. Because the consumer might want to read in the ResponseBody from the error via BufferedSource rather than as a String. Now the problem I see is, since the ResponseBody that we need to close is part of the Exception that is thrown, it can easily be ignored or lost which means people might see this

That the tradeoff of giving user to parse error body. And this is the contract that user must follow always close `ApolloException`

>3 There are other implications having non 200 status codes being returned as exceptions. One that I can think of is how a server error can kill a rx stream from the RxApollo package when really calling onError in streams should be for fatal, unrecoverable, exceptions. See here for more details on that.

That's issue of Rx wrapper if user don't want to break the rx chain he must wrap the error (for example with onErrorResumeNext or smth). So this is not part of ApolloClient itself as it's a pure java module, but we can think about this in context of our RxWrapper

Issue #369, opened 2017-03-29
SQLStore -- Records not updated


BenSchwab:
We are calling `insert` to update records, which apparently does not actually update existing records.

Also, our insert statement is not cleaned for SQL injection.

PR #371, opened 2017-03-29
SQLStore - use compiled sql statements, use update for merge

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/369

### DIFF
--- a/apollo-runtime/src/androidTest/java/com/apollographql/android/cache/normalized/sql/SqlStoreTest.java
+++ b/apollo-runtime/src/androidTest/java/com/apollographql/android/cache/normalized/sql/SqlStoreTest.java
@@ -73,6 +73,17 @@ public class SqlStoreTest {
     assertThat(record).isNull();
   }
 
+  @Test
+  public void testRecordMerge() {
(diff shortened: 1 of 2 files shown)


digitalbuddha:
I'm terrible with sql stuff. Didn't you just remove the compile statement stuff?

BenSchwab:
Haha me too -- it's pretty arcane stuff. I changed the compile statement to use the `insert` / `update` methods outlined here: https://developer.android.com/training/basics/data-storage/databases.html.

I _think_ the advantage is that those methods will properly cleanse the arguments for sql injection attacks.

With the compile sql using a string, I think a user could construct a malicious record key that performs sql injection.

digitalbuddha:
Make sense thanks for explaining! I just assumed CompiledStatements == PreparedStatements.

BenSchwab:
I think there is a difference between what I changed this too, and compiled statements. And compile statements have performance benefits. I'll go ahead and land this so SQLStore is functional, but I'll continue thinking about the optimal structure.

And yeah sqlDelight might be nice :)

PR #375, opened 2017-03-30
Refactor api package

sav007:
This PR:
- rename `com.apollographql.android.api.graphql` to `com.apollographql.apollo.api`
- ~enable back plugin tests~
- ~enable back auto add dep to apollo-runtime module in gradle plugin~
- bump apollo version to 0.2.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jdk:
   - oraclejdk8
 
 script:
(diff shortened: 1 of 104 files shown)


marwanad:
just noticed something, should the `groupID` under gradle.properties should be `com.apollographql.apollo` instead?

sav007:
yeah will do in next PR

PR #379, opened 2017-03-31
Introduce ApolloExceptions

### LINKED ISSUE #367: Consider wrapping any exception with ApolloException
The idea is to introduce new checked exception `ApolloException` and wrap any exception thrown at runtime with this class.

Change the signature of `com.apollographql.android.ApolloCall#execute` and `com.apollographql.android.ApolloCall.Callback#onFailure` to:
```
@Nonnull Response<T> execute() throws ApolloException;

and 

void onFailure(@Nonnull ApolloException t);
```

Introduce 3 codes/types of the `ApolloException`: 
- `ApolloException.PARSE` represents parsing response error (like malformed response, etc.)
- `ApolloException.HTTP` represents any network errors (like statusCode != 200, timeout etc.)
- `ApolloException.RUNTIME` any other errors

sav007:
Closes #367 

Introduces 4 ApolloExceptions:
- `ApolloException` generic any runtime errors should be wrapped into it
- `ApolloHttpException` if http status code != 200
- `ApolloNetworkException` any network, timeouts etc.
- `ApolloParseException` response parse errors

See new API:
https://github.com/apollographql/apollo-android/pull/379/files#diff-18999fee7e5f70c2f862624cbab6d833
https://github.com/apollographql/apollo-android/pull/379/files#diff-b059ebbc703ddbbde44ba44eed598837

@mandrizzle


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloPrefetch.java
```diff
@@ -1,23 +1,33 @@
...
 
   @Nonnull ApolloPrefetch enqueue(@Nullable Callback callback);
 
   ApolloPrefetch clone();
 
   void cancel();
 
-  interface Callback {
-    void onSuccess();
+  abstract class Callback {
```
BenSchwab: Nice, like the option to use a single callback and inheritance, or the split methods.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -25,10 +27,22 @@
...
+  abstract class Callback<T> {
+    public abstract void onResponse(@Nonnull Response<T> response);
+
+    public abstract void onFailure(@Nonnull ApolloException e);
+
+    public void onHttpError(@Nonnull ApolloHttpException e) {
+      onFailure(e);
+    }
+
+    public void onNetworkError(@Nonnull ApolloNetworkException e) {
```
mandrizzle: I think it would be nice to pass in the status code and response here, making this look like:

```java
public void onNetworkError(int status, Response response, @Nonnull ApolloNetworkException e)
```

I fell it is nice not having to dig through the exception, secondly, it makes it clearer what we have to work with

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -25,10 +27,22 @@
...
   @Nonnull ApolloCall<T> clone();
 
-  interface Callback<T> {
-    void onResponse(@Nonnull Response<T> response);
+  abstract class Callback<T> {
+    public abstract void onResponse(@Nonnull Response<T> response);
+
+    public abstract void onFailure(@Nonnull ApolloException e);
+
+    public void onHttpError(@Nonnull ApolloHttpException e) {
```
mandrizzle: I think it would be nice to pass in the status code and response here, making this look like:

```java
public void onNetworkError(int status, Response response, @Nonnull ApolloNetworkException e)
```

I fell it is nice not having to dig through the exception, secondly, it makes it clearer what we have to work with

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -25,10 +27,22 @@
...
+      onFailure(e);
```
mandrizzle: Thoughts on calling `closeQuietly(e.response())` after we call `onFailure`? I think it would prevent a lot of leaked connections

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
sav007: Not sure about this, as it becomes too clumsy to have 3 params to handle one error.
And if user really decides to handle network error then don't see any issue to dig into `ApolloNetworkException` to extract what he needs.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
mandrizzle: Should we have the exception `ApolloNetworkException`? I think an `IOException` is more fitting for this case

sav007: We decided to wrap with ApolloNetworkException that extends ApolloException.
So from our sync API call (`com.apollographql.apollo.ApolloCall#execute`) user should expect only one type of errors `ApolloException`.

PR #385, opened 2017-04-01
Enforce nullable type generation via plugin extension config

sav007:
The issue: sometimes plugin generates guava optionals even if there is no  such dependency in main classpath.

This PR adds plugin configuration option, to enforce nullable type generation:
- `annotated` will generate nullable values with JSR annotation
- `autoOptional` (by default) will check if guava dep present guava optional will be used
- `guavaOptional` use guava optional class
- `apolloOptional` use apollo internal optional class


apollo-gradle-plugin/src/main/java/com/apollographql/android/gradle/ApolloClassGenTask.java
```diff
@@ -42,8 +42,11 @@ void generateClasses(IncrementalTaskInputs inputs) {
     inputs.outOfDate(new Action<InputFileDetails>() {
       @Override
       public void execute(InputFileDetails inputFileDetails) {
+        boolean useOptional = (nullValueType != ApolloExtension.NullableValueType.annotated);
+        boolean useGuava = (nullValueType == ApolloExtension.NullableValueType.guavaOptional
+            || (hasGuavaDep && nullValueType == ApolloExtension.NullableValueType.autoOptional));
```
marwanad: Might as well get rid of the `autoOptional` and the `useOptional` boolean and only have `annotated`, `guavaOptional`, `apolloOptional` be part of the enum.

Less configuration for the user and it's safe to assume that someone doing `ApolloExtension.NullableValueType.guavaOptional` would have guava in classpath.

marwanad:
Would it be better to use the enum under `NullableValueGenerationType.kt` in `ApolloExtension` and pass that along to `GraphQLCompiler`?

apollo-gradle-plugin/src/test/groovy/com/apollographql/android/gradle/integration/BasicAndroidSpec.groovy
```diff
@@ -148,16 +150,35 @@ class BasicAndroidSpec extends Specification {
         "return Currency.class;")
   }
 
-  def "adding generateOptional in Apollo Extension generates classes with apollographql Optional"() {
+  def "adding nullableValueType = `annotated` in Apollo Extension generates classes with Apollo Optional"() {
```
marwanad: *shouldn't generate?

Might be worth checking for the javax import in that case.

PR #388, opened 2017-04-01
Java docs

VisheshVadhera:
Partially addresses #48.

Added docs for the following classes/interfaces:

-  /apollo/api/internal/Utils.java

- ApolloCall.java

- ApolloClient.java

- ApolloPrefetch.java

- HttpCacheControl.java

- Cancelable.java

- RxApollo.java

More docs on the way!


sav007:
pls rebase to latest master

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+/**
+ * ApolloCall is an abstraction for a request that has been prepared for execution. ApolloCall
+ * represents a single request/response pair and cannot be executed twice, though it can be cancelled.
```
sav007: I guess it worth to mention smth like to execute it twice `clone` should be used

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+ */
 public interface ApolloCall<T> extends Cancelable {
 
+  /**
+   * Sends the request immediately and blocks until the response can be processed or is an error.
+   *
+   * @return The successful or failed {@link Response}
+   * @throws ApolloException       if the request could not be executed due to a cancellation, a timeout or a network
```
sav007: or parsing error, the full list of possible error types: 
`com.apollographql.apollo.exception.ApolloHttpException`
`com.apollographql.apollo.exception.ApolloNetworkException`
`com.apollographql.apollo.exception.ApolloParseException`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+   *                               failure
+   * @throws IllegalStateException when the call has already been executed
+   */
   @Nonnull Response<T> execute() throws ApolloException;
 
+  /**
+   * Schedules the request to be executed at some point in the future.
+   * The dispatcher defines when the request will run: usually immediately unless there are several other requests
+   * currently being executed.
```
sav007: Not sure about the mentioned dispatcher, as it's not quite correct as we are actually using 2 dispatchers: 1st used to kickoff the request plus it will go to normalized cache, 2nd is okHttp3 dispatcher that is used internally with http call `okhttp3.Call#enqueue`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+   */
   @Nonnull ApolloCall<T> clone();
 
+  /**
+   * Communicates responses from a server or offline requests.
+   */
   abstract class Callback<T> {
+
+    /**
+     * Gets called when GraphQl response is received successfully.
```
sav007: Should be changed to `Gets called when GraphQl response is received and parsed successfully`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+     *
+     * @param response the GraphQl response
```
sav007: Should we use word `GraphQL` maybe change it to `GraphQL Operation` com.apollographql.apollo.api.Operation

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+     */
     public abstract void onResponse(@Nonnull Response<T> response);
 
+    /**
+     * Gets called when a network exception occurs while communicating with the server, or an unexpected
```
sav007: should we omit `network exception occurs while communicating with the server` and use just `Gets called when an unexpected exception occurs while creating the request or processing the response.`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -13,25 +13,83 @@
...
+     * exception occurs while creating the request or processing the response.
+     */
     public abstract void onFailure(@Nonnull ApolloException e);
 
+    /**
+     * Gets called when an http request error takes place either due to client error (status code >= 400) or due
+     * to server error (status code >= 500).
+     */
```
sav007: I would say any http status code not from 200 series: okhttp3.Response#isSuccessful 
```
public boolean isSuccessful() {
    return code >= 200 && code < 300;
  }
```

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -40,19 +98,40 @@ public void onHttpError(@Nonnull ApolloHttpException e) {
...
+    /**
+     * Gets called when the network request succeeds but there was an error parsing the response.
+     */
     public void onParseError(@Nonnull ApolloParseException e) {
       onFailure(e);
     }
   }
 
+  /**
+   * Factory for creating ApolloCall & ApolloPrefetch objects.
```
sav007: `Factory for creating apollo calls` ?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -124,34 +143,79 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
...
+     */
     public Builder okHttpClient(@Nonnull OkHttpClient okHttpClient) {
       this.okHttpClient = checkNotNull(okHttpClient, "okHttpClient is null");
       return this;
     }
 
+    /**
+     * <p>Set the API server's base url.</p>
+     *
+     * <b>Base URLs should not have a trailing slash</b>
```
sav007: Is it?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -124,34 +143,79 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
...
+     * @return The {@link Builder} object to be used for chaining method calls
+     */
     public Builder httpCache(@Nonnull ResponseCacheStore cacheStore, @Nonnull EvictionStrategy evictionStrategy) {
       this.httpCacheStore = checkNotNull(cacheStore, "cacheStore == null");
       this.httpEvictionStrategy = checkNotNull(evictionStrategy, "evictionStrategy == null");
       return this;
     }
 
+    /**
+     * Set the cache to use as the normalized cache store.
```
sav007: `Set the configuration to be used for normalized cache` ?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -124,34 +143,79 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
...
+     * @param serverUrl the url to set.
+     * @return The {@link Builder} object to be used for chaining method calls
+     */
     public Builder serverUrl(@Nonnull String serverUrl) {
       this.serverUrl = HttpUrl.parse(checkNotNull(serverUrl, "serverUrl == null"));
       return this;
     }
 
+    /**
+     * Set the cache to be used for reading and writing cached http responses.
```
sav007: `Set the configuration to be used for request/response http cache` ?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -124,34 +143,79 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
...
+     */
     public Builder normalizedCache(@Nonnull CacheStore cacheStore,
         @Nonnull CacheKeyResolver<Map<String, Object>> cacheKeyResolver) {
       this.cache = new RealCache(checkNotNull(cacheStore, "cacheStore == null"),
           checkNotNull(cacheKeyResolver, "cacheKeyResolver == null"));
       return this;
     }
 
+    /**
+     * Set the type adapter to use for serializing and de-serializing custom types.
```
sav007: *custom GraphQL scalar types

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
```diff
@@ -1,11 +1,39 @@
...
 
+/**
+ * HttpCacheControl class represents the cache controlling strategies for reading http responses from the cache when a
+ * network request is made.
+ */
 public enum HttpCacheControl {
+
+  /**
+   * Signals the apollo client to first fetch the data from the http cache. If the data is not present in
+   * the cache, then it is fetched from the network.
```
sav007: should we mention `is not present or stale` ?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
```diff
@@ -1,11 +1,39 @@
...
+   */
   CACHE_FIRST("cache-first"),
+  /**
+   * Signals the apollo client to only fetch the data from the http cache. If the data is not present in the cache, an
+   * exception is thrown.
```
sav007: I guess we can mention `ApolloHttpException` is returned

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
```diff
@@ -1,11 +1,39 @@
...
+   */
   CACHE_ONLY("cache-only"),
+  /**
+   * Signals the apollo client to only fetch the data from the network request. If the network request fails, an
+   * exception is thrown.
```
sav007: the same ApolloHttpException

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
```diff
@@ -1,11 +1,39 @@
...
+   */
   NETWORK_ONLY("network-only"),
+  /**
+   * Signals the apollo client to first fetch the data from the network request. If the network request fails, then
+   * the data is fetched from the http cache. If the data in the cache has become stale, an exception is thrown.
```
sav007: ApolloHttpException

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Agreed.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Oh yeah, need to add parsing as one of the possible reasons.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Right.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Oh yeah, missed the parsing part. Thanks for pointing it out.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: I feel we should use GraphQL only because the Response (which contains not just Operation but also contains data and list of errors) object represents the successful or failed response and is not solely representative of  the Operation.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Yeah, this needs to be changed to what you suggested because this is not a specific callback unlike onHttpError or onNetworkError, rather is representative of the overall failure of the entire operation.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: Alright, but what about status>=300 && status<400. Should they also lead to a HttpError? That's why I only mentioned 4XX & 5XX.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
VisheshVadhera: The purpose of this class is to create Call & Prefetch objects. So felt that factory can be used, though if you feel it is unsuitable, we can change it to something like `creates ApolloCall & ApolloPrefetch`.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: Done

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
sav007: yes everything that not in range 2xx leads to HttpError

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
VisheshVadhera: I would say is stale is a better repesentation

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: So I checked a couple of public graphQL apis and all of them seem to be not having a trailing slash. So I assume that should indeed be the case. What is your take on this? Should we add it to the docs?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
```diff
@@ -1,11 +1,39 @@
...
   NETWORK_FIRST("network-first"),
+  /**
+   * Signals the apollo client to first fetch the data from the network request. If the network request fails, then
+   * the data is fetched from the http cache even if it has become stale.
+   */
   NETWORK_BEFORE_STALE("network-before-stale"),
+  /**
+   * Signals the apollo client to mark the data in the cache stale whenever a read takes place next.
+   */
   EXPIRE_AFTER_READ("expire-after-read");
```
VisheshVadhera: @sav007 Do you feel this can be improved further? I feel there is still some scope for improvement for this particular strategy

VisheshVadhera:
Have made the requested changes. Though I still would want clarification on a couple of points. See my last two comments above on the points where I want more clarification.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
sav007: no, that's right. Http cache for this request will be deleted after the first read

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
sav007: I guess it's really depends on server side router configuration, there is no such limitations on client side( on our side). So I would rather to no mention this at all.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCacheControl.java
VisheshVadhera: Alright

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: Cool, so I'll remove the line

VisheshVadhera:
Final set of changes have been made.

VisheshVadhera:
Final set of changes have been made and if everything else looks fine, we can go ahead and merge this branch

digitalbuddha:
👍 I think this is good as first pass. We can refine more in future prs.

PR #395, opened 2017-04-05
Add isCanceled method to the Cancelable interface

### LINKED ISSUE #394: Cancellable interface needs to have isCancelled method

VisheshVadhera:
Closes #394


VisheshVadhera:
Alright.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloWatcher.java
```diff
@@ -62,6 +62,10 @@
     cache.unsubscribe(recordChangeSubscriber);
   }
 
+  @Override public boolean isCanceled() {
+    return !isActive;
```
BenSchwab: nit: Mind changing the field `isActive` to `isCanceled`. Seems to better fit the model

VisheshVadhera: In fact that's what I wanted to do in the first place, but decided against it because I didn't really want to change the existing code. But it makes more sense to change it to isCanceled.

VisheshVadhera:
Added test cases for the following Cancelable classes:

- ApolloCall
- ApolloPrefetch
- ApolloWatcher

Basically took the approach suggested by @BenSchwab and used a countdown latch. Please let me know if this is correct. Other comments/feedback are welcome!

PR #396, opened 2017-04-05
Rename Cache to ApolloStore, CacheStore to Normalized cache; Add ability to clearAll from CacheStore

BenSchwab:


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/CacheStore.java
```diff
@@ -34,4 +34,6 @@
     return aggregatedDependentKeys;
   }
 
+  public abstract void clearAll();
```
sav007: JavaDocs?

sav007:
Just curios what will happen if one thread is in process of merging recordset and another tries to clear all?

Should we call clear all in write transaction?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/CacheStore.java
BenSchwab: Adding in next PR with `CacheHeaders`

BenSchwab:
@sav007 Good point about synchronization. `CacheStore` does not have access to the transactions as it at a lower level. Generally, I anticipated that `CacheStore` should not have it's methods called by clients, and only internally, but `clear` is breaking that paradigm. (I see no need to call `clear` internally.)

So, I think the best bet will be to make the core `CacheStore` read/merge methods synchronized. This isn't needed internally, because we already have a locking mechanism, but shouldn't cause any harm either.

BenSchwab:
So abstract classes can't have synchronized methods -- so switched this over to the iOS architecture. (Users inject the higher level, synchronization aware `Cache`). With the architectures more similar, I feel like it makes more sense to unify the names. 

The general idea here is that `CacheStore` should never be called directly by user. If user disregards, we make no guarantee about thread-safety. On the other hand, `RealCache` is thread safe.

As our architecture is aligned with iOS, we should unify names. Generally, `Cache` -> `ApolloStore` and `CacheStore` -> `NormalizedCache`. 

Will fix names in follow up PR with javadocs.

https://github.com/apollographql/apollo-android/issues/397

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -195,14 +192,10 @@ public Builder httpCache(@Nonnull ResponseCacheStore cacheStore, @Nonnull Evicti
...
      *
-     * @param cacheStore       the cache store to use
-     * @param cacheKeyResolver the cache key resolver
      * @return The {@link Builder} object to be used for chaining method calls
      */
-    public Builder normalizedCache(@Nonnull CacheStore cacheStore,
-        @Nonnull CacheKeyResolver<Map<String, Object>> cacheKeyResolver) {
-      this.cache = new RealCache(checkNotNull(cacheStore, "cacheStore == null"),
-          checkNotNull(cacheKeyResolver, "cacheKeyResolver == null"));
+    public Builder normalizedCache(@Nonnull Cache cache) {
```
sav007: I guess we need to revert this back, as RealCache is supposed to be hidden from user (it's under internal package). Unfortunately java doesn't allow to make it internal per jar file. But the idea that user should not deal with any class from apollo internal package

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealCache.java
```diff
@@ -63,6 +68,15 @@ public RealCache(@Nonnull CacheStore cacheStore,
...
   }
 
+  @Override public void clearAll() {
+    writeTransaction(new Transaction<WriteableCache, Boolean>() {
+      @Override public Boolean execute(WriteableCache cache) {
+        cache.clearAll();
+        return true;
+      }
+    });
+  }
```
sav007: 👍

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlStore.java
```diff
@@ -86,6 +89,10 @@ private SqlStore(ApolloSqlHelper dbHelper, FieldsAdapter parser) {
     return changedKeys;
   }
 
+  @Override public void clearAll() {
```
sav007: We should not forget to add javadoc and mention that this function shouldn't be called directly from client only via Cache

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/CacheStore.java
sav007: We should not forget to add javadoc and mention that this function shouldn't be called directly from client only via Cache.clearAll

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlStore.java
BenSchwab: Yes, technically all methods in this class should not be called directly. Will add in the javadocs. https://github.com/apollographql/apollo-android/issues/397

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
BenSchwab: I think in this case `RealCache` should be moved from the internal package. If we want to expose thread-safe `clear` we will have to allow the user to access the `RealCache` object. The only way to do that and keep the builder patter is to have user inject an instance of `RealCache`.

https://github.com/apollographql/apollo-android/issues/397 I plan to rename it to `ApolloStore` to better align with iOS and web naming.

BenSchwab:
If https://github.com/apollographql/apollo-android/issues/397 sounds good, I'll just wrap it into this PR to save some time

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
sav007: Sorry confused, why do we need to expose `RealCache`. Can't user just call 'apolloClient.cache().clearAll()' or even `apolloClient.clearCache()`?

BenSchwab: Yes, that is also an option.

BenSchwab:
+ Add top level `clearNormalizedCache` to ApolloClient.
+ For custom implementations of the cache, a client may need a way to run operations in a thread-safe manner. So also added `apolloStore()` to ApolloClient, which exposes the thread-safe API.

PR #404, opened 2017-04-09
Polymorphic inline fragments

### LINKED ISSUE #334: Polymorphic Type Handling
Currently the Codegeneration puts Type Instances as fields to an Generic Object.
To make it more clear here is an Example:

For a query like:
```
query HeroForEpisode($ep: Episode!) {
  hero(episode: $ep) {
    ... on Droid {
        name      
    }
    ... on Human {
      name      
    }
  }
}
```

we do get the following Generated as Java Code:

```java
public static class Hero {
  private @Nullable AsHuman AsHuman;

  private @Nullable AsDroid AsDroid;

  public Hero(@Nullable AsHuman AsHuman,
    @Nullable AsDroid AsDroid) {
    this.AsHuman = AsHuman;
    this.AsDroid = AsDroid;
  }

  public @Nullable AsHuman AsHuman() {
    return this.AsHuman;
  }

  public @Nullable AsDroid AsDroid() {
    return this.AsDroid;
  }

  public static class AsHuman {
    private final @Nullable String name;

    public AsHuman(@Nullable String name) {
      this.name = name;
    }

    public @Nullable String name() {
      return this.name;
    }
  }

  public static class AsDroid {
    private final @Nullable String name;

    public AsDroid(@Nullable String name) {
      this.name = name;
    }

    public @Nullable String name() {
      return this.name;
    }
  }
}
```

This makes detection of Subtypes not convenient as you always have to check weather the return type of `as<type>()`is null. In my opinion using `thing.instanceOf(Clazz.class)` produces better code.

My suggestion is to change the generic class to abstract and create instances of the mapped subtypes, so the generated Java Classes would look something like this:

```java
public abstract static class Hero {
  public Hero() {
    
    }

  public static class AsHuman extends Hero {
    private final @Nullable String name;

    public AsHuman(@Nullable String name) {
      this.name = name;
    }

    public @Nullable String name() {
      return this.name;
    }
  }

  public static class AsDroid extends Hero{
    private final @Nullable String name;

    public AsDroid(@Nullable String name) {
      this.name = name;
    }

    public @Nullable String name() {
      return this.name;
    }
  }
}
```

sav007:
Closes #334 

Make inline fragment to extend parent type:
https://github.com/apollographql/apollo-android/pull/404/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46R119
https://github.com/apollographql/apollo-android/pull/404/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46R183

Make type that contains inline fragments to be empty (as all graphql fields are merged into inline fragments by apollo codegen) and have only references to fragments
https://github.com/apollographql/apollo-android/pull/404/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46R247

Make Mapper to return first instantiate of non null inline fragment
https://github.com/apollographql/apollo-android/pull/404/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46R325

Overall overview what the result will be:
https://github.com/apollographql/apollo-android/pull/404/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46


apollo-compiler/src/test/graphql/com/example/simple_inline_fragment/TestQuery.java
```diff
@@ -245,34 +245,26 @@ public AsDroid map(ResponseReader reader) throws IOException {
...
-      }
-
-      public Optional<AsHuman> asHuman() {
-        return this.asHuman;
-      }
-
-      public Optional<AsDroid> asDroid() {
-        return this.asDroid;
+      public Hero() {
+        if (this instanceof AsHuman) {
```
BenSchwab: What would happen if there was a third type such as `asPlant`. And `Plant extends Character`. This query should still return if the `Plant` is a hero, and `name` should be accessible.

apollo-compiler/src/test/graphql/com/example/simple_inline_fragment/TestQuery.java
```diff
@@ -335,10 +323,15 @@ public AsDroid read(String conditionalType, ResponseReader reader) throws IOExce
 
         @Override
         public Hero map(ResponseReader reader) throws IOException {
-          final String name = reader.read(fields[0]);
-          final AsHuman asHuman = reader.read(fields[1]);
-          final AsDroid asDroid = reader.read(fields[2]);
-          return new Hero(name, asHuman, asDroid);
+          final AsHuman asHuman = reader.read(fields[0]);
```
BenSchwab: What would happen if I specified two fragments that acted on Droid? This seems like it would return the first one, making the second one in accessible

sav007: Yes, that's what Martijn mentioned in slack, we don't support several inline fragments to be acted on Droid and on Character at the same time, for now.

>I don’t think we need to complicate it right now, and we should probably just rely on only one inline fragment matching. I suspect we’ll want to keep using subclasses for concrete objects even if we add better support for interfaces later on, and then we could have the classes implement the interface.

apollo-compiler/src/test/graphql/com/example/simple_inline_fragment/TestQuery.java
sav007: We don't support multiple inline fragments (see comment below).
And codegen merges all fields in IR for inline fragments. So you will have access to `name` field as it will be merged into `Plant` fragment

PR #407, opened 2017-04-11
RxJava2 Module

### LINKED ISSUE #345: Add RxJava2 module

VisheshVadhera:
Closes #345 

Added Rx2 wrappers for the following classes in a separate module for Rx2 types:

1. ApolloPrefetch (Completable)
2. ApolloCall (Single)
3. ApolloWatcher (Observable)

Felt that there is no need for an explicit function to get a backpressure supported Rx2 type for ApolloWatcher as we can always convert an Observable to a Flowable with toFlowable method on it.


apollo-rx2support/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=compiler
+POM_NAME=Apollo Android Compiler
+POM_DESCRIPTION=Implementation for the Gradle plugin
+POM_PACKAGING=jar
```
digitalbuddha: Shouldn't it be aar since it depends on apollo runtime  which is aar

VisheshVadhera: Yep it should be aar. Also need to make changes to the other variables. Pushing it.

sav007: it should be jar, and pure java module due to:
https://github.com/apollographql/apollo-android/pull/409

VisheshVadhera: Cool, I'll change it to jar

apollo-rx2support/src/main/java/com/apollographql/android/rx2/Rx2Apollo.java
```diff
@@ -0,0 +1,148 @@
...
+ * and ApolloWatcher types to RxJava 2 sources.
+ */
+public class Rx2Apollo {
+
+  private Rx2Apollo() {
+    throw new AssertionError("This class cannot be instantiated");
+  }
+
+  /**
+   * Converts an {@link ApolloPrefetch} to an asynchronous Observable.
```
digitalbuddha: technically the Observable is not async, it has a callback but until someone uses a scheduler everything with still run synchronous by default

VisheshVadhera: I think I meant ApolloWatcher here not ApolloPrefetch. But anyways even if you feel that Observable here is a sync one I present to you my argument.

So if we assume that on the client side no scheduler is being used, then when the client subscribes to this observable it will simply block waiting for emissions to take place. In that case the client would forever remain blocked because we never call onComplete in the wrapper function for ApolloWacther. The only possible way by which client would be able to move to the next statement of execution is if the other terminal event (which is onError) somehow is emitted. And I think that's not what our intention with ApolloWatcher is. We don't want clients to simply while using ApolloWatcher.

VisheshVadhera:
@digitalbuddha Yeah, I'll just create an issue and start work on it once this gets merged.

digitalbuddha:
Lgtm

digitalbuddha:
You need to exclude rxjava.properties from apk to make build pass

VisheshVadhera:
Fixed the duplicate file exception by adding the following line to apollo-integration's build.gradle

```
packagingOptions {
    exclude 'META-INF/rxjava.properties'
  }
```

@digitalbuddha Also wanted to know what kind of examples do you want me to add to the sample module. Do you just want me to add a simple Rx example for apollocall or you want me to add examples of other apollo types as well?

digitalbuddha:
Just a simple example would be fine showing how to get an observable, how to subscribe to it and how to unsubscribe on destroy.  Hmm CI still failing with weird errors not sure why :-/

apollo-rx2support/build.gradle
```diff
@@ -0,0 +1,58 @@
+buildscript {
+  dependencies {
+    classpath dep.androidPlugin
+  }
+}
+
+apply plugin: 'com.android.library'
+clearSkipApolloRuntimeDep()
+
+android {
```
sav007: pls convert this module to pure java, as runtime and api right now java modules

VisheshVadhera: which would also mean converting the rx2 module to pure java? So I'll create an issue for rx1 module and create a PR for it a well.

VisheshVadhera:
@digitalbuddha The builds are failing with a 137, which is an indication that the gradle process was killed by SIGKILL and is likely due to less memory resources than required. Should we try increasing the heap size?

PR #408, opened 2017-04-11
Rx Completable for ApolloPrefetch

### LINKED ISSUE #389: RxApollo doesn't have a wrapper function for ApolloPrefetch

VisheshVadhera:
closes #389


apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
```diff
@@ -103,6 +106,37 @@ private RxApollo() {
...
+   */
+  @Nonnull public static Completable from(@Nonnull final ApolloPrefetch prefetch) {
+    checkNotNull(prefetch, "prefetch == null");
+
+    return Completable.create(new Completable.OnSubscribe() {
+      @Override public void call(CompletableSubscriber subscriber) {
+        cancelOnCompletableUnsubscribe(subscriber, prefetch);
+        try {
+          prefetch.execute();
+          subscriber.onCompleted();
```
sav007: Do we need to check:
```
if (!subscriber.isUnsubscribed()) {
}
```

apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
```diff
@@ -103,6 +106,37 @@ private RxApollo() {
...
+        } catch (ApolloException e) {
+          Exceptions.throwIfFatal(e);
+          subscriber.onError(e);
```
sav007: Do we need to check:
```
if (!subscriber.isUnsubscribed()) {
}
```

VisheshVadhera: Done

apollo-rxsupport/src/main/java/com.apollographql.android.rx/RxApollo.java
VisheshVadhera: Done

apollo-integration/src/test/java/com/apollographql/apollo/RxApolloTest.java
```diff
@@ -88,7 +92,7 @@
     TestSubscriber<EpisodeHeroName.Data> subscriber = new TestSubscriber<>();
     Subscription subscription = RxApollo
         .from(call)
-        .delay(5, TimeUnit.SECONDS)
+        .delay(RX_DELAY_SECONDS, TimeUnit.SECONDS)
```
digitalbuddha: what is point of the delay? will this slow our tests down by 5 seconds?

VisheshVadhera: The docs for delay say:

> This version of delay operates by default on the computation Scheduler.

which means that its a non blocking call and hence wont block the thread running the test. So the reason behind introducing delay is to let the observable do work in another thread while in the main thread we try to unsubscribe from the it. Basically simulate a real world scenario of unsubscribing from a Subscription.

apollo-integration/src/test/java/com/apollographql/apollo/RxApolloTest.java
```diff
@@ -29,13 +29,17 @@
...
 import rx.observers.TestSubscriber;
 
 import static com.google.common.truth.Truth.assertThat;
 
 public class RxApolloTest {
+
   private ApolloClient apolloClient;
   private MockWebServer server;
+
+  private static final int RX_DELAY_SECONDS = 5;
```
sav007: we can reduce it to 2 I guess

VisheshVadhera: Done

Issue #244, opened 2017-02-24
Enqueue should provide an override to take a handler


brianPlummer:
Enqueue's onResponse() delivers data on background thread. there needs to be a mechanism where results can be delivered on the main thread (handler)

digitalbuddha:
Idea is to allow user to pass in a handler when calling client.enqueue

marwanad:
We can accept an `Executor` instead (to avoid the android dependency) and users would wrap their Handler calls in that. 

I might start working on that later tonight if that's fine.

digitalbuddha:
Sounds good!

sav007:
@marwanad our `:apollo-runtime` module is android lib, so we ok to use `Handler`, and from Android perspective its better to have handler than executor

marwanad:
@sav007 I didn't see using any android components now or in the future planned APIs for cache so I initially wanted to go with an Executor because that way we can just make a normal java lib.
But yeah if we're keeping it as android library, we might as well go with `Handler`. 

The thing about the Handler is there are different methods to schedule messages to the handler so we're making the assumption that the user would prefer `Handler.post()` vs `Handler.postDelayed()` for instance.

An `Executor` would be agnostic to the threading technique used within. Users can still use a `Handler` if they want inside `Executor.execute()`. I feel this is more flexible and more applicable if we just wanna extend the callback execution to anything vs just android ui threads.

sav007:
The whole point of having `Handler` is to provide flexible way for user to get callback on UI thread, that is only the good reason to use handler. With `Executor` that won't be possible, and why do we need to switch to another thread in callback from working thread? All calls is scheduled  on `OkHttp` dispatcher (that internally is using `Executor`).

digitalbuddha:
Is the library java only right now? I think it's a valid argument to keep base library java only if we can. Maybe there is a way we can do this similar to rxjava where RxAndroid does the scheduling on Main thread but through a Scheduler interface. Rxjava stays java only.

marwanad:
> With Executor that won't be possible

Correct me if I'm wrong but wouldn't something like that run on the UI thread?

```
class AnExecutor implements Executor {
      private final Handler mhandler = new Handler(Looper.getMainLooper());

      @Override public void execute(Runnable r) {
        mhandler.post(r);
      }
 }
````

> why do we need to switch to another thread in callback from working thread

[...]

PR #409, opened 2017-04-11
Extract Android specific stuff to separate module

### LINKED ISSUE #347: Consider Creating Apollo-Android Module
Currently we have runtime as AAR since it depends on the SQL stuff (and handler stuff that will land when @brianPlummer is back from vacation).  We should consider making an android only module and having everything else be java only.

sav007:
Closes #347 and #244


apollo-android-support/src/androidTest/java/com/apollographql/apollo/cache/normalized/sql/ApolloPrefetchCallbackTest.java
```diff
@@ -0,0 +1,175 @@
...
+  @Before public void setUp() throws Exception {
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(new OkHttpClient.Builder()
+            .connectTimeout(2, TimeUnit.SECONDS)
+            .readTimeout(2, TimeUnit.SECONDS)
+            .build())
+        .build();
+
+    emptyQuery = new Query() {
```
digitalbuddha: seems like multiple tests use this, can we make a factory to dedup the code? Alternatively put in `TestUilts` or smth

PR #410, opened 2017-04-12
Add Java 8 Optional Api map, flatmap

sav007:


sav007:
@digitalbuddha the real intention of these changes is to have flat map, that will make chain look nicer:

before:
```
response -> response.data()
        .transform(it -> it.node.get())
        .transform(it -> it.asCheckout.get())
        .transform(it -> it.availableShippingRates.get())
```

after:
```
response -> response.data()
        .flatmap(::node)
        .flatmap(::asCheckout)
        .flatmap(::availableShippingRates)
```

PR #413, opened 2017-04-14
Prepare for v0.3.0 release

felipecsl:


felipecsl:
Release draft saved [here](https://github.com/apollographql/apollo-android/releases/tag/untagged-c83e47e8f1671c8d1f8b)

felipecsl:
`apollo-integration` is not happy about the upgrade though. What do you guys usually do in this case? merge with the red status and fix after the snapshot is uploaded?

BenSchwab:
Hmm, seems like this is because apollo-integration (which test apollo-runtime) uses a project dependency instead of relying on the published artifact. But it does not do the same with the gradle plugin.

I think ideally, we would have apollo-integration also depend on a "project version" of the gradle plugin, but not super familiar with gradle plugins, and how to set that up.

in the meantime, seems like mergeing and fixing later is fine. The base branch commit passed all checks.

gradle.properties
```diff
@@ -1,5 +1,5 @@
 GROUP=com.apollographql.apollo
-VERSION_NAME=0.2.3-SNAPSHOT
+VERSION_NAME=0.3.1-SNAPSHOT
```
BenSchwab: I suppose we could just make this `0.3.0`, publish, and then bump again?

gradle/dependencies.gradle
```diff
@@ -1,5 +1,5 @@
 def versions = [
-    apolloVersion    : '0.2.3-SNAPSHOT',
+    apolloVersion    : '0.3.0',
```
marwanad: This variable is used for the sample and integration modules so it should stay as `0.2.3-SNAPSHOT` because v0.3.0 doesn't exist.

marwanad: been out of the loop but if 0.3.0 isn't out yet, this should stay as `0.2.3-SNAPSHOT` because it's used to pull dependencies from maven for the sample and apollo-integration.

felipecsl: 0.3.0 is out alreadu

digitalbuddha:
Can this be closed?  Should Readme be updated?

marwanad:
@felipecsl iirc, the way we do version bumps without having modules complain is bump `VERSION_NAME` and update the .travis file with

`-x apollo-integration -x apollo-gradle-plugin:test`

Then after the merge, re-enable tests in a new commit.

Issue #326, opened 2017-03-15
Add CacheHeaders


BenSchwab:
We need a way to pass metadata for the cache. Examples might be `TTL` or `offlineCritical`.
This header info will be passed to `CacheStore` for read and writes. As `CacheStore` is user configurable, I feel we should make this CacheHeader a key-value String map. This allows a user to add custom CacheHeaders that make sense for their individual implementation of CacheStore.

Couple options on how to implement for the API:

```
public final class CacheControl {
     public CacheControl(CacheFetchPolicy) ...
     public CacheControl(CacheFetchPolicy, Map<String, String> cacheHeaders)
}

public enum CacheFetchPolicy {
  CACHE_FIRST,
  CACHE_ONLY,
  NETWORK_FIRST,
  NETWORK_ONLY,
}
```

or in `ApolloCall` just: 
```
  @Nonnull ApolloCall<T> cacheControl(@Nonnull CacheControl cacheControl);
 @Nonnull ApolloCall<T> cacheControl(@Nonnull CacheControl cacheControl, @Nullable Map<String, String> cacheHeaders);

```

sav007:
But if we talking about `TTL` that completely on the impl of cache store, when writing response to cache impl of normalized cache store should add this to the record

sav007:
As for `offlineCritical ` I'm not sure that we need this option, you can get this functionality by setting `TTL` to some large value, and cache store will keep records for a long time

BenSchwab:
But don't we want different `TTL` for different requests? Some request (say a message thread) should never be evicted (or at least, have a very low eviction priority). Whereas a temporary search results should be evicted quickly.

sav007:
> But don't we want different TTL for different requests

Should it be defined by request or by record? CacheStore implementation when merge records can attach different TTL for different records. The only that we are missing the type of the record.

> Some request (say a message thread) should never be evicted

What if the message referenced to user, this TTL for messages will be used as TTL for user as well?

BenSchwab:
The message/user case you brought up is the perfect example as to why it needs to be on the request:

If a `message` request stores a `user`, and the `message` is marked as "never expire", we expect that anytime I run that request, it will return -- including the `user`.

However, a different request, this time just for a `user`  may be marked as "ttl 1 minute". Although this request accesses the same record in the normalized cache, the cache should actually return `null` if the record hasn't been updated in more than 1 minute.

[...]

PR #414, opened 2017-04-14
Add CacheHeaders

BenSchwab:
Closes https://github.com/apollographql/apollo-android/issues/326

API change: https://github.com/apollographql/apollo-android/pull/414/files#diff-18999fee7e5f70c2f862624cbab6d833R67

While it's up to each NormalizedCache if they want to respect headers (similar to http headers), it is probably good to have a canonized set of headers that caches that apollo ships with obeys, and that all caches are recommended to follow:

Standard cache headers: https://github.com/apollographql/apollo-android/pull/414/files#diff-f8f76830abf8c20e474e6df061d85d63R7


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -101,9 +104,8 @@ private ApolloClient(Builder builder) {
       }
     }
     return new RealApolloCall<T>(operation, serverUrl, httpCallFactory, httpCache, defaultHttpCacheControl, moshi,
-        responseFieldMapper, customTypeAdapters, apolloStore, defaultCacheControl, dispatcher, logger)
-        .httpCacheControl(defaultHttpCacheControl)
```
BenSchwab: @sav007 These didn't seem necessary, am I missing anything?

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
```diff
@@ -53,11 +55,14 @@
     deleteAllRecordsStatement = database.compileStatement(DELETE_ALL_RECORD_STATEMENT);
   }
 
-  @Nullable @Override public Record loadRecord(String key) {
+  @Nullable public Record loadRecord(String key, Map<String, String> cacheHeaders) {
```
sav007: Maybe we should add annotations here:
```
loadRecord(@Nonnull String key,@Nonnull Map<String, String> cacheHeaders)
```

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
```diff
@@ -53,11 +55,14 @@
...
     return selectRecordForKey(key).orNull();
   }
 
-  @Nonnull @Override public Set<String> merge(Record apolloRecord) {
+  @Nonnull public Set<String> merge(Record apolloRecord, Map<String, String> cacheHeaders) {
```
sav007: the same we are missing annotations nullable|nonull for arguments

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
```diff
@@ -40,21 +43,24 @@ protected RecordFieldAdapter recordAdapter() {
   }
 
   /**
-   * @param key The key of the record to read
+   * @param key          The key of the record to read
+   * @param cacheHeaders The cache headers associated with the request which generated this record.
    * @return The {@link Record} for key. If not present return null.
    */
-  @Nullable public abstract Record loadRecord(String key);
+  @Nullable public abstract Record loadRecord(String key, Map<String, String> cacheHeaders);
```
sav007: Should we wrap Map<String, String> into custom object like `CacheHeaders`. So that if we decide to change impl from Map to smth else it will be much easier for refactoring

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/StandardCacheHeaders.java
```diff
@@ -0,0 +1,14 @@
+package com.apollographql.apollo.cache.normalized;
+
+/**
+ * A collection of "standardized" cache headers that an implementation {@link NormalizedCache}
+ * is recommended to respect.
+ */
+public class StandardCacheHeaders {
```
sav007: interface?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -54,14 +57,35 @@
...
 
   /**
+   * Sets the cache headers to use for this call. The default {@link CacheControl} will be used.
+   *
+   * @param cacheHeaders the cache headers that will be passed with records generated from this request to {@link
+   *                     com.apollographql.apollo.cache.normalized.NormalizedCache}. Standardized cache headers are
+   *                     defined in {@link StandardCacheHeaders}.
+   * @return The ApolloCall object with the provided CacheControl strategy
+   */
+  @Nonnull ApolloCall<T> cacheControl(@Nonnull Map<String, String> cacheHeaders);
```
sav007: should we call this `cacheHeaders`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -54,14 +57,35 @@
...
+  /**
+   * Sets the {@link CacheControl} strategy for an ApolloCall object.
+   *
+   * @param cacheControl the CacheControl strategy to set
+   * @param cacheHeaders the cache headers that will be passed with records generated from this request to {@link
+   *                     com.apollographql.apollo.cache.normalized.NormalizedCache}. Standardized cache headers are
+   *                     defined in {@link StandardCacheHeaders}.
+   * @return The ApolloCall object with the provided CacheControl strategy
+   */
+  @Nonnull ApolloCall<T> cacheControl(@Nonnull CacheControl cacheControl, @Nonnull Map<String, String> cacheHeaders);
```
sav007: I guess we don't need this method. As we can do this with chain:
```
call.cacheControl(cacheControl).cacheHeaders(cacheHeaders);
```

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/StandardCacheHeaders.java
```diff
@@ -0,0 +1,14 @@
...
+
+  /**
+   * Records from this request should not be stored in the {@link NormalizedCache}.
+   * This does **not** specify that a request should not be read from a cache.
+   */
+  public static final String DO_NOT_CACHE = "do-not-cache";
```
sav007: Shouldn't it be part of CacheControl? It becomes confusing that we have 2 ways how to control cache. I thought the CacheHeaders should act as a helper to CacheControl. (like TTL) but not control the cache itself.

BenSchwab: I can merge 'CacheHeaders` and `CacheControl` into a single object if that seems better. But they are different concepts.

 I would argue that `CacheControl` should have been named `FetchStrategy` -- it controls how we look up data.

`CacheHeaders` instructs the cache to do certain things -- such as not storing records, or returning null because the TTL is expired.

sav007: Seems the right thing to change the name to `CacheFetchStrategy` plus I would like to merge `HttpCacheControl` with `CacheControl` if it possible, what do you think?

BenSchwab: I'll look into it!

BenSchwab:
I'll morph this into rethinking cache control designs: https://github.com/apollographql/apollo-android/issues/432 for discussion

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/StandardCacheHeaders.java
BenSchwab: Checkstyle complained when I made it an interface because it has no methods.

BenSchwab:
Sorry for the delay on this. Decided to keep this focused only on `CacheHeaders`. `FetchStrategy` will come after.

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
```diff
@@ -66,18 +82,20 @@
     } else {
       Record oldRecord = optionalOldRecord.get();
       changedKeys = oldRecord.mergeWith(apolloRecord);
-      if (!changedKeys.isEmpty()) {
-        updateRecord(oldRecord.key(), recordAdapter().toJson(oldRecord.fields()));
-      }
+      updateRecord(oldRecord.key(), recordAdapter().toJson(oldRecord.fields()));
```
sav007: just to make sure that if condition was removed intentionally

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
```diff
@@ -40,21 +45,24 @@ protected RecordFieldAdapter recordAdapter() {
   }
 
   /**
-   * @param key The key of the record to read
+   * @param key          The key of the record to read.
+   * @param cacheHeaders The cache headers associated with the request which generated this record.
    * @return The {@link Record} for key. If not present return null.
    */
-  @Nullable public abstract Record loadRecord(String key);
+  @Nullable public abstract Record loadRecord(@Nonnull String key, @Nonnull CacheHeaders cacheHeaders);
```
sav007: Should we have overloaded version of this function without argument `CacheHeaders`?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/ApolloCacheHeaders.java
```diff
@@ -0,0 +1,21 @@
+package com.apollographql.apollo.cache;
+
+import com.apollographql.apollo.cache.normalized.NormalizedCache;
+
+/**
+ * A collection of cache headers that Apollo's implementations of {@link NormalizedCache} respect.
+ */
+public final class ApolloCacheHeaders {
```
sav007: interface?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/CacheHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+ * A key/value collection which is sent with {@link com.apollographql.apollo.cache.normalized.Record}
+ * from a {@link com.apollographql.apollo.api.Operation} to the
+ * {@link com.apollographql.apollo.cache.normalized.NormalizedCache}.
+ *
+ * For headers which the default {@link com.apollographql.apollo.cache.normalized.NormalizedCache} respect, see
+ * {@link ApolloCacheHeaders}.
+ */
+public final class CacheHeaders {
+
+  private Map<String, String> headerMap = new LinkedHashMap<>();
```
sav007: final and we don't need  initialization as it will be set in constructor

apollo-runtime/src/main/java/com/apollographql/apollo/cache/CacheHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public static CacheHeaders.Builder builder() {
+    return new Builder();
+  }
+
+  public static final CacheHeaders NONE = CacheHeaders.builder().build();
```
sav007: why not just new CacheHeaders(emptyMap);

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/lru/LruNormalizedCache.java
```diff
@@ -79,16 +81,25 @@
             return record;
           }
         });
+        return record;
```
sav007: but don't we need to check for `EVICT_AFTER_READ` as well here?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/ApolloCacheHeaders.java
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+   * Records from this request should not be stored in the {@link NormalizedCache}.
+   * This does not specify that a request should not be read from a cache. To control where
+   * a response is read use {@link com.apollographql.apollo.cache.normalized.CacheControl}.
+   */
+  public static final String NO_CACHE = "no-cache";
```
sav007: Can we come up with better naming, as without looking into JavaDoc the name confusing, does it mean it won't use cache at all for both reading/writing.

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
BenSchwab: Good catch, I was initially going to put in storing lastModified in this PR, but held off. Will revert this util it makes more sense.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
BenSchwab: That would seem to promote allowing a cache to ignore the` CacheHeaders`, which I don't think we wish to encourage.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/ApolloCacheHeaders.java
BenSchwab: According to our checkstyle, interfaces should have "behavior" and hence methods. I would prefer if this was an interfrace, but would have to change our checkstyle.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/lru/LruNormalizedCache.java
BenSchwab: Oops, good catch. Did not mean to have this return.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/ApolloCacheHeaders.java
BenSchwab: agreed

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -62,6 +64,16 @@
   @Nonnull ApolloCall<T> cacheControl(@Nonnull CacheControl cacheControl);
 
   /**
+   * Sets the cache headers to use for this call. The default {@link CacheControl} will be used.
+   *
+   * @param cacheHeaders the {@link CacheHeaders} that will be passed with records generated from this request to {@link
+   *                     com.apollographql.apollo.cache.normalized.NormalizedCache}. Standardized cache headers are
+   *                     defined in {@link ApolloCacheHeaders}.
+   * @return The ApolloCall object with the provided CacheControl strategy
```
sav007: I think it's copy paste: `...with the provided CacheControl strategy`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -62,6 +64,16 @@
   @Nonnull ApolloCall<T> cacheControl(@Nonnull CacheControl cacheControl);
 
   /**
+   * Sets the cache headers to use for this call. The default {@link CacheControl} will be used.
```
sav007: Is it right `The default {@link CacheControl} will be used.`

apollo-runtime/src/main/java/com/apollographql/apollo/cache/ApolloCacheHeaders.java
sav007: then I guess we need to have private constructor

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/CacheControl.java
```diff
@@ -1,8 +1,12 @@
...
 
+import com.apollographql.apollo.cache.CacheHeaders;
+
 /**
- * CacheControl class represents the cache controlling strategies for reading GraphQL responses back from the
- * {@link NormalizedCache}.
+ * {@link CacheControl} represents strategies for defining what precedence reading a GraphQL responses back
+ * from the {@link NormalizedCache}.
+ *
+ * To control how a cache reads a response see {@link CacheHeaders}.
```
sav007: Is it just for reading `To control how a cache reads`. I thought CacheHeaders is used for writing as well

Issue #317, opened 2017-03-13
"Apollo iOS" Error Message in Apollo-Android


commonsguy:
Attempting to add `__typename` to a query results in:

```
:app:generateDebugApolloIR
Apollo iOS inserts __typename automatically when needed, please do not include it explicitly
Validation of GraphQL query document failed
:app:generateDebugApolloIR FAILED
```

However, this is from a Gradle build of an Apollo-Android project, and there are no iOS projects within kilometers of here, AFAIK.

Should these messages be genericized to just "Apollo"? Or should they be platform-aware?

digitalbuddha:
@martijnwalraven could we make the error message more generic, I think this is comming from code gen.

PR #419, opened 2017-04-16
Misc Fixes

marwanad:
Bump apollo-codegen version, update README for v0.3 release.

Also, removes the unnecessary `skipApolloRuntimeDep` method since project dependencies
are included in the resolved compile dependency sets and so the apollo-integration module will use the project version of `apollo-runtime` (verified that locally).


marwanad:
closes #317 which is fixed by the apollo-codegen version bump.

PR #421, opened 2017-04-17
Builder improvements proposal

hzsweers:
This has a few changes:
- `build()` is now non-mutating, so you can call `build()` multiple times without mutating its internal state and get consistent results at the tradeoff of being a little more verbose.
- This makes the builder accept an HttpCache directly with a caveat that your provided client should also have an interceptor backed by this. This eliminates the part where we manually added the interceptor under the hood (and a bug where, related to the first bullet, we would add multiple copies if `build()` was called more than once!)
- Convenience default build of a `Call.Factory` if one isn't provided, and convenience addition of the provided `HttpCache`'s interceptor onto it if one is present.
- Finally - the builder doesn't have a reference to an `OkHttpClient` for the ApolloClient to construct with - just a `Call.Factory`. This comes with a corresponding `callFactory()` setter, which is a nice feature that allows for a layer of indirection (lazily-initialized http clients, alternate `Call.Factory` implementations, etc)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -200,16 +211,15 @@ public Builder serverUrl(@Nonnull String serverUrl) {
...
-     * @param evictionStrategy EvictionStrategy decides when the data in the cacheStore becomes stale and is no longer
-     *                         valid
+     * @param httpCache The HttpCache to use.
      * @return The {@link Builder} object to be used for chaining method calls
+     * @see HttpCache#interceptor()
      */
-    public Builder httpCache(@Nonnull ResponseCacheStore cacheStore, @Nonnull EvictionStrategy evictionStrategy) {
-      this.httpCacheStore = checkNotNull(cacheStore, "cacheStore == null");
-      this.httpEvictionStrategy = checkNotNull(evictionStrategy, "evictionStrategy == null");
+    public Builder httpCache(@Nonnull HttpCache httpCache) {
```
sav007: HttpCache is in internal package, if we want to expose it then we need to move it from internal and make sure that we can safely open it

sav007: Another issue is with the constructor:
```
HttpCache(@Nonnull ResponseCacheStore cacheStore, @Nonnull EvictionStrategy evictionStrategy,
      @Nonnull ApolloLogger logger)
```
With 3rd argument, logger. So how user will instantiate this class, how logger will be provided?
Maybe instead of argument to constructor make it via setter

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -301,28 +311,38 @@ public Builder logger(@Nullable Logger logger) {
...
-      if (httpCacheStore != null && httpEvictionStrategy != null) {
-        httpCache = new HttpCache(httpCacheStore, httpEvictionStrategy, apolloLogger);
-        okHttpClient = okHttpClient.newBuilder().addInterceptor(httpCache.interceptor()).build();
+      okhttp3.Call.Factory callFactory = this.callFactory;
+      boolean localClient = this.callFactory == null;
+      if (localClient) {
+        callFactory = new OkHttpClient();
+      }
+
+      if (httpCache != null && callFactory instanceof OkHttpClient) {
```
sav007: can we wrap httpCache to Optional

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -301,28 +311,38 @@ public Builder logger(@Nullable Logger logger) {
...
-      apolloLogger = new ApolloLogger(logger);
-      moshi = moshiBuilder.build();
+      ApolloLogger apolloLogger = new ApolloLogger(logger);
+      Moshi moshi = moshiBuilder.build();
 
-      if (httpCacheStore != null && httpEvictionStrategy != null) {
-        httpCache = new HttpCache(httpCacheStore, httpEvictionStrategy, apolloLogger);
-        okHttpClient = okHttpClient.newBuilder().addInterceptor(httpCache.interceptor()).build();
+      okhttp3.Call.Factory callFactory = this.callFactory;
+      boolean localClient = this.callFactory == null;
```
sav007: can we wrap callFactory to Optional

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -173,7 +175,16 @@ private Builder() {
...
-      this.okHttpClient = checkNotNull(okHttpClient, "okHttpClient is null");
+      return callFactory(checkNotNull(okHttpClient, "okHttpClient is null"));
+    }
+
+    /**
+     * Set the custom call factory for creating {@link Call} instances.
+     * <p>
+     * Note: Calling {@link #okHttpClient(OkHttpClient)} automatically sets this value.
+     */
+    public Builder callFactory(Call.Factory factory) {
```
sav007: `@Nonnull Call.Factory factory`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -301,28 +311,38 @@ public Builder logger(@Nullable Logger logger) {
...
-        okHttpClient = okHttpClient.newBuilder().addInterceptor(httpCache.interceptor()).build();
+      okhttp3.Call.Factory callFactory = this.callFactory;
+      boolean localClient = this.callFactory == null;
+      if (localClient) {
+        callFactory = new OkHttpClient();
+      }
+
+      if (httpCache != null && callFactory instanceof OkHttpClient) {
+        if (localClient) {
+          apolloLogger.w("Created local client, adding passed in HttpCache");
```
sav007: That's confusing as we don't create actually local client, we just add interceptor to it.
Maybe it will be better to change it to smth like this:
```
apolloLogger.w("Add passed in HttpCache to provided http client");
```

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -301,28 +311,38 @@ public Builder logger(@Nullable Logger logger) {
...
 
+      ApolloStore apolloStore = this.apolloStore;
       if (cacheFactory.isPresent() && cacheKeyResolver.isPresent()) {
         final NormalizedCache normalizedCache =
             cacheFactory.get().createNormalizedCache(RecordFieldAdapter.create(moshi));
-        this.apolloStore = new RealApolloStore(normalizedCache, cacheKeyResolver.get());
+        apolloStore = new RealApolloStore(normalizedCache, cacheKeyResolver.get());
       }
 
+      ExecutorService dispatcher = this.dispatcher;
```
sav007: I know we forgot about this, but please wrap it to Optional

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
sav007: Pls ignore above comment, we actually create it.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -301,28 +311,38 @@ public Builder logger(@Nullable Logger logger) {
...
+      okhttp3.Call.Factory callFactory = this.callFactory;
+      boolean localClient = this.callFactory == null;
+      if (localClient) {
+        callFactory = new OkHttpClient();
+      }
+
+      if (httpCache != null && callFactory instanceof OkHttpClient) {
+        if (localClient) {
+          apolloLogger.w("Created local client, adding passed in HttpCache");
+          callFactory = ((OkHttpClient) callFactory).newBuilder().addInterceptor(httpCache.interceptor()).build();
```
sav007: should we merge this with condition above, so that this nested if will go away:
```
if (localClient) {
  callFactory = new OkHttpClient().newBuilder().addInterceptor(httpCache.interceptor()).build();
}
```

digitalbuddha:
awaiting few fixes from comments.

hzsweers:
hey - sorry I'm traveling this week. I'll try to get around to this later today

digitalbuddha:
Np! It's not blocking anything. It can wait. Enjoy the travels

digitalbuddha:
Hey hey, mind cleaning up and merging or closing?

hzsweers:
Yep I'll circle back to this tonight or tomorrow!

hzsweers:
(got pulled in to io stuff, this weekend I promise)

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: I'm not really a fan of using Optional in implementation details, would much rather keep it clean. There's quite a few other nullable places that don't use `Optional` internally right now

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: Any further thoughts on opening it up? The current API after the HttpCacheStore change prevents proper use of this builder proposal :/.

Even with it open, it's sort of a weird implicit requirement that it have an interceptor set. I kind of almost wonder if this interceptor should be handled independently somehow. Thoughts?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: See above

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: See above

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: Reworked in the rebased changes

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
hzsweers: Fixed in the rebased changes

hzsweers:
Rebased and replied to comments. 

I'm not really sure what the need is for using Optional in local implementation details and want to ask if it's really necessary, as it just induces allocation overhead and readability complexity for something that the IDE + static analysis can tell you. I can make the change if you feel strongly about it.

The HttpCache interceptor bit makes this really awkward. Currently - a lot of tests will fail unless I force the underlying implementation to always add an interceptor, which is awkward for two reasons:
  1. It doesn't know if one has already been added (potential for duplicates)
  2. It requires everyone to use an OkHttpClient anyway rather than Call.Factory, kind of defeating the purpose of this PR.

We could open HttpCache up and go with the `localClient` check, but it's still an awkward API since it makes implicit requirements to add an interceptor manually before. I think it would do well to move the interceptor bit to something internal, but not really sure how that would look. Thoughts?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -381,35 +398,47 @@ public Builder addApplicationInterceptor(@Nonnull ApolloInterceptor interceptor)
...
-      moshi = moshiBuilder.build();
+      ApolloLogger apolloLogger = new ApolloLogger(logger);
+      Moshi moshi = moshiBuilder.build();
 
+      okhttp3.Call.Factory callFactory = this.callFactory;
+      boolean localClient = this.callFactory == null;
+      if (localClient) {
+        callFactory = new OkHttpClient();
+      }
+
```
sav007: Can we just check if user provided callFactory is instanceof OkHttpClient then add interceptor otherwise just ignore and log warning that http cache won't be enabled unless user adds it by himself.

sav007:
@hzsweers I have only comment but I think this PR looks good as is right now

hzsweers:
So I'm not totally sure how to proceed on this. I can just hook it up if it's an OkHttpClient instance, but doesn't not having the cache be public preclude anyone from trying to hook anything else up via the interceptor? Would be really curious to know your thoughts on what I proposed above

sav007:
> It doesn't know if one has already been added (potential for duplicates)
We can check if `okHttpClient.interceptors()` has already interceptor of type `com.apollographql.apollo.internal.cache.http.HttpCacheInterceptor` (but before make it public)

>It requires everyone to use an OkHttpClient anyway rather than Call.Factory, kind of defeating the purpose of this PR.

So again check if user has provided `OkHttpClient` then auto add interceptor on builder side.
If it was provided just `Call.Factory` then user must add it manually.

So then we need to change `com.apollographql.apollo.ApolloClient.Builder#httpCacheStore(@Nonnull HttpCacheStore cacheStore)` to `httpCache(HttpCache httpCache)`
And make `HttpCache` to be in public package (not internal)

hzsweers:
ok - I've extracted HttpCache + its interceptor to a public package, added the check for the interceptor if it's an okhttpclient instance, and rebased.

I still think it's going to be important to revisit trying to eliminate this asterisk from the builder API, but that can be something for the future :)

hzsweers:
Yep I'll jump on those next, just want to get the refactoring up and ok'd first before proceeding

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -387,38 +406,62 @@ public Builder addApplicationInterceptor(@Nonnull ApolloInterceptor interceptor)
...
+      if (localClient) {
+        callFactory = new OkHttpClient();
+      }
 
+      HttpCache httpCache = null;
       if (httpCacheStore != null) {
         httpCache = new HttpCache(httpCacheStore, apolloLogger);
-        okHttpClient = okHttpClient.newBuilder().addInterceptor(httpCache.interceptor()).build();
+        if (localClient || shouldAddHttpCacheInterceptor(callFactory)) {
+          apolloLogger.w("Created local client, adding passed in HttpCache");
```
sav007: Probably the log is confusing, first we don't always create local client, second user doesn't pass HttpCache but HttpCacheStore

hzsweers:
Sorry been on vacation, going to revisit this when I'm back next week

hzsweers:
Well actually I'm sitting in an airport so rebased now

hzsweers:
Alright this should do it, tests passing locally for me now. Squashed two bugs from before in b2585a5 and also in the rebase where the local `apolloStore` in `build()` wasn't actually getting set, causing NPEs in a lot of tests.

sav007:
Green yay!!!!

hzsweers:
aw you merged a rebase conflict in :|

sav007:
Yeah, sorry thought it would be green and we could merge it

hzsweers:
Rebased again

PR #426, opened 2017-04-18
Add memoizing toString()/hashCode()

### LINKED ISSUE #420: Proposal: Add option to memoize toString()/hashCode()
Since the models are immutable, memoizing hashCode/toString() would be nice. This is a common use case for AutoValue's `@Memoized` extension as well, and would make its implementation a good reference.

sav007:
Closes #420 

Similar result to what AutoValue Memoized generates.

@hzsweers


hzsweers:
So some of us were discussing this and think that the synchronized block is actually overkill, and that the cost of re-calculating is probably less than the cost of the synchronized lock. I'd say just keep it simple and skip it. @digitalbuddha thoughts?

digitalbuddha:
I agree skip it.

sav007:
@digitalbuddha @hzsweers any comments?

PR #428, opened 2017-04-18
Feature application interceptors

### LINKED ISSUE #346: Make Apollo application interceptors to be extendable
Apollo application interceptors should be open and extendable, so user can register own interceptor with ApolloClient (similar to what OkHttp does).

VisheshVadhera:
Closes #346


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -84,6 +88,7 @@ private ApolloClient(Builder builder) {
     this.defaultHttpCacheControl = builder.defaultHttpCacheControl;
     this.defaultCacheControl = builder.defaultCacheControl;
     this.logger = builder.apolloLogger;
+    this.applicationInterceptors = builder.applicationInterceptors;
```
sav007: pls wrap this into `Collections.unmodifiableList()` or even earlier in `ApolloClient.Builder`

sav007: Never mind we don't set in builder the list of interceptors but add them one by one

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -153,6 +149,20 @@ public RealApolloCall(Operation operation, HttpUrl serverUrl, Call.Factory httpC
 
   @Override @Nonnull public RealApolloCall<T> clone() {
     return new RealApolloCall<>(operation, serverUrl, httpCallFactory, httpCache, httpCacheControl, moshi,
-        responseFieldMapper, customTypeAdapters, apolloStore, cacheControl, dispatcher, logger);
+        responseFieldMapper, customTypeAdapters, apolloStore, cacheControl, dispatcher, logger,
+        applicationInterceptors);
+  }
+
+  private List<ApolloInterceptor> getInterceptors() {
```
sav007: just a nit: we usually omit `get` prefixes. So I guess we can make this method to return `RealApolloInterceptorChain` so to have smth like this:
```
private ApolloInterceptorChain prepareInterceptorChain(Operation operation, List< ApolloInterceptor> applicationInterceptors) {
...
return new RealApolloInterceptorChain(operation, interceptors);
}
```

apollo-integration/src/test/java/com/apollographql/apollo/InterceptorTest.java
```diff
@@ -0,0 +1,432 @@
...
+    try {
+      mockWebServer.shutdown();
+    } catch (IOException e) {
+
+    }
+  }
+
+  @Test
+  public void syncApplicationInterceptorCanShortCircuitResponses() throws IOException, ApolloException {
+    mockWebServer.shutdown();
```
sav007: Do we need this?

apollo-integration/src/test/java/com/apollographql/apollo/InterceptorTest.java
```diff
@@ -0,0 +1,432 @@
...
+
+  private ApolloClient getApolloClient(ApolloInterceptor interceptor) {
+    return ApolloClient.builder()
+        .serverUrl(mockWebServer.url("/"))
+        .okHttpClient(okHttpClient)
+        .applicationInterceptor(interceptor)
+        .build();
+  }
+
+  @NonNull private InterceptorResponse getInterceptorResponse(EpisodeHeroName query) {
```
sav007: nit: we don't like `get` prefixes :)

apollo-integration/src/test/java/com/apollographql/apollo/InterceptorTest.java
```diff
@@ -0,0 +1,432 @@
...
+      @Override public void dispose() {
+        //No op
+      }
+    };
+
+    client = getApolloClient(interceptor);
+
+    Response<EpisodeHeroName.Data> actualResponse = client.newCall(query).execute();
+
+    assertThat(expectedResponse.parsedResponse.get()).isEqualTo(actualResponse);
```
sav007: would be nice if we check that other interceptors aren't called.

apollo-integration/src/test/java/com/apollographql/apollo/InterceptorTest.java
VisheshVadhera: Yeah, we do need this, so as to ensure that response is indeed the intercepted response and not  from the server.

apollo-integration/src/test/java/com/apollographql/apollo/InterceptorTest.java
VisheshVadhera: I am wondering how can we test that the other interceptors aren't called? Maybe check that the `httpResponse` & `cachedResponse` are equal to the fake responses that we passed to the `interceptedResponse`? But that would tell us nothing about other interceptors being called or not called.

VisheshVadhera:
@sav007 It would be difficult to move this class to the apollo-runtime module because InterceptorTest class needs objects like NamedCountdownLatch & EpisodeHeroName which are only available in the apollo-integration module. Moving it to apollo-runtime would mean adding apollo-integration as a dependency for apollo-runtime which we don't want.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: Well, the initial thought was to wrap it in an unmodifiable list, but then I realized that we are not exposing the list of interceptors in any way to the clients, so it might as well just remain a simple List.

VisheshVadhera:
@sav007 Changed the method names as requested. If everything else looks fine, we can go ahead and merge it.

VisheshVadhera:
@sav007 Oh btw, do you have any inputs on how we can test that the rest of the interceptors are not called when we short circuit responses?

sav007:
create a unit tests for RealApolloInterceptorChain add couple interceptors and make one of them to return value right away. Then call`interceptorChain.proceed()` and check if only those interceptors who supposed to be are called

VisheshVadhera:
@sav007 Added the test which checks that other interceptors are not called when one of the the interceptors short circuits the response. Also made changes to the docs of the corresponding builder method to reflect the same.

sav007:
@VisheshVadhera sorry to request so much, but could you please add additional test to verify if `com.apollographql.apollo.interceptor.ApolloInterceptorChain#dispose` is called for application interceptors as well.

VisheshVadhera:
Np, I'll be happy to add them, the more tests we have the better it is for
the project.

VisheshVadhera:
@sav007 Added a test case for the dispose method, plus a couple more for asyncInterceptors. Also I realized we don't have tests for ApolloInterceptorChain. Want me to open another PR?

digitalbuddha:
lgtm

digitalbuddha:
Yup please do that in the next pr to keep this one from growing anymore

sav007:
Merging?

digitalbuddha:
Lgtm

PR #431, opened 2017-04-18
ApolloWatcherTest- cancel watcher at end of each test

BenSchwab:
We saw an ApolloWatcherTest spuriously fail on CI. I could not get it fail locally, but trying a few things to make the tests more stable

- Bump timeout to 5 seconds
- Cancel watcher at end of each test
- Remove countdown on errors -- we already timeout if a latch does not countdown
- Some style fixes

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
@@ -13,7 +13,6 @@ import com.apollographql.apollo.exception.ApolloException;
 
 import junit.framework.Assert;
 
-import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
```diff
@@ -34,13 +33,11 @@
 public class ApolloWatcherTest {
   private ApolloClient apolloClient;
   private MockWebServer server;
-  private NormalizedCache normalizedCache;
-  private static final int TIME_OUT_SECONDS = 3;
+  private static final int TIME_OUT_SECONDS = 5;
```
digitalbuddha: this is my least favorite change in this pr.  I don't think the issue is that our code takes an addl 2 seconds to run on a jvm

BenSchwab: Fair enough -- creating a `MockWebServer` is an expensive operation. Wasn't sure how fast our CI instance was, but 2 seconds probably isn't a factor. I'll revert.

BenSchwab:
Added this switch as another preventive measure:

https://github.com/apollographql/apollo-android/pull/431/files#diff-b4677f256e817e6b559705ed46bd59f6R85

A watcher unsubscribes while it is refetching. But it would call `onResponse` before resubscribing. This may be what caused the issue because this would allow the countdown latch to decrement and a new call to be made on another thread before resubscribing.

PR #433, opened 2017-04-19
Apollo integration composite build

### LINKED ISSUE #139: Use buildSrc to add the built plugin to the sample classpath
instead of pulling it from maven

https://docs.gradle.org/current/userguide/organizing_build_logic.html#sec:build_sources

marwanad:
This converts `apollo-integration` module to a[ composite build](https://docs.gradle.org/current/userguide/composite_builds.html) as suggested on #139.

This would basically substitute the dependencies for apollo-integration with project dependencies (including the plugin).

It might seem odd but the reason it's done through shell script is because the conversion requires apollo-integration to not be a sub-module. I don't think Android Studio supports opening multiple projects under the same window so if you're viewing the code for that module it would look like that:

image

(No autocompletion, can't run tests through the IDE ,etc) (I'm not sure if that's fixable so would appreciate any input here)

The shell scripts copies the files under `gradle/composite/`, runs `./gradlew -p apollo-integration build` after all the other modules have been built and then resets everything after. This run would include the substituted dependencies. 

The other alternative is building the plugin under `buildSrc` which would place it in the classpath of the other modules build-scripts but the downside is that it slows down the builds heavily.


sav007:
> (No autocompletion, can't run tests through the IDE ,etc) (I'm not sure if that's fixable so would appreciate any input here)

What that means, we can't write integration tests using IDEA autocomplete, auto compile check?

marwanad:
yeah, since `apollo-integration` will have to be a separate module, so it won't be listed under the project view. The only way to run those tests would be through command line. Even then, writing them could be a pain since there's no auto-complete. You can still open the project in a new window and do the work there but that's still annoying.

The shell script solves this issue, everything works as before and whenever you want to run the integration tests against the project version of the plugin run 
`.buildscript/integration_tests_composite.sh`. On Travis, it would always run that shell script.

sav007:
> The shell script solves this issue, everything works as before and whenever you want to run the integration tests against the project version of the plugin run
.buildscript/integration_tests_composite.sh

Cool, I thought that we will loose the autocomplete and ability to run tests under IDE

PR #434, opened 2017-04-19
Proposal: Make generated class structure even more flat

sav007:
Proposal is to make class generation be more flat and nest response model classes under `Query` itself, and not under `Query.Data`.

Example:
`com.example.hero_name.TestQuery.Hero` (instead of `com.example.hero_name.TestQuery.Data.Hero`)


digitalbuddha:
I think you should make a TestQuery package with individual non nested classes

sav007:
Not sure that making generation without nested classes good idea.

Imagine that my package is `com.example.domain.model`. We have 2 queries: `CollectionWithProductsQuery` and `ProductDetailsQuery`. Both queries define the same model `Product`. 

If we remove nesting then in place where we need to reference both versions of `Product` model we have to specify fully qualified package name:
```
com.example.domain.model.collectionWithProductsQuery.Product product1 = ...
com.example.domain.model.productDetailsQuery.Product product2 = ...
```

With nested we don't have such issue:
```
CollectionWithProductsQuery.Product product1 = ...
ProductDetailsQuery.Product product2 = ...
```

PR #447, opened 2017-04-23
Emit whole response instead just data in Rx wrapper

### LINKED ISSUE #436: Rx Support calls leave errors unreachable
When errors are thrown on a GraphQL request, the response is often still HTTP code 200 from the server, and the client is expected to look at the "errors" section of the JSON response to understand that an error has occurred.

In the current implementation we pass the result through the call ` emitter.onSuccess(response.data());` we prevent the user from checking these errors.

I recommend checking `response.isSuccessful()` first, and passing the response wrapped in an exception to `emitter.onError()` if `response.isSuccessful() == false`. Or alternatively the entire response object should be passed to `onSuccessful`.

sav007:
Closes #436 
Remove response success verification.
Emit whole response in Rx observables

cc @VisheshVadhera @kassim

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/Response.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/Response.java
@@ -4,6 +4,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
+import javax.annotation.Nonnull;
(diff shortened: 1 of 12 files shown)


VisheshVadhera:
@sav007 We forgot to make the corresponding set of changes for RxApollo class.

Issue #451, opened 2017-04-24
Introduce store custom resolver


sav007:
Overview of solution:
http://dev.apollodata.com/react/cache-updates.html#cacheRedirect

First step is to be able read cached response for query directly from store.

sav007:
@martijnwalraven could you please share how this works on iOS if it's implemented

sav007:
@BenSchwab @digitalbuddha @helfer I would love to hear any comments / suggestions.

sav007:
Here is some thoughts.

To be able read cached response for operation we need just substitute the way how resolved root cache record for this operation.

#### Step 1
Introduce custom `cacheKey` in `ApolloCall` that will be used in `ApolloCacheInterceptor` where cached root record is being resolved. 

So instead of:
```java
Record rootRecord = cache.read(CacheKeyResolver.rootKeyForOperation(operation).key());
```

To have smth like this:
```java
Record rootRecord;
if (operation has custom cacheKey) {
 rootRecord = cache.read(operation.cacheKey());
} else {
 rootRecord = cache.read(CacheKeyResolver.rootKeyForOperation(operation).key());
}
```

To pass custom cache key to cache interceptor along with operation object we need to introduce internally `Request` object that will be passed between interceptors and it will incapsulate `operation`, `cacheKey`, `cacheHeaders`, `cachePolicy` etc.

#### Step 2
Calculation of cache keys for operation is done as part of parse process inside `ApolloParseInterceptor`.  So why not to use this and simplify the way how user provides this custom `cacheKey`.  

So let say user from the response of this query:

```graphql
query ListView {
  books {
    id
    title
    abstract
  }
}
```
would be able to easily extract cache key (that has been already calculated during parsing):
```java
cacheKey  = response.data.listView.books[0].cacheKey
```

and pass this key as the custom cache key to the detail query:
```graphql
query DetailView {
  book(id: $id) {
    id
    title
    abstract
  }
}
```

In this case we just need to decide how we can store calculated cache keys for parsed response data and provide access to to them.

digitalbuddha:
I'll look over some more. One comment after reading intro. Rather than if (operation has custom cacheKey) {
 rootRecord = cache.read(operation.cacheKey());
} else {
 rootRecord = cache.read(CacheKeyResolver.rootKeyForOperation(operation).key());
}

We should have a default cacheKeyResolver that a use can override. Code would flow nicer.

digitalbuddha:
[...]

PR #457, opened 2017-04-26
Add auto generation of __typename field

sav007:
Part of #451

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Field.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Field.kt
@@ -12,21 +12,24 @@ data class Field(
     val responseName: String,
     val fieldName: String,
     val type: String,
-    val args: List<Map<String, Any>>?,
+    val args: List<Map<String, Any>>? = null,
     val isConditional: Boolean = false,
-    val fields: List<Field>?,
-    val fragmentSpreads: List<String>?,
-    val inlineFragments: List<InlineFragment>?,
-    val description: String?
+    val fields: List<Field>? = null,
+    val fragmentSpreads: List<String>? = null,
+    val inlineFragments: List<InlineFragment>? = null,
+    val description: String? = null
 ) : CodeGenerator {
-  override fun toTypeSpec(context: CodeGenerationContext): TypeSpec =
-      SchemaTypeSpecBuilder(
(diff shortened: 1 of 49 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Field.kt
```diff
@@ -12,21 +12,24 @@ data class Field(
...
+    val args: List<Map<String, Any>>? = null,
     val isConditional: Boolean = false,
-    val fields: List<Field>?,
-    val fragmentSpreads: List<String>?,
-    val inlineFragments: List<InlineFragment>?,
-    val description: String?
+    val fields: List<Field>? = null,
+    val fragmentSpreads: List<String>? = null,
+    val inlineFragments: List<InlineFragment>? = null,
+    val description: String? = null
```
marwanad: what made those null initializations necessary?

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Fragment.kt
```diff
@@ -10,16 +10,23 @@ import javax.lang.model.element.Modifier
...
     val inlineFragments: List<InlineFragment>,
     val fragmentsReferenced: List<String>
 ) : CodeGenerator {
+
   /** Returns the Java interface that represents this Fragment object. */
   override fun toTypeSpec(context: CodeGenerationContext): TypeSpec =
-      SchemaTypeSpecBuilder(formatClassName(), fields, fragmentSpreads, inlineFragments, context)
+      SchemaTypeSpecBuilder(
+          typeName = formatClassName(),
+          fields = listOf(Field("__typename", "__typename", "String!")) + fields,
```
BenSchwab: nvm, I guess that is just to the query. But this to actually add it as a field. 👍

sav007: yes, it auto adds to query but we still need to generate field for it

apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/ir/Field.kt
sav007: `Field("__typename", "__typename", "String!")` this stuff

PR #458, opened 2017-04-27
Custom cache key resolver

sav007:
Part of #451 

`CacheKeyResolver` is used in 2 places: first is to resolve cache key from the server response to write it to the store and second for reading cached response from the store.
Because of that `CacheKeyResolver` has been updated to handle these 2 cases:
```java
CacheKey fromFieldRecordSet(@Nonnull Field field, @Nonnull Map<String, Object> recordSet);

CacheKey fromFieldArguments(@Nonnull Field field, @Nonnull Operation.Variables variables);
```
`fromFieldRecordSet` will be used when parsing response from server and saving to the store.
`fromFieldArguments` will be used when reading cached response from the store.

Example of `CacheKeyResolver` implementation:
```java
public class IdFieldCacheKeyResolver extends CacheKeyResolver {
  @Nonnull @Override
  public CacheKey fromFieldRecordSet(@Nonnull Field field, @Nonnull Map<String, Object> recordSet) {
    return formatCacheKey((String) recordSet.get("id"));
  }

  @Nonnull @Override
  public CacheKey fromFieldArguments(@Nonnull Field field, @Nonnull Operation.Variables variables) {
    return formatCacheKey((String) field.resolveArgument("id", variables));
  }

  private CacheKey formatCacheKey(String id) {
    if (id == null || id.isEmpty()) {
      return CacheKey.NO_KEY;
    } else {
      return CacheKey.from(id);
    }
  }
}
```

Using string values as names for resolving field cache key in the record set or field arguments is not good idea, as user must have knowledge how this field is named in the server response, but for there is probably no better way how to fix it.


apollo-api/src/main/java/com/apollographql/apollo/api/Field.java
```diff
@@ -103,6 +103,21 @@ public String cacheKey(Operation.Variables variables) {
...
+  @SuppressWarnings("unchecked") public Object resolveArgument(String name, Operation.Variables variables) {
+    Map<String, Object> variableValues = variables.valueMap();
+    Object argumentValue = arguments.get(name);
+    if (argumentValue instanceof Map) {
+      Map<String, Object> argumentValueMap = (Map<String, Object>) argumentValue;
+      if (isArgumentValueVariableType(argumentValueMap)) {
+        String variableName = argumentValueMap.get(VARIABLE_NAME_KEY).toString();
+        return variableValues.get(variableName);
+      } else {
+        return null;
```
digitalbuddha: maybe optional instead of null?

sav007: I would love to have Optional here, but one concern is that this is public API so if user uses Guava for in his application won't it be confusing for him to deal with 2 different implementations of Optional, guava and our version?

apollo-api/src/main/java/com/apollographql/apollo/api/Field.java
```diff
@@ -103,6 +103,21 @@ public String cacheKey(Operation.Variables variables) {
     return String.format("%s(%s)", fieldName(), orderIndependentKey(arguments, variables));
   }
 
+  @SuppressWarnings("unchecked") public Object resolveArgument(String name, Operation.Variables variables) {
```
BenSchwab: Mark as `@Nullable`?

Issue #452, opened 2017-04-24
Introduce RefetchQuerries for mutations


BenSchwab:
Javascript documentation: http://dev.apollodata.com/core/apollo-client-api.html#ApolloClient.mutate

Use case: There is post feed, and a detail page where you can create a post. When we create the post, there may be no way for the normalized cache to know that the new post belongs to the feed query. One option is to directly write into the store, but an even simpler (albeit less efficient) is to specify "RefetchQueries" for the mutation.

Ideally, we will want to batch all the qurries in the refetch query into one POST call.

sav007:
Is this concept called Live Query?

chaudharyzia:
Hi sir,
Con you tell me the syntax of graphql “Mutation” in android.

sav007:
@chaudharyzia what do you mean by syntax?

PR #469, opened 2017-04-28
Espresso idling resource for Apollo GraphQL client

### LINKED ISSUE #429: Espresso Idling resource for ApolloClient

VisheshVadhera:
Closes #429

Added a new ApolloTracker class which keeps track of the running ApolloCall & ApolloPrefetch objects. ApolloIdlingResource basically uses this class (indirectly) to keep a count of the running call objects. When the count becomes zero it changes its state from active to idle. 

ApolloTracker is very similar to the Dispatcher class in okHttpClient, but without the ExecutorService encapsulated inside.


VisheshVadhera:
Please put the review on hold. I just found a flaky test. Will fix it and we can then review this one.

sav007:
```
com.apollographql.apollo.ApolloPrefetchCallbackTest > onNetworkError[test(AVD) - 4.0.4] FAILED 
Test failed to run to completion. Reason: 'Instrumentation run failed due to 'java.lang.AssertionError''. Check device logcat for details
Tests on test(AVD) - 4.0.4 failed: Instrumentation run failed due to 'java.lang.AssertionError'
:apollo-android-support:connectedDebugAndroidTest FAILED
```

VisheshVadhera:
@sav007 Fixing it.

VisheshVadhera:
@sav007 I have fixed the failing tests. Feel free to merge it, whenever you get time :)

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+import com.apollographql.apollo.ApolloCall;
+import com.apollographql.apollo.ApolloPrefetch;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * ApolloCallTracker is responsible for keeping track of running {@link ApolloCall} & {@link ApolloPrefetch} objects.
+ */
+public class ApolloCallTracker {
```
sav007: final

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -151,6 +154,20 @@ public ApolloStore apolloStore() {
     return apolloStore;
   }
 
+  /**
+   * Sets the idleCallback which will be called when this ApolloClient is idle.
+   */
+  public void setIdleCallback(Runnable idleCallback) {
```
sav007: Remove prefix `set`
I would make callback to be interface `interface IdelCallback` instead of generic `Runnable`

sav007: And should it be part of Builder?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -151,6 +154,20 @@ public ApolloStore apolloStore() {
...
+    callTracker.setIdleCallback(idleCallback);
+  }
+
+  /**
+   * Returns the sum of {@link ApolloCall} & {@link ApolloPrefetch} objects which are currently in progress.
+   */
+  public int getRunningCallsCount() {
```
sav007: drop prefix `get` and maybe just call it `activeCallCount`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -151,6 +154,20 @@ public ApolloStore apolloStore() {
...
 
+  /**
+   * Sets the idleCallback which will be called when this ApolloClient is idle.
+   */
+  public void setIdleCallback(Runnable idleCallback) {
+    callTracker.setIdleCallback(idleCallback);
+  }
+
+  /**
+   * Returns the sum of {@link ApolloCall} & {@link ApolloPrefetch} objects which are currently in progress.
```
sav007: * the count of ... calls that are currently in progress

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * <p><b>Note</b>: This method needs to be called right after a prefetch call is completed (whether successful or
+   * failed).</p>
+   */
+  void syncPrefetchFinished(ApolloPrefetch apolloPrefetch) {
+    Runnable idleCallback;
+    int runningCallsCount;
+    synchronized (this) {
+      if (!runningSyncPrefetches.remove(apolloPrefetch)) {
+        throw new AssertionError("Prefetch call wasn't in progress");
```
sav007: Do we really need to throw exception? is it critical?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+ * ApolloCallTracker is responsible for keeping track of running {@link ApolloCall} & {@link ApolloPrefetch} objects.
+ */
+public class ApolloCallTracker {
+
+  private Runnable idleCallback;
+
+  private final Deque<ApolloCall> runningSyncCalls = new ArrayDeque<>();
+  private final Deque<RealApolloCall.AsyncCall> runningAsyncCalls = new ArrayDeque<>();
+  private final Deque<ApolloPrefetch> runningSyncPrefetches = new ArrayDeque<>();
+  private final Deque<RealApolloPrefetch.AsyncCall> runningAsyncPrefetches = new ArrayDeque<>();
```
sav007: Do we need to track async and sync separately ?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -77,6 +78,7 @@ public static Builder builder() {
   private final CacheControl defaultCacheControl;
   private final CacheHeaders defaultCacheHeaders;
   private final ApolloLogger logger;
+  private final ApolloCallTracker callTracker = new ApolloCallTracker();
```
sav007: I guess we need to make it lazy, so if I don't need to tracker we should avoid overhead of  `ApolloCallTracker ` (it uses synchronization blocks a lot)

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+  public ApolloCallTracker() {
+  }
+
+  /**
+   * <p>Adds this {@link ApolloPrefetch} to the underlying data structure keeping track of the in progress synchronous
+   * prefetch objects.</p>
+   *
+   * <p><b>Note</b>: This method needs to be called right before a prefetch call is executed.</p>
+   */
+  synchronized void syncPrefetchInProgress(ApolloPrefetch apolloPrefetch) {
```
sav007: I guess would be better to add `on` prefix `onSyncPrefetchInProgress`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+  private void executeCallBackIfCallsAreFinished(int runningCallsCount, Runnable idleCallback) {
+    if (runningCallsCount == 0 && idleCallback != null) {
+      idleCallback.run();
+    }
+  }
+
+  /**
+   * Returns a total count of in progress {@link ApolloCall} & {@link ApolloPrefetch} objects.
+   */
+  public int getRunningCallsCount() {
```
sav007: rename to `activeCallCount()`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Deque;
+
+/**
+ * ApolloCallTracker is responsible for keeping track of running {@link ApolloCall} & {@link ApolloPrefetch} objects.
+ */
+public class ApolloCallTracker {
+
+  private Runnable idleCallback;
+
+  private final Deque<ApolloCall> runningSyncCalls = new ArrayDeque<>();
```
sav007: Why `Deque` can we use `LinkedHashSet` I guess it will be better for performance?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
VisheshVadhera: Yeah, I feel that's an exceptional circumstance, because any part of code trying to remove a call object which was never added is a programmer's fault and we should stop then and there itself.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: I feel it shouldn't be a part of the builder because it will force the clients to supply the callback at the time of creation of the apolloClient which I feel is a bit limiting. For example if it is being used in conjunction with a DI framework, the clients might not have the access to callback at all, for example in a Dagger2 module which contains a @Provides method for creating ApolloClient.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
VisheshVadhera: I guess so. Don't really have a strong opinion here but I felt that even though sync and async refer to the same object type, it's better to have different deques for tracking async and sync calls, just to keep concerns separate. We don't have a use case right now though wherein we only need to track all the async operations or only the sync operations. 

If you feel otherwise I'll just go ahead and make the change though.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: I am wondering how we are going to lazily create it. If we had a builder method to supply the `IdleCallback` then it would have been possible, but I have already expressed my concern over why we shouldn't take that route. 

Or maybe if we are thinking of creating the ApolloTracker when someone sets the idleCallback, then I am afraid that wouldn't give us an exact count of running calls e.g. if an async call is made and then an idleCallback is set, then the apolloTracker would give us 0 running calls even though we have a running call. 

Do you have anything else in your mind?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
VisheshVadhera: I think yeah, a `LinkedHashSet` makes more sense here and would give us better performance over a `Deque`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
sav007: What I meant is if user doesn't care about keep tracking of calls, then why should we create it at all and have some synchronization overhead.

But at the same time for bigger picture we probably need it.
Then I would call it not a tracker but `ActiveCallRegistry` or smth like this, and  it can be used not only for tracking how many calls we are running right now but potentially for batching requests. In this case we need to include as well watchers.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,174 @@
...
+      runningCallsCount = getRunningCallsCount();
+      idleCallback = this.idleCallback;
+    }
+    executeCallBackIfCallsAreFinished(runningCallsCount, idleCallback);
+  }
+
+  /**
+   * Registers idleCallback which is invoked when the apolloClient becomes idle.
+   */
+  public synchronized void setIdleCallback(Runnable idleCallback) {
```
sav007: I don;t think we need it to be synchronized, it doesn't have any value.

VisheshVadhera: Right.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
sav007: I would probably update this class later to enable this feature https://github.com/apollographql/apollo-android/issues/452

apollo-idling-resource/src/main/java/com/apollographql/apollo/espresso/ApolloIdlingResource.java
```diff
@@ -0,0 +1,61 @@
...
+import com.apollographql.apollo.IdleCallback;
+
+import javax.annotation.Nonnull;
+
+import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
+
+/**
+ * An Espresso {@link IdlingResource} for {@link com.apollographql.apollo.ApolloClient}.
+ */
+public class ApolloIdlingResource implements IdlingResource {
```
sav007: So we are creating new module just for this one class?

apollo-runtime/src/main/java/com/apollographql/apollo/IdleCallback.java
```diff
@@ -0,0 +1,14 @@
+package com.apollographql.apollo;
+
+/**
+ * Callback which gets invoked when the {@link ApolloClient} transitions
+ * from active to idle state.
+ */
+public interface IdleCallback {
```
sav007: Let's make it generic and call `IdleResourceCallback`.
Plus replace `{@link ApolloClient}` to `resource` in java docs

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
VisheshVadhera: Alright.

apollo-idling-resource/src/main/java/com/apollographql/apollo/espresso/ApolloIdlingResource.java
VisheshVadhera: Making this class a part of any other module would have meant that users who don't want an idling resource would also end up getting one. So I felt it's better to keep it in a module of it's own, so that whoever wants an ApolloIdlingResource can simply pull in the dependency from maven.

apollo-runtime/src/main/java/com/apollographql/apollo/IdleCallback.java
VisheshVadhera: Sure

apollo-idling-resource/src/main/java/com/apollographql/apollo/espresso/ApolloIdlingResource.java
marwanad: I feel it would be better to have this be in a separate repo and linked to from apollo-android. What do you think?

VisheshVadhera: Hmm, we can move it to a separate repo which is a part of the apollo-graphql project, probably because it's a good to have feature, but not an absolutely necessary one. Is that what you also think? I don't know what others feel though.

apollo-idling-resource/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=apollo-idling-resource
```
sav007: should we call it then `apollo-espresso-support`

apollo-idling-resource/src/main/java/com/apollographql/apollo/espresso/ApolloIdlingResource.java
```diff
@@ -0,0 +1,61 @@
+package com.apollographql.apollo.espresso;
```
sav007: `com.apollographql.apollo.test.espresso` ?

apollo-idling-resource/gradle.properties
```diff
@@ -0,0 +1,4 @@
...
+POM_NAME=Apollo GraphQL Idling Resource
```
sav007: Apollo GraphQL Espresso Support

apollo-idling-resource/gradle.properties
VisheshVadhera: Hmm, yeah adding espresso makes it clear that this module is about supporting espresso testing. `apollo-idling-resource` sounds vague.

apollo-idling-resource/src/main/java/com/apollographql/apollo/espresso/ApolloIdlingResource.java
VisheshVadhera: Right

apollo-idling-resource/gradle.properties
VisheshVadhera: Done

VisheshVadhera:
@sav007 I think it makes sense to publish jars only as we don't really have any resources in our android  modules. Should I open a new issue, so that this can be tackled in another PR?

VisheshVadhera:
Ok, looks like the build failed because the CI server was unable to download guava :| 

`A problem occurred configuring project ':apollo-android-support'.
> Could not resolve all files for configuration ':apollo-android-support:classpath'.
   > Could not download guava.jar (com.google.guava:guava:18.0)
      > Could not get resource 'https://jcenter.bintray.com/com/google/guava/guava/18.0/guava-18.0.jar'.
         > Response 304: Not Modified has no content!`

sav007:
@digitalbuddha @BenSchwab @marwanad any comments?

marwanad:
I'll take a look tomorrow.

apollo-espresso-support/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=apollo-espress-support
```
marwanad: espress**o**

apollo-espresso-support/gradle.properties
```diff
@@ -0,0 +1,4 @@
...
+POM_NAME=Apollo GraphQL Espresso Support
+POM_DESCRIPTION=An EspressoIdlingResource for Apollo GraphQL Client.
```
marwanad: `EspressoIdlingResource` -> `Espresso Idling Resource`?

apollo-espresso-support/src/main/java/com/apollographql/apollo/test/espresso/ApolloIdlingResource.java
```diff
@@ -0,0 +1,61 @@
...
+import com.apollographql.apollo.IdleResourceCallback;
+
+import javax.annotation.Nonnull;
+
+import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
+
+/**
+ * An Espresso {@link IdlingResource} for {@link com.apollographql.apollo.ApolloClient}.
+ */
+public class ApolloIdlingResource implements IdlingResource {
```
marwanad: make final?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -0,0 +1,175 @@
...
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * ApolloCallTracker is responsible for keeping track of running {@link ApolloCall} & {@link ApolloPrefetch} objects.
+ */
+public final class ApolloCallTracker {
+
+  private IdleResourceCallback idleResourceCallback;
```
marwanad: Do we really need a class for that? Can't we just take a `Runnable`?

sav007: I would not use Runnable for this as this is really callback not some runnable action that we need to run.

apollo-espresso-support/gradle.properties
VisheshVadhera: Alright

apollo-espresso-support/gradle.properties
VisheshVadhera: Nice catch!

apollo-espresso-support/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
VisheshVadhera: Not sure if the top level .gitignore ignores the module specific build directories, so yeah this is indeed required.

VisheshVadhera:
@marwanad Well that was the first thought which came to my mind to combine dispatcher and tracker in one class, but decided against it, because I wanted to keep the concerns separate and have a class whose only concern is to track calls and do nothing more.

sav007:
@VisheshVadhera I see all tests related to cache are failing could you pls fix that we can merge this PR

VisheshVadhera:
On it.

VisheshVadhera:
@sav007 Fixed the failing tests. Feel free to merge this PR.

Issue #467, opened 2017-04-28
Add imperative store API


martinbonnin:
Add something similar to what's described there: https://dev-blog.apollodata.com/apollo-clients-new-imperative-store-api-6cb69318a1e3.

The main use case is to be able to update the cache when a mutation is made. It could be complicated so we might want smaller separate tickets here but didn't find one already so I'm using this one to start the discussion.

martinbonnin:
This might or might not be linked to https://github.com/apollographql/apollo-android/issues/452

sav007:
Hey @martijnwalraven does iOS provide this API `readQuery()`, `readFragment()`, `writeQuery()`, and `writeFragment()`?

sav007:
@BenSchwab @digitalbuddha @marwanad 

After a lot of thoughts regarding `writeQuery/writeFragment` to the normalized store here is the proposal.

To be able to write to the store we need to use the same `ResponseNormalizer` that we are using for reading and parsing JSON response in `ResponseReader`. To use `ResponseNormalizer` we need to get access to generated schema `Field` descriptors the same used  in `Mapper` classes. 
So I suggest to do the similar what we did for mappers and update generated models to have following:

First introduce new interface:

```java
public interface ResponseSerializer {

  void serializeScalarField(Field field, Object value);

  void willSerializeObjectField(Field objectField, Map<String, Object> fieldValues);

  void didSerializeObjectField(Field objectField);

  void didSerializeList(List array);

  void willSerializeListElement(int index);

  void didSerializeListElement(int index);

}
```

 Introduce method in `Operation.Data` and `GraphqlFragment`:

```java
interface Data {
    void serialize(ResponseSerializer serializer);
}

public interface GraphqlFragment {
  void serialize(ResponseSerializer serializer);
}
```
Generate implementation of this `serialize` method that will take the value of field in generated model and pass it to `ResponseSerializer`.

The implementation of `ResponseSerializer` internally will delegate calls to `ResponseNormalizer` and build record set.

The implementation of `writeOperation` will be smth like this:
```java
public final class RealApolloStore implements ApolloStore, ReadableStore, WriteableStore {
  
  public void writeOperation(Operation operation) {
    CacheResponseSerializer responseSerializer = new CacheResponseSerializer(responseNormalizer);
    operation.serialize(responseSerializer);
    merge(responseSerializer.records());
  }

}
```

sav007:
[...]

PR #480, opened 2017-05-03
Read operation/fragment from store

sav007:
Add read function to read operations from ApolloStore.
Prepare class generation for reading fragments from store.
Add function to read fragment from store.

Part of #467


apollo-api/src/main/java/com/apollographql/apollo/api/Fragment.java
```diff
@@ -0,0 +1,7 @@
+package com.apollographql.apollo.api;
+
+/**
+ * Represents a GraphQL fragment
+ */
+public interface Fragment {
```
digitalbuddha: maybe name this `GraphqlFragment` so that autoimport in android will be happier

BenSchwab: +1

PR #482, opened 2017-05-04
Update apollo-codegen last version

gsancho:


marwanad:
Did you run `gradle updateTestIRFiles` after? This task will update our test fixtures if something changed with the new codegen version.


You'll need to have apollo-codegen globally installed, `npm install -g apollo-codegen@0.10.13` and then run `gradle apollo-compiler:updateTestIRFiles` from the project root.

sav007:
Just checked seems new version doesn't introduce changes to generated models. We can merge this PR.

marwanad:
We might want to bump the version number [here](https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/apollo-codegen.properties). I can do it and merge later.

sav007:
@marwanad that version is generated by when you run updateTestIRFiles. But yeah we have to update it as well

Issue #466, opened 2017-04-28
Sample app crashes when clicking an entry


martinbonnin:
`java.lang.NullPointerException: Attempt to invoke virtual method 'com.apollographql.apollo.sample.EntryDetailQuery$PostedBy com.apollographql.apollo.sample.EntryDetailQuery$Entry.postedBy()' on a null object reference`

the bug seems located there: https://github.com/apollographql/apollo-android/blob/50128ca46c6bbe79189dd69ef921c08acfe59d9a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java#L154 

`responseFieldMapper.map(responseReader)` will return an empty `Data` (data.entry == null) and not `null` so it will be considered a cache hit while everything is empty in there...

sav007:
dup of https://github.com/apollographql/apollo-android/issues/454

BenSchwab:
It crashes because of a bug I fix here: https://github.com/apollographql/apollo-android/pull/478/files

...but, this _is_ a real bug that this is considered a cache hit.

Now, according to the model definition, having a null `entry` is a valid state for the model. Hence, we resolve from the cache. This was an intentional design to enable requests that have never been executed, but have all relevant information to be resolved. (Useful for a master-child flow).

But I agree, in this case where the root piece of the query is nullable, it is nonsensical. It is especially a problem with our current fetch strategies. With `CACHE_FIRST` as it means that this request will never go to the network. 

I think we should consider storing all requests under hash of its `queryDocument`, or less aggressively its name (this would help when a nullable field was added in a new version, which would then still resolve). So instead of `[QUERY_ROOT -> Entry]` we have  `[QUERY_ROOT -> EntryDetailQuery -> Entry]`. 

Then we can rely on another mechanism, maybe such as imperative store api (https://github.com/apollographql/apollo-android/issues/467) for the master - child flow.

I'm guessing that iOS has this bug as well @martijnwalraven @kompfner.

thoughts, @digitalbuddha @sav007 @marwanad

PR #484, opened 2017-05-05
Don't allow request to proceed from network/cache if fields missing

BenSchwab:
Closes https://github.com/apollographql/apollo-android/issues/466

There is currently a bug where if a request has all `@nullable` fields and has never been executed before, and is executed with CACHE_FIRST, we will resolve null from cache and never go to network.

Per the graphql spec, requested fields should never be missing -- only null. Additionally, we should respect this in the cache, and store null when the server returns null.

As this is part of the GQL spec, if the server returns missing fields, we should throw an exception.

If there are missing fields when reading from the cache, this is a cache miss. This is an important case because if all the objects in a query are nullable, the cache needs to be able to differentiate between valid nullability, and nullability because a request has never been executed.

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/IntegrationTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/IntegrationTest.java
@@ -191,14 +191,12 @@ public class IntegrationTest {
     latch.awaitOrThrowWithTimeout(TIME_OUT_SECONDS, TimeUnit.SECONDS);
   }
 
-  @Test public void dataEmpty() throws Exception {
+  @Test(expected = ApolloException.class) public void dataEmpty() throws Exception {
     MockResponse mockResponse = mockResponse("ResponseDataEmpty.json");
     server.enqueue(mockResponse);
 
     ApolloCall<HeroName.Data> call = apolloClient.newCall(new HeroName());
-    Response<HeroName.Data> body = call.execute();
-    assertThat(body.data()).isNotNull();
-    assertThat(body.hasErrors()).isFalse();
+    call.execute();
   }
 
   @Test public void dataNull() throws Exception {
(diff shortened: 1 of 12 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/field/CacheFieldValueResolver.java
```diff
@@ -38,7 +40,7 @@ public CacheFieldValueResolver(ReadableCache readableCache, Operation.Variables
     }
   }
 
-  private Record valueFor(Record record, Field.ObjectField field) {
+  private Record valueFor(Record record, Field.ObjectField field) throws IOException {
```
BenSchwab: IOException or ApolloException? -- I choose IOException because that is what our ResponseReader currently expects.

apollo-integration/src/test/java/com/apollographql/apollo/IntegrationTest.java
```diff
@@ -211,6 +209,14 @@
     assertThat(body.hasErrors()).isFalse();
   }
 
+  @Test(expected = ApolloException.class) public void fieldMissing() throws Exception {
```
BenSchwab: @sav007 I changed this to be an exception. This should never happen per the GQL spec, so seems "exceptional".

BenSchwab:
Also, this is how iOS operates: https://github.com/apollographql/apollo-ios/blob/b3cf1c6252caf96d87884ef7d070653cdc9740bc/SharedTests/LoadQueryFromStoreTests.swift#L46

apollo-integration/src/test/java/com/apollographql/apollo/IntegrationTest.java
```diff
@@ -191,14 +191,12 @@
     latch.awaitOrThrowWithTimeout(TIME_OUT_SECONDS, TimeUnit.SECONDS);
   }
 
-  @Test public void dataEmpty() throws Exception {
+  @Test(expected = ApolloException.class) public void dataEmpty() throws Exception {
```
BenSchwab: @sav007 I changed this to be an exception. This should never happen per the GQL spec, so seems "exceptional".

apollo-runtime/src/main/java/com/apollographql/apollo/internal/reader/RealResponseReader.java
```diff
@@ -203,7 +203,7 @@ Boolean readBoolean(Field field) throws IOException {
     }
   }
 
-  @SuppressWarnings("unchecked") private <T> T readCustomType(Field.CustomTypeField field) {
+  @SuppressWarnings("unchecked") private <T> T readCustomType(Field.CustomTypeField field) throws IOException {
```
sav007: Does any instruction below throw IOException?

BenSchwab: Yep, as an exception will now be thrown if the field does not exist in the recordSet.

PR #485, opened 2017-05-07
Refactor HttpCache

sav007:
Remove HttpCacheControl and introduce in HttpCachePolicy instead.
Simplify checking for stale and remove eviction strategy.
Fix issue for IOException with NETWORK_FIRST|NETWORK_BEFORE_STALE fetch strategy.
Rename classes to have Http prefix.


apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/http/Utils.java
```diff
@@ -84,16 +98,51 @@ static void copyResponseBody(Response response, Sink sink) throws IOException {
...
+    return servedDate == null || now - servedDate.getTime() > timeout;
+  }
+
+  private static FetchStrategy fetchStrategy(Request request) {
+    String fetchStrategyHeader = request.header(CACHE_FETCH_STRATEGY_HEADER);
+    if (fetchStrategyHeader == null || fetchStrategyHeader.isEmpty()) {
+      return null;
+    }
+
+    for (FetchStrategy fetchStrategy : FetchStrategy.values()) {
```
BenSchwab: replace with `FetchStrategy.valueof(fetchStrategyHeader)`?

sav007: valueof throws exception if it can't find enum const by name.
> IllegalArgumentException - if this enum type has no constant with the specified name

this one is safe

sav007:
@BenSchwab I know that normalized cache will be much greater than http and probably in most cases will replace it. You are right they are different, and we probably should stop trying to unify them.

But I don't see why not to keep them both. Here are couple points:
- http cache simpler (and it must be) than normalized one and believe some users will use only http cache
- we have prefetch logic based on http cache
- it can be used for network throttling, when user can limit sequential same queries to be sent to server

That is why I bring this PR try to make http cache API as simple as possible. Agree that we need to make normalized cache to be great and make it in 90% to be the number one choice for user.

BenSchwab:
I think ideally we will add prefetch to the normalized cache, and have robust enough fetch strategies  (with a TTL option for example) to achieve request throttling.

I'm fine with keeping the the HttpCache for the time being. Though, I think we should keep an eye on it as the NormalizedCache matures, and as we get feedback on if people are finding the whole NormalizedCaching ecosystem valuable.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/http/HttpCachePolicy.java
```diff
@@ -0,0 +1,164 @@
...
+  public interface WriteOnlyFactory {
+    /**
+     * Obtain new write only cache policy. Cached response will be only written to cache but never read.
+     *
+     * @return obtained cache policy
+     */
+    @Nonnull HttpCachePolicy obtain();
+  }
+
+  public interface NeverExpireFactory {
```
sav007: How about this:

```java
public final class HttpCachePolicy {

  public static final ExpirePolicy CACHE_ONLY = new ExpirePolicy(HttpCacheFetchStrategy.CACHE_ONLY);

  public static final Policy NETWORK_ONLY = new Policy(HttpCacheFetchStrategy.NETWORK_ONLY, 0, null, false);

  public static final ExpirePolicy CACHE_FIRST = new ExpirePolicy(HttpCacheFetchStrategy.CACHE_FIRST);

  public static final ExpirePolicy NETWORK_FIRST = new ExpirePolicy(HttpCacheFetchStrategy.NETWORK_FIRST);

  public static class Policy {
    public final HttpCacheFetchStrategy fetchStrategy;
    public final long expireTimeout;
    public final TimeUnit expireTimeUnit;
    public final boolean expireAfterRead;

    Policy(HttpCacheFetchStrategy fetchStrategy, long expireTimeout, TimeUnit expireTimeUnit,
        boolean expireAfterRead) {
      this.fetchStrategy = fetchStrategy;
      this.expireTimeout = expireTimeout;
      this.expireTimeUnit = expireTimeUnit;
      this.expireAfterRead = expireAfterRead;
    }

    public long expireTimeoutMs() {
      if (expireTimeUnit == null) {
        return 0;
      }
      return expireTimeUnit.toMillis(expireTimeout);
    }
  }

  public static final class ExpirePolicy extends Policy {
    ExpirePolicy(HttpCacheFetchStrategy fetchStrategy) {
      super(fetchStrategy, 0, null, false);
    }

    private ExpirePolicy(HttpCacheFetchStrategy fetchStrategy, long expireTimeout, TimeUnit expireTimeUnit,
        boolean expireAfterRead) {
      super(fetchStrategy, 0, null, false);
    }

    public ExpirePolicy expireAfter(long expireTimeout, @Nonnull TimeUnit expireTimeUnit) {
      return new ExpirePolicy(fetchStrategy, expireTimeout, expireTimeUnit, expireAfterRead);
    }

    public ExpirePolicy expireAfterRead() {
      return new ExpirePolicy(fetchStrategy, expireTimeout, expireTimeUnit, true);
    }
  }
}
```
usage:

``` java
// cache only even if stale
HttpCachePolicy.CACHE_ONLY;
// cache only expire after timeout
HttpCachePolicy.CACHE_ONLY.expireAfter(2, TimeUnit.SECONDS)
// cache only stale expire after read
HttpCachePolicy.CACHE_ONLY.expireAfterRead()
// cache only check if not stale expire after read
HttpCachePolicy.CACHE_ONLY.expireAfter(2, TimeUnit.SECONDS).expireAfterRead()
... the same for CACHE_FIRST, NETWORK_FIRST

// network only - doesn't make sense to have any config
HttpCachePolicy.NETWORK_ONLY

// NETWORK_FIRST without expiration config is network before stale
//  so we don't need additional fetch strategy NETWORK_FIRST_BEFORE_STALE
HttpCachePolicy.NETWORK_FIRST
```

BenSchwab: Yeah, like this much better!

sav007:
@digitalbuddha mind to take a look at this PR?

PR #491, opened 2017-05-13
Prepare for query re-fetching

sav007:
Introduce separate interfaces for query and mutation calls.
Generate OperationName type for operations to be used as identifier for refetching.

Part of #452

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
@@ -5,8 +5,6 @@ import com.apollographql.android.impl.normalizer.HeroAndFriendsNamesWithIDs;
 import com.apollographql.android.impl.normalizer.type.Episode;
 import com.apollographql.apollo.api.Response;
 import com.apollographql.apollo.cache.normalized.CacheControl;
-import com.apollographql.apollo.cache.normalized.CacheKey;
-import com.apollographql.apollo.cache.normalized.CacheKeyResolver;
 import com.apollographql.apollo.cache.normalized.lru.EvictionPolicy;
 import com.apollographql.apollo.cache.normalized.lru.LruNormalizedCacheFactory;
 import com.apollographql.apollo.exception.ApolloException;
@@ -62,7 +60,7 @@ public class ApolloWatcherTest {
     EpisodeHeroName query = EpisodeHeroName.builder().episode(Episode.EMPIRE).build();
     server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
 
-    ApolloWatcher<EpisodeHeroName.Data> watcher = apolloClient.newCall(query).watcher();
+    ApolloQueryWatcher<EpisodeHeroName.Data> watcher = apolloClient.query(query).watcher();
     watcher.enqueueAndWatch(
         new ApolloCall.Callback<EpisodeHeroName.Data>() {
           @Override public void onResponse(@Nonnull Response<EpisodeHeroName.Data> response) {
@@ -84,7 +82,7 @@ public class ApolloWatcherTest {
 
     // Another newer call gets updated information
     server.enqueue(mockResponse("EpisodeHeroNameResponseNameChange.json"));
-    apolloClient.newCall(query).cacheControl(CacheControl.NETWORK_ONLY).execute();
+    apolloClient.query(query).cacheControl(CacheControl.NETWORK_ONLY).execute();
     secondResponseLatch.awaitOrThrowWithTimeout(TIME_OUT_SECONDS, TimeUnit.SECONDS);
     watcher.cancel();
   }
@@ -98,7 +96,7 @@ public class ApolloWatcherTest {
     EpisodeHeroName query = EpisodeHeroName.builder().episode(Episode.EMPIRE).build();
     server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
 
-    ApolloWatcher<EpisodeHeroName.Data> watcher = apolloClient.newCall(query).watcher();
+    ApolloQueryWatcher<EpisodeHeroName.Data> watcher = apolloClient.query(query).watcher();
     watcher.enqueueAndWatch(
         new ApolloCall.Callback<EpisodeHeroName.Data>() {
           @Override public void onResponse(@Nonnull Response<EpisodeHeroName.Data> response) {
@@ -118,7 +116,7 @@ public class ApolloWatcherTest {
     firstResponseLatch.awaitOrThrowWithTimeout(TIME_OUT_SECONDS, TimeUnit.SECONDS);
 
     server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
-    apolloClient.newCall(query).cacheControl(CacheControl.NETWORK_ONLY).enqueue(null);
+    apolloClient.query(query).cacheControl(CacheControl.NETWORK_ONLY).enqueue(null);
 
     // Wait 3 seconds to make sure no double callback.
     // Successful if timeout _is_ reached
@@ -135,7 +133,7 @@ public class ApolloWatcherTest {
     server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
     EpisodeHeroName query = EpisodeHeroName.builder().episode(Episode.EMPIRE).build();
 
-    ApolloWatcher<EpisodeHeroName.Data> watcher = apolloClient.newCall(query).watcher();
+    ApolloQueryWatcher<EpisodeHeroName.Data> watcher = apolloClient.query(query).watcher();
     watcher.enqueueAndWatch(
         new ApolloCall.Callback<EpisodeHeroName.Data>() {
           @Override public void onResponse(@Nonnull Response<EpisodeHeroName.Data> response) {
@@ -157,7 +155,7 @@ public class ApolloWatcherTest {
     HeroAndFriendsNamesWithIDs friendsQuery = HeroAndFriendsNamesWithIDs.builder().episode(Episode.NEWHOPE).build();
 
     server.enqueue(mockResponse("HeroAndFriendsNameWithIdsNameChange.json"));
-    apolloClient.newCall(friendsQuery).cacheControl(CacheControl.NETWORK_ONLY).execute();
+    apolloClient.query(friendsQuery).cacheControl(CacheControl.NETWORK_ONLY).execute();
 
     secondResponseLatch.awaitOrThrowWithTimeout(TIME_OUT_SECONDS, TimeUnit.SECONDS);
     watcher.cancel();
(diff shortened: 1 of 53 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
```diff
@@ -0,0 +1,44 @@
...
+   * <p>Sets a list of GraphQL query names to be re-fetched once this mutation completed.</p>
+   * In order to get queries to be re-fetched you must obtain {@link ApolloQueryWatcher} from provided list of
+   * queries before running this mutation.
+   *
+   * @param operationNames array of {@link OperationName} query names to be re-fetched
+   * @return {@link ApolloMutationCall} that will trigger re-fetching provided queries
+   */
+  @Nonnull ApolloMutationCall<T> refetchQueries(@Nonnull OperationName... operationNames);
+
+  @Nonnull @Override ApolloMutationCall<T> cacheControl(@Nonnull CacheControl cacheControl);
```
sav007: @BenSchwab do you think `cacheControl` makes sense for mutation as it defines fetch strategies and makes sense for queries only? 

Though `cacheHeaders` makes sense cos it defines if response for this mutation should be cached or not.

apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
```diff
@@ -62,7 +60,7 @@ public void testQueryWatcherUpdated_SameQuery_DifferentResults() throws IOExcept
     EpisodeHeroName query = EpisodeHeroName.builder().episode(Episode.EMPIRE).build();
     server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
 
-    ApolloWatcher<EpisodeHeroName.Data> watcher = apolloClient.newCall(query).watcher();
+    ApolloQueryWatcher<EpisodeHeroName.Data> watcher = apolloClient.query(query).watcher();
```
digitalbuddha: can you watch anything other than a query?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -157,6 +153,26 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
     }
   }
 
+  private <D extends Operation.Data, T, V extends Operation.Variables> RealApolloCall<T> newCall(
+      @Nonnull Operation<D, T, V> operation) {
+    ResponseFieldMapper responseFieldMapper = responseFieldMapper(operation);
+    return new RealApolloCall<T>(operation, serverUrl, httpCallFactory, httpCache, defaultHttpCachePolicy, moshi,
```
digitalbuddha: this constructor is wild! is there a way we can encapsulate somehow.  Of the top of my head I see 2 options: builder or organization of params into a few objects.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -157,6 +153,26 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
...
+        responseFieldMapper, customTypeAdapters, apolloStore, defaultCacheControl, defaultCacheHeaders,
+        dispatcher, logger, applicationInterceptors, Collections.<OperationName>emptyList());
+  }
+
+  private ResponseFieldMapper responseFieldMapper(Operation operation) {
+    ResponseFieldMapper responseFieldMapper;
+    synchronized (responseFieldMapperPool) {
+      responseFieldMapper = responseFieldMapperPool.get(operation.getClass());
+      if (responseFieldMapper == null) {
```
digitalbuddha: Sorry if this is out of scope for this pr but is there a way way can do a null object pattern or some other way to show absence without using `null`.  also missing nullability annotations

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
digitalbuddha: how does caching a mutation work?  Like what happens if you try same mutation again will you then get cached result instead of it hitting server?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Factory for creating {@link ApolloMutationCall} calls.
+   */
+  interface Factory {
+    /**
+     * Creates and prepares a new {@link ApolloMutationCall} call.
+     *
+     * @param query the operation which needs to be performed
+     * @return prepared {@link ApolloQueryCall} call to be executed at some point in the future
+     */
```
VisheshVadhera: We need to change the docs to:
/**
     * Creates and prepares a new {@link ApolloMutationCall} call.
     *
     * @param mutation the mutation which needs to be performed
     * @return prepared {@link ApolloMutationCall} call to be executed at some point in the future
     */

apollo-rx2support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
```diff
@@ -34,14 +34,14 @@ private Rx2Apollo() {
   }
 
   /**
-   * Converts an {@link ApolloWatcher} to an asynchronous Observable.
+   * Converts an {@link ApolloQueryWatcher} to an asynchronous Observable.
    *
    * @param watcher the ApolloWatcher to convert.
```
VisheshVadhera: We need to change it to:

 `@param watcher the ApolloQueryWatcher to convert.`

apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
```diff
@@ -35,27 +35,27 @@ private RxApollo() {
   }
 
   /**
-   * Converts an {@link ApolloWatcher} into an Observable. Honors the back pressure from downstream with the back
+   * Converts an {@link ApolloQueryWatcher} into an Observable. Honors the back pressure from downstream with the back
    * pressure strategy {@link rx.Emitter.BackpressureMode#LATEST}.
    *
    * @param watcher the ApolloWatcher to convert
```
VisheshVadhera: @param watcher the ApolloQueryWatcher to convert.

apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
```diff
@@ -35,27 +35,27 @@ private RxApollo() {
...
-  public static <T> Observable<Response<T>> from(@Nonnull final ApolloWatcher<T> watcher) {
+  public static <T> Observable<Response<T>> from(@Nonnull final ApolloQueryWatcher<T> watcher) {
     return from(watcher, Emitter.BackpressureMode.LATEST);
   }
 
   /**
-   * Converts an {@link ApolloWatcher} into an Observable.
+   * Converts an {@link ApolloQueryWatcher} into an Observable.
    *
    * @param watcher          the ApolloWatcher to convert
```
VisheshVadhera: `@param watcher the ApolloQueryWatcher to convert.`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
sav007: We shouldn't not hit the when we run mutations, but we should store the results to cache. So it's should WRITE ONLY. That's why I guess we need to remove `cacheControl` and just have `cacheHeaders` as they tells to save to cache or not

apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
sav007: Not now, we watch only Queries

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
BenSchwab: Yes, cacheControl does not make sense for mutations. It will always be NETWORK_ONLY. If we write to the cache can be controlled by `cacheHeaders`. It's generally important to write mutations results to the cache, as this is what triggers updates to `apolloWatchers`.

sav007:
@BenSchwab @digitalbuddha @marwanad any comments?

sav007:
@BenSchwab just checked we don;t allow to have the same names for queries even in separate packages:
```
There can be only one operation named "CheckoutByIdQuery".
There can be only one operation named "CheckoutByIdQuery".
```

But it even if we allow to have the same names, for refetch we are going to use `OperationName` object itself not string name:
```java
public final class TestQuery implements Query<TestQuery.Data, Optional<TestQuery.Data>, Operation.Variables> {
...

  private static final OperationName OPERATION_NAME = new OperationName() {
    @Override
    public String name() {
      return "TestQuery";
    }
  };

  @Override
  public OperationName name() {
    return OPERATION_NAME;
  }
```

So it will be unique for every query

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
```diff
@@ -0,0 +1,41 @@
...
+
+import javax.annotation.Nonnull;
+
+/**
+ * A call prepared to execute GraphQL mutation operation.
+ */
+public interface ApolloMutationCall<T> extends ApolloCall<T> {
+
+  /**
+   * <p>Sets a list of GraphQL query names to be re-fetched once this mutation completed.</p>
```
BenSchwab: Is this comment correct? Why do you need to obtain a ApolloQueryWatcher first?

sav007: We will use `ApolloQueryWatcher` to be notified when this mutation operation completed. So in order to listen for refetch user should have watcher first right?

BenSchwab: Well, we might want to refetch a query you aren't watching. Like if you have a post feed page, and a child page that creates a post, we might want to refetch the post feed query to update the normalized cache, even if we aren't watching

BenSchwab: Then if you happen to be watching the post page, and something did change, the watcher will be triggered when the refetched query is written to the normalized cache

BenSchwab: I was imagining you would just pass the actual queries to be refetched here instead of an identifier. 

Also, what would happen if I was listening to the same query that was executed twice with different variables. Seems like the tags would collide.

sav007: Yeah, makes sense. Tags can easily collide. So we just provide list of `Query` objects to be refetched, so the only open question how to handle errors, just ignore? Mutation finished with success, but one of the refetched query failed.

sav007: At the same time even if more than one watcher with the same name registered we will trigger them both

BenSchwab: Yeah, I think we would just have to ignore errors. Log them to our ApolloLogger of course.

Good point that we could just refetch all watchers with same tag. It's possible that we would be fetching more than the user actually wants, but if we batch the refetch, it doesn't seem like a big deal.

PR #494, opened 2017-05-15
Builder for no parameters

### LINKED ISSUE #462: Create builder even when no params.
To make the api less confusing lets create a builder even when no params are needed for a query. We should still keep the constructor as a path to creation.

VisheshVadhera:
Closes #462 

Quick question: Is there is a need to add `@Deprecated` to the constructor?


sav007:
No, no need to add @Deprecated.

VisheshVadhera:
Alright

PR #499, opened 2017-05-17
Add Java8 Optional generation

### LINKED ISSUE #496: Add support for using Java 8 Optional for nullable fields
Right now, `apollo-compiler` only supports the following [options for nullable types](https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/NullableValueType.kt):
```kotlin
  ANNOTATED("annotated"),
  APOLLO_OPTIONAL("apolloOptional"),
  GUAVA_OPTIONAL("guavaOptional");
```

We're using the compiler in a server-side project with access to Java 8, and it would be great to be able to use Java 8's `Optional` as well.

sav007:
Closes #496


apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/JavaTypeResolver.kt
```diff
@@ -31,6 +32,7 @@ class JavaTypeResolver(
       when (context.nullableValueType) {
         NullableValueType.APOLLO_OPTIONAL -> parameterizedOptional(javaType)
         NullableValueType.GUAVA_OPTIONAL -> parameterizedGuavaOptional(javaType)
+        NullableValueType.JAVA8_OPTIONAL -> parameterizedJava8Optional(javaType)
```
digitalbuddha: Nitpick I think you can call if just java optional so we don't need to change anything once java 9 comes out

PR #500, opened 2017-05-17
Remove unnecessary null check from equals/hashcode

### LINKED ISSUE #489: Unnecessary null check for a @Nonnull field causes FindBugs violation
In the generated code, field is marked `@NonNull` but in the `equals` method, that field is checked for null causing FindBugs to report violation.

```  public static class Data implements Operation.Data {
    private final @Nonnull Settings settings;

    public Data(@Nonnull Settings settings) {
      this.settings = settings;
    }

    public @Nonnull Settings settings() {
      return this.settings;
    }

    @Override
    public String toString() {
      return "Data{"
        + "settings=" + settings
        + "}";
    }

    @Override
    public boolean equals(Object o) {
      if (o == this) {
        return true;
      }
      if (o instanceof Data) {
        Data that = (Data) o;
        //field is checked for null below.
        return ((this.settings == null) ? (that.settings == null) : this.settings.equals(that.settings));
      }
      return false;
    }
```

sav007:
Closes #489

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/Util.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/android/compiler/Util.kt
@@ -123,9 +123,11 @@ fun TypeSpec.withEqualsImplementation(): TypeSpec {
               } else {
                 it.add("this.\$L == that.\$L", field.name, field.name)
               }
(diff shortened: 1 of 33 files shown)


digitalbuddha:
Lgtm

PR #504, opened 2017-05-20
Introduce explicit re-fetch query for mutation operation.

sav007:
Introduce API to explicit specify queries to be re-fetched after mutation completed.

Part of #452


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/lru/LruNormalizedCache.java
```diff
@@ -85,6 +87,8 @@
         return null;
       }
     } else {
+      Map tmp = new HashMap(lruCache.asMap());
+      tmp = tmp;
```
sav007: shit sorry

sav007:
@BenSchwab @digitalbuddha any comments?

Issue #498, opened 2017-05-16
Directives (@include, @skip) incorrectly marked as "Missing"


BenSchwab:
We currently throw an IO exception if a field is missing from the cache or json response.

```
 @SuppressWarnings("unchecked") private <T> T fieldValue(Record record, Field field) throws IOException {
    String fieldKey = field.cacheKey(variables);
    if (!record.hasField(fieldKey)) {
      throw new IOException("Missing value: " + field.fieldName());
    }
    return (T) record.field(fieldKey);
  }
```

This helps differentiate between missing and optional -- which is important to prevent normalized cache hits on requests that have not been executed.

however, directives are generated as optionals, and if the directive is false, is is not included in the json response which causes an IOException.

The quickest solution would be to not throw an IOExecption in our network value resolver:

```
public final class MapFieldValueResolver implements FieldValueResolver<Map<String, Object>> {

  @SuppressWarnings("unchecked") @Override public <T> T valueFor(Map<String, Object> map, Field field)
      throws IOException {
    if (!map.containsKey(field.responseName())) {
      throw new IOException("Missing value: " + field.responseName());
    }
    return (T) map.get(field.responseName());
  }
}
```

And to just store `null` for a value if it is missing from the response.

My only concern with this is if we think there is value in throwing an error on a missing field in the event a directive is not used, as this would indicate a problem with the server's GraphQL implementation.

sav007:
Yeah I don't know why we throwing exception in `MapFieldValueResolver` if we check if field is missing in `RealResponseReader#checkValue`:

```java
private void checkValue(Object value, boolean optional) {
    if (!optional && value == null) {
      throw new NullPointerException("corrupted response reader, expected non null value");
    }
  }
```

BenSchwab:
I wonder if our cache handle directives at all?

```
query Hero($episode: Episode, $withFriends: Boolean!) {
  hero(episode: $episode) {
    name
    friends @include(if: $withFriends) {
      name
    }
  }
}
```

My understanding is that this would make friends an optional. However, in the normalized cache if another query had fetched friends for that hero, it seems we would resolve the friends, even if $withFriends is false, as the omission logic is supposed to be controlled by the server.

kassim:
Is it important for us (the client) to distinguish the difference between a value that may be `null` and a value that may be missing?

[...]

PR #507, opened 2017-05-29
Fix missing field exception for Directives

kassim:
don’t throw IOException from MapFieldValueResolver when response doesn’t contain field

fixes https://github.com/apollographql/apollo-android/issues/498


sav007:
@kassim could you pls comment this test out: com.apollographql.apollo.IntegrationTest > dataEmpty

kassim:
done

PR #509, opened 2017-06-03
Add re-fetch query watchers

### LINKED ISSUE #452: Introduce RefetchQuerries for mutations

sav007:
Re-fetch query watchers after mutation completed
Refactor ApolloTracker.

Closes #452

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -2,7 +2,6 @@ package com.apollographql.apollo.internal;
 
 import com.apollographql.apollo.ApolloMutationCall;
 import com.apollographql.apollo.ApolloQueryCall;
-import com.apollographql.apollo.ApolloQueryWatcher;
 import com.apollographql.apollo.CustomTypeAdapter;
 import com.apollographql.apollo.api.Operation;
 import com.apollographql.apollo.api.OperationName;
@@ -67,7 +66,7 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
   final List<ApolloInterceptor> applicationInterceptors;
   final List<OperationName> refetchQueryNames;
   final List<Query> refetchQueries;
-  final Optional<QueryFetcher> queryFetcher;
+  final Optional<QueryReFetcher> queryReFetcher;
   final AtomicBoolean executed = new AtomicBoolean();
   volatile boolean canceled;
 
@@ -93,11 +92,12 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
     refetchQueryNames = builder.refetchQueryNames;
     refetchQueries = builder.refetchQueries;
     tracker = builder.tracker;
-    if (refetchQueries.isEmpty() || builder.apolloStore == null) {
-      queryFetcher = Optional.absent();
+    if ((refetchQueries.isEmpty() && refetchQueryNames.isEmpty()) || builder.apolloStore == null) {
+      queryReFetcher = Optional.absent();
     } else {
-      queryFetcher = Optional.of(QueryFetcher.builder()
+      queryReFetcher = Optional.of(QueryReFetcher.builder()
           .queries(builder.refetchQueries)
+          .queryWatchers(refetchQueryNames)
           .serverUrl(builder.serverUrl)
           .httpCallFactory(builder.httpCallFactory)
           .moshi(builder.moshi)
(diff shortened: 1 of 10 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -210,61 +211,69 @@ private RealApolloCall(Builder<T> builder) {
...
-        if (queryFetcher.isPresent()) {
-          queryFetcher.get().refetchAsync(new QueryFetcher.OnFetchCompleteCallback() {
-            @Override public void onFetchComplete() {
-              //noinspection unchecked
-              responseCallback.onResponse(response.parsedResponse.get());
-            }
-          });
-        } else { //noinspection unchecked
-          responseCallback.onResponse(response.parsedResponse.get());
+  private void refetchQueryWatchers() {
```
BenSchwab: Although strange, should this be `refetchRefetchQueryWatchers`?

I'm a little confused by this. We should never need to explicity call refetch on a query watcher, as the point of a query watcher is to refetch itself when a watched key changes.

We only want to refetch the expliciting set refetchQuerires.

BenSchwab: Maybe this goes back to the discussion we had a while ago, I've gotten a little lost, but do we force user's to be watching a query in able to refetch it?

I've also envisioned them to be separate concepts -- you can watch whatever you want, and whenever a dependent key changes you get notified. And when you mutate, you can pass a set of real query objects to refetch -- these will have no direct listeners, _but_ if you are watching somewhere, the normal watcher flow will trigger and notify.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -210,61 +211,69 @@ private RealApolloCall(Builder<T> builder) {
...
+            return;
+          }
+
+          refetchQueryWatchers();
+
+          if (queryFetcher.isPresent()) {
+            queryFetcher.get().refetchAsync(new QueryFetcher.OnFetchCompleteCallback() {
+              @Override public void onFetchComplete() {
+                //noinspection unchecked
+                originalCallback.onResponse(response.parsedResponse.get());
```
BenSchwab: Would it be safe to not block the original call on the refetch queries? I can see the argument for both ways, ideally if it blocked or not would be user configurable.

Additionally, the actually refetch routine seems to be serially executed. Maybe we should investigate just dumping all the refetch querries into one graphql query? Although is may complicate query whitelisting.

sav007:
@BenSchwab 

> Also, do we need to guard against an infinite loop of refetch queries? (Two refetch queries reference each other and ping pong back and forth refetching)

That should not be the case, as refetch queries names can be set only for `ApolloMutationCall`. At the same time you can't watch `ApolloMutationCall` only query can be watched. So we are safe here.

> Would it be safe to not block the original call on the refetch queries? I can see the argument for both ways, ideally if it blocked or not would be user configurable.

Yeah I thought about this and don't have strong opinion on that so not sure what the right way.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
sav007: Sorry what do you mean?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
sav007: > Maybe we should investigate just dumping all the refetch querries into one graphql query? 

Yeah that will be awesome, we def need to explore this, I see though as it won't be trivial. At the same time we can explore overall merging queries (smth like throttling ) into one call.

stubailo:
> Maybe we should investigate just dumping all the refetch queries into one graphql query? Although is may complicate query whitelisting.

A better approach would be to use transport batching probably (sending multiple documents/query IDs in one request), that way you can still whitelist, but they can be sent in one request and batched on the server.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
BenSchwab: 'refetchQueryWatchers` sort of makes it seem like we will be refetching all query watchers - but really we are refetching the set of "refetch" query watchers, right?

It's an internal method, so not a big deal at all.

sav007:
@digitalbuddha any comments?

PR #530, opened 2017-06-18
Remove unused catch blocks.

NightlyNexus:
Writing to a Buffer does no I/O.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
@@ -62,12 +62,7 @@ import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
 
   @Override @Nonnull public InterceptorResponse intercept(Operation operation, ApolloInterceptorChain chain)
       throws ApolloException {
-    try {
-      httpCall = httpCall(operation);
-    } catch (IOException e) {
-      logger.e(e, "Failed to prepare http call");
-      throw new ApolloNetworkException("Failed to prepare http call", e);
-    }
+    httpCall = httpCall(operation);
 
     try {
       return new InterceptorResponse(httpCall.execute());
@@ -82,13 +77,7 @@ import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
       @Nonnull ExecutorService dispatcher, @Nonnull final CallBack callBack) {
     dispatcher.execute(new Runnable() {
       @Override public void run() {
-        try {
-          httpCall = httpCall(operation);
-        } catch (IOException e) {
-          logger.e(e, "Failed to prepare http call");
-          callBack.onFailure(new ApolloNetworkException("Failed to prepare http call", e));
-          return;
-        }
+        httpCall = httpCall(operation);
 
         httpCall.enqueue(new Callback() {
           @Override public void onFailure(Call call, IOException e) {
@@ -112,7 +101,7 @@ import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
     this.httpCall = null;
   }
 
-  private Call httpCall(Operation operation) throws IOException {
+  Call httpCall(Operation operation) {
     RequestBody requestBody = httpRequestBody(operation);
     Request.Builder requestBuilder = new Request.Builder()
         .url(serverUrl)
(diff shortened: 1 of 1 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -112,7 +101,7 @@ public void interceptAsync(@Nonnull final Operation operation, @Nonnull final Ap
     this.httpCall = null;
   }
 
-  private Call httpCall(Operation operation) throws IOException {
+  Call httpCall(Operation operation) {
```
NightlyNexus: unrelated visibility change obviates the need for a synthetic accessor method. will revert if unwanted.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -134,16 +123,24 @@ private Call httpCall(Operation operation) throws IOException {
...
   }
 
-  public static String cacheKey(RequestBody requestBody) throws IOException {
+  public static String cacheKey(RequestBody requestBody) {
     Buffer hashBuffer = new Buffer();
-    requestBody.writeTo(hashBuffer);
+    try {
+      requestBody.writeTo(hashBuffer);
+    } catch (IOException e) {
+      throw new AssertionError(e);
```
sav007: why do we need IOException to be wrapped into Assert exception?

sav007:
Sorry I don't understand why we convert IOException to AssertException. Could you please elaborate what are you trying to fix here?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
sav007: Never mind I think I got what you mean

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -134,16 +123,24 @@ private Call httpCall(Operation operation) throws IOException {
...
   }
 
-  private RequestBody httpRequestBody(Operation operation) throws IOException {
+  private RequestBody httpRequestBody(Operation operation) {
     JsonAdapter<Operation> adapter = new OperationJsonAdapter(moshi);
     Buffer buffer = new Buffer();
-    adapter.toJson(buffer, operation);
+    try {
+      adapter.toJson(buffer, operation);
+    } catch (IOException e) {
```
sav007: Could we log in case of error, I know it should never happen but lets add logs here:
`logger.e(e, "Failed to prepare http call");`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -134,16 +123,24 @@ private Call httpCall(Operation operation) throws IOException {
...
     return RequestBody.create(MEDIA_TYPE, buffer.readByteString());
   }
 
-  public static String cacheKey(RequestBody requestBody) throws IOException {
+  public static String cacheKey(RequestBody requestBody) {
     Buffer hashBuffer = new Buffer();
-    requestBody.writeTo(hashBuffer);
+    try {
+      requestBody.writeTo(hashBuffer);
+    } catch (IOException e) {
```
sav007: Could we log in case of error, I know it should never happen but lets add logs here:
`logger.e(e, "Failed to prepare http call");`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
NightlyNexus: I'd rather not add ambiguity to code by adding cases that can't happen.
If you'd like the change, I'll close this, and any other contributor is free to copy-paste this and modify it.

NightlyNexus:
Feel free to reuse the code.

sav007:
@NightlyNexus why this PR was closed?

PR #532, opened 2017-06-23
Generate response fields outside of response mapper.

sav007:
Response fields are going to be shared between response mapper and marshaller, make them generated under the type as static array instead being part of response mapper.
Rename `api.Field` to `api.ResponseField`

Part of #467


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ResponseFieldSpec.kt
```diff
@@ -0,0 +1,293 @@
...
+class ResponseFieldSpec(
+    val irField: Field,
+    val fieldSpec: FieldSpec,
+    val responseFieldType: ResponseField.Type,
+    val typeConditions: List<String> = emptyList(),
+    val context: CodeGenerationContext
+) {
+
+  fun factoryCode(): CodeBlock {
+    val factoryMethod = FACTORY_METHODS[responseFieldType]!!
```
digitalbuddha: sanity check you wanted to do !! rather than
```
return FACTORY_METHODS[responseFieldType]?.apply{when(){...}}
```
I might have trouble following logic but wanted to confirm it is not possible that FACTORY_METHODS[responseFieldType] nulls out

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ResponseFieldSpec.kt
```diff
@@ -0,0 +1,293 @@
...
+        .beginControlFlow("if (\$LStr != null)", fieldSpec.name)
+        .addStatement("\$L = \$T.valueOf(\$LStr)", fieldSpec.name, fieldSpec.type, fieldSpec.name)
+        .nextControlFlow("else")
+        .addStatement("\$L = null", fieldSpec.name)
+        .endControlFlow()
+        .build()
+    return CodeBlock
+        .builder()
+        .addStatement("final \$T \$LStr = \$L.\$L(\$L[\$L])", ClassNames.STRING, fieldSpec.name, READER_VAR,
+            READ_METHODS[ResponseField.Type.STRING]!!, fieldsVarName, index)
```
digitalbuddha: verifying !! is intentional

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ResponseFieldSpec.kt
sav007: we can't have `factoryMethod = null`, otherwise this is our error (we forgot to add mapping response field type to appropriate factory method)  and we need to crash here.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ResponseFieldSpec.kt
sav007: the same as above

digitalbuddha:
Thanks! Lgtm

PR #540, opened 2017-07-01
Generate verification to ensure nonnull values for model constructor

sav007:
Because of the fact we allow now user to create instances of generated models we must ensure that null values are not accepted for non-null fields

Part of #467


digitalbuddha:
+1

PR #541, opened 2017-07-02
Write fragments to the Apollo Store API

sav007:
Part of #467

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
@@ -381,7 +381,6 @@ public class ResponseWriteTestCase {
     assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().name()).isEqualTo("Leia Organa");
 
     apolloClient.apolloStore().write(
-        query,
         new HeroWithFriendsFragment(
             hero.fragments().heroWithFriendsFragment().__typename(),
             hero.fragments().heroWithFriendsFragment().id(),
@@ -408,15 +407,16 @@ public class ResponseWriteTestCase {
                     )
                 )
             )
-        ), CacheKey.from(hero.fragments().heroWithFriendsFragment().id()));
+        ), CacheKey.from(hero.fragments().heroWithFriendsFragment().id()), query.variables()
+    );
 
     apolloClient.apolloStore().write(
-        query,
         new HumanWithIdFragment(
             "Human",
             "1002",
             "Beast"
-        ), CacheKey.from("1002"));
+        ), CacheKey.from("1002"), query.variables()
+    );
 
     hero = apolloClient.query(query).cacheControl(CacheControl.CACHE_ONLY).execute().data().hero();
     assertThat(hero.__typename()).isEqualTo("Droid");
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
@@ -1,6 +1,7 @@
 package com.apollographql.apollo;
 
 import com.apollographql.apollo.cache.normalized.CacheControl;
+import com.apollographql.apollo.cache.normalized.CacheKey;
 import com.apollographql.apollo.cache.normalized.lru.EvictionPolicy;
 import com.apollographql.apollo.cache.normalized.lru.LruNormalizedCacheFactory;
 import com.apollographql.apollo.exception.ApolloException;
@@ -257,7 +258,7 @@ public class ResponseWriteTestCase {
                 "R222-D222",
                 asList(
                     new HeroWithFriendsFragment.Friend(
-                        "Droid",
+                        "Human",
                         new HeroWithFriendsFragment.Friend.Fragments(
                             new HumanWithIdFragment(
                                 "Human",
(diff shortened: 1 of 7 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStore.java
```diff
@@ -134,9 +134,39 @@
    * @param <D>           type of GraphQL operation data
    * @param <T>           type operation cached data will be wrapped with
    * @param <V>           type of operation variables
+   * @return set of keys of {@link Record} which have changed
    */
-  <D extends Operation.Data, T, V extends Operation.Variables> void write(@Nonnull Operation<D, T, V> operation,
-      @Nonnull D operationData);
+  @Nonnull <D extends Operation.Data, T, V extends Operation.Variables> Set<String> write(
+      @Nonnull Operation<D, T, V> operation, @Nonnull D operationData);
```
sav007: operation needed to resolve variables

sav007:
@BenSchwab @digitalbuddha any comments here?

apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
```diff
@@ -356,4 +357,79 @@ public void operation_with_inline_fragments() throws Exception {
...
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(0).fragments().humanWithIdFragment().name()).isEqualTo("Luke Skywalker");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(1).fragments().humanWithIdFragment().__typename()).isEqualTo("Human");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(1).fragments().humanWithIdFragment().id()).isEqualTo("1002");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(1).fragments().humanWithIdFragment().name()).isEqualTo("Han Solo");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().__typename()).isEqualTo("Human");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().id()).isEqualTo("1003");
+    assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().name()).isEqualTo("Leia Organa");
+
+    apolloClient.apolloStore().write(
+        new HeroWithFriendsFragment(
```
digitalbuddha: thoughts on a builder? With more than 3 params it would be tough to see what corresponds to what.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStore.java
```diff
@@ -134,9 +134,43 @@
...
+  /**
+   * Write operation to the store and publish changes of {@link Record} which have changed.
+   *
+   * @param operation     {@link Operation} response data of which should be written to the store
+   * @param operationData {@link Operation.Data} operation response data to be written to the store
+   * @param <D>           type of GraphQL operation data
+   * @param <T>           type operation cached data will be wrapped with
+   * @param <V>           type of operation variables
+   */
+  <D extends Operation.Data, T, V extends Operation.Variables> void writeAndPublish(
```
digitalbuddha: sorry if I missed something, I'm unclear what publish is.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
```diff
@@ -225,23 +225,63 @@ public RealApolloStore(@Nonnull NormalizedCache normalizedCache, @Nonnull CacheK
...
           operationData.marshaller().marshal(cacheResponseWriter);
-          Collection<Record> records = cacheResponseWriter.normalize(networkResponseNormalizer());
-          merge(records, CacheHeaders.NONE);
+          ResponseNormalizer<Map<String, Object>> responseNormalizer = networkResponseNormalizer();
+          responseNormalizer.willResolveRootQuery(operation);
+          Collection<Record> records = cacheResponseWriter.normalize(responseNormalizer);
+          return merge(records, CacheHeaders.NONE);
+        } catch (IOException e) {
+          logger.e(e, "Failed to write operation data to the store");
+          throw new RuntimeException(e);
```
digitalbuddha: maybe a different exception type, I can't imagine every app wanting to crash or swallow all RuntimeExceptions when there is a failed write.  Makes me also think error policy would be nice.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
```diff
@@ -225,23 +225,63 @@ public RealApolloStore(@Nonnull NormalizedCache normalizedCache, @Nonnull CacheK
...
+        }
+      }
+    });
+  }
+
+  @Override public <D extends Operation.Data, T, V extends Operation.Variables> void writeAndPublish(
+      @Nonnull Operation<D, T, V> operation, @Nonnull D operationData) {
+    Set<String> changedKeys = write(operation, operationData);
```
digitalbuddha: next method you checkNotNull even when nullability annotation exists. Don't particular care if there is the double check as long as it is consistent.

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStore.java
sav007: It publishes keys that will trigger any watchers depended on these keys

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
sav007: It's kind of always the question: do you need to swallow internal error or expose it to user. I would delegate that to user decision, maybe annotate this method that it can throw IOException? To make it explicitly

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
sav007: `@Nonnull` is just for syntax highlighting in IDE, but it doesn't mean user won't pass null as argument, and I think is better to fail first than fail later with unpredictable outcome.

apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
sav007: Yeah I feel we need to generate builder as well, but now question is about method counts. Won't it be too much for one generated model?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
sav007: Actually @digitalbuddha there is next PR that is coming to fix IOException throws mess.
As we moved away from streaming API to read json data there is no need to have readers / writers API to throw IOException as we don't use any IO.

apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
digitalbuddha: Good point can leave for now

apollo-runtime/src/main/java/com/apollographql/apollo/internal/cache/normalized/RealApolloStore.java
digitalbuddha: So add nonNull here?

digitalbuddha: Oops sorry meant maybe add null check here

Issue #551, opened 2017-07-07
Add support for semantic naming


BenSchwab:
on iOS a query like: 
```
query AllStarships {
  allStarships(first: 7) {
    edges {
      node {
        ...starshipFragment
      }
    }
  }
}
```
Would generate a file like class called `AllStarshipsQuery`. This makes sense because the generated object is a thing/noun. when writing the graphql, it often makes sense to omit the Query suffix because the language requires a `query` identifier before the name.

This is important because when we want to whitelist a query iOS may write: 
```
query AllStarships {
}
```
and Android currently may write 
```
query AllStarshipsQuery {
}
```
to generate the same code. However, the whitelist hash would be different.

Not everyone may want to use this semantic naming, so I propose adding it as a gradle plugin flag:
`useSemanticNaming`

cc @kompfner

stubailo:
> However, the whitelist hash would be different.

Is it a problem if iOS and Android generate slightly different queries for you?

BenSchwab:
We haven't decided specifically if iOS and Android will share a common registry, but I could see how it would be convenient to easily check if a specific query is used in android/ios without worrying about different naming conventions

PR #553, opened 2017-07-07
Add option to append query and mutation to operation names

BenSchwab:
closes https://github.com/apollographql/apollo-android/issues/551

Also noticed that we had no tests for mutations in apollo compiler, so added one, and change the name of test files from `TestQuery.***` to `TestOperation.***`.


apollo-gradle-plugin/src/main/java/com/apollographql/apollo/gradle/ApolloExtension.java
```diff
@@ -12,6 +12,7 @@
   private Map<String, String> customTypeMapping = new LinkedHashMap<>();
   private String nullableValueType = NullableValueType.ANNOTATED.getValue();
   private boolean generateAccessors = true;
+  private boolean useSemanticNaming = false;
```
BenSchwab: This makes `useSemanticNaming` default to false. Wouldn't mind  making this true, if we want to default to same behavior as iOS.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationTypeSpecBuilder.kt
```diff
@@ -1,17 +1,29 @@
...
 
 class OperationTypeSpecBuilder(
     val operation: Operation,
-    val fragments: List<Fragment>
+    val fragments: List<Fragment>,
+    useSemanticNaming: Boolean
 ) : CodeGenerator {
-  private val OPERATION_TYPE_NAME = operation.operationName.capitalize()
-  private val DATA_VAR_TYPE = ClassName.get("", "$OPERATION_TYPE_NAME.Data")
+  private val OPERATION_TYPE_NAME: String
```
sav007: nit: not sure why we make these names to be capitalized initially, should be lower camel case I guess

sav007:
What about fragment, should we do the same for them, just for consistency.

apollo-gradle-plugin/src/main/java/com/apollographql/apollo/gradle/ApolloExtension.java
sav007: don't have any objections, I wish we could put this on vote :)

BenSchwab:
sure, I'll bump to 0.40. I decided to default to the same generated code as iOS - so it is a breaking change. However, by setting `useSemanticNaming = false`, it would not be a breaking change.

BenSchwab:
Also regarding not including fragments -- iOS doesn't append "Fragment" to the end of fragments. I think this is the way to go because we currently reference fragments through a `fragments` list, so the code reads clearly. For example: 
```
fragment comparisonFields on Character {
```
would be accessed something like `data.fragments().comparisonFields()` which makes sense.

Issue #432, opened 2017-04-18
Cache Fetch Strategy / Cache Header unification proposal


BenSchwab:
Apollo Client has two levels of caching:

- Response cache
- Normalized cache

Each level has their own `CacheControl` (`HttpCacheControl` and `CacheControl`, respectively). HttpCache also has a basic concept of response headers, and the normalized cache may need something similar.

Inherently, having two caches is complex, and many users may only choose to use one. However, the similarity but minor differences between the two cache types will cause confusion. Additionally, we have designed the Apollo caches to be user configurable. For this to be useful, it must be very clear what the responsibility of a custom cache implementation is. 

The first proposal is to separate "FetchStrategy" from "CacheHeaders".

`FetchStrategy` control where data is looked up. These will include:

 ``` 
CACHE_FIRST,

  CACHE_ONLY,

  NETWORK_FIRST,

  NETWORK_ONLY
```

`CacheHeaders` control *how* a response is cached/and or read. For example, a request may include 'offline-critical'. This indicates that this response should be kept around longer than non `offlineCritical` responses. Another common example may be `no-cache`. `no-cache` is different from a fetch strategy, as we may want to use it with any of the above strategies. The difference is with `no-cache` we will not write the response to the cache. We can canonize some of these as standard headers -- but in general we should keep these light. If someone has a specific use case, they can pass in that header, and implement a cache to handle it.

`CacheHeaders` are intentionally designed to be generic (Map of strings internally). This is because they operate on the user configurable cache. 

Conversely, `FetchStrategies` are designed to be closed. They operate in the main logic of ApolloClient, and cannot be configured by user.

One tricky concept with this design is the concept of staleness, as it bridges between `CacheHeaders` and `FetchStrategies`. A cache has every ability to determine staleness based on internal configuration, and `CacheHeaders`. Therefore, a cache will essentially return a tuple:
`<@nullable response, isStale>`.  To drive this home, all a specific cache implementation is is a function: `(CacheHeaders, key) --> <response, isStale>` 

Semantically, if a response isStale it should *not* be returned to the user as a response. However, in the case that the network fails, it is may be desired to return stale data. This logic must be baked into the core apollo client. Therefore, we can include a boolean flag with the fetchStrategy, such as `builder.setFetchStragegy(CACHE_FIRST, shouldUseStaleDataOnError)`.
 
[...]

PR #556, opened 2017-07-10
Change CacheControl to Interceptor based Fetcher

BenSchwab:
## Motivation

The interceptor architecture we use for executing a request is loosely similar to the future direction Apollo JS is going: https://github.com/apollographql/apollo-fetch. If I understand correctly, the vision is that Apollo Fetch will make it easy to write fetch policies like `CACHE_FIRST` or more complicated policies like `CACHE_FIRST_IF_NOT_STALE_BUT_IF_NO_NETWORK_ALLOW_STALE`.

However, shipping with every possible fetch strategy is impossible -- thus our current `CacheControl` approach based on a set of enums, could be made more flexible. This PR proposes making `CacheControl` an `ApolloInterceptor` that contains logic of how to fetch the request. (And changes the name to `Fetcher` to better align with JS). 

## Fetcher
- Just an `ApolloInterceptor`
- Uses `FetchOptions` to configure fetching preferences
- Can change `CacheHeaders` and a boolean `fetchFromCache` to control how the data will be fetched
- Thus CACHE_FIRST is implemented by first proceeding the chain with a FetchOptions with `fetchFromCache = true` and catching an exception. If exception it will try to proceed with `fetchFromCache = false`

### Benefits
- `ApolloCacheInterceptor` has less logic - all fetching logic is contained in the `Fetcher`
- Fairly easy to implement a custom Fetcher for specific needs
- In basic use case, API is no more complicated

```
 githuntFeedCall = application.apolloClient()
        .query(feedQuery)
        .fetcher(ApolloFetchers.CACHE_FIRST);
```

## Mutli-response
Notably this PR enables a fetcher to respond multiple times. The rationale behind this is that when you use Apollo with a persistent sql cache, currently the only fetch policy that works is `NETWORK_FIRST` which negates the value of the sql / in memory cache. This is because if you use `CACHE_FIRST`, after the first cache population, the data will never be automatically refreshed. This could be handled with some sort of "staleness" system, but again, this would result in the cache data always being held until it is stale. Rather, it would be useful to allow a request to fulfill from cache first, and then also, fetch updated data from the network.

This does complicate the callback API, particularly with ApolloRx, as we need to know when a request will return no more data (`onCompleted`).

This PR ships with one double response fetch policy `CACHE_AND_NETWORK`.

## Additional work
- [x] Integrate with ApolloSupport
- [x] Add tests
- [x] Add documentation
- [x] Make sure cancel and dispose works properly

Previous discussions: https://airbnb.quip.com/BVbnAztIGoLm

closes https://github.com/apollographql/apollo-android/issues/432


BenSchwab:
cc @kompfner @martijnwalraven for thoughts from iOS perspective

BenSchwab:
@sav007 @digitalbuddha Let me know what you think about going in this direction, if things seem good from a conceptual perspective, I'll move forward with writing all the tests, documentation, and code cleanup

sav007:
I love this idea, clean and flexible

apollo-runtime/src/main/java/com/apollographql/apollo/fetcher/Fetcher.java
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo.fetcher;
+
+import com.apollographql.apollo.interceptor.ApolloInterceptor;
+import com.apollographql.apollo.internal.util.ApolloLogger;
+
+public interface Fetcher {
+
+  ApolloInterceptor create(ApolloLogger logger);
```
sav007: should it be renamed for clarity to smth `provideInterceptor`?

apollo-runtime/src/main/java/com/apollographql/apollo/fetcher/Fetcher.java
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo.fetcher;
+
+import com.apollographql.apollo.interceptor.ApolloInterceptor;
+import com.apollographql.apollo.internal.util.ApolloLogger;
+
+public interface Fetcher {
```
sav007: Should be called `ResponseFetcher`?

digitalbuddha:
I like the concept but am not crazy with the builder method name 

   .fetcher(ApolloFetchers.CACHE_FIRST) feels weird. It's not a fetcher but a cache policy or a fetch policy. When  I see a builder method called fetcher I expect to pass in something that has an implementation for fetch.

digitalbuddha:
Ok next time I'll look at code and not just description of pr. It does fetch! I'm with Ivan love it.

apollo-runtime/src/main/java/com/apollographql/apollo/fetcher/CacheAndNetworkFetcher.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private synchronized void handleNetworkResponse(ApolloInterceptor.InterceptorResponse response) {
+    networkResponse = Optional.of(response);
+    dispatch();
+  }
+
+  private synchronized void handleNetworkError(ApolloException exception) {
+    networkException = Optional.of(exception);
+    dispatch();
```
sav007: should we consider to not dispatch if result if it was canceled (disposed)

sav007:
We need to make sure that `cancel` on `RealApolloCall` works as expected, all chained interceptors are been disposed

BenSchwab:
Divided this into runtime changes, rx changes, and test changes.

Also, took a stab at https://github.com/apollographql/apollo-android/issues/560 as this PR required me to address cancellation in the new fetchers.

There is a slight change in behavior of cancellation:

1) If cancel is called before `enqueue` or `execute`, there will be an exception
2) If cancel is called during `execute`, there will be an exception
3) If cancel is called during `enqueue` there will be _no_ exception, and callback will not be triggered

The change is in point (3) which would previously trigger onError. This felt weird to me, as if I call `cancel` I don't expect any callbacks (Rx follows this model).

To avoid the memory leaks https://github.com/apollographql/apollo-android/issues/560 during cancelation a made the callback a WeakReference.

I also more aggressively check if dispose has been called on interceptors to avoid doing unnecessary work.

apollo-runtime/src/main/java/com/apollographql/apollo/fetcher/ApolloResponseFetcher.java
```diff
@@ -0,0 +1,47 @@
+package com.apollographql.apollo.fetcher;
+
+import com.apollographql.apollo.internal.fetcher.CacheFirstFetcher;
+import com.apollographql.apollo.internal.fetcher.CacheOnlyFetcher;
+import com.apollographql.apollo.internal.fetcher.CacheAndNetworkFetcher;
+import com.apollographql.apollo.internal.fetcher.NetworkFirstFetcher;
+import com.apollographql.apollo.internal.fetcher.NetworkOnlyFetcher;
+
+public final class ApolloResponseFetcher {
```
sav007: nit: should be named `ApolloResponseFetchers`, like in java `Executors`, in Rx `Schedulers`

apollo-runtime/src/main/java/com/apollographql/apollo/interceptor/FetchOptions.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Immutable options to configure how a {@link ApolloInterceptorChain} will fetch a response.
+ */
+public final class FetchOptions {
+
+  public final boolean fetchFromCache;
+  public final CacheHeaders cacheHeaders;
+
+  public static final FetchOptions NETWORK = new FetchOptions(false, CacheHeaders.NONE);
```
sav007: nit: `NETWORK_ONLY`?

sav007:
Just curios if `FetchOptions` is really necessary, is behaviour inside the chained interceptors provided by `ResponseFetcher` is not enough?

BenSchwab:
While, I believe it would be possible to write a custom Fetcher (ApolloInterceptor) that could resolve from network / cache, ideally user's can just proceed with the combo of CacheInterceptor/ParseInterceptor/ServerIntercept that we already assemble. It seemed to me the easiest way to pass a configuration to `CacheInterceptor` to instruct it how to proceed is an options object with `fetchFromCache` and `CacheHeaders` both of which many custom fetch strategies would change to achieve different effects (adding a CacheHeader like "ALLOW_STALE" after an initial cache and network miss)

sav007:
We should update java doc for Callback: `public abstract void onResponse(@Nonnull Response<T> response);` to mention that this can be called multiple times, user should be prepared for this

sav007:
And we need to mention that `Callback.onCompleted` won't be called in case of error

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -43,15 +43,16 @@
   void enqueue(@Nullable Callback<T> callback);
 
   /**
-   * Sets the {@link CacheControl} strategy for an ApolloCall object.
+   * Sets the {@link ResponseFetcher} strategy for an ApolloCall object.
    *
-   * @param cacheControl the CacheControl strategy to set
+   * @param fetcher the {@link ResponseFetcher} to use.
    * @return The ApolloCall object with the provided CacheControl strategy
```
sav007: `@return The ApolloCall object with the provided CacheControl strategy` -> 
`@return The ApolloCall object with the provided fetch strategy`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -334,24 +334,24 @@ public Builder defaultHttpCachePolicy(@Nonnull HttpCachePolicy.Policy cachePolic
...
-    public Builder defaultCacheControl(@Nonnull CacheControl cacheControl) {
-      this.defaultCacheControl = checkNotNull(cacheControl, "cacheControl == null");
+    public Builder defaultCacheHeaders(@Nonnull CacheHeaders cacheHeaders) {
+      this.defaultCacheHeaders = checkNotNull(cacheHeaders, "cacheHeaders == null");
       return this;
     }
 
     /**
-     * Set the default {@link CacheHeaders} strategy that will be passed
+     * Set the default {@link ResponseFetcher} that will create a {@link ResponseFetcher} to be passed
```
sav007: smth wrong here `ResponseFetcher` will create a `ResponseFetcher`

apollo-rx2support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
```diff
@@ -66,30 +63,39 @@ private Rx2Apollo() {
   }
 
   /**
-   * Converts an {@link ApolloCall} to a synchronous Single.
+   * Converts an {@link ApolloCall} to an {@link Observable}.
```
sav007: I guess we should provide more info why it's `Observable` on cache and network fetch strategy example

apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
```diff
@@ -81,30 +79,50 @@ private RxApollo() {
   }
 
   /**
-   * Converts an {@link ApolloCall} to a Single.
+   * Converts an {@link ApolloCall} to a Observable.
```
sav007: the same as above

BenSchwab:
@sav007 I guess I should bump the version number again since this is a breaking change? Not sure we have enough new stuff since that last release to publish an official release

sav007:
but we have't released 0.4.0 yet, why do we need to bump it to 0.5.0?

BenSchwab:
Ok, I wasn't sure if we were trying to keep snapshot builds being semver capatabilble. But if not, let's wait to bump!

Issue #550, opened 2017-07-07
Add Operation Id to generated code


BenSchwab:
Many clients will require GraphQL queries to be registered on the server (whitelisting) for security and performance reasons (send up a hash instead of a long query body). Apollo CodeGen added support for this here: https://github.com/apollographql/apollo-codegen/commit/703edf258637232f1015cb33f6501052721ed71d.

Issue #559, opened 2017-07-11
How to handle `mergeInFieldsFromFragmentSpreads` with imperative store api


BenSchwab:
I was investigating updating to the most recent version of apollo codegen to add support for query operation ids. The new version has a cool - but maybe problematic feature -- fragments get spread onto their containing type:

consider:

```
query HeroAndFriendsWithFragments($episode: Episode) {
  hero(episode: $episode) {
    ... HeroWithFriendsFragment
  }
}

fragment HeroWithFriendsFragment on Character {
  id
  name
  friends {
    ... HumanWithIdFragment
  }
}
```

The new IR file:

```
	{
			"filePath": "/Users/ben_schwab/airapps/apollo-android/apollo-integration/src/main/graphql/com/apollographql/apollo/integration/normalizer/HeroAndFriendsWithFragments.graphql",
			"operationName": "HeroAndFriendsWithFragments",
			"operationType": "query",
			"rootType": "Query",
			"variables": [
				{
					"name": "episode",
					"type": "Episode"
				}
			],
			"source": "query HeroAndFriendsWithFragments($episode: Episode) {\n  hero(episode: $episode) {\n    __typename\n    ...HeroWithFriendsFragment\n  }\n}",
			"fields": [
				{
					"responseName": "hero",
					"fieldName": "hero",
					"type": "Character",
					"args": [
						{
							"name": "episode",
							"value": {
								"kind": "Variable",
								"variableName": "episode"
							}
						}
					],
					"isDeprecated": false,
					"deprecationReason": null,
					"fields": [
						{
							"responseName": "__typename",
							"fieldName": "__typename",
							"type": "String!"
						},
						{
							"responseName": "id",
							"fieldName": "id",
							"type": "ID!",
							"description": "The ID of the character",
							"isDeprecated": false,
							"deprecationReason": null
						},
						{
							"responseName": "name",
							"fieldName": "name",
							"type": "String!",
							"description": "The name of the character",
							"isDeprecated": false,
							"deprecationReason": null
						},
						{
							"responseName": "friends",
							"fieldName": "friends",
							"type": "[Character]",
							"description": "The friends of the character, or an empty list if they have none",
							"isDeprecated": false,
							"deprecationReason": null,
							"fields": [
								{
									"responseName": "__typename",
									"fieldName": "__typename",
									"type": "String!"
								}
							],
							"fragmentSpreads": [
								"HumanWithIdFragment"
							],
							"inlineFragments": [
								{
[...]

PR #566, opened 2017-07-12
Update to apollo codegen 0.15.0

BenSchwab:
This is a pre-req for https://github.com/apollographql/apollo-android/issues/550

This includes a change https://github.com/apollographql/apollo-codegen/commit/faa5715aee550f5f1adb04ce08ed0ed523accf8e which enables:

`mergeInFieldsFromFragmentSpreads`.

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/TestOperation.graphql
+++ b/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/TestOperation.graphql
@@ -14,7 +14,7 @@ fragment starshipFragment on Starship {
   pilotConnection {
     edges {
       node {
-        ...pilotFragment
+       name
       }
     }
   }
--- a/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/TestOperation.graphql
+++ b/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/TestOperation.graphql
@@ -14,7 +14,7 @@ fragment starshipFragment on Starship {
   pilotConnection {
     edges {
       node {
-       name
+       ...pilotFragment
       }
     }
   }
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
@@ -24,6 +24,7 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Collections;
 import java.util.Date;
+import java.util.List;
 
 import okhttp3.OkHttpClient;
 import okhttp3.mockwebserver.MockResponse;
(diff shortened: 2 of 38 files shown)


apollo-integration/src/test/java/com/apollographql/apollo/ResponseWriteTestCase.java
```diff
@@ -249,35 +250,63 @@ public void operation_with_fragments() throws Exception {
     assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().id()).isEqualTo("1003");
     assertThat(hero.fragments().heroWithFriendsFragment().friends().get(2).fragments().humanWithIdFragment().name()).isEqualTo("Leia Organa");
 
+    final HumanWithIdFragment supermanAsHumanFragment = new HumanWithIdFragment(
```
BenSchwab: @sav007 This is an example of https://github.com/apollographql/apollo-android/issues/559

The fact that fragment types are merged in now (and a different type) makes the imperative store api a bit more clumsy and the potential problem -- what happens if a user inconsistently creates the two different fragment objects? Which one will win in the merge process?

sav007: Yeah, I have the same thought while I was implementing write operation. I think it is user responsibility to make sure that fragments with the same id have the same data. If they are different then it depends of field tree traversal

sav007:
I see that fields from fragment being merged into query, is it because of logic behind operation id generation?

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/TestOperation.graphql
```diff
@@ -14,7 +14,7 @@ fragment starshipFragment on Starship {
   pilotConnection {
     edges {
       node {
-        ...pilotFragment
+       name
```
sav007: Is it intentionally? As this test was used for testing fragment inside another fragment

BenSchwab: Good catch, was testing something myself and forgot to switch back

BenSchwab:
@sav007 The field merging is separate from operation id generation. @martijnwalraven May be able to give more insight on it.

I will follow up with another PR to enable the operationId flag, but I wanted to do this PR separately to discuss the field spreading, in case there is any changes we want to do to handle it.

BenSchwab:
https://github.com/apollographql/apollo-codegen/pull/147 This is the PR that enables operationId in Apollo codegen. When we turn the flag on we will get an operationId and source with fragments merged on the IR file (this is different than the fragment spreading we see in this PR) and a build time generated json map file of operationId -> source with fragments

BenSchwab:
In 0.15.2 we can turn off the fragment spread. Looks like there will be no IR file changes until I enable the operationId generation. So going to close this, and proceed work on the operationId change.

PR #568, opened 2017-07-14
Add invalidate cached item by key from Apollo store

### LINKED ISSUE #564: Ability to invalidate cached items by key needed.
This discussion started in issue #561. @BenSchwab suggested I create a separate issue report for the following:

----

There currently appears to be no way to invalidate specific items in the cache. From what I can tell, it's an all or nothing proposition with only a `clearAll()` type method being exposed to users of the Apollo SDK. There are valid use cases for needing to invalidate specific cached objects.

For example, let's say you had a list/detail situation. A query is executed to get a list of objects for the UI, each with only a few properties to display summary level info for each. The detail UI makes a query for one of the specific objects and includes more properties. Everything loaded so far is cached. If I wanted to invalidate the cache for the one specific object, ideally I could blow it away using its cache key. I guess I would expect that if I did that, the cached list would also become invalidated since it references the single object which was invalidated.

I don't see how one would do something like that with the current SDK. The only options for cache invalidation seem to be to blow everything away. 

To quote @BenSchwab on this topic:

> Can you make a separate issue for deleting single cache entries? We should be able to do something via the imperative store api.

sav007:
Closes #564

### DIFF
--- a/apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
+++ b/apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
@@ -6,6 +6,7 @@ import android.database.sqlite.SQLiteStatement;
 
 import com.apollographql.apollo.api.internal.Optional;
 import com.apollographql.apollo.cache.CacheHeaders;
+import com.apollographql.apollo.cache.normalized.CacheKey;
 import com.apollographql.apollo.cache.normalized.NormalizedCache;
 import com.apollographql.apollo.cache.normalized.Record;
 import com.apollographql.apollo.cache.normalized.RecordFieldAdapter;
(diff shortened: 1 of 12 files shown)


BenSchwab:
Based #564 it seems like it might be useful to delete all keys for a specific query (dependentKeys). This enables that as we could repeatedly call delete(cachekey), but we can probably get some performance gains if we offer this API directly and run the delete actions in once transaction.

sav007:
Probably makes sense to have such API, to run batch remove in one transaction, at the same time user can do it manually by loop over depended keys and call `apolloStore.remove()` but its will be run in separate transaction for each call.

How about just add a new API call `apolloStore.remove(List<CacheKey>)`?

eschlenz:
What @BenSchwab outlined is specifically what I was after. I see that the ability to remove by key was added, so that alone is a huge improvement. The invalidation of cached objects that refer to an invalidated item would be even better.

PR #569, opened 2017-07-14
Generate operation id and add option to send to server instead of document

BenSchwab:
- Updates apollo codedgen to 0.15.2 (turns off new fragment spread feature)
- Builds now output a `[Variant]OperationIdMap.json` in the same directory of API.json
Developers can write a script to upload this artifact to the server, so that all the operations in the app can be registered.

closes https://github.com/apollographql/apollo-android/issues/550

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: android
 sudo: true
+dist: precise
 
 android:
   components:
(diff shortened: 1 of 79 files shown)


BenSchwab:
@digitalbuddha when you have a chance mind taking a look? Not sure if this is a feature you need at NYT

sav007:
@digitalbuddha after queries I saw from you, this feature will def help you to save bandwidth

PR #573, opened 2017-07-18
Introduce Apollo Store Operation 

### LINKED ISSUE #571: Apollo store API improvements
Problem: current store API implemented as sync operations, but under the hood it could potentially work with SQL database and it's explicitly visible for user.
Solution: improve API to provide 2 ways of making calls as sync operation and via `dispatcher`. In the similar way as `ApolloCall` does.

sav007:
Closes #571


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStoreOperation.java
```diff
@@ -0,0 +1,118 @@
...
+public abstract class ApolloStoreOperation<T> {
+
+  public static <T> ApolloStoreOperation<T> emptyOperation(final T result) {
+    return new ApolloStoreOperation<T>(null) {
+      @Override protected T perform() {
+        return result;
+      }
+
+      @Override public void enqueue(Callback<T> callback) {
+        if (callback != null) {
```
BenSchwab: `callback.get() == null` ?

apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStoreOperation.java
```diff
@@ -0,0 +1,118 @@
...
+      throw new ApolloException("Failed to perform store operation", e);
+    }
+  }
+
+  /**
+   * Schedules operation to be executed in dispatcher
+   *
+   * @param callback to be notified about operation result
+   */
+  public void enqueue(final Callback<T> callback) {
```
BenSchwab: Should callback be `@Nullable`?

Issue #560, opened 2017-07-11
Make sure RealApolloCall doesn't hold reference to user callback after complete or cancel


sav007:
We need to nullify reference to user provided callback in `com.apollographql.apollo.internal.RealApolloCall#enqueue` that if call has been finished or canceled we don't hold the strong reference to user callback any longer.

BenSchwab:
I have to move around some of the cancel/dispose logic in https://github.com/apollographql/apollo-android/pull/556 because of the added `Callback#onComplete` for multi-response fetchers. I can try to fix in that PR.

sav007:
@BenSchwab Yeah I was waiting on your PR to go first anyway and then fix this

sav007:
So I will reassign this to you then

PR #574, opened 2017-07-18
Null out callback in async operations in Call, Prefetch, and Watcher

BenSchwab:
Closes https://github.com/apollographql/apollo-android/issues/560

I played around with weak references, but they seemed a little indeterministic which made testing tough.

- Added more state assertations 
- Because this adds another non-final variable in asyncrhonous code, I added sychronized blocks to the points where the variables are accessed. Because a callback cannot be asserted non-null and called into in one expression, using volatile does not seem sufficient

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -70,6 +70,7 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
   final List<Query> refetchQueries;
   final Optional<QueryReFetcher> queryReFetcher;
   final AtomicBoolean executed = new AtomicBoolean();
+  Optional<Callback<T>> callback = Optional.absent();
   volatile boolean canceled;
   volatile boolean completed;
 
@@ -153,16 +154,22 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
   }
 
   @Override public void enqueue(@Nullable final Callback<T> responseCallback) {
-    if (!executed.compareAndSet(false, true)) {
-      throw new IllegalStateException("Already Executed");
-    }
-    if (canceled) {
-      if (responseCallback != null) {
-        responseCallback.onCanceledError(new ApolloCanceledException("Canceled"));
+    synchronized (this) {
+      if (!executed.compareAndSet(false, true)) {
+        throw new IllegalStateException("Already Executed");
       }
+      this.callback = Optional.fromNullable(responseCallback);
+      if (canceled) {
+        if (this.callback.isPresent()) {
+          this.callback.get().onCanceledError(new ApolloCanceledException("Canceled"));
+          return;
+        } else {
+          throw new IllegalStateException("Call was canceled");
+        }
+      }
+      tracker.registerCall(this);
     }
-    tracker.registerCall(this);
-    interceptorChain.proceedAsync(dispatcher, fetchOptions, interceptorCallbackProxy(responseCallback));
+    interceptorChain.proceedAsync(dispatcher, fetchOptions, interceptorCallbackProxy());
   }
 
   @Nonnull @Override public RealApolloQueryWatcher<T> watcher() {
@@ -191,10 +198,13 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
   }
 
   @Override public void cancel() {
-    canceled = true;
-    interceptorChain.dispose();
-    if (queryReFetcher.isPresent()) {
-      queryReFetcher.get().cancel();
+    synchronized (this) {
+      canceled = true;
+      callback = Optional.absent();
+      interceptorChain.dispose();
+      if (queryReFetcher.isPresent()) {
+        queryReFetcher.get().cancel();
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -153,16 +154,22 @@ private RealApolloCall(Builder<T> builder) {
...
 
   @Override public void enqueue(@Nullable final Callback<T> responseCallback) {
-    if (!executed.compareAndSet(false, true)) {
-      throw new IllegalStateException("Already Executed");
-    }
-    if (canceled) {
-      if (responseCallback != null) {
-        responseCallback.onCanceledError(new ApolloCanceledException("Canceled"));
+    synchronized (this) {
+      if (!executed.compareAndSet(false, true)) {
```
sav007: there is no need this code to be under sync bloc as it Atomic operation

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -153,16 +154,22 @@ private RealApolloCall(Builder<T> builder) {
   }
 
   @Override public void enqueue(@Nullable final Callback<T> responseCallback) {
-    if (!executed.compareAndSet(false, true)) {
-      throw new IllegalStateException("Already Executed");
-    }
-    if (canceled) {
-      if (responseCallback != null) {
-        responseCallback.onCanceledError(new ApolloCanceledException("Canceled"));
+    synchronized (this) {
```
sav007: and do we really need to have sync block?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -191,10 +198,13 @@ private RealApolloCall(Builder<T> builder) {
   }
 
   @Override public void cancel() {
-    canceled = true;
-    interceptorChain.dispose();
-    if (queryReFetcher.isPresent()) {
-      queryReFetcher.get().cancel();
+    synchronized (this) {
```
sav007: do we really need this?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -70,6 +70,7 @@
   final List<Query> refetchQueries;
   final Optional<QueryReFetcher> queryReFetcher;
   final AtomicBoolean executed = new AtomicBoolean();
+  Optional<Callback<T>> callback = Optional.absent();
```
sav007: What do you think to replace it with AtomicReference? And try then remove all sync blocks?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloQueryWatcher.java
```diff
@@ -24,10 +25,10 @@
 
 final class RealApolloQueryWatcher<T> implements ApolloQueryWatcher<T> {
   private RealApolloCall<T> activeCall;
-  private ApolloCall.Callback<T> callback;
+  private Optional<ApolloCall.Callback<T>> callback = Optional.absent();
   private ResponseFetcher refetchResponseFetcher = ApolloResponseFetchers.CACHE_FIRST;
-  private volatile boolean canceled;
-  private boolean executed = false;
+  private boolean canceled;
+  private boolean executed;
```
sav007: I would do the same as we did in `RealApolloCall` with atomic boolean and CAS operation

BenSchwab:
@sav007 I don't think atomic values are sufficient in a lot of cases. We need the entire callback functions to be synchronous in respect to the entire cancelation function. Otherwise, a call to cancel could interleave while an `onResponse` or `enqueue` was ongoing, causing issues. 

For example, take `enqueue`:
```
  if (!executed.compareAndSet(false, true)) {
        throw new IllegalStateException("Already Executed");
      }
      this.callback = Optional.fromNullable(responseCallback);
      if (canceled) {  
        if (this.callback.isPresent()) {
          this.callback.get().onCanceledError(new ApolloCanceledException("Canceled"));
          return;
        } else {
          throw new IllegalStateException("Call was canceled");
        }
      }
    [--- cancel called here in another thread  ---]
      tracker.registerCall(this); // do we actually want to register this call if it's cancelled?
    }
    interceptorChain.proceedAsync(dispatcher, fetchOptions, interceptorCallbackProxy());
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
BenSchwab: I feel like we need the `canceled = true` and `callback = Optional.absent` to happen atomically. Otherwise we could have a state where a user gets a callback, during which she check `call.isCanceled` and sees that evaluate to `true`... probably not too harmful, but it does seem to break our contract.

sav007:
Yeah, I see your point. I guess for now lets leave it. 

I have in mind some very simple state machine (FSM) that has 4 states:
`IDLE` -> `IN_PROGRESS` -> `COMPLETED` -> `CANCELED`. Transferring between states happens with CAS operations. And what nice is that only states `IDLE`, `IN_PROGRESS` keep reference for callback (we start from `IDLE(callback) -> `IN_PROGRESS(callback)` then when we transfer to `COMPLETED` or `CALLED` (which doesn't have strong ref to callback), it will be picked up by GC.

But maybe it's over engineering for now, so let's keep it simple with sync. Maybe later I will share proposal.

BenSchwab:
Took another stab at cleaning it up --

In the version I have, while some are implicit, could be mapped to:
`IDLE`
`IN_PROGRESS`
`CANCELLED`
`TERMINATED` (onFailure or onComplete)

`IDLE` -> `IN_PROGRESS` occurs in enqueue
`IN_PROGRESS` -> `CANCELLED` occurs in cancel
`IN_PROGRESS` -> `TERMINATED` occurs in `onComplete` or `onFailure`

Each of these transitions operates a few Atomic references, which should be operated on in one synchronous transaction:
`tracker`
`callback`
and state flags `terminated` `canceled` `executed`.

The synchronize blocks could be seen as the "transitions" that change values and assert state is accurate.

I can definitely see the value in formalizing this even farther.

BenSchwab:
I'm playing around with the FSM idea. I like it a lot: here's a WIP sketch https://github.com/apollographql/apollo-android/pull/577

BenSchwab:
Going to move forward with #577

PR #577, opened 2017-07-19
Model ApolloCall/Query/Fetcher state as FSM

BenSchwab:
With the addition of multi-response callbacks, the states in ApolloCall / Prefetch / Query watcher  are more complex. We also must be sure to properly track/untrack calls, null out callbacks in certain states.

This is an attempt to use a FSM to model this behavior.

- [x] Migrate Prefetch
- [x] Migrate QueryWatcher

closes https://github.com/apollographql/apollo-android/issues/560

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -124,6 +124,12 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
     Response<T> response;
     try {
       response = interceptorChain.proceed(fetchOptions).parsedResponse.or(Response.<T>builder(operation).build());
+    } catch (Exception e) {
+      if (state.get() == CANCELED) {
+        throw new ApolloCanceledException("Call canceled", e);
+      } else {
+        throw e;
+      }
     } finally {
       terminate();
     }
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -44,7 +44,10 @@ import okhttp3.Call;
 import okhttp3.HttpUrl;
 
 import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
-import static com.apollographql.apollo.internal.CallState.*;
+import static com.apollographql.apollo.internal.CallState.ACTIVE;
+import static com.apollographql.apollo.internal.CallState.CANCELED;
+import static com.apollographql.apollo.internal.CallState.IDLE;
+import static com.apollographql.apollo.internal.CallState.TERMINATED;
 import static java.util.Collections.emptyList;
 
 @SuppressWarnings("WeakerAccess")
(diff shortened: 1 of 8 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -119,102 +119,93 @@ private RealApolloCall(Builder<T> builder) {
...
         .build();
   }
 
-  @Override public void cancel() {
-    canceled = true;
-    interceptorChain.dispose();
-    if (queryReFetcher.isPresent()) {
-      queryReFetcher.get().cancel();
+  @Override public synchronized void cancel() {
+    switch (state.get()) {
```
sav007: In one of my project I had this non blocking:
```
do {
      State oldState = state.get()
      switch (oldState) {
        case ACTIVE: {
          if (state.compareAndSet(oldState, CANCELED)) {
            // ACTIVE -> CANCELED
            return;
          }
          break;
        }
        case IDLE: {
          if (state.compareAndSet(oldState, CANCELED)) {
            // IDLE -> CANCELED
            return;
          }
          break;
        }
        default:
          return;
      }
    } while (true)
```

Just as a suggestion.

BenSchwab: Not quite sure I follow, what's the advantage of the do while loop?

sav007: First you don't need sync block, and do-while requires to try again if another thread changes state (`state.compareAndSet` returned false).

```
do {
      State oldState = state.get()
      switch (oldState) {
        case ACTIVE: {
          // if some other thread changed the state below returns false and we will try again
          // on next iteration of loop
          if (state.compareAndSet(oldState, CANCELED)) {
            // ACTIVE -> CANCELED
            return;
          }
          break;
        }
        case IDLE: {
          if (state.compareAndSet(oldState, CANCELED)) {
            // IDLE -> CANCELED
            return;
          }
          break;
        }
        default:
          return;
      }
    } while (true)
```

BenSchwab: @sav007 But as our state transitions involve mutating state (contained in multiple variables) inside the `if` you have sketched above, couldn't another thread see the new state flag before state transition side effects have finished?

Maybe I'm just not fully grasping your solution -- I won't have time to work on this until this weekend, so feel free to pick this up if you would like in the meantime.

sav007: Agh, right we have mutations.... I hate mutations. Sure then just lets move on with what you have

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -68,9 +72,8 @@
   final List<Query> refetchQueries;
   final Optional<QueryReFetcher> queryReFetcher;
   final boolean sendOperationdIdentifiers;
-  final AtomicBoolean executed = new AtomicBoolean();
-  volatile boolean canceled;
-  volatile boolean completed;
+  final AtomicReference<CallState> state = new AtomicReference<>(IDLE);
```
sav007: I believe we can omit `AtomicReference ` here as we don't use any CAS operations, `volatile` should be enough. The same for callback

Issue #572, opened 2017-07-17
Remove reflection from request parsing


digitalbuddha:
It was observed by a user that large requests lead to lots of reflection creating moshi json adapters through reflection. We need to either generate moshi json adapters or create our own serialization for request objects.  As an example: it took 1200ms to create the Json request string & under 200ms to get a response.

sav007:
it seems we can remove moshi completely and generate marshaller code in the same way as we doing for operation data marshaller

PR #578, opened 2017-07-19
Replace Moshi with generated marshaller for operation json serialization

sav007:
Part of #572 

Next step is to remove Moshi from SqlNormalizedCache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: android
 sudo: true
+dist: precise
 
 android:
   components:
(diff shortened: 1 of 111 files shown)


digitalbuddha:
@sav007 should this be made configurable. I worry that most won't care about the optimization but will care about the additional methods the request parsing generates.

sav007:
@digitalbuddha well I see your point but not sure if that won't be a messy to have it both. I fell like we already have too much configs, and it was our initial idea use generation instead of Moshi.
Plus I feel like we really won't benefit of saving couple methods here. Maybe we can later optimize it a little bit when such concern raised.

jaredsburrows:
@digitalbuddha I completely agree. 

@sav007 In this PR, I did not see you remove the dependency from the `build.gradle`. Did you remove it in another PR?

jaredsburrows:
Are you going to remove moshi completely from this repository? It looks like it is still used: https://github.com/apollographql/apollo-android/search?utf8=%E2%9C%93&q=moshi&type=.

PR #582, opened 2017-07-21
Remove Moshi from cache layer

### LINKED ISSUE #572: Remove reflection from request parsing

sav007:
Update RecordFieldAdapter to use direct json api instead of moshi lib.
Final step to rid off Moshi lib from runtime module.

Closes #572


jaredsburrows:
@sav007 Can we remove it from Apollo completely?

sav007:
@jaredsburrows we can't as it used internally for our apollo-compiler but it's not used at runtime

jaredsburrows:
@sav007 Ok, thank you so much. That is what I meant!

PR #591, opened 2017-07-27
Fix logic in apolloclient build when using callFactory

hzsweers:
This was missing a return in the okhttp block and incorrectly fell through to a true return at the end, which should have actually been false.

Otherwise - if one provided a callfactory rather than an okhttpclient instance, it would fail with a ClassCastException trying to incorrectly add the interceptor


hzsweers:
Test failure seems like a flake? `SocketException`

Issue #562, opened 2017-07-12
Unit tests should not call `Assert.fail` on a background thread


BenSchwab:
A lot of our tests call `Assert.fail` on a background thread. This won't actually fail the test:

https://stackoverflow.com/questions/2596493/junit-assert-in-thread-throws-exception

sav007:
I believe I've fixed this in most case with this PR https://github.com/apollographql/apollo-android/pull/639 closing this

PR #595, opened 2017-07-29
add missing break

martinbonnin:


digitalbuddha:
:+1

sav007:
can we add test case for this please

apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java
```diff
@@ -104,6 +111,50 @@ public void testQueryWatcherUpdated_SameQuery_DifferentResults() throws IOExcept
...
+
+    EpisodeHeroNameQuery query = EpisodeHeroNameQuery.builder().episode(Episode.EMPIRE).build();
+    server.enqueue(mockResponse("EpisodeHeroNameResponseWithId.json"));
+
+    ApolloQueryWatcher<EpisodeHeroNameQuery.Data> watcher = apolloClient.query(query).watcher();
+    watcher.enqueueAndWatch(
+        new ApolloCall.Callback<EpisodeHeroNameQuery.Data>() {
+          @Override public void onResponse(@Nonnull Response<EpisodeHeroNameQuery.Data> response) {
+            if (secondResponseLatch.getCount() == 2) {
+              assertThat(response.data().hero().name()).isEqualTo("R2-D2");
```
sav007: don't use assert in the background thread, as if it fails it will be ignored
https://github.com/apollographql/apollo-android/issues/562

martinbonnin: I see... I just copy/pasted the test above so we might need to fix some of them. We still need to check that so my favorite solution would be to have the callback run in the main thread. I know how to do that in RxJava, not sure we want to use RxJava in these files though. Any idea what would be the best solution here ?

martinbonnin: Actually, I see RxJava in the dependencies of apollo-integration so maybe we could use that. Is it ok to use `AndroidSchedulers.mainThread()` in there or are we pure java ?

sav007: > We still need to check that so my favorite solution would be to have the callback run in the main thread

Yes we have to fix other places as well, and we have a ticket for that

> Is it ok to use AndroidSchedulers.mainThread() in there or are we pure java ?

It's a pure java test. What you can do is use `AtomicReference`  to set all what you need inside callback, and assert then outside callback. (Ugly I know but not sure what else we can do)

martinbonnin: should be good now

martinbonnin:
And somehow tests do not run on ci :-(... Any chance I can get a more verbose output ?

sav007:
```
/home/travis/build/apollographql/apollo-android/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.java:149: error: constructor Record in class Record cannot be applied to given types;
        Record record = new Record("2001");
                        ^
  required: String,Map<String,Object>,UUID
  found: String
  reason: actual and formal argument lists differ in length
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
```

martinbonnin:
Thanks, I just merged master. It should be ok now.

Issue #583, opened 2017-07-22
Introduce Optimistic updates for mutations


sav007:
**The problem**: in most cases the result of mutations are well known even before waiting for server response. The idea behind optimistic updates are pretty straight forward, to be able specify `optimisticResponse` for mutation operation that will be temporary saved into memory cache before network response arrives.

**Proposal**: introduce the layer that is basically separate instance of ApolloStore for optimistic updates only and that based on mem cache. Any queries that try to read response from cache will first hit ApolloStore with optimistic updates and if it can't resolve cache response then  will hit regular ApolloStore with normalized cache. The rollback mechanism for optimistic updates should remove them in case if mutation succeed or failed. 

http://dev.apollodata.com/react/mutations.html#optimistic-ui
https://dev-blog.apollodata.com/mutations-and-optimistic-ui-in-apollo-client-517eacee8fb0

sav007:
@BenSchwab @digitalbuddha @martijnwalraven I'm going to start working on this feature so if you have any thoughts, comments pls share

BenSchwab:
I was thinking of a similar approach! Sgtm

PR #607, opened 2017-08-03
Introduce OptimisticNormalizedCache and API to write/rollback

sav007:
Introduce `OptimisticNormalizedCache` to be always used as first layer in a chain
Introduce `ApolloStore` API to write/rollback optimistic updates
Some small refactoring related to `RecordFieldJsonAdapter` that required only for SQL cache

Part of #583


BenSchwab:
Maybe I'm missing something, but does the following work:

Current key value (not in optimisitc cache):  `A -> 0`
Optimistic update 1 `A -> 1`
Optimistic update 2 `A -.> 2`
rollback optimistic update 2:
expect `A -> 1`, but it seems that rolling back 2 would remove the `A` from the optimistic cache, forgetting about update 1, which is still pending and valid.

sav007:
@BenSchwab yes that is correct, update with version 2 will override update with version 1. And when we rollback the version 2 it will rollback all. This case will happen only when second query was faster than the first one.

In order to support this case I guess we need to keep the whole history of changes for all records in mem cache. I think for now we can keep this simple version of implementation, and I will think how we can solve this case.

sav007:
@BenSchwab Yeah, it will be almost similar to JS: `ApolloMutationCall.optimisticUpdates(...)`

PR #610, opened 2017-08-03
bug/fixNonClonedCall

digitalbuddha:
Currently we are not cloning the `originalCall` when creating an `Observable` this pr will make sure we clone in case a user calls `.retry()`

### DIFF
--- a/apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
+++ b/apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
@@ -79,7 +79,8 @@ public final class RxApollo {
   }
 
   /**
-   * Converts an {@link ApolloCall} to a Observable. The number of emissions this Observable will have
+   * Converts an {@link Apollo
+   } to a Observable. The number of emissions this Observable will have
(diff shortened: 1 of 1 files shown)


apollo-rxsupport/src/main/java/com/apollographql/apollo/rx/RxApollo.java
```diff
@@ -79,17 +79,19 @@ private RxApollo() {
...
    * @param <T>              the value type
    * @param backpressureMode The {@link rx.Emitter.BackpressureMode} to use.
    * @return the converted Observable
    */
-  @Nonnull public static <T> Observable<Response<T>> from(@Nonnull final ApolloCall<T> call,
+  @Nonnull public static <T> Observable<Response<T>> from(@Nonnull final ApolloCall<T> originalCall,
       Emitter.BackpressureMode backpressureMode) {
-    checkNotNull(call, "call == null");
+    checkNotNull(originalCall, "call == null");
+    Call<T> call = originalCall.clone();
```
sav007: `Call<T> call = originalCall.clone();` must be done inside:
```
return Observable.create(new Action1<Emitter<Response<T>>>() {
      @Override public void call(final Emitter<Response<T>> emitter) {
```

PR #612, opened 2017-08-06
clone call for each subscriber

digitalbuddha:
take 2 :-).

We should really add an `isExecuted` to `ApolloCall` similar to `Retrofit`'s Call.  Otherwise there isn't a good way to test this or see whether we actually need to clone.


digitalbuddha:
will reopen

PR #618, opened 2017-08-09
Introduce record version history for optimistic store

### LINKED ISSUE #583: Introduce Optimistic updates for mutations

sav007:
Rolling back optimistic updates is smart now, it keeps the history of record updates.
If mutation finished it's execution it rolls back only its version of the record and keeps the rest versions for still running mutations.

Closes #583

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/OptimisticNormalizedCache.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/OptimisticNormalizedCache.java
@@ -7,12 +7,11 @@ import com.apollographql.apollo.cache.CacheHeaders;
 import com.nytimes.android.external.cache.Cache;
 import com.nytimes.android.external.cache.CacheBuilder;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.List;
+import java.util.LinkedList;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/OptimisticNormalizedCache.java
```diff
@@ -102,31 +101,78 @@
...
+      if (journal.history.isEmpty()) {
+        removedKeys.add(cacheKey);
       }
     }
-    lruCache.invalidateAll(invalidateKeys);
+    lruCache.invalidateAll(removedKeys);
+    return changedCacheKeys;
+  }
+
+  private final class RecordJournal {
```
BenSchwab: Like the name 👍

PR #620, opened 2017-08-10
Treat ID as custom scalar type

### LINKED ISSUE #585: Add ability to use CustomTypeAdapter with "regular" scalars (e.g. ID)
Ideally we should be able to configure the `ApolloClient` to use logic inside a `CustomTypeAdapter` when mapping regular scalar fields, e.g.
```java
return ApolloClient.builder()
    .serverUrl(graphqlEndpoint.toString())
    .okHttpClient(graphQLApolloClient)
    .addCustomTypeAdapter(ScalarType.ID, myIdTypeAdapter) // <--
    .build();
```

sav007:
User now will be able to provide custom type adapter for ID GraphQL type.
By default if mapping is not provided in plugin configuration it will be mapped to String.

Closes #585

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -2,6 +2,7 @@ package com.apollographql.apollo.compiler
 
 import com.apollographql.apollo.compiler.ir.CodeGenerationContext
 import com.apollographql.apollo.compiler.ir.CodeGenerationIR
+import com.apollographql.apollo.compiler.ir.ScalarType
 import com.apollographql.apollo.compiler.ir.TypeDeclaration
 import com.squareup.javapoet.JavaFile
 import com.squareup.moshi.Moshi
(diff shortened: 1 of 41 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -2,6 +2,7 @@ package com.apollographql.apollo.compiler
 
 import com.apollographql.apollo.compiler.ir.CodeGenerationContext
 import com.apollographql.apollo.compiler.ir.CodeGenerationIR
+import com.apollographql.apollo.compiler.ir.ScalarType
 import com.apollographql.apollo.compiler.ir.TypeDeclaration
 import com.squareup.javapoet.JavaFile
 import com.squareup.moshi.Moshi
```
digitalbuddha: unrelated, what is still using Moshi?

apollo-compiler/src/test/graphql/com/example/hero_details_semantic_naming/type/CustomType.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.Class;
+import java.lang.Override;
+import java.lang.String;
+import javax.annotation.Generated;
+
+@Generated("Apollo GraphQL")
+public enum CustomType implements ScalarType {
+  ID {
+    @Override
+    public String typeName() {
```
digitalbuddha: is this same as `graphql` typename or something else?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
sav007: just for compiler, it not used in runtime

apollo-compiler/src/test/graphql/com/example/hero_details_semantic_naming/type/CustomType.java
sav007: yes this is the name of scalar type as it defined in GraphQL schema

sav007:
@digitalbuddha 👍 or 👎

PR #622, opened 2017-08-13
Generate data model builder

### LINKED ISSUE #594: Add `model.newBuilder()` to allow easily cloning data
It would be nice to have an easier way to mutate generated classes instances. Calling the constructor becomes very tedious as data becomes more complex and nested. Ideally, given the query below:

    query {
        video {
            channel {
                name
            }
        }
    }

We could change the `name` field with: 

    builder = data().newBuilder();
    builder.video().channel().withName("newName")
    newData = builder.build()

This would be very handy to update the store with

    store.write(operation, newData)

Not sure how it will affect method count though.

sav007:
Introduce plugin configuration `generateModelBuilder` to enable / disable data model builder generation.
By default it's set to false.

Closes #594

Examples:
https://github.com/apollographql/apollo-android/pull/622/files#diff-1429b2a1e17efe3e3dd15137018df33e

https://github.com/apollographql/apollo-android/pull/622/files#diff-e0f8aa3bbca9447766d25967d5c39f5f

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/BuilderTypeSpecBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/BuilderTypeSpecBuilder.kt
@@ -12,7 +12,7 @@ class BuilderTypeSpecBuilder(
     val typeDeclarations: List<TypeDeclaration>
 ) {
   fun build(): TypeSpec {
-    return TypeSpec.classBuilder(builderClassName)
+    return TypeSpec.classBuilder(builderClass)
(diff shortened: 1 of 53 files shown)


martinbonnin:
Any reason to make this a shallow copy (vs-deep) ? As of now, it doesn't change much. If I want to change a property deep down in `data`, I have to manually create all the intermediate objects

sav007:
@martinbonnin this variant of implementation was more faster as we have already generic code generation for builders in this way. But I see your point could you pls open another issue to make enhancement to the builder?

martinbonnin:
yup, will do

martinbonnin:
https://github.com/apollographql/apollo-android/issues/643

PR #639, opened 2017-08-24
Drop sync API execute call

### LINKED ISSUE #606: Drop synchronous api
As part of discussion in #588 decision was made to drop all Apollo sync operations.
`com.apollographql.apollo.ApolloCall#execute`
`com.apollographql.apollo.ApolloPrefetch#execute`

Any Rx wrappers should be built on top of `com.apollographql.apollo.ApolloCall.Callback` and `com.apollographql.apollo.ApolloPrefetch.Callback` now.

As for the tests there are 2 options: use Rx with blocking operations or create blocking implementation of callback.

sav007:
I know this is a monster PR but I had to fix all tests that we have.
I broke it into separate logical commits, so it will help a little.

Closes #606


apollo-integration/src/test/java/com/apollographql/apollo/ApolloPrefetchTest.java
```diff
@@ -73,137 +70,65 @@ public void setup() {
...
+    server.enqueue(mockResponse("HttpCacheTestAllPlanets.json"));
     faultyCacheStore.failStrategy(FaultyHttpCacheStore.FailStrategy.FAIL_BODY_WRITE);
-    try {
-      apolloClient.prefetch(new AllPlanetsQuery()).execute();
-      fail("exception expected");
-    } catch (Exception expected) {
-    }
+    Rx2Apollo.from(apolloClient.prefetch(new AllPlanetsQuery()))
+        .test()
+        .awaitDone(TIME_OUT_SECONDS, TimeUnit.SECONDS)
```
digitalbuddha: I usually use `awaitTerminalEvent` in cases like this, not sure what difference is with `awaitDone`. From what I gather `awaitTerminalEvent` waits for `onComplete` or `onError` to be called while `awaitDone` merely waits for latch.

sav007: there is not much difference:
```
public final U awaitDone(long time, TimeUnit unit) {
        try {
            if (!done.await(time, unit)) {
                dispose();
            }
        } catch (InterruptedException ex) {
            dispose();
            throw ExceptionHelper.wrapOrThrow(ex);
        }
        return (U)this;
    }
```
and 
```
public final boolean await(long time, TimeUnit unit) throws InterruptedException {
        return done.getCount() == 0 || done.await(time, unit);
    }

public final boolean awaitTerminalEvent(long duration, TimeUnit unit) {
        try {
            return await(duration, unit);
        } catch (InterruptedException ex) {
            Thread.currentThread().interrupt();
            return false;
        }
    }
```

So the first API is just call dispose on timeout and it returns the the TestObserver so you can chain asserts where second just awaits and returns boolean

njuprincerain:
@sav007 I'm trying to use apolloClient in server side Java instead of Android and in some cases I do need synchronous calls instead of async callbacks. What is the right way to have something like `execute()` after this change?

digitalbuddha:
You can use the rxjava bindings or write similar ones for futures

njuprincerain:
@digitalbuddha Thanks for your reply! is it something like ```Rx2Apollo.from(ApolloQueryCall).blockingFirst()```, is it possible not to use Rx2Apollo? since the server is not really react related.

digitalbuddha:
Here's a stackoverflow.com post about how to make a future from a callback. Not idea if best way but it's a way https://stackoverflow.com/a/2180534/1965124

Generally futures are a good low level way to make something synch as you can call future.get()

The answer below the accepted one uses a notifier as an atomic reference that will work too.

PR #645, opened 2017-08-30
allow to merge null values in records

martinbonnin:
If the value was previously absent (and not null), null was not merged


martinbonnin:
Test case added

PR #649, opened 2017-09-01
Add support of GraphQLList of GraphQLLists

### LINKED ISSUE #637: GraphQLList of GraphQLLists?
Good day, I'm having an issue where Android Studio is giving the error `Error:Execution failed for task ':app:generateDebugApolloClasses. com.squareup.javapoet.ParameterizedTypeName cannot be cast to com.squareup.javapoet.ClassName'` when trying to build a query with the following member.
```json
              "name": "myList",
              "description": null,
              "args": [],
              "type": {
                "kind": "LIST",
                "name": null,
                "ofType": {
                  "kind": "LIST",
                  "name": null,
                  "ofType": {
                    "kind": "LIST",
                    "name": null,
                    "ofType": {
                      "kind": "SCALAR",
                      "name": "Float",
                      "ofType": null
                    }
                  }
                }
              }
```

Only happens when there is a LIST of LISTs. Just wondering is this an issue or is there something that needs to be done on my end? I have tried running the query on my local GraphQL and it has no problems fyi. Thanks for any help.

sav007:
Closes #637

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/ResponseField.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/ResponseField.java
@@ -134,48 +134,18 @@ public class ResponseField {
   }
 
   /**
-   * Factory method for creating a Field instance representing {@link Type#SCALAR_LIST}.
+   * Factory method for creating a Field instance representing {@link Type#LIST}.
    *
    * @param responseName alias for the result of a field
    * @param fieldName    name of the field in the GraphQL operation
    * @param arguments    arguments to be passed along with the field
    * @param optional     whether the arguments passed along are optional or required
    * @param conditions   list of conditions for this field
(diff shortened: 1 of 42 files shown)


build.gradle
```diff
@@ -5,6 +5,7 @@ subprojects {
     repositories {
       jcenter()
       maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
+      mavenLocal()
```
digitalbuddha: intentional?

sav007: it helps to run integration tests inside IDE with locally installed new version (new changes) of plugin. But yeah will remove it.

sav007:
@digitalbuddha any other comments?

digitalbuddha:
Nope Thank you for this. Web devs will be happy.

PR #651, opened 2017-09-07
Provide explicit schema path option to gradle plugin

### LINKED ISSUE #633: Provide explicit schema path option to gradle plugin
Per discussion in slack and  https://github.com/apollographql/apollo-android/issues/593 would be great if apollo android plugin has explicit option to provide path to GraphQL schema. This option will have higher priority than schema file auto lookup.

sav007:
Adds new plugin options:
 - `schemaFilePath` path to the schema.json location
 - `outputPackageName` the package name for model generation

Closes #633

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,21 @@ apollo {
 }
 ```
 
+### Explicit Schema location
+By default Apollo-Android tries to lookup GraphQL schema file in `/graphql` folder, the same folder where all your GraphQL queries are stored. 
+For example if query files are located at `/src/main/graphql/com/example/api` then the schema file should be placed to the same location `/src/main/graphql/com/example/api`. Relative path of schema file to `/src/main/graphql` root folder defines the package name for generated models, in our example the package name of generated models will be `com.example.api`.
+
+Alternatively, you can explicitly provide GraphQL schema file location and package name for generated models:
+
+#### Usage
+
+```groovy
+apollo {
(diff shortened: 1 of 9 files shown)


CristianGM:
Thanks for working on that!

May I suggest changing the Readme on the same PR to document the plugin option?

README.md
```diff
@@ -389,6 +389,21 @@ apollo {
...
+
+#### Usage
+
+```groovy
+apollo {
+  schemaFilePath = "/path_to_schema_file/my-schema.json"
+  outputPackageName = "com.my-example.graphql.api"
+}
+```
+
```
sav007: @digitalbuddha could you pld review readme ^^^

README.md
```diff
@@ -389,6 +389,21 @@ apollo {
 }
 ```
 
+### Explicit Schema location
+By default Apollo-Android tries to lookup GraphQL schema file in `/graphql` folder, the same folder where all your GraphQL queries are stored. 
+For example if query files are located at `/src/main/graphql/com/example/api` then the schema file should be placed to the same location `/src/main/graphql/com/example/api`. Relative path of schema file to `/src/main/graphql` root folder defines the package name for generated models, in our example the package name of generated models will be `com.example.api`.
```
digitalbuddha: For example,....    Placed in the same location

Missing comma and change from 'to the same location' to 'in the same location'

PR #654, opened 2017-09-11
Introduce ApolloCall status event notification

### LINKED ISSUE #632: Operation state observer
After a second thought on #588 and #606  here is the proposal to introduce `StateObserver`.

We are going to ~introduce new~ modify API ~to~ of ~`enqueue` operation~ `Callback`:  ~`void enqueue(StateObserver<T> stateObserver)`~ add a new callback `onStateChanged(State<T> newState)` where `State<T>` :

```
public interface State {
   
    public static final class Scheduled implements State {
    }

    public static final class FetchFromCache implements State {
    }

    public static final class FetchFromNetwork implements State {
    }

    public static final class Response<T> implements State {
      public final Response<T> response;

      public Response(Response<T> response) {
        this.response = response;
      }
    }

    public static final class Completed extends State {
    }
  }
```

~With this approach we leave existent API to `enqueue` operation based on callback and introduce new one that explicitly defined as state observer.~

Please feel free to suggest better naming for API. 

@BenSchwab @digitalbuddha @mandrizzle any thoughts?

sav007:
Introduce `ApolloCall.StatusEvent` enum to support different status events that corresponds to action happen to `ApolloCall`.
These events will be emitted via `ApolloCall.Callback#onStatusEvent(StatusEvent)`

See: https://github.com/apollographql/apollo-android/pull/654/files#diff-18999fee7e5f70c2f862624cbab6d833

Closes #632

### DIFF
--- a/apollo-android-support/src/main/java/com/apollographql/apollo/ApolloCallback.java
+++ b/apollo-android-support/src/main/java/com/apollographql/apollo/ApolloCallback.java
@@ -52,10 +52,10 @@ public final class ApolloCallback<T> extends ApolloCall.Callback<T> {
     });
   }
 
-  @Override public void onCompleted() {
+  @Override public void onStatusEvent(@Nonnull final ApolloCall.StatusEvent event) {
     handler.post(new Runnable() {
       @Override public void run() {
-        delegate.onCompleted();
+        delegate.onStatusEvent(event);
       }
     });
   }
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
@@ -87,10 +87,11 @@ public interface ApolloCall<T> extends Cancelable {
     public abstract void onFailure(@Nonnull ApolloException e);
 
     /**
-     * Gets called when the {@link #onResponse(Response)} has been called for the last time. Not called in the
-     * case of an error. {@link #onFailure(ApolloException)} will be called instead.
+     * Gets called whenever any action happen to this {@link ApolloCall}.
+     *
+     * @param event status that corresponds to a {@link ApolloCall} action
      */
-    public void onCompleted() { }
+    public void onStatusEvent(@Nonnull StatusEvent event) { }
 
     /**
      * <p>Gets called when an http request error takes place. This is the case when the returned http status code
@@ -127,5 +128,27 @@ public interface ApolloCall<T> extends Cancelable {
     public void onCanceledError(@Nonnull ApolloCanceledException e) {
       onFailure(e);
     }
+  }
+
+  /**
+   * Represents a status event that corresponds to a {@link ApolloCall} action
+   */
+  enum StatusEvent {
+    /**
+     * {@link ApolloCall} is scheduled for execution
+     */
+    SCHEDULED,
+    /**
+     * {@link ApolloCall} fetches response from cache
+     */
+    FETCH_CACHE,
+    /**
+     * {@link ApolloCall} fetches response from network
+     */
+    FETCH_NETWORK,
+    /**
+     * {@link ApolloCall} is finished its execution
+     */
+    COMPLETED
   }
 }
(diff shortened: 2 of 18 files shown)


apollo-android-support/src/main/java/com/apollographql/apollo/ApolloCallback.java
```diff
@@ -52,10 +52,10 @@ public ApolloCallback(@Nonnull ApolloCall.Callback<T> callback, @Nonnull Handler
     });
   }
 
-  @Override public void onCompleted() {
```
digitalbuddha: Is on completed not a thing anymore?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -87,10 +87,11 @@
     public abstract void onFailure(@Nonnull ApolloException e);
 
     /**
-     * Gets called when the {@link #onResponse(Response)} has been called for the last time. Not called in the
-     * case of an error. {@link #onFailure(ApolloException)} will be called instead.
+     * Gets called whenever any action happen to this {@link ApolloCall}.
+     *
+     * @param event status event that corresponds to the action happen to this call
```
digitalbuddha: Event status event sounds confusing could you reword a bit?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -128,4 +129,26 @@ public void onCanceledError(@Nonnull ApolloCanceledException e) {
       onFailure(e);
     }
   }
+
+  /**
+   * Represents a status event that corresponds to action that happen to {@link ApolloCall}
+   */
+  enum StatusEvent {
+    /**
+     * {@link ApolloCall} scheduled for execution
```
digitalbuddha: Some comments are past tense some are current. Ie scheduled as opposed to fetches. Maybe standardize.

apollo-android-support/src/main/java/com/apollographql/apollo/ApolloCallback.java
sav007: it's now part of `onStatusEvent`, you got notification with `StatusEvent.COMPLETED`

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
sav007: Yeah, open to any suggestions

digitalbuddha: Status that corresponds to a Apollo Call action

sav007: oh, sorry I thought your comment was related to the name of method itself `onStatusEvent`.
:)

apollo-runtime/src/main/java/com/apollographql/apollo/interceptor/ApolloInterceptor.java
```diff
@@ -54,6 +54,13 @@ void interceptAsync(@Nonnull InterceptorRequest request, @Nonnull ApolloIntercep
     void onResponse(@Nonnull InterceptorResponse response);
 
     /**
+     * Called when interceptor starts fetching response from source type
+     *
+     * @param sourceType type of source been used to fetch response from
+     */
+    void onFetch(FetchSourceType sourceType);
```
BenSchwab: What's the difference between this and FETCH_NETWORK/ FETCH_CACHE notifications?

BenSchwab: I guess I'm confused because we already include this data in the `Response` object as a boolean flag, so the StatusEvent callback seems sufficient.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java
```diff
@@ -128,4 +129,26 @@ public void onCanceledError(@Nonnull ApolloCanceledException e) {
...
+     */
+    SCHEDULED,
+    /**
+     * {@link ApolloCall} fetches response from cache
+     */
+    FETCH_CACHE,
```
BenSchwab: Maybe some better documentation if this occurs before the cache fetch begins, or after it completes?

Issue #430, opened 2017-04-18
Testing support for ApolloClient


VisheshVadhera:
@digitalbuddha on slack suggested something like this: 

`MockApolloClient.addResponse(jsonStream).build()`

Instead of pulling in Json from other graphql tools it would be nice if we can feed the response directly into the apolloClient which can then return mock responses.

sav007:
I guess we should take a look at https://github.com/square/retrofit/tree/master/retrofit-mock

digitalbuddha:
I'm not sure if that is right pattern. Retrofit uses proxy methods to
implement an interface. Okhttp does not have that same setup. Maybe
something there though...

CristianGM:
Why not using: https://github.com/square/okhttp/tree/master/mockwebserver ?
( we prefer WireMock but it's heavier )

digitalbuddha:
That's what we're doing now but it would be nice to have first class
testing support for stunning responses.

AdamMc331:
Any updates on how we can do this? I've been trying to find a way to stub responses with mockito and a JSON file I've saved but I can't quite pull this off. I could try MockWebServer but I agree some more first class support for stubbing would be nice.

AdamMc331:
I suppose without the JSON file this is helpful: https://github.com/apollographql/apollo-android/issues/678

PR #660, opened 2017-09-15
Extract operation response parser

### LINKED ISSUE #653: Allow deserializing models with json files in tests
I'm trying to write some unit tests where I need the model objects, but there doesn't seem to be a way of doing this without setting up a mock server that takes in a request, which is a lot of overhead for a unit test.

sav007:
Introduce OperationResponseParser responsible for parsing response from buffered source.
Can be used in tests with mocked raw response string.

Closes #653

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/Response.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/Response.java
@@ -56,6 +56,14 @@ public final class Response<T> {
     return fromCache;
   }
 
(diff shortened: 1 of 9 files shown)


sav007:
@digitalbuddha yes we should definitely think about this API, but for now this PR just refactors internal parsing logic and as a side effect of this it can be used in tests. So I would leave it for now as is and the next step to provide test support on ApolloClient level itself with API you proposed. And we have for this a ticket already https://github.com/apollographql/apollo-android/issues/430 created by you

Issue #570, opened 2017-07-15
Apollo Store as submodule


sav007:
Proposal: extract Apollo store as separate submodule

Motivation: Apollo store became mature and independent thing with own imperative API, we should consider to move it to separate submodule, so `apollo-runtime` will be pure for making network calls, `apollo-store` pure for storing response data. And it will be optional for user in case if he needs only pure network layer.

sav007:
@digitalbuddha @BenSchwab let me know what do you think?

sav007:
<img width="266" alt="screen shot 2017-07-15 at 11 20 43 am" src="https://user-images.githubusercontent.com/902816/28240431-cb2b33fc-694f-11e7-8448-a6883ec40944.png">

All these classes will be moved to this module

digitalbuddha:
Good idea. Was hoping to avoid using it altogether if possible. Sometimes app wants just simple graphql client

BenSchwab:
sgtm

sav007:
Normalized cache is so tightly coupled with the core Apollo runtime that makes almost impossible to extract it as submodule, the effort of doing this is too much than the benefit. Closing it for now.

PR #662, opened 2017-09-17
Simplify fetcher logic by moving FetchOptions to InterceptorRequest

sav007:
`FetchOptions` represents a thin wrapper around one boolean flag `fetchFromCache` and `cacheHeaders`.
The only consumers of this wrapper class are `ApolloInterceptor`'s that takes `InterceptorRequest` as an argument.
The proposal of this PR is to remove `FetchOptions` abstraction and move `fetchFromCache` and `cacheHeaders` to the `InterceptorRequest` itself for now, as it's unclear what other fetch options we are going to have and having `FetchOptions` just to wrap 2 values seems too much.

Part of #570


BenSchwab:
Seems reasonable 👍

PR #680, opened 2017-10-05
Make generateModelBuilder generate deep builders (instead of shallow)

### LINKED ISSUE #643: Make generateModelBuilder generate deep builders (instead of shallow)
Making the generated builders deep would make modifying nested data much easier.

See https://github.com/apollographql/apollo-android/pull/622

sav007:
For each builder generate additional setter with `Mutator<*.Builder>` as an argument:

```
public Builder hero(@Nonnull Mutator<Hero.Builder> mutator) {
  Hero.Builder builder = this.hero != null ? this.hero.toBuilder() : Hero.builder();

  mutator.accept(builder);

  this.hero = builder.build();
  return this;
}
```

That will allow user to do next:
```
data.toBuilder()
  .hero(someHero)
  // or
  .hero(new Consumer<>() {
    void accept(Hero.Builder builder) {
      builder.name("Joe")
       .age(31)
    }
  })
```

This will be even better with lambda support:
```
data.toBuilder()
  .hero(someHero)
  // or
  .hero(builder -> builder
       .name("Joe")
       .age(31)
  )
```

Closes #643

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/internal/Mutator.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/internal/Mutator.java
@@ -1,0 +1,17 @@
+package com.apollographql.apollo.api.internal;
+
+/**
+ * Represents an operation that accepts a single input argument, mutate it and returns no result
+ *
+ * @param <T> the type of the input to the operation
+ */
+public interface Mutator<T> {
+
+  /**
+   * Performs this operation on the given argument.
+   *
+   * @param t the input argument
+   */
+  void accept(T t);
+
+}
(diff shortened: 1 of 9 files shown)


sav007:
cc: @martinbonnin

martinbonnin:
Just to let you know I started looking into this but I have troubles importing the snapshot into my project. Somehow the Consumer<> class is not exported not sure exactly why...

Also it would help to have an integration test where we could demonstrate the usage of this. Not really sure how to do this without adding another submodule as it requires changing some compiler flags.

Anyways, I'll post when I get it working

apollo-api/src/main/java/com/apollographql/apollo/api/internal/Consumer.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Represents an operation that accepts a single input argument and returns no result. Unlike most other functional
+ * interfaces, {@code Consumer} is expected to operate via side-effects.
+ *
+ * <p>This is a <a href="package-summary.html">functional interface</a> whose functional method is {@link
+ * #accept(Object)}.
+ *
+ * @param <T> the type of the input to the operation
+ */
+public interface Consumer<T> {
```
martinbonnin: Did you use Consumer on purpose ? Why not call the class Mutator<T> or BuilderMutator<T> so as to avoid confusion with the other Consumers (reactivex, java.util, com.annimon.stream, maybe others ?)

sav007: Just copied from Java8, but yeah we can rename it

PR #690, opened 2017-10-13
Simplify API for `clearNormalizedCache`

### LINKED ISSUE #689: Clear Normalized cache not working
The query still pulls info from the normalized cache both memory and sql even after using the             "client.clearNormalizedCache()", but it seems to work if i use "client.apolloStore().normalizedCache().clearAll()" instead which isn't thread safe

sav007:
Current API returns `ApolloStoreOperation<Boolean>` in `clearNormalizedCache` method call.
This API is a little confusing as user expects that the normalized cache will be cleared right away but instead he must call `ApolloStoreOperation#execute`.

Closes #689


digitalbuddha:
I agree with Ben main thread is a no no. I think we should act similar to shared preferences and have both an apply and a commit. One is async other is sync. Only difference is that sync should throw exception if called from main thread.

sav007:
Yeah, forgot about the SQL. I guess then will close this PR as it doesn't make sense.

PR #695, opened 2017-10-19
Docs

jbaxleyiii:
Moved the (empty) docs into this repo and setup auto deploys with netlify!


docs/_config.yml
```diff
@@ -0,0 +1,88 @@
...
+    - index
+
+github_repo: apollographql/apollo-android
+content_root: docs/source
+
+social_links:
+  github: 'https://github.com/apollographql'
+  twitter: '@apollographql'
+
+# API keys
```
digitalbuddha: Sanity check, not sure what this is exactly but  should api keys be checked into public repo?

docs/circle.yml
```diff
@@ -0,0 +1,26 @@
+machine:
```
digitalbuddha: Currently we use Travis for CI, do we need to introduce circle as well?

stubailo: No - this file is unnecessary

stubailo:
I'm going to merge this for now. We're not using this URL yet for anything, but I want to finish up the migration of the repos etc. -- this merge doesn't mean we need to start using this site for the docs or anything.

PR #697, opened 2017-10-19
Prepare next development version.

sav007:


digitalbuddha:
I'm assuming the deploy docs task will fail until we merge #695

PR #712, opened 2017-10-29
Add support of enum unknown values

sav007:
**Problem**: whenever server returns new unsupported by client enum value it's a breaking changes, as `ResponseReader` tries to map raw string value to appropriate generated enum constant by `.valueOf`.

**Proposal**: to have a fallback value called `$UNKNOWN`, so that any unsupported by client enum value is mapped to it. This will allow to handle such case gracefully.

Examples:

https://github.com/apollographql/apollo-android/pull/712/files#diff-9a0209432c8d9c39a92ea1e525dbee0d

https://github.com/apollographql/apollo-android/pull/712/files#diff-d88c16e23165a268eb39d908f34c5ab7


jeffreydecker:
@sav007 thanks for getting this taken care of 😄  are there equivalent prs for the other platforms by any chance?

sav007:
@jeffreydecker we will sync tomorrow with iOS regarding this issue

PR #725, opened 2017-11-20
NetworkFirstFetcher fix to return network exception if cache fetch fails

drumtoad:
NetworkFirstFetcher javadoc says that in case of cache fetch failure it should rethrow network exception, but it thrown cache exception instead. Reproducible with sample project(with network off and with no cache). So here's fix to pass network exception on cache fetch fail.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/fetcher/NetworkFirstFetcher.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/fetcher/NetworkFirstFetcher.java
@@ -38,12 +38,28 @@ public final class NetworkFirstFetcher implements ResponseFetcher {
           callBack.onResponse(response);
         }
 
-        @Override public void onFailure(@Nonnull ApolloException e) {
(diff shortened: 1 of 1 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/fetcher/NetworkFirstFetcher.java
```diff
@@ -38,12 +38,28 @@ public void interceptAsync(@Nonnull final InterceptorRequest request, @Nonnull f
...
+              }
+
+              @Override public void onFailure(@Nonnull ApolloException cacheException) {
+                callBack.onFailure(networkException);
+              }
+
+              @Override public void onCompleted() {
+                callBack.onCompleted();
+              }
+            });
```
sav007: I guess we need to check if it wasn't disposed before notifying original callback, like:
```
if (!disposed) {
callBack.onFailure(networkException);
}
```

What do you think?

PR #743, opened 2017-12-04
Remove mutation cache root key

sav007:
Refactor ApolloStore to support query operations only.
Remove cache key for mutation.


lincolnq:
@sav007 Can you explain again why a mutation should not update the normalized cache? I don't think I understand. For my use case this is pretty important -- I have mutations which refetch data as part of the mutation, and Apollo should not require additional server round-trips in order to populate that cache. I very well may be confused -- however, my understanding of the principle of the normalized cache is that mutations are supposed to update it -- at least, Apollo's own javascript client seems documented to do so (see https://www.apollographql.com/docs/react/advanced/caching.html#automatic-updates) 

If this is something that Apollo Android will never support, I could use better instructions for building this infrastructure for myself. Is the recommended approach to call apolloStore().writeFragment() when my mutation returns, to populate the cache manually?

PR #751, opened 2017-12-08
Making generated OPERATION_NAME public

### LINKED ISSUE #750: Make generated `static final OperationName OPERATION_NAME` public?
We set up our mockwebserver tests using the query name to match a test json fixture and right now we're using `Strings`. This works but it's prone to errors, it would be helpful if we could access the generated `MyOperationQuery.OPERATION_NAME` static const value at compile time.

I'm aware we can access the name at Runtime via `MyOperationQuery.Builder().build().name().name()` but that doesn't work for annotations.

murki:
Closes #750


sav007:
Thanks!

Issue #757, opened 2017-12-13
Readme Instructions for Http Caching are not up to date


dmathewwws:
Instructions in Readme use old ResponseCacheStore. 

See pull request attached

PR #758, opened 2017-12-13
Update README.md

dmathewwws:
Instructions in Readme use old ResponseCacheStore. Updated to ApolloHttpCache


dmathewwws:
For #757

Issue #763, opened 2017-12-15
ApolloCacheHeaders.DO_NOT_STORE not handled for LruNormalizedCache


wijnanb:
version: 0.4.3

We are using the in-memory cache and are trying to exlude some queries of being cached.

We use the `ApolloCacheHeaders.DO_NOT_STORE` header to achieve this, but the query is still cached.

When debugging the code we noticed that the function `NormalizedCache.merge()` (see [source](https://github.com/apollographql/apollo-android/blob/e2af80859cb49013793459930f0145d30f46abf5/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java#L74)) is never called where the `DO_NOT_STORE` header is handled.

This is how the cache is initialized:

```
EvictionPolicy evictionPolicy = EvictionPolicy.builder()
                .maxSizeBytes(5 * 1024 * 1024) // 5MB in memory cache
                .expireAfterWrite(30, TimeUnit.MINUTES)
                .build();

LruNormalizedCacheFactory normalizedCacheFactory = new LruNormalizedCacheFactory(evictionPolicy);
```

And the header is set to the query like this:

```
CacheHeaders cacheHeaders = CacheHeaders.builder().addHeader(ApolloCacheHeaders.DO_NOT_STORE, "true").build();

apolloClient.query(query)
	.cacheHeaders(cacheHeaders)
	.enqueue(callback);


```

Any ideas on this?

Thanks
Bert

PR #765, opened 2017-12-19
handle ApolloCacheHeaders.DO_NOT_STORE if merging collection of records

wijnanb:
Closes issue #763

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
@@ -102,6 +102,10 @@ public abstract class NormalizedCache {
     checkNotNull(recordSet, "recordSet == null");
     checkNotNull(cacheHeaders, "cacheHeaders == null");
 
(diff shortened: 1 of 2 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.java
```diff
@@ -102,6 +102,10 @@
     checkNotNull(recordSet, "recordSet == null");
     checkNotNull(cacheHeaders, "cacheHeaders == null");
 
+    if (cacheHeaders.hasHeader(ApolloCacheHeaders.DO_NOT_STORE)) {
+      return Collections.emptySet();
+    }
+
```
sav007: I believe we need to move this condition up to and remove it from `SqlNormalizedCache`:  `com.apollographql.apollo.internal.interceptor.ApolloCacheInterceptor#interceptAsync`

Why

wijnanb: Hi @sav007 

I am picking this one up again.

Have been looking in the code on how to implement it higher up in the code, but cannot find how to do it.

Can we merge it like this? It fixes a critical bug for our application where we need to control which queries are NOT cached.

PR #772, opened 2017-12-21
Implement "useJavaBeansSemanticNaming"

### LINKED ISSUE #749: Support for `getFieldName` type methods?
Now that we are not able to use `generateAccessors`, It would be nice to have the option for generating method names prepended with `get` so that in Kotlin consumer code we can still use property access syntax.

E.g.:
`@Nonnull Hero hero();` makes usage like `movie.hero().name() // Kotlin`
versus `@Nonnull Hero getHero();` allows usage like `movie.hero.name // Kotlin`

Any thoughts? Really appreciate everyone's work.

nubstick27:
When Java classes are consumed by Kotlin code, property access syntax is
enabled for any methods with prepended get (e.g. `getFoo()`). This means
that the consuming Kotlin code can access those fields like `thing.foo`.

Code generation with `useJavaBeansSemanticNaming` enabled should modify
any accessor functions only to support this use.

Boolean fields ought to be prefixed with `is` rather than `get` for the
most semantic naming (boxed and unboxed).

The property is added to the ApolloExtension to be passed to the
CodeGenerationContext so that the value can be defined by users of the
library.

Closes #749

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -27,7 +27,8 @@ class GraphQLCompiler {
         nullableValueType = args.nullableValueType,
         ir = ir,
         useSemanticNaming = args.useSemanticNaming,
-        generateModelBuilder = args.generateModelBuilder
+        generateModelBuilder = args.generateModelBuilder,
+        useJavaBeansSemanticNaming = args.useJavaBeansSemanticNaming
     )
     ir.writeJavaFiles(
         context = context,
@@ -84,6 +85,7 @@ class GraphQLCompiler {
       val nullableValueType: NullableValueType,
       val useSemanticNaming: Boolean,
       val generateModelBuilder: Boolean,
+      val useJavaBeansSemanticNaming: Boolean,
       val outputPackageName: String?
   )
 }
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Util.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Util.kt
@@ -14,6 +14,11 @@ private val JAVA_RESERVED_WORDS = arrayOf(
 
 fun String.escapeJavaReservedWord() = if (JAVA_RESERVED_WORDS.contains(this)) "${this}_" else this
 
+fun String.toJavaBeansSemanticNaming(isBooleanField: Boolean): String {
+  val prefix = if (isBooleanField) "is" else "get"
+  return "$prefix${capitalize()}"
+}
+
 fun TypeName.overrideTypeName(typeNameOverrideMap: Map<String, String>): TypeName {
   if (this is ParameterizedTypeName) {
     val typeArguments = typeArguments.map { it.overrideTypeName(typeNameOverrideMap) }.toTypedArray()
--- a/README.md
+++ b/README.md
@@ -437,6 +437,34 @@ apollo {
 }
 ```
 
+### Java Beans Semantic Naming for Accessors
+By default, the generated classes have accessor methods whose names are identical to the name of the Schema field.
+
+```query Foo { bar }```
+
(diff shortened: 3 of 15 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -84,6 +85,7 @@ class GraphQLCompiler {
       val nullableValueType: NullableValueType,
       val useSemanticNaming: Boolean,
       val generateModelBuilder: Boolean,
+      val prependGetter: Boolean,
```
sav007: should we call this: `useJavaBeansSemanticNaming`?

apollo-compiler/src/test/graphql/com/example/prepend_getter/TestOperation.graphql
```diff
@@ -0,0 +1,5 @@
+query TestQuery {
+  hero {
+    name
+  }
+}
```
sav007: Could you please update query to include fragments and inline fragments. I believe there will be some other places where we need to update not just for `Field`

nubstick27: Oh, that's a great callout, thanks.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
nubstick27: 👍

brigade-juni:
Please let me know if you would prefer me to squash everything~

Thanks for your time

nubstick27:
Rebased and re-worded commit

nubstick27:
any guidance on debugging the travis failure?
```
com.apollographql.apollo.ApolloWatcherTest > testQueryWatcherUpdated_Store_write FAILED
    java.lang.AssertionError at ApolloWatcherTest.java:169
        Caused by: java.lang.AssertionError at ApolloWatcherTest.java:169
```
above failure doesn't happen locally for me when i run `.buildscript/integration_tests_composite.sh`

sav007:
Yeah, it's a well known issue with this flaky test, we need to take a look at it. Just try to build again

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Util.kt
```diff
@@ -14,6 +14,8 @@ private val JAVA_RESERVED_WORDS = arrayOf(
 
 fun String.escapeJavaReservedWord() = if (JAVA_RESERVED_WORDS.contains(this)) "${this}_" else this
 
+fun String.toJavaBeansSemanticNaming() = "get${this.capitalize()}"
+
```
digitalbuddha: Is it always going to be get? What if we have a variable named enabled, shouldn't it be `isEnabled`?

nubstick27: Good point -- Booleans ought to be `isFoo` rather than `getFoo` to be most semantic

brigade-juni:
Hi @digitalbuddha -- I've updated as per feedback from you and ivans. Please take another look~

nubstick27:
Hi @digitalbuddha -- I've updated as per feedback from you and ivans. Please take another look~

nubstick27:
I've added a blurb to the README

digitalbuddha:
Thank you for the contribution!

sougat818:
Hey guys, thanks for all the amazing work here. I tried to use this feature on 0.4.3 . However I got the error - 
> Could not set unknown property 'useJavaBeansSemanticNaming' for object of type com.apollographql.apollo.gradle.ApolloExtension.

I couldn't find 0.4.3 version of the apollo-compiler on mvn repos as well. Am I missing something here?

nubstick27:
hello @sougat818 -- I don't believe this commit has been released yet. Are you able to use the 4.4-SNAPSHOT version?

PR #777, opened 2017-12-24
Remove synthetic accessors from source files

### LINKED ISSUE #741: Synthetic accessors need to be removed.
There are right now around 200 synthetic accessor methods in the entire project, which need to be removed in order to reduce the count of methods in the dex file.

VisheshVadhera:
Closes #741 

Also I believe it would be a good idea to create an annotation `@VisibleToAvoidSynthetics`, something similar to the `@Thunk` used in AOSP ([here](https://android.googlesource.com/platform/packages/apps/Launcher3/+/master/src/com/android/launcher3/util/Thunk.java)) and use it for all the members whose visibility has been changed in this PR. This would prevent accidentally changing their visibility back to private in future. 

Is this something that should be added as well?


sav007:
I think it's fine without annotation for now. Unfortunately error-prone doesn't support rule to check synthetic accessors.

Issue #754, opened 2017-12-13
Add support of subscriptions


sav007:
**Protocol description:** https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md

**Public GraphQL server with subscription support** https://github.com/apollographql/GitHunt-API

sav007:
blocked on https://github.com/apollographql/apollo-codegen/issues/316
cc @martijnwalraven

AndreiHarasemiuc:
Hello, sorry to bother, but do you guys have an ETA on this feature?

sav007:
the PR is up, waiting on review, the only missing part for now are tests and javadocs.
Should be merged soon.

bhuvan0616:
Any update on this feature?

devagul93:
Any update on This?

Queike:
Any update on this feature?

pleerock:
Issue seems to be merged but there are no docs on how to use it?

edwrios:
Any update on this feature?

wawhal:
Hey @sav007 where is this documented?
Couldn't find it anywhere in the docs/README.

PR #780, opened 2018-01-02
Introduce subscription call

sav007:
- subscription manager
- web socket subscription transport
- apollo subscription call

Part of #754 

Missing:
- [ ] JavaDocs
- [x] Tests
- [x] sample
- [x] Rx wrappers
  
To test it locally you need to setup:
- https://github.com/apollographql/GitHunt-API
- https://github.com/apollographql/githunt-react

In githunt react example you can add a new comments that will be reflected on android emulator via subscription.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -5,6 +5,7 @@ import com.apollographql.apollo.api.Operation;
 import com.apollographql.apollo.api.OperationName;
 import com.apollographql.apollo.api.Query;
 import com.apollographql.apollo.api.ScalarType;
+import com.apollographql.apollo.api.Subscription;
 import com.apollographql.apollo.api.cache.http.HttpCache;
 import com.apollographql.apollo.api.cache.http.HttpCachePolicy;
 import com.apollographql.apollo.api.internal.Optional;
@@ -22,8 +23,13 @@ import com.apollographql.apollo.internal.ApolloCallTracker;
 import com.apollographql.apollo.internal.ApolloLogger;
 import com.apollographql.apollo.internal.RealApolloCall;
 import com.apollographql.apollo.internal.RealApolloPrefetch;
+import com.apollographql.apollo.internal.RealApolloSubscriptionCall;
 import com.apollographql.apollo.internal.ResponseFieldMapperFactory;
 import com.apollographql.apollo.internal.cache.normalized.RealApolloStore;
+import com.apollographql.apollo.internal.subscription.NoOpSubscriptionManager;
+import com.apollographql.apollo.internal.subscription.RealSubscriptionManager;
+import com.apollographql.apollo.internal.subscription.SubscriptionManager;
+import com.apollographql.apollo.subscription.SubscriptionTransport;
 import com.apollographql.apollo.response.ScalarTypeAdapters;
 
 import java.io.IOException;
@@ -62,7 +68,8 @@ import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
  *
  * <p>See the {@link ApolloClient.Builder} class for configuring the ApolloClient.
  */
-public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutationCall.Factory, ApolloPrefetch.Factory {
+public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutationCall.Factory, ApolloPrefetch.Factory,
+    ApolloSubscriptionCall.Factory {
 
   public static Builder builder() {
     return new Builder();
@@ -82,6 +89,7 @@ public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutati
   private final ApolloCallTracker tracker = new ApolloCallTracker();
   private final List<ApolloInterceptor> applicationInterceptors;
   private final boolean sendOperationIdentifiers;
+  private final SubscriptionManager subscriptionManager;
 
   ApolloClient(HttpUrl serverUrl,
       Call.Factory httpCallFactory,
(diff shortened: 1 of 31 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloSubscriptionCall.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.apollographql.apollo.api.Response;
+import com.apollographql.apollo.api.Subscription;
+import com.apollographql.apollo.exception.ApolloCanceledException;
+import com.apollographql.apollo.exception.ApolloException;
+
+import javax.annotation.Nonnull;
+
+public interface ApolloSubscriptionCall<T> {
+  void start(@Nonnull Callback<T> callback) throws ApolloCanceledException;
```
digitalbuddha: Cancelled (missing T)

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloSubscriptionCall.java
```diff
@@ -0,0 +1,178 @@
...
+  private final AtomicReference<CallState> state = new AtomicReference<>(IDLE);
+  private SubscriptionManagerCallback<T> subscriptionCallback;
+
+  public RealApolloSubscriptionCall(Subscription<?, T, ?> subscription, SubscriptionManager subscriptionManager) {
+    this.subscription = subscription;
+    this.subscriptionManager = subscriptionManager;
+  }
+
+  @Override
+  public void start(@Nonnull Callback<T> callback) throws ApolloCanceledException {
```
digitalbuddha: is `start` standard? thoughts on `connect` or `subscribe`?

sav007: `subscribe` / `unsubscribe` then?

digitalbuddha: any idea what apollo js calls it?

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -231,6 +249,8 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
     Optional<Logger> logger = Optional.absent();
     final List<ApolloInterceptor> applicationInterceptors = new ArrayList<>();
     boolean sendOperationIdentifiers;
+    String subscriptionServerUrl;
+    WebSocket.Factory webSocketFactory;
```
nubstick27: How do you feel about making ApolloClient to have `SubscriptionTransport.Factory` directly to abstract WebSocket (implementation detail) from subscription API?

sav007: Sorry I didn't get it, you want ApolloClient to implement `SubscriptionTransport.Factory`?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloSubscriptionCall.java
sav007: looking into https://github.com/apollographql/apollo-client/blob/master/docs/source/features/subscriptions.md

seems it's called `subscribeToMore`

sav007: but if we call it subscribe then our API will look like this:
```
 application.apolloClient()
        .subscribe(new RepoCommentAddedSubscription(repoFullName))
        .subscribe(new ApolloSubscriptionCall.Callback<RepoCommentAddedSubscription.Data>() {
          @Override public void onResponse(@Nonnull Response<RepoCommentAddedSubscription.Data> response) {
            Log.d("Subscription!!!", response.data().toString());
          }

          @Override public void onFailure(@Nonnull ApolloException e) {
            Log.e("Subscription!!!", "", e);
          }

          @Override public void onCompleted() {
            Log.d("Subscription!!!", "Completed!");
          }
        });
```

where first subscribe returns `ApolloSubscriptionCall` and second is actually start subscription

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
nubstick27: Ah, sorry no -- I meant ApolloClient might hold reference to SubscriptionTransport.Factory directly rather than explicit reference to WebSocket.Factory

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -434,12 +457,26 @@ public ApolloClient build() {
...
       ApolloStore apolloStore = this.apolloStore;
+      Optional<NormalizedCacheFactory> cacheFactory = this.cacheFactory;
+      Optional<CacheKeyResolver> cacheKeyResolver = this.cacheKeyResolver;
       if (cacheFactory.isPresent() && cacheKeyResolver.isPresent()) {
         final NormalizedCache normalizedCache = cacheFactory.get().createChain(RecordFieldJsonAdapter.create());
         apolloStore = new RealApolloStore(normalizedCache, cacheKeyResolver.get(), scalarTypeAdapters, dispatcher,
             apolloLogger);
       }
 
+//      Optional<WebSocket.Factory> webSocketFactory = this.webSocketFactory;
```
digitalbuddha: remove?

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
```diff
@@ -132,6 +133,37 @@ private Rx2Apollo() {
     });
   }
 
+  @Nonnull public static <T> Observable<Response<T>> from(@Nonnull final ApolloSubscriptionCall<T> call) {
```
digitalbuddha: We need to worry about backpressure here, maybe Flowable would be more right

digitalbuddha:
I did not review with fine tooth comb but overall it looks reasonable. Will try to get a few others to look over.

Issue #783, opened 2018-01-10
Serialize custom scalar type list as json


knordbo:
Similar to #752 there is the case where a have a custom scalar type is a list. #752 already fixes the case when its a map, so it is a small fix to make it work for lists as well.

sav007:
Merged

PR #784, opened 2018-01-10
Serialize custom scalar type list as json string

knordbo:
Solves #783


apollo-runtime/src/main/java/com/apollographql/apollo/internal/response/RealResponseReader.java
```diff
@@ -265,9 +265,9 @@ private void checkValue(ResponseField field, Object value) {
   }
 
   @SuppressWarnings("unchecked") private static String normalizeCustomTypeValue(Object value) {
-    if (value instanceof Map) {
+    if (value instanceof Map || value instanceof Iterable) {
```
sav007: I would change it to `value instanceof List` as it always will be a list

PR #788, opened 2018-01-11
add OPERATION_ID as static field.

martinbonnin:
I'm using this to write a OkHttpInterceptor that reads the 'X-APOLLO-OPERATION-ID' and displays debug information based on the Operation


sav007:
You probably forgot to commit test fixtures.

martinbonnin:
Indeed, Thanks !

PR #816, opened 2018-02-02
remove the section about kotlin & plugin order until we figure out wh…

martinbonnin:
…at's exactly happening.

remove the reference to generateAccessors that was removed some time ago


martinbonnin:
See https://github.com/apollographql/apollo-android/issues/817

PR #833, opened 2018-02-25
Refactor code to avoid line length rule violation

ghostbuster91:
The true question is: how the code which fails build got merged at the first place? :thinking:


ghostbuster91:
But it failed my build locally. So you mean that you are not running them on CI, yes?

PR #836, opened 2018-02-26
Remove delays from tests

ghostbuster91:
Should be merged after #834

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/ApolloCancelCallTest.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/ApolloCancelCallTest.java
@@ -13,15 +13,14 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import java.io.IOException;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
+import okhttp3.Dispatcher;
 import okhttp3.OkHttpClient;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 
 public class ApolloCancelCallTest {
-  private static final long TIME_OUT_SECONDS = 3;
   private ApolloClient apolloClient;
   @Rule public final MockWebServer server = new MockWebServer();
   private MockHttpCacheStore cacheStore;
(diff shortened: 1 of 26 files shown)


ghostbuster91:
Can I rewrite ApolloInterceptorTest to use Rx2Apollo?

sav007:
> Can I rewrite ApolloInterceptorTest to use Rx2Apollo?

Sure, if it help to reduce the code and make it simpler go ahead

ghostbuster91:
OK I think I'm done with that. There is still a delay in `ApolloIdlingResourceTest` but it is in different module and that would requires sharing some utils between them, so I will make it in another PR.

apollo-integration/src/test/java/com/apollographql/apollo/ApolloCancelCallTest.java
```diff
@@ -13,23 +13,24 @@
...
   private MockHttpCacheStore cacheStore;
 
   @Before
   public void setup() {
     cacheStore = new MockHttpCacheStore();
-    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    OkHttpClient okHttpClient = new OkHttpClient.Builder()
+        .dispatcher(new Dispatcher(Utils.immediateExecutorService()))
+        .build();
     apolloClient = ApolloClient.builder()
```
sav007: We should set `.dispatcher(Utils.immediateExecutorService())` for `ApolloClient` as well, otherwise it will use the default one: `com.apollographql.apollo.ApolloClient.Builder#defaultDispatcher`

ghostbuster91: Absolutely, good caught :)

ghostbuster91:
Unfortunately I failed in simplifying `ApolloCancelCallTest::cancelCallAfterEnqueueNoCallback`

ghostbuster91:
Can we merge it before any merge conflicts emerge?

apollo-integration/src/test/java/com/apollographql/apollo/ApolloCancelCallTest.java
```diff
@@ -46,30 +49,27 @@ public void cancelCallBeforeEnqueueCanceledException() throws Exception {
...
-        .setBodyDelay(TIME_OUT_SECONDS / 2, TimeUnit.SECONDS));
-
+    OkHttpClient okHttpClient = new OkHttpClient.Builder()
+        .dispatcher(new Dispatcher(Utils.immediateExecutorService()))
+        .build();
+    apolloClient = ApolloClient.builder()
+        .serverUrl(server.url("/"))
+        .okHttpClient(okHttpClient)
+        .httpCache(new ApolloHttpCache(cacheStore, null))
+        .build();
```
sav007: The same issue as above: `.dispatcher(Utils.immediateExecutor())`

ghostbuster91: Actually it wasn't a coincidence that I left it here without `immediateExecutor`. 
We cannot just add here `immediateExecutor` because when doing all sequentially the values will be emmited before cancelling the call, so the assertions will fail. What we need here is an executor which can be controlled, and can fire events like `TestScheduler`.  I tried to write something like that but without luck. So for now at least I think that it has to stay as it is.

PR #840, opened 2018-03-01
Format cache keys as sorted json structure (cross platform cache sharing)

tuncaulubilge:
We have been working on sharing cache between android and react-native in a hybrid app but had some issues with different client libraries having separate approaches.There are some critical differences between `apollo-client` and `apollo-android` in terms of how they handle **CacheKey** in normalized cache. This PR aims to fix this issue by converting android cache key notation to json standards.

Apollo Client (JS) uses json notation while naming their cache keys in store, whereas apollo-android has its own structure without curly brackets or quotes. 

**Sample:**
apollo-client:
`hero({"color":"blue","episode":"JEDI"})`

apollo-android:
`hero(color:blue,episode:JEDI)`

With this PR, both platforms generates the cache key in the same manner. See updated tests for the new cache key formats.

cc @fourlastor who worked with me on this

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/ResponseField.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/ResponseField.java
@@ -169,7 +169,7 @@ public class ResponseField {
    *
    * @param responseName     alias for the result of a field
    * @param fieldName        name of the field in the GraphQL operation
-   * @param conditionalTypes conditional GraphQL types
+   * @param typeConditions conditional GraphQL types
    * @return Field instance representing {@link Type#FRAGMENT}
    */
   public static ResponseField forFragment(String responseName, String fieldName, List<String> typeConditions) {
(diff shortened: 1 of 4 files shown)


apollo-api/src/main/java/com/apollographql/apollo/api/ResponseField.java
```diff
@@ -277,30 +277,47 @@ private String orderIndependentKey(Map<String, Object> objectMap, Operation.Vari
...
+  private String asJsonValue(Object o) {
+    if (o instanceof Boolean
+        || o instanceof Byte
+        || o instanceof Character
+        || o instanceof Double
+        || o instanceof Float
+        || o instanceof Integer
+        || o instanceof Long
+        || o instanceof Short) {
+      return o.toString();
```
sav007: 1. It can be simplified, just check for `instanceof Number` and it will cover `Byte, Double, Float, Integer, Long, Short`.
2. Not sure that `Character` is supported json value type, should it be wrapped in ""?

tuncaulubilge: Good point, fixed in ec8b540

Issue #850, opened 2018-03-07
GraphQL Subscription not support connectionParams.  


jchbh-duplicate:
In subscriptions-transport-ws, the client allow to send connectionParams when connect_init, which is a core function to support JWT or other token for the subscription.

Can we offer something that can send this, and if it is possible, can also expose the socket object too so we can manual reconnect to server when refresh the token.

jchbh-duplicate:
Right now I just copy the WebSocketSubscriptionTransport code and change the send() method to handle OperationClientMessage.Init with another logic.

PR #865, opened 2018-04-03
Add subscription connection parameters as payload of init message

sav007:
Closes https://github.com/apollographql/apollo-android/issues/850


apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.java
```diff
@@ -40,10 +42,19 @@ public String toJsonString() {
 
   public static final class Init extends OperationClientMessage {
     private static final String TYPE = "connection_init";
+    private final Map<String, Object> connectionParams;
+
+    public Init(@Nonnull Map<String, Object> connectionParams) {
```
digitalbuddha: Uppercase method name

Issue #895, opened 2018-04-27
Automatic Websocket Reconnection 


Syl1234:
I sucessfully managed to use ApolloClient in order to deal with GraphQL Subscription over websocket. I can see events onNext happening. Unfortunatly when my server is facing difficulties, ApolloClient is not able to reconnect using the same websocket.

`

     private ApolloSubscriptionCall<Subscription.Data> getMySubscriptionCall() {
        Subscription = Subscription.builder().build();
        return apolloClient.subscribe(Subscription);
    }

/* */

        okHttpClient = new OkHttpClient.Builder()
                .addInterceptor(new LoggingInterceptor())
                .pingInterval(30, TimeUnit.SECONDS)
                .retryOnConnectionFailure(true)
                .build();

        apolloClient = ApolloClient.builder()
                .serverUrl(env.getBaseUrl())
                .okHttpClient(okHttpClient)
                .subscriptionTransportFactory(new WebSocketSubscriptionTransport.Factory(env.getSubscriptionBaseUrl(), okHttpClient))
                .build();

        disposables.add(Rx2Apollo.from(getMySubscriptionCall())
                .subscribeOn(Schedulers.io())
                .subscribeWith(new Subscriber()));
`

Is there any other particular configuration that i must add in order to reconnect the client ?

Thx for advices

sav007:
Have you tried to `unsubscribe` and then `subscribe` after some delay like 10-20 seconds? After 10 sec timeout we will kill web socket connection.

sav007:
Actually I guess we need to add support of heart beat from server side:
https://github.com/apollographql/subscriptions-transport-ws/blob/7482425bb27139d99f45c88ac46c7223c210fbf9/src/message-types.ts#L7

PR #922, opened 2018-05-23
Add support connection keep alive message (aka heartbeat)

sav007:
Add support of heartbeat messages sent by the subscription server.

Closes https://github.com/apollographql/apollo-android/issues/895


sav007:
Not sure that we need the max timeout, as it's totally up to user to decide, if they want never reconnect than so be it. Otherwise what the max should be 30 sec or 1 hr?

The only I think we need to switch off this feature by default, and only enable it if user explicitly provide timeout that `> 0`

digitalbuddha:
cool all set then

Issue #921, opened 2018-05-23
Operation name: header vs body


abredykhin:
I'm using apollo-android 0.5.0 and see that library sends the operation name in header as opposed to including it into the body of my query/mutation. Our server written in Go just sends back `Must provide an operation.` error, and it seems that it doesn't look into headers for this info. 

Any suggestions on how to deal with this?

sav007:
That strange that your server requires it as per spec http://graphql.org/learn/serving-over-http/#post-request it's optional. But we can def add it

abredykhin:
@sav007 Somehow the problem is caused by content type header being spelled incorrectly. Apollo sends 'CONTENT_TYPE' instead of 'Content-Type'. Changing header name fixes the issue

abredykhin:
After extensive debugging I found the root cause for this issue. Our server was incorrectly processing the content type of requests. Apollo sends them as `application/json; charset=utf-8`. After modify this to just `application/json` my requests started to go through. 

Now I'm not really sure if really need that linked PR

PR #924, opened 2018-05-23
Send operation name in the POST payload

sav007:
Closes https://github.com/apollographql/apollo-android/issues/921


abredykhin:
I think fix to `content-type` header is also required to really fix this

Issue #932, opened 2018-05-31
Unable to compile because of java error inside generated apollo class


CrisCR:
I'm facing a strange probleme where I can't run my project because of a local variable which is accessed from within inner class, so it needs to be declare final.

```
if (availabilitiesRanges.defined) {
          writer.writeList("availabilitiesRanges", availabilitiesRanges.value != null ? new InputFieldWriter.ListWriter() {
            @Override
            public void write(InputFieldWriter.ListItemWriter listItemWriter) throws IOException {
              for (List<Integer> $item : availabilitiesRanges.value) {
                listItemWriter.writeList($item != null ? new InputFieldWriter.ListWriter() {
                  @Override
                  public void write(InputFieldWriter.ListItemWriter listItemWriter) throws
                      IOException {
                    for (Integer $$item : $item) { <----- the error
                      listItemWriter.writeInt($$item);
                    }
                  }
                } : null);
              }
            }
          } : null);
        }
```

How can I fix it?

dependencies:
classpath 'com.apollographql.apollo:apollo-gradle-plugin:0.5.1-SNAPSHOT'
implementation 'com.apollographql.apollo:apollo-runtime:0.5.0'

Apollo codegen version: 0.19.1

highthem:
same issue here :(

sav007:
I'm confused why do you have this issue as we have test for this https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/com/example/input_object_type/type/ReviewInput.java#L265

And that block is completely legit  as `$item` is define in `for (List<Integer> $item` that I assume final by definition

sav007:
Seems issue is specific for Java7

PR #934, opened 2018-05-31
Fix issue iterator is not final used in nested inner class

sav007:
The issue is specific to Java7 and can't be reproduced on Java8. Apparently the iterator declared in for loop on Java7 must be with final keyword if it's going to be used inside the nested inner class.

Closes https://github.com/apollographql/apollo-android/issues/932


CrisCR:
Is there a snapshot available for this fix?

Issue #937, opened 2018-06-02
Unable to compile when graphql schema has list of `Date` inside a list


CrisCR:
My graphql schema contains `timeSlots: [[Date!]!]` , which causes a Java compile error
![captureapollo](https://user-images.githubusercontent.com/9980130/40880243-5eb4b656-66ad-11e8-8d3f-8637a08aa808.PNG)
the `java.util.Date` isn't import and it seems that `(CustomType.[DATE!]);` needs to be declare as `(CustomType.DATE);`

Gradle conf:
```
classpath 'com.apollographql.apollo:apollo-gradle-plugin:0.5.1-SNAPSHOT'

apollo {
    customTypeMapping['Date'] = "java.util.Date"
}
implementation 'com.apollographql.apollo:apollo-runtime:0.5.0'
```

Apollo codegen version: 0.19.1

PR #938, opened 2018-06-04
Fix issue with code generation for list of list of custom scalar types

sav007:
Closes https://github.com/apollographql/apollo-android/issues/937

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,18 @@ ApolloClient.builder()
   .build();
 ```
 
+If you have compiler warnings as errors (`options.compilerArgs << "-Xlint" << "-Werror"`)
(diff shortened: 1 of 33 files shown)


digitalbuddha:
fyi ```com.apollographql.apollo.compiler.CodeGenTest > generateExpectedClasses[mutation_create_review] FAILED
    org.junit.ComparisonFailure at CodegenTest.kt:24```

Also seems like a  lot of javadoc errors but that didn't fail the build not sure why. I'm also unclear why it says "Build Successful" at the end but "Build Failed" halfway through

sav007:
We have 2 builds, the last one that says Build Success is for running our integration tests only.

PR #946, opened 2018-06-09
Add docs related to Queries.

GurpreetSK95:
Add documentation for queries:
- Introduction
- Creating queries
- Typed query results
- Support For Cached Responses
- Next steps (Might need to skip/tone down)

- [ ] feature
- [ ] blocking
- [x] docs

### DIFF
--- a/docs/source/essentials/queries.md
+++ b/docs/source/essentials/queries.md
@@ -1,0 +1,280 @@
+---
+title: Queries
+---
+
+Fetching data in a simple, predictable way is one of the core features of Apollo Client. In this guide, you'll learn how to Query GraphQL data and use the result in your application.
+You'll also learn how Apollo-android Client simplifies your data management code by tracking different error states for you.
+
+This page assumes some familiarity with building GraphQL queries. If you'd like a refresher, we recommend [reading this guide](http://graphql.org/learn/queries/) and practicing [running queries in GraphiQL](https://graphql.github.io/swapi-graphql/).
+Since Apollo Client queries are just standard GraphQL, anything you can type into the GraphiQL query explorer can also be put into `.graphql` files in your project.
+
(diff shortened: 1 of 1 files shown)


docs/source/essentials/queries.md
```diff
@@ -0,0 +1,280 @@
...
+            Log.i(TAG, response.toString());
+          }
+
+          @Override public void onFailure(@NotNull ApolloException e) {
+            Log.e(TAG, e.getMessage(), e);
+          }
+        }, uiHandler));
+```
+
+> By default, Apollo will deliver query results on a background, so you have to provide a handler in enqueue if you're using the result to update the UI.
```
digitalbuddha: Maybe add a note about the android support which has a handler you can use to deliver on Main thread

docs/source/essentials/queries.md
```diff
@@ -0,0 +1,280 @@
...
+      }
+    }, uiHandler));
+}
+```
+
+Because the above query won't fetch `appearsIn`, this property is not part of the returned result type and cannot be accessed here.
+
+<h2 id="cache-policy">Support For Cached Responses</h2>
+
+As explained in more detail in [the section on watching queries](watching-queries.html), Apollo-android client allows you to keep a client-side cache of query results, making it suitable for use even while offline.
```
digitalbuddha: I would remove anything about caching and make caching its own document

digitalbuddha:
Actually ignore comment to link to tests as must integration tests are related to caching and we can link there instead

docs/source/essentials/queries.md
GurpreetSK95: On a side note, the `wrap` method(line 34) in ApolloCallback.java under apollo-android-support is used just for tests. Can it be marked `@VisibleForTesting` if it is not supposed to be used as a public API?

docs/source/essentials/queries.md
GurpreetSK95: Done. Will create a new PR after this.

PR #948, opened 2018-06-12
Make IR generation workable anywhere

xrd:
If you run the task `./gradlew :apollo-compiler:updateTestIRFiles` from the command line, IR generation works fine.

If you run from the Android Studio gradle menu, the task fails because the working directory is mistakenly set to `apollo-compiler` (so paths look like `apollo-compiler/apollo-compiler/...` and IR fails). 

This PR makes it so that the task figures out its contextual working directory and does the right thing.

### DIFF
--- a/apollo-compiler/gradle/update-test-IR-files.gradle
+++ b/apollo-compiler/gradle/update-test-IR-files.gradle
@@ -14,6 +14,9 @@ class UpdateTestIRFilesTask extends DefaultTask {
       return
     }
 
+    // When this task runs from the Android Studio gradle task menu, it sets working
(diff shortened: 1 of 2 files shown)


xrd:
@digitalbuddha I added a few more commits to make this less platform specific. Just FYI.

sav007:
@xrd can we merge this PR?

xrd:
@sav007 It works for me both from AS gradle menu and the command line. I say yes.

xrd:
@sav007 Is my comment clear? Do you need more information from me. The PR is ready to merge.

Issue #955, opened 2018-06-16
Mapping an interface with both fields and inline fragments returns `null` if the object isn't an instance of any fragments


benkuhn:
Example:

```
# schema.graphql
interface Fruit { weight: Int! }
type Apple implements Fruit { redness: Float! }
type Banana implements Fruit { curvature: Float! }
type Starfruit implements Fruit { numPoints: Int! }
type Query {
    inventory: [Fruit]!
}
```

```
# FruitQuery.graphql
query FruitQuery {
    inventory {
        weight
        ...on Apple {
            redness
        }
        ...on Banana {
            curvature
        }
        # no query for Starfruit but I still expect to
        # see the weight of any starfruit in the inventory
    }
}
```

The generated mapping code for `Inventory.Mapper.map` will look something like:

```
public Inventory map(reader) {
    final AsApple asApple = reader.readConditional(...)
    if (asApple != null) {
        return asApple;
    }
    // same with asBanana
    return null
}
```

At the end, there should be a line that maps an object with *only* the `weight` field into a concrete subtype of `Inventory`. Otherwise, a response of `{"__typename": "Starfruit", "weight": 1}` will be mapped to `null`.

(Side note: I haven't tested this exact example because I got stuck in Gradle sync hell when I tried to make a minimal repro. In a closed-source codebase on v0.5.0 I have an example that is, AFAICT, exactly the same but with more fragments and the names changed, but it's possible I've missed an essential feature. Happy to do further debugging if someone with a working build can't reproduce this.)

josephyanks:
+1, we are experiencing this same issue. This also apparently doesn't affect the iOS variant of our app. We have a workaround where we specify every single type as a different fragment, but it should be able to fall back on the base type. Can provide code examples as well if need be.

benkuhn:
Ah, here's an example of the bug in the test suite:

https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/fragment/HeroDetails.java#L76

As you can see, the mapper will return null unless the returned object has a typename of `Droid`, even though the query asks for the `name` even of non-Droid characters.

benkuhn:
@sav007 would love any feedback on the PR!

sav007:
[...]

PR #964, opened 2018-06-24
Proof of concept: fix null return value from interface mapper code

benkuhn:
Attempt a fix of #955.

The code is probably very bad because I don't really understand how the Apollo IR/codegen works yet. However, I do think it generates the code I want.

This breaks 10 tests which currently do not generate a base interface implementation (instead returning null if no inline fragment matches). I think all those tests currently encode buggy behavior, though am not sure. If there's an easy way to commit the new generated code to make the tests pass, I can do that.


sav007:
another PR is up https://github.com/apollographql/apollo-android/pull/971 closing this one

PR #971, opened 2018-06-30
Fix issue with inline fragments mapped to null

sav007:
If response is an instance of inline fragment that is not in the requested list we map it null.
These changes generates surrogate inline fragment class that mapper will fallback to if response typename doesn't match to requested inline fragment conditions.

Example: https://github.com/apollographql/apollo-android/pull/971/files#diff-4e9825e5c07ebb7d5462ee1ca6f79e46

Closes https://github.com/apollographql/apollo-android/issues/955


digitalbuddha:
looks good.

benkuhn:
Awesome!! Thanks so much for helping with this--really happy that that I can backwards-compatibly add new interface subtypes now :)

PR #1002, opened 2018-07-20
Update cache section

wtrocki:
## Motivation

Documentation snippets for HTTP cache using invalid directory. By default application do not have access to /cache directory. Caching will fail silently.
Common aproach for cache is to use application directory as this space is reserved for the application and it will be removed along with app. 

This simple suggestion to documentation will help developers to get cache working out of the box and with the most appropriate folder.

## Issue

AppolloHttpCache

screen shot 2018-07-20 at 11 50 19 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ Raw HTTP Response Cache:
 
 ```java
 //Directory where cached responses will be stored
-File file = new File("/cache/");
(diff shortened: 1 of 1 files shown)


digitalbuddha:
Thank you for your contribution

Issue #1001, opened 2018-07-20
RuntimeException: Unsupported record value type: class java.util.LinkedHashMap


tuncaulubilge:
Trying to use a custom scalar type but normalisation inside SqlNormalizedCache fails with the following error inside `RecordFieldJsonAdapter`

```
RuntimeException: Unsupported record value type: class java.util.LinkedHashMap
```
The data it's trying to parse is a mixed json array, and we are serialising it as a string using CustomTypeAdapters. However,  RecordFieldJsonAdapter used by SqlNormalizedCache does not use custom type adapters and tries to serialise it as a LinkedHashMap. 

Here is a sample from our json:
```
      "byline": [
        {
          "name": "author",
          "attributes": {
            "slug": "sample-author"
          },
          "children": [
            {
              "name": "text",
              "attributes": {
                "value": "Sample Author"
              },
              "children": []
            }
          ]
        },
        {
          "name": "inline",
          "children": [
            {
              "name": "text",
              "attributes": {
                "value": ", Sample title"
              },
              "children": []
            }
          ]
        }
      ],
```
This `byline` has a custom scalar type defined in the schema and being parsed as a string. We don't want to use strict types on this object as it's highly volatile. Is there a way to customize json serialisation? Am I missing something in the schema?

sav007:
If you register custom type adapter `com.apollographql.apollo.response.CustomTypeAdapter` in `T decode(@NotNull CustomTypeValue value)` you will get `com.apollographql.apollo.response.CustomTypeValue.GraphQLJsonString` that is JSON string. 
Then it's up to you to define how to parse it and convert to custom type you want.

tuncaulubilge:
That's exactly what I'm doing, and apollo is able to parse it as a string.

The problem occurs when apollo tries to persist this data using the SqlNormalizedCache. So, my guess is apollo is not using the custom type adapters while trying to serialise data for persistence.

PR #1010, opened 2018-07-25
Add LinkedHashMap type support to RecordFieldJsonAdapter

tuncaulubilge:
See: https://github.com/apollographql/apollo-android/issues/1001

RecordFieldJsonAdapter does not support LinkedHashMap at the moment, so it's unable to parse json objects if no custom types are defined. This bug is blocking SqlNormalisedCache from persisting these objects to the database.


apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.java
```diff
@@ -84,6 +85,15 @@ private static void writeJsonValue(Object value, JsonWriter jsonWriter) throws I
         writeJsonValue(item, jsonWriter);
       }
       jsonWriter.endArray();
+    } else if (value instanceof LinkedHashMap) {
```
sav007: I think we should use more generic type `Map` instead as we can easily change internal logic and use other  implementation of `Map` in future.

PR #1025, opened 2018-08-21
Stop subscriptions

JeffMcKnight:
Add methods to start and stop the GraphQL subscription manager

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -153,6 +153,20 @@ public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutati
   }
 
   /**
+   * Set {@link SubscriptionManager} to a connectible state.
+   */
+  public void startSubscriptions() {
+    subscriptionManager.start();
+  }
+
+  /**
+   * Unsubscribe from all active subscriptions, and disconnect the web socket.
+   */
+  public void stopSubscriptions() {
+    subscriptionManager.stop();
+  }
+
+  /**
    * @return The default {@link CacheHeaders} which this instance of {@link ApolloClient} was configured.
    */
   public CacheHeaders defaultCacheHeaders() {
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
@@ -97,6 +97,34 @@ public final class RealSubscriptionManager implements SubscriptionManager {
     });
   }
 
+  /**
+   * Set the {@link RealSubscriptionManager} to a connectible state. It is safe to call this method
+   * at any time.  Does nothing unless we are in the stopped state.
+   */
+  @Override
+  public void start() {
+    synchronized (this) {
+      if (state == State.STOPPED) {
+        setStateAndNotify(State.DISCONNECTED);
+      }
+    }
+  }
+
+  /**
+   * Unsubscribe from all active subscriptions, and disconnect the web socket.  It will not be
+   * possible to add new subscriptions while the {@link SubscriptionManager} is stopping
+   * because we check the state in {@link #doSubscribe(Subscription, Callback)}
+   */
+  @Override
+  public void stop() {
+    setStateAndNotify(State.STOPPING);
+    for (SubscriptionRecord eachSubscriptionRecord : subscriptions.values()) {
+      doUnsubscribe(eachSubscriptionRecord.subscription);
+    }
+    disconnect(true);
+    setStateAndNotify(State.STOPPED);
+  }
+
   public void addOnStateChangeListener(@NotNull OnStateChangeListener onStateChangeListener) {
     onStateChangeListeners.add(checkNotNull(onStateChangeListener, "onStateChangeListener == null"));
   }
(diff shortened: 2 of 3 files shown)


sav007:
Could you please elaborate more on `start` feature?

I can understand why `stop`  all subscriptions makes sense, but what is the point of having `start`?

JeffMcKnight:
The `start` feature re-enables the auto-connect feature in `doSubscribe()`, which we disable when we call `stop()` (by setting the state to `STOPPING`/`STOPPED`).  We want to disable the auto-connect feature so that new `subscribe()` calls will not cause `transport.connect()` to be called.  

The use case where we need it is when the Android app is stopping, in which case we want to close the web socket gracefully, and be sure it isn't inadvertently restarted.  I think the edge case we're concerned about is when a user navigates to a Fragment that initiates a subscription in `onStart()`, and then immediately exits the Activity (which would call `stop()` in the Activity `onStop()`) .  In that case, there is a possible race condition between the `subscribe` call and the `doUnsubscribe()` calls in `stop()` (since `subscribe()` runs `doSubscribe()` on the `dispatcher` thread pool).

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -152,6 +152,20 @@ public static Builder builder() {
     return new RealApolloSubscriptionCall<>(subscription, subscriptionManager);
   }
 
+  /**
+   * Set {@link SubscriptionManager} to a connectible state.
+   */
+  public void startSubscriptions() {
```
sav007: My concern is regarding naming here. `startSubscriptions` is misleading as first thought that I had by looking at API was impression that all subscriptions that were active before and stopped via `stopSubscriptions ` will be re-started again.

While `stopSubscriptions` is partially correct (except it disables any new subscriptions), as it really stops all active subscriptions.

Not sure if `enableSubscriptions / disableSubscriptions` is better.

@digitalbuddha do you have any ideas how to make this API better?

JeffMcKnight: That naming seems good.  Changes pushed.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
```diff
@@ -106,6 +135,10 @@ public void removeOnStateChangeListener(@NotNull OnStateChangeListener onStateCh
   }
 
   void doSubscribe(Subscription subscription, SubscriptionManager.Callback callback) {
+    if (state == State.STOPPING || state == State.STOPPED) {
+      callback.onError(new ApolloSubscriptionException("Illegal state: " + state.name()));
```
sav007: Can we add more descriptive message. As it will be reported to user so we want to make it more user friendly

sav007:
Sorry for the delay was off the grid, could you please add unit tests as well

sav007:
And try to rebase to the latest master, not sure why CI is failing

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
onlynone: All set. Let me know what you think of the current message.

onlynone:
The code has been rebased on top of master, unit tests added, and error message extended to be more informative to the user. Let me know if there's anything else to do.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
```diff
@@ -97,6 +97,35 @@ public void run() {
...
+   * {@link #disconnect(boolean)} because we want to disconnect even if, somehow, a new subscription
+   * is added while or after we are doing the {@link #doUnsubscribe(Subscription)} loop.
+   */
+  @Override
+  public void stop() {
+    setStateAndNotify(State.STOPPING);
+    for (SubscriptionRecord eachSubscriptionRecord : subscriptions.values()) {
+      doUnsubscribe(eachSubscriptionRecord.subscription);
+    }
+    disconnect(true);
```
sav007: Should this be synchronized?

```
    synchronized (this) {
      setStateAndNotify(State.STOPPING);
....
      disconnect(true);
    }
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
```diff
@@ -131,7 +166,7 @@ void doUnsubscribe(Subscription subscription) {
     SubscriptionRecord subscriptionRecord;
     synchronized (this) {
       subscriptionRecord = subscriptions.remove(subscriptionId);
-      if (subscriptionRecord != null && state == State.ACTIVE) {
+      if ((subscriptionRecord != null) && (state == State.ACTIVE || state == State.STOPPING)) {
         transport.send(new OperationClientMessage.Stop(subscriptionId));
       }
```
sav007: I think that we shouldn't call `startInactivityTimer ` in case of `state == State.STOPPING`. Doesn't make any sense right?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
onlynone: yeah, looks like it.

onlynone:
I'm not sure what's wrong with the build, the errors and warnings seem unrelated to the code change. @sav007 any idea?

sav007:
Yeah seems like it was a bad merge, will fix master

sav007:
I will merge this PR, even it's red. As I need to do some small changes to SubscriptionManager how  onTerminated callback is propagated.

digitalbuddha:
Do it!

Issue #698, opened 2017-10-21
Support feature modules


yenkenobi:
We have an implementation of instant app that uses feature modules instead of library modules. Currently, we have a workaround where the feature module inherits the library module that utilizes the apollo plugin. However, this has increased the build time and has increased complexity of the project. It would be nice to add plugin support for the feature module type.

digitalbuddha:
I'm not terribly familiar with what this would take. Would love suggestions or more information from community. @marwanad was gracious enough to write the current gradle plugin, not sure if he's familiar with changes needed. @ivans do you instant apps?

marwanad:
I'm not entirely familiar with feature modules. 

By the looks of it, it seems like another Gradle plugin that's applied so simply adding a check: `project.plugins.hasPlugin("com.android.feature")`  before applying the plugin [here](https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy#L41) should do the trick.

theyhoang:
@marwanad just tried to apply such and added the plugin locally:

```
Failed to notify ProjectEvaluationListener.afterEvaluate(), but primary configuration failure takes precedence.
java.lang.IllegalStateException: compileSdkVersion is not specified.
```
Odd since compileSdkVersion is definitely being specified in the feature module. Also tried updating the Android gradle plugin version, but no luck :/

jaredsburrows:
@theyhoang Is this the error you see when applying it to the feature module?

agicquel:
Still no support for feature module ?

kassim:
library modules work fine with Apollo's Gradle plugin - do we know what the actual issue is with Feature Modules?

digitalbuddha:
closing due to inactivity, pls post comment if someone knows what is wrong or what we need to do to fix it :-)

agicquel:
Hello there. I've been looking into the code to find out what is causing the bug with android feature module. After searching with the error message "Apollo plugin couldn't be applied. The Android or Java plugin must be configured first", I've find the class ApolloPlugin.groovy of the package "com.apollographql.apollo.gradle" inside the apollo-gradle-plugin folder.

The method apply wasn't checking the Android Feature plugin : 
```
@Override
  void apply(Project project) {
    this.project = project
    if (project.plugins.hasPlugin(AppPlugin)
        || project.plugins.hasPlugin(LibraryPlugin)
[...]

PR #1067, opened 2018-09-28
[ISSUE-698] Added support for Instant Apps

16BitsOfAwesome:
Issue #698 

I started working with feature modules for my app, and Apollo not supporting it has become a blocker. As explained in https://developer.android.com/topic/google-play-instant/getting-started/feature-plugin, developers are expected to add `apply plugin: 'com.android.feature'` to the top of their feature's build.gradle. Currently, Apollo only recognizes `com.android.application` and the other java plugins.

Here is the suggested edit from that ticket.


16BitsOfAwesome:
Thanks @digitalbuddha! Is there a particular process to get his merged in?

sav007:
Merged should be available in snapshot version soon.

PR #1073, opened 2018-10-09
Rx support for ApolloStoreOperation

jchudev:
Adding Rx support for ApolloStoreOperation. If a user needs to perform any operations with RxJava on the store they will be able to just use RxApollo

- [x] feature
- [ ] blocking
- [ ] docs

### DIFF
--- a/apollo-rx-support/src/main/java/com/apollographql/apollo/rx/RxApollo.java
+++ b/apollo-rx-support/src/main/java/com/apollographql/apollo/rx/RxApollo.java
@@ -6,6 +6,7 @@ import com.apollographql.apollo.ApolloPrefetch;
 import com.apollographql.apollo.ApolloQueryWatcher;
 import com.apollographql.apollo.ApolloSubscriptionCall;
 import com.apollographql.apollo.api.Response;
+import com.apollographql.apollo.cache.normalized.ApolloStoreOperation;
 import com.apollographql.apollo.exception.ApolloException;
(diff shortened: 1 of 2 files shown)


apollo-rx-support/src/main/java/com/apollographql/apollo/rx/RxApollo.java
```diff
@@ -170,6 +171,33 @@ private RxApollo() {
...
 
+  /**
+   * Converts an {@link ApolloStoreOperation} to a Observable.
+   *
+   * @param operation        the ApolloStoreOperation to convert
+   * @param <T>              the value type
+   * @param backpressureMode The {@link rx.Emitter.BackpressureMode} to use.
+   * @return the converted Observable
+   */
+  @NotNull public static <T> Observable<T> from(@NotNull final ApolloStoreOperation<T> operation,
```
digitalbuddha: can a store operation return more than 1 result?

jchudev: https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/cache/normalized/ApolloStoreOperation.java#L82

Seems to only call `onSuccess` once for 1 result

digitalbuddha: I think we can get away with a rx single with no Backpressure

Issue #1066, opened 2018-09-28
difficulty making this work with kotlin dsl/gradle 4.10.2


xenoterracide:
ok, to start this conversation, it's not that I can't make it work at all... it works with *my* plugin, which apply's JavaLibrary, but if I don't use my plugin, it doesn't think `java-library` is loaded.

plugins should react to others
one can do:
```
plugins.withId("some.other.plugin") {
    // action
}
```
action will be executed *if* and  *once*  that plugin is applied

```
buildscript {
    repositories {
        jcenter()
    }
    dependencies {
        classpath( "com.apollographql.apollo:apollo-gradle-plugin:0.5.0")
    }
}

plugins {
    `java-library`
    id("com.xenoterracide.gradle.sem-ver").version("0.5.3")
    id("com.apollographql.android")

}
```

results in

```
> Configure project :
Evaluating root project 'apollo-client' using build file '/Users/calebcushing/IdeaProjects/graphql2redox/modules/apollo-client/build.gradle.kts'.

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/calebcushing/IdeaProjects/graphql2redox/modules/apollo-client/build.gradle.kts' line: 10

* What went wrong:
Plugin [id: 'com.apollographql.android'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (plugin dependency must include a version number for this source)
```
so I add (per gradle docs) to `settings.gradle.kts`
```
pluginManagement {
    resolutionStrategy.eachPlugin {
        logger.info("{}", requested.id)
        if (requested.id.namespace == "com.apollographql") {
            useModule("com.apollographql.apollo:apollo-gradle-plugin:0.5.0")
        }
    }

    repositories {
        gradlePluginPortal()
        jcenter()
    }
}
```
which results in

```
An exception occurred applying plugin request [id: 'com.apollographql.android', artifact: 'com.apollographql.apollo:apollo-gradle-plugin:0.5.0']
> Failed to apply plugin [id 'com.apollographql.android']
   > Apollo plugin couldn't be applied. The Android or Java plugin must be configured first
```

had [a discussion on gradle community slack](https://gradle-community.slack.com/archives/CA7UM03V3/p1538143791000100?thread_ts=1538143527.000100&cid=CA7UM03V3)

> plugins should react to others
> one can do:
> ```
> plugins.withId("some.other.plugin") {
>     // action
> }
> ```
[...]

PR #1120, opened 2018-11-21
Make plugin work with kotlin DSL/4.10.2

lgawin:
I hope this is what was discussed with gradle community #1066 to make the plugin immune to order of applying and what will enable `apollo-gradle-plugin` to work with Kotlin DSL gradle scripts (`build.gradle.kts`).

As it is commented in #1066, it requires some configuration in **root** `settings.gradle.kts` (or just `settings.gradle`) to indicate repository where plugin is available

```
pluginManagement {
    resolutionStrategy.eachPlugin {
        if (requested.id.namespace == "com.apollographql") {
            useModule("com.apollographql.apollo:apollo-gradle-plugin:$requested.version")
        }
    }

    repositories {
        gradlePluginPortal()
        jcenter()
    }
}
...
```

and then one can apply the plugin in submodule (e.g. `graphql/build.gradle.kts`) this way:

```
plugins {
    kotlin("jvm")
    id("com.apollographql.android") version "put-version-here"  // <<<
}

dependencies {
    implementation(kotlin("stdlib-jdk7"))
    implementation("com.apollographql.apollo:apollo-runtime:put-version-here") // <<<
    ...
}
...
```

Plugin should be used with any plugin that applies `java-base` plugin - those are: `java`, `java-library`, `kotlin`, all the android ones `com.android.application`, `com.android.library`, and more (instant app, feature, ...).
Otherwise it will fail with error `"Apollo plugin couldn't be applied without Android or Java or Kotlin plugin."`

### DIFF
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy
@@ -1,8 +1,6 @@
 package com.apollographql.apollo.gradle
 
-import com.android.build.gradle.AppPlugin
-import com.android.build.gradle.InstantAppPlugin
-import com.android.build.gradle.LibraryPlugin
+
 import com.android.build.gradle.api.BaseVariant
 import com.apollographql.apollo.compiler.GraphQLCompiler
 import com.google.common.base.Joiner
@@ -15,8 +13,6 @@ import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.internal.AbstractTask
 import org.gradle.api.internal.file.FileResolver
-import org.gradle.api.plugins.JavaLibraryPlugin
-import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.compile.AbstractCompile
 import org.gradle.api.tasks.compile.JavaCompile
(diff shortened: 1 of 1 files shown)


apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy
```diff
@@ -40,11 +36,14 @@ class ApolloPlugin implements Plugin<Project> {
   @Override
   void apply(Project project) {
     this.project = project
-    if (hasAndroidPlugin(project) || hasJavaPlugin(project)) {
+    project.plugins.withId("java-base") {
```
sav007: Interesting, could you please point to any documentation regarding `java-base`. Is it true that android projects has `java-base` plugins? What about pure kotlin projects?

lgawin: The idea comes from this comment: https://github.com/apollographql/apollo-android/issues/1066#issue-364923607, so from the conversation @xenoterracide had with someone from gradle community. 
That mentions:

> that you may really want to depend on java-base that both java/-library extend

nothing about android plugins, but I checked that (see below).

For `JavaPlugin` and `JavaLibrary` it obviously does the thing:

- `JavaPlugin` (`` `java` ``) - applies `JavaBasePlugin`
https://github.com/gradle/gradle/blob/master/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaPlugin.java#L245

- `JavaLibraryPlugin` (`` `java-library` ``) - applies `JavaPlugin`
https://github.com/gradle/gradle/blob/master/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaLibraryPlugin.java#L53

and for android-specific plugins:
- `AppPlugin` (`` `com.android.application` ``)
- `LibraryPlugin` (`` `com.android.library` ``)
- `InstantAppPlugin` (`` `com.android.instantapp` ``)

all above inherit from android's `BasePlugin` (`com.android.build.gradle.BasePlugin`) which applies `JavaBasePlugin` in `com.android.build.gradle.BasePlugin#configureProject()`  method call

screen shot 2018-11-23 at 06 30 15

which is called for all following locations:

screen shot 2018-11-23 at 06 32 46


(this is for android gradle plugin 2.3.3 which is currently used in the project - btw. it's quite outdated, latest stable is 3.2.1 at the moment and there're new plugins added)

- `KotlinPlugin` - applies `BaseJavaPlugin`
https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPlugin.kt#L345

Issue #1055, opened 2018-09-11
Support APQs (automated-persisted queries) in Android


toby78:
Feature request:

Support APQs (automated-persisted queries) in Android client - as only GET will be supported for CDN caching (e.g. Akamai, Centurylink etc...)

eberhara:
+1 plsss

sav007:
Just for the reference: we have feature persisted queries but it not automatic. The hash of the query is static, generated at build time, the previous idea was that this mapping of hash to query collected from the client ant stored on the server side at compile time.

In order to support automatic persisted query we can reuse the hash that we are currently generating: 
https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/com/example/hero_name/TestQuery.java#L30

Couple links for the ref:
https://github.com/apollographql/apollo-link-persisted-queries#protocol
https://blog.apollographql.com/persisted-graphql-queries-with-apollo-client-119fd7e6bba5
https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de

halysongoncalves:
How to change the client to use GET http?

sav007:
right now we can't, it's a new future to be added.

toby78:
I have started adding it here:

https://github.com/toby78/apollo-android/blob/persisted_queries/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java

...due to lack of time this can take a while. I already added a testcase with variables.

nightbear1009:
hi  is there any example to use http GET with persisted Queries?
I have already .enableAutoPersistedQueries(true)  but the request is still http POST

PR #1137, opened 2018-11-29
Add support for Auto Persisted Queries

sav007:
Add support of Auto Persisted Queries for Apollo Link protocol https://github.com/apollographql/apollo-link-persisted-queries#protocol

There are 2 features missing from the original issue:
- Use Get requests (will probably extracted as separate issue)
- Auto disable Auto Persisted Queries whenever server returns `PersistedQueryNotSupported`

Part of https://github.com/apollographql/apollo-android/issues/1055

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
@@ -71,7 +71,7 @@ public class ApolloAutoPersistedQueryInterceptor implements ApolloInterceptor {
             return Optional.of(request.toBuilder().sendQueryDocument(true).build());
           }
 
-          if (isPersistedQueryNotFound(response.errors()) || isPersistedQueryNotSupported(response.errors())) {
+          if (isPersistedQueryNotSupported(response.errors())) {
             // TODO how to disable Automatic Persisted Queries in future and how to notify user about this
             logger.e("GraphQL server doesn't support Automatic Persisted Queries");
             return Optional.of(request.toBuilder().sendQueryDocument(true).build());
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
@@ -1,0 +1,102 @@
+package com.apollographql.apollo.internal.interceptor;
+
+import com.apollographql.apollo.api.Error;
+import com.apollographql.apollo.api.Response;
+import com.apollographql.apollo.api.internal.Function;
+import com.apollographql.apollo.api.internal.Optional;
+import com.apollographql.apollo.exception.ApolloException;
+import com.apollographql.apollo.interceptor.ApolloInterceptor;
+import com.apollographql.apollo.interceptor.ApolloInterceptorChain;
+import com.apollographql.apollo.internal.ApolloLogger;
+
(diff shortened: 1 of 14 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
```diff
@@ -0,0 +1,102 @@
...
+    return response.parsedResponse.flatMap(new Function<Response, Optional<InterceptorRequest>>() {
+      @NotNull @Override public Optional<InterceptorRequest> apply(@NotNull Response response) {
+        if (response.hasErrors()) {
+          if (isPersistedQueryNotFound(response.errors())) {
+            logger.w("GraphQL server couldn't find Automatic Persisted Query for operation name: "
+                + request.operation.name().name() + " id: " + request.operation.operationId());
+            return Optional.of(request.toBuilder().sendQueryDocument(true).build());
+          }
+
+          if (isPersistedQueryNotFound(response.errors()) || isPersistedQueryNotSupported(response.errors())) {
```
digitalbuddha: Is first condition same as one above?

sav007: oh, my bad oversight. Thx for catching it.

rubensousa:
Why isn't OPERATION_ID the hash of the query document?
I can't manage to generate a valid sha256 hash that's the same as any OPERATION_ID for the queries that we have in the app I'm working.

We're doing APQ by calculating the sha256 hash of the request body in a custom Interceptor for now, but we want to avoid calculating the hash again since we already have OPERATION_ID.

Didn't you face this issue while working on this feature here?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -138,33 +138,29 @@ Call httpCall(Operation operation, CacheHeaders cacheHeaders) throws IOException
...
+    jsonWriter.name("extensions")
+        .beginObject()
+        .name("persistedQuery")
+        .beginObject()
+        .name("version").value(1)
+        .name("sha256Hash").value(operation.operationId())
+        .endObject()
+        .endObject();
+    if (writeQueryDocument) {
+      jsonWriter.name("query").value(operation.queryDocument().replaceAll("\\n", ""));
```
rubensousa: Are you replacing the newlines here because operation id is not the hash of operation.queryDocument()?

sav007:
@rubensousa are you referring to this issue https://github.com/apollographql/apollo-android/issues/1155 ?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
rubensousa: Operation ID is the hash of the query document, right? If the query document you are sending has new lines removed, won't the hash be different? Or is the hash calculated with the new lines already removed?

rubensousa:
@sav007 Yes, that's the problem I was talking here. Let's continue discussion over there.

looploooop:
#1161

donaldarmstrong:
This change prevents using the apollo-android client to be used against AWS AppSync.  The resulting POST request result in a 400 from AWS. The offending block in the post body that creates the 400 is 
```
"extensions": {
    "persistedQuery": {
      "version": 1,
      "sha256Hash": "d25023a7359ae2dcab513cfc2d4e640bbcaddcac1c9a55529957e0542b0ca962"
    }
  },
  ```

Issue #1127, opened 2018-11-26
Add kotlin Coroutines wrappers for operations 


digitalbuddha:
Similar to rx support I'll be adding kotlin support to allow defferred apollo operations

tylerbwong:
Hey @digitalbuddha, I'm curious about what this would look like for the most basic use-case. Would it be something like this?
```kt
suspend fun <T> ApolloCall<T>.await(): T? = suspendCancellableCoroutine { continuation ->
    enqueue(
        object : ApolloCall.Callback<T>() {
            override fun onResponse(response: Response<T>) = continuation.resume(response.data())
            override fun onCanceledError(ex: ApolloCanceledException) { continuation.cancel(ex) }
            override fun onFailure(e: ApolloException) = continuation.resumeWithException(e)
            override fun onNetworkError(e: ApolloNetworkException) = continuation.resumeWithException(e)
            override fun onParseError(e: ApolloParseException) = continuation.resumeWithException(e)
            override fun onHttpError(e: ApolloHttpException) = continuation.resumeWithException(e)
        }
    )
}
```

martinbonnin:
I would expect more something along `Deferred<T>` ? 

Maybe 

     object CoroutineApollo {
         fun from( ApolloCall<T> call): Defferred<T>   
     } 

We actually need this as well, I might try to open a first PR in the coming days.

martinbonnin:
Actually, since ApolloCall are more like `Observables` than `Singles`, I guess that translates in coroutines to `Channels` and not `Deferred` ?

martinbonnin:
A very naive first draft: => https://github.com/martinbonnin/apollo-android/commit/92c3cce56bafab8f6eb6b1631417e88df13bc469

As said above, this doesn't work as soon as the call emits multiple values.

martinbonnin:
First try at Channels:

    fun <T> ApolloCall<T>.toChannel(): Channel<Response<T>> {
        val channel = Channel<Response<T>>(Channel.UNLIMITED)
        enqueue(object : ApolloCall.Callback<T>() {
            override fun onResponse(response: Response<T>) {
                channel.offer(response)
            }
    
            override fun onFailure(e: ApolloException) {
                channel.close()
            }
        })
    
        return channel
    }
    
This seems to work pretty smoothly. I'm not 100% sure about error handling yet

digitalbuddha:
Similar what to do with cancelations?

martinbonnin:
[...]

PR #1169, opened 2019-01-03
Coroutines extensions

martinbonnin:
This adds the following coroutines extensions and their unit tests based on the rxjava2 ones:

    ApolloWatcher.toChannel()
    ApolloCall.toChannel()
    ApolloPrefetch.toJob()


apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -0,0 +1,68 @@
...
+    channel.invokeOnClose {
+        cancel()
+    }
+    enqueueAndWatch(ChannelCallback(channel))
+
+    return channel
+}
+
+fun ApolloPrefetch.toJob(): Job {
+    val deferred = CompletableDeferred<Unit>()
```
sav007: should we cancel `ApolloPrefetch` call whenever job is canceled via `invokeOnCompletion `?

apollo-integration/src/test/java/com/apollographql/apollo/CoroutinesApolloTest.kt
```diff
@@ -0,0 +1,171 @@
...
+                .okHttpClient(okHttpClient)
+                .normalizedCache(LruNormalizedCacheFactory(EvictionPolicy.NO_EVICTION), IdFieldCacheKeyResolver())
+                .build()
+    }
+
+    @Test
+    fun callProducesValue() {
+        server.enqueue(mockResponse(FILE_EPISODE_HERO_NAME_WITH_ID))
+
+
```
sav007: nit: extra line

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
martinbonnin: Good catch, I'll add this

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.apollographql.apollo.exception.ApolloException
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.Channel
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+private class ChannelCallback<T>(val channel: Channel<Response<T>>): ApolloCall.Callback<T>() {
+
+    override fun onResponse(response: Response<T>) {
+        channel.offer(response)
```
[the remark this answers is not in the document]
martinbonnin: This is why I went with UNLIMITED actually:

> When `capacity` is [Channel.UNLIMITED] -- it creates `LinkedListChannel`.
>   This is a channel with linked-list buffer of a unlimited capacity (limited only by available memory).
>   Sender to this channel never suspends and [offer] always returns `true`.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    override fun onStatusEvent(event: ApolloCall.StatusEvent) {
+        if (event == ApolloCall.StatusEvent.COMPLETED) {
+            channel.close()
+        }
+    }
+}
+
+fun <T> ApolloCall<T>.toChannel(): Channel<Response<T>> {
+    val channel = Channel<Response<T>>(Channel.UNLIMITED)
```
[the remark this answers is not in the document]
sav007: Yeah, `RENDEZVOUS ` is not permitted. But if you provide for example `10` if I'm not wrong it will buffer only last 10 messages, but I can be wrong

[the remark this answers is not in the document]
sav007: And I see that `CONFLATED ` can be very useful in many cases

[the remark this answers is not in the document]
martinbonnin: I could make an enum that maps to either `UNLIMITED` or `CONFLATED` and have that as an argument ? 
Would there be a use case for passing an arbitrary capacity like`10` ? That complexifies quite a bit the handling of `channel.offer`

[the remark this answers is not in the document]
sav007: Yeah, I think for now it's fine to support only 2, to make it simple

sav007:
I think one more extension is missing for `ApolloSubscriptionCall` ?

martinbonnin:
I added the `capacity` param, some javadoc and ApolloSubscriptionCall.toChannel() as well.

bhuvan0616:
I have a doubt.
Channel is for continuous communication right? If yes, it can be implemented for subscription. But, for query and mutation suspendCancellableCoroutine is good i think.

martinbonnin:
@bhuvan0616 ApolloCall can emit 2 values if you want to retrieve cached + network responses. 

We could certainly add `ApolloCall.toDeferred()` for simple use cases where only one value is expected. Would that help ?

bhuvan0616:
> @bhuvan0616 ApolloCall can emit 2 values if you want to retrieve cached + network responses so you need a Channel if you want to be able to retrieve both.
> 
> We could certainly add `ApolloCall.toDeferred()` for simple use cases where only one value is expected. Would that help ?
> 
> See also discussion there #1127

Thanks I understood. if you provide Deferred object, the async runs on IO context?

martinbonnin:
@bhuvan0616 not sure I understand. The use case would be like this:

        val deferredResponse = apolloClient.query(EpisodeHeroNameQuery(Input.fromNullable(Episode.EMPIRE))).toDeferred()
        deferredResponse.await()

sav007:
hey @digitalbuddha any feedback on this?

digitalbuddha:
I've been hoping to get someone that dues Coroutines to review. I'll compare to retrofit impl tomorrow

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -0,0 +1,174 @@
...
+ * to retrieve cached and network response, use {@link toChannel} instead.
+ *
+ * @param <T>  the value type.
+ * @return the deferred
+ */
+fun <T> ApolloCall<T>.toDeferred(): Deferred<Response<T>> {
+    val deferred = CompletableDeferred<Response<T>>()
+
+    deferred.invokeOnCompletion {
+        cancel()
```
digitalbuddha: needs to be wrapped in a `if (deferred.isCancelled) {`

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -0,0 +1,174 @@
...
+    }
+    enqueue(object: ApolloCall.Callback<T>() {
+        override fun onResponse(response: Response<T>) {
+            deferred.complete(response)
```
digitalbuddha:
```
          if (response.isSuccessful) {
            deferred.complete(response.body()!!)
          } else {
            deferred.completeExceptionally(HttpException(response))
          }
```

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
digitalbuddha: I think the way you have it written it will always cancel on completion rather than when it is cancelled

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
martinbonnin: Do we want this ? The RxJava2 wrappers emit `Responses` and not `data`

See https://github.com/apollographql/apollo-android/blob/37b9389f2736e132612720017c2ecfd639b0f0e2/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java#L83

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
martinbonnin: Correct. Although that should not change much, right? Calling `appolloCall.cancel()` after the call is completed shouldn't do much ?

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
nukeforum: The upside of returning just the data and throwing otherwise is that there is no `isSuccessful` song and dance at the call site. The inconsistency isn't great, but I think that for Rx, it's good to have `Responses` emitted to `onNext` since there's other "in process" stuff that can go wrong with an Rx mechanism, and those should be caught by `onError`.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
sav007: Yeah it's fine to call cancel, in case it already canceled / completed cancel will do nothing

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
sav007: We shouldn't emit only data, as `Response` has other useful fields:
```
public final class Response<T> {
  private final Operation operation;
  private final T data;
  private final List<Error> errors;
  private Set<String> dependentKeys;
  private final boolean fromCache;
```

So complete with response is fine

PR #1170, opened 2019-01-04
Apollo-LiveData-Support

skydoves:
Apollo-LiveData module
- LiveDataResponse
The LiveDataResponse class maps `Response` to `LiveDataResponse`.
There are Success, Failure, UnsuccessfulResponse, Complete sealed type classes.
RxJava can handle the state of response about onNext, onError, onComplete.
But live data cannot handle and emit values about the many kinds of response.
So I think `LiveData` need some wrapper class about the response.

- LiveDataApollo
 The `LiveDataApollo` class provides methods for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to LiveData sources.

- AplloExtension
 The `ApolloExtension` class provides extensions for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to LiveData sources. It based on `LiveDataApollo` class.

`LiveDataApollo` & `AplloExtension` use case.
```kotlin
// LiveDataApollo
LiveDataApollo.
        from(apolloClient.query(EpisodeHeroNameQuery(Input.fromNullable(Episode.EMPIRE))))
        .observe(this, Observer {
        // do something
    })

// AplloExtension
apolloClient.query(EpisodeHeroNameQuery(Input.fromNullable(Episode.EMPIRE))).asLiveData()
       .observe(this, Observer {
      // do something
       })
```

`AplloExtension` and `LiveDataResponse` use case.
```kotlin
val liveData = apolloClient.mutate(FacebookSignInMutation.builder()
            .facebookToken(facebookToken).build()).asLiveData()
        
liveData.value.let {
     when(it) {
         is ApiResponse.Success -> signInLiveData.postValue(it.data)
         is ApiResponse.Failure -> messageLiveData.postValue(it.errorMessage)
      }   
 }
```


apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * API Success response class.
+     *
+     * [data] is an Optional. (There are responses without data)
+     */
+    class Success<out T>(response: Response<T>) : LiveDataResponse<T>() {
+        val operation = response.operation()
+        val data: T? = response.data()
+        val dependentKeys: Set<String> = response.dependentKeys()
+        val fromCache: Boolean = response.fromCache()
```
sav007: not sure if these `destructuring` are required. Is there any reason to not have just `val response:Response<T>`?

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * API Failure response class.
+     *
+     * ## Throw Exception case.
+     * Communication is not successful or an exception related to the model deserialization occurs.
+     *
+     * ## API Network format error case.
+     * API communication conventions do not match or applications need to handle exceptions or errors.
+     */
+    class Failure<out T>(private val error: Exception, response: Response<out T>? = null) : LiveDataResponse<T>() {
```
sav007: How about to define 2 types of errors: one when smth related to request execution (basically in case of any ApolloException) and errors returned by GraphQL server (response has some errors)

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,227 @@
...
+import com.apollographql.apollo.cache.normalized.ApolloStoreOperation
+import com.apollographql.apollo.exception.ApolloException
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * The LiveDataApollo class provides methods for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to LiveData
+ * sources.
+ */
+@Suppress("unused")
+object LiveDataApollo {
```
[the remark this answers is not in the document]
skydoves: In Kotlin, The default visibility, used if there is no explicit modifier, is public.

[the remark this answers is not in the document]
skydoves: In Kotlin, The default visibility, used if there is no explicit modifier, is public.
Do you mean the `object` should be `internal`?

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.apollographql.apollo.api.Error
+import com.apollographql.apollo.api.Response
+
+/**
+ * The LiveDataResponse class maps [Response] to [LiveDataResponse].
+ * There are Success, Failure, UnsuccessfulResponse, Complete sealed type classes.
+ */
+
+@Suppress("unused")
+sealed class LiveDataResponse<out T> {
```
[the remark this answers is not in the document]
skydoves: :white_check_mark: Ah, I think it is better too.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
skydoves: :white_check_mark: Ah, I did not think about `destructuring`. It will be fixed to `var`.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,227 @@
...
+            var started = AtomicBoolean(false)
+            override fun onActive() {
+                super.onActive()
+                if (started.compareAndSet(false, true)) {
+                    call.enqueue(object : ApolloCall.Callback<T>() {
+                        override fun onResponse(response: Response<T>) {
+                            postValue(LiveDataResponse.of { response })
+                        }
+
+                        override fun onFailure(e: ApolloException) {
```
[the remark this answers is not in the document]
skydoves: :white_check_mark: Thank you :) It will be added about handling `StatusEvent`!

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
skydoves: As you can see, the `companion object` creates two type of Failures about `ApolloException` and 
`response.errors`.  The `of` method will create Failure class If `response.hasErrors()` is true.
But the Failure class is handling `Exception` now. I think it is too ambiguous. So It will fixed to

skydoves: As you can see, the `companion object` creates two types of Failures about `ApolloException` and 
`response.errors`.  The `of` method will create Failure class If `response.hasErrors()` is true.
But the Failure class is handling just `Exception` now. I think it is too ambiguous. So It will be fixed to `ApolloException` and `toString` override.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloExtension.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * The ApolloExtension class provides extensions for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to LiveData
+ * sources.
+ */
+
+/**
+ * Converts an [ApolloCall] to a LiveData.
+ */
+fun <T> ApolloCall<T>.asLiveData(): LiveData<ApolloLiveDataResponse<T>> {
```
sav007: One more note, to be aligned with Coroutine ext module lets use `to` prefix (`toLiveData`) instead of `as`.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
sav007: But there is an issue with that. We have 2 fields that indicates about the error. I believe that these 2 errors are mutual exclusive (I hope they are as in case of exception we can't really have response, on the other hand in case of GraphQL errors we can't have exception). 

That's the reason why we have in `com.apollographql.apollo.ApolloCall.Callback` 2 calls:
```
    /**
     * Gets called when GraphQL response is received and parsed successfully. Depending on the
     * {@link ResponseFetcher} used with the call, this may be called multiple times. {@link #onCompleted()}
     * will be called after the final call to onResponse.
     *
     * @param response the GraphQL response
     */
    public abstract void onResponse(@NotNull Response<T> response);

    /**
     * Gets called when an unexpected exception occurs while creating the request or processing the response.
     * Will be called at most one time. It is considered a terminal event. After called,
     * neither {@link #onResponse(Response)} or {@link #onCompleted()} will be called again.
     */
    public abstract void onFailure(@NotNull ApolloException e);
```
In case of GraphQL server error `onResponse` will be called.

But they are mutual exclusive. So my suggestion is to make `Failure` to be sealed class with 2 subclasses: `ApolloException` and `GraphQLError` (not sure that names are good enough, might be better).

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * API Success response class.
+     *
+     * [data] is an Optional. (There are responses without data)
+     */
+    class Success<T>(var response: Response<T>) : ApolloLiveDataResponse<T>() {
+        var operation = response.operation()
+        var data: T? = response.data()
+        var dependentKeys: Set<String> = response.dependentKeys()
+        var fromCache: Boolean = response.fromCache()
```
sav007: Sorry I might have provided misleading comment before.
I meant to remove all these fields and just leave `val response`.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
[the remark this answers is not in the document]
sav007: Yeah make `internal object LiveDataApollo`. I don't see the reason why to expose it if it used internally only

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,240 @@
...
+     * @throws NullPointerException if originalCall == null
+     */
+    fun <T> from(@NonNull call: ApolloCall<T>): LiveData<ApolloLiveDataResponse<T>> {
+        checkNotNull(call, "call == null")
+        return object : LiveData<ApolloLiveDataResponse<T>>() {
+            var started = AtomicBoolean(false)
+
+            @MainThread
+            override fun setValue(value: ApolloLiveDataResponse<T>?) {
+                started.set(true)
```
sav007: do we need this `started.set(true)`?

I think setting it in `onActive` would be enough?

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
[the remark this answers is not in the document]
skydoves: :white_check_mark: Ah, I think it is better too. Because In Java also can use` ApolloExtension` like this. 
```java
ApolloExtensionKt.toLiveData(call)
```
It will be fixed!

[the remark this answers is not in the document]
skydoves: :white_check_mark: Ah, I think it is better too. Because In Java also can use` ApolloExtension` like this. 
```java
LiveDataApolloExtensionKt.toLiveData(call);
```
So I will fix the extension file package : `ApolloExtension.kt` to `LiveDataApolloExtension.kt`.
And object LiveDataApollo will be internal.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloExtension.kt
skydoves: :white_check_mark: Thank you! I will be fixed.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+     * ## API Network format error case.
+     * API communication conventions do not match or applications need to handle exceptions or errors.
+     */
+    class Failure<out T>(val exception: ApolloException, response: Response<out T>? = null) : ApolloLiveDataResponse<T>() {
+        val errorMessage: String? = exception.localizedMessage
+        val errors: List<Error>? = response?.errors()
+
+        override fun toString(): String {
+            if (exception.localizedMessage == "UnsuccessfulResponse") {
+                return "[ApiResponse.Failure: ${errors.toString()}"
```
digitalbuddha: Either extraneous or missing `[`

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+         *
+         * [Failure] factory function. Only receives [ApolloException] arguments.
+         */
+        fun <T> error(ex: ApolloException) = Failure<T>(ex)
+
+        /**
+         * ApiResponse Factory
+         *
+         * [f] Create ApiResponse from [Response] returning from the block.
+         * If [Response] has not errors, It will create [ApolloLiveDataResponse.Success]
```
digitalbuddha: has no errors

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+         * ApiResponse Factory
+         *
+         * [Failure] factory function. Only receives [ApolloException] arguments.
+         */
+        fun <T> error(ex: ApolloException) = Failure<T>(ex)
+
+        /**
+         * ApiResponse Factory
+         *
+         * [f] Create ApiResponse from [Response] returning from the block.
```
digitalbuddha: What does `[f]` represent?

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
+package com.apollographql.apollo.livedata
+
+import com.apollographql.apollo.api.Error
+import com.apollographql.apollo.api.Response
+import com.apollographql.apollo.exception.ApolloException
+
+/**
+ * The ApolloLiveDataResponse class maps [Response] to [ApolloLiveDataResponse].
+ * There are Success, Failure, UnsuccessfulResponse, Complete sealed type classes.
```
digitalbuddha: '& Complete'

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+@Suppress("unused", "HasPlatformType")
+sealed class ApolloLiveDataResponse<out T> {
+
+    /**
+     * API Success response class.
+     *
+     * [data] is an Optional. (There are responses without data)
```
digitalbuddha: is optional

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
```diff
@@ -0,0 +1,80 @@
...
+         * [Failure] factory function. Only receives [ApolloException] arguments.
+         */
+        fun <T> error(ex: ApolloException) = Failure<T>(ex)
+
+        /**
+         * ApiResponse Factory
+         *
+         * [f] Create ApiResponse from [Response] returning from the block.
+         * If [Response] has not errors, It will create [ApolloLiveDataResponse.Success]
+         * If [Response] has errors, It will create [ApolloLiveDataResponse.Failure] using [UnsuccessfulResponse]
```
digitalbuddha: it

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed. Thank you so much!

digitalbuddha:
I'd like to get someone with more LiveData experience to review this tomorrow. Otherwise looks good!

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: `[f]` represents function, returns `Response<T>`.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/ApolloLiveDataResponse.kt
skydoves: :white_check_mark: It fixed. Thank you!

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,240 @@
...
+     *
+     * @param call the ApolloCall to convert.
+     * @param T    the value type.
+     * @return the converted LiveData.
+     * @throws NullPointerException if originalCall == null
+     */
+    fun <T> from(@NonNull call: ApolloCall<T>): LiveData<ApolloLiveDataResponse<T>> {
+        checkNotNull(call, "call == null")
+        return object : LiveData<ApolloLiveDataResponse<T>>() {
+            var started = AtomicBoolean(false)
```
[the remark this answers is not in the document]
skydoves: The `onActive()` method is called when the LiveData object has an active observer.

LiveData.java
```java
void activeStateChanged(boolean newActive) {
            if (newActive == mActive) {
                return;
            }
            // immediately set active state, so we'd never dispatch anything to inactive
            // owner
            mActive = newActive;
            boolean wasInactive = LiveData.this.mActiveCount == 0;
            LiveData.this.mActiveCount += mActive ? 1 : -1;
            if (wasInactive && mActive) {
                onActive();
            }
            if (LiveData.this.mActiveCount == 0 && !mActive) {
                onInactive();
            }
            if (mActive) {
                dispatchingValue(this);
            }
        }
```

So I think it needs to a flag for prevent to call `ApolloCall` whenever observers are changed.
This [SingleLiveEvent](https://github.com/googlesamples/android-architecture/blob/dev-todo-mvvm-live/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/SingleLiveEvent.java) is an example of a single emitter.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataResponse.kt
skydoves: I agree to it.  It seems to need specifying the Failures(exception case, error case).
It will be fixed. Thank you so much!

apollo-integration/src/test/java/com/apollographql/apollo/ApolloLiveDataTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun callProducesValue() {
+        mockWebServer.enqueue(mockResponse(FILE_EPISODE_HERO_NAME_WITH_ID))
+
+        val observer = mock<Observer<ApolloLiveDataResponse<EpisodeHeroNameQuery.Data>>>()
+        val liveData = apolloClient.query(EpisodeHeroNameQuery(Input.fromNullable(Episode.EMPIRE))).toLiveData()
+        liveData.observeForever(observer)
```
JoseAlcerreca: Have you considered LiveDataTestUtil?
https://github.com/googlesamples/android-architecture-components/blob/master/GithubBrowserSample/app/src/test-common/java/com/android/example/github/util/LiveDataTestUtil.kt

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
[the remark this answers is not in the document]
yigit: it is always called on the main thread. and if you happen to set any value, observers will still not receive it so you don't need to avoid setting value when that flag goes off.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,241 @@
...
+     * Converts an [ApolloPrefetch] to a LiveData.
+     *
+     * @param prefetch the ApolloPrefetch to convert.
+     * @return the converted LiveData.
+     * @throws NullPointerException if prefetch == null
+     */
+    fun <T> from(@NonNull prefetch: ApolloPrefetch): LiveData<ApolloLiveDataResponse<T>> {
+        checkNotNull(prefetch, "prefetch == null")
+        return object : LiveData<ApolloLiveDataResponse<T>>() {
+            var started = AtomicBoolean(false)
```
yigit: i think i'm confused w/ the started flag (sorry idk about Apollo).
Is there a chance that setValue mgiht be called outside this class (notice, it is protected so someone holding an instance of LiveData cannot call it).

Seems like you only need a normal boolean that gets checked in onActive. You would not need an atomic because onActive is called only on the main thread.

apollo-integration/src/test/java/com/apollographql/apollo/ApolloLiveDataTest.kt
skydoves: Hi, I added `LiveDataTestUtil` on Apollo-integration module, 
and it used in `ApolloLiveDataTest` class.

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
skydoves: Hi, appreciate your detail review!
The started `AtomicBoolean` flag is removed.

apollo-integration/src/test/java/com/apollographql/apollo/ApolloLiveDataTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    private lateinit var apolloClient: ApolloClient
+    private lateinit var mockWebServer: MockWebServer
+
+    private val FILE_EPISODE_HERO_NAME_WITH_ID = "EpisodeHeroNameResponseWithId.json"
+    private val FILE_EPISODE_HERO_NAME_CHANGE = "EpisodeHeroNameResponseNameChange.json"
+    private val FILE_HERO_AND_FRIEND_NAMD_CHANGE = "HeroAndFriendsNameWithIdsNameChange.json"
+
+    @Rule
+    @JvmField
+    val instantExecutorRule = InstantTaskExecutorRule()
```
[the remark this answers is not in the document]
skydoves: Hi, appreciate your detail review.
`InstantTaskExecutorRule` is replaced by `TaskExecutor`.

Yeboster:
Nice to hear this implementation, I'm enjoying using LiveData with Apollo, would be awesome to have an implementation like this. Nice work!

apollo-livedata-support/src/main/java/com/apollographql/apollo/livedata/LiveDataApollo.kt
```diff
@@ -0,0 +1,191 @@
...
+   * Converts an [ApolloStoreOperation] to a LiveData.
+   *
+   * @param operation the ApolloStoreOperation to convert.
+   * @param T         the value type.
+   * @return the converted LiveData.
+   */
+  fun <T> from(@NonNull operation: ApolloStoreOperation<T>): LiveData<T?> {
+    checkNotNull(operation, "operation == null")
+    return object : LiveData<T?>() {
+      var onPosted = false
```
sav007: do we need this flag?

sav007:
One more question, I see that we `enqueue` call but never  cancel it. If we have no more observables should we cancel the active call? So the flow will be smth like this:
```
onActive() {
  call.clone().enqueue()
}

onInactive() {
  call.cancel()
}
```

sav007:
Please rebase agains latest master

skydoves:
@sav007 
Please review the latest commits

digitalbuddha:
One more rebase and I feel this is good to go. Great job!!

tasomaniac:
Closing this since it is superseeded by #1320

PR #1202, opened 2019-02-05
Fix name conflict with Object

lawrencebautista:
I didn't test this but I think this will fix this issue.


digitalbuddha:
you should test it

sav007:
Out of date.

PR #1211, opened 2019-02-09
Preserve null list items when parsing the response

### LINKED ISSUE #1209: Null list items are dropped when parsing the response
If the GraphQL response contains an array with some `null` items, those items are dropped when mapping the response to a Java list. That's unexpected and in our case undesired.

`[1, null, 3]` is obviously not the same as `[1, 3]`: in the first case there are 3 items, in the second case only 2.

There's a unit test already proving this behaviour so it must be "by design": [ResponseReaderTest.readScalarListWithNulls()](https://github.com/apollographql/apollo-android/blob/1.0.0-alpha5/apollo-runtime/src/test/java/com/apollographql/apollo/internal/reader/ResponseReaderTest.java#L725), but why? And is there any way to change this behaviour?

I'm not aware of other JSON parsers behaving in this way.

mirkonasato:
Fixes #1209

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/NormalizedCacheTestCase.java
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/NormalizedCacheTestCase.java
@@ -154,11 +154,32 @@ public class NormalizedCacheTestCase {
           @Override
           public boolean test(Response<HeroAppearsInQuery.Data> response) throws Exception {
             assertThat(response.hasErrors()).isFalse();
-            assertThat(response.hasErrors()).isFalse();
             assertThat(response.data().hero().appearsIn()).hasSize(3);
             assertThat(response.data().hero().appearsIn().get(0).name()).isEqualTo("NEWHOPE");
             assertThat(response.data().hero().appearsIn().get(1).name()).isEqualTo("EMPIRE");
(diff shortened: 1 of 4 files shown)


Hemantkumar-Chigadani:
Any update on this PR, waiting for more inputs?

digitalbuddha:
Need to double check with @sav007 why the null check was added

sav007:
Sorry for the delay.
Changes look legit, but could you please write some tests for Normalized cache as well to be sure no unexpected behaviour over there?

mirkonasato:
@sav007 good thinking: I can see an `IllegalStateException: Stack is empty` in `ResponseNormalizer.didResolveList()`. Need to take a bit more time and understand who's supposed to call this `ResolveDelegate` and when.

mirkonasato:
Ok, added a test to `NormalizedCacheTestCase`, that was initially failing and is now passing after calling `resolveDelegate.didResolveNull()` in `RealResponseReader`.

To be honest there isn't much documentation or tests around `ResolveDelegate` so this is just my best guess of when it should be called.

@sav007 kindly have a look when you have a chance.

Hemantkumar-Chigadani:
@sav007 and @digitalbuddha Could you be able to merge this to master? Thanks

PR #1214, opened 2019-02-20
Fix the ConcurrentModificationException in RealSubscriptionManager#stop

baconz:


sav007:
Hm, I'm curious how this can be reproduced? Just looking into the code all accessors are protected by lock it shouldn't have any concurrent modification. Could you please provides step to reproduce?

baconz:
The issue was that we were calling  `doUnsubscribe()` inside of
the for-each loop in `stop()`. Since `doUnsubscribe()` calls `remove()` on
the subscription hash, we were modifying the hash, without letting the
implicit iterator from the for-each loop know about the cahnge. This
results in a `ConcurrentModificationException`.

I've modified one of the test cases to reproduce the issue. If you revert
my change to `stop()` and run the test, you'll see the CME.

sav007:
Agh right, thx for fixing it.

Issue #1215, opened 2019-02-22
parsing fails for null data field in com.apollographql.apollo.response.OperationResponseParser 


gshcherb:
According to https://facebook.github.io/graphql/draft/#sec-Data

null data value is a valid value and can be present in the response. Current implementation of `OperationResponseParser` just takes null value and treats it as valid map with data.

This later causes a crash in 

`com.apollographql.apollo.response.OperationResponseParser.parse` when null is queried for data.

sav007:
PR was merged.

PR #1216, opened 2019-02-22
Added null check for response 'data' field before parsing payload

gshcherb:
https://github.com/apollographql/apollo-android/issues/1215

According to https://facebook.github.io/graphql/draft/#sec-Data
null data value is a valid value and can be present in the response.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java
@@ -49,8 +49,8 @@ public final class OperationResponseParser<D extends Operation.Data, W> {
     checkNotNull(payload, "payload == null");
 
     D data = null;
-    if (payload.containsKey("data")) {
-      Map<String, Object> buffer = (Map<String, Object>) payload.get("data");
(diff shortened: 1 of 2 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java
```diff
@@ -49,8 +49,8 @@ public OperationResponseParser(Operation<D, W, ?> operation, ResponseFieldMapper
     checkNotNull(payload, "payload == null");
 
     D data = null;
-    if (payload.containsKey("data")) {
-      Map<String, Object> buffer = (Map<String, Object>) payload.get("data");
+    Map<String, Object> buffer = (Map<String, Object>) payload.get("data");
+    if (buffer!=null) {
```
sav007: Am I missing smth but the changes are identical to what was before?

sav007: Sorry my bad you are totally right.
Could you please add some test as well?

gshcherb: Added simple passing unit test.

Issue #1183, opened 2019-01-17
Add http header with request scoped state


westd:
Hi

I'm using apollo-android from a multithreaded jvm server application (not android).  I need to add a http header to my graphql request that contains some state specific to the request (let's say something like a user id)

I can't just change the user id property of an OKHttp interceptor to add the header because another thread might get there first and use the wrong user id.

I tried to work around using a thread local variable but it doesn't work because the OK http request runs on a different thread to my original request thread.

Any suggestions on how I might do this?
Thanks

sav007:
Well it's quite unusual feature, just one thought can you use multiple OkHttpClients for different users, but share the same connection pool, dispatchers? In this case each OkHttpClient will have own interceptors but use the same resources.

westd:
Hi Ivan thanks for your suggestion I’ll look into that.

I wouldn’t consider it unusual though. It’s just a spring webmvc server making GraphQL requests. When a user makes a request for particular webpage it’s typical there will be some request state to send to the GraphQL server: the auth token if nothing else. 

A future solution would be to a headers parameter when making the request?

TarasMel:
Hi have little problem, i need get some items from Response Header, can you tell me how i can get it...
Our API gives response.body() with some data, and JWT are in headers, but i can't get it.
Any suggestions on how I might do this?
Thanks

BenSchwab:
Hi @sav007 👋 Airbnb actually uses per request headers frequently for logging purposes. I feel like need to clone OkHttp for every request that needs custom headers (a lot for us) is a bit clumsy. How do you feel about adding this as an option?

digitalbuddha:
Hi Ben! Ivan is on vacation this week but I don't see any reason why it would be a problem to add as a feature. Would someone from your end be able to contribute it?

BenSchwab:
Yeah, happy to take a crack at it and dust of my apollo knowledge :)

BenSchwab:
Just an update -- we are using the cloned http workaround, which does work okay, so this is low-pri for me. Should get to it in the next few weeks, but if someone else would like to take a crack at it sooner, please do!

tmtrademarked:
We've started implementing some GraphQL stuff here, and for various reasons, we have use cases which require per-request headers. I decided to take a stab at implementing this in the referenced PR. Take a look and let me know if you think this is a reasonable road to go down?

sav007:
I think we can close this one.

PR #1234, opened 2019-03-26
Adding support for per-request headers

tmtrademarked:
This PR adds support for the concept of per-request headers. These are introduced by adding a RequestHeaders structure which is similar to the existing CacheHeaders structure. These are configured on ApolloQueryCalls and provided to a new interceptor which adds the specified headers to the call when executed.

This PR also fixes an issue with the integration tests - SHA hashes can be different on different file systems based on encoding. On my local machine, at least, these hashes were different than the hard-coded values. So replacing the integration test SHAs with constants worked better. (Note - the tests don't seem to fully pass for me with `./gradlew check`, with some lint errors in files I didn't touch)

This should help to address https://github.com/apollographql/apollo-android/issues/1183.


sav007:
Looks great, question should `ApolloMutationCall` has the same API?

tmtrademarked:
@sav007 - great callout. I can't see any good reason why this shouldn't work for mutations too, so I've added that API.

PR #1295, opened 2019-05-13
Release 1.0.0

sav007:


upload_archives.sh
```diff
@@ -0,0 +1,10 @@
+./gradlew clean :apollo-android-support:bintrayUpload
```
digitalbuddha: Why run clean between each one?

sav007: No reason just to be sure. :)

Issue #1296, opened 2019-05-14
Incorrect generation of Kotlin enum classes with long name


alexxxdev:
example
```Kotlin
// AUTO-GENERATED FILE. DO NOT MODIFY.
//
// This class was automatically generated by Apollo GraphQL plugin from the GraphQL queries it found.
// It should not be modified by hand.
//
package ...

import kotlin.String

enum class DeliveryAddressStatusEnum(val rawValue: String) {

...
    companion object {
        fun safeValueOf(rawValue: String): DeliveryAddressStatusEnum = values().find { it.rawValue
                == rawValue } ?: UNKNOWN__
    }
}
```
:kaptGenerateStubsDebugKotlin FAILED: Expecting an element in line ```== rawValue } ?: UNKNOWN__```

For short class names, everything is generated correctly:
```Kotlin
companion object {
        fun safeValueOf(rawValue: String): TypeEnum = values().find { it.rawValue == rawValue
                } ?: UNKNOWN__
    }
```

digitalbuddha:
Bad line wrapping thanks for reporting

sav007:
Oh it seems the issue with the Kotlin poet, will check if it's true will open an issue.

sav007:
Issue https://github.com/square/kotlinpoet/issues/695 has been created. Will try to fix it on our end.

jayresta:
Hi- even on current master I am still seeing a variant of this issue for a generated Kotlin enum type which is exactly 45 characters long. The generated line wraps the { bracket after the values().find which causes a compile error.

  companion object {
    fun safeValueOf(rawValue: String): MyGeneratedTypeWhichIs45CharactersLong = values().find
        { it.rawValue == rawValue } ?: UNKNOWN__
  }

PR #1304, opened 2019-05-16
Fix bad line wrapping for enum safeValueOf Kotlin model generation

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1296


digitalbuddha:
Curious why didn't a test fail for this if it was a compilation error?

sav007:
The issue that we don't have such test, our schema defines `Episode` enum that has short name. Kotlin Poet always wrap lines by spaces https://github.com/square/kotlinpoet#spaces-wrap-by-default

Issue #1224, opened 2019-03-11
Coroutines, "Failed to resolve"


Dragons0458:
Hello, I was trying to use the coroutines but when trying to use the implementation I get the following error: "ERROR: Failed to resolve: com.apollographql.apollo: apollo-coroutine-support: 1.0.0-alpha5" Any suggestions? It's the only one that does not work, thank you!

Dragons0458:
ok, add this:

`maven {
        url  "https://dl.bintray.com/apollographql/android" 
    }`

Update README please, thank you! :D

AdamMc331:
Saw this ticket was closed because you figured it out, but I'm adding a readme inside the coroutines module for this: https://github.com/apollographql/apollo-android/pull/1307

PR #1307, opened 2019-05-16
Added documentation for coroutines module for #1224.

AdamMc331:
I had trouble adding the coroutines module just like [this person](https://github.com/apollographql/apollo-android/issues/1224) did and that ticket appears to have been closed before anyone noticed so I added a the docs inside the coroutine module folder.


digitalbuddha:
Thank you for contribution

PR #1319, opened 2019-05-22
fix use get

nightbear1009:


digitalbuddha:
I'm confused what this change is doing

jbttn:
I'm not the author of this change, but looking at it closer it seems like the `prepareInterceptorChain` function is using the value of `useHttpGetMethodForQueries`.  Specifically this line:

```
interceptors.add(new ApolloServerInterceptor(serverUrl, httpCallFactory, httpCachePolicy, false, scalarTypeAdapters,
        logger, useHttpGetMethodForQueries));
```

It looks like the value is being read before it has been set, which this diff seems to correct.

nightbear1009:
> I'm confused what this change is doing
just like @jbttn says
the useHttpGetMethodForQueries is being read before it has been set

Issue #1278, opened 2019-04-28
Add code style file to project and/or linting rules


AdamMc331:
I noticed in the contribution guide that this should be standard Java formatting, specifically four space indents: https://github.com/apollographql/apollo-android/blob/master/Contributing.md

but just about everywhere in the project things are 2. Is this something we could change?

If there's an agreed upon code style we can add an XML file to the project so that those who clone this repository will adhere to the same standard. We could also implement something like [ktlint](https://github.com/pinterest/ktlint).

digitalbuddha:
Sure! Feel free to clean up the formating to standard java style. The 2 space style was due to legacy code ported from another project

AdamMc331:
I don't think this is something we can make one PR for. For example, if I were to run Android Studio's code format tool on the root package, it would change 645 files. I would suspect we _don't_ want to do that, even if it's code formatting only. 

Here's my idea at a potential solution:

Start by creating a code styles XML file with some of the standard Android formatting rules, and submit a PR with just that. Get some feedback/collaboration on that, and make sure it's something we agree on.

Then, we could do one of these approaches (or a mix):

- Leave it be, and have people who edit new files remember to auto format them (if we add the formatting file to git then it will happen automatically)
- Start taking an iterative approach, and either run a formatting process on a module by module (or even smaller, package by package) basis until we've worked through the whole thing. 

Personally, I think a mix is good. I'd also have to do more research but I think there is a way we could add lint checks to the CI process, and set our current error count as a threshold to make sure the project doesn't go below that, and then iteratively over time we can modify it. 

What are your thoughts? Also maybe this is pointless maybe 2 space indentation is fine. If that's the case, we could still do a PR with formatting or add checks, but then we'd also want to update that contribution guide.

digitalbuddha:
I'd prefer to keep it as 2 space indentation, otherwise feel free to add lint/checks and fix other violations.  I'll update the contributing guide once something is added.

PR #1326, opened 2019-05-25
Added code styles xml file for #1278.

AdamMc331:
So the inspiration behind this PR is that when I first forked Apollo, AS on my computer wanted to automatically indent everything to 4 spaces. I realized that wasn't the standard for the project, and updated it locally, but wanted to help save the hassle for others.

If we commit this code styles XML file to the repo, then when others fork it, or run auto format on their (updated) branches, they won't have to worry about that trap. 

I chose not to reformat any files at the same time just to provide emphasis on the styles that are being merged in, and making sure those are approved. Then fixing any malformed files can be done as a separate effort or simultaneous with other PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .gradle
-/.idea
+.idea/*.xml
+.idea/caches/
+.idea/libraries/
 *.iml
 build
 local.properties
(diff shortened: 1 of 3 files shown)


AdamMc331:
I just noticed this was never merged when I was combing thru open PRs. Do we still find this useful? We can close if not.

martinbonnin:
I got just bitten by that (see https://github.com/apollographql/apollo-android/pull/1577#discussion_r327918973). I think we should merge it.

Issue #1322, opened 2019-05-24
Cloudflare 400 Bad Request + extensions.persistedQuery in request


vikrama:
As per #1161 , I tried using the latest SNAPSHOT but I'm still seeing the `extensions. persistedQuery` getting added to the query and resulting in a 400 Bad Request. I noticed that it wasn't our server that throws fits, but our DNS provider Cloudflare. Just by removing the `extensions` object from the json manually (curl request), Cloudflare lets it through.

Ref: https://www.apollographql.com/docs/apollo-server/features/apq#cdn

Any idea where the problem is?

_Update:_
Seeing problems with:
apollo-plugin, apollo-runtime, apollo-rx2: 1.0.0-alpha5, 1.0.0, 1.0.1-SNAPSHOT-*

For some reason, things work fine with the following combo and I don't see any `extensions.persistedQuery` getting added by default.
apollo-plugin: 1.0.0-alpha4
apollo-runtime, apollo-rx2: 1.0.0-alpha3

knordbo:
Seeing the same issue. Should `ApolloServerInterceptor` do a check to see if `persistedQuery` is enabled before sending the payload?

sav007:
Sorry confused what the issue? Did you try latest release?

knordbo:
The problem is that despite disabling `persistedQuery`, `extensions` is still added to the request. https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java#L198 
The current workaround which is a bit nasty is to make an interceptor that removes it.

Many GraphQL servers do not accept `extensions` in the payload.

PR #1340, opened 2019-06-01
Don't send extensions in query payload when auto persistence disabled

sav007:
Don't send `extensions` section with query sha if query auto persistence feature is disabled.

Closes https://github.com/apollographql/apollo-android/issues/1322


brockgibson:
Curious what the timeline might be to getting this change released. It's causing issues in our client-server interactions as well.

PR #1342, opened 2019-06-02
Propagate network error for CacheAndNetworkFetcher

### LINKED ISSUE #1207: CacheAndNetworkFetcher not propagating Network errors
Hello,

We've noticed that `CacheAndNetworkFetcher` by design will not return network errors if the cached data was successfully returned.  I would guess the intention was to silent fail when the cache data is available. 
However, it does not cover the use case where we want to show an error to the user in the UI.

Is there any strong opinions about this? 
I would suggest to change it from:
```
// Only send the network result after the cache result has been dispatched
      if (dispatchedCacheResult) {
        if (networkResponse.isPresent()) {
          originalCallback.onResponse(networkResponse.get());
          originalCallback.onCompleted();
        } else if (networkException.isPresent()) {
          if (cacheException.isPresent()) {
            // Both cache and network errored out
            originalCallback.onFailure(networkException.get());
          } else {
            // Cache was successful. Just terminate.
            originalCallback.onCompleted();
          }
        }
      }
```

to 

```
// Only send the network result after the cache result has been dispatched
      if (dispatchedCacheResult) {
        if (networkResponse.isPresent()) {
          originalCallback.onResponse(networkResponse.get());
          originalCallback.onCompleted();
        } else if (networkException.isPresent()) {
            originalCallback.onFailure(networkException.get());
        }
      }
```

With this approach it is up to the UI to ignore and fail silent if it desires.

Before I send out a PR wanted to check the reasoning behind this logic.

BTW, on iOS ApolloClient all cache errors are swallowed, and any network error is propagated.
Perhaps it is a good idea to keep a consistency and make it more flexible.

sav007:
Closes #1207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ android:
 
 script:
   - ./gradlew clean build connectedCheck -x checkstyleTest --stacktrace --max-workers=2 -x apollo-integration:build -x apollo-integration:connectedCheck
-# - .buildscript/integration_tests_composite.sh
+  - .buildscript/integration_tests_composite.sh
 
 before_script:
   - android list target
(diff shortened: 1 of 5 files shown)


sav007:
Yeah it was disabled when new module was introduced, now we need to enable it back

Issue #1317, opened 2019-05-21
useHttpGetMethodForQueries is not working


woodyLam:
I want to using HTTP GET method for Auto Persisted Queries, but all requests still using POST method even I set useHttpGetMethodForQueries = true

`ApolloClient.builder().useHttpGetMethodForQueries(true).enableAutoPersistedQueries(true)`

sav007:
This flag is used for queries only. Mutations are not supported. Are you sure that query operations still sent as POST?

woodyLam:
yes, i am using query operations and all query sent as POST

wooyukit:
What do you mean about mutations query?

nightbear1009:
I find out in RealApolloCall.java  at line 119 - 121

interceptorChain = prepareInterceptorChain(operation);
optimisticUpdates = builder.optimisticUpdates;
useHttpGetMethodForQueries = builder.useHttpGetMethodForQueries;

should be like as below  **prepareInterceptorChain** is using **useHttpGetMethodForQueries**, but **useHttpGetMethodForQueries** is not assigning value until line 121

useHttpGetMethodForQueries = builder.useHttpGetMethodForQueries;
interceptorChain = prepareInterceptorChain(operation);
optimisticUpdates = builder.optimisticUpdates;

wooyukit:
I tried to move up `useHttpGetMethodForQueries = builder.useHttpGetMethodForQueries;` but it doesn't work as well.

Codebear98:
The above code seems have been fixed at 3eafed98afb936c3b37ecb42c3e2a5cb6312a217 in Master.
Nows
`    enableAutoPersistedQueries = builder.enableAutoPersistedQueries;
    useHttpGetMethodForPersistedQueries = builder.useHttpGetMethodForPersistedQueries;
    interceptorChain = prepareInterceptorChain(operation);
`

However, for issue we really wanna do is [integrate APQs with CDN](https://www.apollographql.com/docs/apollo-server/features/apq/#cdn) which requires HttpGetMethod.

We wanna implement similar field of  `useGETForHashedQueries` which sends with HttpGetMethod when it's a Persisted Query request, otherwise send HttpGetMethod  or HttpPostMethod depends on global config of `useHttpGetMethodForQueries`.

*PR* for this https://github.com/apollographql/apollo-android/pull/1346

sav007:
Fixed.

PR #1346, opened 2019-06-04
Support send HttpGetMethod for Persisted Queries

Codebear98:
- Support APQs with CDN through useHttpGetMethodForPersistedQueries

Ref. [Integrate APQs with CDN](https://www.apollographql.com/docs/apollo-server/features/apq/#cdn)

Implemented `useHttpGetMethodForPersistedQueries` which sends with HttpGetMethod when it's a persisted query request, otherwise send HttpGetMethod  or HttpPostMethod depends on global setting of `useHttpGetMethodForQueries`.

**Example**
```
ApolloClient.builder()
.enableAutoPersistedQueries(true)
.useHttpGetMethodForPersistedQueries(true)
.build()
```

ref. #1317 #1055


apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
```diff
@@ -88,7 +88,8 @@ public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull f
         callBack.onFetch(FetchSourceType.NETWORK);
 
         try {
-          if (useHttpGetMethodForQueries && request.operation instanceof Query) {
+          if ((request.useHttpGetMethodForQueries || useHttpGetMethodForQueries)
```
sav007: hm I guess you need to rebase as it should be only `request.useHttpGetMethodForQueries && request.operation instanceof Query`?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
```diff
@@ -68,13 +70,13 @@ public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull f
           if (isPersistedQueryNotFound(response.errors())) {
             logger.w("GraphQL server couldn't find Automatic Persisted Query for operation name: "
                 + request.operation.name().name() + " id: " + request.operation.operationId());
-            return Optional.of(request.toBuilder().sendQueryDocument(true).build());
+            return Optional.of(request.toBuilder().sendQueryDocument(true).useHttpGetMethodForQueries(false).build());
```
sav007: I think this should fallback to whatever user configured  in `com.apollographql.apollo.ApolloClient.Builder#useHttpGetMethodForQueries` rather than always `false`?

sav007:
Also pls there is check style error in `ApolloAutoPersistedQueryInterceptor`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
Codebear98: Thanks for reviewing the PR.

As 'useHttpGetMethodForQueries' is passing through from ApolloClient as a global setting of using Get Method.
and
request.useHttpGetMethodForQueries is passing from previous interceptor chains. eg. the APQs interceptor.

so either one is true should use GET Method. Is it acceptable?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
Codebear98: Yes, it will fallback in later Interceptor chains.

According to `com.apollographql.apollo.internal.RealApolloCall.java` line. 355
'prepareInterceptorChain', the `com.apollographql.apollo.ApolloClient.Builder#useHttpGetMethodForQueries` is passing to `ApolloServerInterceptor` instead of `ApolloAutoPersistedQueryIntercaptor`.

Under this consideration,
Should we also pass useHttpGetMethodForQueries to ApolloAutoPersistedQueryIntercaptor?

sav007:
Could you please rebase.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
sav007: Please rebase as `useHttpGetMethodForQueries ` is now passed inside the `InterceptorRequest`. So I assume we should fall back to the original `request.useHttpGetMethodForQueries`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
Codebear98: ok get it, code is outdated, going to fix it.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
Codebear98: Got it & updated, now it will respect to the old `request.useHttpGetMethodForQueries` for retrying as normal request after APQs is failed.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java
Codebear98: updated as above suggestion.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
Codebear98: May i clarify the original usage of 
`ApolloClient#useHttpGetMethodForQueries` 
Is it for all requests to use GETMethod regardless APQs or not? for all Queries right?

Just wanna make sure the useHttpGetMethodForPersistedQueries doesn't duplicate the function.

Codebear98:
> Could you please rebase.

hi, good to merge?

sav007:
You probably should have rebased instead of merging as right now it's almost impossible to review this PR `+4,414 −597 ` as it includes all others and its super hard to do a final review. Could you pls rebase your commits on latest master?

Codebear98:
> You probably should have rebased instead of merging as right now it's almost impossible to review this PR `+4,414 −597 ` as it includes all others and its super hard to do a final review. Could you pls rebase your commits on latest master?

I'm sorry, not sure how to fix it from this PR.
I created another PR #1376 with sqush commits, will do rebase instead of merge for later sync.
Thanks!

Issue #1352, opened 2019-06-06
compileJava fails with `no source files` if the project is 100% kotlin and generateKotlinModels = true


martinbonnin:
> Task :compileJava FAILED

    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':compileJava'.
    > no source files

This comes from there: https://github.com/apollographql/apollo-android/blob/b327bc79a94e5a1c7ebef5e42554dc86a12c1225/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy#L105

If the directory is empty, the java compiler does not complain but in that case it contains kotlin files that make javac not happy.

alexxxdev:
now it causes problems with `compile%name%Kotlin`
for any value of `generateKotlinModels`
project is 100% kotlin

sav007:
Just to make sure, you tried the latest snapshot version? Could you please provide a log with the stacktrace for you build?

sav007:
Just tried on my sample app (pure Kotlin) couldn't reproduce any errors

alexxxdev:
@sav007 in logs only https://gist.github.com/alexxxdev/3a2581b9da277868ef99015d535394c6
and errors in unresolved classes, which do not generate

alexxxdev:
also i did 
`git clone https://github.com/apollographql/apollo-android.git`
and
`./gradlew clean build connectedCheck -x checkstyleTest --stacktrace --max-workers=2 -x apollo-integration:build -x apollo-integration:connectedCheck`
I getting tighter error https://gist.github.com/alexxxdev/f12fee47cad255b4ab0c32bf24cc1e0e

martinbonnin:
@alexxxdev I can reproduce with `./gradlew :apollo-kotlin-sample:compileDebugKotlin` from apollo-android master, I'm looking into it.

martinbonnin:
This seems to be a side effect of `afterEvaluate` in the android case. It looks like `registerJavaGeneratingTask` is not picking up the codegen tasks and it works again if I remove the `afterEvaluate`. @sav007 feel free to revert https://github.com/apollographql/apollo-android/pull/1353 for now as it might take a bit of time to investigate.

martinbonnin:
I have been looking into that a bit today. Somehow, `compileDebugKotlin` does not depend on `generateDebugSources` if `afterEvaluate` is used. Not sure yet why.

sav007:
[...]

PR #1353, opened 2019-06-06
do not try to compile java files if we don't generate any

martinbonnin:
Should fix https://github.com/apollographql/apollo-android/issues/1352

### DIFF
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloPlugin.groovy
@@ -101,9 +101,11 @@ class ApolloPlugin implements Plugin<Project> {
     apolloIRGenTask.dependsOn(sourceSetIRTask)
     apolloClassGenTask.dependsOn(sourceSetClassTask)
 
-    JavaCompile compileTask = (JavaCompile) project.tasks.findByName("compile${taskName.capitalize()}Java")
-    compileTask.source += project.fileTree(sourceSetClassTask.outputDir)
-    compileTask.dependsOn(apolloClassGenTask)
+    if (project.apollo.generateKotlinModels == false) {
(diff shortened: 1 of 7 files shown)


martinbonnin:
Test added there: https://github.com/apollographql/apollo-android/blob/95b0e854eb5953d95acceabb474a8e193b9af3f3/apollo-gradle-plugin/src/test/testProject/kotlin/build.gradle

As it's using the snapshots, it will fail until new snapshosts are deployed. I'm not sure how this was handled before ?

sav007:
Do you know why integration test fails on CI? As it a composite test and should run with the latest changes to plugin without waiting for snapshot to be published.

martinbonnin:
Pretty sure I didn't do the composite thing. Right now it uses the snapshot plugin which is why it fails. I'll look into composite tests

martinbonnin:
It should be good. 2 points worth noting:

* I had to move the creation of the apollo tasks to `afterEvaluate`. This might or might not have side effects but it's the only way I found to know reliably the value of apollo.generateKotlinModels and configure the java task.
* The other tests simply add the apollo-api code to the java sourceSet which defeats the value of this test (testing that the project compiles without any java file). In order to still have the api dependency, I included the snapshot so the test is run with a local plugin and a snapshot api. This might or might not be an issue...

sav007:
I will take a look on my sample project with snapshot version if `afterEvaluate` caused any issues, it shouldn't but will check.

alexxxdev:
> I will take a look on my sample project with snapshot version if `afterEvaluate` caused any issues, it shouldn't but will check.

@sav007 @martijnwalraven it seems
now it causes problems with `compile%name%Kotlin`
for any value of `generateKotlinModels`
project is 100% kotlin

`/build/apollo-codegen/` empty & `/build/generated/source/apollo` doesn't exist

AllanWang:
^ For me the generated sources actually do exist, but they don't get resolved properly; gradle won't allow me to use it.

#1372

PR #1357, opened 2019-06-07
Fix usages of internal OkHttp APIs

### LINKED ISSUE #1347: Apollo uses internal OkHttp APIs
This is using non-public API and causes compatibility issues. Case and point - `DiskLruCache.create` no longer works in the soon-to-be-released OkHttp 4. The entire `DiskLruCache` is actually internal to OkHttp, I think it would be best if this were just copied directly in rather than used from OkHttp

ZacSweers:
Resolves #1347. Doesn't entirely remove everything (some internal usages still in tests), but makes the prod code not use them anymore. This is using sources from OkHttp 3.14.2


ZacSweers:
I wasn't entirely sure what you wanted to do with logging in DiskLruCache, so left those commented out for now for discussion

ZacSweers:
The CI failure seems unrelated, maybe a flaky test?

sav007:
As for the disk lru cache, should we use the Jake's version https://github.com/JakeWharton/DiskLruCache instead of copied version?

ZacSweers:
I think the okhttp version is more up to date, and also would be a better guarantee of no subtle behavior changes in this switch

ZacSweers:
Rebased and fixed the issues I think

PR #1376, opened 2019-06-18
Support send HttpGetMethod for Persisted Queries

Codebear98:
- Support APQs with CDN through useHttpGetMethodForPersistedQueries

Ref. [Integrate APQs with CDN](https://www.apollographql.com/docs/apollo-server/features/apq/#cdn)

Implemented `useHttpGetMethodForPersistedQueries` which sends with HttpGetMethod when it's a persisted query request, otherwise send HttpGetMethod  or HttpPostMethod depends on global setting of `useHttpGetMethodForQueries`.

**Example**
```
ApolloClient.builder()
.enableAutoPersistedQueries(true)
.useHttpGetMethodForPersistedQueries(true)
.build()
```

ref. #1317 #1055

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
@@ -20,7 +20,6 @@ public class ApolloAutoPersistedQueryInterceptor implements ApolloInterceptor {
 
   private final ApolloLogger logger;
   private volatile boolean disposed;
-  private boolean useHttpGetMethodForQueries;
 
   final boolean useHttpGetMethodForPersistedQueries;
 
@@ -34,12 +33,10 @@ public class ApolloAutoPersistedQueryInterceptor implements ApolloInterceptor {
   public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull final ApolloInterceptorChain chain,
       @NotNull final Executor dispatcher, @NotNull final CallBack callBack) {
 
-    useHttpGetMethodForQueries = request.useHttpGetMethodForQueries;
-
     InterceptorRequest newRequest = request.toBuilder()
             .sendQueryDocument(false)
             .autoPersistQueries(true)
-            .useHttpGetMethodForQueries(useHttpGetMethodForQueries || useHttpGetMethodForPersistedQueries)
+            .useHttpGetMethodForQueries(request.useHttpGetMethodForQueries || useHttpGetMethodForPersistedQueries)
             .build();
     chain.proceedAsync(newRequest, dispatcher, new CallBack() {
       @Override public void onResponse(@NotNull InterceptorResponse response) {
@@ -81,21 +78,13 @@ public class ApolloAutoPersistedQueryInterceptor implements ApolloInterceptor {
             logger.w("GraphQL server couldn't find Automatic Persisted Query for operation name: "
                 + request.operation.name().name() + " id: " + request.operation.operationId());
 
-            return Optional.of(request.toBuilder()
-                    .sendQueryDocument(true)
-                    .autoPersistQueries(true)
-                    .useHttpGetMethodForQueries(useHttpGetMethodForQueries)
-                    .build());
+            return Optional.of(request);
           }
 
           if (isPersistedQueryNotSupported(response.errors())) {
             // TODO how to disable Automatic Persisted Queries in future and how to notify user about this
             logger.e("GraphQL server doesn't support Automatic Persisted Queries");
-            return Optional.of(request.toBuilder()
-                    .sendQueryDocument(true)
-                    .autoPersistQueries(true)
-                    .useHttpGetMethodForQueries(useHttpGetMethodForQueries)
-                    .build());
+            return Optional.of(request);
           }
         }
         return Optional.absent();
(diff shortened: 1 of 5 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
```diff
@@ -20,15 +20,27 @@
 
   private final ApolloLogger logger;
   private volatile boolean disposed;
+  private boolean useHttpGetMethodForQueries;
```
sav007: I guess this flag can be local in `interceptAsync `?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
```diff
@@ -20,15 +20,27 @@
...
   public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull final ApolloInterceptorChain chain,
       @NotNull final Executor dispatcher, @NotNull final CallBack callBack) {
-    InterceptorRequest newRequest = request.toBuilder().sendQueryDocument(false).build();
+
+    useHttpGetMethodForQueries = request.useHttpGetMethodForQueries;
+
+    InterceptorRequest newRequest = request.toBuilder()
+            .sendQueryDocument(false)
+            .autoPersistQueries(true)
+            .useHttpGetMethodForQueries(useHttpGetMethodForQueries || useHttpGetMethodForPersistedQueries)
```
sav007: nit:  think can be replaced with:
```
.useHttpGetMethodForQueries(request.useHttpGetMethodForQueries || useHttpGetMethodForPersistedQueries)
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
```diff
@@ -68,13 +80,22 @@ public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull f
...
           }
 
           if (isPersistedQueryNotSupported(response.errors())) {
             // TODO how to disable Automatic Persisted Queries in future and how to notify user about this
             logger.e("GraphQL server doesn't support Automatic Persisted Queries");
-            return Optional.of(request.toBuilder().sendQueryDocument(true).build());
+            return Optional.of(request.toBuilder()
+                    .sendQueryDocument(true)
+                    .autoPersistQueries(true)
+                    .useHttpGetMethodForQueries(useHttpGetMethodForQueries)
```
sav007: I believe we don't need this as `request ` is our original request (not modified). So it will have already set this flag.

See line with:
```
chain.proceedAsync(newRequest, dispatcher, new CallBack() {
      @Override public void onResponse(@NotNull InterceptorResponse response) {
        if (disposed) return;
        Optional<InterceptorRequest> retryRequest = handleProtocolNegotiation(request, response);
```

`request` is original request not modified.

Codebear98: I see, we can access to original request.

However, we still need to send the httpbody with `extension.persistedQuery` for creating PQ cache in server for later request. So we need `.sendQueryDocument(true) and .autoPersistQueries(true)` and  can't guarantee original request those value on.

to combine the above two comment, probably modified as following? 

```
return Optional.of(request.toBuilder()
                    .sendQueryDocument(true)
                    .autoPersistQueries(true)
                    .useHttpGetMethodForQueries(request.useHttpGetMethodForQueries)

```

Codebear98: yes you r right, as original request is preserved, can reuse that one. Fixed as suggestion above.

Beware that we still need to send the http body with `extension.persistedQuery` for creating PQ cache in server for later request. So we need `.sendQueryDocument(true) and .autoPersistQueries(true)` and guarantee original request has those value on.

And it's value depends in RealApolloCall.java line. 139
```
    ApolloInterceptor.InterceptorRequest request = ApolloInterceptor.InterceptorRequest.builder(operation)
        .cacheHeaders(cacheHeaders)
        .requestHeaders(requestHeaders)
        .fetchFromCache(false)
        .optimisticUpdates(optimisticUpdates)
        .useHttpGetMethodForQueries(useHttpGetMethodForQueries)
        .autoPersistQueries(enableAutoPersistedQueries)
        .build();
    interceptorChain.proceedAsync(request, dispatcher, interceptorCallbackProxy());
  }
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
Codebear98: removed

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloAutoPersistedQueryInterceptor.java
Codebear98: replaced, thanks!

PR #1387, opened 2019-06-23
RxJava Support section fixed & improved

blipinsk:
"RxJava section" was majorly outdated and confusing.

This PR consists of:
1. Rework of the main part of the section.
2. Removed `ApolloWatcher` example:
   * _It was outdated. It was missing `Response` and was **not** using `ApolloQueryWatcher` but the `ApolloWatcher`._
   * _It didn't really serve any purpose, because it was different from the `ApolloCall` example only by the added `.watcher()` call. Its purpose wasn't to explain how to create an `ApolloQueryWatcher` and this is what it was doing. It was just a dead weight in the "RxJava Support" section._ 
3. `Response` added to the other example.
4. The old "Download" section was moved to where it belongs.


blipinsk:
@digitalbuddha any idea why travis might be failing?

Issue #1384, opened 2019-06-21
models are generated but not resolved


teknolojia:
I have been using this for about 3 months and everything was fine. But just recently I get **unresolved reference** every where apollo model is imported, every model is red. 
```
    //Apollo
    //implementation 'com.apollographql.apollo:apollo-runtime:1.0.0-alpha5'
    implementation 'com.apollographql.apollo:apollo-http-cache:1.0.0-alpha5'
    implementation "com.apollographql.apollo:apollo-coroutine-support:1.0.0-alpha5"

    //remove the snapshot when stable is released.
    implementation 'com.apollographql.apollo:apollo-runtime:1.0.1-SNAPSHOT'

```
I am using snapshot because of upload functionality.

I tried to move `apply plugin: 'com.apollographql.android'` just below  `apply plugin: 'com.android.application'` like sav007 did in the sample [here](https://github.com/apollographql/apollo-android/commit/8882237cfa40e07955dac83e5f88b30c5f776d88) but still everything is unresolved. What should I do
```
apply plugin: 'com.android.application'
apply plugin: 'kotlin-android'
apply plugin: 'kotlin-android-extensions'
apply plugin: 'kotlin-kapt'
apply plugin: "androidx.navigation.safeargs"
apply plugin: 'com.apollographql.android'
```

Dragons0458:
I have the same problem.

2 days ago they updated the plugin.

![imagen](https://user-images.githubusercontent.com/3160847/59953067-ba3c3500-943b-11e9-909f-7a5fcc8f16a7.png)

Thanks for your help :D

teknolojia:
@Dragons0458 for the time being I just removed snapshot. Hope we find solution soon.

AllanWang:
I think it's related to https://github.com/apollographql/apollo-android/issues/1381

Look in your *.iml file, and check that the test source line doesn't exist. You can also "reload" it by setting test to true then back to false.

In the end I'm probably just  going to add it as a srcDirs since even that solution is temporary

https://github.com/apollographql/apollo-android/issues/1372#issuecomment-502552241

Dragons0458:
Hello! with the #1381 my problem was solved, thank you very much! :D

sav007:
Just to clarify:

1) Is this issue happens for pure kotlin projects? 
2) Do you set in your apollo configuration `generateKotlinModels = true`?

AllanWang:
For me it's a pure kotlin project without generated kotlin models.

[...]

Issue #1381, opened 2019-06-19
*.iml problem with Android Studio


AllanWang:
It appears as though apollo constantly generates the following in the *.iml file:

```
<sourceFolder url="file://$MODULE_DIR$/build/generated/source/apollo/classes" isTestSource="true" generated="true" />
```

note that beforehand, there is another line:

```
<sourceFolder url="file://$MODULE_DIR$/build/generated/source/apollo/classes" isTestSource="false" generated="true" />
```

where `isTestSource` is false.

The line with `isTestSource="true"` seems to suggest to the IDE that the generated classes can only be used in the test folder, making it complain for all usages in the source.
If we remove that line, everything works, but it gets reset whenever the project is cleaned/built.

twobeeronebread:
Thanks @AllanWang this worked as a workaround for me too.

sav007:
Is that happening on pure kotlin project?

AllanWang:
Yes

sav007:
I believe it was fixed, closing for now.

PR #1396, opened 2019-06-27
Move after evaluate down to the place where it really needed

sav007:
Part of https://github.com/apollographql/apollo-android/issues/1352
Possible fix for https://github.com/apollographql/apollo-android/issues/1384 and https://github.com/apollographql/apollo-android/issues/1381


sav007:
CC @martinbonnin

Issue #1413, opened 2019-07-04
Generate visitor for compile-time safe consumption of Polymorphic datatypes 


narenkmanoharan:
Currently working on integrating Apollo Client into our network stack and we have a lot of polymorphic datatypes in our schema. Having to manually check the type of the object and then casting them is a pain. And so exposing a generated visitor for all every interface which has multiple implementations would allow for easier handling of data. **The primary benefit of this solution is that whenever a new implementation is introduced in the hierarchy the existing code does not compile until you handle the new type in the consuming visitor method.** I've demonstrated how the generated code would look like below. 

#### Generated visitor and the consuming method: 
```java
default <T> T visit(Visitor<T> visitor) {
  if (this instanceof AsHuman) {
    return visitor.visit((AsHuman) this);
  } else if (this instanceof AsDroid) {
    return visitor.visit((AsDroid) this);
  } else if (this instanceof AsCharacter) {
    return visitor.visit((AsCharacter) this);
  }
  return visitor.visitDefault(this);
}

interface Visitor<T> {
  T visitDefault(@NotNull Hero hero);

  T visit(@NotNull AsHuman asHuman);

  T visit(@NotNull AsDroid asDroid);

  T visit(@NotNull AsCharacter asCharacter);
}
```

#### Can be consumed as:

```java
hero.visit(object: Hero.Visitor<String> {
  override fun visitDefault(hero: Hero): String {
    TODO("not implemented")
  }

  override fun visit(asHuman: AsHuman): String {
    TODO("not implemented")
  }

  override fun visit(asDroid: AsDroid): String {
    TODO("not implemented")
  }

  override fun visit(asCharacter: AsCharacter): String {
    TODO("not implemented")
  }
})

hero.visit(new Hero.Visitor<String>() {

  @Override
  public String visitDefault(@NotNull Hero hero) {
    return TODO("not implemented");
  }

  @Override
  public String visit(@NotNull TestQuery.AsHuman asHuman) {
    return TODO("not implemented");
  }

  @Override
  public String visit(@NotNull TestQuery.AsDroid asDroid) {
    return TODO("not implemented");
  }

  @Override
  public String visit(@NotNull TestQuery.AsCharacter asCharacter) {
    return TODO("not implemented");
  }
);

```

Currently working on a PR to generate the above visitor. Will put it up soon.

sav007:
Fixed.

PR #1414, opened 2019-07-04
Add visitor spec to generate visitor interface and compatible visitor method for Java

narenkmanoharan:
[Related issue for context](https://github.com/apollographql/apollo-android/issues/1413)

Add visitors for easier and compile-time safe consumption of polymorphic data types in the schema.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/VisitorSpec.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/VisitorSpec.kt
@@ -13,24 +13,9 @@ class VisitorInterfaceSpec(
 ) {
 
   fun createVisitorInterface(): TypeSpec {
-    val visitDefaultMethod = MethodSpec.methodBuilder(DEFAULT_VISITOR_METHOD_NAME)
-      .returns(VISITOR_TYPE_VARIABLE)
-      .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
-      .addParameter(ParameterSpec.builder(
-        ClassName.get("", schemaType.simpleName()),
-        schemaType.simpleName().decapitalize()
-      ).addAnnotation(Annotations.NONNULL).build())
-      .build()
-
-    val visitMethods = implementations.map { classNames ->
-      MethodSpec.methodBuilder(VISITOR_METHOD_NAME)
-        .returns(VISITOR_TYPE_VARIABLE)
-        .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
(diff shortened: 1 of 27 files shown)


apollo-compiler/src/test/graphql/com/example/inline_fragment_for_non_optional_field/TestQuery.java
```diff
@@ -177,6 +177,23 @@ public Hero read(ResponseReader reader) {
...
     ResponseFieldMarshaller marshaller();
 
+    default <T> T visit(Visitor<T> visitor) {
+      if (this instanceof AsHuman) {
+        return visitor.visit((AsHuman) this);
+      } else if (this instanceof AsCharacter) {
+        return visitor.visit((AsCharacter) this);
+      }
+      return visitor.visitDefault(this);
+    }
```
sav007: So the default implementation of function in interface will lock Apollo to be used only with Java8.
Should we guard this feature by gradle plugin configuration flag then?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/VisitorSpec.kt
```diff
@@ -0,0 +1,93 @@
...
+        else -> visitorConsumer
+          .nextControlFlow(
+            "else if (this instanceof \$T)",
+            ClassName.get("", className.simpleName())
+          )
+          .addStatement(
+            "return visitor.visit((\$T) this)",
+            ClassName.get("", className.simpleName())
+          )
+      }
```
sav007: nit: expression 
```
.addStatement(
            "return visitor.visit((\$T) this)",
            ClassName.get("", className.simpleName())
          )
```
Can be moved out from both branches in when expression:
```
when {
}
.addStatement(
            "return visitor.visit((\$T) this)",
            ClassName.get("", className.simpleName())
          )
```

narenkmanoharan: done

apollo-compiler/src/test/graphql/com/example/inline_fragment_for_non_optional_field/TestQuery.java
narenkmanoharan: Yeah. My thinking here was that most people on Android are already on Java 1.8 compatibility. But yeah, that's a big assumption. I'm gonna add an option in the ApolloExtension to option into generating visitors and so only those who opt into it needs to add the source/target compatibility. I'll be sure to leave a note in the readme as well.

Issue #1426, opened 2019-07-12
Bad Kotlin generation for nullable enum type


michgauz:
Hello, I have found a small issue with Kotlin model generation.

I have : 
- a `Gender` enum class (Kotlin generated class)
- a nullable `Gender` field named `gender`

Then when reading the response, we have something like that : 
`val gender = Gender.safeValueOf(reader.readString(RESPONSE_FIELDS[6]))`

As `reader.readString(RESPONSE_FIELDS[6])` could be null and `safeValueOf` take a non-nullable parameter, it crashs

PR #1427, opened 2019-07-12
Allow only a non-nullable parameter for 'setValueOf'

michgauz:
This intends to fix issue : https://github.com/apollographql/apollo-android/issues/1426

Not sure this is the appropriate place to check the nullability...


sav007:
Thanks for the PR, but the fix is not so simple as it seems, we have to take the nullability of destination field into the account:
- if the target field (`gender` in your case) is nullable then generate the fix you put
- if the target field is not nullable then it should be as it was before `"%T.safeValueOf(%L.readString(%L))`

Another note just to give you a context if you touch any generation code, you need to update test fixtures, usually the flow is to run `com.apollographql.apollo.compiler.CodeGenTest` and then copy and replace generated files from `apollo-compiler/build/generated/source/apollo/classes/sources` to `/apollo-compiler/src/test/graphql`

I can take this issue and put a PR.

sav007:
Pls check this PR https://github.com/apollographql/apollo-android/pull/1428

I guess we can close this one.

michgauz:
Thanks for the feedback and the explanation 👍

Issue #1203, opened 2019-02-06
Detection of 'GlobalApolloCodegen' happens too early in Gradle configuration


tasomaniac:
When the following option is used `systemProp.apollographql.useGlobalApolloCodegen=true`, Apollo plugin spends some time to validate the installation of the global apollo codegen. 

This validation happens on configuration time. Even-though everything is cached in a module, this happens every-time, anybody runs any Gradle task in the project. 

This takes maybe around couple seconds but it also depends on the computer. Would be good to do this lazily.

PR #1447, opened 2019-07-20
Allow a noverify option for apollo codegen

ZacSweers:
This command gets run on every plugin configuration unnecessarily, and I think it's reasonable to add a noverify option for "I know what I'm doing" cases. Can add an info log statement if desired.

This adds a `apollographql.useGlobalApolloCodegen.noVerify` option to skip this


tasomaniac:
Interesting that you worked on this 2 days ago. I felt like I can took care of #1203 and I handled this without a flag. 

#1451 fixes the issue mentioned above too.

I basically moved this to a place where the actual code generation is run. There since the `apollo-codegen` tool is already run, #1451 does not really add any overhead.

ZacSweers:
That PR works too. I'm fine with either

tasomaniac:
The other one is merged. This can be closed now. Fingers crossed for the next release 🤞

ZacSweers:
👍

PR #1448, opened 2019-07-20
Use standard logging API for quieter codegen verification

ZacSweers:
Printing this to stdout means it gets printed every time and is quite noisy. This changes it to log the start and happy path to info, and warn on errors


sav007:
I think this PR will address this issue https://github.com/apollographql/apollo-android/pull/1451

tasomaniac:
Yep. I am going to update it real quick. Also these messages are not going to be this frequent anyways. They will only be logged when actual codegen runs.

sav007:
I believe we can close this as it was addressed as part of https://github.com/apollographql/apollo-android/pull/1451

PR #1451, opened 2019-07-22
Fix: Speed up build configuration time

### LINKED ISSUE #1203: Detection of 'GlobalApolloCodegen' happens too early in Gradle configuration

tasomaniac:
### Problem

When `useGlobalApolloCodegen` is used, installation is verified in every Gradle run causing slower builds. More info #1203 
This PR speeds up build configuration time by lazily verifying apollo-codegen installation.

### Considerations

Move Apollo global verification to the beginning of the task to speed up configuration process.

This I/O operation (version check) currently runs in configuration time and take up to 5sec. This runs in every Gradle execution (even when unrelated to Apollo)

This commit fixes #1203 and moves verification to a later stage.

This means that we cannot easily fall-back to local installation that easily. That's why an exception is thrown with an appropriate message. User should disable using this feature if they don't have global installation.

### Screenshot

Screen Shot 2019-07-24 at 00 41 40

### DIFF
--- a/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloSystemCodegenGenerationTask.groovy
+++ b/apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloSystemCodegenGenerationTask.groovy
@@ -2,6 +2,7 @@ package com.apollographql.apollo.gradle
 
 import org.gradle.api.GradleException
 import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.logging.Logger
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.AbstractExecTask
 import org.gradle.api.tasks.Input
@@ -20,7 +21,7 @@ class ApolloSystemCodegenGenerationTask extends AbstractExecTask<ApolloSystemCod
   ApolloSystemCodegenGenerationTask() {
     super(ApolloSystemCodegenGenerationTask.class)
     doFirst {
-      verifySystemApolloCodegenVersion()
+      verifySystemApolloCodegenVersion(logger)
     }
   }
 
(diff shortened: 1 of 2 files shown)


apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloSystemCodegenGenerationTask.groovy
```diff
@@ -32,4 +37,24 @@ class ApolloSystemCodegenGenerationTask extends AbstractExecTask<ApolloSystemCod
...
+      if (version == APOLLOCODEGEN_VERSION) {
+        println("Found required 'apollo-codegen@$APOLLOCODEGEN_VERSION' version.")
+        println("Skip apollo-codegen installation.")
+      } else {
+        throw new GradleException("Required 'apollo-codegen@$APOLLOCODEGEN_VERSION' version but found: $version. Consider disabling `apollographql.useGlobalApolloCodegen` property.")
+      }
+    } catch (Exception exception) {
+      throw new GradleException("Failed to verify system 'apollo-codegen' version. Consider disabling `apollographql.useGlobalApolloCodegen` property.", exception)
+    }
+  }
```
sav007: nit: should it be used `logger.info`  instead of `println` ?

tasomaniac: Addressed in https://github.com/apollographql/apollo-android/pull/1451/commits/b5eb9ba6d54ac6ebe211fbdaf5ba6f538f1cab80

tasomaniac:
DO NOT MERGE yet. I decided to take one more iteration and make this detection better.

tasomaniac:
Ready to go! Also updated the PR desc to include the output.

PR #1463, opened 2019-07-26
[Sample] Always use repo version in samples

tasomaniac:
## Problem

The samples use a certain published version of Apollo. When doing a release, it is super easy to forget to update sample versions. 
Also, this means that the current changes cannot be tested on samples since it requires publishing them in the first place. 

## Solution

Move samples into sub-folder names `samples`. Didn't use `apollo` prefix here to make a clear distinction between production code vs sample.

Use composite builds to always use the current apollo tooling.

This is also beneficial in the future if we add more samples (e.g non-android kotlin project)


apollo-compiler/build.gradle
```diff
@@ -1,8 +1,8 @@
 apply plugin: 'java'
 apply plugin: 'kotlin'
 
-targetCompatibility = JavaVersion.VERSION_1_6
-sourceCompatibility = JavaVersion.VERSION_1_6
+targetCompatibility = JavaVersion.VERSION_1_7
```
tasomaniac: ⚠️  This was needed since composite builds require all sub-projects to have compatible java version. 

This is the only place where I don't feel comfortable. Please let me know if this is an issue.

tasomaniac: Here is what I got without this change.

image

digitalbuddha:
We used to do similar as you are suggesting for nyt libs but ran into issues where a version did not publish and we were not aware since everything pointed to project rather than maven deps. With your change is there any safety to validate that new versions post correctly?

tasomaniac:
That's pretty much the same as what you described. But since it is easy to forget to update the version, using the samples for this purpose is not 100% safe against a version skipping publish. 

IMO, the publish task should only be successful if all artifacts are fully published. That's how I would put safety for that.

sav007:
Could you pls rebase.

tasomaniac:
Rebase done :+1:

PR #1470, opened 2019-08-01
Make androidPlugin dependency optional

tasomaniac:
The plugin works fine and quite well actually with kotlin or java only projects without Android. So having android plugin as dependency should not be mandatory. Non-android projects need to download the Android plugin and its all transitive dependencies which is quite big.

Having `compileOnly` should make this possible.

### DIFF
--- a/apollo-gradle-plugin/build.gradle
+++ b/apollo-gradle-plugin/build.gradle
@@ -26,6 +26,7 @@ dependencies {
   fixtureClasspath dep.androidPlugin
 }
 
+// Inspired by: https://github.com/square/sqldelight/blob/83145b28cbdd949e98e87819299638074bd21147/sqldelight-gradle-plugin/build.gradle#L18
 // Append any extra dependencies to the test fixtures via a custom configuration classpath. This
 // allows us to apply additional plugins in a fixture while still leveraging dependency resolution
 // and de-duplication semantics.
--- a/apollo-gradle-plugin/build.gradle
+++ b/apollo-gradle-plugin/build.gradle
@@ -14,6 +14,7 @@ dependencies {
   implementation dep.gradleNodePlugin
   implementation dep.moshi
 
+  testImplementation dep.androidPlugin
   testImplementation dep.junit
   testImplementation(dep.spock) {
     exclude module: 'groovy-all'
--- a/apollo-gradle-plugin/build.gradle
+++ b/apollo-gradle-plugin/build.gradle
@@ -7,11 +7,11 @@ sourceSets.main.groovy.srcDirs = ["src/main/java", "src/main/groovy"]
 
 dependencies {
   compileOnly gradleApi()
+  compileOnly dep.androidPlugin
 
   implementation localGroovy()
   implementation project(':apollo-compiler')
   implementation dep.gradleNodePlugin
-  implementation dep.androidPlugin
   implementation dep.moshi
 
   testImplementation dep.junit


tasomaniac:
Do you know what was wrong with CI? Rebased now in order to retry.

sav007:
>What went wrong:
Execution failed for task ':apollo-gradle-plugin:compileTestGroovy'.
> Unable to load class com.apollographql.apollo.gradle.ApolloPlugin due to missing dependency com/android/build/gradle/api/BaseVariant

sav007:
Somehow I guess your changes to make it optional breaks the compilation of tests

tasomaniac:
Hmm, I had this before in another Gradle plug-in project. Somehow the Gradle test setup is very special and has problems with these optional dependencies even though it works fine with real projects (when integrated).

I will have a look in detail. Thanks for finding the issue for me.

tasomaniac:
Now the test compilation is good. But the test projects still do not find the Android plugin. Tbh, with `plugins {}` in the tests, they should be found by id automatically. This also works in other projects. I couldn't figure it out.

tasomaniac:
@martinbonnin tests run successfully 👍I also added a comment. Please have a look.

martinbonnin:
We might have to revert this one unfortunately, I just tried with mavenLocal() and master and it fails :-(

It works on tests because we inject the android plugin to the GradleRunner but in other projects it will most likely fail because com.android.build.gradle.api.BaseVariant is needed to load the ApolloPlugin class.

I'll try once the snapshot reaches sonatype (somehow, it's still not there as I'm speaking) but I doubt it will work.

martinbonnin:
Confirmed with the snapshot from this afternoon.

tasomaniac:
Hmm, it was working for me with the samples. Do you have a regular Android project? Or is it a Kotlin project?

martinbonnin:
Non-android kotlin project.

tasomaniac:
I have a branch where I was going to add non-Android sample. I should open a PR soon.

I don't think reflection is needed. Just isolating Android API calls should be enough. Are you getting ClassNotFoundException?

martinbonnin:
Yes, ClassNotFoundException on BaseVariant when loading the ApolloPlugin class.

martinbonnin:
I think it was on oracle jdk 8 (I'm mobile now but I can check later). Maybe different JVM will load classes differently ?

martinbonnin:
Better stacktrace: 

    Caused by: java.lang.NoClassDefFoundError: com/android/build/gradle/api/BaseVariant
            at org.gradle.internal.reflect.ClassInspector.inspectClass(ClassInspector.java:69)
            at org.gradle.internal.reflect.ClassInspector.visitGraph(ClassInspector.java:55)
            at org.gradle.internal.reflect.ClassInspector.inspect(ClassInspector.java:35)
            at org.gradle.internal.instantiation.AbstractClassGenerator.inspectType(AbstractClassGenerator.java:213)
            at org.gradle.internal.instantiation.AbstractClassGenerator.generateUnderLock(AbstractClassGenerator.java:161)
            ... 162 more

That points to https://github.com/gradle/gradle/blob/v5.4.1/subprojects/model-core/src/main/java/org/gradle/internal/reflect/ClassInspector.java#L69.

So the class would load just fine but it seems that gradle, while asking reflectively for the methods of `ApolloPlugin` triggers the loading of `BaseVariant`

PR #1476, opened 2019-08-06
Modernize build and fix deprecated gradle API usages

ZacSweers:
This fixes a number of deprecated API usages found in my own project, and also brings the project up to speed with modern gradle and AGP versions


gradle/dependencies.gradle
```diff
@@ -19,7 +19,7 @@ ext.androidConfig = [
 ]
 
 ext.dep = [
-    androidPlugin           : 'com.android.tools.build:gradle:3.3.0',
+    androidPlugin           : 'com.android.tools.build:gradle:3.4.1',
```
DSteve595: 3.4.2?

apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloExtension.groovy
```diff
@@ -46,7 +46,7 @@ class ApolloExtension {
     outputPackageName = project.objects.property(String.class)
     outputPackageName.set("")
 
-    customTypeMapping = project.objects.property(Map.class)
+    customTypeMapping = project.objects.mapProperty(String.class, String.class)
```
tasomaniac: I believe these were included in Gradle 4.x. Readme says 4.3 is required. Would be good to double check if 4.3 is fine or need to update the requirement in the Readme.

gradle/dependencies.gradle
ZacSweers: Will fix 👍

apollo-gradle-plugin/src/main/groovy/com/apollographql/apollo/gradle/ApolloExtension.groovy
ZacSweers: Can check, but also  feel like only supporting current major version should be fine too. Gradle is already on 5.5, and these APIs are also not stable yet

tasomaniac: My Sample changes is merged. If you update your branch then you will have the sample. Sample now uses a separate project with composite builds. We can quickly try with different set of Gradle versions there. By changing the wrapper on the sample.

ZacSweers:
So the minimum supported gradle version for current AGP (3.4.2) is 5.1.1. I think that should settle it?

tasomaniac:
Hmm, think this as minSdk vs compileSdk on Android. This plugin depends on AGP 3.4.2 but works fine with older versions. In fact it work with java/Kotlin only projects where there is no AGP dependency at all. 

It's just that we need to figure out which versions we support at minimum and document that in README

tasomaniac:
Hmm, think this as minSdk vs compileSdk on Android. This plugin depends on AGP 3.4.2 but works fine with older versions. In fact it work with java/Kotlin only projects where there is no AGP dependency at all. 

It's just that we need to figure out which versions we support at minimum and document that in README

ZacSweers:
So I'm personally a big proponent of just supporting the current major version. For AGP that's 3.4.0, and for gradle that's 5.x. We can even actually easily set up tests for this on travis ([example](https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/.travis.yml#L9-L12)), but it's much harder to test this regularly if it requires lowering both AGP and gradle versions every time

tasomaniac:
Gradle testing APIs support integration testing the plugin across different versions. Testing against different version of AGP is also possible. Of course that is out of scope here.

tasomaniac:
Looks like there are some changes which affects the integration module

> Failed to apply plugin [id 'com.apollographql.android']
>    Could not get unknown property 'SourceDirectorySet' for source set android test of type com.android.build.gradle.internal.api.DefaultAndroidSourceSet.

ZacSweers:
Fixed the source directory set bit, just accidentally lost an import in a rebase.

As for testing gradle 4.3, even if I lower back down to AGP 3.3.0, the min gradle version is still 4.10.1. This does fail on one new source directory set API

3.2.1 requires gradle 4.6 (feb 2018). This also fails on the new source directory set API

4.3 was October 2017. Not sure


So in short - this would require gradle 5.x effectively. I think this is reasonable as it matches the current "major" AGP version as well

ZacSweers:
Note the test failures appear unrelated, and I think just a change in how the gradle test kit works now. Will investigate

ZacSweers:
I've partially addressed the new test requirements, but now I don't know what to do to fix all the tasks seeing themselves as up to date every test. I suspect it's something wrong with the incremental tasks logic in apollo itself

tasomaniac:
In another Gradle plugin, we've had similar issues while trying to update to Gradle 5.x. Another problem was that it would use much more memory which results in ~5x slower CI builds. https://github.com/novoda/gradle-static-analysis-plugin/pull/169

What about reverting the Gradle update. Or update to latest 4.10 but just fix depreciations. Then, that can be merged fast and then a PR updating Gradle can be open.

ZacSweers:
I don't quite understand, as 4.10 is what we started with and per my last comment, some of these deprecation fixes require 5.x. Would prefer to fix here rather than backtrack

tasomaniac:
I was just suggesting to take smaller steps so that this can be merged faster. Its just a suggestion. If it can be fixed here, that's better

ZacSweers:
Finally figured out the issue and resolved with 57f829c. Should be good to go now 🤞 

Would like to get resolution for the gradle versioning bit. I think supporting the latest gradle (5.x) and AGP (3.4.x, which requires gradle 5.1+) is a very sane baseline support strategy.

apollo-gradle-plugin/src/test/groovy/com/apollographql/apollo/gradle/unit/ApolloAndroidPluginSpec.groovy
```diff
@@ -73,8 +73,7 @@ class ApolloAndroidPluginSpec extends Specification {
 
     then:
     project.android.sourceSets.all { sourceSet ->
-      assert (sourceSet.extensions.findByName("graphql")) != null
```
tasomaniac: Really small comment. I find having hard coded `graphql` in tests is better in terms of quality assurance.

Imagine a case where somebody changes `GraphQLSourceDirectorySet.NAME` parameter, that will be a breaking change. The previous version of these tests will catch it whereas the new one will just accept them.

tasomaniac:
@ZacSweers can you also update the README mentioning the minimum Gradle requirement. I checked it with your branch, everything works as expected with `Gradle 5.1.1` (which is also what is required by the latest stable Android plugin)

ZacSweers:
@tasomaniac done and also opportunistically updated to gradle 5.6

sav007:
Is it ready to merge?

tasomaniac:
I think so.

PR #1504, opened 2019-08-16
[WIP] Convert apollo-api to kotlin

tasomaniac:
This PR converts the apollo-api module to Kotlin. Only the `internal` package with `Absent` and `Optional` are not converted.


apollo-runtime/src/main/java/com/apollographql/apollo/internal/InternalUtils.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Checks if two {@link Set} are disjoint. Returns true if the sets don't have a single common element. Also returns
+     * true if either of the sets is null.
+     *
+     * @param setOne the first set
+     * @param setTwo the second set
+     * @param <E>    the value type contained within the sets
+     * @return True if the sets don't have a single common element or if either of the sets is null.
+     */
+    static <E> boolean areDisjoint(Set<E> setOne, Set<E> setTwo) {
```
tasomaniac: Moved this here instead of converting to Kotlin since it was only used in `runtime` module.

apollo-api/src/main/java/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo.api
+
+import java.io.File
+
+data class FileUpload(@JvmField val mimetype: String, @JvmField val file: File)
```
tasomaniac: `@JvmField` is used for backward compatibility. Would be preferable to actually remove this but that would make a breaking change and Java clients need to use `getFile()` etc.

tasomaniac:
Currently integration tests fail because some code is not visible. I will have a look.

sav007:
Curious if we can avoid any JVM classes like: `IOException`, `java.util.Collections`, `java.util.Set` ?

Meantime I will investigate how to remove the http cache out from the api module.

tasomaniac:
@sav007 actually, all `java.util.Map`, `java.util.Set` are removed from the Kotlin code. 

I only have `java.util.Collections.unmodifiableList`. I basically kept that to be backward compatible in terms of behavior. Once this is done and released, I think we should do another iteration to make it `Kotlin Common` compatible (aka no `java.util` usage)

sav007:
@tasomaniac I think we should create a new module `apollo-api-core` move there files required only by code generation (such as `ResponseField`, `Query` etc.). The rest leave them in java and rename `apollo-api` to `apollo-api-jvm`. Make `apollo-api-jvm` depend on `apollo-api-core`.

The reason is not all classes that in current `apollo-api` module are needed for KN, only small subset. The rest required by current version of runtime.

What do you think?

tasomaniac:
This sounds good. This was a good exercise anyways. Would it be possible to turn these steps into tasks so that community can contribute more here? Or is it too involved?

Issue #1367, opened 2019-06-13
Trying to change graphql default directory to /java - Sub directories not kept for imports


bellierjonathan:
Hello,

I use this options to change `graphql` default directory folder with this :

```
schemaFilePath = project.projectDir.absolutePath + '/src/main/java/com/xxxxx/backendlibrary/network/schema.json'
outputPackageName = 'com.xxxxx.networklibraryGql'
```

(In my `graphql` default folder I had this tree) : 
```
/graphql/*.graphql
/graphql/user/*.graphql
/graphql/page/*.graphql
```
Before using this new option I was able to use this import : `com.xxxxx.networklibraryGql.user.****`. Apollo was builing my classes with keeping folders tree.

And when I use the new `outputPackageName` option it's not getting my sub folder `user` correctly like by default. And all of my generated classes are generated in the same folder.
All my classes are here `com.xxxxx.networklibraryGql.****`

Is it possible to keep my sub folders architecture with custom path?

sav007:
Yeah seems like we should do this.

tasomaniac:
What is the proposed solution here? outputPackageName is singular. Reading the name, I would expect it to cause all generation to have a single package name. 

If you want to have a hierarchy, have them under a package structure, just like your code and don't use outputPackageName. Wouldn't that simply reolve your issue?

sav007:
@tasomaniac 
> If you want to have a hierarchy, have them under a package structure, just like your code and don't use outputPackageName. Wouldn't that simply reolve your issue?

Yes this should work but I guess the intention here is to use relative folder structure as package with a base provided in `outputPackageName` configuration. 

So in original case the intended outcome will be:
```
com.xxxxx.networklibraryGql.*
com.xxxxx.networklibraryGql.user.*
com.xxxxx.networklibraryGql.page.*
``` 

Does it make sense?

tasomaniac:
Yes but you could easily achieve the same by putting them inside `com/xxxx/networklibrary` folders, no? I mean, you can potentially add this as a feature but it makes it more complex with not much additional benefit.

sav007:
I believe that in case of setting explicitly target package, the folder structure where queries are stored is not respected as subpackages, all generated models will be stored on the same provided package level.

Just a context, when you set explicitly path to schema it is required to provide a package where all models are been generated. One common use when you need to explicitly set location for schema is when it shared between iOS, Android possibly with the back end, outside of the Android project

martinbonnin:
[...]

PR #1507, opened 2019-08-17
First drop for rewriting the plugin in Kotlin

martinbonnin:
This is a first drop to rewrite the gradle plugin in Kotlin in order to make it more maintainable. To test, you can generate the artifacts locally (`./gradlew install`) and then apply `com.apollographql.apollo` instead of `com.apollographql.android`

    buildscript {
        repositories {
            mavenLocal()
        }
        dependencies {
            classpath("com.apollographql.apollo:apollo-gradle-plugin-kotlin:1.0.3-SNAPSHOT")
        }
    }
    plugins {
        id("com.apollographql.apollo")
    }

The new plugin drops support for the npm IR codegen to use the new antlr generator :rocket: . That simplifies the logic quite a lot!

The gradle DSL is not 100% backward compatible with the previous `com.apollographql.android` plugin. I'm actually not sure we want to support all the corner cases from the previous plugin. 2  things that break right now are:

1. paths configuration: This plugin assumes `schema.json`, `*.graphql` and `*.gql` are all in the same folder under `src/{sourceSetName}/{graphqlOrSomeOtherRoot}/com/example/...`. There's a lot of logic in [CodegenGenerationTaskCommandArgsBuilder.java](https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/main/java/com/apollographql/apollo/gradle/CodegenGenerationTaskCommandArgsBuilder.java) that I'm not sure I'm getting right. We certainly need to be able to customize the location to a certain extent but that could certainly be clarified.
2. build.gradle.kts compatibility: As the ApolloExtension is now in Kotlin, accessing properties with `setGenerateKotlinModels(true)` doesn't work anymore and it'll have to be written as `generateKotlinModels = true`. Depending how many users use the Kotlin DSL, that might or might not be an issue.

I'll continue working on this branch to improve backward compatibility but since it's a completely separate artifact, there should be no harm merging it in master either. And opening other pull requests for the specific points above (or others!).


sav007:
@martinbonnin oh you should probably have waited until new code gen released, as we can drop all tasks related to node, npm old codegen. There will be only 1 task left.

sav007:
Sorry never mind above comment, I see that it's a completely new plugin.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
```diff
@@ -0,0 +1,243 @@
...
+    const val TASK_GROUP = "apollo"
+  }
+
+  lateinit var apolloExtension: ApolloExtension
+
+  override fun apply(project: Project) {
+    apolloExtension = project.extensions.create("apollo", ApolloExtension::class.java)
+
+    // the extension block has not been evaluated yet, register a callback once the project has been evaluated
+    project.afterEvaluate {
```
sav007: That's was the main reason why extension was written on grovvy with magic of `Property` that will be evaluated later during execution, so they are placeholders for values. Is there any way we can do this in Kotlin?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
```diff
@@ -0,0 +1,243 @@
...
+      afterEvaluate(it)
+    }
+  }
+
+  private fun afterEvaluate(project: Project) {
+    val generateClassesTask = project.tasks.register("generateApolloClasses") {
+      it.group = TASK_GROUP
+    }
+
+    val androidExtension = project.extensions.findByName("android")
```
sav007: What about pure Kotlin projects? I guess this plugin won't support them right now?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
```diff
@@ -0,0 +1,243 @@
...
+
+    apolloExtension.validate()
+
+    when {
```
sav007: nit: should it be
```
when(androidExtension) {
  is LibraryExtension ->
  is AppExtension -> 
  else ->
```

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
```diff
@@ -0,0 +1,243 @@
...
+        val sourceSets = javaPlugin.sourceSets
+
+        sourceSets.forEach {
+          val apolloVariant = ApolloVariant(
+              name = it.name,
+              sourceSetNames = listOf(it.name)
+          )
+
+          val task = registerVariantTask(project, apolloExtension, apolloVariant) { task ->
+            if (!apolloExtension.generateKotlinModels) {
```
sav007: I believe if this is a plugin `apollo-gradle-plugin-kotlin` for Kotlin projects this code gen flag must be always true?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
```diff
@@ -0,0 +1,117 @@
...
+import com.apollographql.apollo.compiler.NullableValueType
+import org.gradle.api.Action
+
+open class ApolloExtension {
+
+  var nullableValueType = NullableValueType.ANNOTATED.value
+  var useSemanticNaming = true
+  var generateModelBuilder = false
+  var useJavaBeansSemanticNaming = false
+  var suppressRawTypesWarning = false
```
sav007: As this plugin for Kotlin projects should we remove these codegen configurations:
```
generateModelBuilder
useJavaBeansSemanticNaming
suppressRawTypesWarning
generateVisitorForPolymorphicDatatypes
```

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
```diff
@@ -0,0 +1,117 @@
...
+  fun services(action: Action<Services>) {
+    action.execute(services)
+  }
+
+  class Services {
+    val all = mutableListOf<Service>()
+
+    fun create(name: String, action: Action<Service>) {
+      val service = Service(name)
+      println("create $name")
```
sav007: nit: should it be removed?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  var services = Services()
+  var sourceSet = SourceSet()
+
+  /**
+   * Deprecated
+   */
+  fun sourceSet(action: Action<SourceSet>) {
+    action.execute(sourceSet)
+  }
```
sav007: why it's deprecated?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
```diff
@@ -0,0 +1,117 @@
+package com.apollographql.apollo.gradle.kotlin
+
+import com.apollographql.apollo.compiler.NullableValueType
+import org.gradle.api.Action
+
+open class ApolloExtension {
```
sav007: Question, current implementation of plugin invalidates `UP-TO-DATE` state if any changes made to the plugin configuration (don't remember how the tracking exactly works in Gradle), does this plugin provides the same feature? Like if we change the configuration for custom scalar type mapping will Gradle detect and invalidate the cache and re-run code gen?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: We can certainly do the same in Kotlin but that wouldn't remove the race between:

1. The Android variants or java/kotlin sourcesets are available
2. The buildscript is evaluated and the extension populated.

I believe it only works with the current plugin because the android plugin itself calls `afterEvaluate`

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: I'm thinking we could use the same plugin for kotlin and java, which is why I used the generic `com.apollographql.apollo` as a plugin id. The artifact name is suffixed by `-kotin` because it's written in kotlin but it should be applicable to other projects.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: I'll add a test for that.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: I was thinking the plugin could be used for java projects as well since we want to continue supporting java ?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: I'm not super comfortable with the sourceSet block. Things that bug me:

* I find the "sourceSet" naming weird. Does it relate to .graphql files or schema.json ? 
* Since we allow multiple schema.json, we should allow multiple "sourceSet"
* The packageName should be tied to a schema.json. Or computed automatically from each source file relative directory ? 

This is why I'm trying to introduce the `Service` class, similar to a `retrofit` service. A service would have:

* a schema.json
* a list of .graphql files
* it's own include/exclude filters

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: I can't find anymore the `afterEvaluate` I thought I had seen in the Android plugin...
Still, removing the `afterEvaluate` makes it implicit that the extension is populated when the tasks are configured.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: androidExtension could also be `InstantAppExtension` or something else. I'm not sure what to do in that case. I guess we can throw an exception

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
sav007: Originally `sourceSet ` was designed to give a support to provide explicitly schema location (only one schema for now, to support multiple we can support multiple  source sets) and exclude / include filters. It's related to both schema and GraphQL files that required (as source) to generate models.

In your case, could you please provide how is it going to look like in `gradle.build` plugin configuration section, to configure schema location, filters and target package?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: The test is already there actually :) https://github.com/martinbonnin/apollo-android/blob/a651a10cd6ddf00a160defcc6d97f6b08e063243/apollo-gradle-plugin-kotlin/src/test/kotlin/com/apollographql/apollo/gradle/AndroidLibraryTests.kt#L87

I guess this works because `customTypeMapping` is marked as @Input to the task ? So if it changes the task will be re-run ? Gradle magic :sparkles:

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: This is the part I'm not sure about. Do you know if anyone has graphql files in multiples subfolders ? Like so:

    src/main/graphql/com/example/schema.json
    src/main/graphql/com/example/movies/MovieQueries.graphql
    src/main/graphql/com/example/characters/CharactersQueries.graphql    

Being able to specify a single `outputPackageName` in that case feels weird as I would expect 

    com.example.movies.MovieQueries.kt
    com.example.characters.CharacterQueries.kt

martinbonnin: Currently, I'm thinking about exposing something like this: 

    apollo {
        services {
            create("githunt") {
                schemaFilePath = "graphql/com/githunt/schema.json"
                graphqlFilesFolder = "graphql/com/githunt/"
                exclude = listOf("README\\.md")
                // leave the default
                //include = listOf("*\\.graphql")
                // append apollo to not clash with other classes
                packageName = "com.githunt.apollo"
            }
            create("starwars") {
                schemaFilePath = "graphql/com/starwars/schema.json"
                // use defaults for everything else
            }
        }
    }

This obviously doesn't work if some users have multiple subfolders

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: Ultimately, we could have the plugin make `generateKotlinModels=true` by default if a kotlin plugin is detected.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: I updated the proposal above a bit. I think the most important change is that `packageName` should only affect the shared classes, not the individual classes. We can always map the current configuration and provide compatibility methods that would print a deprecation warning but I think this would be a good thing moving forward.

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: To build on my previous answer, we could move the validation of the apollo extension to the execution phase (when taskAction) is called and used `Property` everywhere. That should work since the extension is certainly populated by the time execution starts.

The only thing left would be the `generateKotlinModels` field that I believe is required earlier, to wire the `compileJava/compileKotlin` to the `codegenTasks` ([link](https://github.com/martinbonnin/apollo-android/blob/53e52ebfd0c8a74eb20af9b5addc7b4636c2db69/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt#L72))

That I'm not 100% sire how to do yet...

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
sav007: > Do you know if anyone has graphql files in multiples subfolders ? 
```
src/main/graphql/com/example/schema.json
src/main/graphql/com/example/movies/MovieQueries.graphql
src/main/graphql/com/example/characters/CharactersQueries.graphql 
```

Yeah I saw couple usages  of that, for example to group queries by feature.

> Being able to specify a single `outputPackageName` in that case feels weird as I would expect

I believe there is a bug to fix this, as you are right the `outputPackageName` should be a as a root package. 
https://github.com/apollographql/apollo-android/issues/1367

Agree on your comment / proposal there to rename `packageName ` to `rootPackageName `.

> `services {`

Is there any way to flatten it a little bit, like:
```
apollo {
    service("githunt") {
            // Places to look for graphql files relative to the current sourceSet (e.g src/main/{sourceFolder})
            // You can place .graphql or .gql files in sub directories to use different namespaces
            // By default, the plugin looks in the "graphql" directory
            sourceFolders = listOf("graphql/githunt")
            // Place where the schema file is relative to the current project. For multi-module setups, this can go outside the module but in
            // that case, you'll want to define packageName too else everything will end up at the root of the namespace
            // By default, the plugin will look for a schema.json file in sourceFolders.
            schemaFile = "../common/graphql/src/main/graphql/githunt/com/githunt/schema.json"
            // The package name used for common classes (Custom Types and Fragments).
            // By default, this is computed from the path of schemaFile relative to sourceFolders or the empty string if it the schemaFile is not a child of sourceFolders.
            // Note that this does not affect the packageName of queries and mutation.
            packageName = "com.githunt.apollo"
            // list of files to exclude
            exclude = listOf("**/schema.*")
        }
        service("starwars") {
            sourceFolders = listOf("graphql/starwars")
            // use defaults for everything else
        }
}
```

Another question `service("githunt")` I assume that `githunt` is a name, where it's going to be used?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
sav007: We can try to register both tasks for Java and Kotlin, and skip execution depending on the flag `generateKotlinModels ` and if Kotlin plugin is applied?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
sav007: Hm, interesting, I guess for now we can say it's unsupported and revisit later.

tasomaniac:
Hey, this is great work. 2 overall questions:

1) Would it be possible to keep the original tests so that we see how much compatible they are?
    - If the consumer project is in Kotlin, consuming a plugin written in groovy is already not ideal. I'm guessing that Kotlin users would be fine with little incompatibilities. But on the other hand, it would be great to keep it backward-compatible for regular Gradle consumers.
2) The change to make Android Plugin optional for non-Android users just went in #1511 #1470 and the next release going to have that. It looks like this reverts that. https://github.com/apollographql/apollo-android/pull/1507/files#diff-3b2becbc11a165eb91e36d977167d605R31 Would it be possible to keep that behavior?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
```diff
@@ -0,0 +1,247 @@
...
+    const val TASK_GROUP = "apollo"
+  }
+
+  lateinit var apolloExtension: ApolloExtension
+
+  override fun apply(project: Project) {
+    apolloExtension = project.extensions.create("apollo", ApolloExtension::class.java)
+
+    // the extension block has not been evaluated yet, register a callback once the project has been evaluated
+    project.afterEvaluate {
```
tasomaniac: This looks like a drawback compared to previous implementation. Would it be possible to use `project.plugins.withId` or similar dynamic callbacks to trigger the configuration more reactively based on other applied projects?

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: > Agree on your comment / proposal there to rename packageName to rootPackageName 

If we want to make it 100% backward compatible, we'll have to keep both for some time. I'll investigate this.

> Is there any way to flatten it a little bit 

Yes, this is also what SQL Delight is doing. The reason for the "create" is that a lot of apis come from gradle where everything is a `NamedDomainObjectContainer` so in groovy you could just write `githunt()` instead of `create("githunt")` and it would add an object named "githunt" automagically.

We could do that as well (See [here](https://github.com/square/sqldelight/blob/bc5a8a3d3a6169a4d73efdc4b54fb215e3ccfd55/sqldelight-gradle-plugin/src/main/kotlin/com/squareup/sqldelight/gradle/SqlDelightExtension.kt#L27) for the magic):

    apollo {
        githunt {
            sourceFolders
            ...
        }
        starwars {
            ...
        }
        generateKotlinModels = true
    }

I'm personally not a fan as it mixes types but this might appeal to groovy users.

martinbonnin:
@tasomaniac Thanks for the feedback !

About 1. Yes, this would be the idea. I'd like this plugin to be 100% backward compatible for Groovy users. Kotlin DSL users will have to make some adjustements (mainly use property access syntax instead of setters: replace `setGenerateKotlinModels(true)` with `generateKotlinModels = true`. Since it looks nicer in kotlin, I hope users can adopt that quickly...

Abuot 2. Yes, I'll backport (or is it a forward port ?? :D) the change

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: The issue is if we add an empty file set to the javaCompile input it'll break (see https://github.com/apollographql/apollo-android/pull/1353). Or can we tell everyone to skip the `compileJava` based on `generateKotlinModels` ?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: I'm not sure about that. See https://github.com/apollographql/apollo-android/pull/1507#discussion_r314950718. 

We could use properties and postpone most of the extension resolution to the execution phase. The problem is the `generateKotlinModels` that requires to wire/setup tasks differently.

Despite all the lazy configuration and properties magic, it looks like everyone is still relying on `afterEvaluate` for these use cases...

martinbonnin: I'm not sure about that. See #1507 (comment).

We could use properties and postpone most of the extension resolution to the execution phase. The problem is the generateKotlinModels that requires to wire/setup tasks differently.

Despite all the lazy configuration and properties magic, it looks like everyone is still relying on afterEvaluate for these use cases...

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: The main issue is https://github.com/apollographql/apollo-android/pull/1353. Once we add an empty source set to `compileJava` it will trigger an error... And I'm not sure if we can skip `compileJava` completely...

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloExtension.kt
martinbonnin: About the "githunt" name, it's not actually used, it just serves as an identifier, like `signingConfigs` or `flavors` in android builds. Right now I'm using it in the generated output folder but that's about it...
I don't mind too much removing it altogether:

    apollo {
        service {
            sourceFolders = 
            ...
        }
        service {
            ...
        }
        generateKotlinModels = true
    }

But it feels a bit weird

apollo-gradle-plugin-kotlin/src/main/kotlin/com/apollographql/apollo/gradle/kotlin/ApolloPlugin.kt
martinbonnin: I'm dumping this here while I'm thinking about it: We could have one artifact with 2 pluginIds. One would generate only java and the other only kotlin. Not sure what are the repercussions of that but that could be an option.

martinbonnin:
Superseded by https://github.com/apollographql/apollo-android/pull/1577

PR #1511, opened 2019-08-19
Fix ClassNotFoundException problem after #1470

tasomaniac:
## Problem

After #1470 was merged, non-Android pure Kotlin projects would not download/pull Android plugin anymore. This causes `ClassNotFoundException` on `BaseVariant` class from Android. 

## Solution

Split task creation for Android and Jvm only projects. `TaskConfigurator` is introduced to capture the common functionality. Yes, used inheritance so far 🤦‍♂ Groovy does not make it easy.

Android and Jvm subclasses encapsulates fully the task configuration for 2 different platforms.

**Note:** I verified that this works with @martinbonnin 's project: https://github.com/martinbonnin/testAndroidCompileOnly But it would be good if somebody else also does it.

### DIFF
--- a/apollo-gradle-plugin/build.gradle
+++ b/apollo-gradle-plugin/build.gradle
@@ -15,6 +15,7 @@ dependencies {
 
   implementation localGroovy()
   implementation project(':apollo-compiler')
+  implementation dep.guavaJre
   implementation dep.gradleNodePlugin
   implementation dep.moshi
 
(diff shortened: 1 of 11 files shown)


apollo-gradle-plugin/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
 
   implementation localGroovy()
   implementation project(':apollo-compiler')
+  implementation dep.guavaJre
```
tasomaniac: Android was pulling guava too and we were using it. Now, it is explicitly defined.

sav007:
It seems plugin tests don't like changes.

martinbonnin:
`ApolloPlugin.APOLLO_CODEGEN_GENERATE_TASK_NAME` became `TaskConfigurator.APOLLO_CODEGEN_GENERATE_TASK_NAME`.

tasomaniac:
Sorry folks, I fell asleep while waiting for the CI. It was late 😛

PR #1525, opened 2019-08-23
Update readme

martinbonnin:
* drop Rx1 support
* drop global apollo-codegen documentation (it is still in the code but there should be no need to advertise it anymore)
* add documentation about coroutines support
* add documentation about espresso support

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,14 @@ Apollo-Android is designed primarily with Android in mind but you can use it in 
 - [Consuming Code](#consuming-code)
 - [Custom Scalar Types](#custom-scalar-types)
 - [Support For Cached Responses](#support-for-cached-responses)
-- [RxJava Support](#rxjava-support)
+- [RxJava2 Support](#rxjava2-support)
+- [Coroutines Support](#coroutines-support)
+- [Android Espresso Support](#android-espresso-support)
 - [Gradle Configuration of Apollo Android](#gradle-configuration-of-apollo-android)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -505,6 +506,52 @@ disposable.clear();
...
+
+implementation 'com.apollographql.apollo:apollo-coroutines-support:x.y.z'
+```
+
+## Android Espresso support
+
+The Apollo GraphQL client comes with a [IdlingResource](https://developer.android.com/training/testing/espresso/idling-resource) to use during your Android UI tests.
+
+```kotlin
+IdlingRegistry.getInstance().register(ApolloIdlingResource.create("idlingResourceName", null))
```
tasomaniac: null? Afaik, the second is the ApolloClient. It's recommended to actually do this only once. I think we should make that clear. Registering IdlingResource with the same name twice would crash.

README.md
```diff
@@ -605,17 +652,6 @@ apollo {
 }
 ```
 
-### Use system pre-installed `apollo-codegen`
```
tasomaniac: Should we really remove this? I think it is still helpful until npm is completely removed

martinbonnin: The reasonning was that the README.md is for new users and we want to onboard them on the new CodeGen but I don't mind keeping that a bit longuer. I'll revert.

README.md
martinbonnin: Right. I'll add something

Issue #1533, opened 2019-08-23
New codegen sends the same fragment definition multiple times in the same query


martinbonnin:
This query:

```
fragment PullRequestFragment on PullRequest {
  title
}

fragment RepoFragment on Repository {
  pullRequests(first: 100) {
    nodes {
      ...PullRequestFragment
    }
  }
}

fragment RefFragment on Ref {
  associatedPullRequests {
    nodes {
      ...PullRequestFragment
    }
  }
}

query GetPullRequests {
  apollo: repository(owner: "apollographql", name: "apollo-android") {
    ...RepoFragment
    defaultBranchRef {
      ...RefFragment
    }
  }
  martin: repository(owner: "martinbonnin", name: "apollo-android") {
    pullRequests(first: 100) {
      nodes {
        ...PullRequestFragment
      }
    }
  }
}
```

Generates this code:

```    val QUERY_DOCUMENT: String = """
        |query GetPullRequests {
        |  apollo: repository(owner: "apollographql", name: "apollo-android") {
        |    __typename
        |    ...RepoFragment
        |    defaultBranchRef {
        |      __typename
        |      ...RefFragment
        |    }
        |  }
        |  martin: repository(owner: "martinbonnin", name: "apollo-android") {
        |    __typename
        |    pullRequests(first: 100) {
        |      __typename
        |      nodes {
        |        __typename
        |        ...PullRequestFragment
        |      }
        |    }
        |  }
        |}
        |fragment PullRequestFragment on PullRequest {
        |  __typename
        |  title
        |}
        |fragment RepoFragment on Repository {
        |  __typename
        |  pullRequests(first: 100) {
        |    __typename
        |    nodes {
        |      __typename
        |      ...PullRequestFragment
        |    }
        |  }
        |}
        |fragment RefFragment on Ref {
        |  __typename
        |  associatedPullRequests {
        |    __typename
        |    nodes {
        |      __typename
        |      ...PullRequestFragment
        |    }
        |  }
        |}
        |fragment PullRequestFragment on PullRequest {
        |  __typename
        |  title
        |}
        """.trimMargin()
```

The server returns an error because `PullRequestFragment` is defined twice

PR #1536, opened 2019-08-23
Fix issue with duplicated fragment definition in query

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1533


martinbonnin:
Works like a charm :+1: :rocket:

Issue #1522, opened 2019-08-23
Code generation change 1.0.2 -> 1.1.0


AOrobator:
When upgrading from 1.0.2 to 1.1.0, my models no longer respect the package they are declared in. They now always choose the top level. It was nice to have the separation.

martinbonnin:
Can you provide more details ? Ideally a small repository to reproduce or the layout of your source and generated files before and after the upgrade ?

AOrobator:
I have mutations and queries that live in the signup package. 
<img width="290" alt="Screen Shot 2019-08-23 at 9 24 47 AM" src="https://user-images.githubusercontent.com/4106884/63595862-f305a100-c587-11e9-8eef-b96b4d2b0b07.png">

on v.1.0.2, if I wanted to import a query, I'd use

`com.geneva.chat.api.signup.MyQuery`

on v1.1.0, I have to use

`com.geneva.chat.api.MyQuery`

because the package it's in is ignored.

sav007:
Hm, interesting. Will check why but it seems latest changes formatting package name by schema location rather than query location.

Issue #1516, opened 2019-08-21
Add `./gradlew :downloadXYZSchema` 


martinbonnin:
We could add a task inside the gradle plugin to automatically download/update schema.json. 

With the new DSL proposed in https://github.com/apollographql/apollo-android/pull/1507, that could be something like below:

    apollo {
        service("githunt") {
                sourceFolders = listOf("graphql/githunt")
                schemaFile = "src/main/graphql/githunt/com/githunt/schema.json"
                                
                // specifying the endpoint will add a  `updateGithuntSchema` task. Cal ./gradlew updateGithuntSchema to update the schema
                enpoint {
                    url = "https://api.githunt.com/graphql"
                    // queryParameters and headers will be used when updating the schema
                    queryParameters("apiKey", "API_KEY_HERE")
                    headers("Authorization", "Bearer TOKEN_HERE")
                }
        }
    }

martinbonnin:
Closed in the incubating plugin

PR #1539, opened 2019-08-24
Fix issue generated package name doesn't respect graphql subfolders

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1522
Closes https://github.com/apollographql/apollo-android/issues/1367

Check the tests to get the idea of rules applied when formatting target package names:
https://github.com/apollographql/apollo-android/pull/1539/files#diff-dab7ed2442415d0be3727ce680c61c19

CC: @martinbonnin

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomEnumTypeSpecBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomEnumTypeSpecBuilder.kt
@@ -49,6 +49,6 @@ class CustomEnumTypeSpecBuilder(
           .build()
 
   companion object {
-    fun className(context: CodeGenerationContext): ClassName = ClassName.get(context.packageNameProvider.typesPackageName(), "CustomType")
+    fun className(context: CodeGenerationContext): ClassName = ClassName.get(context.packageNameProvider.typesPackageName, "CustomType")
   }
 }
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/FragmentsResponseMapperBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/FragmentsResponseMapperBuilder.kt
@@ -106,7 +106,7 @@ class FragmentsResponseMapperBuilder(
       fragments
           .map { it.type.unwrapOptionalType(withoutAnnotations = true) as ClassName }
           .map {
-            val mapperClassName = ClassName.get(context.packageNameProvider.fragmentsPackageName(), it.simpleName(),
+            val mapperClassName = ClassName.get(context.packageNameProvider.fragmentsPackageName, it.simpleName(),
                 Util.RESPONSE_FIELD_MAPPER_TYPE_NAME)
             FieldSpec.builder(mapperClassName, it.mapperFieldName(), Modifier.FINAL)
                 .initializer(CodeBlock.of("new \$T()", mapperClassName))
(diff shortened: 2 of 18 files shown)


martinbonnin:
Hi! Thanks for looking into this. A few questions/comments:

1. Are we ok to break backward compatibility ? This PR will most likely require some users to adapt (those relying on flattening the package hierarchy or having the schema.json outside src/main/graphql). Also with this PR, schema.json is required to be a child of a `graphql`directory. I don't think we want to enforce that for people sharing their schema.
2. The more I think about it, the less I think we need the `rootPackageName` option. It's only needed if the `schema.json` is outside the regular location. Without `rootPackageName`, the rules could be as simple as:
2.1 `Types` and `Fragment` use the directory of  the`schema.json` file.
2.2 `Operations` use the directory of the `.graphql` files they are located in.
2.3 If you want to use a `schema.json` outside the regular location, we provide `schemaFilePath` and `schemaFileLocation` for that.

Also, I was planning to deprecate a lot of these options in the incubating plugin at the same time as proper multi-schema support. We might as well do everything there.

sav007:
@martinbonnin 

> Are we ok to break backward compatibility ? This PR will most likely require some users to adapt (those relying on flattening the package hierarchy or having the schema.json outside src/main/graphql). 

`flattening the package hierarchy` it's a bug not a feature, actually I believe that old code generator was preserving the subfolder structure, and it was broken for new one. So we shouldn't expect of breaking changes.

> Also with this PR, schema.json is required to be a child of a graphqldirectory. I don't think we want to enforce that for people sharing their schema.

If I'm right this requirement was from day one to keep everything in `src/*/graphql` folder, schema, queries. Check `com.apollographql.apollo.gradle.GraphQLSourceDirectorySet` line with
```
 SourceDirectorySet sourceSet = objectFactory.sourceDirectorySet(NAME, String.format("%s GraphQL source", NAME));
   >>>>  sourceSet.srcDir("src/" + name + "/graphql");
```

> The more I think about it, the less I think we need the rootPackageName option. It's only needed if the schema.json is outside the regular location. Without rootPackageName, the rules could be as simple as:

Yeah, maybe you are right. Having `schemaFilePath` should be enough. Except the case where you want to make it the  schema and queries locations to be flat (`src/main/grapqhl`) but provide the root package for generate models (`com.sample.api`).

martinbonnin:
> flattening the package hierarchy it's a bug not a feature, actually I believe that old code generator was preserving the subfolder structure, and it was broken for new one. So we shouldn't expect of breaking changes.

I agree it's a bug. It has been there for a long time though. I believe specifying `outputPackageName` has been flattening the hierarchy for both npm and new codegen since the introduction of `outputPackageName`. I'm not sure what broke with the antlr parser since that touched mostly the IR generation, not too much the actual classes.

> If I'm right this requirement was from day one to keep everything in src/*/graphql folder, schema, queries. 

That's only a requirement for the .graphql operation files. The schema file could be in any directory as long as `schemaFilePath` was specified. I don't know how many users had that but since the original use case was to share schema.json with ios clients, it's likely that they don't reside under a `src/main/graphql` tree.

sav007:
Yeah I see your point. I think we will have to introduce this breaking changes eventually to fix the issue, do you think if it's a good time right now to do it, as we bump major release version from 1.0.2 to 1.1.0? Or should we keep behavior as it was before and fix it later?

martinbonnin:
@digitalbuddha we can provide a deprecation period for people to update before we break the api. Also, if we break this api, I'd like to introduce proper multi-schema handling (see https://github.com/apollographql/apollo-android/pull/1507#discussion_r315400046). That could also bring support for downloading the schema without having to rely on node (https://github.com/apollographql/apollo-android/issues/1516). I started working on this in a kotlin-written plugin. I'll update it.

sav007:
Out of date.

PR #1544, opened 2019-08-27
Fix 1522 package name regression

martinbonnin:
This PR should fix https://github.com/apollographql/apollo-android/issues/1522 and make the ExperimentalApolloCodegenTask behave like the regular one. 

It doesn't fix https://github.com/apollographql/apollo-android/issues/1367 yet as over time the logic that governs schema location/package name has grown complex and it's hard to make it evolve without breaking the existing. I'd like to do that once we can simplify the plugin and rewrite it in kotlin (see https://github.com/apollographql/apollo-android/pull/1507)


martinbonnin:
CI is failing. It's also failing on latest master (https://travis-ci.org/apollographql/apollo-android/builds/577042272).

That sounds like a corrupted cache. It looks like `~/.gradle/wrapper/dists/gradle-5.6-all/4chhpeaqtrfx4ih6mafwwr40u/gradle-5.6/lib//ant-1.9.14.jar` is missing.
@sav007 can you delete the travis cache and relaunch the build ?

PR #1551, opened 2019-08-28
Add serverUrl setter

BenjaminFaal:
In certain situations you want to be able to change the server URL without creating a whole new client. For example a Spring Bean.


sav007:
`ApolloClient` is designed as completely immutable class, if you do want to change the server url it should be a new instance. The recommended  way here would be to introduce API that will create a new `ApolloClient.Builder` from existing instance of `ApolloClient` (similar way how OkHttpClient does with `okhttp3.OkHttpClient#newBuilder`)

sav007:
Out of date.

PR #1555, opened 2019-08-29
Remove old NPM code generator

sav007:
- clean up gradle plugin
- refactor gradle plugin tests by removing anything related to Node tasks
- remove reference to old apollo codegen npm module
- clean up graphql compiler tests


sav007:
cc @martinbonnin

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
```diff
@@ -114,6 +106,8 @@ class CodeGenTest(val pkgName: String, val args: GraphQLCompiler.Arguments) {
       return File("src/test/graphql/com/example/")
           .listFiles()
           .filter { it.isDirectory }
+          // TODO figure out what to do with these cases
+          .filter { it.name != "nested_inline_fragment" && it.name != "reserved_words" && it.name != "scalar_types" }
```
martinbonnin: Is this something we don't support with the new antlr IRGen ?

martinbonnin:
Awesome! I was going to ask when we could remove this code.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
sav007: well the issue with these test fixtures that they were generated manually, there are no corresponding GraphQL queries. IR.json files for these test were written by hand. With new code gen we need to modify schema itself, it's very tedious task. I will try to do this in next PR.

Issue #1573, opened 2019-09-05
Documentation is not helpful for setting up Apollo in a non-Android environment.


danielPerez97:
The project is called apollo-android but the documentation says you can use this for a non-android setup with gradle. However, when it comes time to apply the plugin in the instructions with the sample setup, `apply plugin: 'com.apollographql.android'` is used. 

Could non-android users get clearer documentation for setting it up in an environment that isn't android? And may I ask why is this project not taking more of an SQLDelight approach and being agnostic to the environment apart from the extra artifacts it might have and instruct you to use in their README.md?

Thanks for any help.

sav007:
Non android users can use apollo as long as Gradle is used as build tool. There is no need to have 
`apply plugin: 'com.apollographql.android'`

>SQLDelight approach and being agnostic to the environment apart from the extra artifacts

Not sure I got the comment, but again you can use apollo outside of android projects.

danielPerez97:
https://github.com/cashapp/sqldelight

They show off SQLDelight as a general-purpose SQLite library, with compatibility for JVM and Android(and iOS because of Kotlin/Native, but this library doesn't use that). They don't purport that it is an Android-first library, and show examples for setting it up in each environment. Why does Apollo not do the same?

 `apply plugin: 'com.apollographql.android'`

I understand this library is Android-first, but the merits of it's usage go beyond Android as well. I think it would be a wise to be a bit more agnostic of Android but show how to use it in both JVM and Android platforms, per SQLDelight's style. I'll try to throw in a pull request this weekend.

martinbonnin:
@sav007 `apply plugin: 'com.apollographql.android'` is needed, even for non-android projects, right ?

For what it's worth, the plugin rewrite I'm working on (https://github.com/apollographql/apollo-android/pulls) will use `com.apollographql.apollo` as plugin id and deprecate `com.apollographql.apollo`. That should clear up some confusion. I'll update the PR this weekend. 

Also, we should certainly rename the repo from `apollo-android` to something else at some point, especially with multiplatform in mind (https://github.com/apollographql/apollo-android/issues/1455). Not sure what a good name would be though `apollo-jvm` doesn't work for native `apollo-kotlin` doesn't work with java. Maybewe should just call that `graphqldelight` ;)

tasomaniac:
Really looking forward to seeing the new plugin. New id makes sense. 

[...]

PR #1578, opened 2019-09-09
Added wordage to clarify non-Android usage.

danielPerez97:
Added this to help clarify to non-Android users how to set up this project. #1573

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ buildscript {
   }
 }
 ```
+(If you are not using Android, this whole block will go at the top of your build.gradle file, not settings.gradle!)
 
 Then add the following to your app's build.gradle dependencies:
 
@@ -85,11 +86,12 @@ dependencies {
 }
 ```
 
-The plugin can then be applied as follows within your app module's `build.gradle` :
+The plugin can then be applied as follows within your app module's `build.gradle`, or :
 
 ```groovy
 apply plugin: 'com.apollographql.android'
 ```
+(If you are not on Android, still apply this artifact in your build.gradle file, after the `buildscript` block.)
 
 The Android Plugin must be applied before the Apollo plugin.
 For Kotlin users, the Apollo plugin must be applied before the Kotlin plugins.


README.md
```diff
@@ -47,6 +47,7 @@ buildscript {
   }
 }
 ```
+(If you are not using Android, this whole block will go at the top of your build.gradle file, not settings.gradle!)
```
martinbonnin: It shouldn't go in settings.gradle on Android either.

martinbonnin:
@danielPerez97 Thanks for looking into this. I don't think the procedure is that different for `Android` vs `non-Android`. Maybe renaming the main title of the page from `Apollo GraphQL Client for Android` to `Apollo GraphQL Client for Android and other JVM` would be enough ?

README.md
danielPerez97: Right, I think the instructions are clear for how to do it in Android. but when you say things like `The plugin can then be applied as follows within your app module's build.gradle :`, it implies two build.gradle files, and since it was project/app module level, I figured I had to put it in `settings.gradle`. Putting the `buildscript` block in `settings.gradle` was the mistake I made until I figured out to put it in my normal(and only) `build.gradle`. Part of it was me being an idiot, but clarification for JVM can't hurt, no?

danielPerez97:
I agree renaming would be appropriate, clarifying instructions is what I think should be achieved atm.

README.md
digitalbuddha: Settings.gradle is never mentioned. If we need to clarify not to put it there should we also clarify to not put it in gradle.properties or somewhere else. I would suggest talking about where to put it rather than mentioning where not to put it

danielPerez97: 'settings.gradle' never got mentioned because this looks like an android-first library and I haven't seen a 'settings.gradle' file in an android project. It isnt my library and if it continued to look android-first that would be fine, I just want to suggest clarification in the documentation for non-android users.

martinbonnin: As @digitalbuddha mentionning places *not* to put the plugin adds to the cognitive load so it might actually confuse users more.

settings.gradle is not related to android or non-android. I think it has been there for a long time in gradle, especially for [multi-module projects](https://guides.gradle.org/creating-multi-project-builds/)

sav007:
Out of date.

PR #1586, opened 2019-09-13
Coroutines: Swallow exception when calling channel.offer on a closed channel

### LINKED ISSUE #1585: Coroutines: Channel.offer can still be called after channel has been closed.
Hi!

We use a wrapper around `ApolloSubscriptionCall<T>.toChannel` extension and we're having some crashes although we use it inside a `CoroutineScope` with an exception handler.

Error:
```
com.apollographql.apollo.internal.subscription.ApolloSubscriptionTerminatedException: 
  at com.myapp.extensions.ApolloExtensionsKt$toChannelHandleTerminate$2.onTerminated (ApolloExtensions.kt:39)
  at com.apollographql.apollo.internal.RealApolloSubscriptionCall$SubscriptionManagerCallback.onTerminated (RealApolloSubscriptionCall.java:168)
  at com.apollographql.apollo.internal.subscription.RealSubscriptionManager.onConnectionClosed (RealSubscriptionManager.java:284)
  at com.apollographql.apollo.internal.subscription.RealSubscriptionManager$SubscriptionTransportCallback$4.run (RealSubscriptionManager.java:457)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:764)
```

After investigation, we found that this is caused by some concurrency issues inside `ApolloSubscriptionCall.Callback`.

The method `channel.offer(response)` can still be called after the channel has been closed. And since the callback use a separate Thread, the exception isn't caught by the coroutine's exception handler, causing a crash.

EtienneTheScore:
Fixes #1585

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -135,7 +135,13 @@ fun <T> ApolloSubscriptionCall<T>.toChannel(capacity: Int = Channel.UNLIMITED): 
         }
 
         override fun onResponse(response: Response<T>) {
-            channel.offer(response)
+            try {
+                channel.offer(response)
+            } catch (ex: Exception) {
+                // Swallow exception if the channel is closed.
+                // Due to concurrency, this can be called between channel.close and cancel().
+                // For the same reason, we cannot use channel.isClosedForSend.
+            }
         }
 
         override fun onFailure(e: ApolloException) {
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -14,7 +14,13 @@ import kotlin.coroutines.resumeWithException
 private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall.Callback<T>() {
 
     override fun onResponse(response: Response<T>) {
-        channel.offer(response)
+        try {
+            channel.offer(response)
+        } catch (ex: Exception) {
+            // Swallow exception if the channel is closed.
+            // Due to concurrency, this can be called between channel.close and cancel().
+            // For the same reason, we cannot use channel.isClosedForSend.
+        }
     }
 
     override fun onFailure(e: ApolloException) {
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -18,8 +18,6 @@ private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall
             channel.offer(response)
         } catch (ex: Exception) {
             // Swallow exception if the channel is closed.
-            // Due to concurrency, this can be called between channel.close and cancel().
-            // For the same reason, we cannot use channel.isClosedForSend.
         }
     }
 
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -8,6 +8,7 @@ import com.apollographql.apollo.api.Response
 import com.apollographql.apollo.exception.ApolloException
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.ClosedSendChannelException
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 
@@ -16,8 +17,10 @@ private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall
     override fun onResponse(response: Response<T>) {
         try {
             channel.offer(response)
-        } catch (ex: Exception) {
-            // Swallow exception if the channel is closed.
+        } catch (ex: ClosedSendChannelException) {
+            // Swallow exception if the channel has already been closed without a cause.
+        } catch (ex: ApolloException) {
+            // Swallow exception if the channel has already been closed with an ApolloException.
         }
     }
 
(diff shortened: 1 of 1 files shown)


apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -14,7 +14,11 @@ import kotlin.coroutines.resumeWithException
 private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall.Callback<T>() {
 
     override fun onResponse(response: Response<T>) {
-        channel.offer(response)
+        try {
+            channel.offer(response)
+        } catch (ex: Exception) {
+            // Swallow exception if the channel is closed.
```
martinbonnin: Can this catch block be more specific and not catch all Exceptions ?

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -135,7 +139,13 @@ fun <T> ApolloSubscriptionCall<T>.toChannel(capacity: Int = Channel.UNLIMITED):
         }
 
         override fun onResponse(response: Response<T>) {
-            channel.offer(response)
+            try {
+                channel.offer(response)
+            } catch (ex: Exception) {
```
martinbonnin: Same here.

martinbonnin:
Thanks! Can add a test case in `CoroutinesApolloTest` ? That'll help understanding the issue as catching  all exceptions could potentially hide some some legit errors.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
EtienneTheScore: It's possible to just catch `ClosedSendChannelException` (in case the channel is closed without cause) and `ApolloException` (in case the channel is closed by `onFailure(e: ApolloException)`)

EtienneTheScore:
Yes, I didn't find any test regarding subscriptions in `CoroutinesApolloTest`.

I'm almost certain that it can only happens on subscriptions since it's a concurrency issue.
I still preferred to also add this try catch to `ChannelCallback` just in case `onResponse` is called after `onFailure` or `onStatusEvent(ApolloCall.StatusEvent.COMPLETED)` for one reason or another.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -135,7 +142,17 @@ fun <T> ApolloSubscriptionCall<T>.toChannel(capacity: Int = Channel.UNLIMITED):
...
+            try {
+                channel.offer(response)
+            } catch (ex: ClosedSendChannelException) {
+                // Swallow exception if the channel has already been closed without a cause.
+                // Due to concurrency, this can be called between channel.close and cancel().
+                // For the same reason, we cannot use channel.isClosedForSend.
+            } catch (ex: ApolloException) {
+                // Swallow exception if the channel has already been closed with an ApolloException.
+                // Due to concurrency, this can be called between channel.close and cancel().
+                // For the same reason, we cannot use channel.isClosedForSend.
```
martinbonnin: So the scenario would be?

* `onFailure(apolloSubscriptionTerminatedException)` is called, that closes the channel.
* `onResponse(response)` is called. We try to offer on a now closed channel.

That contradicts https://github.com/apollographql/apollo-android/blob/9a0e00364c3d52e4f0ae7fc5f17c274f11a32d4f/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloSubscriptionCall.java#L67: onResponse should not be called after onFailure. Or is there another path leading here ?

martinbonnin:
I can reproduce something similar to you case by tweaking a bit the extensions (see https://github.com/martinbonnin/apollo-android/commit/f6eac0cd46154551e649882c902c3a05eb199e19)

This obviously exagerates the delay but I still believe it's representative of a valid use case.

Calling `channel.cancel()` will produce a `CancellationException: LinkedListChannel was cancelled` so this should maybe be caught as well.

All in all:
1. I don't see how the `ApolloException` case can happen without breaking the `onFailure` contract (see my comment above). If that's the case, we should certainly fix the bug there.
2. We should certainly handle both `ClosedSendChannelException` and `CancellationException`
3. I'm a bit uncomfortable with the apollo callback being called from the apollo dispatcher. I think callback.onResponse, etc... should be called from the thread that called `enqueue` in the first place (potentially configurable with a custom executor).

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
EtienneTheScore: I didn't find how this is possible but we had that kind of crash:
```
com.apollographql.apollo.exception.ApolloNetworkException: 
  at com.apollographql.apollo.internal.RealApolloSubscriptionCall$SubscriptionManagerCallback.onNetworkError (RealApolloSubscriptionCall.java:150)
  at com.apollographql.apollo.internal.subscription.RealSubscriptionManager$SubscriptionRecord.notifyOnNetworkError (RealSubscriptionManager.java:409)
  at com.apollographql.apollo.internal.subscription.RealSubscriptionManager.onTransportFailure (RealSubscriptionManager.java:226)
  at com.apollographql.apollo.internal.subscription.RealSubscriptionManager$SubscriptionTransportCallback$2.run (RealSubscriptionManager.java:439)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:764)
Caused by: java.io.EOFException: 
  at okio.RealBufferedSource.require (RealBufferedSource.java:61)
  at okio.RealBufferedSource.readByte (RealBufferedSource.java:74)
  at okhttp3.internal.ws.WebSocketReader.readHeader (WebSocketReader.java:117)
  at okhttp3.internal.ws.WebSocketReader.processNextFrame (WebSocketReader.java:101)
  at okhttp3.internal.ws.RealWebSocket.loopReader (RealWebSocket.java:274)
  at okhttp3.internal.ws.RealWebSocket$2.onResponse (RealWebSocket.java:214)
  at okhttp3.RealCall$AsyncCall.execute (RealCall.java:203)
  at okhttp3.internal.NamedRunnable.run (NamedRunnable.java:32)
```

martinbonnin: What about:

```kotlin
            try {
                channel.offer(response)
            } catch (ex: ClosedSendChannelException) {
                // Swallow exception if the channel has already been closed.
            } catch (ex: CancellationException) {
                // Swallow exception if the channel has already been cancelled.
            } catch (ex: ApolloException) {
                throw Exception("onResponse called after onFailure", ex)
           }
```

This will most likely not fix your issue in the last case but it will give more information where the problem might come from ?

sav007:
Out of date.

PR #1592, opened 2019-09-16
close #1008 Allow shared custom types

### LINKED ISSUE #1008: Support shared custom types with multiple modules
If you have two modules (let's call them A and B) which both have queries that use a shared common custom type, the `CustomType` enum values will get generated in both. Also, if those two modules have different output package names, that `CustomType` class will be duplicated in each. If the same package name is desired in both, then there is a conflict of class definitions.

If you're willing to accept the duplication by using different package names, now the module in which you create your `ApolloClient` (let's call this module C) must depend on both A and B because they have different `CustomType` enums and you need to add the appropriate call to `addCustomTypeAdapter()`. That unfortunately ties all of these modules into one big build dependency tree.

I believe the solution is to allow specifying external `CustomType` enums. You could then generate the `CustomType` in C, and A and B would have a dependency on C. This way A and B will not have to generate the `CustomType` enums and they can use the same `CustomType` enum in C. The `ApolloClient` created in C can now call `addCustomTypeAdapter()` without needing a dependency on A and B. A prerequisite of building this feature would probably be enabling running apollo-codegen on a module without a query, specifying custom types to generate. As far as I know this is not currently possible. You would, of course, have to be able to explicitly specify custom types as there would be no way of knowing which queries are in use (and thus which custom types you would need for these queries) unless you wanted to generate custom types for the entire schema which seems not ideal. I don't think this is an issue from the user side as you have to create adapters for these custom types anyway.

This might be a bit complicated given the interaction with apollo-cli, unfortunately, but seems theoretically possible, and has benefits for larger projects.

budius:
The suggested approach for the issue ([LINK HERE](https://github.com/apollographql/apollo-android/issues/1008)) was to somehow the gradle plugin would track in-between modules compilation but that has two issues IMHO:
- it's really complex to execute,
- it doesn't cope with a module from an AAR

The suggested approach here uses from the facts that:
- ScalarType already has a typeName() parameter of String; and
- The typename is unique within a GraphQl Server

From this approach it is just a matter of mapping from a String (instead of the class) to the desired type.  
The gradle plugin will re-generate CustomType with their own separate package name for each module,
but those will be ProGuarded and the type defined on the ApolloClient.builder() will be used and most importantly,shared between different modules.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/response/ScalarTypeAdapters.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/response/ScalarTypeAdapters.java
@@ -3,6 +3,7 @@ package com.apollographql.apollo.response;
 import com.apollographql.apollo.api.FileUpload;
 import com.apollographql.apollo.api.ScalarType;
 
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


sav007:
Sorry, but I'm not quite sure how this PR addresses the linked issue by changing the internal map of adapters from `Map<ScalarType, CustomTypeAdapter>` to `Map<String, CustomTypeAdapter>`?

budius:
@sav007 let me explain in an example (which is exactly what I've done on my project and running a local build of the apollo client)

- module/aar `network` creates the `ApolloClient` and register a `CustomTypeAdapter` for the type `CustomType.ISO8601DATETIME` and transforms the ISO8601 String into an `epochInMs : Long`
- Internally this adapter gets registered as `customAdapters.put("ISO8601DateTime" = adapterInstance`; With a string, not with `com.mycompany.network.apollo.CustomType` class
- `network` exposes on the public api `fun <D : Operation.Data, T, V : Operation.Variables> query(query: Query<D, T, V>): LiveData<Response<T>>`, allowing any consumer of this library to do any Apollo call.
- module `app-UI` uses the module/aar `network`
- module `app-UI` register apollo-plugin and passes the same type mapping the `network` module did:
```
apollo {
	customTypeMapping = [
			"ISO8601DateTime": "java.lang.Long"
	]
}
```
- the apollo-plugin on the `app-UI` generates queries specific to the UI from the same schema, and all `ISO8601DateTime` will be created as `Long`.
- Now the `app-UI` on runtime will be executing queries "using" the custom type `com.mycompany.myapp.apollo.CustomType` (a different class from the one on the `network`, but, as the mapping uses the `String typeName()` and not the Class, it will effectively be sharing the type adapter registered on the `network`.

---------------

Ideally, the apollo-plugin would generate a `META-INF/apollo.mappings` on the `network` module, so the plugin on the `app-UI` would read the mappings from it. Removing the need to declare again on the gradle file.
But I believe this merge request is a good start and that it solves the problem.

PR #1595, opened 2019-09-17
Supress "UNUSED_ANONYMOUS_PARAMETER" to prevent warnings in generated Kotlin code

jbttn:
We treat warnings as errors, so this warning prevents our project from compiling successfully.  This change suppresses the warning in the generated code.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
@@ -17,8 +17,8 @@ internal object KotlinCodeGen {
 
   val suppressWarningsAnnotation = AnnotationSpec
       .builder(Suppress::class)
-      .addMember("%S, %S, %S, %S", "NAME_SHADOWING", "LocalVariableName", "RemoveExplicitTypeArguments",
-          "NestedLambdaShadowedImplicitParameter")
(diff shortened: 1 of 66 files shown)


jbttn:
For what it's worth it looks like this warning popped up between 1.0.1 and 1.0.2.  Here are some snippets from one of our generated files:

1.0.1:
```
      operator fun invoke(reader: ResponseReader): Student {
        val __typename = reader.readString(RESPONSE_FIELDS[0])
        val fragments = reader.readConditional(RESPONSE_FIELDS[1]) { conditionalType, reader ->
          val studentData = if (StudentData.POSSIBLE_TYPES.contains(conditionalType))
              StudentData(reader) else null
          Fragments(
            studentData = studentData!!
          )
        }

        return Student(
          __typename = __typename,
          fragments = fragments
        )
      }
```

1.0.2:
```
      operator fun invoke(reader: ResponseReader): Student {
        val __typename = reader.readString(RESPONSE_FIELDS[0])
        val fragments = reader.readConditional(RESPONSE_FIELDS[1]) { conditionalType, reader ->
          val studentData = StudentData(reader)
          Fragments(
            studentData = studentData
          )
        }

        return Student(
          __typename = __typename,
          fragments = fragments
        )
      }
```

The problem is the `conditionalType` parameter.  It looks like it was used in `1.0.1`, but is no longer used in `1.0.2` and onward.  I'm not sure if it would make more sense to remove the unused variable or just take this suppression change.

sav007:
Thx for the PR, I guess we can merge it as a temp fix for now. Proper fix is coming.

PR #1599, opened 2019-09-19
Add option for outputting transformed queries

### LINKED ISSUE #1597: Would like a way to manually get transformed query document
I'm dealing with an API that has persisted operations, but doesn't support Apollo's APQ. I have to manually upload a manifest of the queries. If I do this using the human-friendly .graphql queries, they're missing the `__typename`s needed, and the operation hashes don't match.
I'd like a way to get the `__typename`'d query documents at build time. They're present in the generated query classes.

DSteve595:
Currently, if I want access to the "transformed" queries that have `__typename` and bundled fragments, I need to get it from the generated query classes' bodies. This adds an option to output them to a specified directory.
For my use case, this will allow manual creation of persisted operations that require the real body of the queries being made.
Fixes #1597.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Apollo-Android is designed primarily with Android in mind but you can use it in 
   - [Java Beans Semantic Naming for Accessors](#java-beans-semantic-naming-for-accessors)
   - [Explicit Schema location](#explicit-schema-location)
   - [Kotlin model generation (experimental)](#kotlin-model-generation-experimental)
+  - [Transformed queries](#transformed-queries)
 - [License](#license)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update --> 
@@ -668,6 +669,17 @@ By default Apollo Gradle plugin generates Java models but you can configure it t
 ```groovy
 apollo {
   generateKotlinModels = true
+}
+```
+
+### Transformed queries
+When Apollo-Android executes your queries, the actual queries sent to the server differs slightly from what was given; specifically, type hints are added to variable-type fields, and fragments are bundled. These differences don't affect typical use. If you want access to these transformed queries, Apollo Gradle plugin can save them to a build directory for you. This can be useful if you need to upload a query's exact content to a server that doesn't support automatic persisted queries.
+
+#### Usage
+
+```groovy
+apollo {
+  generateTransformedQueries = true
 }
 ```
 
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -52,6 +52,9 @@ class GraphQLCompiler {
     }
 
     args.transformedQueriesOutputDir?.let { transformedQueriesOutputDir ->
+      if (transformedQueriesOutputDir.exists()) {
+        transformedQueriesOutputDir.deleteRecursively()
+      }
       val transformedQueryOutput = TransformedQueryOutput(
           packageNameProvider
       )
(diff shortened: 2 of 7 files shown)


DSteve595:
Out of laziness, does anyone mind adding this to the readme? I don't feel like installing NPM for doctoc 😛



```markdown
### Transformed queries
When Apollo-Android executes your queries, the actual queries sent to the server differs slightly from what was given; specifically, type hints are added to variable-type fields, and fragments are bundled. These differences don't affect typical use. If you want these transformed queries to be saved to a directory, you can provide a path to one. This can be useful if you need to upload a query's exact content to a server that doesn't support automatic persisted queries.

#### Usage

```groovy
apollo {
  transformedQueriesOutputDir = "$projectDir/myTransformedQueries"
}
```
```

**Comment by martinbonnin** [comment]
TIL about doctoc :). I've been updating the README "by hand" until now. I didn't realize there was a tool for that. What happens if you just edit the table of contents manually ? Does it break future usages of doctoc ? 

In all cases, I'll port that to the incubating plugin: https://github.com/apollographql/apollo-android/pull/1577. We certainly want `transformedQueriesOutputDir` to be per service (or schema, whatever we end up naming that thing).

Or should we instead hardcode the outputDir ? Something like `build/generated/apollo/{service}/{variant}/transformedQueries` ? And just expose a flag to the users:

```
apollo  {
    generateTransformedQueries = true|false
}```

That leaves less opportunity for users to mess things up.

DSteve595:
Thanks for the quick review @martinbonnin!
>I've been updating the README "by hand" until now

I'll do the same then. Doesn't look bad at all.

I totally missed multiple-schema setups. Hardcoding the location is probably a good idea, as long as we can publicly expose that directory. Would that be done through the extension? How are variant-specific properties usually accessed? I know with AGP, it's typically with `(application|library)Variants.all { ... }`, but how would that work here?

Also, do you think it makes sense to split this into a separate Gradle task? It's currently piggybacking off of `generateClasses`.

martinbonnin:
> I totally missed multiple-schema setups. Hardcoding the location is probably a good idea, as long as we can publicly expose that directory. Would that be done through the extension? How are variant-specific properties usually accessed? I know with AGP, it's typically with android.(application|library)Variants.all { ... }, but how would that work here?

What do you mean by "publicly expose that directory"? I was thinking users could just go to `build/output/apollo/transformedQueries/{service}/{variant}` and get the transformed queries from there ? Then you can do whatever is needed with the transformedQueries. Or do you need this to be an input to some other gradle tasks in maybe other plugins?

> Also, do you think it makes sense to split this into a separate Gradle task? It's currently piggybacking off of generateClasses.

We should certainly measure it but I'm hoping writing the transformed queries is fast enough that we don't need a separate gradle task. Maybe as an optimisation in a second step ?

DSteve595:
>publicly expose that directory

Having a `File` that always points to where these outputs go. It'd be good for using it with other tasks, like you mentioned. If and when the folder changes in plugin updates, it'll be easier to migrate if the path isn't hardcoded from the user's perspective.

>I'm hoping writing the transformed queries is fast enough

It probably is. Just writes an already-built string to a file. Will keep.

martinbonnin:
> Having a File that always points to where these outputs go

What I'm thinking of right now is putting this File in `ServiceVariant` and expose the list of `ServiceVariant` list in the `ApolloExtension` as a `NamedDomainObjectContainer` (or some other gradle objects that allows to listen for changes, can't remember exactly). Other plugins could then "react" on the apollo plugin being applied and watch the list:

```
apolloExtension.serviceVariants.all {
    // do something with it.transformedQueriesDirectory
}
```

DSteve595:
That sounds perfect. I'm not sure how to do that, but I'll try to figure it out. Any resources you can point to would be appreciated.

martinbonnin:
I had mainly the incubating plugin in mind (which is not merged yet). I would do something like this: https://github.com/martinbonnin/apollo-android/commit/7ab78cb855bca8c1b95ff722ecec8e3a85702a97. 

I took DefaultDomainObjectSet from the gradle plugin but it looks like it's internal gradle API I'm not sure how safe it is to use it, maybe there's a better candidate for "reactive" lists of obejcts.

DSteve595:
Is `service` a concept in the current Gradle plugin? Not sure how I'd get it (or an analogue) from `TaskConfigurator::createCodegenTask`

martinbonnin:
I have no idea how to do this with the current plugin :-/
Feel free to merge as is and we'll add multi schema etc later.

DSteve595:
Thanks for taking the time to demo it in the incubating plugin.
I left a TODO for it. The directories are still determined by package, so maybe that'll be enough?

PR #1630, opened 2019-09-26
Incubating plugin improvements

martinbonnin:
A few fixes/improvements:
* Use `service0, service1, ...` instead of plain `0, 1` as service names as it gives more hints about what the name is about.
* Make the autocomplete work in the IDE for kotlin models
* Expose CompilationUnit for other plugins/build logic to intercept outputs/transformedQueries
* Fix modifying schema.json did not rebuild the sources
* Put sources under `generated/source/` (same as kapt, BuildConfig)

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt
@@ -49,12 +49,11 @@ object AndroidTaskConfigurator {
     )
 
     registerVariantTask(project, apolloVariant) { serviceVariantTask ->
+      variant.registerJavaGeneratingTask(serviceVariantTask.get(), serviceVariantTask.get().outputDir)
       if (apolloExtension.generateKotlinModels) {
+        // For java, this is done by registerJavaGeneratingTask
         androidExtension.sourceSets.first { it.name == variant.name }.kotlin!!.srcDir(serviceVariantTask.get().outputDir)
         project.tasks.named("compile${variant.name.capitalize()}Kotlin").configure { it.dependsOn(serviceVariantTask) }
-      } else {
-        // apparently, this does everything automagically
-        variant.registerJavaGeneratingTask(serviceVariantTask.get(), serviceVariantTask.get().outputDir)
       }
     }
   }
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt
@@ -51,7 +51,7 @@ object AndroidTaskConfigurator {
     registerVariantTask(project, apolloVariant) { serviceVariantTask ->
       variant.registerJavaGeneratingTask(serviceVariantTask.get(), serviceVariantTask.get().outputDir)
       if (apolloExtension.generateKotlinModels) {
-        // For java, this is done by registerJavaGeneratingTask
+        // For java models, this is done by registerJavaGeneratingTask
         androidExtension.sourceSets.first { it.name == variant.name }.kotlin!!.srcDir(serviceVariantTask.get().outputDir)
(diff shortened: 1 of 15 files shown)


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt
```diff
@@ -49,12 +49,11 @@ object AndroidTaskConfigurator {
     )
 
     registerVariantTask(project, apolloVariant) { serviceVariantTask ->
+      variant.registerJavaGeneratingTask(serviceVariantTask.get(), serviceVariantTask.get().outputDir)
       if (apolloExtension.generateKotlinModels) {
+        // For java, this is done by registerJavaGeneratingTask
```
DSteve595: Does this mean 'java as opposed to android'?

martinbonnin: For android "java models". This file is android-only.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/ApolloCodegenTask.kt
```diff
@@ -58,6 +55,9 @@ open class ApolloCodegenTask : SourceTask() {
     return super.getSource()
   }
 
+  @Internal
+  var schemaFile: File? = null
```
sav007: sanity check, I assume that `schemaFile`  might come from the configuration. If it's marked as internal will this task be NOT UP-TO-DATE if user changes in plugin configuration location to schema?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/CompilationUnit.kt
```diff
@@ -0,0 +1,148 @@
+package com.apollographql.apollo.gradle
```
sav007: nit: seems like the indents in this file is 4 spaces instead of 2?

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
```diff
@@ -63,9 +68,42 @@ class UpToDateTests {
...
+
+      result = TestUtils.executeTask("generateApolloClasses", dir, "-i")
+
+      assertEquals(TaskOutcome.SUCCESS, result.task(":generateApolloClasses")!!.outcome)
+      assertThat(dir.generatedChild("main/service0/com/example/DroidDetailsQuery.java").readText(), not(containsString("classification")))
+    }
+  }
+
+  @Test
+  fun `change schema file rebuilds the sources`() {
```
sav007: neat!

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/ApolloCodegenTask.kt
martinbonnin: The magic comes from `apolloCodeGenTask.source(schemaFile)` there: https://github.com/martinbonnin/apollo-android/blob/7a74704161a8c21f4e6acd261a9aa937118ecede/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/ApolloPlugin.kt#L191.

Since the schemaFile is added to the source files, it's taken into account for up-to-date checks. 

Something I wonder is how Gradle treats `Input` properties. Would it just check the path ? Or also modification date, possibly content ? I'll look that up but for now I assumed that having the schema as source does the job. And the test passes so it should be all good.

Moving forward, I'm not sure we need ApolloCodeGenTask to inherit from `SourceTask`. Maybe it can just be a `DefaultTask` if we declare inputs correctly.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/CompilationUnit.kt
martinbonnin: Oh my, sorry about that. Could we merge that PR: https://github.com/apollographql/apollo-android/pull/1326 ?

Issue #1607, opened 2019-09-20
Support for replacing query with operation ID


tmtrademarked:
This is a slightly specific feature request, but I'm curious if other users have a need for something like this as well. In our tech stack, we use a Rails server running the GraphQL Pro ruby gem (https://graphql.pro), which implements persisted queries slightly differently than Apollo's Auto Persisted Queries feature.

We're trying to figure out if there's a good way to use the rest of the Apollo client side stack here, but basically just swap out the persisted query on our side. Technically, this is pretty easy if you can control where the server request gets written - you just need to *not* write the "query" into the JSON, and instead provide "operationID" with a value.

So I think the more generalizeable form of this feature request is something like "Allow us to plug into ApolloServerInterceptor and replace the query body with a key/value of the application's choosing." On our end, we can easily map from the operation name to the appropriate value to send - there's just no obvious way to hook into the server interceptor that I can see, since it's the last interceptor in the stack.

What does this community think? I don't think such a feature would be particularly difficult to implement, and it would definitely enhance our ability to use this stack. I'm happy to propose some sort of interface/design here for feedback if there's interest, but don't want to waste anyone's time barking up a tree that's either not worthwhile, already been solved, or otherwise not valued here.

tmtrademarked:
As I dig a little further, this may already be supported today via an ApplicationInterceptor. It seems like the combination of an interceptor that sets the `sendQueryDocument` field and a custom Operation wrapper could possibly provide the mechanism I need here. I'll investigate this further.

DSteve595:
Similar scenario here. I think it might be possible to do this with a custom re-implementation of `ApolloNetworkInterceptor`, but that might break the chain and potentially cause other problems.

DSteve595:
Did some deeper digging. I think the best way to achieve this is to use a custom client, avoiding `ApolloClient` entirely. This seems counter-intuitive, but apollo-runtime's public API exposes almost everything needed to run the queries ourselves.

Here's an example of manually running a query with OkHttp:
```kotlin
val scalarTypeAdapters = ScalarTypeAdapters(emptyMap())

fun <D : Operation.Data, T, V : Operation.Variables, O : Operation<D, T, V>> execute(
  operation: O,
  usePersistedId: Boolean = true
): Response<T> {
  return Request.Builder()
    .url("https://gql.yourapi.example/")
    .post(createApolloRequestBody(operation, usePersistedId))
    .build()
    .let(okHttpClient::newCall)
    .execute()
[...]

PR #1631, opened 2019-09-26
Rework transformed query output to include separate fragments

DSteve595:
I made a mistake thinking `sourceWithFragments` was the `__typename`-ified version of the document, while `source` wasn't.
`source` actually _does_ have the `__typename`s added.
Including the fragments in the transformed queries actually opens up potential problems for the workflow I described in #1597 (and likely the workflow in #1607), because APIs that required persisted operations but don't support APQ generally still follow the `queries/`+`mutations/`+`fragments/` pattern, and expect fragments to still be separated as they are in the developer-facing source.
This changes TransformedQueryOutput to save the transformed operations and transformed fragments separately, as they exist in the source. Fragments that are defined in operations (e.g. `query MyQuery { ...MyFragment } fragment MyFragment { }` are appended to the operation's file, resulting in no changes other than the addition of `__typename`s.
Also, no longer using `packageNameProvider` to get the destinations for these files, since those are more intended for Java/Kotlin source.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/TransformedQueryOutput.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/TransformedQueryOutput.kt
@@ -1,36 +1,44 @@
 package com.apollographql.apollo.compiler
 
-import com.apollographql.apollo.compiler.ir.Operation
+import com.apollographql.apollo.compiler.ir.CodeGenerationIR
 import java.io.File
 
 internal class TransformedQueryOutput(
     private val packageNameProvider: PackageNameProvider
 ) {
-  private var transformedQueries: List<TransformedQuery> = emptyList()
+  private var transformedDocuments: List<TransformedDocument> = emptyList()
 
-  fun visit(operations: List<Operation>) {
-    transformedQueries = transformedQueries + operations.map { operation ->
+  fun visit(ir: CodeGenerationIR) {
+    transformedDocuments = transformedDocuments + ir.operations.map { operation ->
       val targetPackage = packageNameProvider.operationPackageName(operation.filePath)
-      TransformedQuery(
-          queryName = operation.operationName,
-          queryDocument = operation.source,
+      TransformedDocument(
+          name = operation.operationName,
+          document = operation.source,
+          packageName = targetPackage
+      )
+    }
+    transformedDocuments = transformedDocuments + ir.fragments.map { fragment ->
+      val targetPackage = packageNameProvider.fragmentsPackageName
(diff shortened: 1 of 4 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/TransformedQueryOutput.kt
```diff
@@ -1,36 +1,45 @@
...
+    transformedDocuments = transformedDocuments + ir.operations.map { operation ->
+      TransformedDocument(
+          document = operation.source,
+          filePath = operation.filePath.relativePathToGraphql()!!
+      )
+    }
+    transformedDocuments = transformedDocuments + ir.fragments.map { fragment ->
+      TransformedDocument(
+          document = fragment.source,
+          filePath = fragment.filePath!!.relativePathToGraphql()!!
```
DSteve595: When would `fragment.filePath!!` be null? I can't find any repro cases

DSteve595:
^ force-push was a no-op, checking to see if that build failure is reproducible

DSteve595:
hey, it was!

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/TransformedQueryOutput.kt
sav007: I guess it left from old code gen, feel free to make it non-optional

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/TransformedQueryOutput.kt
```diff
@@ -1,36 +1,45 @@
...
-        resolve("${transformedQuery.queryName}.graphql")
-      }.writeText(transformedQuery.queryDocument)
-    }
+    transformedDocuments
+      .groupBy { it.filePath } // Multiple documents can be from the same file, e.g. fragments
+      .forEach { (filePath, transformedDocuments) ->
+        val outputFile = outputDir.resolve(filePath).also {
+          it.parentFile.mkdirs()
+        }
+        transformedDocuments.forEachIndexed { i, transformedDocument ->
```
sav007: nit: can be shortened 
```
transformedDocuments
  .joinToString(separator = "\n\n") { it.document }
  .also { outputFile.writeText(it) }
```

DSteve595: There are multiple different filePaths, that'll bunch them all together

PR #1638, opened 2019-09-29
Change output directories into providers

DSteve595:
The output directories for generated classes and transformed queries are currently `File`s in both `CompilationUnit` and `ApolloCodegenTask`. To better enable [lazy configuration](https://docs.gradle.org/current/userguide/lazy_configuration.html), they should instead be `Provider<File>`s. This PR does that using `DirectoryProperty`s.
`CompilationUnit` is currently used both as a configuration for `ApolloCodegenTask` _and_ as a user-facing way to get properties relating to Apollo, and this two-way dependency makes the setup a bit more complicated. Currently the directories are set within the `CompilationUnit` and then applied to the codegen task, but in order for the Providers to carry the dependencies of the task's outputs, the `DirectoryProperty`s in the `CompilationUnit` need to be set _from_ the task's `@OutputDirectory`-annotated properties. `setupOutputProperties` is the solution I came up for this. If you think of something cleaner, please suggest it.

This enables users to configure their tasks such that relying on Apollo's output directories automatically sets up their dependencies without needing to know about the exact tasks Apollo is running to output them. An example configuration is added as a test.

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/DefaultCompilationUnit.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/DefaultCompilationUnit.kt
@@ -1,60 +1,32 @@
 package com.apollographql.apollo.gradle
 
 import com.apollographql.apollo.compiler.*
+import com.apollographql.apollo.gradle.api.CompilationUnit
 import org.gradle.api.Project
-import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.file.Directory
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.util.PatternSet
 import java.io.File
 
-class CompilationUnit(
-    val serviceName: String,
-    val variantName: String,
+class DefaultCompilationUnit(
+    override val serviceName: String,
+    override val variantName: String,
+    override val androidVariant: Any?,
     val files: List<File>,
     val schemaFile: File?,
     val schemaPackageName: String,
     val rootPackageName: String,
-    val androidVariant: Any?,
     project: Project
-) {
-  val name = "${variantName}${serviceName}"
-  val outputDir: DirectoryProperty = project.objects.directoryProperty()
-  val transformedQueriesDir: DirectoryProperty = project.objects.directoryProperty()
+): CompilationUnit {
+  val name = "${variantName}${serviceName.capitalize()}"
+  val outputDirectory = project.buildDir.child("generated", "source", "apollo", "classes", variantName, serviceName)
+  val transformedQueriesDirectory = project.buildDir.child("generated", "transformedQueries", "apollo", variantName, serviceName)
 
-  /**
-   * Sets the output properties for [codegenTask] and this CompilationUnit such that [Provider]s
-   * accessed via this CompilationUnit carry dependencies on [codegenTask]'s outputs.
-   *
-   * @see org.gradle.api.provider.Property.set
-   */
-  internal fun setupOutputProperties(
-      codegenTask: ApolloCodegenTask,
-      willGenerateTransformedQueries: Boolean,
-      project: Project
-  ) {
-    codegenTask.outputDir.apply {
-      set(project.buildDir.child("generated", "source", "apollo", "classes", variantName, serviceName))
-      finalizeValue()
-    }
-    outputDir.apply {
-      set(codegenTask.outputDir)
-      disallowChanges()
-    }
-
-    codegenTask.transformedQueriesOutputDir.apply {
-      if (willGenerateTransformedQueries) {
-        set(project.buildDir.child("generated", "transformedQueries", "apollo", variantName, serviceName))
-      }
-      finalizeValue()
-    }
-    transformedQueriesDir.apply {
-      set(codegenTask.transformedQueriesOutputDir)
-      disallowChanges()
-    }
-  }
+  override lateinit var outputDir: Provider<Directory>
+  override lateinit var transformedQueriesDir: Provider<Directory>
 
   companion object {
-    fun from(project: Project, apolloVariant: ApolloVariant, service: Service): CompilationUnit {
+    fun from(project: Project, apolloVariant: ApolloVariant, service: Service): DefaultCompilationUnit {
       val sourceSetNames = apolloVariant.sourceSetNames
 
       val schemaFilePath = service.schemaFilePath
@@ -92,7 +64,7 @@ class CompilationUnit(
 
       val schemaPackageName = schemaFile.canonicalPath.formatPackageName(dropLast = 1) ?: ""
 
-      return CompilationUnit(
+      return DefaultCompilationUnit(
           serviceName = service.name,
           variantName = apolloVariant.name,
           files = files,
@@ -104,7 +76,7 @@ class CompilationUnit(
       )
     }
 
-    fun default(project: Project, apolloVariant: ApolloVariant): List<CompilationUnit> {
+    fun default(project: Project, apolloVariant: ApolloVariant): List<DefaultCompilationUnit> {
       val sourceSetNames = apolloVariant.sourceSetNames
       val schemaFiles = findFilesInSourceSets(project, sourceSetNames, ".") {
         it.name == "schema.json"
@@ -123,7 +95,7 @@ class CompilationUnit(
 
             val name = "service${i++}"
 
-            CompilationUnit(
+            DefaultCompilationUnit(
                 serviceName = name,
                 variantName = apolloVariant.name,
                 files = files,
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/DefaultCompilationUnit.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/DefaultCompilationUnit.kt
@@ -19,8 +19,8 @@ class DefaultCompilationUnit(
     project: Project
 ): CompilationUnit {
   val name = "${variantName}${serviceName.capitalize()}"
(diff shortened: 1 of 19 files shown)


apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
```diff
@@ -264,4 +264,24 @@ class ConfigurationTests {
...
\ No newline at end of file
+
+  @Test
+  fun `compilation unit directory properties carry task dependencies`() {
+    withSimpleProject("""
+      apollo {
+        generateTransformedQueries = true
+        
+        compilationUnits.all { compilationUnit ->
+          def customTaskProvider = tasks.register("customTask" + compilationUnit.name) {
```
DSteve595: Side note: `name` is a bit clunky to use since it's comprised of '`variant name` + `service name`', but neither are capitalized

cc @martinbonnin

martinbonnin: `name` is really only used to make the `NamedDomainObjectContainer` happy but it's not really used. If there's another way to do `compilationUnits.all {}` without having to give CompilationUnit a `name`, I'd prefer this. I only went for `NamedDomainObjectContainer` bevause it's ubiquitous in the gradle code I've seen so far but it might be overkill.

martinbonnin:
@DSteve595 Regarding `CompilationUnit.setupOutputProperties`, what about simply adding `ApolloCodegenTask.androidVariant: Any?` ? Then instead of writing:

```groovy
      apollo {
        generateTransformedQueries = true        

        compilationUnits.all { compilationUnit ->
          tasks.register("customTask" + compilationUnit.name) {
            inputs.dir(compilationUnit.outputDir)
            inputs.dir(compilationUnit.transformedQueriesDir)
          }
        }
      }
```

we could just use the actual task.

```groovy
  tasks.withType<ApolloCodegenTask>().all { codeGenTask ->
    tasks.register("customTask" + codeGenTask.name.capitalize()) {
      inputs.dir(codeGenTask.outputDir)
      inputs.dir(codeGenTask.transformedQueriesDir)
    }
  }
```
Since `outputDir` and `transformedQueriesDire` are of type `Property`, I think that would work. You could also expose `serviceName` and `variantName` in addition to `androidVariant` to be able to customize the task more.

`CompilationUnit` would just become some internal class not exposed to users and outside the task graph. We could even move it to `com.apollographql.apollo.gradle.internal` to discourage anyone to leak that.

DSteve595:
That would be simpler. However, it then couples the user-facing way of getting that info with the task itself, which I think would be helpful to avoid. I did it this way initially because I thought CompilationUnit was intended to be user-facing (i.e. your `apollo.compilationUnits.all { }` example), so that might've been a wrong assumption.
I don't feel strongly about it (and it's not up to me anyway, lol), but it seems like not exposing the details of ApolloCodegenTask will help make it so the library can make changes without impacting users. For example, a while ago we talked about a potential future optimization where outputting transformed queries is done in a separate task from class generation. If/when that ever got implemented, the notion of accessing things through `ApolloCodegenTask` might not make sense anymore, particularly `transformedQueriesDir` (which the codegen task would then know nothing about). If we instead just exposed the configuration to the user and let them set up their build script that way, the underlying tasks could change as freely as needed because they're just implementation details.

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
DSteve595: Gotcha, forgot the container requires a name (an override property would this more obvious, thanks Groovy)

DSteve595:
Also, `withType` eagerly configures tasks.

martinbonnin:
`withType(Class, Action)` eagerly configures tasks but I "think" `withType(Class).configureEach(Action)` is OK (see the [table at the end of the doc](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview))

What bugs me a bit with the current approach is that `setupOutputProperties` is called because the `ApolloCodegenTask` is configured eagerly [there](https://github.com/DSteve595/apollo-android/blob/6e3b895a5bc1bbe0f1a25eb2d2b42418a57742f9/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/AndroidTaskConfigurator.kt#L54). If it weren't for that, I believe `compilationUnit.outputDir` would just be empty and creating customTasks based on that would fail. 

Maybe another approach would be to actually create a `generateTransformedQueries` task that for now just depends on `ApolloCodegenTask` but could be different if we decide to change the implementation ? 

My current mental model of a gradle plugin is that: 
* The extension is the plugin configuration (== input of the plugin)
* The tasks are what is exposed to the outside world (gradle-core and users)(== output of the plugin)

That's why I think it's not that bad (maybe preferable?) to put this in a task. I'm still learning though...

DSteve595:
>withType(Class).configureEach(Action) is OK

Oops, you're totally right.

>What bugs me a bit with the current approach is that setupOutputProperties is called because the ApolloCodegenTask is configured eagerly there. If it weren't for that, I believe compilationUnit.outputDir would just be empty and creating customTasks based on that would fail.

Good point. This is actually simple to fix; change the arg into `TaskProvider` and use `flatMap { it.outputDir }` rather than just `outputDir`.

I wonder if there's any harm in having the extension be more than just a holder for properties. It does seem like Gradle is moving in the direction of plugins exposing properties rather than tasks, but I'm also just learning, so don't take that as gospel.

martinbonnin:
> change the arg into TaskProvider and use flatMap { it.outputDir } rather than just `outputDir`

Do you mean:

```
  internal fun setupOutputProperties(
      taskProvider: TaskProvider<ApolloCodegenTask>,
      willGenerateTransformedQueries: Boolean,
      project: Project
  ) {
    ...
    outputDir.set(taskProvider.flatMap { it.outputDir })
    ...
```

Is that going to change anything ? If we remove the `ApolloCodegenTask` eager configuration, `setupOutputProperties` will not be called until the tasks are resolved and `CompilationUnit.outputDir` will have no way to know it depends on `ApolloCodegenTask` until `ApolloCodegenTask` is configured. Calling `customTaskmainservice0` will read an empty input property as it won't depend on `generateMainService0ApolloClasses` ? Or will it ?

> It does seem like Gradle is moving in the direction of plugins exposing properties rather than tasks

Indeed. Works for me.

martinbonnin:
Proposition to break a bit the circular dependency there: https://github.com/DSteve595/apollo-android/pull/1

CompilationUnit will be used to configure the `ApolloGenTask`. It's also augmented with `outputDir: Provider<Directory>` and `transformedQueries: Provider<Directory>` that are taken from the taskProvider.

We might want to split the `CompilationUnit` into an "internal" part and an "api/public" part but that can be done later.

DSteve595:
>Calling customTaskmainservice0 will read an empty input property as it won't depend on generateMainService0ApolloClasses ? Or will it ?

The `flatMap` should mean it does. I can verify, but I'm pretty sure that's how the API is intended to work.

martinbonnin:
> The flatMap should mean it does. 

But that `flatMap`will only be called from the configuration of ApolloCodegen, right ? 

And since ApolloCodegen is not a dependency yet, it will not be configured when executing `customTaskmainservice0`

DSteve595:
I don't think so. Since the `flatMap` call is on the `TaskProvider`, Gradle is able to understand that this dependency is something that requires the task to be configured. So if the custom task is configured, Gradle will automatically force the ApolloCodegenTask to be configured, which is what we want.

DSteve595:
Confirmed. Try this in a test project:
```groovy
abstract class CodegenTask extends DefaultTask {
    public Property<String> thing = project.objects.property(String)
}

abstract class MyTask extends DefaultTask {
}

def codegenTaskProvider = tasks.register('codegenTask', CodegenTask) {
    thing.set("hello")
    println "Codegen task is being configured!"
}

tasks.register('myTask', MyTask) { myTask ->
    myTask.inputs.property('thing', codegenTaskProvider.flatMap { it.thing })
    doLast {
        println "Property is: ${myTask.inputs.properties['thing']}"
    }
}
```
`./gradlew build` does not configure `codegenTask`.
`./gradlew myTask`: `codegenTask` is configured, and property is set.

martinbonnin:
My point was that `setupOutputProperties` was called from `configure(ApolloCodegenTask)` and not from `configure(CustomTask)` but I think we're saying the same thing.

DSteve595:
Oh, I see now. Yeah, that probably would've been a problem. Good catch.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/DefaultCompilationUnit.kt
```diff
@@ -19,8 +19,8 @@ class DefaultCompilationUnit(
     project: Project
 ): CompilationUnit {
   val name = "${variantName}${serviceName.capitalize()}"
-  val outputDirectory = project.buildDir.child("generated", "source", "apollo", "classes", variantName, serviceName)
-  val transformedQueriesDirectory = project.buildDir.child("generated", "transformedQueries", "apollo", variantName, serviceName)
+  internal val outputDirectory = project.buildDir.child("generated", "source", "apollo", "classes", variantName, serviceName)
+  internal val transformedQueriesDirectory = project.buildDir.child("generated", "transformedQueries", "apollo", variantName, serviceName)
```
DSteve595: @martinbonnin This makes sense, right?

martinbonnin: Yep, definitely :+1:

PR #1646, opened 2019-10-03
Compile samples in CI

martinbonnin:
Add the samples to Travis CI. Since they are now a composite build, they were not compiled from the root project automatically.

@tasomaniac I guess the gradle wrapper under `samples/gradlew` is needed to open the `samples` directory in IDEA ? Is there any chance we could merge it with the gradle wraper at the root of `apollo-android` ? Not only would that save some space but it would also ensure the gradle version stay in sync (right now we have gradle 5.6 in root vs 5.5.1 in samples).

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
   </state>
 </component>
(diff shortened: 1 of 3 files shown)


tasomaniac:
This option is really the best when using Gradle plugins in samples. Otherwise we cannot use the plugin in the source code. In the past it was using "some" published version and that was also really old 😔

The solution to not synced Gradle versions is to have Gradle version also defined in the script where we define versions and dependencies. And then use that in both root and sample projects.

tasomaniac:
Btw, we could also just remove the wrapper from sample.

martinbonnin:
Closing in favor of https://github.com/apollographql/apollo-android/pull/1649

(Re: the wrapper, I used symlinks in that other PR so intelliJ diesn't ask and we still keep gradle versions in sync)

Issue #1034, opened 2018-08-29
Request is sent with unnecessary whitespace


Stonepaw:
I've been experimenting with this library and noticed that the outgoing requests are sent in the POST body with unnecessary white space characters and line breaks. 

For example the query:
```
mutation CreateSession { 
  createSession {
    token
    content
  }
}
```
Is sent as a POST request with a body of:
<code>{"query":"mutation CreateSession { &nbsp;createSession { &nbsp;&nbsp;&nbsp;__typename &nbsp;&nbsp; token &nbsp;&nbsp; content &nbsp;}}","operationName":"CreateSession","variables":{}}</code>

Is there an option to normalize this behavior and strip the excessive white space and line break characters?

ralphie9224:
@sav007 Picking this up if it's available

PR #1650, opened 2019-10-03
Fix issue with request query document having unnecessary white space

ralphie9224:
This addresses the "Request is sent with unnecessary whitespace #1034 " issue. The actual change was to the regex in the below file. Everything else was test updates.

https://github.com/apollographql/apollo-android/compare/master...ralphie9224:rafa/multi-line-request-bug?expand=1#diff-f69f381fa2b8bcb44194e8a9a5b19b56L206


ralphie9224:
@martinbonnin 
1. Yes that is correct. This leaves out a few extra whitespaces.

2. Kind of answering the last part of 1 as well. The reasoning behind it is that when the code is generated, the QUERY_DOCUMENT string is nicely formatted and human readable. If we formatted the code to be on one line, it's a bit harder to parse. I actually considered suggesting that instead of the code that's printed in the interceptor being one line, to actually have it displayed in the same way that it's displaying in the generated code, but figured we were optimizing for space on the logcat.

3. I'm assuming you're referring to the unit tests and not the Query Test classes. I updated them manually. Not sure how it could be done automatically, but it could be good to look into it

Let me know if any of this is unclear. 

As an aside. This is my first PR for this project. I guess one of y'all usually merges them?

martinbonnin:
All clear for me, thanks ! @sav007 can you look into it? 

About 3. I was thinking maybe a gradle task: `./gradlew :project:updateTestFixtures`. That would save time when changing these kind of things. I'll look into it.

sav007:
There is an issue with these changes. Generated `OPERATION_ID` (this field calculated during model generation as sha256 hash from query document) won't match with the hash generated on the server side, means query auto persistence feature will be broken.

sav007:
I guess easier fix will be in the `ApolloServerInterceptor` instead of:
```
.name("sha256Hash").value(operation.operationId())
```

caclulate sha hash from the actual minified query document.

ralphie9224:
@sav007 Interesting, why is this the case? The `operation.operationId()` we get locally should be the same as the one that gets generated on the server side, regardless of how we represent the entire query document shouldn't it?

martinbonnin:
@sav007 the interceptor will be too late. I'm using operationId from the generated sources for debug purposes. 

Also, we now generate `transformedQueries` as output. I guess these must match what is in sent over the network and if we transform in the interceptor, it might not match ?

sav007:
@ralphie9224 the value for `operation.operationId()` is generated during code gen, that will take original query (with spaces). But to the server we will send different query document (without spaces) that will lead to different hash.

@martinbonnin well if you use `operationId` as just plain, transparent id than it's fine as they will be stable. If you use `operationId` value  as the hash value for query document, then it will be different from what we really going to send to the server. 
As for `transformedQueries ` that definitely an issue. 

Potentially we can do this on codegen level (strip spaces) but then the generated `QUERY_DOCUMENT` won;t be readable

ralphie9224:
@sav007 ah I understand what you're saying now. Yea Martin had brought up doing it as a part of the codegen above, which was something I thought about doing but for the same readability reasons didn't think it was the best bet. The other option would be to not strip the newline characters when we print. That way in the logcat we would show the full JSON block.

If I understand correctly, it sounds like that might be the best option over the original proposed solution of hashing the minified query document given the `transformQueries`?

sav007:
What if we introduce some static function `minify(queryDocument)` to `apollo-api` module that will be called before sending query document to the server and calculating hash, plus it will be called before dumping queries during model generation into disk (for `transformQueries` feature)?

martinbonnin:
@sav007 I use `query.OPERATION_ID` from an okhttp interceptor to match it to the HTTP header. Something like this:
```
        when (request.header("X-APOLLO-OPERATION-ID")) {
            GetSpaceQuery.OPERATION_ID -> {}
```

`query.OPERATION_ID` and `query.QUERY_DOCUMENT` are part of the public API now so we can't really remove them. I think it's just more consistent to minify `QUERY_DOCUMENT` in the generated code. I don't mind it too much. Unless there's a bug in apollo-android, this matches closely my `.graphql` files so I can look there. And it's not that hard to read anyway since the variables will keep their names (it's not like proguard). If really needed, we can add a comment with the whitespace and/or a plugin configuration to disable minification.

sav007:
Value from `query.OPERATION_ID` will always match the value in http header as it set from operation id:
```
.header(HEADER_APOLLO_OPERATION_ID, operation.operationId())
```

The only difference will be the actual request payload sent to the server and  its `sha256` in `persistedQuery.sha256Hash` field. 

Do you think that having formatted `query.QUERY_DOCUMENT` doesn't provide any benefit for the developer?

ralphie9224:
@sav007 @martinbonnin In the best case, I would rather have the query document formatted in the generated code _and_ the logs. It's very hard to parse in the logs

martinbonnin:
I just realized that the `'\n'` stripping is done in 2 places at the moment:

* in [ApolloServerInterceptor.httpPostRequestBody](https://github.com/apollographql/apollo-android/blob/dfd47c7676bd987eb579c04c00df506a5995ecec/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java#L229)
* In [TypeSpec.Builder.addOperationId](https://github.com/apollographql/apollo-android/blob/04c86857aecbd3e4f6c2d54dcb60088ba90a1fa8/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationTypeSpecBuilder.kt#L60)

I missed that before, I thought it was just done in the `ApolloServerInterceptor`. 

Now I see how the [minify proposal](https://github.com/apollographql/apollo-android/pull/1650#issuecomment-539809222) can solve this. Having a single method where this happens will definitely help. 

I would personally minimize as soon as possible (i.e. in generated code) as it enforces every place downstream uses the same minimized value. We could also work on tooling (intellij plugin) to prettify the values when needed. But I don't mind too much if `QUERY_DOCUMENT` is a bit verbose at the moment.

sav007:
Addressed via https://github.com/apollographql/apollo-android/pull/1705

PR #1654, opened 2019-10-04
move build scripts to kotlin

martinbonnin:
Migrate (most of) the build scripts to kotlin. I didn't do the composite projects from https://github.com/apollographql/apollo-android/pull/1649 to avoid potential conflicts.

This pull request is mostly a language translation with the behavior unchanged. This makes the files a bit more verbose (especially all the extra properties stuff). The plan is to refactor them progressively, potentially introducing a `buildSrc` module (but not 100% sure as the dependencies are also consumed by test projects). 

It should be easier to refactor with autocomplete and type safety.


sav007:
Conflicts :(

martinbonnin:
I just rebased and took this opportunity to move the composite build script to kotlin as well

Issue #1647, opened 2019-10-03
Main Thread remains blocked by the ApolloServerInterceptor


nicusorflorin:
Sometimes running a query using apollo results in the application getting an anr caused by the ApolloServerInterceptor 

```
"main@3653" prio=5 waiting for monitor entry
  java.lang.Thread.State: BLOCKED
	 waiting for OkHttp https://streaming-api-bar/...@4905 to release lock on <0x2229> (a com.apollographql.apollo.internal.interceptor.ApolloServerInterceptor)
	  at com.apollographql.apollo.internal.interceptor.ApolloServerInterceptor.dispose(ApolloServerInterceptor.java:90)
	  at com.apollographql.apollo.internal.interceptor.RealApolloInterceptorChain.dispose(RealApolloInterceptorChain.java:42)
	  at com.apollographql.apollo.internal.RealApolloCall.cancel(RealApolloCall.java:187)
	  - locked <0x2228> (a com.apollographql.apollo.internal.RealApolloCall)
	  at com.apollographql.apollo.rx2.Rx2Apollo$6.dispose(Rx2Apollo.java:232)
	  at io.reactivex.internal.disposables.DisposableHelper.dispose(DisposableHelper.java:124)
	  at io.reactivex.internal.operators.observable.ObservableCreate$CreateEmitter.dispose(ObservableCreate.java:121)
	  at io.reactivex.internal.disposables.DisposableHelper.dispose(DisposableHelper.java:124)
	  at io.reactivex.internal.operators.observable.ObservableSubscribeOn$SubscribeOnObserver.dispose(ObservableSubscribeOn.java:73)
	  at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.dispose(ObservableObserveOn.java:146)
	  at io.reactivex.internal.disposables.DisposableHelper.dispose(DisposableHelper.java:124)
	  at io.reactivex.observers.DisposableObserver.dispose(DisposableObserver.java:91)
	  at com.ownzones.app.ui.accountsettings.AccountSettingsActivityViewModel$makeUpdateUserInfoCall$observer$1.onNext(AccountSettingsActivityViewModel.kt:110)
	  at com.ownzones.app.ui.accountsettings.AccountSettingsActivityViewModel$makeUpdateUserInfoCall$observer$1.onNext(AccountSettingsActivityViewModel.kt:99)
[...]

PR #1663, opened 2019-10-09
Fix possible deadlock with http call execution

sav007:
- Relax global (on `ApolloServerInterceptor`) synchronization and use non blocking structures instead.
- Cancel previous http call request whenever new one is requested to execute
- Don't deliver http response result for previous http call if new one is requested to execute

Closes https://github.com/apollographql/apollo-android/issues/1647


coachroebuck:
Hi...

We might want to revisit this issue. We use your library in one of our bigger mobile apps, and began to see issues related to line 142 of the ApolloServerInterceptor.java file. 

We are seeing issues in which a different instance of the Call instance is being passed as a parameter into onResponse(). Because line 142 returns false, that callback is never invoked. In our case the app appears to hang during our API request.

I tried looking for some type of unit tests related to this component. It looks like this apparently doesn't exist. I will offer to go out of my way to write a bunch of unit tests, and read the details of this particular defect to gain a better understanding.

PR #1666, opened 2019-10-09
Allow to update test fixtures while running tests

martinbonnin:
This adds support for:

* `./gradlew :apollo-compiler:test -DupdateTestFixtures=true`
* `./gradlew :apollo-integrationr:test -DupdateTestFixtures=true`

When `-DupdateTestFixture=true` is specified, the test won't fail if the actual results do not match the expected ones. Instead, they will overwrite them. This can be used when making modifications to the compiler and/or interceptors to update the tests all at once.

As a nice side effect, since expected strings are in separate files, they are a bit easier to read as intellij can syntax-highlight them.

PS: all the `Utils.INSTANCE` changes are from converting Utils.java to Utils.kt.


martinbonnin:
CI fails because the current composite build does not apply the kotlin plugin. That also implies that coroutines tests are not run on the CI at the moment. 
I'll fix this once https://github.com/apollographql/apollo-android/pull/1649 is merged, it'll be easier.

sav007:
#1649 was merged

martinbonnin:
I just rebased on master. It should actually pass without any change on master since the composite build applies the kotlin plugin :crossed_fingers:

martinbonnin:
Had to rebase again against the kotlin build scripts :sweat_smile:

ericguffey:
Ty

PR #1673, opened 2019-10-13
Promote incubating plugin

martinbonnin:
This PR prepares the incubating plugin for the big jump :drum: :crossed_fingers:. Let me know if you think this is too early. As it's not 100% backward compatible, I bumped the version to 1.3.x and added a "migrating" section in the README.

In details, this PR:

* Renames `apollo-gradle-plugin` to `apollo-gradle-plugin-deprecated`
* Renames `apollo-gradle-plugin-incubating` to `apollo-gradle-plugin`
* Updates the deploy, travis scripts and a few other build scripts
* Updates the README

Depending the release planning, I would let this live in the snapshots for a couple of weeks and try to get more feedback before an actual 1.3.x release.


README.md
```diff
@@ -5,7 +5,7 @@
...
+Apollo-Android is a GraphQL compliant client that generates Java or Kotlin models from standard GraphQL queries.  These models give you a typesafe API to work with GraphQL servers.  Apollo will help you keep your GraphQL query statements together, organized, and easy to access from your code. Change a query and recompile your project - Apollo code gen will rebuild your data model.  Code generation also allows Apollo to read and unmarshal responses from the network without the need of any reflection (see example generated code below).  Future versions of Apollo-Android will also work with AutoValue and other value object generators.
```
sav007: > Future versions of Apollo-Android will also work with AutoValue and other value object generators.

Agh, not sure why this statement / promise was added at first place a long time ago, if you already touched readme could you please remove it.

sav007:
Looks great! Before merging I need to do some dance with Bintray and JCenter.
Just to confirm we introduce only one new artifact / module here called  `apollo-gradle-plugin-deprecated`, right?

martinbonnin:
>Just to confirm we introduce only one new artifact / module here called apollo-gradle-plugin-deprecated, right?

Correct. The incubating plugin will overwrite `apollo-gradle-plugin`. The current plugin will be in `apollo-gradle-plugin-deprecated` until we remove it in a future revision.

README.md
martinbonnin: It's removed

martinbonnin:
Putting this Pull Request on hold as there are still questions regarding:

* whether we should allow overriding queries in android variants
* whether we could offer a `sourceSet` api to configure the sourceFolders

tasomaniac:
After discussing this offline, I kind of agree that integrating fully into SourceSet mechanism can be an overkill (although it was powerful). Also it is ugly in Kotlin.

My suggestions regarding above,

- What about having `FileCollection` everywhere whenever files/folders are involved. Paths as strings are really bad for Gradle cache mechanisms. FileCollection is great since it will support multiple folders + merging multiple folders from other variants

- related: go over the public API and streamline naming. Currently, usage of `path` is inconsistent.

tasomaniac:
OMG!! So many conflicts. Sorry @martinbonnin 

While talking to Martin, I've realized that our use-case with older plugin is already supported with advanced usage in `compilationUnits`. With #1743 that is now encapsulated in a DSL with a better support.

#1746 also adds support for absolute paths instead of just relative. 

What's missing: 

- More documentation about `onCompilationUnits` DSL
- Support `SourceDirectorySet` instead of `DirectoryProperty` since it is much more flexible.
- Move `sources` DSL into `compilerParams`

doc/migrating-to-1-3.md
```diff
@@ -0,0 +1,106 @@
+# Migrating to 1.3
+
+The 1.3 plugin introduces several breaking changes to better deal with multiple GraphQL endpoints (services), lazy configuration and package name configuration. These changes only affect the plugin. The generated code is unchanged and your consumer code doesn't require any change but your `build.gradle` (or `build.gradle.kts`) files might.
+
+### Plugin id is now `com.apollographql.apollo`
+
+We renamed the plugin id from `com.apollographql.android` to `com.apollographql.apollo` to make it clear that the plugin works also for non Android projects. `com.apollographql.android` will be removed in a future revision.
+
+```groovy
+// Don't do this
```
tasomaniac: You could say `instead of`

doc/migrating-to-1-3.md
```diff
@@ -0,0 +1,106 @@
...
+apply plugin: 'com.apollographql.android'
+
+// Do this
+apply plugin: 'com.apollographql.apollo'
+```
+### Specifying schema and graphql files location
```
tasomaniac: I think it would be nice to keep the order according to order of which these take affect. Mentioning first about service usage and multiple service support can come first. Then this section with even more custom configuration may come afterwards. What do you think?

tasomaniac: Or maybe having a link to other plugin configuration doc can be an option too.

martinbonnin: Yes, for both.

martinbonnin:
There are a lot of outdated comments in this PR, I'll make a new one.

Issue #1674, opened 2019-10-15
Unknown scalar types can't be automatically parsed as Objects


DSteve595:
It seems infeasible to have to add an entry in `apollo { customTypeMapping = [ ... ] }` every time a new scalar is added to the schema, so I'm trying to have scalars "fall back" to being `String`s containing their values. When I do:
```kotlin
OperationResponseParser(
  operation,
  operation.responseFieldMapper(),
  ScalarTypeAdapters(emptyMap())
).parse(httpResponseBody)
```
This fails when hitting any scalar:
>java.lang.IllegalArgumentException: Can't map GraphQL type: MyScalar to: class java.lang.Object. Did you forget to add custom type adapter?

The lowest-maintenance fix I found for this is to create a `ScalarTypeAdapters` with each `CustomType` pointing to an adapter that simply returns `.value`.
```kotlin
val scalarTypeAdapters: ScalarTypeAdapters = run {
  val rawStringCustomTypeAdapter = object : CustomTypeAdapter<Any> {
    override fun decode(value: CustomTypeValue<*>): Any = value.value
    override fun encode(value: Any): CustomTypeValue<*> = CustomTypeValue.fromRawValue(value)
  }
  ScalarTypeAdapters(
    CustomType.values().associate { it to rawStringCustomTypeAdapter }
  )
}
```
This works, but seems a bit wasteful, especially as more scalar types are added.
Is it possible to clean this up?

PR #1685, opened 2019-10-17
Add default custom scalar type adapter for Object

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1674


sav007:
@martinbonnin @DSteve595

PR #1694, opened 2019-10-21
Add gitattributes file for compiler tests' expected outputs

DSteve595:
Currently, `apollo-compiler`'s tests fail on Windows because `checkTestFixture` directly compares file contents, but Windows writes newlines differently from Unix.

image

~This fixes that by stripping carriage return (`\r`) characters from the compared strings.~
Now using gitattributes

### DIFF
--- a/apollo-compiler/src/test/graphql/.gitattributes
+++ b/apollo-compiler/src/test/graphql/.gitattributes
@@ -1,0 +1,3 @@
+# apollo generates code without carriage return. Make sure git does not add them when checking out the test fixtures
+/*.kt eof=lf
+/*.java eof=lf


martinbonnin:
Thanks for catching this. 2 comments:

1. I opened an issue to see if it's easy to run Travis on Windows (https://github.com/apollographql/apollo-android/issues/1696). That might help maintain windows in the long run.
2. I'm not sure where the `\r` come from. There shouldn't be any `\r` in the commited test fixtures. Are they added by `File.readText()` ? That would seem weird to me.

DSteve595:
>That might help maintain windows in the long run.

That'd be good of course, but realistically I think I'm in the minority when it comes to dev environments. If it adds too much complexity/annoyance to maintain, I totally understand leaving it out.

>I'm not sure where the \r come from

Checking it out..

DSteve595:
I believe the `\r`s are added by Git during checkout, respecting [the core.autocrlf setting](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings).

martinbonnin:
TIL git changes my data :scream:...

What about a .gitattributes file ? 

```
# apollo generates code without carriage return. Make sure git does not add them when checking out the test fixtures
apollo-compiler/src/test/graphql/com/example/* text eof=lf
```

Granted there's very little change but if for some reason, codegen starts generating `\r`, I would still like the tests to fail.

DSteve595:
That might make sense. Will leave it for a repo owner to decide. (cc @sav007)

sav007:
I think it makes totally sense to add `.gitattributes` file for test fixtures. @DSteve595 could you pls add it?

DSteve595:
Will do once I'm back on my Windows machine.

DSteve595:
Very sorry for the silence, I completely forgot about this. Thanks to @martinbonnin for reminding me about it.

I just pushed a directory-scoped gitattributes file defining LF line endings for .kt and .java files in the test dir.

Issue #1454, opened 2019-07-23
Support lazy task configuration for Gradle plugin


tasomaniac:
The Gradle plugin does not use the new lazy task configuration. This means that its tasks are eagerly created + it causes other dependant Android/Java/Kotlin tasks to be eagerly realized.

Integrate Lazy task configuration for build speed improvements. 

More info: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

martinbonnin:
It's not possible at the moment because `Variant.registerJavaGeneratingTask` does not allow for lazy configuration. There's an issue in the google bug tracker: https://issuetracker.google.com/issues/117343589

tasomaniac:
Yes I realized this shortly after trying to solve this. I forgot to come back and write it here. Thanks @martinbonnin for that. 

As far as I learned, Android team will replace this with a newer lazy FileProvider/Directory approach and will not directly work with Tasks. That's definitely better but that means that this is basically blocked.

Considering the schedule of Android Studio releases, this may take up to 1 year.

PR #1715, opened 2019-10-28
Incubating plugin: lazy configuration and gradle-y things

martinbonnin:
Sorry again for the huge PR. What I thought was replacing `Strings` with `Property<String>` took me way further than I'd like to admit.  

The main motivation for this PR was to allow wiring non-apollo tasks to the input of the `ApolloCodegenTask`. For an exemple, if you have a `genGraphqlTask` task that generates the .graphql files:

```kotlin
configure<ApolloExtension> {
  compilationUnits.all {
    this.setSources(genGraphqlTask.flatMap { it.outputDir })
  }
}
```

See LazyTests.kt for the complete exemple. Also, since everything is lazy, the configuration time should be better if no apollo task is required although this would have to be measured. 

Other things changed by this PR:

* `api` vs `internal` separation: `api` only contains a minimal set of interfaces now with the implementation being under an `internal` namespace so anyone using it should be clearly aware they're doing something dangerous.
* tasks/object construction is more gradle-idiomatic: they are abstract classes so that gradle can automagically instanciate the various properties and `Closure` overloads for methods that take an `Action`
* `service.schemaFilePath` has been renamed `service.schemaPath`. It is now relative to the current source set (`src/main/graphql/$schemaPath`). It is mainly so that it is consistent with `sourceFolder`. It's also been made optional and will default to the first `.json` file found in `sourceFolder`
* `sourceFolder` is now optional as well and defaults to `"."`
* `build.gradle.kt` won't be able to use the property syntax anymore (`useSemanticNaming = true // that won't work`). That's because `useSemanticNaming` and other parameters are gradle Properties and cannot just be used as kotlin class members.


martinbonnin:
Ping @tasomaniac @DSteve595

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -0,0 +1,273 @@
...
+  override val outputDir = project.objects.directoryProperty()
+  override val transformedQueriesDir = project.objects.directoryProperty()
+
+  init {
+    if (!compilerParams.generateKotlinModels.isPresent) {
+      compilerParams.generateKotlinModels.set(false)
+    }
+  }
+
+  fun sources():Sources {
```
DSteve595: Missing space

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ObjectFactoryExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.apollographql.apollo.gradle.internal
+
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
+//inline fun <reified T:Any> ObjectFactory.property(t: T): Property<T> {
+//  return property<T>(t::class.java).apply { this.set(t) }
+//}
```
DSteve595: Remove?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ResolvedCompilerParams.kt
```diff
@@ -0,0 +1,70 @@
...
+//        throw IllegalArgumentException("""
+//        apollo.outputPackageName is not supported anymore as it doesn't work for multiple services and also flattens the packages.
+//
+//      """.trimIndent() + ApolloPlugin.useService(apolloExtension.schemaFilePath, apolloExtension.outputPackageName))
+//      }
+//
+//      return params
+//    }
+//  }
+//}
```
DSteve595: ?

DSteve595:
Nice!

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ObjectFactoryExtensions.kt
martinbonnin: Yep

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ResolvedCompilerParams.kt
martinbonnin: obsolete

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -26,8 +26,9 @@ class GraphQLCompiler {
         generateVisitorForPolymorphicDatatypes = args.generateVisitorForPolymorphicDatatypes
     )
 
-    if (args.packageNameProvider.schemaPackageName.isNotBlank()) {
-      File(args.outputDir, args.packageNameProvider.schemaPackageName.replace('.', File.separatorChar)).deleteRecursively()
+    val schemaPackageName = (args.packageNameProvider as? DeprecatedPackageNameProvider)?.schemaPackageName
```
tasomaniac: Just a question. If this is not the `Deprecated` one, the operation gets ignored. Is that expected?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/AndroidTaskConfigurator.kt
```diff
@@ -43,6 +41,7 @@ object AndroidTaskConfigurator {
     return container
   }
 
+  // TODO: make this lazy
```
tasomaniac: I would put a link to this issue. https://github.com/apollographql/apollo-android/issues/1454 Since it is blocked, it would be useful.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
martinbonnin: This deletes the output directory (I guess to prevent stale files). But this is already done there: https://github.com/apollographql/apollo-android/blob/e1084c9d9e6a47fc1c12a230c5c543e9cd2d9368/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt#L77

So it's no use doing it twice. I just didn't touch the current behaviour for the deprecated plugin to be on the safe side.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
```diff
@@ -122,12 +122,12 @@ class CodeGenTest(val folder: File) {
 
       val outputPackageName = "com.example.${folder.name}"
 
-      val packageNameProvider = PackageNameProvider(
+      val packageNameProvider = DeprecatedPackageNameProvider(
```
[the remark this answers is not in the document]
martinbonnin: Right, no need to use `DeprecatedPackageNameProvider` here, I'll do a proper test implementation (especially since it's doing almost nothing...)

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -1,60 +1,23 @@
...
-   * By default, this is the directory where the schema.json is stored or "." if the schema is outside.
-   * You need to define this if you have two or more services whose schema is stored outside src/{foo}/graphql.
-   * If not, you can certainly omit it.
-   */
-  var sourceFolderPath: String? = null
+  val schemaPath: Property<String>
+  fun schemaPath(schemaFilePath: String)
 
-  var rootPackageName: String? = null
+  val sourceFolder: Property<String>
```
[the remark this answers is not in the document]
martinbonnin: The thing is android variants... `sourceFolder` has to be a string because the plugin will look into:

* src/main/graphql/$sourceFolder
* src/debug/graphql/$sourceFolder
* src/demoDebug/graphql/$sourceFolder

Service is really DSL oriented, it's just a helper to configure the tasks. If the user wants to do more configuration, he/she will have to use the `CompilationUnit`.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/AndroidTaskConfigurator.kt
martinbonnin: 👍 will do

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -1,60 +1,23 @@
...
 
-  /**
-   * Place where the graphql files are searched.
-   * This path is relative to the current sourceSet (e.g src/{foo}/graphql/{sourceFolderPath})
-   * By default, this is the directory where the schema.json is stored or "." if the schema is outside.
-   * You need to define this if you have two or more services whose schema is stored outside src/{foo}/graphql.
-   * If not, you can certainly omit it.
-   */
-  var sourceFolderPath: String? = null
+  val schemaPath: Property<String>
```
tasomaniac: Similar to my comment below, this could be better to be `RegularFileProperty`. I found myself setting this as file all the time.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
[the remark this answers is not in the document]
tasomaniac: Thinking about this again, I think it would be good to support Android Variants in more explicit way and keep these as files. 

This `service` idea can work pretty well with variants. Imagine you have a service and you configure common parts first (just like in `defaultConfig` in Android plugin). And then you can do per-variant configuration that would override the common configuration.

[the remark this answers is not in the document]
martinbonnin: Something like below?

```groovy
apollo {
  service("github") {
    // will only look in src/main/graphql/github by default
    sourceFolder = file("github") 

    productFlavors {
      demo {
        // will only look in src/main/graphql/github-demo for the demo flavor
        sourceFolder = file("github-demo") 
      }
    }
  }
}
```

I don't like this too much as this is against the android standard sourceSets resolution (https://developer.android.com/studio/build/build-variants#sourceset-build). In the example above, files in `src/debug/graphql/github` would never be picked up.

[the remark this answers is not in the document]
martinbonnin: We might want to block https://github.com/apollographql/apollo-android/pull/1673 until we decide something there. Once the plugin is promoted, it'll be much more difficult to do impactful changes.

PR #1725, opened 2019-11-03
Make okhttp and apollo api deps in http-cache

ZacSweers:
These are both part of the cache's public API, and gives warnings in the IDE when they're hidden from the classpath.

image


martinbonnin:
You'll need to apply the `java-library` plugin for `api` to be available.

ZacSweers:
To fix the build or you think that’s an undue requirement?

martinbonnin:
I guess it's a reasonable addition, not sure why it's not there already actually.

PR #1736, opened 2019-11-06
Add proper support for custom JSON object / list scalar types

sav007:
Add support for encoding / decoding custom GraphQL scalar type value represented by JSON object into Map.
Add support for encoding / decoding custom GraphQL scalar type value represented by JSON list into List.


martinbonnin:
Just curious: are JSON objects/lists scalar types actually part of the spec or just some handy extensions that we allow ?

sav007:
@martinbonnin Well I didn't find in spec that custom scalar type should be strictly a string, and looking into the introspection query response, this is how scalar type is defined:
```
{
          "kind": "SCALAR",
          "name": "Date",
          "description": "The `Date` scalar type represents date format.",
          "fields": null,
          "inputFields": null,
          "interfaces": null,
          "enumValues": null,
          "possibleTypes": null
        }
```

At the same time custom scalar type probably should be true scalar. But in the past I remember that it was a feature request to add support to map custom  scalar type into `Map`, like for instance to be able to map geo coordinates `{"longitude": ..., "latitude": }` into `Map`.

So yes it's probably smth we offer as convenience.

sav007:
For sure, we just provide a default adapter for list and map.

taion:
They are explicitly allowed by the spec, which is why `GraphQLJSON` works. Just somewhat underdocumented.

PR #1737, opened 2019-11-06
Incubating Plugin: Workaround metaspace OOM

martinbonnin:
We're running 40+ tests from the same gradle daemon and this fails on CI (and sometimes on my machine too) due to a lack of MetaSpace. That seems to indicate a ClassLoader leak but it might as well be something else so I'm just restarting the daemon every 8 tests here.

### DIFF
--- a/apollo-gradle-plugin-incubating/build.gradle.kts
+++ b/apollo-gradle-plugin-incubating/build.gradle.kts
@@ -28,6 +28,11 @@ dependencies {
 }
 
 tasks.withType<Test> {
+  // Restart the daemon once in a while or we end up running out of MetaSpace
+  // It's not clear if it's a real ClassLoader leak or something else. The okio timeout thread does hold some ClassLoaders
+  // for up to 60s. The heap dumps also show some process reaper threads but it might just as well be a temporary thing, not sure.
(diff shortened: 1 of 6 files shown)


apollo-gradle-plugin-incubating/build.gradle.kts
```diff
@@ -28,6 +28,11 @@ dependencies {
 }
 
 tasks.withType<Test> {
+  // Restart the daemon once in a while or we end up running out of MetaSpace
+  // It's not clear if it's a real ClassLoader leak or something else. The okio timeout thread does hold some ClassLoaders
+  // for up to 60s. The heap dumps also show some process reaper threads but it might just as well be a temporary thing, not sure.
+  // See https://github.com/gradle/gradle/issues/8354
+  setForkEvery(8L)
```
sav007: Wow, TIL.

martinbonnin: Yea, that stuff is super weird. I tried to look a bit into the heap dumps but it's tough to debug. I guess real users won't load and unload the pugin 40+ times so we should be fine for now.

PR #1739, opened 2019-11-06
Incubating Plugin: Do not allow queries or schema variant overrides

martinbonnin:
Similar to what the Android Plugin does for java sources (https://developer.android.com/studio/build/build-variants#sourceset-build), we should not allow overrides over variants. The layout below will throw an exception:

```
// duplicate schema.json
src/freeDebug/graphql/com/example/schema.json
src/main/graphql/com/example/schema.json

// duplicate Query.graphql
src/freeDebug/graphql/com/example/Query.graphql
src/main/graphql/com/example/Query.graphql
```

This makes it explicit that files are mutually exclusive for a given variant.

ping @tasomaniac


tasomaniac:
LGTM

PR #1740, opened 2019-11-06
Incubating plugin: do not resolve properties in parent DSL scopes

martinbonnin:
This prevents these kinds of scenarios where the user would inadvertently call a method/property on a parent scope:

```kotlin

      configure<ApolloExtension> {
        service("starwars") {
          // this should trigger a compilation error
          schemaFilePath.set("foo")
        }
      }
```

I did this pull request as an exercise but I have mixed feelings about it. I'm not sure what other plugins/gradle do. If they allow this behavior, it's maybe better to favor consistency.

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
@@ -1,6 +1,7 @@
 package com.apollographql.apollo.gradle.api
 
 import com.apollographql.apollo.gradle.internal.DefaultService
+import groovy.lang.Closure
 import org.gradle.api.Action
 import org.gradle.api.DomainObjectCollection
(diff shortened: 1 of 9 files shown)


martinbonnin:
@DSteve595 @sav007 I've been thinking a bit about this and I would say we should wait a bit more before we merge this. `@DslMarker` is most likely ok because I don't think anything from `ApolloExtension` is needed inside `Service` but limiting the scope of the gradle closure to the receiver only prevents is even more restrictive. That will prevent to access methods on the `project` instance like the build directory or extra properties for an exemple.

Unless you tell me otherwise, I'll close this pull request for now.

martinbonnin:
@tasomaniac can you clarify ? Are you saying we should merge this pull request ? Wouldn't it be a problem to not have access to the `project` object in groovy from the DSL ?

tasomaniac:
🙌 Let's keep an eye on this. This can be tricky with Groovy. Our build scripts are in groovy. If I notice anything weird with the incubating plugin, I will report.

PR #1743, opened 2019-11-07
[Incubating Plugin] Provide a DSL to change compilationUnit settings

tasomaniac:
- Introduce new `dsl` called `sources` to set the graphql sources.
- `rootPackageName` is moved to `CompilerParams`. 
- Introcuded `compilerParams {}` DSL to override any `compilerParams` specific value per `compilationUnit`
- Remove the long `setSources`, the new DSL is much more useful and provide the same functionality

It would look like:

```
apollo {
  onCompilationUnits {
    compilerParams {
      rootPackageName("com.example.api")
    }
    sources {
      graphqlDir(files('some/path'))
      schemaFile(file('schema.json'))
    }
  }
}
```

## Tests

`LazyTests` uses the `setSources` function and they're still green. I would like to write more tests to see the whole DSL functionality works.

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
@@ -2,12 +2,10 @@ package com.apollographql.apollo.gradle.api
 
 import com.apollographql.apollo.gradle.internal.DefaultService
 import org.gradle.api.Action
-import org.gradle.api.DomainObjectCollection
 import org.gradle.api.provider.Property
 
 interface ApolloExtension: CompilerParams {
 
-  val compilationUnits: DomainObjectCollection<CompilationUnit>
   fun onCompilationUnits(action: Action<CompilationUnit>)
 
   fun service(name: String, action: Action<DefaultService>)
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
@@ -8,6 +8,7 @@ import org.gradle.api.provider.Property
 interface ApolloExtension: CompilerParams {
 
   val compilationUnits: DomainObjectCollection<CompilationUnit>
+  fun onCompilationUnits(action: Action<CompilationUnit>)
 
   fun service(name: String, action: Action<DefaultService>)
 
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
@@ -17,9 +17,7 @@ interface CompilationUnit {
   val outputDir: Provider<Directory>
   val transformedQueriesDir: Provider<Directory>
 
-  fun compilerParams(closure: Closure<*>)
   fun compilerParams(action: Action<CompilerParams>)
-  fun sources(closure: Closure<*>)
   fun sources(action: Action<Sources>)
 
   class Sources(
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
@@ -5,6 +5,7 @@ import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.Provider
+import java.io.File
 
 interface CompilationUnit {
   val name: String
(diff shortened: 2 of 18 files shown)


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
```diff
@@ -17,6 +17,12 @@ interface CompilationUnit {
   val outputDir: Provider<Directory>
   val transformedQueriesDir: Provider<Directory>
 
+  fun configure(configure: Action<Params>)
   fun setSources(rootFolder: Provider<Directory>)
-  fun setSources(rootFolders: FileCollection, graphqlFiles: FileCollection, schemaFile: Provider<RegularFile>, rootPackageName: Provider<String>)
+
+  data class Params(
+      val graphqlFolder: DirectoryProperty,
```
martinbonnin: Maybe make this a `SourceDirectorySet` so we can add multiple folders and add ìncludes`/`excludes` ?

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
```diff
@@ -17,6 +17,12 @@ interface CompilationUnit {
   val outputDir: Provider<Directory>
   val transformedQueriesDir: Provider<Directory>
 
+  fun configure(configure: Action<Params>)
```
martinbonnin: How is that going to be working from kotlin build scripts ? Can kotlin "recognize" the `configure` or does it have to be called explicitely ?

tasomaniac: Both in Groovy and Kotlin, this needs to be called explicitly. I figured, `compilationUnit.all` is a callback with possibly multiple units. People may want to do it conditionally.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilationUnit.kt
tasomaniac: The problem with that is that `SourceDirectorySet` and `AndroidSourceDirectorySet` are 2 separate classes with 100% same API surface. It is hard to support both.

If we would have full `SourceSet` support, then this would make sense. But now, it is also hard to create a new `SourceDirectorySet` since it is created by the official plugins.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
```diff
@@ -8,6 +8,7 @@ import org.gradle.api.provider.Property
 interface ApolloExtension: CompilerParams {
 
   val compilationUnits: DomainObjectCollection<CompilationUnit>
```
martinbonnin: Maybe remove this from the public API now that we have `onCompilationUnits` ? That'll make it easier to change implementation later if needed

tasomaniac: Good idea 🙌 I was going to ask that actually. Android plugin and Gradle provides both time to time. Especially Gradle is full of overloaded functions doing the same thing 🙃 That's why I was not sure. Will make it private

martinbonnin: I think it would make sense to have it public if the collection were writable by users but it's not, we don't want anyone to write something in there.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -204,28 +209,32 @@ class DefaultCompilationUnit(
...
-    fun fromFiles(project: Project, apolloExtension: DefaultApolloExtension, apolloVariant: ApolloVariant): List<DefaultCompilationUnit> {
+    fun fromFiles(project: Project, apolloExtension: DefaultApolloExtension, apolloVariant: ApolloVariant): DefaultCompilationUnit {
       val sourceSetNames = apolloVariant.sourceSetNames
       val schemaFiles = findFilesInSourceSets(project, sourceSetNames, ".") {
         it.name == "schema.json"
       }
+      require(schemaFiles.size == 1) {
+        """|
+          |By default only one schema.json file is supported. If you have multiple schema.json,
+          |Please use a service instead:
```
martinbonnin: Maybe "Please use multiple services instead" ?

tasomaniac: Makes a lot of sense 💯

tasomaniac: With that, I will also change the example just below to have 2 services.

tasomaniac: From the test, here is how it will look like: 

```
A problem occurred configuring root project 'testProject'.
> By default only one schema.json file is supported.
  Please use multiple services instead:
  apollo {
    service("serviceName") {
      sourceFolder = "com/example/abc/"
    }
    service("serviceName") {
      sourceFolder = "com/example/"
    }
  }
```

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -219,7 +219,7 @@ open class DefaultCompilationUnit @Inject constructor(
           apolloVariant.sourceSetNames,
           project
       ).apply {
-        apolloVariant.androidVariant
+        androidVariant = apolloVariant.androidVariant
```
martinbonnin: Oh my, thanks !

PR #1751, opened 2019-11-09
[Incubating Plugin] Little clean-up

tasomaniac:
- Unused toPackageName is removed
- File.child() is moved to Plugin since it is only used there.
- Also done small improvements


martinbonnin:
We might be able to remove `File.child` altogether. I added it because of the windows file separator but File(parent, "src/main/graphql") might work equally well.

tasomaniac:
I found it quite useful in cases where some paths are variables. It's up to you. It's also easier to remove once it's only inside the plugin. 😉

martinbonnin:
No need to rush the removal, it's not very complicated code in all cases.

PR #1753, opened 2019-11-10
[Incubating plugin] CompilerParams overrides + SourceDirectorySet all the things 

martinbonnin:
This pull request moves the sources configuration to `CompilerParams`. All of `ApolloExtension`, `Service` and `CompilationUnit` implement `CompilerParams` so we can have a nice cascade of params overrides:

- global params are in the `ApolloExtension`. This is also were they used to be so it's backward compatible.
- service params are in `Service`.
- lastly, users can override on a `CompilationUnit` basis.

Also the graphql sources are now a `SourceDirectorySet` so it's easier to add multiple folders, includes, excludes, etc... for more advanced use cases.

Very basic use case:
```kotlin
// no apollo block, one service, reads everything under src/$foo/graphql
```

Simple use case:
```kotlin
// multiple service, the plugin will find the graphql files and schema:
apollo {
  service("starwars") {
    sourceFolder = "starwars"
  }
  service("githunt") {
    sourceFolder = "githunt"
  }
}
```

Advanced use case:
```kotlin
// task inputs wiring, SourceDirectorySet customization etc...
apollo {
  service("starwars") {
  }
  service("githunt") {
  }

  onCompilationUnits {
    graphqlSourceDirectorySet.srcDir(installTask.flatMap { it.outputDir })
    graphqlSourceDirectorySet.exclude("**/*.gql")
    // do something with androidVariant, outputDir and/or transformedQueries
  }
}
```

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
@@ -151,21 +151,26 @@ open class ApolloPlugin : Plugin<Project> {
         it.group = TASK_GROUP
         it.description = "Generate Apollo models for ${compilationUnit.name.capitalize()} GraphQL queries"
 
-        val sources = compilationUnit.sources()
+        val compilerParams = compilationUnit
+            .withFallback(project.objects, compilationUnit.service)
+            .withFallback(project.objects, compilationUnit.apolloExtension)
 
-        it.graphqlFiles.setFrom(sources.graphqlFiles)
-        it.rootFolders.set(sources.rootFolders.map { it.absolutePath })
-        it.schemaFile.set(sources.schemaFile)
+        compilationUnit.setSourcesIfNeeded(compilerParams.graphqlSourceDirectorySet, compilerParams.schemaFile)
 
-        it.nullableValueType.set(compilationUnit.compilerParams.nullableValueType)
-        it.useSemanticNaming.set(compilationUnit.compilerParams.useSemanticNaming)
-        it.generateModelBuilder.set(compilationUnit.compilerParams.generateModelBuilder)
-        it.useJavaBeansSemanticNaming.set(compilationUnit.compilerParams.useJavaBeansSemanticNaming)
-        it.suppressRawTypesWarning.set(compilationUnit.compilerParams.suppressRawTypesWarning)
-        it.generateKotlinModels.set(compilationUnit.compilerParams.generateKotlinModels)
-        it.generateVisitorForPolymorphicDatatypes.set(compilationUnit.compilerParams.generateVisitorForPolymorphicDatatypes)
-        it.customTypeMapping.set(compilationUnit.compilerParams.customTypeMapping)
-        it.rootPackageName.set(compilationUnit.compilerParams.rootPackageName)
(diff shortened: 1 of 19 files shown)


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
```diff
@@ -151,29 +151,34 @@ open class ApolloPlugin : Plugin<Project> {
...
-        it.nullableValueType.set(compilationUnit.compilerParams.nullableValueType)
-        it.useSemanticNaming.set(compilationUnit.compilerParams.useSemanticNaming)
-        it.generateModelBuilder.set(compilationUnit.compilerParams.generateModelBuilder)
-        it.useJavaBeansSemanticNaming.set(compilationUnit.compilerParams.useJavaBeansSemanticNaming)
-        it.suppressRawTypesWarning.set(compilationUnit.compilerParams.suppressRawTypesWarning)
-        it.generateKotlinModels.set(compilationUnit.compilerParams.generateKotlinModels)
-        it.generateVisitorForPolymorphicDatatypes.set(compilationUnit.compilerParams.generateVisitorForPolymorphicDatatypes)
-        it.customTypeMapping.set(compilationUnit.compilerParams.customTypeMapping)
-        it.rootPackageName.set(compilationUnit.compilerParams.rootPackageName)
+        val compilerParams = compilationUnit
```
tasomaniac: This is nice. The rest is more clear. It is also more clear here that we do fallbacks.

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/util/TestUtils.kt
```diff
@@ -26,7 +26,9 @@ object TestUtils {
 
     block(dest)
 
-    dest.deleteRecursively()
+    // It's ok to not delete the directory as it will be deleted before next test
+    // During developement, it's easy to keep the testProject around to investigate if something goes wrong
+    // dest.deleteRecursively()
```
tasomaniac: At Novoda Static Analysis Plugin, we would use jUnit temp folder rule to create the test project into a random folder and then only delete the folder if the test is successful

tasomaniac: https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/test/groovy/com/novoda/test/TestProjectRule.groovy#L59

martinbonnin: I actually like that the test is always in the same location so that I can use my bash history to replay any commands. I tend to do that a lot with stuff like: 

* `./gradlew -p apollo-gradle-plugin-incubating/build/testProject -Dorg.gradle.debug=true`
* `find apollo-gradle-plugin-incubating/build/testProject/build/generated/sources/apollo `

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
martinbonnin: Yes, exactly. The only special case now is with `generateKotlinModels`.

PR #1754, opened 2019-11-10
New Kotlin extensions to RxJava2 support and Input API

tasomaniac:
Add Kotlin extensions to RxJava artifact. 

## Note

Added the extensions followed by a method used by Refrofit, AutoDispose and others. With this approach, there is not need to publish a new artifact. AutoDispose deprecated their `ktx` artifacts earlier this year. 
https://github.com/uber/AutoDispose/

This is done basically with combination of `compileOnly` Kotlin stdlib and `@JvmSyncthetic` annotation. The annotation makes it impossible for Java callers to call these methods.

Kotlin needs to be configured for them to be called. It then magically works. 🎉 

## Extension Function API (Feedback needed)

Just `rx()` function is added to replace `Rx2Apollo.from()`. These are mere replacements and are not the best. 

`rxQuery`, `rxMutate`, `rxPrefetch` and `rxSubscribe` are extensions on `ApolloClient` and imitates the already available API in `ApolloClient` and provides much better API for Kotlin users.

Samples are also updated to demonstrate the new API. Here is a gist of the diff:

```
val call = apolloClient.query(repositoryDetailQuery)
Rx2Apollo.from(call)
```

becomes

```
apolloClient.rxQuery(repositoryDetailQuery)
```

If you need to configure the `call` that has been created. You could also do:

```
apolloClient.rxQuery(query) {
    httpCachePolicy(HttpCachePolicy.CACHE_FIRST)
}
```

### DIFF
--- a/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
+++ b/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
@@ -1,29 +1,50 @@
 @file:Suppress("NOTHING_TO_INLINE")
+@file:JvmName("KotlinExtensions")
 
 package com.apollographql.apollo.rx2
 
-import com.apollographql.apollo.*
-import com.apollographql.apollo.api.*
+import com.apollographql.apollo.ApolloCall
+import com.apollographql.apollo.ApolloClient
+import com.apollographql.apollo.ApolloMutationCall
+import com.apollographql.apollo.ApolloPrefetch
+import com.apollographql.apollo.ApolloQueryCall
+import com.apollographql.apollo.ApolloQueryWatcher
+import com.apollographql.apollo.ApolloSubscriptionCall
+import com.apollographql.apollo.api.Mutation
+import com.apollographql.apollo.api.Operation
+import com.apollographql.apollo.api.Query
+import com.apollographql.apollo.api.Response
+import com.apollographql.apollo.api.Subscription
 import com.apollographql.apollo.cache.normalized.ApolloStoreOperation
-import io.reactivex.*
+import io.reactivex.BackpressureStrategy
+import io.reactivex.Completable
+import io.reactivex.Flowable
+import io.reactivex.Observable
+import io.reactivex.Single
+import io.reactivex.annotations.CheckReturnValue
 
 @JvmSynthetic
+@CheckReturnValue
 inline fun ApolloPrefetch.rx(): Completable =
     Rx2Apollo.from(this)
 
 @JvmSynthetic
+@CheckReturnValue
 inline fun <T> ApolloStoreOperation<T>.rx(): Single<T> =
     Rx2Apollo.from(this)
 
 @JvmSynthetic
+@CheckReturnValue
 inline fun <T> ApolloQueryWatcher<T>.rx(): Observable<Response<T>> =
     Rx2Apollo.from(this)
 
 @JvmSynthetic
+@CheckReturnValue
 inline fun <T> ApolloCall<T>.rx(): Observable<Response<T>> =
     Rx2Apollo.from(this)
 
 @JvmSynthetic
+@CheckReturnValue
 inline fun <T> ApolloSubscriptionCall<T>.rx(
     backpressureStrategy: BackpressureStrategy = BackpressureStrategy.LATEST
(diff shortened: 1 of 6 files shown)


martinbonnin:
NIce! 

There was some discussion whether we should introduce an artifact called `apollo-ktx` and add all the Kotlin in there (including coroutines-support) or whether we should just add Kotlin to the individual artifacts and hide them from java.
I guess we're not adding `apollo-ktx` then, right ?

tasomaniac:
This is far better approach, imho. AutoDispose also have a Coroutines artifact which separate. I think it makes sense to keep Coroutines separate since it is not really just ktx (Kotlin extensions) but a separate interop library.

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
```diff
@@ -0,0 +1,109 @@
+@file:Suppress("NOTHING_TO_INLINE")
+@file:JvmName("KotlinExtensions")
```
AOrobator: If we plan on introducing more kotlin extensions that aren't related to Rx, maybe we'd want to rename this? I don't see anything wrong with keeping the original filename of RxJavaExtensions.

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
```diff
@@ -0,0 +1,109 @@
...
+    query: Query<D, T, V>,
+    configure: ApolloQueryCall<T>.() -> Unit = {}
+): Observable<Response<T>> = query(query).apply(configure).rx()
+
+/**
+ * Creates a new [ApolloMutationCall] call and then converts it to a [Single].
+ */
+@JvmSynthetic
+@CheckReturnValue
+inline fun <D : Operation.Data, T, V : Operation.Variables> ApolloClient.rxMutate(
```
AOrobator: There should also be a `Single<T>` version for queries

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
tasomaniac: I've thought about that. It's not meant for Java users. It's just a friendly name if anybody discovers. Since it is inside `rx2` package, it is unique across modules.

I've looked at beat practices established by others. Both AutoDispose and Retrofit use the same name.

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
tasomaniac: In our project, we have that. Should we add it for convenience? The problem is that there might be multiple emissions based on the configuration. If there are multiple emissions, then it will just crash.

Users can always use `toSingleOrError()` which makes it more explicit.

AOrobator: I'm assuming there'd be multiple emissions if there was a cache involved? Are there any other uses cases that would trigger multiple emissions? I agree that explicit errors would be better, makes for less surprises

AOrobator: My version of `toSingle` is a bit more involved.

```kotlin
data class ApolloErrorException(val errors: List<Error>) : Throwable()

val <T> Response<T>.exception: ApolloErrorException
  get() = ApolloErrorException(errors())

fun <T> Response<T>.responseToSingle(): Single<Response<T>> =
  if (hasErrors()) {
    Single.error(exception)
  } else {
    Single.just(this)
  }

fun <T> Observable<Response<T>>.toSingle(): Single<Response<T>> =
  retryWithBackoffAndJitter()
      .flatMapSingle { it.responseToSingle() }
      .singleOrError()
```

tasomaniac: Yes, multiple emissions would happen only when there is a cache.

I think for these extensions, it is more logical to keep them consistent with the original Java integration. That's why I don't want to have any custom logic in Kotlin extensions.

I think your suggestion is great. It would be great if you can contribute that back into Java implementation since it is not Kotlin-only.

PR #1755, opened 2019-11-10
Make java models accessible from kotlin

martinbonnin:
When we call `registerJavaSources()`, the kotlin plugin has already configured its sources and therefore misses the new generated folder.

This comes from the fact that we use `project.afterEvaluate{}` to make sure   apolloExtension.generateKotlinModels has been populated correctly and we therefore postpone the task registration to after the Android and Kotlin plugins have done their magic.

This feels a bit awkward but I'm not sure there's a nice solution here. Maybe have 2 different plugins: one that generates Kotlin and the other one that generates Java. That would simplify things quite a bit at the expense of breaking a bit more backward compatibility. 

Also that would mean users woulnd't be able to configure java/kotlin models on a per-service anymore but maybe it's not that bad. Actually defaulting to generating kotlin models if the kotlin plugin is applied would certainly be a better first experience for most users.


tasomaniac:
This has been always tricky with plugin generated code vs Kotlin. This looks like a good solution.

PR #1761, opened 2019-11-12
Generate function to parse operation response

sav007:
- Add new API `parse` function to `Operation` interface for parsing operation GraphQL response
- Introduce util class `SimpleOperationResponseParser` to delegate operation response parsing
- Update generated models implement `Operation.parse` function

This is final PR to make it possible use generated models without runtime module.

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
@@ -44,6 +44,13 @@ public interface Operation<D extends Operation.Data, T, V extends Operation.Vari
    */
   @NotNull String operationId();
 
+  /**
+   * Parses provided GraphQL operation raw response
+   *
+   * @param response operation raw response to parse
+   * @param scalarTypeAdapters configured instance of custom GraphQL scalar type adapters
+   * @return parsed GraphQL operation {@link Response}
+   */
   @NotNull Response<T> parse(@NotNull Map<String, Object> response, @NotNull ScalarTypeAdapters scalarTypeAdapters);
 
   /**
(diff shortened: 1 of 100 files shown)


apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
```diff
@@ -44,6 +44,15 @@
...
   @NotNull String operationId();
 
+  /**
+   * Parses provided GraphQL operation raw response
+   *
+   * @param response operation raw response to parse
+   * @param scalarTypeAdapters configured instance of custom GraphQL scalar type adapters
+   * @return parsed GraphQL operation {@link Response}
+   */
+  @NotNull Response<T> parse(@NotNull Map<String, Object> response, @NotNull ScalarTypeAdapters scalarTypeAdapters);
```
martinbonnin: How hard would it be to support a streaming API ? 
Something like:
`@NotNull Response<T> parse(@NotNull okio.Source source, @NotNull ScalarTypeAdapters scalarTypeAdapters);`

sav007: It's not hard at all, but keep in mind Kotlin Native, until we get `Okio` as KN library we shouldn't use any JVM libs.

martinbonnin: Looks like `Okio` is multiplatform actually: https://square.github.io/okio/changelog/#version-230.
[Source](https://square.github.io/okio/2.x/okio/okio/-source/) is written as supporting native/js

sav007: Will definitely take a look

Issue #1748, opened 2019-11-08
Replacement for the old DebugAPI.json files


jaregier:
We recently tried upgrading our Apollo to use the latest compiler.  What we found was that a file that was generated as part of the old compiler, `DebugAPI.json` (or `ReleaseAPI.json`, whatever build variant you are running) is no longer generated as part of the codegen step.  I understand that the new compiler does things differently, so let me explain the use case:

We use persisted queries, but we don't use [automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/).  We manually sync the persisted queries to our server as part of a CI step.  Previously, we would use the `DebugAPI.json` and send it to our server, but really all we needed was a mapping of operationId -> sourceWithFragments.

Now that this file is no longer generated, we'll need to find another way to get this mapping.  We aren't seeing any mapping files generated as part of the build process; we've also tried using the [extract](https://github.com/apollographql/apollo-tooling#apollo-clientextract-output) command from the CLI interface, but differences in spacing and sorting seem to be causing it to generate signatures differently from what Apollo is generating for our Operation IDs.

TLDR: Is there a way for us to generate a json file that maps the operation IDs to the sourceWithFragments with the new compiler?

martinbonnin:
You can look into `transformedQueries`: https://github.com/apollographql/apollo-android/issues/1597

If I understand correctly, this was added specifically for this use case.

jaregier:
We tried with the transformed queries, and while they do output the actual queries that Apollo uses under the hood, there's no Operation ID output to go with them (at least from what I can see).  We were hoping we could use those queries and run them through Apollo CLI to get the Operation ID, but the CLI seems to have slightly different logic before it generates the signature (different whitespace between the two and it also seems to be sorting the fragments), so when the sha256 is run on them, they don't match.

@DSteve595 Are you approaching persisted queries in a similar way?  If so, did you run into this same hash problem?

DSteve595:
In our case, the backend provides operation IDs for given queries. We ignore the operation ID created by the compiler, for the reasons you mentioned. Is your backend able to provide them?

martinbonnin:
@DSteve595 , if the server returns a serverID, somehow the client needs to remember this and map from operationId to serverId, right ? Or a way to go from MyQuery to serverId ? I can see how someone could just use apollo-runtime and just strip the operation while using OperationID in all cases.

[...]

PR #1763, opened 2019-11-12
[Transformed Queries]: make them match Operation.queryDocument() exactly

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/1748#issuecomment-552508744 for the background info.

Also, this Pull Request enables the incubating plugin in the composite build by default (samples + integration tests)


sav007:
Not sure that this PR addresses the original issue, as the actual query sent to the server will be minified by `QueryDocumentMinifier.minify`, where `TransformedQueryOutput` uses original query source with fragments, without minification.

https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/com/example/hero_name/TestQuery.kt#L98

If we want to have exact query that is been sent to the server we need to do the same in `TransformedQueryOutput` as well, minify `sourceWithFragments`.

martinbonnin:
@sav007 minification should be done there: https://github.com/apollographql/apollo-android/pull/1763/files#diff-a5ec6e9e52d3c16ef3c048d1ed8cac12R31

sav007:
Agh, my bad missed that.

PR #1764, opened 2019-11-12
[Incubating Plugin] Also compile test variants

martinbonnin:
This adds `generateTest${service}ApolloSources` tasks and makes the compileJava/Kotlin tasks depend on it. This will compile files in:

```
src/test/graphql/com/example/schema.json
src/test/graphql/com/example/query.graphql
```

Also, on non-android projects, adding a sourceSet will also compile the associated graphql files:

```
configurations {
    fooCompile
}

sourceSets {
   foo { // the plugin will create generateFoo${service}ApolloSources
     compileClasspath = fooCompile
   }
}

// You'll need to add apollo-api to the foo configuration
dependencies {
   fooCompile "com.apollographql.apollo:apollo-api:$versions.apollo"
}
```

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/JvmTaskConfigurator.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/JvmTaskConfigurator.kt
@@ -11,7 +11,10 @@ object JvmTaskConfigurator {
   fun getVariants(project: Project): NamedDomainObjectContainer<ApolloVariant> {
     val container = project.container(ApolloVariant::class.java)
 
-    listOf("main", "test").forEach {name ->
+    val javaPlugin = project.convention.getPlugin(JavaPluginConvention::class.java)
+    val sourceSets = javaPlugin.sourceSets
+
+    sourceSets.map { it.name }.forEach {name ->
       val apolloVariant = ApolloVariant(
           name = name,
           sourceSetNames = listOf(name),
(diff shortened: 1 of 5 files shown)


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/JvmTaskConfigurator.kt
```diff
@@ -10,33 +10,42 @@ object JvmTaskConfigurator {
...
   }
 
   fun registerGeneratedDirectory(project: Project, compilationUnit: DefaultCompilationUnit, codegenProvider: TaskProvider<ApolloGenerateSourcesTask>) {
     val javaPlugin = project.convention.getPlugin(JavaPluginConvention::class.java)
     val sourceSets = javaPlugin.sourceSets
-    val name = compilationUnit.variantName
+    val sourceSetName = compilationUnit.variantName
+
+    val taskName = when(compilationUnit.variantName) {
+      "main" -> ""
```
tasomaniac: Just double checking. Main name is omit only when there is a single source set. If there are multiple, main is used explicitly. I.e compileMainKotlin

martinbonnin: Thanks for the heads up :+1: , I'll add a test and patch this.

tasomaniac: I'm 90% sure but please verify this with tests beforehand :)

martinbonnin: Looks like it's still called `compileJava`/`compileKotlin`... I'll try to make a small sample repo

martinbonnin: I added the check to one of the tests: https://github.com/apollographql/apollo-android/pull/1764/files#diff-b308f8b88ded6b4cfed0ca665d8de6f8R161

PR #1768, opened 2019-11-14
Reorganize documentation

martinbonnin:
Over time, the main README.md has grown a lot. This pull request takes it and moves some of the more advanced topics in sub-pages. Hopefully, that makes the README.md more readable while still keeping the essential information on the main page.

This pull request is mainly about small rewording and moving things around, most of the content is still the same. A few notable changes:

* removed the incubating plugin section as it's already obsolete given the configuration moved to SourceDirectorySets. I'll update it again when we promote the plugin.
* added a sub-page for subscriptions and another one for persisted queries that are mostly empty right now. 

I'm also planning to write a short style guide somewhere so we're consistent on things like whether we use first voice or passive voice and write sample code in java or kotlin.


tasomaniac:
Why not put them under `docs`? Docs website is far better in terms of discoveribility and style. Plus, having `docs` and `doc` next to each other is really confusing

README.md
```diff
@@ -1,118 +1,69 @@
...
 repositories {
     jcenter()
-    maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
 }
 
 dependencies {
-  implementation 'com.apollographql.apollo:apollo-runtime:1.2.2-SNAPSHOT'
+  implementation("com.apollographql.apollo:apollo-runtime:x.y.z")
+  
+  // The code generated by the plugin uses jetbrains annotations, so you will need
```
tasomaniac: I don't think this is a hard requirement. What about something like `you may need to include this as compile time dependency if it has not been already available (i.e Kotlin projects already have it)`?

README.md
```diff
@@ -255,552 +197,41 @@ apollo {
...
-- `NETWORK_ONLY` - Fetch a response from the network only, ignoring any cached responses.
-- `CACHE_FIRST` - Fetch a response from the cache first. If the response doesn't exist or is expired, then fetch a response from the network.
-- `NETWORK_FIRST` - Fetch a response from the network first. If the network fails and the cached response isn't expired, then return cached data instead.
+## Intellij Plugin
 
-For `CACHE_ONLY`, `CACHE_FIRST` and `NETWORK_FIRST` policies you can define the timeout after what cached response is treated as expired and will be evicted from the http cache, `expireAfter(expireTimeout, timeUnit)`.`
+The [JS Graphql Intellij Plugin](https://jimkyndemeyer.github.io/js-graphql-intellij-plugin/) provides auto-completion, error highlighting, and go-to-definition functionality for your graphql files. You can create a [.graphqlconfig](https://jimkyndemeyer.github.io/js-graphql-intellij-plugin/docs/developer-guide#working-with-graphql-endpoints-and-scratch-files) file in order to use GraphQL scratch files to work with your schema outside product code, e.g. by writing temporary queries to test resolvers.
```
tasomaniac: This is amazing information. We have been using this for a while. Thanks for adding it 🙌

tasomaniac:
Once this is merged, I will add some more info into `RxJava` page about the new Kotlin extensions usage.

martinbonnin:
> Why not put them under docs? Docs website is far better in terms of discoveribility and style. Plus, having docs and doc next to each other is really confusing

Good point. I mainly wanted to keep this PR (relatively) short and only about reorganization while keeping everything hosted on github. I'm a bit reluctant to move the files over to `docs` because:

* doing this would splits information between github (README.md) and www.apollographql.com (advanced topics) and I would much rather have a single source of truth.
* I know nothing about how `docs` work and how to run this on my local machine (but I'll definitely give it a shot)
* Ultimately, I'd like to deploy javadoc and I was thinking about doing that in github pages. We could then have everything in github pages  with mkdocs, ala [okhttp](https://square.github.io/okhttp/).

If the markdown used by docs (https://docs-docs.netlify.com/docs/docs/) is compatible with github markdown, I was hoping we could easily change if needed ?

README.md
martinbonnin: Wansn't sure about that one, I changed the wording a bit.

README.md
martinbonnin: It was there before actually, just burried in the 800+ lines of the README.md

designatednerd:
@martinbonnin It's definitely compatible with GFM, I use it for all the iOS docs.

In order to run the docs generator locally, you `cd` into `docs` then run `npm install` to get the assorted dependencies, and `npm start` to start the server. You then get a `localhost:8000` preview of your docs, and when you save something in `docs` the preview regenerates. 

I'd love to get the Android docs to a point where we can get them integrated with the main docsets on the Apollo page, but I definitely understand wanting to get them organized first. 

Let me know if y'all need any help!

designatednerd:
@martinbonnin worth shoving these under `docs` just to see how things go with the doc generator? Or would you prefer to do that in a separate PR?

tasomaniac:
Just like the README index, the sidebar index may need to be changed accordingly. It's pretty easy tho: https://github.com/apollographql/apollo-android/blob/master/docs/gatsby-config.js#L14

PR #1784, opened 2019-11-18
Extract apollo-api-core module common for both JVM and Kotlin models

sav007:
Introduce new `apollo-api-core` module, used by both JVM and Kotlin generated models.
Convert classes extracted from `apollo-api` to Kotlin and fix code generation issues related to the nullability type declarations.

With a great help from https://github.com/apollographql/apollo-android/pull/1504, thanks to @tasomaniac 

Another step towards KN.


sav007:
Sorry for such large PR, but there is no other way to do this migration.

martinbonnin:
Nice ! 

About naming: Shouldn't `apollo-api-core` be the real `apollo-api` and `apollo-api` moved to `apollo-api-java` ? And maybe move `Logger ` and `HttpCache` to separate Kotlin modules ?

Also I was thinking about adding some japicmp to make sure the binary compatibility is preserved while moving to kotlin. I don't know if that can be made retroactively ?

apollo-api-core/src/main/kotlin/com/apollographql/apollo/api/CustomTypeAdapter.kt
```diff
@@ -0,0 +1,62 @@
...
+ *  val builder = ApolloClient.builder()
+ *    .addCustomTypeAdapter(CustomType.DATE, dateCustomTypeAdapter)
+ * ```
+ *
+ * In the example code above, the `CustomType.DATE` is the class generated by the compiler.
+ */
+interface CustomTypeAdapter<T> {
+
+  /**
+   * De-serializes the value to the custom scalar type. Usually used in parsing the GraphQL response.
```
tasomaniac: In KDoc, Jetbrains usually make the params and return the part of the sentence and do not include separate `@param / @return` section. Something like

```
De-serializes the [value] and return the custom scalar type [T]. Usually used in parsing the GraphQL response
```

martinbonnin:
Should we close this one now that a lot of `apollo-api` is in kotlin ?

tasomaniac:
Probably yes.

sav007:
Yeah closing, this PR was used just as reference, now it doesn't make any sense.

PR #1788, opened 2019-11-19
[Android Plugin] remove "helper" setters for Properties

martinbonnin:
These setters were called from groovy and were displaying a deprecation warning during build. By removing them, groovy will directly set the property. 

In kotlin it will be either:

```
generateKotlinModels.set(true)
```
or 

```
generateKotlinModels(true)
```

Maybe we could remove the `fun generateKotlinModels(Boolean)` as it's just a short hand to setting the property. I'm not sure what the conventions are (if there are any?)


tasomaniac:
I was recently thinking about this. Would it be possible to introduce a setter as an extension fun in Kotlin. First of all, it will not confuse Groovy and mess with Gradle magic.

In Kotlin you could actually have the DSL to look like `generateKotlinModels = true`

Answering the convention: I've seen plugins using the Property with the set function call (1st example above) Here is one: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/README.md#configuration

martinbonnin:
> Answering the convention: I've seen plugins using the Property with the set function call (1st example above) Here is one: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/README.md#configuration

Indeed, it's also written black on white in the gradle docs: https://docs.gradle.org/current/userguide/lazy_configuration.html#sec:lazy_configuration_faqs

```Avoid simplifying calls like obj.getProperty().get() and obj.getProperty().set(T) in your code by introducing additional getters and setters.```

I just removed all of the "helper" functions. So now everything is a `Property`

tasomaniac:
Well, for Groovy, that's for sure since nice property setter syntax is supported magically. Do you think this statement is also correct for Kotlin?

martinbonnin:
> Well, for Groovy, that's for sure since nice property setter syntax is supported magically. Do you think this statement is also correct for Kotlin?

Kotlin users will have to use `.set()` from my understanding. I think it's also what the AGP started doing ?

tasomaniac:
Well, AGP is not yet optimizing their DSL for Kotlin users.

IMO, it is totally ok to remove these right now to have a consistent API. If we want to make the Kotlin DSL nicer, that can be done additionally.

martinbonnin:
> Well, AGP is not yet optimizing their DSL for Kotlin users.

Who is ? :sweat_smile: 

> IMO, it is totally ok to remove these right now to have a consistent API. If we want to make the Kotlin DSL nicer, that can be done additionally.

Yep. Also since this is the first release of the new plugin, I'd like to have the smallest possible API surface. It'll be way easier to add things if needed than remove things without breaking things there and there.

martinbonnin:
Some more discussion about how to make Kotlin DSL look better there: https://github.com/gradle/gradle/issues/9268

That's a lot of information to take and I'm not sure what the conclusion of that thread is but as I said above, I think it's best to start with the smallest API possible (therefore Kotlin users will have to call `.set`) and see if we can make this better with more usage and feedback.

tasomaniac:
Plus, Gradle team might provide out of the box better DSL support.

PR #1791, opened 2019-11-20
[Android Support] Revive the connectedTests

martinbonnin:
The android connectedTest have been skipped for some time now. 

I'm trying to revive them. They are mainly about the SQLite helper. There were also a couple of checks for `ApolloCallback` that were not compiling anymore and that I removed. We could certainly deprecate and remove `ApolloCallback` now that we have `RxJava2` and `Coroutines` extensions. I don't think a lot of users would rely on this.


martinbonnin:
Tests passed \o/

martinbonnin:
> they could be converted to use Robolectric too.

Yep, I was thinking about that but it looks like SQLiteOpenHelper is not provided by Roboelectric. Maybe if we move everything to androidx but not sure...

tasomaniac:
`androix.sqlite` group is supposed to work well with Robolectric.

martinbonnin:
Great. I created an issue to move to androidx: https://github.com/apollographql/apollo-android/issues/1793

PR #1795, opened 2019-11-22
[Build scripts] Introduce japicmp 

martinbonnin:
While testing the incubating plugin, I realized that https://github.com/apollographql/apollo-android/pull/1759 broke some of the API compatibility by moving classes from the `com.apollographql.apollo.response` package to the `com.apollographql.apollo.api` package.

~I moved the classes back to their original packageName while still keeping them in `apollo-api` so that they are still usable for anyone depending on `apollo-runtime`~ (EDIT: this will come in a separate PR)

Introducing jcmpapi will help detect such cases more easily. This should also help with the kotlin migration (https://github.com/apollographql/apollo-android/pull/1784). 

Feedback welcome. ~If there's no objection, I'll make the build fail on API/ABI differences starting with 1.2.2 (or 1.3 whichever comes first).~

Right now, the way to run japicmp is to do it manually by running `./gradlew japicmp` from the root of the project. I'm looking into ways to automate this. Breaking the build might be a bit overeacting as we have some internal classes that are allowed to break. Also the incubating plugin will most likely break a lot of public api so we'll need to have a way to override japicmp if needed.

I created https://github.com/apollographql/apollo-android/issues/1814 to discuss this further

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,14 +6,34 @@ buildscript {
   repositories {
     maven { url = uri("https://plugins.gradle.org/m2/") }
     google()
+    jcenter()
   }
 
   dependencies {
     classpath(groovy.util.Eval.x(project, "x.dep.android.plugin"))
     classpath(groovy.util.Eval.x(project, "x.dep.gradleErrorpronePlugin"))
+    classpath(groovy.util.Eval.x(project, "x.dep.gradleJapiCmpPlugin"))
     classpath(groovy.util.Eval.x(project, "x.dep.kotlin.plugin"))
   }
 }
+
+fun baselineJar(project: Project, version: String): File {
+  val group = project.property("GROUP")
+  val artifactId = project.property("POM_ARTIFACT_ID")
+  try {
+    val jarFile = "$artifactId-${version}.jar"
+    project.group = "virtual_group_for_japicmp" // Prevent it from resolving the current version.
+    // see https://github.com/apollographql/apollo-android/issues/1325
+    project.repositories.maven("https://dl.bintray.com/apollographql/android")
+    val dependency = project.dependencies.create("$group:$artifactId:$version@jar")
+    return project.configurations.detachedConfiguration(dependency).files
+        .first { (it.name == jarFile) }
+  } finally {
+    project.group = group!!
+  }
+}
+
+val rootJapiCmp = tasks.register("japicmp")
 
 subprojects {
   apply {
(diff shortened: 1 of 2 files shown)


martinbonnin:
Of course, another option is to introduce a breaking change. We very well do a big 2.0 release with:

* kotlin-stdlib dependency
* incubating plugin
* `apollo-api` usable without `apollo-runtime`

martinbonnin:
@tasomaniac sorry for the PR dance, I removed the package renaming from this PR as it had so many conflicts it's going to be easier to make it a separate one. This also makes this pull request only about japicmp. I guess more atomic PRs is better. I'll open a new PR for the package renaming in a bit.

build.gradle.kts
```diff
@@ -46,6 +66,24 @@ subprojects {
     resolutionStrategy.force(groovy.util.Eval.x(this@subprojects, "x.dep.errorProneCore"))
   }
 
+  // TODO: Make this lazy
+  this@subprojects.afterEvaluate {
+    val jarTask = this@subprojects.tasks.findByName("jar") as? org.gradle.jvm.tasks.Jar
```
tasomaniac: You could remove afterEvaluate by doing something like this but it will not be lazy since it will configure the jar task, regardless. 

```
this@subprojects.tasks.matching { it.name == "jar" }.all { jarTask ->
}

```

martinbonnin: I actually tried something similar (with tasks.withType instead of tasks.matching) and it did not got well :-/. It failed on:

```
* What went wrong:
A problem occurred configuring project ':apollo-android-support'.
> Could not create task ':apollo-android-support:androidJavadocsJar'.
   > DefaultTaskContainer#register(String, Class, Action) on task set cannot be executed in the current context.
```

I'm not sure what's happening exactly. Maybe creating a task from another task configuration isn't allowed. That wouldn't surprise me too much.

tasomaniac: Oh yeah, I hit this before. It is not allowed. As far as I learned the solution is to create dummy tasks that just depends on the real task based on `jar` being available. 

So you always create the task but only run if conditions are met.

martinbonnin: I must be missing something, even calling `dependsOn` is not allowed it seems: https://github.com/martinbonnin/apollo-android/blob/94ae0e40777372218fe3e4fb5f2ccfcc0578ee33/build.gradle.kts#L87

martinbonnin: Ok, I needed to do everything from `all` and not `configureEach`: https://github.com/martinbonnin/apollo-android/commit/9a143e118272c23e514a7bcedcd03f0706064f97

martinbonnin: Ok so using `all` doesn't work as it's eager so it's actually worse than everything else as it forces the configuration of *all* the tasks...

tasomaniac: `findByName` would also force the configuration of all the tasks, no?

martinbonnin: > findByName would also force the configuration of all the tasks, no?

Indeed. The problem is that it forces configuring the actual Task which assumes a `jar` task is present. Oh well, I guess I can skip configuration in that case...

martinbonnin: But... since `all` forces the configuration of the `actualJapicmp`task, it gets configured before the `jar` task is added by the gradle-mvn-push script and we're back to `project.afterEvaluate {}`...

PR #1796, opened 2019-11-22
Parse generated GraphQL operation models from okio source

sav007:
Parse GraphQL operation response from `okio.BufferedSource` rather than from raw map.

- Move classes related to parsing JSON to api module
- Generate API to parse operation response from `okio.BufferedSource`


apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
```diff
@@ -44,14 +46,14 @@
    */
   @NotNull String operationId();
 
-  /**
+    /**
```
tasomaniac: Accidental?

PR #1802, opened 2019-11-25
Provide API to serialize GraphQL operation variables as JSON string

sav007:
This is the final piece  to make possible usage of Apollo generated models without runtime.

- Move supporting files to work with JSON into `apollo-api` module
- Introduce new API `Operation.Variable#marshal` to serialize variables as JSON string
- Update docs

Another step towards KN.

### DIFF
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
@@ -107,6 +107,16 @@ public interface Operation<D extends Operation.Data, T, V extends Operation.Vari
     /**
      * Serializes variables as JSON string to be sent to the GraphQL server.
      *
+     * @return JSON string
+     * @throws IOException
+     */
+    public final String marshal() throws IOException {
+      return marshal(ScalarTypeAdapters.DEFAULT);
+    }
+
+    /**
+     * Serializes variables as JSON string to be sent to the GraphQL server.
+     *
      * @param scalarTypeAdapters adapters for custom GraphQL scalar types
      * @return JSON string
      * @throws IOException
--- a/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
+++ b/apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.api;
 
+import com.apollographql.apollo.api.internal.json.InputFieldJsonWriter;
+import com.apollographql.apollo.api.internal.json.JsonWriter;
+import okio.Buffer;
 import okio.BufferedSource;
 import org.jetbrains.annotations.NotNull;
 
@@ -49,7 +52,7 @@ public interface Operation<D extends Operation.Data, T, V extends Operation.Vari
   /**
    * Parses provided GraphQL operation raw response
    *
-   * @param source for operation raw response to parse
+   * @param source             for operation raw response to parse
    * @param scalarTypeAdapters configured instance of custom GraphQL scalar type adapters
    * @return parsed GraphQL operation {@link Response}
    */
@@ -99,6 +102,24 @@ public interface Operation<D extends Operation.Data, T, V extends Operation.Vari
         @Override public void marshal(InputFieldWriter writer) {
         }
       };
+    }
+
+    /**
+     * Serializes variables as JSON string to be sent to the GraphQL server.
+     *
+     * @param scalarTypeAdapters adapters for custom GraphQL scalar types
+     * @return JSON string
+     * @throws IOException
+     */
+    public final String marshal(@NotNull final ScalarTypeAdapters scalarTypeAdapters) throws IOException {
+      final Buffer buffer = new Buffer();
+      final JsonWriter jsonWriter = JsonWriter.of(buffer);
+      jsonWriter.setSerializeNulls(true);
+      jsonWriter.beginObject();
+      marshaller().marshal(new InputFieldJsonWriter(jsonWriter, scalarTypeAdapters));
+      jsonWriter.endObject();
+      jsonWriter.close();
+      return buffer.readUtf8();
     }
   }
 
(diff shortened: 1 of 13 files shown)


apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
```diff
@@ -100,6 +103,24 @@ protected Variables() {
...
     }
+
+    /**
+     * Serializes variables as JSON string to be sent to the GraphQL server.
+     *
+     * @param scalarTypeAdapters adapters for custom GraphQL scalar types
+     * @return JSON string
+     * @throws IOException
+     */
+    public final String marshal(@NotNull final ScalarTypeAdapters scalarTypeAdapters) throws IOException {
```
tasomaniac: Would it make sense to have an overloaded function without the scalarTypeAdapters? It would make it easier for those who don't have anything custom.

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
```diff
@@ -49,7 +52,7 @@
   /**
    * Parses provided GraphQL operation raw response
    *
-   * @param source for operation raw response to parse
+   * @param source             for operation raw response to parse
```
martinbonnin: typo ?

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
martinbonnin: If we do this for `Variables.marshal`, we should also certainly do it for `Operation.parse()` ? Another option would be to add a SingleTon empty scalaraTypeAdapters ?  

```  public static ScalarTypeAdapters Empty = new ScalarTypeAdapters(new HashMap<ScalarType, CustomTypeAdapter>())
```

doc/no-runtime.md
```diff
@@ -15,9 +15,23 @@ If for some reason you want to use your own network layer and don't want to use
...
+    final Query query = ...;
+
+    // if you have custom scalar types, provide proper instance of ScalarTypeAdapters with your own custom adapters
+    final ScalarTypeAdapters scalarTypeAdapters = new ScalarTypeAdapters(Collections.<ScalarType, CustomTypeAdapter>emptyMap());
+
+    final String requestPayload = "{" +
+        "\"operationName\": " + query.name().name() + ", " +
+        "\"query\": " + query.queryDocument() + ", " +
+        "\"variables\": " + query.variables().marshal(scalarTypeAdapters) +
+        "}";
```
martinbonnin: Should we add a helper function that does this ? Something like:

```
fun Operation.payload(): String
```

?

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
tasomaniac: That's a good idea. My comment was also just a simple idea and not a blocker for this PR. 

I actually believe it would be better as a follow-up.

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
tasomaniac: I think the Java code style forces param definitions to be on the same column. I noticed this in earlier PRs too. Would be good to align.

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
sav007: sure makes sense, will address this in next PR.

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
sav007: Yeah, it's formatting JavaDocs, will check how to disable this rule.

doc/no-runtime.md
sav007: The initial reason why I introduced marshalling for variables only, is to give dev flexibility on deciding how to send GraphQL request to the server: use POST, GET, send just operatation ID, or bundle with bunch of extension attributes etc.

But I see values of having simple case, as most of the time it will be plain http POST request. 

How about if we introduce new API to the `com.apollographql.apollo.api.Operation` like `payload():String` or `postPayload():String` or  `requestPayload()`. Not sure about how to name it, any thoughts?

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
martinbonnin: Should we switch to dokka anyways ?

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
sav007: Added new API:
```
    public final String marshal() throws IOException {
      return marshal(ScalarTypeAdapters.DEFAULT);
    }
```

For parse it will be in the next PR, as it will involve to update code generation fixtures.

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
sav007: I guess when we migrate these interfaces to Kotlin

doc/no-runtime.md
martinbonnin: I was suggesting `Operation.payload(): String` but maybe we want to keep `Operation` transport agnostic ? 

In that case, I'm not sure where to put it... Maybe we can just wait to see how users are going to use the API before we add some API that is never going to be used in real life 😅

tasomaniac: I see. If there are 3 different usages, I would create utility functions (extensions in Kotlin) decoupled from the actual implementation to provide such extra nice to have functionality. 

This could be separate class (maybe in separate package, maybe not). It could even be separate artifact

sav007: Ok lets leave this as is for now, and we can improve later when we migrate everything to Kotlin.

Issue #1722, opened 2019-11-01
Isn't cache supposed to work with GraphQL Subscriptions?


yankeppey:
I found that Normalized Cache does not work with GraphQl Subscriptions (opened via websockets), and I didn't find anything in the apollo-android code that would prove the opposite.

Could you confirm it and probably clarify the reasons why it was designed so?

I found a workaround for that but this should probably be done by default:
```
Rx2Apollo.from(call)
            .flatMapSingle((Response<D> response) -> Rx2Apollo
                .from(apolloClient.apolloStore().writeAndPublish(subscription, response.data()))
                .onErrorReturnItem(false) // no need to cancel subscription because of the failed cache update
                .map($ -> response))
```

Thanks!

sav007:
Currently subscriptions doesn't support normalized cache. Not sure if it make sense to add such support as subscriptions represents live queries via web socket connection.

yankeppey:
Thanks for the confirmation! 
IMHO it makes perfect sense to utilize normalized cache here, as it's the most common use case of websockets: delivery of some updates initiated by the backend to the app.
I'm not sure what to do if CacheKeyResolver produces `NO_CACHE_KEY`, but for the data which we can calculate cache key for (using `id`  or smth), why wouldn't we want to put it into the cache?

sav007:
hm, maybe you are right.

So you want to be able get cached response for the subscription before running a subscription?

wtrocki:
@sav007 There is a pattern that teams widely using. Generally focus is to use cache as the source of the data and use subscriptions to refill that data. More about this approach in this blog:
https://medium.com/@wtr/data-query-patterns-for-graphql-clients-af66830531aa#1150

In Apollo-client (js) users can simply use subscribeToMore to get more data

wtrocki:
@sav007 We have investigated having interceptor layer attached to subscriptions. Are you open to accept Pull requests on this or you do think that this is bad idea?

I have seen so many workarounds for lack of cache on subscriptions and I think that there should be at least the ability to attach this extra interceptor for developers who would like to use cache as a source of truth. Without that, it is hard to work with cache enabled app. 

Let's imagine that we have a query cached on some separate activity. Now we subscribed to receive new results that are pushed directly to UI (which is Kinda antipattern on it's own). Then we go back to previous view and go back to original view again. Since results are in the cache we get old results. Most of the developers at this point are really confused and simply do:

[...]

PR #1804, opened 2019-11-26
Add support of normalized cache to subscriptions

sav007:
Enable and provide new API to set subscription normalized cache policies:
- `NO_CACHE` bypass cache, don't fetch or cache from / to normalized cache
- `NETWORK_ONLY` always fetch from network but cache response to normalized cache
- `CACHE_AND_NETWORK` fetch from cache first then from network

Closes https://github.com/apollographql/apollo-android/issues/1722

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -9,6 +9,7 @@ import com.apollographql.apollo.api.Subscription;
 import com.apollographql.apollo.api.cache.http.HttpCache;
 import com.apollographql.apollo.api.cache.http.HttpCachePolicy;
 import com.apollographql.apollo.api.internal.Optional;
+import com.apollographql.apollo.api.internal.Supplier;
 import com.apollographql.apollo.cache.CacheHeaders;
 import com.apollographql.apollo.cache.normalized.ApolloStore;
 import com.apollographql.apollo.cache.normalized.ApolloStoreOperation;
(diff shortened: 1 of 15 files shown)


sav007:
ping @martinbonnin @tasomaniac

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -615,8 +622,14 @@ public ApolloClient build() {
       SubscriptionManager subscriptionManager = new NoOpSubscriptionManager();
       Optional<SubscriptionTransport.Factory> subscriptionTransportFactory = this.subscriptionTransportFactory;
       if (subscriptionTransportFactory.isPresent()) {
+        final ApolloStore finalApolloStore = apolloStore;
         subscriptionManager = new RealSubscriptionManager(scalarTypeAdapters, subscriptionTransportFactory.get(),
-            subscriptionConnectionParams, dispatcher, subscriptionHeartbeatTimeout);
+            subscriptionConnectionParams, dispatcher, subscriptionHeartbeatTimeout,
+            new Supplier<ResponseNormalizer<Map<String, Object>>>() {
+              @Override public ResponseNormalizer<Map<String, Object>> get() {
+                return finalApolloStore.networkResponseNormalizer();
```
martinbonnin: Why does this have to be lazy ?

sav007: The issue that instance of `ResponseNormalizer` is stateful, means we need to create a new instance everytime we parse response. So it's not lazy but rather a factory to be called whenever we parse new response from the server.

PR #1815, opened 2019-11-29
[Runtime] Unbreak api compatibility

martinbonnin:
Move JsonDataException, JsonEncodingException, CustomTypeAdapter, CustomTypeValue and ScalarTypeAdapters back to their original package name


martinbonnin:
@sav007 @tasomaniac can you check this one quickly so that we can merge it before there are conflicts all over the place again ?

martinbonnin:
There's no rush for a new version as the package renaming change has never been released. What I would do is:
- 1.2.2 with this PR
- promote incubating plugin just after 1.2.2
- make a 1.3 with the incubating plugin a bit later. That leaves a few days with the incubating plugin in master, feels safer.

tasomaniac:
This plan sounds good.

Issue #1800, opened 2019-11-25
How to get access to raw JSON of response? [question]


NatElkins:
Currently I'm using generated Kotlin classes to access a GraphQL API.  At one point, I need to get access to the response type in JSON form.  If I used Jackson to serialize it to JSON, it usually works just fine, unless I'm using fragments.  If I'm using fragments, I get something similar to the following:

```json
{
        "__typename": "Channel",
        "fragments": {
            "commonChannelInfo": {
                "__typename": "Channel",
                "id": "96895296-a3ec-45c3-bc85-2ff229cb5fde",
                "name": "#lips",
                "description": "Crowd favorites sealed with a kiss 💋",
                "facets": [
                    "Makeup"
                ],
                "subscription_status": false,
                "answered_users": [
                    {
                        "__typename": "User",
                        "avatar": "329b29b8-cae9-4c0d-9f01-3c36ad3b4fe2"
                    },
                    {
                        "__typename": "User",
                        "avatar": "329b29b8-cae9-4c0d-9f01-3c36ad3b4fe2"
                    },
                    {
                        "__typename": "User",
                        "avatar": "329b29b8-cae9-4c0d-9f01-3c36ad3b4fe2"
                    },
                    {
                        "__typename": "User",
                        "avatar": "329b29b8-cae9-4c0d-9f01-3c36ad3b4fe2"
                    }
                ],
                "stats": {
                    "__typename": "ChannelStats",
                    "question": 647,
                    "answer": 1525,
                    "subscription": 610
                }
            }
        }
    }
```

If I were using "normal" GraphQL, the fragments property wouldn't be there.  Is there a way to get access to what the GraphQL is responding with directly?  Or alternatively, is there a known workaround?

Thanks!

martinbonnin:
If you really need, you can access the raw json from an okhttp interceptor. But apollo-android should give you access to all the data from your queries using a typesafe API, which is usually better than using Jackson or something else to parse json. Is there a specific use case where you need access to the json from some data that is not in your query ? 

See also https://github.com/apollographql/apollo-android/issues/1771

NatElkins:
@martinbonnin Thanks for the tip, I'll give that a try.  It's not about the data per se, it's about the shape of the data, I have a need to have it represented as JSON (in the same way that it gets returned by GraphQL if you were to access it directly).  Other times it's useful to have it in the type safe API, because I need to apply some kind of transofm

[...]

Issue #1771, opened 2019-11-15
Support return raw JSON


minhbnt-1512:
Apollo Android Client returns classes that are auto-generated by apollo. This is weird because after query I have to map the auto-generated classes (Ex: GetUserQuery.Data) to the data model (Ex: UserData). What can I do to the query return raw JSON instead of auto-generated classes? With JSON, I can parse it to the data model easier.

martinbonnin:
@minhbnt-1512 this is actually one of the main advantages of using `apollo-android`. Using the generated models guarantee that they match your graphql schema (nullable/nonnullable types and type safety in general). 

If you don't want that, I guess you could simply use a http client like OkHttp but you lose a lot of benefits of using `apollo-android`

wulf0r:
I agree that using generated classes to have the responses well-typed is exactly what want apollo to do. 

However, I think making the raw response json available is still needed.

My usecase is that Shopify's GraphQL  returns "extension" fields for every request and these fields needed to deal with server-side throtteling. Ex:

```
    "extensions": {
        "cost": {
          "requestedQueryCost": 1,
          "actualQueryCost": 1,
          "throttleStatus": {
             "maximumAvailable": 1000,
             "currentlyAvailable": 999,
             "restoreRate": 50
           }
        }
    }


I can't include this in my queries as it will give an error so I will just have to get via the raw response

martinbonnin:
I see. Could you somehow could use an interceptor (either okhttp or apollo) for this ?

wulf0r:
Yes, that's how I ended up solving it :) Of course this code needs the approtriate data classes to work but this should get anyone started that has the same issue:

     OkHttpClient.Builder().addInterceptor {chain ->
            val original = chain.request()
            val builder = original.newBuilder().method(original.method(), original.body())
            builder.header("X-Shopify-Access-Token", accessToken )
            var response = chain.proceed(builder.build())
            val body = response.body() ?: return@addInterceptor response
            do {
                val bodyText = response.peekBody(Long.MAX_VALUE).charStream().readText()
                val json = Gson().fromJson(bodyText,ResponseJson::class.java)
                if( json.errors?.any { it.isThrottledError() } == true) {
                    with(json.extensions.cost) {
                        val neededAvailable = requestedQueryCost - throttleStatus.currentlyAvailable
                        val timeToWaitInMS = min(floor(neededAvailable / throttleStatus.restoreRate) * 1000, 500.0)
[...]

PR #1817, opened 2019-12-01
Add support for GraphQL extensions

sav007:
According to the spec https://github.com/graphql/graphql-spec/blob/master/spec/Section%207%20--%20Response.md#response-format response map may also contain an entry with key extensions.

Closes https://github.com/apollographql/apollo-android/issues/1771
Closes https://github.com/apollographql/apollo-android/issues/1800

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -66,22 +66,20 @@ subprojects {
     resolutionStrategy.force(groovy.util.Eval.x(this@subprojects, "x.dep.errorProneCore"))
   }
 
-  if (name != "apollo-compiler" && name != "apollo-gradle-plugin-incubating" && name != "apollo-coroutines-support") {
-    // TODO: Make this lazy
-    this@subprojects.afterEvaluate {
-      val jarTask = this@subprojects.tasks.findByName("jar") as? org.gradle.jvm.tasks.Jar
-      if (jarTask != null) {
-        val japiCmp = this@subprojects.tasks.register("japicmp", me.champeau.gradle.japicmp.JapicmpTask::class.java) {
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -66,20 +66,22 @@ subprojects {
...
-        oldClasspath = this@subprojects.files(baselineJar(this@subprojects, "1.2.1"))
-        newClasspath = this@subprojects.files(jarTask.archiveFile)
-        ignoreMissingClasses = true
-        packageExcludes = listOf("*.internal*")
-        onlyModified = true
-        txtOutputFile = this@subprojects.file("$buildDir/reports/japi.txt")
-      }
-      rootJapiCmp.configure {
-        dependsOn(japiCmp)
+  if (name != "apollo-compiler" && name != "apollo-gradle-plugin-incubating" && name != "apollo-coroutines-support") {
```
sav007: I had to disable registration of `japicmp` task for these modules, otherwise build fails with this error:
```
Execution failed for task ':apollo-android:apollo-compiler:compileKotlin'.
> Could not create task ':apollo-android:apollo-compiler:japicmp'.
   > Collection contains no element matching the predicate.
```
locally (weird that at the same time it seems its ok on CI).

tasomaniac: Can you run from command line with --stacktrace option with more details?

martinbonnin: I've seen this during development but it's working well here, even on this branch. Maybe it's due to the publishing configuration that creates some extra 'jar' tasks if you have things like BINTRAY_USER defined ?

Issue #1605, opened 2019-09-20
Modify generated code visibility in Kotlin


NitroG42:
It would be useful to be able to change default visibility of apollo generated class to be able to set *internal* visibility so using apollo in a child module of a project won't "leak" class in modules that import it. 
It would affect Fragment, Mutation, Query, etc...

I don't think i would be useful in java (or maybe be able to choose package or public visibility).

It should be a Gradle Configuration like 
```
apollo {
  generateKotlinModels = true
  kotlinVisibility = "internal"
}
```
It should changes all generated code visibility, to avoid difficult implementation.

It would be useful for my current project but maybe I'm the only one, what do you think ?

sav007:
That seems like reasonable feature request.

PR #1818, opened 2019-12-01
Add configuration to generate Kotlin models with internal modifier

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1605

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -36,7 +36,8 @@ class GraphQLCompiler {
           ir = ir,
           customTypeMap = args.customTypeMap,
           useSemanticNaming = args.useSemanticNaming,
-          packageNameProvider = args.packageNameProvider
+          packageNameProvider = args.packageNameProvider,
+          generateAsInternal = args.generateAsInternal
       ).write(args.outputDir)
     } else {
       ir.writeJavaFiles(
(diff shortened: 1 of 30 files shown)


tasomaniac:
I am thinking now whether if we should segregate Kotlin only configuration extensions vs others.

martinbonnin:
> I am thinking now whether if we should segregate Kotlin only configuration extensions vs others in the incubating plug-in

At some point I was contemplating have 2 pluginIds: `com.apollographql.apollo.java` and `com.apollographql.apollo.kotlin`. Not sure if it's really worth it or if it will bring confusion...

tasomaniac:
2 separate plugins may mean also code duplication. I'm not sure. Somehow separation would be useful tho. Right now, it is getting confusing too. Like this internal modifier is available to Java users. They don't need it.

martinbonnin:
> 2 separate plugins may mean also code duplication. 

Not necessarily. We can use the same jar, 2 pluginIds and reuse as much of the plugin code as we want.
* We'll know even before `Plugin.apply()` what kind of source file is going to be generated => no more need for `afterEvaluate` to setup the sourceSets
* Only expose in the gradle configuration the parameters that are useful for the targetted language.

> Somehow separation would be useful tho. Right now, it is getting confusing too

Agreed. We could do something like:

```kotlin
apollo {
    outputPackageName.set("com.example")
    kotlinOptions {
       generateAsInternal.set(true)
    }
    javaOptions {
       nullableValueType.set("annotated")
    }
}
```

> Like this internal modifier is available to Java users

It's caught at runtime at the moment with checks like this:
https://github.com/martinbonnin/apollo-android/blob/bea404bc589e298715c8aad3ea5d0675165b3181/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt#L123
Agreed compile time would be better.

PR #1819, opened 2019-12-01
Added release scripts and allow Travis to upload to bintray

martinbonnin:
⚠️ Once this PR merged, Travis will upload to bintray automatically each time there is a tag. (at least theoretically...)

The workflow is:

1. The maintainer runs `./scripts/release.kts` from his local dev machine.
2. The script removes the `-SNAPSHOT` suffix, commits and tags version `v1.x.y`
3. The script bumps version to the next patch version and commits
4. The maintainer pushes master (this could be automated in a future PR but I prefer to keep a manual step for the time being)
5. Travis will build tag `v1.x.y` and `scripts/deploy.sh` will see it's from a tag and deploy to bintray with the encrypted credentials from `.travis.yml`

This means anyone with write access to the repo can trigger a new release. It's also safer because releases are built from CI in a reproducible environment.


sav007:
Am I right, in case of manual release (without using `release.kts` script) the steps are following:
1. Create new release branch 
2. Remove `-SNAPSHOT` suffix, commit
3. Create tag
4. Bump version and add suffix back, commit
5. Push tag and branch

martinbonnin:
>Am I right, in case of manual release (without using release.kts script) the steps are following:

Mostly yes, just unsure about release branches. The release.kts script works directly on master. This could also be done using branches and pull requests but it'll be a bit more involved.

Issue #1823, opened 2019-12-03
The apollo-api pom file does not pull okio as a dependency


martinbonnin:
In the latest version, we added a version of `Operation.parse` that operates on a `Okio.Source`

`okio` is correctly marked as an `api` dependency of `apollo-api` but it does not appear in the resulting pom file: 
https://bintray.com/apollographql/android/download_file?file_path=com%2Fapollographql%2Fapollo%2Fapollo-api%2F1.2.2%2Fapollo-api-1.2.2.pom

We might want to move to the `maven-publish` plugin, not sure how easy that is.

PR #1824, opened 2019-12-03
[Build scripts] Update maven publishing

martinbonnin:
This is a tentative to bring the publish scripts to Kotlin and to the new`maven-publish` plugin.

The new tasks should be:

* `publishMavenPublicationToMavenLocal` to publish to ~/.m2
* `publishMavenPublicationToUnitTestRepository` to publish to $rootProject/build/localMaven for plugin unit tests.
* `publishMavenPublicationToOssRepository` to publish snapshots
* `bintrayUpload` doesn't change

This should hopefully fix https://github.com/apollographql/apollo-android/issues/1823 although I haven't tested since it requires a bintray upload.


martinbonnin:
> I wonder if it is possible to publish a SNAPSHOT version to try it out.

Same here 🙋‍♂🙂 . Every commit on master pushes to sonatype so we'll see quite fast if the SNAPSHOTS still work...

martinbonnin:
Looks liked it worked well: https://travis-ci.org/apollographql/apollo-android/builds/622706218

Is there a way to verify the upload time of a snapshot ?

Issue #1686, opened 2019-10-17
singularization of names ending with a


jporcelli-tt:
Is there a flag to not have the kotlin code gen change classes that end with a to end with um?

martinbonnin:
You have `apollo.useSemanticNaming`. Is this what you are looking for ?

jporcelli-tt:
yes that would be nice. I have classes that end with a, e.g. Cta, and there getting renamed to Ctum

sav007:
Yeah the issue that it's very hard to handle this case, our rule for your case is: `"([ti])a$" to "$1um"`

https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt#L31

Not really sure how to handle this as `Cta` is not a word even. You can always try to use alias with different name to avoid this.

jporcelli-tt:
Other people I work with on iOS have had similar complaints for different words, naming. I think making this optional would be best. Is there a reason that this is necessary to do?

martinbonnin:
@jporcelli-tt Do you still have the problem on 1.3.0 ?

sav007:
Closing, out of date.

PR #1829, opened 2019-12-05
Add customizable singularization rules

EricKuck:
Since updating to the kotlin codegen, we've run into lots of issues with bizarre singularization in our project. Examples include `Pies` -> `Py`, `Ctas` -> `Ctum`, and `Slice` -> `Slouses`. Another user opened an issue to report this problem a while back: #1686.

This PR addresses the reality that it is impossible to accurately model the English language in a project like this. Additionally, any fixes to the Singularization rules would end up breaking existing codebases, which isn't ideal. Rather than adding crazy complicated NLP features to the library (which would still have issues), I opted to instead allow users to define custom singularization rules at the project level. These custom rules will take priority over any of the singularization rules built into the plugin. Example usage from our project based on this fork follows:

```
    apollo {
        generateKotlinModels = true
        customSingularizationRules = [
            "(slice)\$": "\$1",
            "(pie)s\$": "\$1",
            "(cta)s\$": "\$1",
        ]
    }
```


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -12,11 +12,13 @@ class GraphQLCompiler {
   fun write(args: Arguments) {
     val ir = args.ir
     val customTypeMap = args.customTypeMap.supportedTypeMap(ir.typesUsed)
+    val customSingularizationRules = args.customSingularizationRules.map { Pair(it.key, it.value) }.toHashSet()
```
sav007: just a curious why not to leave it as Map but convert to hash set? As it looks like with map it will be easier to work later in `String.singularize`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt
```diff
@@ -9,7 +8,12 @@ import java.util.regex.Pattern
  * The singularization methods were heavily based off rogueweb <link> https://code.google.com/archive/p/rogueweb/source</link>
  */
 
-fun String.singularize(): String {
+fun String.singularize(customMappings: Set<Pair<String, String>>): String {
+  val customMapping = customMappings.lastOrNull { match(it.component1(), this) }
```
sav007: nit: you can use destructuring here:
```
val (pattern, value) = customMappings....
```

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
```diff
@@ -72,6 +72,28 @@ class UpToDateTests {
     File(dir, "build.gradle").writeText(text.replace(apolloBlock, ""))
   }
 
+  //TODO: finish this!
```
sav007: is this still WIP and  this TODO will be addressed later in this PR?

apollo-gradle-plugin/src/test/groovy/com/apollographql/apollo/gradle/integration/BasicAndroidSpec.groovy
```diff
@@ -188,6 +188,30 @@ class BasicAndroidSpec extends Specification {
         "return Currency.class;")
   }
 
+  def "adding a custom singularization rule to the build script re-generates the CustomType class"() {
+    setup: "a testProject with a previous build and an apollo extension appended"
+    replaceTextInFile(new File("$testProjectDir/build.gradle")) {
+      it.replace("apollo {", "apollo {\n customSingularizationRules=[\"F(ilm)s\\\$\" : \"SingularF\\\$1\"]\n")
+    }
+
+    System.err.println(new File("$testProjectDir/build.gradle").getText('UTF-8'))
```
sav007: nit: looks like should be removed?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
EricKuck: Cause I did a bunch of refactoring and forgot to change it back to a Map! Great catch, thanks.

EricKuck: Cause I did a bunch of refactoring and forgot to change it back to a Map! Great catch, thanks.

EricKuck: Cause I did a bunch of refactoring and forgot to change it back to a Map! Great catch, thanks.

apollo-gradle-plugin/src/test/groovy/com/apollographql/apollo/gradle/integration/BasicAndroidSpec.groovy
EricKuck: Oof...yes, will remove, sorry about that

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
EricKuck: Not a WIP anymore, just a sloppy commit. Thanks.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
EricKuck: Actually putting to back to a Map made the `singularize` method a bit more confusing, as you have to iterate the keys, then use that to fetch the matching value (which is nullable). Happy to change it if you want, but personally I think the logic for `Set<Pair<String, String>>` ends up being a bit clearer.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt
EricKuck: Destructuring would work great with `last`, but unfortunately it doesn't work so well with `lastOrNull`.

EricKuck:
Just pushed a commit with debug things removed. Happy to change the other things if desired as well.

martinbonnin:
Sorry I'm a bit late to the party but why not disabling singularization altogether ? Is there a specific reason it is needed ?

EricKuck:
I don’t really know why it’s included, but since a fair amount of work went into singularization on both iOS and Android, it seemed like it was a key feature of the library. If disabling it altogether is the preferred method of fixing this problem I’m happy to open another PR that does that instead.

martinbonnin:
ping @sav007. Wouldn't it be simpler to allow disabling singularization altogether ?

EricKuck:
While I’m still happy to open a different PR to disable singularization altogether, it seems to me that there are two possible scenarios here, neither of which would be addressed by making it disableable.

1) Singularization is a useful feature of the library, in which case it would be unfortunate if people had to opt out if it entirely due to a handful of mistakes it makes.

2) Singularization isn’t useful, in which case it should be removed from the library, or at least off by default.

Given that someone spent a decent amount of time adding singularization, it seems as though #1 is believed to be the case. If this is correct, shouldn’t we work to make it accessible to everyone instead of telling people they can’t use it if their schema contains models with names it doesn’t work with?

martinbonnin:
I'm wondering if we're not victims of the sunk cost fallacy here. If we're in case #1, there surely was a reason but I'd like to find out before we invest even more and make any further decision even more complicated.

designatednerd:
So I took some time to talk to other client devs at Apollo about this and the much much much easier method of doing this is using aliases when requesting fields, then using the alias to generate the customized field name. 

Y’all are very correct about trying to singularize things ourselves being error prone and getting into some NLP stuff that’s waaaay out of scope. 

Recommendation would be to deprecate and later remove this feature in favor of aliases on individual fields going forward, particularly since this gives developers more control about when this is applied and when it’s not.

EricKuck:
The idea of developer-provided aliases makes a lot of sense to me (much better than trying to fix errors after they've popped up!). Given that this issue is creating some super wacky code in our project, I'm fully on board with opening another PR with whatever suggestions we decide on.

Is there consensus for the alias idea in place of singularization? If so, what would the plan for deprecating singularization be? On one hand, removing it right now would break people's codebases. On the other hand, it's fairly buggy and will continue creating issues for people moving from Java codegen to Kotlin codegen until it's gone.

designatednerd:
In terms of how other libraries in the Apollo Comic Universe™ handle it, aliasing instead of trying to do NLP is definitely the strategy we plan to use going forward.

In terms of how to deprecate it in a way that makes sense...uhhhhhhhh.... *sprints away leaving @sav007 and @martinbonnin to answer this question*

martinbonnin:
Regarding deprecation, what about adding `CompilerParams.useSingularization` to the incubating gradle plugin 

```kotlin
  /**
   * Whether to singularize fields
   */
  val useSingularization: Property<Boolean>
```

We could then either:
1. leave it true by default with a warning that it will change in the future.
2. or change it to default to false as soon as 1.3.0 is released.

I think I would prefer 2. if users will need to update their code eventually, the sooner the better. To be nice, we could even check if any type in the schema would be impacted based on the current rules. If there are, fail early with a nice error message like: 

```
singularization is disabled by default since 1.3.0, to keep your code compatible, please add this to your build.gradle file:
apollo {
   useSingularization.set(false)
}
```

EricKuck:
This sounds like a good plan to me.

I'm a little unclear what you mean about failing early though. If we error out on detecting a type that would be affected by the change, the only way to get your code to compile again would be to re-enable singularization.

martinbonnin:
> If we error out on detecting a type that would be affected by the change, the only way to get your code to compile again would be to re-enable singularization.

I was thinking that if we detect an affected type and `compilerParams.useSingularization` is not explicitely set, we throw an Exception. To make the exception go away, users would have to explicitely set `useSingularization`. Thinking about it again, this will also affect new users so it's not ideal but maybe an acceptable middle ground between giving existing users some context about the required changes and ease of use for newcomers ?

Also, I'm not sure how many schemas in general are affected by this. Is this a corner case or would 90% of the schemas contain types affected by singularization ?

sav007:
Hey folks, TBH this singularization is pain for a long time. It works great in very simple cases but but awful in corner cases (I wish this feature was never introduced at first place).

I think introducing new flag `useSingularization` with default value `false` in incubating gradle plugin makes totally sense to me. And I agree  with @martinbonnin we should remove it altogether in future so if user still uses it `useSingularization=true` we can produce warning.

tasomaniac:
I think overall the plan looks good. I just want to give 1 more suggestion with the goal to keep things simple. 

What about removing it all together (without any Gradle flag) but still generate them with deprecated warning? In generated Kotlin code, we can even provide a quick fix for the users and they can fix them just using `Option + Enter`.

EricKuck:
Opened a new PR, which adds the flag for disabling singularization altogether: https://github.com/apollographql/apollo-android/pull/1880

Issue #1825, opened 2019-12-03
Request hangs and callback is not invoked if a custom Call.Factory decorates a Call


steventannz:
**Summary**
All requests hang and callbacks (onError/onResponse) are not invoked if a custom Call.Factory decorates a Call. 
**Description**
Related to the changes made in https://github.com/apollographql/apollo-android/pull/1663 

Reproduced in the Kotlin sample project: https://github.com/steventannz/apollo-android/commit/43211d4130afa575209b535c7e1ead799de8080f

You can reproduce this by adding a custom call factory to the ApolloClient and and wrapping the Call returned by newCall in a decorated version.
```
ApolloClient.builder()
                .serverUrl("url")
                .callFactory {
                  DecoratedCall(client.newCall(it))
                }.build()
```

where client is an okHttp instance and DecoratedCall is 

```    
class DecoratedCall(val call: Call) : Call {
        override fun enqueue(responseCallback: Callback) {
            call.enqueue(responseCallback)
        }

        override fun isExecuted(): Boolean {
            return call.isExecuted
        }

        override fun timeout(): Timeout {
            return call.timeout()
        }

        override fun clone(): Call {
            return call.clone()
        }

        override fun isCanceled(): Boolean {
            return call.isCanceled
        }

        override fun cancel() {
            call.cancel()
        }

        override fun request(): Request {
            return call.request()
        }

        override fun execute(): Response {
            return call.execute()
        }

    }
```
Any requests made by this client will hang and not invoke onResponse or onFailure.

This is happening because line 142 in ApolloServerIntercptor.java is using an AtomicReference which has been set to the DecoratedCall produced by the custom Call.Factory. It is calling compareAndSet with the Call provided as part of the onResponse parameters, which is the inner undecorated OkHttp RealCall. As they are not the same reference, this returns false and does not invoke the callbacks.

ApolloServerIntercptor.java
```
      @Override
      public void onResponse(@NotNull Call call, @NotNull Response response) {
        if (httpCallRef.compareAndSet(call, null)) {
          if (disposed) return;
          callBack.onResponse(new ApolloInterceptor.InterceptorResponse(response));
          callBack.onCompleted();
        }
      }
```

OkHttp's RealCall.execute() 
```
...
responseCallback.onResponse(RealCall.this, response);
...
```

**Version**
Started happening in 1.2.1.
Working ok in 1.2.0

PR #1836, opened 2019-12-10
Add an integration test case for call factory that decorates calls

steventannz:
Tests for #1825 to prevent regressions

I have checked that the tests pass with the fix in #1828 and fail without it.


martinbonnin:
Thanks ! Any reason to write the test in Java rather than Kotlin ? Not than anything wrong with Java but I would expect new code to be more and more written in Kotlin ?

steventannz:
Absolutely no reason for it to be in Java. I just noticed that all the code that I had looked so far in this codebase had been in Java, so I figured at the time I'd stick to the status quo. More than happy to convert this to Kotlin, here's the PR: #1847

Issue #1794, opened 2019-11-22
Support hashing of queries other than SHA256


SaurabhKukreja:
**Is your feature request related to a problem? Please describe.**
We at Wayfair use a **different hashing mechanism** than standard SHA256 supported by Apollo, We have implemented it by forking but would like to share the solution with the apollo community so everyone can use it.

**Describe the solution you'd like**
We would like to contribute by adding a generic method where client can tell which hashing algo to use for the queries. We already have the solution and would like to see if apollo wants to support this algorithm that will work for other teams and organizations wanting to use it.

sav007:
We would love to add such support, please feel free to open PR.

martinbonnin:
@SaurabhKukreja Not sure what you had in mind but maybe make this a `HashingAlgorithm` class ?  not just a string like "SHA256" ? This way it's complely customizable and we can have `HashingAlgorithmSha256`, `HashingAlgorithmMD5`, etc...

I guess this should be a gradle configuration ? That would help ensure that transformed queries use the same algorithm as the runtime ?

SaurabhKukreja:
Sure. will make it generic as possible

boopathi:
@SaurabhKukreja I'm interested in this topic too regarding a generic solution (instead of just choosing a different hashing algorithm, calling a service / a CLI for getting the ID) and created the above tagged issue #1834 . Is there an existing PR? I can contribute to that.

martinbonnin:
We discussed this with @DSteve595 and the conclusion was that doing this 100% customizable while still keeping the gradle tasks incremental & cacheable would certainly require to split the current task in 2/3 tasks:

* generateIR
* generateHashes
* generateCode

where generate Code would take the input from generateIR and generateHashes (or something similar).

PR #1841, opened 2019-12-10
[feature] Provide a way to look at query generated operation output.

Laimiux:
The `apollo-tooling`(https://github.com/apollographql/apollo-tooling) API provides a way to generate a json file such as using `operationIdsPath`.
```json
{
  "$operation_id" : {
    "name"   : "$name",
    "source" : "$transformedQuery"
  }
}
```
Real example:
```json
{
  "8fe6f24988733e5c84d90cfcdcdc9590987654321b651d3ce35889bf5b5b477a": {
    "name": "CurrentLocation",
    "source": "query CurrentLocation {\n  currentLocation {\n    __typename\n    coordinates {\n      __typename\n      latitude\n      longitude\n    }\n    zipCode\n  }\n}"
  },
  "d11117159ca53b57bbee9c1b3cd64a3fe24a72d584f04aa1fa68eb098863edcd": {
    "name": "CurrentUser",
    "source": "query CurrentUser {\n  currentUser {\n    __typename\n    id\n    fullName\n    avatarImage {\n      __typename\n      url\n    }\n  }\n}"
  }
}
```

Currently, there is no way to access such information in a format outside of `.kt` or `.java` classes for this plugin.

TODO:
- [x] add a test

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -53,6 +53,14 @@ class GraphQLCompiler {
       val transformedQueryOutput = TransformedQueryOutput(args.packageNameProvider)
       transformedQueryOutput.apply { visit(ir) }.writeTo(transformedQueriesOutputDir)
     }
+
+    args.operationOutputDir?.let { operationIdOutputDir ->
+      if (operationIdOutputDir.exists()) {
+        operationIdOutputDir.deleteRecursively()
+      }
+      val operationIdOutput = OperationOutput(args.packageNameProvider)
+      operationIdOutput.apply { visit(ir) }.writeTo(operationIdOutputDir)
+    }
   }
 
   private fun CodeGenerationIR.writeJavaFiles(context: CodeGenerationContext, outputDir: File) {
@@ -113,6 +121,8 @@ class GraphQLCompiler {
     val OUTPUT_DIRECTORY = listOf("generated", "source", "apollo", "classes")
     @JvmField
     val TRANSFORMED_QUERIES_OUTPUT_DIRECTORY = listOf("generated", "apollo", "transformedQueries")
+    @JvmField
+    val OPERATION_OUTPUT_DIRECTORY = listOf("generated", "apollo", "operationOutput")
   }
 
   data class Arguments(
@@ -123,6 +133,7 @@ class GraphQLCompiler {
       val packageNameProvider: PackageNameProvider,
       val generateKotlinModels: Boolean = false,
       val transformedQueriesOutputDir: File? = null,
+      val operationOutputDir: File? = null,
       val generateAsInternal: Boolean = false,
 
       // only if generateKotlinModels = false
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -54,12 +54,12 @@ class GraphQLCompiler {
       transformedQueryOutput.apply { visit(ir) }.writeTo(transformedQueriesOutputDir)
     }
 
-    args.operationOutputDir?.let { operationIdOutputDir ->
-      if (operationIdOutputDir.exists()) {
-        operationIdOutputDir.deleteRecursively()
+    args.operationOutputDir?.let { operationOutputDir ->
+      if (operationOutputDir.exists()) {
+        operationOutputDir.deleteRecursively()
       }
-      val operationIdOutput = OperationOutput(args.packageNameProvider)
-      operationIdOutput.apply { visit(ir) }.writeTo(operationIdOutputDir)
+      val operationOutput = OperationOutput(args.packageNameProvider)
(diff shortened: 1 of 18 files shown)


martinbonnin:
Can you share a bit more what your use case is ? There are some related questions about customizing the hash function (https://github.com/apollographql/apollo-android/issues/1794) and maybe we can hit two birds with one stone ?

Laimiux:
I need to whitelist the queries by sending some information such as generated `operation id`, `operation name` and `transformed query` to the backend. The iOS counterpart is using `apollo-tooling` and `operationIdsPath` flag to generate a json file to send to the backend. I currently don't see a good way to generate this output.

Laimiux:
Regarding the hash function, the existing mechanism for hash generation works for me.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Operation.kt
```diff
@@ -15,6 +15,7 @@ data class Operation(
     val fields: List<Field>,
     val filePath: String,
     val fragmentsReferenced: List<String>,
+    // NOTE: this property is never correctly set!
     val operationId: String
```
martinbonnin: Could we remove it ?

Laimiux: Yep, can do that. One concern is that currently, we have a few places calling
```QueryDocumentMinifier.minify(operation.sourceWithFragments).sha256()```. Given the discussion around different id hash generation, it might make sense to centralize the generation of this ID so we don't accidentally have a mismatch.

File: apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt [hunk 2]
```diff
@@ -53,6 +53,14 @@ class GraphQLCompiler {
       val transformedQueryOutput = TransformedQueryOutput(args.packageNameProvider)
       transformedQueryOutput.apply { visit(ir) }.writeTo(transformedQueriesOutputDir)
     }
+
+    args.operationOutputDir?.let { operationIdOutputDir ->
```
martinbonnin: Nitpicking: `operationIdOutputDir` could be renamed to `operationOutputDir` so that it's consistent with other use cases.

martinbonnin:
@DSteve595 would you be ok using the operationOutput json instead of the transformed queries ? They overlap in terms of functionality with the operationOutput also having `id`

DSteve595:
What does that json look like?

martinbonnin:
I understand it would look like this (@Laimiux please correct me if I'm wrong):

```
{
  "$operation_id" : {
    "name"   : "$name",
    "source" : "$transformedQuery"
  }
}```

So it would contain all the transformedQueries along with the operationId and operationName in a single big json file.

**Comment by DSteve595** [comment]
That seems totally fine. Would be nice if there were a way to get the values through the DSL, otherwise consumers have to use a json parser.

**Comment by Laimiux** [comment]
Yep, that's the format. The only minor correction is that it generates a json file for each `.graphql` file (not a single json file for all .graphql).

@DSteve595 could you explain what you mean by DSL? Are you talking about providing a lambda/interface to the apollo plugin which will be called with some data structures that represent the codegen report?

**Comment by DSteve595** [comment]
Yes. Preferably a Property.

**Comment by martinbonnin** [comment]
@Laimiux does it have to be one json file per operation ? I understood the apollo tools were generating a single file so maybe it's better if we stick to what they're doing ?

@DSteve595 for up-to-date checks to work, I think it'll have to be a `DirectoryProperty` or `RegularFileProperty` (depending if it's one json per operation or one big json for all). What we could do is provide a `OperationOutput.deserialize()` to transform the directory/file into data structures.

**Comment by Laimiux** [comment]
@martinbonnin I initially had a single-file per module, but for some reason split it up into multiple json files. As this does not seem to be preferable, I'll revert the change.

@tasomaniac the output JSON structure is the same as what `apollo-tooling` library provides with `operationIdsPath` parameter.  (https://github.com/apollographql/apollo-tooling#apollo-clientcodegen-output)

For context `generate.ts` file:
```ts
interface OperationIdsMap {
  name: string;
  source: string;
}

function writeOperationIdsMap(context: CompilerContext) {
  let operationIdsMap: { [id: string]: OperationIdsMap } = {};
  Object.keys(context.operations)
    .map(k => context.operations[k])
    .forEach(operation => {
      operationIdsMap[operation.operationId!] = {
        name: operation.operationName,
        source: operation.source
      };
    });
  fs.writeFileSync(
    context.options.operationIdsPath,
    JSON.stringify(operationIdsMap, null, 2)
  );
}
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Operation.kt
Laimiux: Removed the property.

Laimiux:
Reverted back to a single JSON file per module. Currently, it will be placed in `build/apollo/${serviceName}/${variant}/operationOutput/OperationOutput.json`.

martinbonnin:
Follow up PR is there: https://github.com/apollographql/apollo-android/pull/1859

tasomaniac:
@Laimiux The docs were moved to `docs/source` folder. If you can fix the conflicts, then this can be shipped :shipit:

Laimiux:
Rebased + fixed conflicts.

PR #1845, opened 2019-12-11
[Gradle Plugin] Remove after evaluate

martinbonnin:
Attempt at removing `project.afterEvaluate` from the plugin following a discussion with the Gradle team. The idea is to:

* register all possible tasks: `generate${apolloVariant}${service}${language}ApolloSources` regardless of what the user configured in the extension:
  * generateMainStarwarsJavaApolloSources
  * generateMainStarwarsKotlinApolloSources
  * generateDebugStarwarsJavaApolloSources
  * generateDebugServiceJavaApolloSources
  * etc...
* enable/disable the tasks during task configuration, once the `apolloExtension` has been configured and `generateKotlinModels` is known.

This is in theory harmless in terms of performance since the extra tasks are simply "registered" and  registration is very fast but the Angroid Gradle Plugin configure them all eagerly for `registerJavaGeneratingTask` so maybe it adds a bit of configuration time, not sure.

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
@@ -95,15 +95,12 @@ open class ApolloPlugin : Plugin<Project> {
          * services or set generateKotlinModels. Here we disable the ones that do not make sense once we
          * know the actual user configuration.
          */
-        var enabled = true
-        if (generateKotlinModels != compilationUnit.kotlin) {
-          enabled = false
-        }
-        if (!compilationUnit.service.isUserDefined && hasUserDefinedServices()) {
-          enabled = false
-        }
 
-        it.enabled = enabled
+        it.enabled = when {
+          generateKotlinModels != compilationUnit.kotlin -> false
+          !compilationUnit.service.isUserDefined && hasUserDefinedServices() -> false
+          else -> true
+        }
 
         compilationUnit.setSourcesIfNeeded(compilerParams.graphqlSourceDirectorySet, compilerParams.schemaFile)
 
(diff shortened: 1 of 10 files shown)


tasomaniac:
Just read the description. This is also the approach that I followed when I migrated some plugins to lazy task configuration. I had to register "meta" tasks to just trigger other tasks depending on the configuration.

I think it is a quite a bit of burden to plugin developers to deal with this. But it is better for users at the end. 👍

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
```diff
@@ -157,6 +88,23 @@ open class ApolloPlugin : Plugin<Project> {
...
 
+        val generateKotlinModels = compilerParams.generateKotlinModels.getOrElse(false)
+
+        /**
+         * To avoid project.afterEvaluate, all tasks are registered before we know if the user configured
+         * services or set generateKotlinModels. Here we disable the ones that do not make sense once we
+         * know the actual user configuration.
+         */
+        var enabled = true
+        if (generateKotlinModels != compilationUnit.kotlin) {
```
tasomaniac: This could be nicer with a when statement. Right now I'm on mobile. I'm thinking something like this 

```
it.enabled = when {
    condition -> false
    else -> true
}
```

martinbonnin: yup, much nicer!

martinbonnin:
> I think it is a quite a bit of burden to plugin developers to deal with this. But it is better for users at the end. 👍

Completely agree. Actually, in this specific case, it can even be argued if it really brings something to the average user as it adds tasks and complexity.

The only case I can think of is if another plugin consumes the Apollo api then maybe it would be easier to interact with. But I'm not even sure about that since we now have a nice public api that should abstract the timing/calbacks issues.

martinbonnin:
I'm closing this one for now, I'll reopen it if we find a good reason to.

PR #1849, opened 2019-12-11
Add customizable persisted query id generation

### LINKED ISSUE #1794: Support hashing of queries other than SHA256

boopathi:
Add functionality to override the default ID generation logic (`sha256(minifiedQueryStr)`).

The Gradle plugin now takes a new option - `operationIdGenerator` - an `instance` which implements the interface `OperationIdGenerator`. The implementation uses two overrides -  a function `apply` which takes in the document string and returns the id, and a `version` parameter to indicate updates to implementation to invalidate the cache.

+ fix #1794 

Example usage in a project: 

inline in **build.gradle**:

```grade
import com.apollographql.apollo.compiler.OperationIdGenerator

class MyIdGenerator implements OperationIdGenerator {
    String apply(String operationDocument, String operationFilepath) {
        return operationDocument().toString();
    }
    String version = "v1"
}

apollo {
  operationIdGenerator = new MyIdGenerator()
}
```

or as a separate project 

current project's **build.gradle**:

```gradle
import com.example.MyIdGenerator;

apollo {
  operationIdGenerator = new MyIdGenerator()
}
```

the IdGenerator project - **MyIdGenerator.kt**:

```kt
package com.example

import com.apollographql.apollo.compiler.OperationIdGenerator

class MyIdGenerator: OperationIdGenerator {
    override fun apply(operationDocument: String, operationFilepath: String): String {
        return operationDocument.length.toString()
    }
    override val version = "v1"
}
```


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt
```diff
@@ -18,6 +19,10 @@ abstract class ApolloGenerateSourcesTask : DefaultTask() {
   @get:Optional
   abstract val customTypeMapping: MapProperty<String, String>
 
+  @get:Input
+  @get:Optional
+  abstract val customIdGenerator: Property<String>
```
martinbonnin: Thinking out loud: what about just making this? 

```kotlin
  @get:Optional
  abstract val customIdGenerator: CustomIDGenerator
```

Then we would rely on the fact that changing the `CustomIDGenerator` would invalidate the whole build as it would change the build scripts somehow. I'm not sure about that but it should, right ? In all cases, it's not too different than the current String property + reflection and provides compile time type safety.

boopathi: If we use Optional, Internal, how can we pass the IdGenerator from the build.gradle file? 

I tried this change and the tests fail with `Could not generate a decorated class for class com.apollographql.apollo.gradle.internal.ApolloGenerateSourcesTask` .

martinbonnin: Oopps, sorry, the property shouldn't be abstract in that case, I edited my comment above

boopathi: I don't understand how we can pass `customIdGenerator` to Apollo's task `:generateApolloSources` if this is not an `Input`.

martinbonnin: `Input` is used for up-to-date checks but ultimately, `ApolloGenerateSourcesTask` is a regular java class so you can set fields like you would do in regular Kotlin:

```kotlin
// ApolloPlugin.kt
     project.tasks.register(taskName, ApolloGenerateSourcesTask::class.java) {
        it.customIdGenerator = apolloExtension.customIdGenerator
        ...
```

```build.gradle.kts
apollo {
   customIdGenerator = {transformedQuery ->
       myHashFunction(transformedQuery)
   }
}
```

boopathi: Thank you. I updated both the plugins to use this pattern and pushed the changes.

martinbonnin: The thing is I'm not sure how this plays with up-to-date checks, cacheability, etc... I guess we'll have to add tests for that.

boopathi: Can you help me write these tests?

Should I overwrite the existing integration tests' `testProject/.../build.gradle` to use a custom id generator or should I create a new project for testing this?

kenyee:
Another fair question is how many different ways are there to generate graphQL hash IDs?
I.e. do you need full customizability or just need to support "de facto" standards?
Passing in classes seems a little complicated...though it is an interesting way to send in a custom hash method...

martinbonnin:
I think one of the use cases is to have the hash generated server side (ping @DSteve595). In this case, the hash function will have to run custom code to make a round trip to the server.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt
tasomaniac: I think this would break cacheability. And it is not easy to write tests for build cache behavior.

DSteve595:
Yeah, anything that can be done to keep that hashing cacheable+lazy, because it can potentially be expensive.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
```diff
@@ -34,6 +35,13 @@ interface CompilerParams {
    */
   val customTypeMapping: MapProperty<String, String>
 
+  /**
+   * For custom persisted query Ids.
+   *
+   * null by default
```
tasomaniac: `null by default` is not really explanatory. 

What about this:

```
If not provided, default hashing algorithm (sha256) will be used
```

martinbonnin: It would be nice to include implementations for Sha1 and Md5: `Md5IdGenerator`, `Sha1IdGenerator`. Can be done in a later PR though

tasomaniac:
Just an idea: Would it be possible to separate query generation and hashing as 2 separate tasks? Maybe then, it could be easier to keep the task cacheable.

We don't really need to register hashing function as input. Inputs are the GraphQL operations, hashes are the output. If we assume that the hashing algorithm does not change all the times then it should be fine.

If hashing supported in backend changes, that's a bigger migration anyways.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/OperationTypeBuilder.kt
```diff
@@ -25,11 +27,14 @@ internal fun Operation.ast(
     isSubscription() -> OperationType.Type.SUBSCRIPTION
     else -> throw IllegalArgumentException("Unsupported GraphQL operation type: $operationType")
   }
+
+  val operationId = customIdGenerator?.apply(sourceWithFragments) ?: QueryDocumentMinifier.minify(sourceWithFragments).sha256()
```
martinbonnin: Don't we need to apply the `customIdGenerator` to the minified source ? This is what is going to be sent over the wire.

Also, we could introduce a `Sha256IdGenerator` so that this code ultimately becomes:

```kotlin
val operationId = (customIdGenerator ?: Sha256IdGenerator()).apply(QueryDocumentMinifier.minify(sourceWithFragments)) 
```

martinbonnin:
> If hashing supported in backend changes, that's a bigger migration anyways.

Good point. Maybe add `CompilerParams.customIdGeneratorVersion` to force a rebuild if the server changes ? Can certainly be done in a separate Pull Request too.

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt
martinbonnin: Turns out gradle 6 doesn't like Optional + Internal. The CI currently fails with:

```
Property 'customIdGenerator' getter 'getCustomIdGenerator()' annotated with @Internal should not be also annotated with @Optional. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```
So I guess it's just`Internal` then ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/OperationTypeBuilder.kt
boopathi: I updated this change to use minified code to `apply`. Also, I add the Sha256IdGenerator.

sav007:
I like the idea of providing flexibility on custom query hash / id generator.
Couple suggestions:
- what do you think of renaming `customIdGenerator ` to `queryHashGenerator`?
- as @martinbonnin suggested we should provide basic implementations for SHA and MD5 and by default it will be SHA (`queryHashGenerator = ApolloQueryHashGenerator.SHA256`). Introduce smth like:
```
  final class ApolloQueryHashGenerator {
     static final QueryHashGenerator SHA256
     static final QueryHashGenerator MD5
     ....
```

sav007:
After digging into our code here is another thing to keep in mind related to the query auto persistence feature. Apollo GraphQL server always expects query hash to be SHA256 to be sent in `extensions { sha256Hash`. I guess if client enables `enableAutoPersistedQueries` then `customIdGenerator ` should be ignored and SHA256 should be used instead.

martinbonnin:
@boopathi I added a compiler test for the CustomIdGenerator there: https://github.com/martinbonnin/apollo-android/commits/custom-id-generator-tests. Feel free to cherry pick it. Also this branch contains a fix for java codegen.

I'll try to add some plugin tests to this branch as well.

martinbonnin:
Plugin tests there: https://github.com/martinbonnin/apollo-android/commit/440b781f0824efbba944d4d06dd0334b75a12388. It turns out that changing the `build.gradle` file is not enough to mark the `generateApolloSources` task as not up-to-date (The second test fails at the moment).

What about adding a version to `CustomIdGenerator` ?

```kotlin
interface CustomIdGenerator {
  fun apply(queryString: String): String

  /**
   * The version of the CustomIdGenerator
   * 
   * You should change the version every time the implementation of the CustomIdGenerator
   * changes to let gradle and build tools know that they have to re-generate the 
   * resulting files.
   */
  val version: String
}
```

Then make `customIdGenerator.version` an input of `generateApolloSourcesTask`. It's a simple fix and allows for 100% correct gradle checks. The only downside is that users will have to remember to update but that seems fair enough.

boopathi:
> Then make customIdGenerator.version an input of generateApolloSourcesTask

How do I annotate version inside CustomIdGenerator as Input?

martinbonnin:
@boopathi something like this should do the job:

```
  @Input
  @Optional
  fun customIdGeneratorVersion() = customIdGenerator.orNull?.version 
```

marwinxxii:
What do you think about providing a query file name as an argument along with the query itself?
This would allow to have a mapping of query file to persisted query id.

Would be useful in case queries are registered separately from gradle plugin and we simply want to provide a mapping from query to id.
This can be workarounded by going through all of the query files and finding which one matches the query, but this does not seem nice to me.

Example:

```
queries = [
   "myQuery.graphql": "id1",
   "yourQuery.graphql": "id2"
]
apollo {
  customIdGenerator = { query, queryFile ->
    queries.find { it.key == queryFile }.value ?: throw RuntimeException("Query is not registered")
  }
}
```

(`queryFile` could be also a full path)

apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt
```diff
@@ -18,6 +19,13 @@ abstract class ApolloGenerateSourcesTask : DefaultTask() {
   @get:Optional
   abstract val customTypeMapping: MapProperty<String, String>
 
+  @get:Internal
+  abstract val customIdGenerator: Property<CustomIdGenerator>
+
+  @Input
+  @Optional
+  fun customIdGeneratorVersion() = customIdGenerator.orNull?.version
```
boopathi: how should this be used? The UP_TO_DATE checks doesn't say it's updated when version is changed.

martinbonnin: It looks like gradle requires Java Beans syntax:

```suggestion
  fun getCustomIdGeneratorVersion() = customIdGenerator.orNull?.version
```

martinbonnin: It's also what makes the gradle 6.0 tests fail:

```
Type 'ApolloGenerateSourcesTask': non-property method 'customIdGeneratorVersion()' should not be annotated with: @Input, @Optional. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/CustomIdGeneratorTests.kt
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Test
+  fun `changing the customIdGenerator recompiles sources`() {
+    val apolloConfiguration = """
+      class MyIdGenerator implements CustomIdGenerator {
+          String apply(String queryString, String queryFilepath) {
+              return queryString.length().toString();
+          }
+          String version = "1.0.0-SNAPSHOT"
```
martinbonnin: Nitpicking: What about `CustomIdGenerator-v1` or something similar here ? `-SNAPSHOT` feels very maven-oriented and isn't very relevant here.

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/CustomIdGeneratorTests.kt
```diff
@@ -0,0 +1,74 @@
...
+      val gradleFile = File(dir, "build.gradle").readText()
+
+      File(dir, "build.gradle").writeText("import com.apollographql.apollo.compiler.CustomIdGenerator\n$gradleFile")
+
+      var result = TestUtils.executeTask("generateApolloSources", dir)
+
+      Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":generateApolloSources")!!.outcome)
+
+      File(dir, "build.gradle").replaceInText("queryString.length()", "(queryString.length() * 2)")
+      File(dir, "build.gradle").replaceInText("1.0.0-SNAPSHOT", "2.0.0-SNAPSHOT")
```
martinbonnin: Same as above: what about `CustomIdGenerator-v2` ?

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/CustomIdGeneratorTests.kt
boopathi: I just used it to be unique to search and replace below. I'll change it to the `MyIdGenerator-v1`

martinbonnin:
For the record, we should certainly add the default implementations for MD5 and Sha1 mentioned there: https://github.com/apollographql/apollo-android/pull/1849#issuecomment-568485214 and the override for auto-persisted queries mentioned there: https://github.com/apollographql/apollo-android/pull/1849#issuecomment-568494236

But there's already a lot of discussion in this pull request so I think we can merge it and open new ones to follow-up.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
```diff
@@ -0,0 +1,24 @@
+package com.apollographql.apollo.compiler
+
+interface CustomIdGenerator {
+  fun apply(
+      /**
+       * The minified query string source with fragments
```
sav007: nit: maybe it's better to say `GraphQL operation document to be sent to the server`

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
```diff
@@ -0,0 +1,24 @@
...
+       */
+      queryString: String,
```
sav007: nit: how about name it as `operationDocument`? As it will be used not only for queries but for subscriptions as well.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -119,6 +122,7 @@ class GraphQLCompiler {
       val ir: CodeGenerationIR,
       val outputDir: File,
       val customTypeMap: Map<String, String>,
+      val customIdGenerator: CustomIdGenerator? = null,
```
sav007: nit: should we make it as non optional? With default value `Sha256IdGenerator `

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
```diff
@@ -0,0 +1,24 @@
+package com.apollographql.apollo.compiler
+
+interface CustomIdGenerator {
```
sav007: nit: should we just call it `OperationIdGenerator` instead? And make `CodeGenerationContext` always have value set to Sha256 ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * The version of the CustomIdGenerator
+   *
+   * You should change the version every time the implementation of the CustomIdGenerator
+   * changes to let gradle and build tools know that they have to re-generate the
+   * resulting files.
+   */
+  val version: String
+}
```
sav007: Another suggestion, what do you think of adding default implementation here:
```
interface OperationIdGenerator {
...
  object Sha256: OperationIdGenerator { ... }
  object Md5: OperationIdGenerator { ... }
}
```

So on user side it will look like this:
```
apollo {
  operationIdGenerator = OperationIdGenerator.Md5
}
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
boopathi: There was `customTypeMapping`. I followed the same convention. Should I change it to `OperationIdGenerator`? Or 

the interface name as `OperationIdGenerator` and the input name as `customIdGenerator`. 

```
apollo {
  customIdGenerator = new OperationIdGenerator() {
    // ...
  }
}
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
boopathi: I did try this before. Looks like, in Kotlin, the optional parameters are a compile time thing. So, like in tests, when you have 

```kt
val operationIdGenerator = when (folder.name) { 
  // ... 
  else -> null
}

GraphQLCompiler.Arguments(
  operationIdGenerator = operationIdGenerator,
)
```

the null doesn't cast to the missing parameter and there is no value to represent a missing parameter because it's a compile time thing. There were workarounds like a secondary constructor or using `if (x == null) Arguments() else Arguments(p = p)`. Can you suggest something regarding this?

martinbonnin: You could just do:

```kotlin
      val customIdGenerator = when (folder.name) {
        "custom_id_generator" -> object : CustomIdGenerator {
          override fun apply(queryString: String, queryFilepath: String): String {
            return "hash"
          }

          override val version: String = "1"
        }
        else -> OperationIdGenerator.Sha256
      }
```

boopathi: The default generator (`OperationIdGenerator.Sha256`) is now specified in 3 places (not including tests) - 

1. Arguments, 
1. plugin (arguments construction - default instead of null
1. plugin-incubating (arguments construction - default instead of null)

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
boopathi: Instead of an `object`, I used a class: In the Kotlin plugin-incubating, I'm able to use the default value during arguments construction like so 

```
Arguments(operationIdGenerator = OperationIdGenerator.Sha256)
```

But in the Groovy version, Kotlin `object`s are `java.lang.Class`es and are not castable/assignable to class `OperationIdGenerator`. I get this error
  + `Cannot cast object 'class com.apollographql.apollo.compiler.OperationIdGenerator$Sha256' with class 'java.lang.Class' to class 'com.apollographql.apollo.compiler.OperationIdGenerator'`

boopathi:
Thanks for the reviews.

@sav007 @martinbonnin can you verify the changes?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/CustomIdGenerator.kt
martinbonnin: @boopathi you should be able to reference a kotlin object using `.INSTANCE`. Something like this: 

```
OperationIdGenerator gen = OperationIdGenerator.Sha256.INSTANCE
```

martinbonnin: That being said, it makes the groovy experience a bit weird so I like better keeping the class and not using object there.

louistsaitszho:
@martinbonnin @sav007 is there gonna be a 1.2.3 this month, or does it need to wait for #1889?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/OperationTypeBuilder.kt
```diff
@@ -1,14 +1,15 @@
...
 import com.apollographql.apollo.compiler.escapeKotlinReservedWord
 import com.apollographql.apollo.compiler.ir.Operation
-import com.apollographql.apollo.compiler.sha256
 import com.apollographql.apollo.internal.QueryDocumentMinifier
 
 internal fun Operation.ast(
     operationClassName: String,
-    context: Context
+    context: Context,
+    operationIdGenerator: OperationIdGenerator
```
Laimiux: Could we add an extension function to this
```kotlin
fun OperationIdGeneration.apply(operation: Operation): String {
  return apply(QueryDocumentMinifier.minify(sourceWithFragments), filePath)
}
```

Issue #1842, opened 2019-12-10
Generated Java code using internal classes 


robcolburn:
**Summary**
Hi there, I'm using a collection of scripts from a former author that manually invoke the compiler in a similar fashion to ApolloCodegenTask.groovy, however from Java code. Then generated code seems to attempt to use internal API. Formerly this collection of scripts used apollo-android 1.0.0, and I'm attempting to update that code to 1.2.1 or 1.2.2.

**Description**
The new generated code seems to want to make use of internal classes. Should I be trying to get compiler to not do this with arguments? Or, should I try to vend these internal classes publicly to the app.

```
import com.apollographql.apollo.api.internal.Mutator;
import com.apollographql.apollo.api.internal.SimpleOperationResponseParser;
import com.apollographql.apollo.api.internal.UnmodifiableMapBuilder;
import com.apollographql.apollo.api.internal.Utils;
```

Here are some examples I'm seeing

```java
      public Builder userLists(@NotNull Mutator<List<UserList.Builder>> mutator) {
        Utils.checkNotNull(mutator, "mutator == null");
        List<UserList.Builder> builders = new ArrayList<>();
        if (this.userLists != null) {
          for (UserList item : this.userLists) {
            builders.add(item != null ? item.toBuilder() : null);
          }
        }
        mutator.accept(builders);
        List<UserList> userLists = new ArrayList<>();
        for (UserList.Builder item : builders) {
          userLists.add(item != null ? item.build() : null);
        }
        this.userLists = userLists;
        return this;
      }
```

```java
  @Override
  @NotNull
  public Response<FollowCount.Data> parse(@NotNull final BufferedSource source,
      @NotNull final ScalarTypeAdapters scalarTypeAdapters) throws IOException {
    return SimpleOperationResponseParser.parse(source, this, scalarTypeAdapters);
  }
```

```java
  public static class Data implements Operation.Data {
    static final ResponseField[] $responseFields = {
      ResponseField.forList("userLists", "userLists", new UnmodifiableMapBuilder<String, Object>(2)
      .put("id", new UnmodifiableMapBuilder<String, Object>(2)
        .put("kind", "Variable")
        .put("variableName", "userId")
        .build())
      .build(), true, Collections.<ResponseField.Condition>emptyList())
    };
```

**Version**
1.2.x - though there is more internal use in 1.2.2 than 1.2.1

martinbonnin:
[...]

PR #1852, opened 2019-12-11
[Incubating Plugin] make sure all apollo dependencies are the same version

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/1842

This will throw if the runtime (or any other dependency) version doesn't match the gradle plugin version (or any other).

### DIFF
--- a/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
+++ b/apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
@@ -5,6 +5,8 @@ import com.apollographql.apollo.gradle.api.ApolloSourceSetExtension
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.ConfigurationContainer
 import org.gradle.api.plugins.ExtensionAware
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.util.GradleVersion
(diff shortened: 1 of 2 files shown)


apollo-gradle-plugin-incubating/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
```diff
@@ -260,11 +262,35 @@ open class ApolloPlugin : Plugin<Project> {
       registerCodegenTasks(project, apolloExtension)
 
       registerDownloadSchemaTasks(project, apolloExtension)
+
+      checkVersions(project)
+    }
+
+    data class Dep(val name: String, val version: String?)
+
+    fun getDeps(configurations: ConfigurationContainer): List<Dep> {
```
tasomaniac: Does this mean that we resolve dependencies during configuration time? If so, I wonder if that can be done lazily.

martinbonnin: I was under the impression that gradle would resolve the dependencies during configuration in all cases. If not then it's going to add some processing to each and every gradle build :/.

What do you mean by `lazily`? maybe just run the check along the `compileApolloSources` tasks? And not if it's already up-to-date ?

tasomaniac: I'm not sure about the solution itself myself. Maybe do this when one of our tasks started.

martinbonnin: On my machine, once the gradle cache warmed up, it takes 7ms to execute this check, maybe it's not too bad ?

tasomaniac: I don't think this is a blocker anyways. I think this becomes little annoyance if you have really bad connection or start the project offline etc.

PR #1859, opened 2019-12-14
Remove transformed queries

martinbonnin:
Now that https://github.com/apollographql/apollo-android/pull/1841 is merged, this PR can be merged.

This PR:

* removes the transformed queries as `operationOutput.json` contains more data
* makes `operationOutputFile` a `RegularFileProperty` instead of a `DirectoryProperty`
* adds `OperationOutput(File)` to transform OperationOutput.json to data structures easier to work with. There an exemple in the integration tests: https://github.com/apollographql/apollo-android/compare/master...martinbonnin:remove-transformed-queries?expand=1#diff-2966339ccc9c3f57b884f0a635458863R12. Would that be ok ?


martinbonnin:
Instead of using `transformedQueriesDir` with a flat dir of all transformedQueries, you would now have to use `operationOutputFile` that is a json that maps operationIds to transformedQueries.

You can use `OperationOutput(file: File)` to create pojos from the json file.

@DSteve595 is it ok for you ?

DSteve595:
Sounds great. Thank you all for your hard work.

doc/persisted-queries.md
```diff
@@ -0,0 +1,17 @@
+# Persisted queries
+
+## Automatic persisted queries
+
+Apollo supports [Automatic Persisted Queries](https://www.apollographql.com/docs/apollo-server/performance/apq/).
+
+// TODO: add some sample code
```
tasomaniac: Is there any sample code really? I think it may be just enabling the flag in Gradle. The rest will be handled.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -46,23 +47,16 @@ class GraphQLCompiler {
       )
     }
 
-    args.transformedQueriesOutputDir?.let { transformedQueriesOutputDir ->
-      if (transformedQueriesOutputDir.exists()) {
-        transformedQueriesOutputDir.deleteRecursively()
+    args.operationOutputFile?.let { operationOutputFile ->
+      val dir = operationOutputFile.parentFile
+      if (dir.exists()) {
+        dir.deleteRecursively()
```
tasomaniac: Are you sure about deleting the parent directory with all its content? What if the directory is shared with other output? What about just deleting the provided file?

apollo-gradle-plugin/src/test/groovy/com/apollographql/apollo/gradle/integration/BasicAndroidSpec.groovy
```diff
@@ -350,7 +350,7 @@ class BasicAndroidSpec extends Specification {
   }
 
   def cleanupSpec() {
-    FileUtils.deleteDirectory(testProjectDir)
+    //FileUtils.deleteDirectory(testProjectDir)
```
tasomaniac: Can this be a mistake?

doc/persisted-queries.md
martinbonnin: My bad, it's a bad merge, there's already more documentation under `docs/source/advanced/persisted-queries.md`. I'll remove that file.

apollo-gradle-plugin/src/test/groovy/com/apollographql/apollo/gradle/integration/BasicAndroidSpec.groovy
martinbonnin: Definitely a mistake, I do this during debugging, I forgot to uncomment. Thanks for noticing !

apollo-integration/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ dependencies {
   add("implementation", "com.apollographql.apollo:apollo-rx2-support")
   add("implementation", "com.apollographql.apollo:apollo-coroutines-support")
   add("implementation", "com.apollographql.apollo:apollo-http-cache")
+  add("implementation", "com.apollographql.apollo:apollo-compiler")
```
tasomaniac: CI errors with duplicate class error. This may be the cause.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
martinbonnin: Yep, feels safer, will do 👍

apollo-integration/build.gradle.kts
martinbonnin: yea, looks like it pulls both antlr and antlr-runtime at the same time...

tasomaniac: Would `compileOnly` work here? If not, you can provide a lambda block and call `exclude` function to exclude `antlr`. Not sure if that Gradle API is `Kotlin` friendly.

martinbonnin: I just removed the `antlr-runtime` dependency from apollo-compiler altogether 🤷‍♂: https://github.com/apollographql/apollo-android/pull/1859/commits/2c9672d966992dab93102af8ad6698c8c9724d58#diff-8d72e2bd6e0f687326024c0f1306de87L18

What's weird is that both `antlr` and `antlr-runtime` seem to include the same duplicate classes. I would have expected one to depend on the other. Anyways, it looks like it fixed the build.

martinbonnin:
I guess it's a breaking change to whoever is relying on transformed queries. It doesn't change anything to the runtime and/or codegen. Overall, I would say users of transformed queries are advanced enough to migrate to operation output but it's definitely worth a mention in the changelog.

tasomaniac:
I guess mention in the changelog would be enough. Their build configuration will fail to find the old property.

PR #1862, opened 2019-12-17
Apollo coroutines are now on jcenter

martinbonnin:
Remove previous references to `dl.bintray.com`


sav007:
Wow, how did you manage to fix this?

martinbonnin:
Bintray fixed it. I think they might have been confused between  `apollo-coroutine-support` and `apollo-couroutines-support` and maybe missed that we need the later to be promoted. In all cases, it's looking good now.

PR #1863, opened 2019-12-17
Support for custom query hashing

ketulshah1947:
This PR allows MD5 and SHA256 hashing with the flexibility of adding other algorithms.

You can provide the hashing algorithm in the build.gradle file of the project consuming this library.

apollo {
        hashingAlgorithm = "MD5"
}

Currently supported types of Hashing algorithms are MD5 and SHA256.

If no parameter (or invalid parameter) provided, it will default back to "SHA256" algorithm which is Apollo's default option.


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+import java.util.Locale
+
+class HashingAlgorithms(
+    private val algorithm: String?
+) {
+
+    fun applyHashing(text: String): String {
+        return when ((algorithm ?: SHA256).toUpperCase(Locale.ENGLISH)) {
```
kenyee: should this be an enum instead of specific string?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
```diff
@@ -0,0 +1,40 @@
+package com.apollographql.apollo.compiler
+
+import java.math.BigInteger
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+import java.util.Locale
+
+class HashingAlgorithms(
+    private val algorithm: String?
```
kenyee: why can't this default to SHA256 instead of being nullable?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    fun applyHashing(text: String): String {
+        return when ((algorithm ?: SHA256).toUpperCase(Locale.ENGLISH)) {
+            SHA256 -> sha256(text)
+            MD5 -> md5(text)
+            else -> sha256(text)
+        }
+    }
+
+    private fun sha256(text: String): String {
```
kenyee: would probably be better as a base class of Algorithm w/ a method named "encode(text): String"

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
```diff
@@ -0,0 +1,40 @@
...
+        val bytes = text.toByteArray(charset = StandardCharsets.UTF_8)
+        val md = MessageDigest.getInstance("SHA-256")
+        val digest = md.digest(bytes)
+        return digest.fold("") { str, it -> str + "%02x".format(it) }
```
kenyee: can use string interpolation instead of concat'ing strings

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    private fun md5(text: String): String {
+        val digest = MessageDigest.getInstance("MD5")
+        digest.update(text.toByteArray(Charset.forName("US-ASCII")), 0, text.length)
+        val magnitude = digest.digest()
+        return String.format("%0" + (magnitude.size shl 1) + "x", BigInteger(1, magnitude))
```
kenyee: ditto string interpolation

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
kingsleyliao: Enum would allow for compile time checking so that some of the tests won't be necessary

ketulshah1947: This string comes from project that consumes Apollo library. Not sure how that project can pass enum as gradle parameter.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
ketulshah1947: Fixed.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
ketulshah1947: Fixed.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
ketulshah1947: Fixed.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
ketulshah1947: Fixed.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/HashingAlgorithms.kt
kenyee: The project can reference classes from the plugin via an import AFAIK.

ketulshah1947: Do you have a snippet that I can refer?

martinbonnin:
Any reason to not use https://github.com/apollographql/apollo-android/pull/1849 that allows for complete customization of the hash function ?

PR #1867, opened 2019-12-22
Fix Rx extension configure lambda do not use the returned instance

### LINKED ISSUE #1861: RxJavaExtensions rxQuery and rxMutate operate on original call
**Summary**
`ApolloClient.rxQuery` and `ApolloClient.rxMutate` extensions do not configure original call and pass original one instead

**Description**
If we look at one of the extensions:
```
inline fun <D : Operation.Data, T, V : Operation.Variables> ApolloClient.rxQuery(
    query: Query<D, T, V>,
    configure: ApolloQueryCall<T>.() -> Unit = {}
): Observable<Response<T>> = query(query).apply(configure).rx()
```

We can see that `.apply(configure)` returns the original call from `query(query)` hence whatever I do within `configure` lambda will be ignored. 

The same applies to all other methods that use `configure` lambda.

**Version**
1.2.2

**Proposal**
Lambda should return a new `ApolloQueryCall` object and it could look like this:

```
inline fun <D : Operation.Data, T, V : Operation.Variables> ApolloClient.rxQuery(
    query: Query<D, T, V>,
    configure: ApolloQueryCall<T>.() -> ApolloQueryCall<T> = { this }
): Observable<Response<T>> = query(query).configure().rx()
```

tasomaniac:
More info on the bug is in #1861
Fixes #1861

### DIFF
--- a/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
+++ b/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/RxJavaExtensions.kt
@@ -59,8 +59,8 @@ inline fun <T> ApolloSubscriptionCall<T>.rx(
 @CheckReturnValue
 inline fun <D : Operation.Data, T, V : Operation.Variables> ApolloClient.rxQuery(
     query: Query<D, T, V>,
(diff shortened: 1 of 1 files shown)


tasomaniac:
Would it make sense to do a release for this fix? Or are we going to do another release soon anyways?

PR #1878, opened 2019-12-26
Move IdlingResources to AndroidX libraries

### LINKED ISSUE #1877: Migrate IdlingResources to AndroidX
ApolloIdlingResources still uses old `android.support` libraries. It works fine thanks to Android's Jetifier. 

But still, it would be good to migrate now. 

The module name is called `apollo-espresso-support`. It would be good to use this chance to rename it to `apollo-idling-resource`. Since this will be a breaking change, it would be easier for consumers with module name change.

tasomaniac:
Fixes #1877


apollo-idling-resource/src/test/java/com/apollographql/apollo/test/espresso/ApolloIdlingResourceTest.java
```diff
@@ -1,53 +1,48 @@
...
   private ApolloIdlingResource idlingResource;
   private ApolloClient apolloClient;
-  @Rule public final MockWebServer server = new MockWebServer();
-  private OkHttpClient okHttpClient;
 
   private static final long TIME_OUT_SECONDS = 3;
   private static final String IDLING_RESOURCE_NAME = "apolloIdlingResource";
 
-  private static final Query EMPTY_QUERY = new Query() {
+  private static final Query<Operation.Data, Object, Operation.Variables> EMPTY_QUERY = new Query<Operation.Data, Object, Operation.Variables>() {
```
tasomaniac: Can be seen as random changes. I just went over and fixed warnings etc here.

martinbonnin:
Are we sure we want to change the name of the artifact ? `espresso` is more generic, meaning we can add things like matchers or some other stuff in the future if we want to. Also, I'd say it's more easily identifiable to newcomers ? I guess more people know what `espresso` is compared to `idling-resources` ?

tasomaniac:
I don't think Apollo would need anything else. That's why espresso-support is misleading IMHO. There are other libraries doing something very similar without Espresso word in them. Like `RxIdler` or `okhttp-idling-resource`.

This is technically a major breaking change. It is not backward compatible to support library espresso users. If it is not likely that we go for version 2.0, then I would say that the name change makes sense.

martinbonnin:
Sounds good !

tasomaniac:
I was gonna go for `apollo-espresso-idling-resource`. That's more descriptive but really long 😅

PR #1880, opened 2020-01-02
Added flag to disable type singularization

EricKuck:
Picking up where https://github.com/apollographql/apollo-android/pull/1829 left off…

Since updating to the kotlin codegen, we've run into lots of issues with bizarre singularization in our project. Examples include `Pies` -> `Py`, `Ctas` -> `Ctum`, and `Slice` -> `Slouses`. Another user opened an issue to report this problem a while back: #1686.

This PR addresses the reality that it is impossible to accurately model the English language in a project like this. Additionally, any fixes to the Singularization rules would end up breaking existing codebases, which isn't ideal. Rather than adding crazy complicated NLP features to the library (which would still have issues), I opted to instead allow users to disable singularization altogether. Example usage from our project based on this fork follows:

```
    apollo {
        generateKotlinModels = true
        singularizeTypes = false
    }
```

Note: There was discussion in the other PR about defaulting this to false in various cases, but it seemed like a consensus wasn't quite reached. This PR simply adds the flag to disable singularization. Any future work with throwing warnings or defaulting to false can either come later or can be added to this PR if there is a concrete plan in place.


EricKuck:
I see that the tests failed on travis, but they're passing locally so I'm not sure what to change. Any tips?

tasomaniac:
The following test is failing. Perhaps the flag is not set properly for this test

com.apollographql.apollo.compiler.CodeGenTest > generateExpectedClasses[src/test/graphql/com/example/singularize_types] FAILED
    java.lang.Exception at CodegenTest.kt:49

martinbonnin:
I'm pretty sure this comes from the changes in master. Travis builds the merge so there might be some changes to the generated files in master that are not in the fixtures from the branch. It should be reproducible if you rebase the branch.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
```diff
@@ -73,6 +73,10 @@ class CodeGenTest(val folder: File) {
       } else {
         emptyMap()
       }
+      val singularizeTypes = when (folder.name) {
```
tasomaniac: Wouldn't this be more simple if you just do `folder.name == "singularize_types"`

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/SchemaTypeSpecBuilder.kt
```diff
@@ -134,7 +134,7 @@ class SchemaTypeSpecBuilder(
 
   private fun nestedTypeSpecs(): List<Pair<String, TypeSpec>> {
     return fields.filter(Field::isNonScalar).map {
-      it.formatClassName() to it.toTypeSpec(
+      it.formatClassName(context) to it.toTypeSpec(
```
martinbonnin: It feels a bit heavy handed to pass `context` around when all that's needed is `singularizeTypes` but I'm no expert of these classes so works for me as is too :)

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
```diff
@@ -72,6 +72,27 @@ class UpToDateTests {
     File(dir, "build.gradle").writeText(text.replace(apolloBlock, ""))
   }
 
+  fun `setting singularizeTypes in the build script re-generates the CustomType class`(dir: File) {
```
tasomaniac: Is this test working?

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
```diff
@@ -72,6 +72,27 @@ class UpToDateTests {
...
+        singularizeTypes = false
+      }
+    """.trimIndent()
+
+    File(dir, "build.gradle").appendText(apolloBlock)
+
+    val result = TestUtils.executeTask("generateApolloSources", dir)
+
+    // modifying the customTypeMapping should cause the task to be out of date
+    // and the task should run again
```
martinbonnin: Bad copy/paste. Also this method is unused at the moment. Either call it from `complete test()` or just remove it alltogether, I don't think we need up-to-date checks for each and every parameter as they all work the same.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
EricKuck: Yes, but every other check here uses when, so I made sure my code fit in with the existing patterns.

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
EricKuck: It's removed now, per @martinbonnin's comment

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/UpToDateTests.kt
EricKuck: Good call, removed this test altogether.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/SchemaTypeSpecBuilder.kt
EricKuck: Updated to only pass in the boolean.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
tasomaniac: I would personally prefer simpler solutions over consistency just for consistency's sake. 😅 Resolving this now since it is personal opinions at the end.

EricKuck:
After seeing the support for this PR, I went back and integrated this branch into our project. Honestly I think the other solution might have been better. Without any sort of singularization I'm left with some non-ideal class names. For examples, all elements of any "edges" array are now of type `Edges` vs `Edge` (`List<Edges>` vs `List<Edge>`). This is to be expected without singularization, but it's also pretty weird.

I imagine most schemas are going to have plenty of arrays of elements that are named as the plural form of that element, so this is going to create some weird code for lots of people. I'm obviously not a core contributor to this project, but I do think perhaps this isn't the way to go.

martinbonnin:
> For examples, all elements of any "edges" array are now of type Edges vs Edge (List<Edges> vs List<Edge>). This is to be expected without singularization, but it's also pretty weird.

Yikes, it does feel weird indeed. 
I'm starting to wondering if we can't make something better... There are 2 different names involved there:

* `edges` is the name of the field
* `Edge` is the name of the type of each item in the `edges` list.

This is beyond my current knowledge of how codegen works but couldn't we have something like this ? 

```
val edges: List<Edge>
```

That would be the best of both worlds ?

EricKuck:
I see no reason Apollo uses field names instead of the schema type names, but IMO keeping type names consistent between backend and frontend would be nice. This would be a pretty easy change to make - I already have it working locally with no issues (so far). Any objections to closing this PR and opening yet another one that adds a compiler param for `useSchemaTypeNames`?

EricKuck:
Ah, I take that back - I do see why type name wasn't used. One query/fragment might have several fields that all have the same schema type, but with different fields selected. Example:

```
fragment User on SchemaUser {
  homeAddress {
    street
  }
  workAddress {
    street
    city
  }
}
```
Assuming that both `homeAddress` and `workAddress` in the above example are instances of the schema type `Address`, they will be generated as `Address` and `Address1`. This would not only make code harder to read, but would also potentially cause future issues if something in the compiler changes that results in the models switching names somehow.

Perhaps this could be addressed by aliases, but then we end up with confusing logic for using field names if they're aliased, else using schema type names.

martinbonnin:
Indeed. The same kind of problem already happens when using field names actually. Using version 1.2.2, our codebase is full of `Node1`. With your previous example:

```graphql
fragment User on SchemaUser {
  homeAddress {
    coordinates {
      latitude
    }
  }
  workAddress {
    coordinates {
      latitude
      longitude
    }
  }
}
```

This would currently produce `Coordinate`, `Coordinate1`, `HomeAddress` and `WorkAddress`. 
With`useSchemaTypeNames = true`, it would produce `Coordinate`, `Coordinate1`, `Address` and `Address1`. 

That being said, not having to deal with suffixed type names would definitely be better. 
All in all, if there isn't a 100% perfect solution, maybe we can improve the current one:

1. I think there is definitely a bug that we currently singularize non-list fields:

With this:

```graphql
fragment User on SchemaUser {
  addresses(page: Int) {
    hasNextPage
    nodes {
      street
    }
  }
}
```

Codegen produces `val pullRequests: PullRequest`. I don't think we want that in all cases so we can remove that source of singularization already.

2. For list fields, if we can't avoid singularization, maybe we can add a simple rule that if part of the field name contains the type name modulo 's', we use that as singular. I think that would fix most of your issues ? 

3. If that's not enough, users can control the generated type names using aliases and we can document that. In that case, I guess we shouldn't apply singularization to the alias ?

ping @designatednerd @sav007

sav007:
@EricKuck 
> I see no reason Apollo uses field names instead of the schema type names,

That't correct, you can have several different fields with the same type. On top of that we should consider field aliases.

It looks like there is no perfect solutions here, as for some cases we need inflector for others we don't.

We could use Graphql directive https://graphql.org/learn/queries/#directives

So for instance:
```
query {
   pies @singular(word: "pie")
   ctas @singular(word: "cta")
}
```

To provide hint for the code generation but that requires user to alter GraphQL schema to include `singular` directive spec.

Another crazy idea,  we can think of, use special comment:
```
query {
   pies #${singular: "pie"}
   ctas #${singular: "cta"}
}
```

Or we can return back to have gradle configuration.

tasomaniac:
How does iOS handle this?

designatednerd:
> How does iOS handle this?

@tasomaniac Poorly :trollface: . Right now we're using a javascript singularization method but [it's also doing weird things](https://github.com/apollographql/apollo-ios/issues/955). I tried ripping out the JS stuff but as you've seen, it makes a bunch of stuff really odd in terms of `List<Edges>` and the like. 

I like the idea of allowing customizable singularization. I'm looking at potentially using something like [InflectorKit](https://github.com/mattt/InflectorKit), which has a bunch of common singularizations but allows for customizability, as part of the Swift codegen rewrite. Is there something similar out there in Kotlin/Java that you might be able to call from the gradle script?

martinbonnin:
> Is there something similar out there in Kotlin/Java that you might be able to call from the gradle script?

I found a few Java libs but with multiplatform around the corner, we should certainly refrain from adding more Java dependencies. 

Also, as I said above, compared to a generic Inflector lib, we have more context because we have the schema type name in addition to the field name. Type name will be singular and field name plural most of the times. Maybe we can use that to increase the precision without adding an external lib ?

EricKuck:
FWIW InflectorKit is extremely simple. It's very close to what already exists in this project, just with more/better rules. https://github.com/mattt/InflectorKit/blob/master/InflectorKit/TTTStringInflector.m

How about this for a game plan:
- Disable singularization for non-lists (no flag, just disable it for everyone, since it's a bug)
- Update the inflector in this project to match InflectorKit rules (and therefore match Apollo iOS)
- See what happens next? If there are still issues, add a gradle config option like my previous PR or somehow integrate schema type names like @martinbonnin suggested.

designatednerd:
>  (and therefore match Apollo iOS)

To be clear, this won't match iOS yet, this is something i'm working on.

EricKuck:
@designatednerd After looking into InflectorKit some more, I'd probably recommend against using it in the iOS lib. It's pretty out of date and still causes some pretty obvious issues (slice -> slouse for example) that have been fixed upstream. InflectorKit is based on [Rails's inflector](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/inflections.rb#L35), which has much more up to date regexes. When I open the PR with these changes I'll just grab the latest from Rails.

PR #1882, opened 2020-01-03
[Incubating plugin] Add gradle build cache test

martinbonnin:
~This PR is dependent on https://github.com/apollographql/apollo-android/pull/1849. Do not merge until https://github.com/apollographql/apollo-android/pull/1849 is merged.~ 
This is ok to merge now.

Add a test to check that build cache (https://docs.gradle.org/current/userguide/build_cache.html) is working correctly with the customIdGenerators.

### DIFF
--- a/apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/OperationIdGeneratorTests.kt
+++ b/apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/OperationIdGeneratorTests.kt
@@ -41,19 +41,6 @@ class CustomIdGeneratorTests {
 
   @Test
   fun `changing the operationIdGenerator recompiles sources`() {
-    val apolloConfiguration = """
-      class MyIdGenerator implements OperationIdGenerator {
-          String apply(String operationDocument, String operationFilepath) {
-              return operationDocument.length().toString();
-          }
-          String version = "MyIdGenerator-v1"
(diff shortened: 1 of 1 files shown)


tasomaniac:
Using a custom build cache folder is a good idea. I didn't know that this was possible. 👍 

I wonder if we can find a generic way to test if build cache is working. We could have one test without even configuring the plugin with custom props. That would be easy.

martinbonnin:
> I wonder if we can find a generic way to test if build cache is working. We could have one test without even configuring the plugin with custom props. That would be easy.

Not sure I get it. What are custom props here? `apollo. customIdGenerator ` ? This isn't really required but the customIdGenerator triggered the discussion about build cache so I decided to add the test in this context. It can certainly be added for other tests too. Or did you want to add it systematically for all tests ?

tasomaniac:
Sorry my comment here was just brain-storming and was out of scope of this PR. I was wondering if it would be possible to write (smoke) tests for build cache. They would run over PRs and detect any accidental build cache breakages. It is just an idea.

martinbonnin:
We could but I don't think build cache would break a lot. While the gradle plugin has some complexity, it should be self contained enough so that it doesn't break too much?

martinbonnin:
I'm merging this one because it's a small one and I want to open a PR to promote the incubating plugin and this would conflict. Let me know you want me to change anything.

tasomaniac:
All good. I meant to approve it actually. 👍

PR #1888, opened 2020-01-07
Singularization fixes

EricKuck:
Picking up where #1880 left off…

Since updating to the kotlin codegen, we've run into lots of issues with bizarre singularization in our project. Examples include Pies -> Py, Ctas -> Ctum, and Slice -> Slouses. Another user opened an issue to report this problem a while back: #1686.

This PR addresses this issue in 2 ways, neither of which are guaranteed to work for all cases. For our project, all 3 instances of bad singularization have been fixed by this PR. Two were fixed by disabling singularization on non-arrays and one was fixed by updating the singularization rules.

1) Singularization was disabled for everything that is not an element of an array. It seems like it was a bug that these classes were ever singularized in the first place.
2) Singularization rules were synced from ruby on rails's inflector, which seems to be where the rules in this project came from in the first place (via rogueweb).

Note: While I was working on Inflector.kt, I also did some minor optimizations. Now arrays will only be created once and each regex will only be compiled once.

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
+++ b/apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
@@ -117,7 +117,7 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
     }
   }
 
-  data class AllStarship(
+  data class AllStarships1(
     val __typename: String,
     /**
      * A list of edges.
@@ -139,7 +139,7 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
           ResponseField.forList("edges", "edges", null, true, null)
           )
 
-      operator fun invoke(reader: ResponseReader): AllStarship {
+      operator fun invoke(reader: ResponseReader): AllStarships1 {
         val __typename = reader.readString(RESPONSE_FIELDS[0])
         val edges = reader.readList<Edge>(RESPONSE_FIELDS[1]) {
           it.readObject<Edge> { reader ->
@@ -147,7 +147,7 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
           }
 
         }
-        return AllStarship(
+        return AllStarships1(
           __typename = __typename,
           edges = edges
         )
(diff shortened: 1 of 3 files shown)


apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
```diff
@@ -117,7 +117,7 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
     }
   }
 
-  data class AllStarship(
+  data class AllStarships1(
```
martinbonnin: Interesting that it suffixes the class now. I think it should be ok without the `1` since it's nested ? 

```
class AllStarships {
    class AllStarships {
    }
}
```

That should be valid, right ?

martinbonnin:
Thanks ! One thing we should keep in mind is that this is going to break consumer code. I'd say we should at least keep this for 1.3.x (that will also break some thing around plugin configuration) and make a warning in the README.md.

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
EricKuck: Seems like it should be, yeah. This should probably be fixed with the 1.3.0 release so you don't risk doing 2 consecutive breaking changes releases. That being said, it's unrelated to this PR so I won't fix it here.

martinbonnin: Sounds good. We should just remember to fix it in 1.3.0. Maybe it's time we document the coming breaking changes somewhere so we don't forget anything...

martinbonnin: 1.3.0 release checklist there : https://github.com/apollographql/apollo-android/issues/1889

sav007: It's ok for Kotlin, but not for Java, this is not valid in Java:
```
public class Test {
  static class Test {
    
  }
}
```

We might need to relax our Kotlin version of class name clash resolution and don't include top class into used keywords. But for Java we can't do the same.

martinbonnin: Thanks for the explanation!

Issue #1883, opened 2020-01-03
Java and Kotlin code generation ignores `@include`


morganlutz:
**Summary**
I am trying to upgrade from `1.0.0` to the latest version of apollo-android, `1.2.2`. However, the Java and Kotlin generated code seem to ignore the `@include` logic that I have defined in my fragment. The full fragment definition is included in the description.

The fragment optionally includes extra details using `@include(if: withDetails: Boolean)`. In the feed I don't need the extra details. In a detail page I do need the extra details. Some of the extra details are marked as `NON_NULL` in the schema, because if I request them then I should receive them.

Previously, the Java generated code marked these additional details as `@Nullable` regardless of whether `NON_NULL` was present in the schema. Now, it marks these objects as `@NotNull`.

I saw that apollo-android supports Kotlin now, so I tried migrating to Kotlin code generation to see if it was any different, but the extra details in the generated data class are non-null instead of optional. Whenever I make a call to fetch the object without the extra details, it throws an exception / fails to parse the http response.

I think this is a bug since it was working previously and the same fragment definition works well in the iOS app. If it's not a bug - What is the right way to define a fragment to optionally include attributes? Or, is there a recommended work-around?

**Description**
`@include` documentation: https://graphql.org/learn/queries/#directives
Schema - https://api.kivaws.org/graphql

Fragment definition:

```
fragment LoanFragment on LoanBasic {
    id
    status
    name
    use
    loanAmount
    plannedExpirationDate
    loanFundraisingInfo {
        fundedAmount
        reservedAmount
    }
    image {
        url(presetSize: loan_retina)
    }
    geocode {
        country {
            isoCode
            name
        }
    }
    ... on LoanBasic @include(if: $withDetails) {
        raisedDate
        description
        whySpecial
        activity {
            name
        }
        sector {
            id
            name
        }
        hasCurrencyExchangeLossLenders
        terms {
            lenderRepaymentTerm
        }
        repaymentInterval
        paidAmount
        userProperties {
            lentTo
        }
        isMatchable
        matcherName
    }
    ... on LoanPartner @include(if: $withDetails) {
        partnerName
        partner {
            id
            image {
                url(customSize: "h300")
            }
            loanAlertText
            chargesFeesInterest
            riskRating
        }
    }
}
```

A specific example using `Sector`, which we do not fetch in the feed, but do fetch on a detail page:

Version `1.0.0` generated Java code:

```
  /**
   * The sector is a more general classification of the loan than activity
   */
[...]

PR #1891, opened 2020-01-09
Add support directives to inline fragments

sav007:
- Address issue with directives with inline fragments
- Simplify and unify code generation for  regular fragments and inline fragments
- Address issue with fragment serialization in `ResponseWriter`

Closes https://github.com/apollographql/apollo-android/issues/1883


martinbonnin:
Should we move `ResponseField`, `ResponseReader` and `ResponseWriter` to some `internal` package ? This PR changes the API of these classes and I guess it's not a big deal because there is very little use case to use them from client code?

Are there any valid use cases though ? If not, we could make it explicit that this is only to be used by internal code ?

sav007:
Yeah I guess the should be internal, don't quite remember why these classes were exposed as public.

sav007:
Will move these classes to internal sub-package in my next PR

apollo-api/src/test/java/com/apollographql/apollo/api/internal/SimpleResponseReaderTest.java
```diff
@@ -95,7 +95,7 @@
     }
   }
 
-  @Test public void readDouble() {
+  @Test public void readD1ouble() {
```
tasomaniac: typo here.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ResponseFieldSpec.kt
```diff
@@ -212,40 +230,26 @@ class ResponseFieldSpec(
...
-            .returns(FRAGMENTS_CLASS)
-            .addParameter(ParameterSpec.builder(String::class.java, CONDITIONAL_TYPE_VAR).build())
-            .addParameter(RESPONSE_READER_PARAM)
-            .addStatement("return \$L.map(\$L, \$L)", FRAGMENTS_CLASS.mapperFieldName(),
-                RESPONSE_READER_PARAM.name, CONDITIONAL_TYPE_VAR)
-            .build())
-        .build()
-    return CodeBlock.of("final \$T \$L = \$L.\$L(\$L, \$L);\n", normalizedFieldSpec.type, fieldSpec.name,
-        readerParam, READ_METHODS[responseFieldType], fieldParam, readerTypeSpec)
+//  private fun readFragmentCode(readerParam: CodeBlock, fieldParam: CodeBlock): CodeBlock {
```
tasomaniac: should this be deleted?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/SchemaTypeSpecBuilder.kt
```diff
@@ -433,18 +449,18 @@ class SchemaTypeSpecBuilder(
   }
 
   private fun inlineFragmentsResponseMapperSpec(nameOverrideMap: Map<String, String>,
```
tasomaniac: `codestyle`: Looks like here, it would be better to put the first param in a new line

PR #1907, opened 2020-01-14
Use the correct OperationIdGenerator for OperationOutput

tasomaniac:
Use the correct OperationIdGenerator for OperationOutput

Since only used in single place, move sha256 implementation and make it private

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
@@ -30,7 +30,7 @@ interface OperationIdGenerator {
       return operationDocument.sha256()
     }
 
-    override val version = "1.0"
+    override val version = "sha256-1.0"
 
     private fun String.sha256(): String {
       val bytes = toByteArray(charset = StandardCharsets.UTF_8)
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.compiler
 
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+
 interface OperationIdGenerator {
   fun apply(
       /**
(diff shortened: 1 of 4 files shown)


tasomaniac:
@Laimiux After your contribution, a custom hashing (operation id) support was added in #1849. Can you have a look if this looks correct?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
```diff
@@ -22,11 +25,18 @@ interface OperationIdGenerator {
    */
   val version: String
 
-  class Sha256: OperationIdGenerator {
+  class Sha256 : OperationIdGenerator {
     override fun apply(operationDocument: String, operationFilepath: String): String {
       return operationDocument.sha256()
     }
 
     override val version = "1.0"
```
tasomaniac: Should we have a name like `Sha256-1.0`?

I feel like `1.0` is really generic and another algoritm should not really have the same version.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/operationoutput/OperationOutputWriter.kt
```diff
@@ -1,27 +1,24 @@
...
+import com.apollographql.apollo.compiler.OperationIdGenerator
 import com.apollographql.apollo.compiler.ir.CodeGenerationIR
 import com.apollographql.apollo.compiler.ir.Operation
-import com.apollographql.apollo.compiler.sha256
 import com.apollographql.apollo.internal.QueryDocumentMinifier
 import java.io.File
-import com.squareup.moshi.Moshi
-import com.squareup.moshi.Types
 
-internal class OperationOutputWriter(val packageNameProvider: PackageNameProvider) {
```
tasomaniac: `packageNameProvider` was never used.

Laimiux:
Looks good to me!

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
martinbonnin: +1. We should definitely encourage unique version names.

tasomaniac: Just pushed a commit. Plz have a look.

Issue #1906, opened 2020-01-14
[Incubating Plugin] Duplicate classes in tests when using `graphqlSourceDirectorySet`


martinbonnin:
The incubating plugin allows to specify the schema and graphql files using something like this:

```kotlin
apollo {
   schemaFile = ...
   graphqlSourceDirectorySet.srcDirs(...)
}
```

It's nice and concise but it applies the same configuration to all variants including test variants so we end up with duplicates classes when compiling tests.

Two possible solutions would be:

1. Remove `graphqlSourceDirectorySet` from `CompilerParams` and only expose it from `CompilationUnit`. It means users would have to do more complex configuration in the general case:

```kotlin
 apollo {
   onCompilationUnits {
        if (name == 'mainService') { // if we don't do this, Apollo will generate classes also for `test`s
            schemaFile = ...
            graphqlSourceDirectorySet.srcDirs += ...
            graphqlSourceDirectorySet.include("**/*.graphql")
        }
    }
}
```

2. Do some special handling for tests and only apply the top level configuration to non-test variants but it breaks a bit the consistency of the API

```kotlin
apollo {
   schemaFile = ...
   graphqlSourceDirectorySet.srcDirs(...)

   test {
      schemaFile = ...
      graphqlSourceDirectorySet.srcDirs(...)   
   }
}
```

PR #1910, opened 2020-01-15
[Incubating Plugin] Added various tests for compilationUnit override support

tasomaniac:


apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
```diff
@@ -351,6 +351,59 @@ class ConfigurationTests {
     }
   }
 
+  @Test
+  fun `sources can be overridden in compilationUnits`() {
+    withSimpleProject("""
+      apollo {
+        schemaPath = "com/some/other/schema.json"
```
martinbonnin:
```suggestion
        schemaFile = file("com/some/other/schema.json")
```

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
```diff
@@ -351,6 +351,59 @@ class ConfigurationTests {
...
+        sourceFolder = "com/some/other"
```
martinbonnin: `sourceFolder` does not exist on the extension, it's only in services

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
tasomaniac: Changing to this actually makes the test go green. I am not even sure if we have a bug to fix. 🤷‍♂️

apollo-gradle-plugin-incubating/src/test/kotlin/com/apollographql/apollo/gradle/test/ConfigurationTests.kt
tasomaniac:
```suggestion
```

tasomaniac:
@martinbonnin the issues I had (that led me work on these tests) were related to #1906 

Even putting `if (variantName == 'main')` and similar didn't work in our project last night :) They actually work now. That's why I'm not sure if we have a bug for this. 

Still, I think these tests can be merged since they're not bad.

PR #1914, opened 2020-01-16
Documentation update for 1.3.0

martinbonnin:
This adds a 'Migrating to 1.3.0' section to the main README.md and updates the documentation about the plugin.


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
```diff
@@ -15,75 +15,105 @@ import org.gradle.api.provider.Provider
...
    * suppressRawTypesWarning will add the appropriate SuppressWarning annotation
    *
-   * false by default
+   * Default value: false
    */
   val suppressRawTypesWarning: Property<Boolean>
 
   /**
-   * Whether to suffix your queries, etc.. with `Query`, etc..
+   * When true, Apollo-Android will make sure all the generated classes end with 'Query' or 'Mutation'.
```
tasomaniac: Small: I know the whole project is still called Apollo Android but it would be better to avoid Android in docs if it is not Android specific. What about just saying Apollo? Or you can go passive and say "all generated classes will end with ..."

martinbonnin:
I tried to make the migrating documentation a bit less scary. It's hard to give a general migration as all changes are mostly independent so in the end we'd have to explain all of the changes. i.e. someone might be impacted by singularization and not multi-service or the other way round or maybe Kotlin DSL, it's hard to tell in advance.

tasomaniac:
The last added clarifying sentence is awesome.

PR #1916, opened 2020-01-18
Generate Operation.Data toJson implementation

### LINKED ISSUE #1881: Add toJson method for writing response objects to strings
I'm trying to add a "mock" mode to my app where the server logic is implemented in-app.  I've got a mock okhttp Client built but I would like to be able to automatically turn the generated response objects into json strings to put in the response body.  I can write my own `ResponseWriter` but that feels janky and unnecessary.

```
fun ResponseWriter.toJson(): String?
```

Let me know if this is something you'd be amenable to and I can make a PR.  Or if it's already possible and I'm just not seeing the API.

sav007:
Introduce new `toJson` API to make it easier to serialize `Operation.Data` response into Json.

Example of how to dump response data into JSON format:
```
  Response<AllPlanetsQuery.Data> response = ...;
  String json = response.data().toJson("  ");
```

Closes #1881


apollo-api/src/main/java/com/apollographql/apollo/api/Operation.java
```diff
@@ -78,6 +78,59 @@
...
+     *        }
+     *      }
+     *    }
+     * }</pre>
+     *
+     * @param indent the indentation string to be repeated for each level of indentation in the encoded document. Must be a string
+     * containing only whitespace,
+     * @return json representation of GraphQL operation response
+     */
+    String toJson(@NotNull String indent);
```
tasomaniac: Would it make sense to put this first since the other one overloads this function?

apollo-compiler/src/test/graphql/com/example/union_inline_fragments/TestQuery.kt
```diff
@@ -410,6 +411,14 @@ class TestQuery : Query<TestQuery.Data, TestQuery.Data, Operation.Variables> {
       }
     }
 
+    override fun toJson(indent: String): String = toJson(indent, DEFAULT)
+
+    override fun toJson(indent: String, scalarTypeAdapters: ScalarTypeAdapters): String {
```
tasomaniac: I wonder if there is a way to move this implementation to somewhere we can share so that we don't have to duplicate in every class generated.

If we target Java8, then we could use default method implementations.

tasomaniac:
That sounds like a good approach. We already have setup for Kotlin extension methods. Static method for Java with additional Kotlin extension would be nice.

sav007:
@martinbonnin you are right, `toJson` isn't really "must have" API, it's more like convenient or nice to have feature and making it as utility function totally makes sense to me.

We can come up with utility class `OperationResponseJsonSerializer` (maybe there is a better name, open to any suggestions):
```
public final class OperationResponseWriter {
  private constructor() { }
  
   public static String toJson(@NotNull Operation.Data data, @NotNull String indent) {
    return toJson(data, indent, ScalarTypeAdapters.DEFAULT);
  }

  public static String toJson(@NotNull Operation.Data data, @NotNull String indent, @NotNull ScalarTypeAdapters scalarTypeAdapters) {
    ....
  }
```

martinbonnin:
Maybe `OperationDataJsonSerializer.serialize(operationData, "    ")` ? That'll keep the same philosophy as `QueryDocumentMinifier.minify(queryDocument)`

PR #1922, opened 2020-01-21
toJson kotlin extension

tasomaniac:
Add `toJson` Kotlin extension for Kotlin users. I didn't go for `serialize` since as an extension, that is not super clear. `data.toJson` looks nicer.

Doc updates:
- Add documentation about the new `OperationDataJsonSerializer`.
- Split `no-runtime` doc into multiple sections.
- Also add new Kotlin extension in 
there

### DIFF
--- a/docs/source/advanced/no-runtime.md
+++ b/docs/source/advanced/no-runtime.md
@@ -10,13 +10,15 @@ For this, remove the `com.apollographql.apollo:apollo-runtime`dependency and rep
   implementation("com.apollographql.apollo:apollo-api:x.y.z")
 ```
 
+## Parsing HTTP body
+
 All `Operation` instances provide an API to parse `Response` from raw `okio.BufferedSource` source that represents http response body returned by the GraphQL server.
 If for some reason you want to use your own network layer and don't want to use fully featured `ApolloClient` provided by `apollo-runtime` you can use this API:
 
 ```java
     okhttp3.Response httpResponse = ...;
 
-    Response<Query.Data> response = new Query().parse(httpResponse.body().source());
+    Response<Operation.Data> response = new Query().parse(httpResponse.body().source());
 ```
 
 If you do have custom GraphQL scalar types, pass properly configured instance of `com.apollographql.apollo.response.ScalarTypeAdapters`:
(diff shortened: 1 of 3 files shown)


docs/source/advanced/no-runtime.md
```diff
@@ -10,13 +10,15 @@ For this, remove the `com.apollographql.apollo:apollo-runtime`dependency and rep
...
+## Parsing HTTP body
+
 All `Operation` instances provide an API to parse `Response` from raw `okio.BufferedSource` source that represents http response body returned by the GraphQL server.
 If for some reason you want to use your own network layer and don't want to use fully featured `ApolloClient` provided by `apollo-runtime` you can use this API:
 
 ```java
     okhttp3.Response httpResponse = ...;
 
-    Response<Query.Data> response = new Query().parse(httpResponse.body().source());
+    Response<Operation.Data> response = new Query().parse(httpResponse.body().source());
```
tasomaniac: I thought operation here would be more correct. `Query` was referring to a generated type, right?

martinbonnin: I think `Query` was referring to https://github.com/apollographql/apollo-android/blob/65c4fff8b98f9c8f5949adb6a3b2e694489dfee0/apollo-api/src/main/java/com/apollographql/apollo/api/Query.java.

But this all should work with Mutation as well so using `Operation.Data` seems better 👍

docs/source/advanced/no-runtime.md
```diff
@@ -26,9 +28,31 @@ If you do have custom GraphQL scalar types, pass properly configured instance of
 
     ScalarTypeAdapters scalarTypeAdapters = new ScalarTypeAdapters(<provide your custom scalar type adapters>);
 
-    Response<Query.Data> response = new Query().parse(httpResponse.body().source(), scalarTypeAdapters);
+    Response<Operation.Data> response = new Query().parse(httpResponse.body().source(), scalarTypeAdapters);
+```
+
+## Converting Query.Data back to JSON
```
martinbonnin: Does that belong to the `no-runtime` doc ? I think the original use case was for testing with a mock server. Not sure where to put it though... 😅

tasomaniac: Suggestions are welcome. All of the stuff mentioned in this doc are in api module rather than runtime. I didn't really want to create a new file.

tasomaniac: Should I move out?

martinbonnin: I don't really have a better idea where to put this doc. Maybe at some point do a 'testing' page ? But I guess `no-runtime` is fine for now, sorry for the noise.

tasomaniac: Not at all. It's useful feedback.

PR #1933, opened 2020-01-24
Remove bintray plugin and upload marker to bintray

martinbonnin:
It turns out that the `maven-publish` plugin is very capable of uploading to bintray. Use that and remove a dependency.

I also took the opportunity to upload the [Gradle plugin marker](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_markers) to ~jcenter~ bintray. I made it manually for 1.3.0. It should be done by the CI for next versions.
Promotion to jcenter is not possible at the moment because "there's no binary file, aar, jar, ... in the package". I emailed bintray support about it.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -37,5 +37,9 @@ else
   echo "Deploy to Gradle portal..."
   ./gradlew :apollo-gradle-plugin:publishPlugin -Pgradle.publish.key=$GRADLE_PUBLISH_KEY -Pgradle.publish.secret=$GRADLE_PUBLISH_SECRET
   echo "Deployed to Gradle portal!"
+  echo "Deploy Gradle plugin marker to bintray..."
+  # We override the artifact_id for the marker else it is uploaded at the same coordinates as apollo-gradle-plugin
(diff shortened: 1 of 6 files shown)


scripts/deploy.sh
```diff
@@ -32,7 +32,11 @@ if [ "$TRAVIS_TAG" == "" ]; then
   echo "Skipping release deployment: not a tag"
 else
   echo "Deploy to bintray..."
-  ./gradlew bintrayUpload -Pbintray.user="${BINTRAY_USER}" -Pbintray.apikey="${BINTRAY_API_KEY}"
+  ./gradlew publishDefaultPublicationToBintrayRepository -Pbintray.user="${BINTRAY_USER}" -Pbintray.apikey="${BINTRAY_API_KEY}"
   echo "Deployed to bintray!"
+  echo "Deploy gradle plugin marker to bintray..."
+  # We override the artifact_id for the marker else it is uploaded at the same coordinates as apollo-gradle-plugin
+  ./gradlew publishApolloGradlePluginPluginMarkerMavenPublicationToBintrayRepository -Pbintray.user="${BINTRAY_USER}" -Pbintray.apikey="${BINTRAY_API_KEY}" -PPOM_ARTIFACT_ID=com.apollographql.apollo.gradle.plugin
```
tasomaniac: Wouldn't it be possible to keep the artifact id in gradle files?

martinbonnin: Not really, the marker is a special case since it belongs to the `apollo-gradle-plugin` module.
The problem is that the repository depends on the publication and the `maven-publish` plugin is designed for these things to be independent. We could also duplicate the bintray repository, that would be an option but it'll require slightly more changes

Issue #1850, opened 2019-12-11
Support hash-only queries in GraphQL Subscriptions


ketulshah1947:
**Is your feature request related to a problem? Please describe.**
Currently, Apollo sends the hashed query in the "id" field and full-text query in the "query" field as part of WebSocket message (Reference: OperationClientMessage.Start.writeToJson()). At Wayfair, for security purposes, we want to avoid full-text queries and send only hashed ones. Is there any existing support available for it? If not, and if allowed, I would like to implement it.

**Describe the solution you'd like**
We can accept a parameter from the project Gradle file to decide whether to use a hashed-only query or not. Based on that field we can avoid full-text queries. So it will only send hash over WebSocket. Even if connection is not secure, the query will not be compromised.

sav007:
Not sure if this makes sense but you can try to use query auto persistence feature for this by setting `enableAutoPersistedQueries = true`, here is what it does:
- Apollo will try to send query without query document just with  `extensions { sha256Hash `
- It will send another request with query document only when the GraphQL server returns error such as `PersistedQueryNotFound` or `PersistedQueryNotSupported`

I guess in your case if your server doesn't have auto persisted query feature then technically it should work, though this workaround looks very hacky.

ketulshah1947:
@sav007 I tried but it only works in query and mutation. For subscription cases, it doesn't work. Do you have any insight on it?

sav007:
Yeah, subscriptions is not supported at the moment, sorry I missed this part in the issue description. I think we can add it. I can take a look after the holidays.

ketulshah1947:
@sav007 Sure. I can create PR if you like.

PR #1943, opened 2020-01-26
Add support for auto persisted subscriptions

sav007:
- Introduce new configuration flag to enable subscription auto persistence feature (it was intentionally introduced as a separate flag to avoid messing with regular subscriptions, though I'm not really confident if need it, we might use the same flag as for regular queries `enableAutoPersistedQueries`)
- Update `Start` protocol message to support auto persistence
- Handle subscription `protocol negotiation errors` and resend `Start` message with subscription document
- Tests.

Closes https://github.com/apollographql/apollo-android/issues/1850

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
@@ -1,5 +1,6 @@
 package com.apollographql.apollo.internal.subscription;
 
+import com.apollographql.apollo.api.Error;
 import com.apollographql.apollo.api.Response;
 import com.apollographql.apollo.api.ResponseFieldMapper;
 import com.apollographql.apollo.api.Subscription;
@@ -39,6 +40,8 @@ public final class RealSubscriptionManager implements SubscriptionManager {
   static final int CONNECTION_KEEP_ALIVE_TIMEOUT_TIMER_TASK_ID = 3;
   static final long CONNECTION_ACKNOWLEDGE_TIMEOUT = TimeUnit.SECONDS.toMillis(5);
   static final long INACTIVITY_TIMEOUT = TimeUnit.SECONDS.toMillis(10);
+  static final String PROTOCOL_NEGOTIATION_ERROR_NOT_FOUND = "PersistedQueryNotFound";
+  static final String PROTOCOL_NEGOTIATION_ERROR_NOT_SUPPORTED = "PersistedQueryNotSupported";
 
   Map<UUID, SubscriptionRecord> subscriptions = new LinkedHashMap<>();
   volatile SubscriptionManagerState state = SubscriptionManagerState.DISCONNECTED;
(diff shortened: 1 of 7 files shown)


sav007:
cc @ketulshah1947

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.java
```diff
@@ -62,26 +62,48 @@ public Init(@NotNull Map<String, Object> connectionParams) {
     private static final String JSON_KEY_QUERY = "query";
     private static final String JSON_KEY_VARIABLES = "variables";
     private static final String JSON_KEY_OPERATION_NAME = "operationName";
+    private static final String JSON_KEY_EXTENSIONS = "extensions";
+    private static final String JSON_KEY_EXTENSIONS_PERSISTED_QUERY = "persistedQuery";
+    private static final String JSON_KEY_EXTENSIONS_PERSISTED_QUERY_VERSION = "version";
+    private static final String JSON_KEY_EXTENSIONS_PERSISTED_QUERY_HASH = "sha256Hash";
```
kenyee: Should this just be queryHashMethod? It's not necessarily sha256

apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java
```diff
@@ -130,15 +130,15 @@ public OperationResponseParser(Operation<D, W, ?> operation, ResponseFieldMapper
...
+            return parseError(reader.toMap());
           }
         });
       }
     });
   }
 
   @SuppressWarnings("unchecked")
-  private Error readError(Map<String, Object> payload) {
+  public static Error parseError(Map<String, Object> payload) {
```
kenyee: Nice renaming of this method 🙂

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.java
sav007: that part is really interesting, GraphQL spec doesn't have any info regarding this. BTW I noticed the same issue with regular queries, we sending `sha256Hash ` but as you said it can be different (md5). As far as I remember Apollo server expects `sha256Hash`. 

What about your use case, what GraphQL server are you using? Are you using sha256 as hash? 

I guess the only option right now is to make it configurable.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.java
```diff
@@ -165,13 +170,16 @@ void doSubscribe(Subscription subscription, SubscriptionManager.Callback callbac
...
         subscriptions.put(subscriptionId, new SubscriptionRecord(subscriptionId, subscription, callback));
+
         if (state == SubscriptionManagerState.DISCONNECTED) {
           state = SubscriptionManagerState.CONNECTING;
           transport.connect();
         } else if (state == SubscriptionManagerState.ACTIVE) {
-          transport.send(new OperationClientMessage.Start(subscriptionId.toString(), subscription, scalarTypeAdapters));
+          transport.send(
+              new OperationClientMessage.Start(subscriptionId.toString(), subscription, scalarTypeAdapters, autoPersistSubscription,
+                  false)
```
tasomaniac: This looks little weird in terms of code style. Would be nice to leave this false alone

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.java
SaurabhKukreja: @sav007 . This part is already merged, but would it be better to just name it `hash` as it could be anything. 

We use md5

Issue #1936, opened 2020-01-24
Add default value for __typeName in generated Kotlin model objects


edenman:
```
  data class MyEntity(
    val __typename: String,
    val id: Int
  )
```
For creating mock response objects, it would be helpful to not have to specify the `__typename`, especially since Fragments are now verifying the `__typename` when parsing the json.  Any chance we could update the codegen to give a default value for this field?  I'm happy to do the PR if this would be acceptable

martinbonnin:
I bumped into that as well. I was using `POSSIBLE_TYPES[0]` but this is not available in 1.3.0 anymore.

sav007:
@martinbonnin is there any other use case where you used `POSSIBLE_TYPES `?

martinbonnin:
Creating mock objects was the only use case where I used `POSSIBLE_TYPES`. It felt a bit like a hack so I don't mind too much seeing it go away but having a way to get the `__typename` of a model would be nice.

edenman:
How would folks feel about just removing it from the constructor entirely? Doesn’t seem like you’d ever actually want to pass in a different value

sav007:
We can't just remove it as it heavily used in normalized cache. But we can definitely make it optional.

tasomaniac:
I think @edenman meant removing only from the constructor. It would still stay as a property and would be part of de/serialization.

edenman:
(finally updated and wanted to say my mock mode implementation thanks you for this change!)

PR #1944, opened 2020-01-26
Generate default value for __typename field

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1936

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
+++ b/apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
@@ -71,7 +71,7 @@ data class TestQuery(
   override fun parse(source: BufferedSource): Response<Data> = parse(source, DEFAULT)
 
   data class HeroWithReview(
-    val __typename: String,
+    val __typename: String = "Human",
     /**
      * What this human calls themselves
      */
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/FragmentTypeBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/FragmentTypeBuilder.kt
@@ -11,7 +11,7 @@ import com.apollographql.apollo.compiler.ir.FragmentRef
 internal fun Fragment.ast(context: Context): ObjectType {
   val typeRef = context.registerObjectType(
       name = fragmentName.capitalize().escapeKotlinReservedWord(),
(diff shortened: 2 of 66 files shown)


apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
```diff
@@ -71,7 +71,7 @@ data class TestQuery(
   override fun parse(source: BufferedSource): Response<Data> = parse(source, DEFAULT)
 
   data class HeroWithReview(
-    val __typename: String,
+    val __typename: String = "Human",
```
tasomaniac: This is not supposed to be changed right? What about moving it out from the constructor and keep it as a field.

martinbonnin: Apparently this is still required for the cache: https://github.com/apollographql/apollo-android/issues/1936#issuecomment-578431702

I wonder if this is a technical limitation or if, for the same generated model, there's a use case for passing different values ? Maybe with inheritance ?

martinbonnin:
Looks good to me, thanks!
One question: would this be needed for java codegen as well or is there another way to do it in java ?

edenman:
1) I would definitely prefer the value just be removed from the constructor entirely, and just hardcode the value in the codegen.  During deserialization from json you could still confirm that the `__typename` is there in the json and that it has the expected value, and throw an exception if it's different?
2) This could certainly be done in Java too if we take this approach.  Harder to do if we want to keep it part of the constructor, would just have to be a constructor overload that omits that parameter.

Let me know if I can help!

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
sav007: It can be different. Right now the default value is taken from the definition of GraphQL schema type. But in case of interface the real value for `__typename` depends on the returned instance.
For example having this query:
```
query {
  hero {
    __typename
    name
  }
}
```

Object `hero` is defined as  `Character` in GraphQL schema. So default value will be `"Character"`. But at runtime it can be one of  possible types `Human` or `Droid` https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/schema.json#L637.

sav007:
@edenman pls see my comment above why it can't be removed from the constructor.

I will make changes to the Java code gen as well.

edenman:
Ah. Makes sense, thanks!

PR #1947, opened 2020-01-27
Very minor documentation updates

louistsaitszho:


martinbonnin:
I might have closed this one a bit too fast actually. Would'nt it make sense to have the method called `onCompilationUnit` instead of `onCompilationUnits` ?

louistsaitszho:
@martinbonnin maybe, but `onCompilationUnits` is what the gradle plugin understands.

martinbonnin:
What I'm saying is that maybe we should fix the plugin instead of the doc ? Having a plural `compilationUnits` with a single `CompilationUnit` parameter sounds a bit weird.

louistsaitszho:
Sounds reasonable, but for those who are using 1.3.0 now, it would be better to have the documentation that doesn't trick them. Will have that documentation update included in that PR.

Issue #1927, opened 2020-01-22
Gradle error from generateApolloClasses


yusufa2z:
**Summary**
When I follow the steps in the readme, the generateApolloClasses step fails.

**Description**
The step 
> ./gradlew :generateApolloClasses

Gives the error 

> Task 'generateApolloClasses' not found in root project 'My Application'.

When I run
> ./gradlew tasks

It shows generateApolloClasses as one of the available Apollo tasks. If I then run Gradle with the --info option, it gives some complaint about an enum:
> Analytics other plugin to proto: Unknown plugin type com.apollographql.apollo.gradle.ApolloPlugin expected enum COM_APOLLOGRAPHQL_APOLLO_GRADLE_APOLLOPLUGIN

**Version**
Apollo Plugin version 1.2.3
Gradle version 6.1-rc-1

tasomaniac:
Can you try without colon before the task name?

martinbonnin:
The Analytics warning is unrelated. I have it as well, it seems to come from the Android Gradle Plugin: https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/profile/AnalyticsUtil.java?autodive=0%2F#383

ys1382:
@tasomaniac you are right, removing the colon fixed the issue, although the colon is in the readme. Perhaps it is a change for Gradle 6?

martinbonnin:
That was a mistake, sorry. Fix is there: https://github.com/apollographql/apollo-android/pull/1950/files

tasomaniac:
@ys1382 colon prefix means root project in Gradle. It is not a change with Gradle 6. It basically depends on how you setup the project. Without colon, it searches for the given task everywhere.

PR #1950, opened 2020-01-27
Remove colon in README.md

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/1927#issuecomment-578908932

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ apollo {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tasomaniac:
I think we may further clarify. Just building the project automatically run this task since the plugin setups the necessary dependencies.

This task is good to run explicitly to verify that it works.

PR #1951, opened 2020-01-27
Rename all singular CompilationUnit references to singular

louistsaitszho:
From #1947: rename almost all references of `CompilationUnit` to singular.

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
@@ -33,11 +33,11 @@ interface ApolloExtension: CompilerParams {
 
   /**
    * [CompilationUnit] are created by the plugin based on the registered [Service]s and variants.
-   * [onCompilationUnits] allows to retrieve and configure them.
+   * [onCompilationUnit] allows to retrieve and configure them.
    *
    * @param action: the configure action for the [CompilationUnit]
    */
-  fun onCompilationUnits(action: Action<CompilationUnit>)
+  fun onCompilationUnit(action: Action<CompilationUnit>)
 
   /**
    * @Deprecated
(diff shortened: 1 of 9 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -15,13 +15,13 @@ open class DefaultApolloExtension(val project: Project)
   val services = mutableListOf<DefaultService>()
 
   /**
-   * compilationUnits is meant to be consumed by other gradle plugin.
+   * compilationUnit is meant to be consumed by other gradle plugin.
    * The apollo plugin will add the {@link CompilationUnit} as it creates them
    */
-  internal val compilationUnits = project.container(CompilationUnit::class.java)
+  internal val compilationUnit = project.container(CompilationUnit::class.java)
```
tasomaniac: This one is a container (which extends from Set) and internal. I think we should keep this one plural.

martinbonnin: +1

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -15,13 +15,13 @@ open class DefaultApolloExtension(val project: Project)
...
 
-  override fun onCompilationUnits(action: Action<CompilationUnit>) {
-    compilationUnits.all(action)
```
martinbonnin: What about keeping the old one and doing:

```
 fun onCompilationUnits(action: Action<CompilationUnit>) {
    throw Exception("please use onCompilationUnit {} instead")  
}
```

And then we remove that old version in a couple of months ?  This API is new from 1.3.0 so presumably not that many people are using it but that'll make a nicer deprecation path for the users that use it.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloExtension.kt
```diff
@@ -39,6 +39,14 @@ interface ApolloExtension: CompilerParams {
    */
   fun onCompilationUnit(action: Action<CompilationUnit>)
 
+  /**
+   * @Deprecated
+   *
+   * Please see [onCompilationUnit]
+   */
+  @Deprecated("There is a typo. Please use onCompilationUnit instead", replaceWith = ReplaceWith("onCompilationUnit"))
```
martinbonnin: Nice :+1:

PR #1958, opened 2020-01-29
Avoid field names clash

### LINKED ISSUE #1953: Android Apollo (GraphQL) autogeneration issue with Kotlin
I used to use an Android library Apollo (GraphQL) version 1.2 with Java models autogeneration. Today I updated to version 1.3 with possibility to generate Kotlin (model) data classes. But I have got an issue. In the Schema I have an Integer parameter:

```
...
{
   "name": "size",
   "description": "",
   "type": {
      "kind": "SCALAR",
      "name": "Int",
      "ofType": null
   },
   "defaultValue": null
}
...
```
Apollo generates code, where it uses Kotlin synthetic sugar by lambda statement "apply{}" on List parameter. Inside it applies to a property of generated class "size", but inside "apply{}" the "size" is the property of the "apply{}" context, not of the generated class. And, of course, "size" does not have "size.defined" and "size.value":

```
data class MembersQuery(
  val mandatorNo: Int,
  val memberNoFilter: Input<List<Int?>> = Input.absent(),
  val memberId: Input<String> = Input.absent(),
  val page: Int,
  val size: Input<Int> = Input.absent()
) : Query<MembersQuery.Data, MembersQuery.Data, Operation.Variables> {
  @Transient
  private val variables: Operation.Variables = object : Operation.Variables() {
    override fun valueMap(): Map<String, Any?> = mutableMapOf<String, Any?>().apply {
      this["mandatorNo"] = mandatorNo
      if (memberNoFilter.defined) this["memberNoFilter"] = memberNoFilter.value
      if (memberId.defined) this["memberId"] = memberId.value
      this["page"] = page
      if (size.defined) this["size"] = size.value
    }
...
}
```
Roll back to java generation is not good idea as I have changed many things, expecting of the Kotlin magic. Change the generated class does not have sense, as it will be regenerated and rewritten. What to do in the case?

sav007:
- Fix issue query argument with name `size` clashing with `Map.size`
- Fix potential field names clash with lambda arguments such as `writer`, 'listItemWriter', 'value' by prefixing last ones with `_`

Closes #1953

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
+++ b/apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
@@ -47,10 +47,12 @@ data class TestQuery(
       this["greenValue"] = this@TestQuery.greenValue
     }
 
-    override fun marshaller(): InputFieldMarshaller = InputFieldMarshaller { _writer ->
-      if (episode.defined) _writer.writeString("episode", episode.value?.rawValue)
-      _writer.writeInt("stars", stars)
-      _writer.writeDouble("greenValue", greenValue)
+    override fun marshaller(): InputFieldMarshaller = InputFieldMarshaller { writer ->
+      if (this@TestQuery.episode.defined) {
+        writer.writeString("episode", this@TestQuery.episode.value?.rawValue)
+      }
+      writer.writeInt("stars", this@TestQuery.stars)
+      writer.writeDouble("greenValue", this@TestQuery.greenValue)
     }
   }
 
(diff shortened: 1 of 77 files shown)


martinbonnin:
Am I correct by saying that if someone decides to name a field `_writer` it might clash again ? Or does the GraphQL spec forbids underscores ?

tasomaniac:
I have a similar question like Martin asked. I was actually going to take a look at this issue. I was thinking it can also be solved by not using `apply` and also by not letting scopes being mixed up when we use fields from generated code. Does that make sense?

martinbonnin:
I guess the 100% correct solution is to scope all fields (`this@...`) ? If it is, this is going to be quite verbose 😅

tasomaniac:
This issue would exist only in 2 scenario. Please correct me if I'm wrong. If we do `inner class` or use higher order functions (lambdas) with receiver (e.g `apply`). I would say let's just stop using apply and similar functions and problem will be mostly solved.

sneronovAdviqo:
When can we expect the fix?

sav007:
Yes you are both right about `_writer` can clash again, though it's less risky than it was before.
I will try go with Martin's suggestion and use `this` prefix.

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
```diff
@@ -42,15 +42,17 @@ data class TestQuery(
   @Transient
   private val variables: Operation.Variables = object : Operation.Variables() {
     override fun valueMap(): Map<String, Any?> = mutableMapOf<String, Any?>().apply {
```
tasomaniac: Just want to compare this without using `apply` here. I would look like:

```
override fun valueMap(): Map<String, Any?> {
  val map = mutableMapOf<String, Any?>()
  if (episode.defined) map["episode"] = episode.value
  map["stars"] = stars
  return map
}
```

sav007: well local variable `map` can collide if there is variable with the same name. To be 100% safe we need use `this@`

sneronovAdviqo:
On version 1.3.2 the same problem!

Issue #1955, opened 2020-01-28
[Codegen] Generated objects/fields on Interfaces/Implementations


aveiros:
**Summary**
When generating code for a Non-Null GraphQL Field and using concrete types, generated code for the field becomes optional.

**Description**
Consider the following schema

```
type Test {
   items: [Item!]!
}

interface Item {
    id: String!
}

type ItemA implements Item {
    id: String!
    a: String!
}

type ItemB implements Item {
    id: String!
    b: String!
}

type Query {
    getTest: Test
}
```

**Case 1**
If we don't specify the concrete types on the query
```
getTest {
     items {
        id
     }
}
```
The java code for the items field is
```
List<Item>
```

**Case 2**
Once we specify the the concrete types on the query
```
getTest {
     items {
        id
       ... on ItemA {
          a
       }
       ... on ItemB {
          b
       }
     }
}
```
The java code for the items field is
```
Optional<List<Item>>
```
Actual: a field (non null) is optional, once concrete types are specified
Expected: a field (non null) is not optional, once concrete types are specified

**Version**
v1.2.3

PR #1960, opened 2020-01-30
Field with inline fragment forced to be optional

sav007:
¯\_(ツ)_/¯ who knows why this condition was in place, maybe it became irrelevant after couple iterations.

Fix `inline_fragment_for_non_optional_field` fixture as it wasn't right before, this fixture was supposed to test this particular case.

Closes https://github.com/apollographql/apollo-android/issues/1955

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,11 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <option name="OTHER_INDENT_OPTIONS">
+      <value>
+        <option name="INDENT_SIZE" value="2" />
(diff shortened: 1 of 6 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -1,5 +1,11 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <option name="OTHER_INDENT_OPTIONS">
+      <value>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="TAB_SIZE" value="2" />
+      </value>
+    </option>
```
sav007: this changes somehow needed for having 2 spaces for any non source code files

PR #1971, opened 2020-02-05
Sanitize yield keyword in Kotlin

### LINKED ISSUE #1957: Identifier 'yield' is reserved
**Summary**
Error during build kotlin application: Identifier 'yield' is reserved. Use backticks to call it: `yield` 

**Description**
I have field called 'yield' in graphql query.
I generate sources using:	
./gradlew generateApolloSource
with generateKotlinModels=true

**Version**
v1.3.0

tasomaniac:
Fixes #1957 

For some reason `yield` is not mentioned in the below list. I verified that compilation actually fails with `yield`. That's why I also wanted to keep tests included for Kotlin sanitization.

https://kotlinlang.org/docs/reference/keyword-reference.html

I also double checked with "Soft keywords" mentioned in the 2nd half of the page. They are fine. We don't really need to escape those.


tasomaniac:
Also added `it` and `field` as reserved. Right now, those also conflict. With the latest PR from @sav007 we will be starting to use explicit names instead of `it` but still it would be good to reserve that too.

PR #1978, opened 2020-02-07
Use `KotlinProjectExtension` instead of the reflection magic to get a kotlin sourceDirectorySet

martinbonnin:
I'm not sure why I didn't do that in the first place. I guess I just did see that... :man_shrugging: 

In all cases, it's less code and more idiomatic.


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/JvmTaskConfigurator.kt
```diff
@@ -40,9 +39,9 @@ object JvmTaskConfigurator {
     }
 
     val sourceDirectorySet = if (!compilationUnit.generateKotlinModels()) {
```
tasomaniac: Little unrelated but I have a suggestion: Inverting this condition makes it hard to understand. I saw Kotlin in the else and confused because I didn't see the negation

martinbonnin: :+1: Indeed, I just swapped the branches.

tasomaniac: I love your commit message 😊

PR #1982, opened 2020-02-08
[Plugin] Fix compile task naming

martinbonnin:
Closes https://github.com/apollographql/apollo-android/issues/1981

The java plugin removes the dash when generating the task name, do the same in the apollo plugin.

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/GUtil.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/GUtil.kt
@@ -22,9 +22,6 @@ import java.util.regex.Pattern
 object GUtil {
   private val WORD_SEPARATOR: Pattern = Pattern.compile("\\W+")
 
-  fun String.uncapitalize(): String {
-    return if (isNotEmpty() && this[0].isUpperCase()) substring(0, 1).toLowerCase() + substring(1) else this
-  }
   fun toCamelCase(string: CharSequence?, lower: Boolean): String? {
     if (string == null) {
       return null
@@ -40,7 +37,7 @@ object GUtil {
         continue
       }
       if (lower && first) {
-        chunk = chunk.uncapitalize()
+        chunk = chunk.decapitalize()
         first = false
       } else {
         chunk = chunk.capitalize()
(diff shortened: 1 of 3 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/JvmTaskConfigurator.kt
```diff
@@ -2,6 +2,7 @@ package com.apollographql.apollo.gradle.internal
 
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
+import org.gradle.api.internal.jvm.ClassDirectoryBinaryNamingScheme
```
tasomaniac: Hmm, I'm not sure about using internal APIs. This bite me back a lot of times before. Task names are public. Gradle and Kotlin team will never change task names like compileKotlin or compileFooJava. But they may do internal refactoring which even removed/renames this class.

Little duplication of code where you need to remove `Main` is not too bad.

martinbonnin: Shoots, I didn't see this was internal. You're completely right, that'll break if/when gradle decides to move things around. I'll try to duplicate the logic here. It's not only about the "Main" thing but also about `-` and maybe some other characters ?

martinbonnin: It should be better now. The java plugin uses a regex to find word boundaries, which explains why `-` was considered a regex.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/GUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+object GUtil {
+  private val WORD_SEPARATOR: Pattern = Pattern.compile("\\W+")
+
+  fun String.uncapitalize(): String {
+    return if (isNotEmpty() && this[0].isUpperCase()) substring(0, 1).toLowerCase() + substring(1) else this
+  }
```
sav007: nit: 🤔 I think Kotlin 'String.decapitalize()' provides the same function?

martinbonnin: :+1: I was looking for `uncaptalize`, I missed `decapitalize`. Thanks for the heads up!

PR #1983, opened 2020-02-08
Update test fixtures

martinbonnin:
That should make travis happy again. I guess this was a collision between #1958 (yield) and #1971 (name clashes)
@sav007 @tasomaniac can you check ?


martinbonnin:
I think Travis does merge master when building the PR. But these 2 PR where open at the same time so at the time Travis compiled them both were fine. If you were to relaunch the build on the 2nd PR after the 1st were merged, it would have failed. 
We could enforce a rebase on master before merging to prevent that. That will create some friction though so I'm not sure I want it 😅

Issue #1455, opened 2019-07-23
Add KotlinNative support


sav007:
After introducing support for Kotlin model generation there is nothing stopping us from adding Kotlin Native support and make it as multi platform library. This feature can be broken into 2 phases:

Phase 1:
- [x] Introduce pure Kotlin version of `apollo-api` module, avoid any JVM references
- [ ] Introduce light weight implementation of `apollo-runtime` with `ResponseReader / ResponseWriter / InputFieldWriter` support only based on multi platform Kotlinx Serialization JSON library.
- [x] Investigate any changes required by Gradle Plugin
- [x] Investigate multi platform artifacts publishing 
- [x] Provide sample App

After this phase Apollo enables basic support for multi platform projects and provides these functionalities:
- generate Kotlin models
- serialize / deserialize models to / from raw JSON (we can utilize multi platform library such as  Kotlinx Serialization for JSON support)
- support generating cache keys for normalized cache implementation

Phase 2 might include adding support for multi platform network client, http / normalized cache.

designatednerd:
@sav007 let me know if you need any help testing K/N support from iOS, I would be VERY happy to help 😁

tasomaniac:
First step could be to actually migrate to Kotlin from Java source code

Sarthak30:
Is anyone working on it? I would like to start working on it with Phase 0 being moving code from Java to Kotlin and then carrying on to Phase 1 and Phase 2.

sav007:
@Sarthak30 sure you can take a look at:

> Introduce pure Kotlin version of apollo-api module, avoid any JVM references

So basically we should take existing api module create new version `apollo-api-kotlin`.

Sarthak30:
So can you create a repo for this, where I can maybe start committing code and maintain a project there? How do you normally go around this? Or should I create a repo in my profile and maybe you can fork it some time?

martinbonnin:
@Sarthak30 we do [Pull Requests](https://help.github.com/en/articles/about-pull-requests). 

There's one that has just been opened for moving the `apollo-api` module to kotlin => https://github.com/apollographql/apollo-android/pull/1504

Although I realize it converts in-place. That could be risky. @tasomaniac shouldn't we setup some japicmp before the migration to ensure everything behaves the same ? Like in okhttp [did when they switched to kotlin](https://github.com/square/okhttp/issues/4723)

tasomaniac:
[...]

Issue #1697, opened 2019-10-21
Split Gradle Plugin in 2 artifacts, api & implementation


martinbonnin:
* `apollo-gradle-plugin-api` will contain the "public" gradle model part.
* `apollo-gradle-plugin` will contain the implementation. It will depend on `apollo-gradle-plugin-api`

This will ensure that consumers of the plugin will not depend on implementation details that can change.

tasomaniac:
Is this still valid for the incubating plugin?

martinbonnin:
Yes, the incubating plugin has different package names (`com.apollographql.gradle.api` and `com.apollographql.gradle.internal` respectively) but not different maven artifacts. 

The idea would be to ship 2 different maven artifacts so that developers building on-top of the plugin would not pull the internal dependencies by mistake. 

This ticket is very low priority as the impact isn't huge. Also I think java/gradle is moving towards more capable packages and metadata (https://blog.gradle.org/gradle-metadata-1.0) so we maybe that could be used as well.

martinbonnin:
Closing as stale. Will reopen if needed

PR #1985, opened 2020-02-10
Reorganize api-module

sav007:
As a prerequisite for migrating `apollo-api` module to Kotlin, unfortunately these changes are not binary compatible with previous version.

- Extract http cache related API classes to separate module `apollo-http-cache-api`
- Move several classes under `internal` package as they shouldn't have been public at first place (`JsonDataException`, `JsonEncodingException`, `InputFieldMarshaller`, `InputFieldWriter`, `ResponseFieldMapper`, `ResponseFieldMarshaller`, `ResponseReader`, `ResponseWriter`)
- Move couple classes in `apollo-api` module under `api` subpackage (`ApolloLogger`, `QueryDocumentMinifier`, `CustomTypeAdapter`, `CustomTypeValue`, `ScalarTypeAdapters`)

Part of https://github.com/apollographql/apollo-android/issues/1455


martinbonnin:
Nice ! That's looking good :).

One still that still bugs me a little bit is how we use the term `api`. We have 3 usages right now:

1. for `apollo-api`, `internal` is a subpackage of `api`
2. for `apollo-gradle-plugin`, `internal` is a sibling pacakge of `api`
3. for `apollo-http-cache`, we now have two modules: `apollo-http-cache` and `apollo-http-cache-api`

I think 3. is the way to go because it will prevent to leak internal classes to consumers. Actually I wanted to do the same for the gradle plugin (See https://github.com/apollographql/apollo-android/issues/1697).
Maybe we should take this opportunity to split the `apollo-api` module in two new modules ?
* `apollo-core`
* `apollo-core-api`

But it adds a bit of complexity because it basically duplicates all modules. Also there's https://github.com/apollographql/apollo-android/pull/1784 incoming that introduces `apollo-api-core`, which I haven't had the time to dig into a lot yet. 

All in all, I'm not sure what to do with all this so feel free to merge this pull request as it's a nice reorganisation. We can always figure out things later (if things need figuring out...)

tasomaniac:
I never worked with a big enough library to split everything into api vs implementation to completely hide internal implementation details from consumers. In smaller libraries, I tried that and found out that it's an overkill. `internal` naming in the package is a decent solution, imho.

martinbonnin:
@tasomaniac I agree it might be overkill but should wee at least try to stay consistent ? If we're not doing it for `apollo-gradle-plugin`, then we shouldn't do it for `apollo-http-cache` ?

tasomaniac:
Good point! Do we actually need to split the http cache into 2? What was the reason behind that?

sav007:
Hey folks.

@martinbonnin @tasomaniac 
> Do we actually need to split the http cache into 2

We do need 2 modules because `apollo-http-cache-api` required by both `apollo-http-cache` and `apollo-runtime` and it has minimum set of supporting interfaces / classes, without them `apollo-runtime` / `apollo-http-cache` won't even compile.

You can use `apollo-runtime` without `apollo-http-cache` but you can't use it without `apollo-http-cache-api`, as it's just won't compile or crash in runtime with class not found exception. Our runtime heavily integrated with http cache supporting files.

@martinbonnin 
> Maybe we should take this opportunity to split the apollo-api module in two new modules ?
> - apollo-core
> - apollo-core-api

I was thinking of leave `apollo-api` module to be generic for each platform (basically this is what you are  suggesting `apollo-core-api`).  Then for JVM platform we are going to have `apollo-api-jvm` that will depend on `apollo-api`.

Means that we will extract everything JVM specific from our current `apollo-api` module into `apollo-api-jvm`. Convert `apollo-api` into pure Kotlin module to be used in KN.

What do you think, does it make sense?

sav007:
Btw, should we bump version to `1.4.0-SNAPSHOT`?

tasomaniac:
I guess we cannot keep everything under API module either since we don't have cache related stuff there.

martinbonnin:
Hi !

> We do need 2 modules because apollo-http-cache-api required by both apollo-http-cache and apollo-runtime

Got it 👍 

> Means that we will extract everything JVM specific from our current apollo-api module into apollo-api-jvm. Convert apollo-api into pure Kotlin module to be used in KN

Sounds good. 
`apollo-api` will still contain a bunch of `internal` classes (all the ResponseReader/Writer things), which is why I would have maybe gone with the `apollo-core` naming instead. Using `apollo-api`, it's not 100% clear whether it's the API for users or for the rest of apollo-android. But again, I'm not really sure where I'm going with this so we can decide later.

> Btw, should we bump version to 1.4.0-SNAPSHOT?

Yes ! 🚀 We can always branch and cherry pick if we need to make a maintainance release

tasomaniac:
💃

martinbonnin:
🚀

Issue #1993, opened 2020-02-12
java.lang.ArrayIndexOutOfBoundsException on generated code


pcarrier:
**Summary**
Invalid deserialization code is generated for a fragment like this:
```
fragment FriendDetails on Character {
  ...CharacterDetails
  ... on Human {
    homePlanet
  }
  ... on Droid {
    primaryFunction
  }
}
```
leading to `ArrayIndexOutOfBoundsException` during deserialization.

**Description**
The issue can be reproduced against `master` from the `pcarrier/20200212-repro` branch.
Execute `com.example.mixed_fragments_on_fragment.TestExecution`.
(I couldn't find a better way to write a runtime test against the generated code)

`RESPONSE_FIELDS` is of length 3:
```
    private val RESPONSE_FIELDS: Array<ResponseField> = arrayOf(
        ResponseField.forString("__typename", "__typename", null, false, null),
        ResponseField.forFragment("__typename", "__typename", listOf(
          ResponseField.Condition.typeCondition(arrayOf("Human"))
        )),
        ResponseField.forFragment("__typename", "__typename", listOf(
          ResponseField.Condition.typeCondition(arrayOf("Droid"))
        ))
        )
```

`FriendDetails.invoke` goes up to index 3:
```
    operator fun invoke(reader: ResponseReader): FriendDetails = reader.run {
      val __typename = readString(RESPONSE_FIELDS[0])
      val fragments = Fragments(reader)
      val asHuman = readFragment<AsHuman>(RESPONSE_FIELDS[2]) { reader ->
        AsHuman(reader)
      }
      val asDroid = readFragment<AsDroid>(RESPONSE_FIELDS[3]) { reader ->
        AsDroid(reader)
      }
      FriendDetails(
        __typename = __typename,
        fragments = fragments,
        asHuman = asHuman,
        asDroid = asDroid
      )
    }
  }
```

**Version**
Reproduced by our customer on `v1.3.0`, and through this testbed on `master`.

pcarrier:
`com.apollographql.apollo.compiler.codegen.kotlin.KotlinCodeGen.toMapperFun` calls `mapIndexed`, but `FieldType.Fragments` does not actually consume an index.

pcarrier:
Pushed a fix.

PR #1994, opened 2020-02-13
Fix regression issue field indexes messed up for queries with fragments

sav007:
Fix regression issue introduced after adding support fragments with directives. When query contains fragments and inline fragments at the same level, indexes used to reference GraphQL response fields (in `ResponseField[]` array) get out of order. This happens because we removed `ResponseField.forString` code generation for surrogate field `Fragments`. This fix reverts back changes that caused index to be out of order.

For info see original issue, but this is where it gets messed up:
https://github.com/apollographql/apollo-android/pull/1994/files#diff-4784749789d3edec94f885e8663e6246R81

Because this was missing:
https://github.com/apollographql/apollo-android/pull/1994/files#diff-4784749789d3edec94f885e8663e6246R46

Closes https://github.com/apollographql/apollo-android/issues/1993


sav007:
cc @designatednerd

designatednerd:
Thanks! @pcarrier How's this lookin' to you?

pcarrier:
@designatednerd I have to accept my limitations and admit that I have yet to grasp the design of this compiler and associated runtime decisions, so I'll happily defer to @sav007. Thanks so much for the quick turnaround; how quickly can we hope to land into published artifacts?

pcarrier:
@designatednerd I have to accept my limitations and admit that I have yet to grasp the design of this compiler and its generated Kotlin code to reason effectively about its correctness, but this absolutely prevents the bug as we've observed it. Thanks so much for the quick turnaround; how quickly do we expect to land such bugfixes into published artifacts?

designatednerd:
I'll leave code review to @martinbonnin and @tasomaniac, but it would be helpful to ship this ASAP since it's a regression.

apollo-compiler/src/test/graphql/com/example/unique_type_name/HeroDetailQuery.kt
```diff
@@ -62,6 +62,7 @@ class HeroDetailQuery : Query<HeroDetailQuery.Data, HeroDetailQuery.Data, Operat
 
     companion object {
       private val RESPONSE_FIELDS: Array<ResponseField> = arrayOf(
+          ResponseField.forString("__typename", "__typename", null, false, null),
```
tasomaniac: Does it make sense that the same line is next to each other here?

sav007: Do you mean the same line of  `ResponseField.forString("__typename", "__typename", null, false, null)`?  If yes that's correct, first line used by actual field `__typename` and the second one is not involved in parsing response, but it needs to be there to have right field index.

I'm thinking of making changes to remove `RESPONSE_FIELDS ` all together, as index based access to the response field meta looks fragile.

PR #2003, opened 2020-02-19
Rafa/documentation mirroring

ralphie9224:
I think this looks alright based on @martinbonnin original requirements. I didn't see any relative links, but I might have missed it as I was skimming through it, feel free to point them out.


scripts/deploy.sh
```diff
@@ -25,6 +25,9 @@ else
   echo "Snapshot deployed!"
 fi
 
+
+cat README.md | grep -v 'the official docs' > docs/source/essentials/get-started.md
```
ralphie9224: Wondering if we could just not have a `get-started.md` file and just have this script here. It would essentially create the `get-started.md` file when it gets clobbered with the `cat README.md`

martinbonnin: Not sure I follow. The original intent was to remove `get-started.md` from source control since it duplicates a lot of information with `README.md`

martinbonnin: > if we delete the get-started.md from source control, all we'd have to do is run this cat README.md | grep -v 'the official docs' > docs/source/essentials/get-started.md in order for it to actually show up when it's deployed on gatsby.

Exactly. This avoids duplicating the file in source control. You will need to do that from netlify.yml though as `deploy.sh` might not be recognized by neetlify/gatsby

netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
   base    = "docs/"
   publish = "docs/public/"
-  command = "gatsby build --prefix-paths && mkdir -p docs/android && mv public/* docs/android && mv docs public/ && mv public/docs/android/_redirects public"
+  command = "gatsby build --prefix-paths && mkdir -p docs/android && mv public/* docs/android && mv docs public/ && mv public/docs/android/_redirects public && cat README.md | grep -v 'the official docs' > docs/source/essentials/get-started.md'"
```
martinbonnin: You should invert the commands, `getting-started.md` is an input to `gatsby build` and should be present before the command is executed.

netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
   base    = "docs/"
   publish = "docs/public/"
-  command = "./generate-docs.sh"
+  command = "chmod +x ./generate-docs.sh"
```
martinbonnin: This won't work as it just changes the mode, it doesn't execute the script.
I believe you can change the mode in git itself. Change the permission on your machine and commit the file.

docs/generate-docs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+## Set up documentation so README.md is reflected onto apollographql.com/docs/android
```
martinbonnin: Can you add `set -e` at the beginning of this script to exit with an error if any of the commands failed ? (see https://intoli.com/blog/exit-on-errors-in-bash-scripts/).

This was the case before due to the `&&`. We should keep this behaviour.

martinbonnin:
@ralphie9224 it turns out there's a dead link in the main `README.md`: 

change 
```
[plugin-configuration](#configuration-reference) 
```

with an absolute link like it is done in the end of the readme:

```
[plugin-configuration.md](https://www.apollographql.com/docs/android/gradle/plugin-configuration/)
```

docs/generate-docs.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+
+echo -e "---\ntitle: Get started\ndescription: Learn how to quickly set up Apollo Client in your Android app\n---\n\n$(cat ../README.md | grep -v 'the official docs')" > source/essentials/get-started.md
```
tasomaniac: One question: are the URL links ok? I'm not 100 sure but there is a slight difference between how links to other docs work between README.md and others. Either we would need to convert those links or keep the compatible.

martinbonnin: > One question: are the URL links ok?

I think they are. Gatsby actually has a check that fails if they're not ok.

Most of the url in the README.md are absolute, which explains why they work. There's only one that's not absolute and it's an anchor: `[downloading a schema file](#downloading-a-schemajson-file).`

tasomaniac: Anchors work fine. And you're right, gatsby verify links. I think we should be ok, then. We cannot put relative links to code in the repo, we will need to put full links.

tasomaniac: Should we merge this then?

martinbonnin: LGTM! I made a rebase there: https://github.com/apollographql/apollo-android/pull/2053

Issue #2005, opened 2020-02-19
Job Cancellation Exception


msya:
**Summary**

I am getting a `kotlinx.coroutines.JobCancellationException: Job was cancelled; job=StandaloneCoroutine{Cancelled}` exception when I use the `toFlow` method on the Apollo client.

**Description**

This repo https://github.com/msya/ApolloAndroidPlayground has an example app that shows the problem. I have a repo that creates an Apollo client and uses the `toFlow` extension to get the Flow. 

```
class Repo {

    fun getData(): Flow<Response<TestQuery.Data>> {
        val client = ApolloClient.builder().serverUrl("https://api.kivaws.org/graphql").build()
        return client.query(TestQuery.builder().build()).toFlow()
    }
}
```

I also have a ViewModel that launch a coroutines in the view model scope. In this coroutine, I am collecting from the Flow. 

```
class MyViewModel: ViewModel() {

    private val repo = Repo()

    val scope = CoroutineScope(Dispatchers.Main)

    fun getData() {
        scope.launch {
            repo.getData()
                .collect {
                    println(it)
                }
        }
    }

    fun cancel() {
        scope.cancel()
    }

}
```

However, if I call cancel on the Job returned by the coroutine. I am getting the cancellation exception - `kotlinx.coroutines.JobCancellationException: Job was cancelled; job=StandaloneCoroutine{Cancelled}`. In the activity below, I am canceling the call explicitly on a button click.

```
class MainActivity : AppCompatActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)

        val model: MyViewModel by viewModels()

        make_request?.setOnClickListener {
            model.getData()
        }

        cancel?.setOnClickListener {
            model.cancel()
        }

    }
}
```

I suspect this problem is due to the clone method in the `callbackFlow` in the `toFlow` extension. If you remove the clone method, the exception does not occur. 

**Version**
1.3.2

martinbonnin:
For some context, the `clone` was added to address this bug: https://github.com/apollographql/apollo-android/issues/1986

diegocunhawarren:
Hey @martinbonnin, looking the changes regarding this issue, the class `ApolloSubscriptionCall` has the same issue about emit when the channel is closed, I created a custom extension like yours and just fixed to me!

CureleaAndrei:
[...]

PR #2007, opened 2020-02-20
Coroutines: make sure we do not crash when calling `offer`

martinbonnin:
Closes https://github.com/apollographql/apollo-android/issues/2005

See https://github.com/Kotlin/kotlinx.coroutines/issues/974
See https://github.com/apollographql/apollo-android/issues/1585

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -13,9 +13,8 @@ import kotlinx.coroutines.flow.*
 
 private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall.Callback<T>() {
 
-  @ExperimentalCoroutinesApi
   override fun onResponse(response: Response<T>) {
-    if (!channel.isClosedForSend) {
+    runCatching {
       channel.offer(response)
     }
   }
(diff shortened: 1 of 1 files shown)


apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -13,9 +13,8 @@ import kotlinx.coroutines.flow.*
 
 private class ChannelCallback<T>(val channel: Channel<Response<T>>) : ApolloCall.Callback<T>() {
 
-  @ExperimentalCoroutinesApi
   override fun onResponse(response: Response<T>) {
-    if (!channel.isClosedForSend) {
+    runCatching {
```
martinbonnin: @sav007 I feel the `runCatching` approach is safer since there could potentially be a race condition where the channel is closed just after the `if` before. Let me know if `isClosedForSend` was used to catch other cases.

msya: To avoid having to wrap the offer in a `runCatching`, you could create an extension on a Channel.

```
fun <E> SendChannel<E>.safeOffer(value: E) = runCatching {
        offer(value)
}

martinbonnin: @msya, for 2 instances only, I'd rather not pollute the `Channel` namespace with an extension function.

PR #2014, opened 2020-02-24
Introduce new apollo-api-jvm module

sav007:
Move all JVM platform related artifacts to the new created module `apollo-api-jvm`.
This module now is required by any code targeting JVM platform.

Remaining classes in `apollo-api` will be converted to Kotlin in next PR (some utils functions will be moved to `apollo-api-jvm` after this as well)

`apollo-api` is going to be a core api module in pure Kotlin.

Part of https://github.com/apollographql/apollo-android/issues/1455


martinbonnin:
Looking at the `import java.xyz` in the `apollo-api-jvm`, almost everything seems to be about collections:

*  Collections
* HashMap
* Set
* ...

Can't this easily be replaced by Kotlin collections ? 
Then there would only be `Optional` left because it implements `Serializable` and this is a JVM-only thing. But for the rest, couldn't everything could be in `apollo-api` in pure Kotlin ?

tasomaniac:
In a Kotlin Multiplatform module, these can co-exist. Especially assuming that we will need the same functionality on Kotlin native target, we could just keep them in a single module.

sav007:
@tasomaniac Agh, sh.... I totally forgot that it will be different targets and JVM will have all jvm related stuff. 

Silly me, time to retire.

tasomaniac:
That's my experience so far. I never did that on a library context tho.

PR #2021, opened 2020-02-27
Convert top level api classes to Kotlin

sav007:


apollo-api/src/main/java/com/apollographql/apollo/api/Error.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Represents an error response returned from the GraphQL server
+ */
+data class Error(
+    /**
+     * Server error message
+     */
+    @JvmField
+    val message: String,
```
martinbonnin: Is the `@JvmField` required here ? It exposes a field publicly for no apparent reason.

apollo-api/src/main/java/com/apollographql/apollo/api/Error.kt
```diff
@@ -0,0 +1,74 @@
...
+     */
+    @JvmField
+    val customAttributes: Map<String, Any?> = emptyMap()
+) {
+
+  /**
+   * Returns server error message.
+   */
+  @Deprecated(message = "Use property instead", replaceWith = ReplaceWith(expression = "message"))
+  fun message(): String? = message
```
martinbonnin: +1 for the notice 👍 
I wonder how that's going to play with Java if we decide to remove `@JvmField` as mentioned above.

apollo-api/src/main/java/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo.api
+
+import java.io.File
+
+data class FileUpload(@JvmField val mimetype: String, @JvmField val file: File)
```
martinbonnin: Why do we need FileUpload to be a data class ?

apollo-api/src/main/java/com/apollographql/apollo/api/Error.kt
```diff
@@ -0,0 +1,74 @@
+package com.apollographql.apollo.api
+
+/**
+ * Represents an error response returned from the GraphQL server
+ */
+data class Error(
```
martinbonnin: Why do we need `Error` to be a data class ?

apollo-api/src/main/java/com/apollographql/apollo/api/Operation.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    open fun marshaller(): InputFieldMarshaller {
+      return InputFieldMarshaller { }
+    }
+
+    /**
+     * Serializes variables as JSON string to be sent to the GraphQL server.
+     */
+    fun marshal(): String {
```
martinbonnin: This used to have `throw IOException` in the signature. I don't think it's a huge problem since I think `marshal` is only used internally ? If that's the case, it would be worth hiding that in the `.internal` package

martinbonnin:
japicmp report there: https://storage.googleapis.com/martin-playground.appspot.com/japicmp.html. I put a few comments inline in the code as the discussion will be easier with some context.

Some more global remarks:

1. Class file format changed from 50 (java 6) to 51 (java 7). I think this is fine ? 
2. A lot of classes/methods became final, which is a good thing in my opinion.
3. java.lang.Boolean is now the primitive non-boxed `boolean`, which I think is also a good thing.
4. data classes: I'm afraid using data classes broadly will not only add unecessary bytecode but also might convey the wrong intent. When I'm seeing a data class, I'm looking for a usage of `.copy()` or `.equals()` and it doesn't seem we need it at the moment ?
5. `@JvmField`: I would drop the `@JvmField` annotation right now. Java can use `.getField()` instead of `.field()` and we won't have to maintain `@JvmField` in the long run.

Overall looks great :). One small bonus I'd like to have is the ability to keep the git history accross file rename. I'm not sure how doable that is but I'll look into it.

martinbonnin:
I tried to do the "2 steps" commit thing (move first, modify then) to try and keep file history but it's not working great either. See https://github.com/apollographql/apollo-android/compare/master...martinbonnin:convert-to-kt-keep-history. 
Both intelliJ and git do not follow accross file renames by default. It works on the commandline if one adds the `--follow` option but if manual input is needed, we might as well lookup the java file directly.

apollo-api/src/main/java/com/apollographql/apollo/api/Error.kt
sav007: On Java as far as I can see it just marked as deprecated without replace action, just a message. I think we are good here.

sav007:
>Class file format changed from 50 (java 6) to 51 (java 7). I think this is fine ?

Yeah, we should be fine

> data classes: I'm afraid using data classes broadly will not only add unecessary bytecode but also might convey the wrong intent. When I'm seeing a data class, I'm looking for a usage of .copy() or .equals() and it doesn't seem we need it at the moment ?

Agree, will clean up

> @JvmField: I would drop the @JvmField annotation right now. Java can use .getField() instead of .field() and we won't have to maintain @JvmField in the long run.

Was trying to make it backward compatible with Java but I guess went too far, will remove those that not required.

martinbonnin:
@sav007 Looking into this again, I realized I read the japicmp report wrong. We now generate Java 6 while it was 7 before so Kotlin is actually more compatible that java 🙃

PR #2026, opened 2020-03-01
config okHttp timeout

silencecorner:
when schema.json too large or network slow，may generate socket exception！set properties in jvmargs:
```
org.gradle.jvmargs=-DokHttp.connectTimeout=60 -DokHttp.readTimeout=60
```

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
@@ -10,6 +10,7 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.TaskAction
 import java.io.ByteArrayOutputStream
+import java.util.concurrent.TimeUnit
 
 abstract class ApolloDownloadSchemaTask : DefaultTask() {
   @get:Input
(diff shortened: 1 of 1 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
```diff
@@ -66,7 +67,11 @@ abstract class ApolloDownloadSchemaTask : DefaultTask() {
         .url(urlBuilder.build())
         .build()
 
-    val response = OkHttpClient().newCall(requestBuilder.build()).execute()
+    val response = OkHttpClient.Builder()
+        .connectTimeout(System.getProperty("okHttp.connectTimeout","10").toLong(), TimeUnit.SECONDS)
+        .readTimeout(System.getProperty("okHttp.readTimeout","10").toLong(), TimeUnit.SECONDS)
```
martinbonnin: Thanks for the contribution 👍 . What about making the default a bit bigger like "60" ? Most of the times, this will be called from the dev machine or from CI so I think it's more important that the task actually downloads something compared to speed.

tasomaniac:
```suggestion
        .connectTimeout(System.getProperty("okHttp.connectTimeout", "60").toLong(), TimeUnit.SECONDS)
        .readTimeout(System.getProperty("okHttp.readTimeout", "60").toLong(), TimeUnit.SECONDS)
```

tasomaniac: I made the change through Github UI 😄

tasomaniac:
Thanks for contribution @silencecorner I will update the advanced documentation after merging this.

tasomaniac:
Added here: https://www.apollographql.com/docs/android/essentials/get-started/#downloading-a-schema

PR #2028, opened 2020-03-02
Convert bunch of internal classes to Kotlin

sav007:
Not all, but those shared between Java and Kotlin models generation.


apollo-api/src/main/java/com/apollographql/apollo/api/internal/InputFieldMarshaller.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import java.io.IOException
+
+interface InputFieldMarshaller {
+  @Throws(IOException::class)
+  fun marshal(writer: InputFieldWriter)
+
+  companion object {
+    @JvmSynthetic
+    inline operator fun invoke(crossinline block: (InputFieldWriter) -> Unit) = object : InputFieldMarshaller {
```
martinbonnin: Is there a place where `invoke` is used as a function call ? If it doesn't need to be an operator, maybe call it `from` or `wrap` ?

```kotlin
    override fun marshaller(): InputFieldMarshaller = InputFieldMarshaller.from { writer ->
      writer.writeBoolean("withDetails", this@TestQuery.withDetails)
      writer.writeBoolean("skipHumanDetails", this@TestQuery.skipHumanDetails)
    }
```

Or with a top-level function, like coroutines use a lot ? 

```kotlin
    override fun marshaller(): InputFieldMarshaller = inputFieldMarshaller { writer ->
      writer.writeBoolean("withDetails", this@TestQuery.withDetails)
      writer.writeBoolean("skipHumanDetails", this@TestQuery.skipHumanDetails)
    }
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/response/RealResponseReader.java
```diff
@@ -286,22 +286,22 @@ private void checkValue(ResponseField field, Object value) {
       return (String) value;
     }
 
-    @Override public Integer readInt() {
+    @Override public int readInt() {
```
martinbonnin: Using primitive types looks way better 👍

apollo-api/src/main/java/com/apollographql/apollo/api/internal/InputFieldMarshaller.kt
sav007: It's used in input types for example https://github.com/apollographql/apollo-android/pull/2028/files#diff-431dfc1d4ab56f3b8039eb0ddb8547fdR107

Let's keep it for now as SAM conversion for interfaces is coming in  Kotlin 1.4 https://blog.jetbrains.com/kotlin/2019/12/what-to-expect-in-kotlin-1-4-and-beyond/?_ga=2.136948336.1505843568.1583040891-1122893.1461098344

After that we will remove these functions entirely.

Issue #1970, opened 2020-02-05
Apollo files are not generated after upgrading to 1.3.0.


sabrinaj:
**Summary**
Apollo files are not generated with Apollo 1.3.0.

**Description**
Hello,

I'm struggling to upgrade to v1.3.0, apollo files are not generated with it. I have the following tree:
```
- app
- moduleA => module where I want to have apollo generated classes with "my.package.common" as package
- someGitModule
	schema.json
	- FeatureA
		- Fragment
			fragmentA.qraphql
		- Query
			queryA.graphql
	- FeatureB
		- Fragment
			fragmentB.qraphql
		- Query
			queryB.graphql
```
			
			
**Apollo v1.2.x:** Works fine, I get all files from all features generated in moduleA with "my.package.common" as package

```
android {
	sourceSets {
		main.graphql.srcDirs += "../../someGitModule"
	}
}

apollo {
    generateKotlinModels.set(true)

    sourceSet {
        schemaFile = "../../someGitModule/schema.json"
    }
    outputPackageName = "my.package.common"
}
```

**Apollo v1.3:** files are not generated

```
apollo {
    onCompilationUnits { // and not "onCompilationUnit " like in ReadMe, how come?
        schemaFile = file("../../someGitModule/schema.json")
        rootPackageName = "my.package.common"
        graphqlSourceDirectorySet.srcDirs += "../../someGitModule/.."
    }

    generateKotlinModels.set(true)
}
```

- Should the "apollo {}" be inside "android {}"? 
- What would be the proper syntax with `graphqlSourceDirectorySet.srcDir` to have same result as in v1.2.x? With my current syntax, it seems to generate apollo files but only for test (folders `debugUnitTest/debugAndroidTest/releaseUnitTest` and `debug` is missing), did I miss something?

**Version**
1.3.0

martinbonnin:
Assuming you have your apollo configuration in moduleA/build.gradle, you only need to move up once ("../someGitModule" and not "../../someGitModule") to point to your git module. I'm not sure why you had to move up twice before. 

Can you try the following:

```groovy
apollo {
    onCompilationUnits { // this will become `onCompilationUnit` in 1.3.1
        schemaFile = file("../someGitModule/schema.json")
        rootPackageName = "my.package.common"
        graphqlSourceDirectorySet.srcDirs += file("../someGitModule/")
        graphqlSourceDirectorySet.include("**/*.graphql")
    }

    generateKotlinModels.set(true)
}
```

[...]

PR #2032, opened 2020-03-04
Expose Compilation Units earlier

martinbonnin:
Fixes https://github.com/apollographql/apollo-android/issues/1970

Also fix doc

Sorry for the big diff, most of it is a schema file for tests. I'm hoping git does the right thing and just stores this once.


README.md
```diff
@@ -304,9 +304,9 @@ apollo {
 // With:
 apollo {
   onCompilationUnit {
-     schemaFile = "/path/to/your/schema.json"
+     schemaFile.set("/path/to/your/schema.json")
```
japplin: Shouldn't this be
 ```
schemaFile = file("/path/to/your/schema.json")
```

Comment (martinbonnin) [reply README.md]: Yup, my bad, I wanted to make something that worked in kotlin and groovy but forgot the `file` part...

Comment (martinbonnin) [reply README.md]: I think `schemaFile.set(file("/path/to/your/schema.json"))` would work equally well in Groovy and Kotlin.

**Comment by japplin** [comment]
I tried running the tests locally but it seems I'm missing some steps for setting up the environment, anyway I think this project matches 
```
withProject(
        false,
        listOf(androidLibraryPlugin, kotlinAndroidPlugin, apolloPlugin),
        """
      apollo {
        onCompilationUnit {
          schemaFile = file("../schema.json")
          rootPackageName = "com.company.module"
          generateKotlinModels = true
        }
      }
    """.trimIndent()) { dir ->
```

File: apollo-gradle-plugin/testProjects/README.md [hunk 2]
```diff
@@ -0,0 +1,13 @@
+# testProjects
+
+This directory contains projects used by the plugin tests.
+
+You can find other files in [../src/test/files](../src/test/files). These projects are programmatically constructed but it becomes harder and harder to maintain so new test projects should be added here. It will duplicate some files and schemas but git should do the correct thing and only store objects once.
```
martinbonnin: @tasomaniac if ok with you, I'll switch to using plain test directories for plugin tests instead of programmatically constructing them like it was the case before. It's just easier to maintain and debug.

tasomaniac: Let's try and see. In other projects, I've also made them programmatically since they are dynamic and some pieces needs to change.

For the first time here, I've seen the approach is slightly different. I think it makes sense.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt
```diff
@@ -99,16 +99,22 @@ open class ApolloPlugin : Plugin<Project> {
...
+          compilationUnit.outputDir.set(codegenProvider.flatMap { it.outputDir })
+          compilationUnit.operationOutputFile.set(codegenProvider.flatMap { it.operationOutputFile })
+
+          /**
+           * Order matters here. See https://github.com/apollographql/apollo-android/issues/1970
+           * We want to expose the `CompilationUnit` to users before the task is configured by
+           * AndroidTaskConfigurator.registerGeneratedDirectory so that schemaFile and sourceDirectorySet are
+           * correctly set
+           */
+          apolloExtension.compilationUnits.add(compilationUnit)
```
tasomaniac: :100:

Issue #2017, opened 2020-02-25
compiler cannot cope with schema that contains byte order mark


simonmarshall:
**Summary**
a schema.json file extracted from the aws appsync console contains a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8) which confuses the apollo-compiler: 
```
Failed to parse GraphQL schema introspection query from `***/schema.json`: Use JsonReader.setLenient(true) to accept malformed JSON at path $
```
see the schema.json in [schema.zip](https://github.com/apollographql/apollo-android/files/4250490/schema.zip).

**Description**
i realise that (1) the compiler largely hands off parsing to moshi and (2) moshi has a `lenient` flag i could possibly use to suppress it, but (a) i cannot use (2) since i am using the [apollo-client-maven-plugin](https://github.com/aoudiamoncef/apollo-client-maven-plugin) which does not expose this flag and (b) it's _not_ malformed json it's a standard BOM that prefixes the json file contents (and which i want to be parsed _without_ leniency).

i think the correct fix is for `com.apollographql.apollo.compiler.parser.Schema#JsonReader.locateSchemaRootNode()` to skip any [byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark) before finding the beginning of the json object.

**Version**
apollo-compiler-1.2.3.

sav007:
There was a debate in Moshi regarding this issue as well, and it looks like issue should be fixed on the client side:

>Implementations MUST NOT add a byte order mark to the beginning of a JSON text. In the interests of interoperability, implementations that parse JSON texts MAY ignore the presence of a byte order mark rather than treating it as an error.

sav007:
Merged.

PR #2039, opened 2020-03-06
Ignore BOM symbol when parsing GraphQL schema

sav007:
Closes https://github.com/apollographql/apollo-android/issues/2017


simonmarshall:
thanks @sav007 
there was a bommed schema attached, if u need unit test data

PR #2044, opened 2020-03-06
Convert next portion of api module files to Kotlin

sav007:


apollo-api/src/main/java/com/apollographql/apollo/api/CustomTypeValue.kt
```diff
@@ -4,9 +4,7 @@ package com.apollographql.apollo.api
  * A wrapper class for representation of custom GraphQL type value, used in user provided [CustomTypeAdapter]
  * encoding / decoding functions.
  **/
-abstract class CustomTypeValue<T> private constructor(
-    @JvmField val value: T
-) {
+sealed class CustomTypeValue<T> (@JvmField val value: T) {
```
tasomaniac:
```suggestion
sealed class CustomTypeValue<T>(@JvmField val value: T) {
```

apollo-api/src/main/java/com/apollographql/apollo/api/internal/json/BufferedSourceJsonReader.kt
```diff
@@ -55,7 +55,7 @@ class BufferedSourceJsonReader(private val source: BufferedSource) : JsonReader
   override var failOnUnknown = false
 
   @Throws(IOException::class)
-  override fun beginArray() {
+  override fun beginArray(): JsonReader {
```
tasomaniac: One thing that I do for self returning functions is to use apply

```
override fun beginArray(): JsonReader = apply {
  // do stuff
}
```

Then you don't have to `return this`

apollo-api/src/main/java/com/apollographql/apollo/api/internal/json/InputFieldJsonWriter.kt
```diff
@@ -0,0 +1,222 @@
...
+import java.io.IOException
+
+class InputFieldJsonWriter(
+    private val jsonWriter: JsonWriter,
+    private val scalarTypeAdapters: ScalarTypeAdapters
+) : InputFieldWriter {
+
+  @Throws(IOException::class)
+  override fun writeString(fieldName: String, value: String?) {
+    if (value != null) {
```
tasomaniac: For consistency, maybe we can invert this conditional and have `value == null` first.

apollo-api/src/main/java/com/apollographql/apollo/api/internal/json/JsonEncodingException.kt
```diff
@@ -13,13 +13,9 @@
...
-/** Thrown when the data being parsed is not encoded as valid JSON. */
-public final class JsonEncodingException extends IOException {
-  public JsonEncodingException(String message) {
-    super(message);
-  }
-}
+/**
+ * Thrown when the data being parsed is not encoded as valid JSON.
+ */
+class JsonEncodingException(message: String) : RuntimeException(message)
```
martinbonnin: What is the reason to make it a `RuntimeException` instead of an `IOException` like it used to be ?

tasomaniac: Are you suggesting to keep it as IOException but skip having @Throws annotation? I assume this turns checked exception to unchecked. In Kotlin the difference is thinner.

sav007: Agh, it was my mistake. Will revert it back

PR #2049, opened 2020-03-07
Merge Hotfix 1.3.3 into master

tasomaniac:
Rewrote the fix in `ScalarTypeAdapters` in kotlin


apollo-api/src/main/java/com/apollographql/apollo/api/ScalarTypeAdapters.kt
```diff
@@ -31,7 +35,19 @@ class ScalarTypeAdapters(customAdapters: Map<ScalarType, CustomTypeAdapter<*>>)
     private val DEFAULT_ADAPTERS = mapOf(
         String::class.java to object : DefaultCustomTypeAdapter<String>() {
           override fun decode(value: CustomTypeValue<*>): String {
-            return value.value.toString()
+            return if (value is GraphQLJsonList || value is GraphQLJsonObject) {
+              val buffer = Buffer()
+              JsonWriter.of(buffer).use { writer ->
```
tasomaniac: Rewrote this bit in Kotlin. Using `use` is quite nice. I checked, it also ignores any issues if `writer.close()` fails. Just like in the original fix: https://github.com/apollographql/apollo-android/pull/2037

PR #2053, opened 2020-03-07
Avoid doc duplication by using README.md on gatsby as well

martinbonnin:
This is the same as https://github.com/apollographql/apollo-android/pull/2003 rebased and with the timeout doc from @tasomaniac added to `README.md`


README.md
```diff
@@ -200,17 +200,22 @@ apollo {
...
 ```
-./gradlew :shared:downloadApolloSchema -Pcom.apollographql.apollo.endpoint=https://your.graphql.endpoint -Pcom.apollographql.apollo.schema=src/main/graphql/com/example/schema.json  "-Pcom.apollographql.apollo.headers=Authorization=Bearer YOUR_TOKEN" "-Pcom.apollographql.apollo.query_params=key1=value1&key2=value2"
+./gradlew :module:downloadApolloSchema -Pcom.apollographql.apollo.endpoint=https://your.graphql.endpoint -Pcom.apollographql.apollo.schema=src/main/graphql/com/example/schema.json  "-Pcom.apollographql.apollo.headers=Authorization=Bearer YOUR_TOKEN" "-Pcom.apollographql.apollo.query_params=key1=value1&key2=value2"
 ```
 
 The `com.apollographql.apollo.headers` and `com.apollographql.apollo.query_params` properties both take a query string where key and values should be URL encoded.
 
+The default timeout for download operation is 1 minute. If you have a large `schema.json`, you may want to increase the timeout. Do that by adding the following into `gradle.properties`:
+```
+org.gradle.jvmargs=-DokHttp.connectTimeout=60 -DokHttp.readTimeout=60
```
martinbonnin: @silencecorner @tasomaniac Do you know if there's a specific reason to use system properties (-D) over project properties (-P) ? There's a bunch of both in the project and I feel it would be more clean to use the same everywhere but maybe not ?

tasomaniac: As far as I know, it is harder to put project properties as persistent. Like putting them in `gradle.properties`

I had this in mind while reviewing the prev PR. But I've seen other `-D` in the project. I didn't know we also have `-P`. I would also go for consistency.

martinbonnin: If wish there were a single way to do it :). 
I'll move the downloadSchema task config to `-D` so we have `-D` everywhere

tasomaniac: I saw you were using task inputs as parameters which works even without -P or -D.

martinbonnin: Ah yes, that's even better 👍

tasomaniac: I think we can have 2. Either that which looks much better. Or -D for the things that are more persistent. Like timeout setting.

martinbonnin:
Thanks @ralphie9224 !

PR #2054, opened 2020-03-08
Update Okio and Okhttp

tasomaniac:
Okio and Okhttp also brings kotlin stdlib dependencies. So I thought it would be a good step towards Kotlin MPP

- This also forces us to use Gradle metadata (must have for Kotlin MPP).
- And also for Java 8
- Also upgraded Gradle and Kotlin
  - Gradle for better Metadata support
  - Kotlin because 1.3.50 has a bug about metadata


gradle/dependencies.gradle
```diff
@@ -39,11 +38,9 @@ ext.dep = [
         runtime: "org.antlr:antlr4-runtime:$versions.antlr4",
     ],
     apollo                : [
-        plugin          : "com.apollographql.apollo:apollo-gradle-plugin:$versions.apollo",
-        runtime         : "com.apollographql.apollo:apollo-runtime:$versions.apollo",
-        rx2             : "com.apollographql.apollo:apollo-rx2-support:$versions.apollo",
```
tasomaniac: these were never used since integration tests use the real module instead of latest maven dependency.

apollo-runtime/build.gradle.kts
```diff
@@ -19,7 +19,6 @@ dependencies {
   add("testImplementation", groovy.util.Eval.x(project, "x.dep.junit"))
   add("testImplementation", groovy.util.Eval.x(project, "x.dep.truth"))
   add("testImplementation", groovy.util.Eval.x(project, "x.dep.okHttp.mockWebServer"))
-  add("testImplementation", groovy.util.Eval.x(project, "x.dep.okHttp.testSupport"))
```
tasomaniac: Looks like this was never used. And also not published anymore as of OkHttp 4.x

tasomaniac:
Build seems to be failing because of checkstyle setup.

gradle/dependencies.gradle
```diff
@@ -9,13 +9,12 @@ def versions = [
     javaPoet              : '1.9.0',
     jetbrainsAnnotations  : '13.0',
     junit                 : '4.12',
-    kotlin                : '1.3.50',
+    kotlin                : '1.3.70',
```
martinbonnin: 🚀

apollo-android-support/build.gradle.kts
```diff
@@ -11,8 +11,8 @@ extensions.findByType(BaseExtension::class.java)!!.apply {
   compileSdkVersion(groovy.util.Eval.x(project, "x.androidConfig.compileSdkVersion").toString().toInt())
 
   compileOptions {
-    sourceCompatibility = JavaVersion.VERSION_1_7
-    targetCompatibility = JavaVersion.VERSION_1_7
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
```
martinbonnin: I think this is ok but I guess it's worth double checking and give a warning in release notes if needed. I started a thread over there: https://github.com/apollographql/apollo-android/issues/2057

apollo-rx2-support/build.gradle.kts
```diff
@@ -4,8 +4,8 @@ plugins {
 }
 
 java {
-  targetCompatibility = JavaVersion.VERSION_1_7
-  sourceCompatibility = JavaVersion.VERSION_1_7
+  targetCompatibility = JavaVersion.VERSION_1_8
+  sourceCompatibility = JavaVersion.VERSION_1_8
```
martinbonnin: Should this be factored-in the root `build.gradle.kts` file ? Make sure all sub-projects target JAVA 8 ? Or are Android projects different ?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```
martinbonnin: I was thinking of bumping Gradle to latest version (6.2.2 IIRC). It should be safe to use the latest Gradle version since we have test that check against older versions.

apollo-android-support/build.gradle.kts
tasomaniac: Google nowadays recommends libraries to target Java 8 to move the platform forward. Very few of the AndroidX libs already do that. OkHttp too. 

Technically we don't have to do it for all modules. But for consistency that would be good. I will check if this can be done in root.

gradle/wrapper/gradle-wrapper.properties
tasomaniac: I actually did in previous commits. My reasoning was that Gradle metadata support officially become stable in 6.0. But I had an issue with Checkstyle task and reverted.

build.gradle.kts
```diff
@@ -56,6 +53,21 @@ subprojects {
...
+      sourceCompatibility = JavaVersion.VERSION_1_8
+      targetCompatibility = JavaVersion.VERSION_1_8
+    }
+  }
+
+  plugins.withType(org.gradle.api.plugins.JavaPlugin::class.java) {
+    extensions.configure(JavaPluginExtension::class.java) {
+      sourceCompatibility = JavaVersion.VERSION_1_8
+      targetCompatibility = JavaVersion.VERSION_1_8
+    }
```
martinbonnin: 👍

PR #2062, opened 2020-03-10
Introduce Kotlin Multiplatform into apollo-api module

tasomaniac:
For now, we only have `jvm` as a target. That means that all changes in JVM should be backward compatible.

## What is next?

- Maven publish should be tested (probably needs to be adjusted) More info: https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#publishing-a-multiplatform-library
- FileUpload, `ScalarTypeAdapters` implementations only exist in jvm. So in a potential native implementation, these need to be reimplemented. 
- `BigDecimal` is typealiased to `java.math.BigDecimal`. Investigation showed that using `ULong` in possible native impl should be enough.
- `Optional`, its siblings and `UnmodifiableMapBuilder` classes are kept in Java and only ever going to be available for Java code generation. This should just work for now.

## Unit tests

Kept all tests in Java ~~(Checkstyle suddenly started to yell and had to little style fixes)~~

In `apolli-api` module, we only have `jvmTest` target. Once we have actual multi-platform code, this needs to be moved into `commonTest` as much as possible.


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CustomTypeValue.kt
```diff
@@ -13,11 +16,12 @@ sealed class CustomTypeValue<T>(@JvmField val value: T) {
     override fun equals(other: Any?): Boolean {
       if (this === other) return true
       if (other !is GraphQLString) return false
+      if (value != other.value) return false
       return true
     }
 
     override fun hashCode(): Int {
-      return javaClass.hashCode()
```
tasomaniac: Here `hashCode` has been always the same value. Was that intentional?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReader.kt
```diff
@@ -125,12 +124,12 @@ class SimpleResponseReader private constructor(
         val conditionValue = variableValues[condition.variableName] as Boolean
         if (condition.inverted) {
           // means it's a skip directive
-          if (java.lang.Boolean.TRUE == conditionValue) {
```
tasomaniac: I want to say that this is the same as before. But I'm not sure.

apollo-api/src/jvmMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo.api
+
+/**
+ * Represents a custom GraphQL scalar type
+ */
+actual interface ScalarType {
+  actual fun typeName(): String
+
+  fun javaType(): Class<*>
```
tasomaniac: I'm not 100% sure about this. Right now, this is the only choice to keep it compatible. This means, `javaType` will work as before for Jvm/Android.

Right now, this is also used by generated code. So, it would be blocker moving forward to figure this out.

apollo-api/gradle.properties
```diff
@@ -1,4 +1,4 @@
-POM_ARTIFACT_ID=apollo-api
+POM_ARTIFACT_ID=apollo-api-jvm
```
sav007: is it intentionally? I thought we keep the name of artifact the same.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/BigDecimal.kt
```diff
@@ -0,0 +1,7 @@
+package com.apollographql.apollo.api
+
+expect class BigDecimal(value: String?) {
```
sav007: I have in one of my project this https://gist.github.com/sav007/8bf365e7eaa86f4f4b6c890ea3ce47e6

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CustomTypeValue.kt
sav007: Probably oversight

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
```diff
@@ -3,8 +3,6 @@ package com.apollographql.apollo.api
 /**
  * Represents a custom GraphQL scalar type
  */
-interface ScalarType {
+expect interface ScalarType {
   fun typeName(): String
-
-  fun javaType(): Class<*>
```
sav007: Wonder if we should replace this with `KClass`, the only place where it used is to find default scalar type adapter.
`com.apollographql.apollo.api.ScalarTypeAdapters`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReader.kt
sav007: Yeah, that's fine. It was bad port to kotlin.

apollo-api/build.gradle.kts
```diff
@@ -1,18 +1,36 @@
...
+    withJava()
+  }
 
-  add("api", groovy.util.Eval.x(project, "x.dep.kotlin.stdLib"))
-  add("api", groovy.util.Eval.x(project, "x.dep.okio"))
+  sourceSets {
+    val commonMain by getting {
+      dependencies {
+        implementation(kotlin("stdlib-common"))
+        api("com.squareup.okio:okio-multiplatform:2.4.3")
```
sav007: nit: should it be moved to dep.gradle?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo.api
+
+expect class ScalarTypeAdapters(customAdapters: Map<ScalarType, CustomTypeAdapter<*>>) {
```
sav007: What holds us to move this class to common?

apollo-api/src/jvmMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
sav007: We need to think about this, this is used only to locate default scalar type adapter for system types like Int, Boolean, String etc.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
tasomaniac: I tried but it is not possible. It is because the code generation uses this to implement custom types and KClass would not compile in Java code generation.

If we replace it somehow with KClass, then it can be easily moved to common

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
tasomaniac: Answered here https://github.com/apollographql/apollo-android/pull/2062#discussion_r390797343

apollo-api/gradle.properties
tasomaniac: I was just trying stuff. I reverted this soon after pushing.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
sav007: We can replace this with raw string, I can make codegen changes for that.

apollo-api/src/jvmMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
tasomaniac: This will be done in a follow-up by also changing the code generation to accomodate.

tonilopezmr:
This is awesome!

PR #2063, opened 2020-03-13
Refactor `ScalarType`

sav007:
Replace `javaType(): Class<*>` with `className(): String` for better support of KN


tasomaniac:
I had a quick look but didn't continue since this needs rebasing to master. I guess keeping the new implementation in commonMain sourceSet would be fine.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
```diff
@@ -1,10 +1,117 @@
...
+    } as CustomTypeAdapter<T>
+  }
 
   companion object {
-    val DEFAULT: ScalarTypeAdapters
+    @JvmField
+    val DEFAULT = ScalarTypeAdapters(emptyMap())
+
+    private val DEFAULT_ADAPTERS = emptyMap<String, CustomTypeAdapter<*>>() +
+        createDefaultScalarTypeAdapter("java.lang.String", "kotlin.String") { value ->
```
tasomaniac: Is `java.lang.String` enough or should we also check other Strings? Like I wonder what would happen on iOS. Of course can be tackled later.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
```diff
@@ -1,10 +1,117 @@
...
+        createDefaultScalarTypeAdapter("java.lang.Double", "kotlin.Double") { value ->
+          when (value) {
+            is GraphQLNumber -> value.value.toDouble()
+            is GraphQLString -> value.value.toDouble()
+            else -> throw IllegalArgumentException("Can't decode: $value into Double")
+          }
+        } +
+        createDefaultScalarTypeAdapter(FileUpload::class.qualifiedName!!) { value ->
+          when (value) {
+            is GraphQLNumber -> value.value.toDouble()
```
tasomaniac: Why does this say Double for FileUpload?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
sav007: on iOS it must be `kotlin.String` as generated code is platform agnostic (common).

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
sav007: ogh, jeez my bad

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
tasomaniac: 👍 sounds good.

PR #2069, opened 2020-03-15
Upload MPP artifacts to OSS/Bintray

martinbonnin:
This moves most of the logic to the build.gradle.kts file and uploads the `Default` or `KotlinMultiplatform/Jvm` publications depending the projects.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -264,6 +264,15 @@ fun Project.configurePublishing() {
           password = findProperty("bintray.apikey") as String?
         }
       }
+      maven {
+        // Same as the regular bintray repository with the plugin marker hardcoded
+        name = "bintrayMarker"
+        url = uri("https://api.bintray.com/maven/apollographql/android/com.apollographql.apollo.gradle.plugin/;publish=1;override=1")
+        credentials {
+          username = findProperty("bintray.user") as String?
+          password = findProperty("bintray.apikey") as String?
+        }
+      }
 
       maven {
         name = "oss"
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -276,3 +285,48 @@ fun Project.configurePublishing() {
...
+      println("Snapshot deployed on OSS!")
+    }
+  } else {
+    println("Skipping snapshot (branch=$branch pullRequest=$pullRequest)")
+  }
+
+  if (tag.isNullOrBlank().not()) {
+    println("Deploying release to Bintray and Gradle Plugin Portal...")
+    dependsOn(publishTasks("bintray"))
+    dependsOn(":apollo-gradle-plugin:publishApolloGradlePluginPluginMarkerMavenPublicationToBintrayMarkerRepository")
```
martinbonnin: That's a mouthful but I'm not sure it can be significantly reduced...

build.gradle.kts
```diff
@@ -276,3 +285,48 @@ fun Project.configurePublishing() {
...
   }
 }
+
+fun publishTasks(repository: String): List<Task> {
+  val name = repository.capitalize()
+  return subprojects.flatMap {
+    if (it.name == "apollo-api") {
+      // apollo-api is a MPP project
+      tasks.matching {
+        it.name == "publishKotlinMultiplatformPublicationTo${name}Repository"
```
tasomaniac: publishAll task also runs the task starting with publishMetadata, I think that is needed for Gradle metadata.

With this approach we will need to maintain this list. What is the issue with `publishAll...`?

martinbonnin: > With this approach we will need to maintain this list. What is the issue with publishAll...?

We could make it work by removing the `default` publication on `apollo-api` and `apollo-gradle-plugin` and leave the MPP and `java-gradle-plugin` plugins do their work. That'll actually be leaner.

But then comes `apollo-gradle-plugin`. This project has 2 publications:

1. `plugin`
2. `ApolloGradlePluginPluginMarker`

and 2 repositories:

1. `https://api.bintray.com/maven/apollographql/android/apollo-gradle-plugin/;publish=1;override=1`
2. `https://api.bintray.com/maven/apollographql/android/com.apollographql.apollo.gradle.plugin/;publish=1;override=1`

We want 1. to be only published in 1. and 2. to be only published in 2.
If we just call `publishAllPublicationsToBintray`, we will end up with the marker in the wrong path. Since repos and publication are orthogonal, I'm not sure how to do this. I'll investigate a bit.

tasomaniac: Didn't realize Gradle Plugin complexity while commenting this out. Not sure what is the best solution here either.

If we go with this I would recommend adding some more in this list. Like metadata one. Basically running publish task on the apollo-api with `--dry-run` should give the all tasks that are available.

build.gradle.kts
```diff
@@ -276,3 +285,48 @@ fun Project.configurePublishing() {
...
+            || it.name == "publishJvmPublicationTo${name}Repository"
```
tasomaniac: What about this?
```suggestion
        it.name in arrayOf(
            "publishKotlinMultiplatformPublicationTo${name}Repository",
            "publishJvmPublicationTo${name}Repository"
        )
```

build.gradle.kts
tasomaniac: I just run it, we now have `publishIosPublicationToPluginTestRepository` and `publishIosSimPublicationToPluginTestRepository`

martinbonnin:
I'm closing this PR . I'll make a new one after https://github.com/apollographql/apollo-android/pull/2084 that uses the `publishAllPublication` for everything *but* `apollo-gradle-plugin`

PR #2070, opened 2020-03-15
Add iOS targets to API module

sav007:
- Refactor couple classes to make easier to add support for iOS targets
- Provide iOS implementations of expected classes


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -1,5 +1,19 @@
 package com.apollographql.apollo.api
 
-expect class FileUpload constructor(mimetype: String, file: File) {
-  val mimetype: String
+class FileUpload(val mimetype: String, val filePath: String) {
```
tasomaniac: Changing from File to String would be additional binary incompatible change. Are we ok with that?

martinbonnin: Could we keep the `File` constructor in the jvm artifact and expose both ?

tasomaniac: I guess Ivan wanted to keep the full implementation in the common code. To do that you unfortunately need to use expect/actual and have different implementations in common/jvm/ios

sav007: yeah if we keep file then we need to provide full implementation of File, where on iOS it's completely different how you work with files. There is no such thing as File, and for example to read text on iOS:
```
NSString.create(contentsOfFile = path).toString()
```

Secondly, this file upload is hack, there is no GraphQL specs, it supported probably by just one server only. Not sure if it widely used as the common practice is to use just plain http upload to let's say S3, then get url and run mutation to update.

So I would make such changes anyway to avoid extra complexity.

sav007: TBH I was against this feature at first place, I think you shouldn't use GraphQL to upload files.

apollo-api/build.gradle.kts
```diff
@@ -4,6 +4,21 @@ plugins {
 }
 
 kotlin {
+  data class IosTarget(val name: String, val preset: String, val id: String)
```
designatednerd: Totally optional but is it worth making this `iOSTarget` so it's easier to parse visually? I think I'm fine with `iosTargets` below but the uppercase here threw me off

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
designatednerd: FWIW, there is a [graphQL multi-part request spec](https://github.com/jaydenseric/graphql-multipart-request-spec) that's mostly intended to handle `Upload` types. 

However as Apollo [we only recommend it for proof-of-concept and recommend only sending the URL the file was uploaded to via GraphQL](https://blog.apollographql.com/%EF%B8%8F-graphql-file-uploads-with-react-hooks-typescript-amazon-s3-tutorial-ef39d21066a2) in most normal cases.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -1,5 +1,19 @@
...
+
+    if (mimetype != other.mimetype) return false
+    if (filePath != other.filePath) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = mimetype.hashCode()
+    result = 31 * result + filePath.hashCode()
```
designatednerd: Will Kotlin not auto-generate `hashCode` for MPP stuff?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/ResponseJsonStreamReader.kt
```diff
@@ -104,7 +104,7 @@ open class ResponseJsonStreamReader(
     return when {
       isNextNull -> skipNext().let { null }
       isNextBoolean -> nextBoolean(false)
-      isNextNumber -> BigDecimal(nextString(false))
+      isNextNumber -> BigDecimal(nextString(false)!!)
```
designatednerd: Is this something you really want to crash on if it's not working?

sav007: It's not informative but `false` signals here that field is non optional, so we will never crash here but in other place where we resolve field value (inside `nextString` function)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
sav007: What do you mean? It won't as `FileUpload` is not a `data` class.

tasomaniac: Why is it not? 😊

designatednerd: Ah that's what I missed, that it's not a `data class`. However @tasomaniac poses a good question 😄

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/ResponseJsonStreamReader.kt
designatednerd: Shouldn't the value from that be non-nullable if that's the case?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
martinbonnin: I'd like to avoid data classes in public APIs unless really needed as they make it quite hard to ensure source/binary compatibility.
See https://jakewharton.com/public-api-challenges-in-kotlin/

designatednerd: And that's a great answer as to why not do it that way. Thank you @martinbonnin!

tasomaniac: Oh yes. Makes sense.

Issue #2040, opened 2020-03-06
Code generator generates code for invalid mutation


wskinner-eml:
**Summary**
Apollo generates code for queries that have type errors. When a query has a type error, I would expect the code generation to fail.

**Description**
I created a simple [schema](https://gist.github.com/wskinner-eml/a429fe21f5fe0e57a0744860bcc3246a) to isolate this issue. The schema has a single table with a single integer column, `job_id`. 
![Screen Shot 2020-03-05 at 5 14 38 PM](https://user-images.githubusercontent.com/43324541/76113258-a8ce9500-5f98-11ea-8973-aeeb026af2de.png)

My mutation is below:
```graphql
mutation AddRow($job_id: Int!) {
    insert_test(objects: [{job_id: $job_id}]) {
        returning {
            job_id
        }
    }
}
```
Running `gradle generateApolloSources` results in correct code:
```kotlin
data class AddRowMutation(
  val job_id: Int
) : Mutation<AddRowMutation.Data, AddRowMutation.Data, Operation.Variables> {
...
}
```
If I change the mutation parameter to a string, Apollo still generates code. This mutation:
```graphql
mutation AddRow($job_id: String!) {
    insert_test(objects: [{job_id: $job_id}]) {
        returning {
            job_id
        }
    }
}
```
produces this code:
```kotlin
data class AddRowMutation(
  val job_id: String
) : Mutation<AddRowMutation.Data, AddRowMutation.Data, Operation.Variables> {
...
}
```

**Version**
1.3.2

wskinner-eml:
Hey @sav007 - since you are assigned, can you comment on whether this is a legitimate issue, & and when or if it will be addressed?

sav007:
Yes the issue legit, as right now we missing proper validation of argument types during query parsing. I'm going to take a look this soon.

PR #2073, opened 2020-03-16
Refactor GraphQL query validation 

sav007:
Refactor GraphQLDocumentParser:
- Extract exceptions
- Extract some utils extensions
- Extract pos parse validations

Add tests for variables validation
Add fragment validation tests

Part of  https://github.com/apollographql/apollo-android/issues/2040


apollo-compiler/src/test/validation/graphql/FragmentUnknowType.error
```diff
@@ -0,0 +1,7 @@
+Failed to parse GraphQL file /FragmentUnknowType.graphql (7:25)
```
tasomaniac: Typo in the file name. 
```suggestion
Failed to parse GraphQL file /FragmentUnknownType.graphql (7:25)
```

sav007:
This PR is just refactoring (preparation) before the actual fix. To fix original issue is more complicated than I anticipated

tasomaniac:
Yep, I got that from the description but put it wrongly. I think it would be valuable to do it as a hotfix once you have the fix.

PR #2075, opened 2020-03-17
`ApolloClient` API improvements

### LINKED ISSUE #2071: Expose getter for ApolloClient.serverUrl
no

I would like a simple public getter for `ApolloClient.serverUrl`. There are times where I need to rebuild the client when the serverUrl changes (our app supports dynamic URLs). But I would like to know what the "current" url is before I decide to re-build. Example

```
val initialApolloClient = ApolloClient.builder()
        .serverUrl("some URL")
        .build()

// ...some time later
val newClient = if (initialApolloClient.serverUrl == "some new URL") {
    // URLs changed, rebuild
     ApolloClient.builder()
            .serverUrl("some new URL")
            .build()
} else {
    // nothing changed, use initial client
    initialApolloClient
}
```

tasomaniac:
Add getters for ApolloClient properties

Add newBuilder method to `ApolloClient` to be able to copy instance. Had to change some internal APIs in order to accommodate `newBuilder` functionality.

Fixes #2071

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -395,7 +395,7 @@ public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutati
     Builder() {
     }
 
-    Builder(ApolloClient apolloClient) {
+    private Builder(@NotNull ApolloClient apolloClient) {
       callFactory = apolloClient.httpCallFactory;
       serverUrl = apolloClient.serverUrl;
       httpCache = apolloClient.httpCache;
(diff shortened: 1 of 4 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -222,11 +222,20 @@ public void disableSubscriptions() {
 
   /**
    * @return The default {@link CacheHeaders} which this instance of {@link ApolloClient} was configured.
+   * @deprecated Use getDefaultCacheHeaders() instead
    */
+  @Deprecated
```
tasomaniac: We had other instances where we had `get`. Plus, it is nicer in Kotlin. That's why for consistency I wanted to deprecate this one.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -339,6 +395,24 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
     Builder() {
     }
 
+    Builder(ApolloClient apolloClient) {
```
martinbonnin:
```suggestion
    private Builder(@NotNull ApolloClient apolloClient) {
```

* private because there shouldn't be a reason to call it instead of `newBuilder()`, right ?
* @NotNull because the IDE suggested it 🙃

PR #2084, opened 2020-03-21
Publishing: fix KMP artifacts ids

martinbonnin:


apollo-api/build.gradle.kts
```diff
@@ -58,17 +58,6 @@ kotlin {
...
 }
 
-publishing {
-  publications.withType<MavenPublication>().apply {
-    val kotlinMultiplatform by getting {
-      artifactId = "apollo-api-multiplatform"
-    }
-    val jvm by getting {
-      artifactId = "apollo-api"
-    }
```
martinbonnin: I put this logic in the main `build.gradle.kts` so it applies to other KMP module if we have any.

tasomaniac: We actually don't need to do this if users enable Gradle Metadata. This is kind of for backward compatibility. 

We could as well ask our users to enable Gradle Metadata. That way they will just use `apollo-api` and the correct one will be used automatically.

martinbonnin: Am I right that Gradle Metadata is enabled by default on Gradle 6.0 ? If yes, then I'm in favor of using `apollo-api-jvm` so that it matches the default KMP publishing scheme.

tasomaniac: I thought it was but when I tried locally I still needed the enable function in settings.gradle

PR #2088, opened 2020-03-21
Update agp to 3.6

martinbonnin:
Update Android Gradle Plugin version.
AGP 3.6 is especially nice because it publishes  softwareComponents ready to be consumed by `maven-publish`. See https://issuetracker.google.com/issues/37055147

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -1,6 +1,6 @@
 def versions = [
     minAndroidPlugin      : '3.4.2',
-    androidPlugin         : '3.6.0',
+    androidPlugin         : '3.6.1',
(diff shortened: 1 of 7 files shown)


gradle/dependencies.gradle
```diff
@@ -1,5 +1,6 @@
 def versions = [
-    androidPlugin         : '3.4.2',
+    minAndroidPlugin      : '3.4.2',
+    androidPlugin         : '3.6.0',
```
tasomaniac: Should we use 3.6.1?

martinbonnin: Yep, good catch 👍

Issue #1975, opened 2020-02-06
Fragments fail to select on an interface that is different from the response interface


gsabran:
**Description**

Let's consider
```graphql
# schema
interface FooInterface {
  foo: String
}
interface BarInterface {
  bar: String
}
type FooBar implements FooInterface & BarInterface {
  foo: String
  bar: String
}
type Query {
  getFoo: FooInterface
}
schema {
  query: Query
}

# query
query GetFoo {
  getFoo {
    foo
    ... on BarInterface {
      bar
    }
  }
}
```
The codegen fails saying ```Fragment cannot be spread here as objects of type `FooInterface` can never be of type `BarInterface` ``` which is inaccurate.

Maybe related to https://github.com/apollographql/apollo-android/pull/1633

sav007:
I might missing something here but:
- `getFoo: FooInterface'
-  `BarInterface` is not a subtype of `FooInterface`
-  `FooBar ` is not used by anyone

So looks like error is legit or sample is not valid.

gsabran:
Hum. I don't think there's a notion of subtype in GQL, only types that conform to interfaces, and unions that are a list of types.

`getFoo: FooInterface` means that the result of `getFoo` will be any type that conforms to interface `FooInterface`. For instance the result could be a `FooBar`. So it is possible to have an inline fragment on `BarInterface`.

The compiler could maybe throw an error if none of the types implementing `FooInterface` didn't implement `BarInterface`, even though this is not forbidden by the GQL specs afaik.

sav007:
https://spec.graphql.org/June2018/#sec-Interfaces

gsabran:
Thanks for fixing! 🎉

PR #2090, opened 2020-03-22
Fix issue with fragment condition type coercing

sav007:
Closes https://github.com/apollographql/apollo-android/issues/1975

It might be not easy to get the idea what is going on in the original issue, here is another attempt of explanation.

Imagine we have 2 GraphQL interfaces `Foo` and `Bar` and we have object `FooBar` that confirms  to both these interfaces
```
interface Foo {
   foo: String
}

interface Bar {
  bar:String
}

object FooBar: Foo, Bar {
  foo: String
  bar:String
}
```

Now we have query with selection that returns field with  type `Foo`:
```
query TestQuery {
  foo { # field with Foo type 
    foo
  }
}
```

So the issue is during parsing and validating query we don't allow to use inline fragment on interface `Bar`:
```
query TestQuery {
  foo {
    foo
   ... on Bar { # current implementation gives error that fragment cannot be spread here
     bar
  }
}
```

This is an issue, as we have object `FooBar` that confirms to BOTH interfaces `Foo` and `Bar`, means response might have field that has `FooBar` type so we SHOULD allow to use inline fragment on type `Bar`.


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
```diff
@@ -387,38 +387,55 @@ class GraphQLDocumentParser(val schema: Schema, private val packageNameProvider:
...
+      // carry over meta info such as: `description`, `isDeprecated`, `deprecationReason`
+      val decoratedFields = parentFields.map { parentField ->
+        when (schemaType) {
+          is Schema.Type.Interface -> schemaType.fields?.find { it.name == parentField.fieldName }
+          is Schema.Type.Object -> schemaType.fields?.find { it.name == parentField.fieldName }
+          is Schema.Type.Union -> schemaType.fields?.find { it.name == parentField.fieldName }
+          else -> null
+        }?.let { field ->
+          parentField.copy(
+              description = field.description ?: "",
```
tasomaniac: You might want to keep the original value if null. Like `description = field.description ?: description`

Right now, this always overrides it with empty String if field description is null

PR #2091, opened 2020-03-22
Move to GitHub actions

martinbonnin:
Move to Github Actions for MacOS (and potentially later) windows builds.
Also update publishing for KMP artifacts based on https://github.com/apollographql/apollo-android/pull/2069
See https://github.com/martinbonnin/apollo-android/runs/525635751?check_suite_focus=true for an example build


build.gradle.kts
```diff
@@ -300,4 +319,47 @@ fun PublicationContainer.setDefaultPomFields() {
...
   }
+}
+
+tasks.register("publishIfNeeded") {
+  val eventName = System.getenv("GITHUB_EVENT_NAME")
+  val ref = System.getenv("GITHUB_REF")
+  println("publishIfNeeded eventName=$eventName ref=$ref")
+
+  if (eventName == "push" && ref == "ref/heads/master") {
+    println("Deploying snapshot to OSS...")
```
sav007: nit: I thought Gradle project provides logger, should we use proper `project.logger`?

sav007:
Also one question, is there any benefit of making  build / check each apollo module as separate action instead of one large build `./gradlew clean connectedCheck`?

scripts/ci.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "no" |avdmanager create avd --force -n test -k 'system-images;android-28;default;x86_64'
+
+"$ANDROID_HOME"/emulator/emulator -avd test -no-window -gpu swiftshader_indirect -no-snapshot -noaudio -no-boot-anim &
+
+# theorically, we should wait here for the emulator to boot but compilation is slow enough
+# that the emulator will be up by the time we reach the connectedTests. This saves writing
+# some bash code and a few seconds of total build time.
+
+./gradlew clean connectedCheck
+./gradlew build -x checkstyleTest --stacktrace --max-workers=2
```
sav007: 🤔 isn't `./gradlew clean connectedCheck` will do the same? I mean build first?

tasomaniac: I think that is assemble. 
Build is everything including tests, static analysis etc

martinbonnin: I think `build` run tests while `connectedCheck` doesn't. I did this in order to fail faster on the connectedCheck since setting up an android emulator was a bit painful but everything can be put back to the same gradle invocation and gradle should parallelize appropriately.

build.gradle.kts
martinbonnin: Yep, that'll allow users to shut down the logs if they really want to. I should certainly do this inside the Gradle plugin as well.

martinbonnin:
> Also one question, is there any benefit of making build / check each apollo module as separate action instead of one large build ./gradlew clean connectedCheck?

I don't think it's worth it. We want to build sample and integration tests and these depend on compiling most of the other modules. The benefit we would have by sharding per module is potentially offloading compiler/plugin tests to other VMs but I would say it's a lot of added complexity to save 1 or 2 minutes of compile time.

We should ultimately shard per platform though: run macOS and Windows in parallel

build.gradle.kts
martinbonnin: For the record, I just checked the apollo-gradle-plugin module and it doesn't print to stdout at all so we're all good 🎉

build.gradle.kts
```diff
@@ -300,4 +319,47 @@ fun PublicationContainer.setDefaultPomFields() {
...
+  if (eventName == "push" && ref == "ref/heads/master") {
+    project.logger.log(LogLevel.LIFECYCLE, "Deploying snapshot to OSS...")
+    dependsOn(subprojects.flatMap {
+      tasks.matching {
+        if (it.name == "apollo-gradle-plugin") {
+          it.name in arrayOf("publishDefaultPublicationToOjoRepository",
+              "publishApolloGradlePluginPluginMarkerMavenPublicationToOjoRepository")
+        } else {
+          it.name == "publishAllPublicationsToOjoRepository"
+        }
```
LouisCAD, line 338: @martinbonnin 
How can `it.name == "apollo-gradle-plugin"` be true and `it.name in arrayOf("publishDefaultPublicationToOjoRepository",
              "publishApolloGradlePluginPluginMarkerMavenPublicationToOjoRepository")` be true at the same time?

The content of that `if` branch looks incorrect as to what would be a intent that makes sense.

martinbonnin: Yup, that's a bug, the first `it` should be the project, the second `it` should be the task...
We have a very complicated bintray setup right now because we have 1 bintray package for each artifact, which doesn't really scale. I'm planning to streamling this for 2.0 and this code should actually disappear

PR #2106, opened 2020-03-27
Fix duplicate class generation in test folders

tasomaniac:
Modify and auto-add sourceSets only for non-test variants

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
@@ -78,14 +78,18 @@ abstract class DefaultCompilationUnit @Inject constructor(
 
   private fun SourceDirectorySet.findSources(schemaFile: RegularFileProperty) {
     when {
-      apolloVariant.isTest -> srcDirFromVariant(apolloVariant)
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -55,7 +55,7 @@ subprojects {
   }
 
   plugins.withType(com.android.build.gradle.BasePlugin::class.java) {
-    extension.compileOptions {
```
tasomaniac: I just noticed that this is deprecated in AGP 3.6.1 and removed in 4.0

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -65,29 +66,38 @@ abstract class DefaultCompilationUnit @Inject constructor(
...
+    }
+
+    if (!schemaFile.isPresent) {
+      schemaFile.set { resolveSchema(sourceDirectorySet) }
+    }
+  }
+
+  private fun SourceDirectorySet.findSources(schemaFile: RegularFileProperty) {
+    when {
+      apolloVariant.isTest -> srcDirFromVariant(apolloVariant)
```
tasomaniac: Starting from suggestions in the IDE, I ended up refactoring this to make it easier to follow. This is the only behavior change.

I recommend reviewing this in "split" view.

PR #2108, opened 2020-03-27
Fix IdlingResources issue with Watcher

### LINKED ISSUE #1805: ApolloIdlingResource is never idle when used with query watchers
We're using `ApolloIdlingResource` in Espresso tests. It worked well with `ApolloQueryCall`s, but when we transitioned to `ApolloQueryWatcher` for the queries, Espresso tests always hang on the `ApolloIdlingResource` since it's never idle.

I understand it's a tricky situation, because technically a `watcher` is not _idle_ until it's cancelled. However that makes `ApolloIdlingResource` completely useless in Espresso tests, since we'd also like to test specifically the watcher integration (elements getting updated etc.) and besides it just doesn't make sense to cancel it early.

So my question is, is `ApolloIdlingResource` inherently incompatible with query watchers, or it can be made to e.g. idle after first response arrives? It does make some sense in my mind, but I haven't throught it through very much. Pinging @VisheshVadhera since I see you in other idling resource discussions here :)

tasomaniac:
Fixes #1805 

## Solution

Adding a Watcher adds a call internally so that already contributes to the callCount. Watcher itself should not contribute to the call count.

Test case contributed by @odlund 👏

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
@@ -10,6 +10,7 @@ import com.apollographql.apollo.api.Mutation;
 import com.apollographql.apollo.api.Operation;
 import com.apollographql.apollo.api.OperationName;
 import com.apollographql.apollo.api.Query;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/ApolloCallTracker.java
```diff
@@ -160,12 +164,13 @@ void registerMutationCall(@NotNull ApolloMutationCall apolloMutationCall) {
...
     OperationName operationName = apolloMutationCall.operation().name();
     registerCall(activeMutationCalls, operationName, apolloMutationCall);
+    activeCallCount.incrementAndGet();
   }
 
   /**
    * <p>Removes provided {@link ApolloMutationCall} that finished his execution, if it is found, else throws an
    * {@link AssertionError}.</p>
-   *
+   * <p>
```
tasomaniac: I can revert these if wanted. IntelliJ automatically put them

PR #2112, opened 2020-03-28
Apollo Cache module

tasomaniac:

I took the opportunity to "try" moving cache related code into a new module. `apollo-cache`. This is currently only used by `runtime` and should not introduce any behavior changes. Doing this would make much easier to introduce Kotlin Multiplatform in the future since we can do iteratively. 

Now we have:
`apollo-http-cache`
`apollo-http-cache-api`
`apollo-cache`


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/FieldValueResolver.kt
```diff
@@ -0,0 +1,7 @@
+package com.apollographql.apollo.api.internal
+
+import com.apollographql.apollo.api.ResponseField
+
+interface FieldValueResolver<R> {
```
tasomaniac: This and below 2 classes are shared between regular response parsing and cached response parsing. Moving them to `api` in internal package is the easiest solution. Let me know if you have any concerns. The other option would be to create a new module just for there. Like `http-cache-api` one.

apollo-cache/src/main/java/com/apollographql/apollo/internal/cache/normalized/CacheFieldValueResolver.java
```diff
@@ -31,7 +30,7 @@ public CacheFieldValueResolver(ReadableStore readableCache, Operation.Variables
 
   @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
   @Override public <T> T valueFor(Record record, ResponseField field) {
-    switch (field.type()) {
+    switch (field.getType()) {
```
tasomaniac: Just some clean-up after the migration to Kotlin in our `apollo-api` module. More of these should be done.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/FieldValueResolver.kt
sav007: I think it should be ok, in future it can be used by KN runtime.

tasomaniac:
I don't really know what's wrong with this PR. Only this one has issues with GitHub Actions and unfortunately we don't see any logs there. 😔 Not sure how to proceed. All checks are green locally.

PR #2113, opened 2020-03-28
Silence sdkmanager in CI builds

tasomaniac:
Send sdkmanager stdout to >/dev/null because otherwise CI output becomes unmanageably big.


martinbonnin:
👍 I always end up looking at the raw output 😅

PR #2114, opened 2020-03-29
Move migration guide

tasomaniac:
Move migration guide into its own file since we already released version 1.4

Preview: 
https://5e8069a32e7fca00080ebe9f--apollo-android-docs.netlify.com/docs/android/essentials/migration/


docs/source/essentials/migration.md
```diff
@@ -0,0 +1,232 @@
+---
+title: Migration Guides
```
martinbonnin: Nitpicking: call this file migrating-to-1-3-x.md ? We will most likely need migrating-to-2-0-x.md very soon :)

tasomaniac: I was thinking we can add 2.x also here. If we use a header it will also show up at the right nav bar.

martinbonnin: 👍

PR #2120, opened 2020-03-29
[Gradle Plugin] Support Kotlin multiplatform

### LINKED ISSUE #2086: Gradle Plugin adjustments for Kotlin Multiplatform
Current plugin has some assumptions about Kotlin JVM plugin and they need very small adjustments to work with Kotlin Multiplatform projects.

My initial investigation showed that Android flavors and build types support will make it very difficult.

In a Multiplatform project, we cannot have flavor specific common code since iOS and other platforms do not have the same concept. Perhaps when KMP is detected we should only support `commonMain` sourceSet.

tasomaniac:
Support Kotlin Multiplatform projects in the Gradle Plugin

- Apollo Compiler uses multiplatform compatible Throws and IOException
- Since generated code supports "common" code, it should be compatible with all targets.
- Currently only supports `commonMain` and `commonTest`. Looks like supporting per target would make it extremely complex.
- For multiplatform projects, Kotlin code generation cannot be turned off.

Also increased Gradle's memory since with iOS builds, it consumes much more memory.

Fixes #2086

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
@@ -4,8 +4,8 @@ import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FunSpec
 
-val MULTIPLATFORM_THROWS = ClassName.bestGuess("com.apollographql.apollo.api.internal.Throws")
-val MULTIPLATFORM_IO_EXCEPTION = ClassName.bestGuess("okio.IOException")
+private val MULTIPLATFORM_THROWS = ClassName("com.apollographql.apollo.api.internal", "Throws")
+private val MULTIPLATFORM_IO_EXCEPTION = ClassName("okio", "IOException")
 
 fun FunSpec.Builder.throwsMultiplatformIOException() = throws(MULTIPLATFORM_IO_EXCEPTION)
 
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
@@ -1,0 +1,18 @@
+package com.apollographql.apollo.compiler.codegen.kotlin
+
+import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FunSpec
+
+val MULTIPLATFORM_THROWS = ClassName.bestGuess("com.apollographql.apollo.api.internal.Throws")
+val MULTIPLATFORM_IO_EXCEPTION = ClassName.bestGuess("okio.IOException")
+
+fun FunSpec.Builder.throwsMultiplatformIOException() = throws(MULTIPLATFORM_IO_EXCEPTION)
+
+/**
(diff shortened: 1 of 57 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FunSpec
+
+val MULTIPLATFORM_THROWS = ClassName.bestGuess("com.apollographql.apollo.api.internal.Throws")
+val MULTIPLATFORM_IO_EXCEPTION = ClassName.bestGuess("okio.IOException")
+
+fun FunSpec.Builder.throwsMultiplatformIOException() = throws(MULTIPLATFORM_IO_EXCEPTION)
+
+/**
+ * User instead of Kotlin poet throws since we have our own Throws class for Kotlin Multiplatform
```
martinbonnin, line 13:
```suggestion
 * Use this instead of Kotlin poet `throws` since we have our own Throws class for Kotlin Multiplatform
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
```diff
@@ -0,0 +1,18 @@
+package com.apollographql.apollo.compiler.codegen.kotlin
+
+import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FunSpec
+
+val MULTIPLATFORM_THROWS = ClassName.bestGuess("com.apollographql.apollo.api.internal.Throws")
```
martinbonnin, line 7:
```suggestion
private val MULTIPLATFORM_THROWS = ClassName.bestGuess("com.apollographql.apollo.api.internal.Throws")
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
```diff
@@ -0,0 +1,18 @@
...
+val MULTIPLATFORM_IO_EXCEPTION = ClassName.bestGuess("okio.IOException")
```
martinbonnin, line 8:
```suggestion
private val MULTIPLATFORM_IO_EXCEPTION = ClassName.bestGuess("okio.IOException")
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -104,12 +105,13 @@ abstract class DefaultCompilationUnit @Inject constructor(
     }
   }
 
-  fun generateKotlinModels(): Boolean {
-    return generateKotlinModels.orElse(service.generateKotlinModels).orElse(apolloExtension.generateKotlinModels).getOrElse(false)
+  fun generateKotlinModels(): Boolean = when {
+    project.isKotlinMultiplatform -> true
```
martinbonnin, line 109: 👍

martinbonnin:
Looks good to me 🚀 

I added an ios test to the gradle tests. I think we don't want too many of these tests because they take a lot of time. Maybe one or two is enough + a sample ios app ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGenUtil.kt
tasomaniac: I didn't see the constructor when writing this. 😊

Issue #2097, opened 2020-03-23
When an object of a union type receives an unknown type, ApolloParseException is thrown


francescocervone:
**Summary**
I expect that unions behave similar to enum types, that is, mapping unknown new types into an `Unknown` generated type.

**Description**
Let's say you have a union defined as:
```
union Action = Action1 | Action2
```
We deploy our app with a schema that support these two action.
At some point, the backend adds `Action3` to the union `Action`.
```
union Action = Action1 | Action2 | Action3
```
I expect that older apps don't break if `Action3` is received. If that's the intended behavior, it would mean that every time the backend adds a new type to a union we need to deprecate all older apps and release a new app that supports the new schema.

We always map unions into fragments so we don't know actually if this issue happens only with fragments.
When the fragment gets instantiated:
- With the Kotlin code generator an `IllegalStateException` wrapped into an `ApolloParseException` is thrown
- With the Java code generator a `NullPointerException` wrapped into an `ApolloParseException` is thrown

Currently, looks like the iOS generator handles this use case as expected. Our iOS app doesn't have the same issue.

**Version**
1.3.3 (tested with both Kotlin and Java code generators).

P.S.
I'm not sure this is a bug or an intended behavior. It looked more like a bug, so I opened this issue as a bug.

### Edit
I've tested the use case without using fragments. When not using fragments for unions, everything works as expected: if client app receives an unsupported type from the backend, no `ApolloParseException` is thrown. Quoting my previous example, the unknown action `Action3` is mapped into an object like this:
```
Action(
  asAction1 = null,
  asAction2 = null
)
```

sav007:
Could you please share your sample of query that causes this issue?

sav007:
Also could you please provide a full stack of the exception

francescocervone:
@sav007 

Imagine that the schema of the compiled app has:
```
union LoginResponse = LoginSuccess
```
At some point, the backend adds the `LoginFailure` type, and the updated schema is:
```
union LoginResponse = LoginSuccess | LoginFailure
```
Even if the schema has been updated by the backend, out there we still have apps distributed with the old schema, withtout `LoginFailure`.

This is a sample mutation:
```graphql
mutation LoginMutation($username: String!, $password: String!) {
    login(username: $username, password: $password) {
        ...LoginResponseFragment
    }
}

fragment LoginResponseFragment on LoginResponse {
    ... on LoginSuccess {
        token
    }
}
```

[...]

PR #2125, opened 2020-03-31
Fix issue with fragment field type conditions

sav007:
When fragment defined on the same type as parent field, don't generate fragment field type conditions otherwise it breaks parsing response for any new type.

See original issue for more details. 

Now for cases when fragment is defined on type of parent field we don't generate type conditions:
```
query TestQuery {
  hero { # hero here has type Character
    ...HeroDetails
  }  
}

fragment HeroDetails on Character {
   ...
}
```

https://github.com/apollographql/apollo-android/pull/2125/files#diff-988687da010965f3832fb8c9216ff862

Closes https://github.com/apollographql/apollo-android/issues/2097

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/simple_fragment/TestQuery.java
+++ b/apollo-compiler/src/test/graphql/com/example/simple_fragment/TestQuery.java
@@ -360,9 +360,7 @@ public final class TestQuery implements Query<TestQuery.Data, Optional<TestQuery
 
       public static final class Mapper implements ResponseFieldMapper<Fragments> {
         static final ResponseField[] $responseFields = {
-          ResponseField.forFragment("__typename", "__typename", Arrays.<ResponseField.Condition>asList(
-            ResponseField.Condition.typeCondition(new String[] {"Human", "Droid"})
-          )),
+          ResponseField.forFragment("__typename", "__typename", Collections.<ResponseField.Condition>emptyList()),
           ResponseField.forFragment("__typename", "__typename", Arrays.<ResponseField.Condition>asList(
             ResponseField.Condition.typeCondition(new String[] {"Human"})
           ))
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/response/RealResponseReader.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/response/RealResponseReader.java
@@ -199,7 +199,7 @@ public final class RealResponseReader<R> implements ResponseReader {
       resolveDelegate.didResolveNull();
       result = null;
     } else {
-      CustomTypeAdapter<T> typeAdapter = scalarTypeAdapters.adapterFor(field.getScalarType());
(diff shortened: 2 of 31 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/response/RealResponseReader.java
```diff
@@ -199,7 +199,7 @@ public RealResponseReader(Operation.Variables operationVariables, R recordSet,
       resolveDelegate.didResolveNull();
       result = null;
     } else {
-      CustomTypeAdapter<T> typeAdapter = scalarTypeAdapters.adapterFor(field.scalarType());
+      CustomTypeAdapter<T> typeAdapter = scalarTypeAdapters.adapterFor(field.getScalarType());
```
tasomaniac: It would be good to hold off for the depreciation changes in bug fixes. They are because of Kotlin changes and will make it harder to cherry pick into 1.4.x branch.

apollo-compiler/src/test/graphql/com/example/simple_fragment/TestQuery.java
```diff
@@ -360,9 +360,7 @@ public int hashCode() {
 
       public static final class Mapper implements ResponseFieldMapper<Fragments> {
         static final ResponseField[] $responseFields = {
-          ResponseField.forFragment("__typename", "__typename", Arrays.<ResponseField.Condition>asList(
-            ResponseField.Condition.typeCondition(new String[] {"Human", "Droid"})
```
tasomaniac: Just double checking. Is this intentional? I don't really understand what's going on here and Im curious how this works.

sav007: Sure, I will try to explain.

Let's imagine that we have this query:
```
query TestQuery {
  hero { # hero here has type Character
    ...HeroDetails
  }  
}
fragment HeroDetails on Character {
   name
   ...
}
```

Here, `hero` field has interface type `Character` with `Human`, `Droid` implementation in schema v1. 
Now if fragment `HeroDetails` is defined on interface type `Character` (not on concrete type) we were generating meta field like:
```
ResponseField.forFragment("__typename", "__typename", Arrays.<ResponseField.Condition>asList(ResponseField.Condition.typeCondition(new String[] {"Human", "Droid"})
```
with `Human` and `Droid` type conditions, means that in order to read fragment from the response it MUST be on of these types.

Now the issue that with a new version of schema that introduces new implementation of interface type `Character` let's say `Mandalorian` we are going to fail to read fragment (even when it's defined on base interface type `Character`) because we locked ourselves and expect only `Human` or `Droid` types.

This PR fixes this issue by generating this meta field instead:
```
ResponseField.forFragment("__typename", "__typename", Collections.<ResponseField.Condition>emptyList())
```
basically we remove type conditions if fragment defined on interface or union type.

tasomaniac: Thanks 👍

PR #2127, opened 2020-03-31
Apollo cache module

tasomaniac:

I took the opportunity to "try" moving cache related code into a new module. `apollo-cache`. This is currently only used by `runtime` and should not introduce any behavior changes. Doing this would make much easier to introduce Kotlin Multiplatform in the future since we can do iteratively. 

Now we have:
`apollo-http-cache`
`apollo-http-cache-api`
`apollo-cache`


martinbonnin:
Am I correct that `apollo-cache` is only about normalize cache ? If yes, I'd name this module `apollo-normalized-cache` to make it clear that it's different from the `http-cache`.

tasomaniac:
Yes! That's makes sense a lot. Thanks for the suggestion.

tasomaniac:
Will close this for now. I will try one more time with a new module name in a separate PR

PR #2136, opened 2020-04-05
Kotlin Multiplatform Sample

tasomaniac:
Added `multiplatform` sample where it has 3 folders:
- kmp-ios-app
- kmp-android-app
- kmp-lib-sample

Name suggestions for modules are welcome 😄 

## Considerations

- I had the problem described here. It looks like targeting 1.8 for all is safer. Android tools are handling everything for a long time now. https://stackoverflow.com/questions/48988778/cannot-inline-bytecode-built-with-jvm-target-1-8-into-bytecode-that-is-being-bui/56996020#56996020

- The implementations on `ResponseReader` gave me an `AbstractMethodError` crash. Looks like the issue is resolved in Kotlin 1.4. This is because the implementation was in Java. Converting the implementation to Kotlin fixed it. https://youtrack.jetbrains.com/issue/KT-18489

- iOS app only logs into console for now. Next step will be to implement actual application.

## XCode Console Screenshot

Screen Shot 2020-04-05 at 2 07 54 AM


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
```diff
@@ -21,7 +20,6 @@ interface ResponseReader {
 
   fun <T : Any> readObject(field: ResponseField, objectReader: ObjectReader<T>): T?
 
-  @JvmSynthetic
```
martinbonnin, line 24: Not sure I follow. Why removing that/Why did we need that in the first place ?

apollo-runtime/build.gradle.kts
```diff
@@ -1,13 +1,14 @@
...
-  add("api", project(":apollo-api")) // apollo-espresso-support uses some apollo-api internals
-  add("api", project(":apollo-http-cache-api"))
-  add("api", groovy.util.Eval.x(project, "x.dep.okHttp.okHttp"))
-  add("compileOnly", groovy.util.Eval.x(project, "x.dep.jetbrainsAnnotations"))
-  add("implementation", groovy.util.Eval.x(project, "x.dep.cache"))
+  api(project(":apollo-api"))
+  api(project(":apollo-http-cache-api"))
+  api(groovy.util.Eval.x(project, "x.dep.kotlin.stdLib"))
+  api(groovy.util.Eval.x(project, "x.dep.okHttp.okHttp"))
+  implementation(groovy.util.Eval.x(project, "x.dep.cache"))
```
martinbonnin, line 11: So we don't need the `jetbrainsAnnotations` at compileTime anymore ?

gradle/dependencies.gradle
```diff
@@ -1,6 +1,6 @@
 def versions = [
     minAndroidPlugin      : '3.4.2',
-    androidPlugin         : '3.6.1',
+    androidPlugin         : '3.6.2',
```
martinbonnin, line 3: 👍

apollo-runtime/build.gradle.kts
tasomaniac: Once we add Kotlin we don't need them anymore. They're actually transitively added already.

apollo-api/build.gradle.kts
```diff
@@ -58,6 +60,10 @@ kotlin {
   }
 }
 
+tasks.withType<KotlinCompile> {
+  kotlinOptions.jvmTarget = "1.8"
+}
+
```
martinbonnin, line 66: Maybe put this in `apollo-android/build.gradle.kts` next to the others ? https://github.com/apollographql/apollo-android/blob/3e37125cc4f98e6e981c37e877c1cc4f61f17604/build.gradle.kts#L57

samples/multiplatform/kmp-android-app/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    id("com.android.application")
+    id("org.jetbrains.kotlin.multiplatform")
+}
+apply(plugin = "kotlin-android-extensions")
```
martinbonnin, line 7: No plugin marker for this one guess ? 

Also, I'd like to remove this to use ViewBinding instead at some point but definitely not the priority.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
tasomaniac: This somehow allowed us not to implement these methods in Java since they were not visible to Jvm. I think using JvmDefault while targeting 1.8 would work but that's not available in Kotlin common.

samples/multiplatform/kmp-android-app/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    id("com.android.application")
+    id("org.jetbrains.kotlin.multiplatform")
```
martinbonnin, line 5: Do we actually need the multiplatform plugin here ? This module will always be just Android, right ?

samples/multiplatform/kmp-android-app/src/main/kotlin/com/apollographql/apollo/kmpsample/data/ApolloCoroutinesService.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.apollographql.apollo.kmpsample.fragment.RepositoryDetail
+import com.apollographql.apollo.kmpsample.type.OrderDirection
+import com.apollographql.apollo.kmpsample.type.PullRequestState
+import com.apollographql.apollo.kmpsample.type.RepositoryOrderField
+import kotlinx.coroutines.invoke
+
+/**
+ * An implementation of a [GitHubDataSource] that shows how we can use coroutines to make our apollo requests.
+ */
+class ApolloCoroutinesService(private val apolloClient: ApolloClient) {
```
martinbonnin, line 19: It would be nice to move the ApolloCoroutinesService class to the multiplatform lib module since it is common logic between android/ios.

samples/multiplatform/kmp-ios-app/kmp-ios-app/SceneDelegate.swift
```diff
@@ -0,0 +1,111 @@
...
+        let query = GithubRepositoriesQuery(
+            repositoriesCount: 50,
+            orderBy: RepositoryOrderField.updatedAt,
+            orderDirection: OrderDirection.desc
+        )
+        let requestPayload = "{" +
+        "\"operationName\": \"" + query.name().name() + "\", " +
+        "\"query\": \"" + query.queryDocument() + "\", " +
+        "\"variables\": " + query.variables().marshal() +
+        "}"
```
martinbonnin, line 30: What about adding `fun Query.toJson(): String` to factor this code ?

samples/multiplatform/kmp-ios-app/kmp-ios-app/SceneDelegate.swift
```diff
@@ -0,0 +1,111 @@
...
+            // Check for Error
+            if let error = error {
+                print("Error took place \(error)")
+                return
+            }
+     
+            // Convert HTTP Response Data to a String
+            if let data = data, let dataString = String(data: data, encoding: .utf8) {
+                query.parse(source: OkioBuffer())
+                print("Response data string:\n" + (dataString.data(using: .utf8)!.prettyPrintedJSONString! as String))
```
martinbonnin, line 54: 🙌 😃

samples/multiplatform/kmp-ios-app/kmp-ios-app/SceneDelegate.swift
tasomaniac: Yes! definitely. I got this from our docs. I compared it with what we have in runtime. Runtime also adds more in case autoPersist is enabled. We should do that and move this to common code

samples/multiplatform/kmp-android-app/src/main/kotlin/com/apollographql/apollo/kmpsample/data/ApolloCoroutinesService.kt
tasomaniac: For sure. I kind of like to do that in a separate PR. That would allow us to remove networking call from `SceneDelegate`. That is also misplaced atm.

apollo-runtime/build.gradle.kts
martinbonnin: Could we remove the `annotations` part in the README.md ?https://github.com/apollographql/apollo-android/blame/master/README.md#L53

This has always confused me.

samples/multiplatform/kmp-ios-app/kmp-ios-app/SceneDelegate.swift
tasomaniac: Created a follow-up on this https://github.com/apollographql/apollo-android/issues/2137

samples/multiplatform/kmp-android-app/build.gradle.kts
tasomaniac: Addressed here: https://github.com/apollographql/apollo-android/pull/2136/commits/3cb048ceaa38c5164bf8cd946160d6f4a7457b60

samples/multiplatform/kmp-android-app/build.gradle.kts
tasomaniac: https://github.com/apollographql/apollo-android/pull/2136/commits/3cb048ceaa38c5164bf8cd946160d6f4a7457b60

tasomaniac:
@martinbonnin it looks like Travis is not going to be an option since MultiplatformTests are always going to fail.

martinbonnin:
@tasomaniac Actions are green again, let's remove the `.travis.yml`and merge 😅

tasomaniac:
Awesome. I'm guessing it was the memory.

martinbonnin:
Looks like it. Having the CI hang instead of throwing an OOM is super weird though. Hopefully that'll stabilize.

PR #2142, opened 2020-04-06
Apollo normalized cache module

tasomaniac:
Previously reviewed in #2112

I took the opportunity to "try" moving cache related code into a new module. `apollo-normalized-cache`. This is currently only used by `runtime` and should not introduce any behavior changes. Doing this would make much easier to introduce Kotlin Multiplatform in the future since we can do iteratively. 

Now we have:
`apollo-http-cache`
`apollo-http-cache-api`
`apollo-normalized-cache`


tasomaniac:
Merging this since it was already approved previously.

PR #2148, opened 2020-04-07
Migrate some normalized cache classes

tasomaniac:
I had an overall look and started with most straight-forward ones.


tasomaniac:
ran japicmp, the changes are addition of `Companion` class, @Metadata annotation and some functions becoming `final`. Nothing unexpected. Merging now.

Issue #2137, opened 2020-04-05
Create function to prepare request body


tasomaniac:
This is a follow-up on this conversation. https://github.com/apollographql/apollo-android/pull/2136#pullrequestreview-387798354

### TODO

Create a function on Operation class to convert it to Json to be used in request body.

This JSON conversion should be done according to the code we already have in `apollo-runtime`: https://github.com/apollographql/apollo-android/blob/58e17b960fdc181d04d904b616a80e80bb8ff453/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloServerInterceptor.java#L210

Ideally it should be used in there + in iOS sample.

After this is done, please update the docs accordingly: https://www.apollographql.com/docs/android/advanced/no-runtime/

sav007:
@tasomaniac I can take it if you haven't started yet

tasomaniac:
I haven't started on any of them yet. Please go ahead 👍

PR #2150, opened 2020-04-10
Generate 'composeRequestBody' to encode GraphQL POST request payload

sav007:
See https://github.com/apollographql/apollo-android/pull/2150/files#diff-f17956f5f068990894e5e12df8833d96

Closes https://github.com/apollographql/apollo-android/issues/2137

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
@@ -10,6 +10,7 @@ import com.apollographql.apollo.api.internal.json.JsonWriter
 import com.apollographql.apollo.api.internal.json.use
 import okio.Buffer
 import okio.BufferedSource
+import okio.ByteString
 import okio.IOException
 import kotlin.jvm.JvmField
 
@@ -59,6 +60,22 @@ interface Operation<D : Operation.Data, T, V : Operation.Variables> {
    */
   @Throws(IOException::class)
   fun parse(source: BufferedSource): Response<T>
+
+  /**
+   * Composes POST JSON-encoded request body to be sent to the GraphQL server.
+   *
+   * Optional [scalarTypeAdapters] must be provided in case when this operation defines variables with custom GraphQL scalar type.
+   *
+   * *Example*:
+   * ```
+   * {
+   *    "query": "query TestQuery($episode: Episode) { hero(episode: $episode) { name } }",
+   *    "operationName": "TestQuery",
+   *    "variables": { "episode": "JEDI" }
+   * }
+   * ```
+   */
+  fun composeRequestBody(scalarTypeAdapters: ScalarTypeAdapters = ScalarTypeAdapters.DEFAULT): ByteString
 
   /**
    * Abstraction for data returned by the server in response to this operation.
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
@@ -62,9 +62,7 @@ interface Operation<D : Operation.Data, T, V : Operation.Variables> {
   fun parse(source: BufferedSource): Response<T>
 
   /**
-   * Composes POST JSON-encoded request body to be sent to the GraphQL server.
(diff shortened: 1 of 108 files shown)


sav007:
Sample + doc  are comming

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/OperationRequestBodyComposer.kt
```diff
@@ -0,0 +1,46 @@
...
+                  .endObject()
+            }
+          }
+          .apply {
+            if (withQueryDocument) {
+              name("query").value(operation.queryDocument())
+            }
+          }
+          .endObject()
+          .close()
```
tasomaniac, line 43: If we use `use` function, then this `close` will be called automatically and safer.

tasomaniac:
This is awesome 💯

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
```diff
@@ -60,6 +61,34 @@ interface Operation<D : Operation.Data, T, V : Operation.Variables> {
...
+   * *Example*:
+   * ```
+   * {
+   *    "query": "query TestQuery($episode: Episode) { hero(episode: $episode) { name } }",
+   *    "operationName": "TestQuery",
+   *    "variables": { "episode": "JEDI" }
+   * }
+   * ```
+   */
+  fun composeRequestBody(scalarTypeAdapters: ScalarTypeAdapters): ByteString
```
martinbonnin, line 76: Just curious: what's the advantage of  exposing a `ByteString` instead of a `String` ? Do we support something else than utf8 ?

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
```diff
@@ -72,6 +75,32 @@ data class TestQuery(
...
+  override fun composeRequestBody(
+    autoPersistQueries: Boolean,
+    withQueryDocument: Boolean,
+    scalarTypeAdapters: ScalarTypeAdapters
+  ): ByteString = OperationRequestBodyComposer.compose(
+    operation = this,
+    autoPersistQueries = autoPersistQueries,
+    withQueryDocument = withQueryDocument,
+    scalarTypeAdapters = scalarTypeAdapters
+  )
```
martinbonnin, line 102: Does this need to be in the generated models ?  What about just having that as an extension function in kotlin ? That's less codegen code to maintain and it'll make it easier to navigate the generated models ?

martinbonnin:
That's a very nice addition :)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
sav007: Sure we can make it as a `String`, I was optimizing for `okhttp3.RequestBody`
https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/RequestBody.kt#L105

in case of string it will convert it to back to byte array, so we are saving here extra conversions `bytestring -> string -> bytestring`.

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
sav007: I see you concern but I think it should be part of generated models for 2 reasons:

- as it's a symmetric to the generated method implementation of `Operation.parse` 
- and it's more nicer API from the iOS user perspective

martinbonnin: > as it's a symmetric to the generated method implementation of Operation.parse

Could `Operation.parse` be made an extension method ? 

> and it's more nicer API from the iOS user perspective

Indeed. Although I guess a lot of users might still consume this API from Kotlin but for those consuming it in Swift, that'll be way better 👍

sav007: Yeah we could make `parse` as extension function but again for Swift consumers it won't be very nice experience.

PR #2151, opened 2020-04-10
More Kotlin in normalized-cache

tasomaniac:
Convert NormalizedCache and its implementations.

- First converted `NormalizedCache` to Kotlin by removing `Optional`.
- And did necessary changes in other classes in Java first. 
- Then converted those as well.
- ~Moved `prettyDump` to test code since it was only used there.~ reverted

### Backward Compatibility

No change in tests.

`nextCache()` is deprecated and kept as is and returns `Optional` for now.


apollo-android-support/build.gradle.kts
```diff
@@ -1,10 +1,4 @@
 import com.android.build.gradle.BaseExtension
-buildscript {
```
tasomaniac, line 2: This is out of scope but it was redundant.

sav007:
>Moved prettyDump to test code since it was only used there.

Let's keep it in prod as it was useful for finding issue with normalized cache in the past. Several times to understand what was wrong or how normalization was working we were using this function.

tasomaniac:
I see, it was indeed part of the public API. Will move back :+1:

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
```diff
@@ -102,40 +99,35 @@
...
-      }
-    });
+    NormalizedCache nextCache = getNextCache();
+    if (nextCache != null) {
+      nextCache.clearAll();
+    }
     clearCurrentCache();
   }
 
   @Override public boolean remove(@NotNull final CacheKey cacheKey, final boolean cascade) {
```
sav007, line 110: I have to double check entire logic of this function, after quick look at the logic that was before looks suspicious.
It looks like we remove only from the `nextCache` if it present and never from current instance 🤔

tasomaniac: Do you mean the current change is suspicious or the old behavior is? This code is quite well tested.

build.gradle.kts
```diff
@@ -71,7 +71,10 @@ subprojects {
   }
 
   tasks.withType<KotlinCompile> {
-    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+    kotlinOptions {
+      jvmTarget = JavaVersion.VERSION_1_8.toString()
+      freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
```
sav007, line 76: what is the implication on the user side, will it be required to include this compiler arg on the library consumer side as well?

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
tasomaniac: From what I understand, `return result || ...` was doing the same. If the result is true, the rest will never be executed. 

On a second look, I don't think the tests cover having a nextCache scenario

build.gradle.kts
tasomaniac: As far as understand, this makes it an implementation detail. Users shouldn't require anything. We can test that with snapshots after merging this one. If it is not the case, it is easy to remove this. The experimental map/set builder API is just nicer.

apollo-android-support/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.java
tasomaniac: I did another iteration to keep it much closer to the original implementation. Behavior wise, it should be the same but I think this will make it easier to review. Let me know if you see any issues.

I am personally not 100% sure how should the implementation be with the `nextCache` involved.

sav007: No sorry for the confusion, I had to mention that it's not about your changes, I meant the logic that we had before.

The issue is that we are removing the next chained cache (`nextCache()`) and if it  succeed we skip the current one as `return result || deleteRecord` won't call `deleteRecord`.

The reason that it's not an issue is because usually `SqlNormalizedCache` is the last one in the chain but implementation is definitely not correct.

I will fix it in separate PR, let's keep it for now as is.

build.gradle.kts
sav007: I just want to make sure that we don't introduce extra requirements for the users to include this compilation flag on their end whenever they want to use `dump()`. 

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-experimental-stdlib-api/
> Beware using the annotated API especially if you're developing a library, since your library might become binary incompatible with the future versions of the standard library.

It can cause issues in the future if they decide to change API.

tasomaniac: Tbh, this is an inline function and it is not something we expose to consumers.In this particular case, it should be fine.

PR #2152, opened 2020-04-10
Move RealApolloStore to normalized-cache

tasomaniac:
When I first created `normalized-cache` module, I thought `RealApolloStore` couldn't be moved. That is actually the only class left in there that is cache related.

By moving `ResponseWriter` and `ResponseReader` to common module, I was able to move `RealApolloStore` too.


apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/RealApolloStore.java
```diff
@@ -120,11 +113,9 @@ public RealApolloStore(@NotNull NormalizedCache normalizedCache, @NotNull CacheK
   @Override @NotNull public ApolloStoreOperation<Boolean> clearAll() {
     return new ApolloStoreOperation<Boolean>(dispatcher) {
       @Override public Boolean perform() {
-        return writeTransaction(new Transaction<WriteableStore, Boolean>() {
-          @Override public Boolean execute(WriteableStore cache) {
-            optimisticCache.clearAll();
-            return Boolean.TRUE;
-          }
+        return writeTransaction(cache -> {
```
tasomaniac, line 116: Just applied IntelliJ's suggestion to convert these to lambdas. I can revert these if not wanted.

sav007:
>By moving ResponseWriter and ResponseReader to common module, I was able to move RealApolloStore 

Concern here is that those classes `ResponseWriter`, `ResponseReader`, `RealApolloStore`  are used only in runtime. If we move them to `api` module we make them available even for ppl who don't want use our `runtime` module. For instance iOS.

Is that what we want? I mean idea of `api` module to include only essential code / classes required by generated code, without leaking any Apollo specific runtime logic.

tasomaniac:
Fair enough. I am converting normalized cache to Kotlin and the multiplatform. Long way to go still. I thought with RealApolloStore, it would be more complete. But this is not strictly necessary.

Thanks a lot for the feedback 🖖

PR #2158, opened 2020-04-12
Replace legacy Android SQL with SqlDelight

### LINKED ISSUE #2116: Migrate to SqlDelight for persistent cache
This PR #2112 introduces a new cache module where we encapsulate all caching functionality.

Right now, the normalized cache with SQLite is **only** supported on Android with `apollo-android-support` module.

Instead, let's use SqlDelight as a backing implementation to support other platforms. With this migration, we can also keep the `SqlNormalizedCache` in the `apollo-cache` module without requiring additional modules like `android-support`

tasomaniac:
After discussing with @martinbonnin we've decided to create a new module `apollo-sqlite-cache` for the new persisted cache implementation.

## Considerations

- `apollo-android-support` now depends on the new module for backward compatibility purposes.
- `ApolloSqlHelper` is kept 100% backward compatible and deprecated.
- `apollo-normalized-cache-api` module is added for shared API between components.

Fixes #2116


apollo-sqlite-cache/src/main/sqldelight/com/apollographql/apollo/cache/normalized/sql/cache.sq
```diff
@@ -0,0 +1,25 @@
...
+CREATE INDEX idx_records_key ON records(key);
+
+recordForKey:
+SELECT key, record FROM records WHERE key=?;
+
+insert:
+INSERT INTO records (key, record) VALUES (?,?);
+
+update:
+UPDATE records SET key=:key, record=:record WHERE key=:key;
```
martinbonnin, line 16: Nitpicking, maybe not update key ? 

```suggestion
UPDATE records SET record=:record WHERE key=:key;
```

apollo-sqlite-cache/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=apollo-sqlite-cache
```
martinbonnin, line 1: Nitpicking about the naming: what about `apollo-normalized-cache-sqldelight` ?  This would:

* put the module next to `apollo-normalized-cache` and `apollo-normalized-cache-api` when sorted alphabetically
* make a clear distinction between 'normalized' caches and 'http' ones.
* allow for sqlite implementation without sqldelight

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.apollographql.apollo.cache.normalized.CacheKey
+import com.apollographql.apollo.cache.normalized.NormalizedCache
+import com.apollographql.apollo.cache.normalized.Record
+import com.apollographql.apollo.cache.normalized.RecordFieldJsonAdapter
+import okio.IOException
+
+class SqlNormalizedCache(
+    private val recordFieldAdapter: RecordFieldJsonAdapter,
+    private val cacheQueries: CacheQueries
+) : NormalizedCache() {
```
martinbonnin, line 14: What about `SQLDelightNormalizedCache` ?

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -0,0 +1,97 @@
...
+        }
+      }
+    }
+  }
+
+  fun selectRecordForKey(key: String): Record? {
+    return try {
+      cacheQueries.recordForKey(key)
+          .executeAsList()
+          .firstOrNull()
```
martinbonnin, line 66:
```suggestion
          .executeAsOneOrNull()
```

martinbonnin:
Looks great! A few suggestions about naming, as always but it's a great step to multiplatform cache.

2 other comments:
* I feel like `CacheKeyResolver` should belong in the `apollo-normalized-cache-api` module ?
* How hard would it be to add a migration test ? Take a DB generated with the previous cache and run queries with the new code to make sure everything runs well?

apollo-sqlite-cache/src/main/sqldelight/com/apollographql/apollo/cache/normalized/sql/cache.sq
tasomaniac: I saw this too but wanted to keep the same as before 😊 This was like this before.

apollo-sqlite-cache/gradle.properties
tasomaniac: What about `apollo-normalized-cache-sqlite`?

I genuinely think that SqlDelight should be implementation detail. We provide persisted normalized cache backed by `SQLite`.

- We wouldn't provide 2nd implementation side by side
- If we want to replace it in the future, it's easier (not exposed)
- If a user wants to have a different implementation, they can always implement `NormalizedCache` from the API

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
tasomaniac: That would be backward incompatible.

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
tasomaniac: I remember there was a specific reason why I chose asList. I will double check.

apollo-sqlite-cache/gradle.properties
martinbonnin: Sounds good!

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
martinbonnin: I see. Since it had queries in the constructor, I assumed it was already backward incompatible but this is only created through `SqlNormalizedCacheFactory`, right ? Then mark the constructor internal maybe ?

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
tasomaniac: Ok, I see now. `executeAsOneOrNull` is more like `singleOrNull` function in Kotlin. If there are more, it crashes. The old logic was going to over the Cursor and picks the first match too. Since `key` in the database is not `UNIQUE`, we may have multiple of them.

Not sure if we should though 😄 Don't really want to change the behavior now. It can be something we may look in the future.

tasomaniac:
All comments are addressed. I'm not sure about testing the old DB automatically. I manually tested it.

I had an already existing sample apps from `master` and upgraded with this and verified that it works. Also ran the tests on the devices with the old DB existing there.

I would write automated tests if we have a real SQLite migration on the table. To be honest, I don't really know how to do it now.

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
tasomaniac: Yes. That I missed. Before the constructor was package private. It is not something users should be instantiating themselves. They should use the factory.

But they may assign `SqlNormalizedCache` to somewhere. The name is still public. I don't like it either :) but we need to keep it.

tasomaniac:
I wanted to demonstrate backward compatibility by keeping the samples as before. But the fact that I put level as Error in deprecation makes the sample fail to compile. Will fix that

apollo-sqlite-cache/src/main/java/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
martinbonnin: Makes senses 👍

tasomaniac:
Btw, SqlDelight also puts their migration verification task as a dependency to `check` task. That's really good. 

image

PR #2160, opened 2020-04-13
KMP sample: coroutines in common code

### LINKED ISSUE #2139: [Sample] Kotlin Multiplatform Sample 2nd iteration
Move [ApolloCoroutinesService](https://github.com/apollographql/apollo-android/pull/2136/files/fcd5ffde1dce0eda6ca3a86bd32b2af2cccc6772#diff-b03228a4a656662ccb8823151edab952) to `kmp-lib-sample` project's common code and share this between Android and iOS.

Recommended to be done after #2137

tasomaniac:
This PR moves some of the logic from `jvmMain` to `commonMain` and make it possible to re-use more code in the iOS app.

Right now, unfortunately, the `print` in the Swift code is not working. I couldn't even debug it. Any help is appreciated.

Fixes #2139

### Screenshots

We are only logging it to Console. Here is a proof that the parsing works. 

The only code we have in Swift looks like this

```
let repository = ApolloiOSRepositoryKt.create()
    DispatchQueue.main.async {
        repository.fetchRepositories { (repositories) in
            print(repositories)
        }
    }
}
```

image


gradle/dependencies.gradle
```diff
@@ -11,7 +11,7 @@ def versions = [
     jetbrainsAnnotations  : '13.0',
     junit                 : '4.12',
     kotlin                : '1.3.71',
-    kotlinCoroutines      : '1.3.1',
+    kotlinCoroutines      : '1.3.5-native-mt',
```
martinbonnin, line 14: Do we need the `native-mt` version, I think I read somewhere that ktor might not be compatible with it ?

tasomaniac: Just a copy paste. It might be the reason why iOS sample does not return results. Will try without
```suggestion
    kotlinCoroutines      : '1.3.5',
```

samples/multiplatform/kmp-lib-sample/src/iosMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloCoroutinesService.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  actual suspend fun fetchCommits(query: GithubRepositoryCommitsQuery): Response<GithubRepositoryCommitsQuery.Data> =
+      suspendCoroutine {
+        doRequest(query)
+      }
+
+  private fun <D : Operation.Data, T, V : Operation.Variables> doRequest(query: Operation<D, T, V>): Response<T> {
+    val request = createRequest(query)
+    val response = NSURLConnection.sendSynchronousRequest(request, null, null)
+    return query.parse(Buffer().write(response!!.toByteString()))
```
sav007, line 45: as far as I remember KN still doesn't support multiple threaded coroutines. Means you can't offload network execution to background thread using coroutines. Coroutines work only on main thread.

But what we can do is similar to Ktor, schedule network execution on background thread parse response  and then switch back to the main thread (at least that how we using it at Shopify).

If you want I can extract network executor that we have and add to this PR.

tasomaniac: That would definitely be welcome.  This was me trying to learn and implement at the same time 😊

PR #2168, opened 2020-04-14
Fix some of the Kotlin Multiplatform dependency issues

### LINKED ISSUE #2166: Importing the project in IJ/AS
I tried to import the project in IntelliJ 2020.1, Android Studio 3.6.2 and Android Studio 4.1 Canary 5, but on each version I'm encountering some problems: 
 - not all references are resolved
 - GraphQL plugin doesn't provide completion in integration test files
 - untracked files are created and pollute the diff

The most success I've had with AS 3.6.2, but even there I see for example:

image

or 

image

(all imports are actually necessary and used)

and these files are always changed:

image

I tried to import the root project (`./settings.gradle`) as well all the composite one (`./composite/settings.gradle`). The final state I was able to arrive at was importing the composite project, syncing/building everything I could (project compiles successfully) such that I can run tests, but there are still tons of unresolved references and lots of the code completion features don't work. Is there some specific configuration I'm missing, or a specific IJ version that's supported? Some set of required plugins?

tasomaniac:
Fixes #2166

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -6,10 +6,13 @@ import com.apollographql.apollo.ApolloQueryWatcher
 import com.apollographql.apollo.ApolloSubscriptionCall
 import com.apollographql.apollo.api.Response
 import com.apollographql.apollo.exception.ApolloException
-import kotlinx.coroutines.*
-import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.Job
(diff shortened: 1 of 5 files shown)


apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -19,7 +22,7 @@ import kotlinx.coroutines.flow.*
  * @return a flow which emits [Responses<T>]
  */
 @ExperimentalCoroutinesApi
-fun <T> ApolloCall<T>.toFlow() = callbackFlow {
+fun <T> ApolloCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
```
martinbonnin, line 25: Isn't type inference supposed to work here ? Or are you adding this for readability reasons ?

tasomaniac: I think originally this was because of `compileOnly`. I added it. But then didn't remove it since It is a good practice to explicitly define types in public APIs.

PR #2169, opened 2020-04-14
Update docs for 2.0.0 release

tasomaniac:


apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/GradleBuildCacheTests.kt
```diff
@@ -7,7 +7,7 @@ import org.junit.Assert
 import org.junit.Test
 import java.io.File
 
-class CacheTests {
+class GradleBuildCacheTests {
```
tasomaniac, line 10: Renamed because it was confusing with other normalized cache related tests.

PR #2170, opened 2020-04-14
Fix removing graphql files does not remove the generated files

tasomaniac:


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt
```diff
@@ -102,7 +116,7 @@ abstract class ApolloGenerateSourcesTask : DefaultTask() {
     val nullableValueTypeEnum = NullableValueType.values().find { it.value == nullableValueType.getOrElse(NullableValueType.ANNOTATED.value) }
     if (nullableValueTypeEnum == null) {
       throw IllegalArgumentException("ApolloGraphQL: Unknown nullableValueType: '${nullableValueType.get()}'. Possible values:\n" +
-          NullableValueType.values().map { it.value }.joinToString("\n"))
+          NullableValueType.values().joinToString(separator = "\n") { it.value })
```
tasomaniac, line 119: This was suggested by IntelliJ

Issue #2140, opened 2020-04-06
License metadata broken


dsvensson:
The release jars files do not inform the MIT license that at least GitHub has assigned to this project, which prevents automatic tooling for classifying this project, using for example: 

https://github.com/jk1/Gradle-License-Report

...but I guess any tool would have the same problem.

tasomaniac:
Do you know from which version is it broken?

tasomaniac:
Will be fixed with 2.0

PR #2172, opened 2020-04-14
Update maven publishing

martinbonnin:
* use the Android software components available with AGP 3.6
* do not create publications if the java-gradle-plugin or org.jetbrains.kotlin.multiplatform did it for us
* properly configure the pom metadata
* remove the bintrayMarker repository
* upload everything to the merged `apollo` package
* since the publications do not overlap anymore, we can call "publishAllPublications"

### DIFF
--- a/apollo-normalized-cache-sqlite/gradle.properties
+++ b/apollo-normalized-cache-sqlite/gradle.properties
@@ -1,4 +1,4 @@
-POM_ARTIFACT_ID=apollo-sqlite-cache
+POM_ARTIFACT_ID=apollo-normalized-cache-sqlite
 POM_NAME=Apollo GraphQL SQLite Cache
 POM_DESCRIPTION=Apollo GraphQL SQLite Cache
 POM_PACKAGING=jar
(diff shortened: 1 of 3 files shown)


tasomaniac:
@martinbonnin can you check if this is ok with these changes? https://github.com/apollographql/apollo-android/issues/2140

martinbonnin:
Yes, that should fix it.

build.gradle.kts
```diff
@@ -209,43 +208,43 @@ fun Project.configurePublishing() {
...
-            artifactId = findProperty("POM_ARTIFACT_ID") as String?
+            pom {
+              artifactId = findProperty("POM_ARTIFACT_ID") as String?
+            }
           }
         }
       }
+
+      withType<MavenPublication>() {
+        setDefaultPomFields()
```
martinbonnin, line 246: This was missing and preventing the license url to be present in the pom files

apollo-normalized-cache-sqlite/gradle.properties
```diff
@@ -1,4 +1,4 @@
-POM_ARTIFACT_ID=apollo-sqlite-cache
+POM_ARTIFACT_ID=apollo-normalized-cache-sqlite
```
tasomaniac, line 1: Ops

PR #2174, opened 2020-04-15
Enable support for generics Objective-C interop

sav007:
With this compiler argument generated Objective-C headers contains info required for generics interop.


tasomaniac:
Is this only for users or should be also enable it at the library level (at apollo-api)?

sav007:
Yeah it can only be enabled on the user side as it compiler arg for generating iOS framework.

Issue #2164, opened 2020-04-14
`asFlow()` leaks cloned watcher


lwasyl:
Current implementation of `asFlow` is flawed in that it will leave the call active even when the flow's coroutine has been cancelled.

You can observe this behavior on my commit https://github.com/lwasyl/apollo-android/commit/9c6fde79844cf8c31e2f5c2b4e042c7b59ad6d30 in `CoroutinesApolloTest#watcherFlowCancellationCancelsWatcher` test. The test is failing but I left the logs to make it clear what's happening:
```
Awaiting first flow
[FoosCall] Got event SCHEDULED
[FoosCall] Got event FETCH_CACHE
[FoosCall] Got event FETCH_NETWORK
[FoosCall] Got response com.apollographql.apollo.api.Response@7e1bcf9c
[FoosCall] Got event COMPLETED
Cancelling first job
Clearing caches
Enqueueing second response
Calling second query
[BarsCall] Got event SCHEDULED
[BarsCall] Got event FETCH_NETWORK
[FoosCall] Got event SCHEDULED
[FoosCall] Got event FETCH_CACHE
[FoosCall] Got event FETCH_NETWORK 
[FoosCall] Got failure com.apollographql.apollo.exception.ApolloNetworkException: Failed to execute http call
[BarsCall] Got response com.apollographql.apollo.api.Response@cab19945
[BarsCall] Got event COMPLETED
Received response: com.apollographql.apollo.api.Response@cab19945
Second query value received
```

As you can see, even after cancelling the first call coroutine, the FoosCall is still active and even tries to reach out to the network (because the cache has been cleared, so behaving as expected as per #2033)

I belive the bug is in the `asFlow` implementation which creates a `clone()` and subscribes to it, but it cancels the originally passed watcher, thus leaving the cloned one open and leaking. Simply capturing `clone`d watcher and cancelling it instead of `this@toFlow` fixes the issue

tasomaniac:
Thanks for the detailed issue description. Your solutions sounds reasonable. Can you open a Pull Request with that fix? It would be awesome to include this to upcoming 2.0.0 release.

lwasyl:
@tasomaniac Opened https://github.com/apollographql/apollo-android/pull/2175

Issue #2033, opened 2020-03-04
Query watcher re-emitting old value after caches are cleared


lwasyl:
I'm facing some weird behavior, which I'm not even sure if it's a bug or intended behavior. If the second, I'm curious what's the reasoning.

I have two different queries, and (probably related) each of them contains objects of the same types: QueryOne fetches `[A]` and `[B]`, and QueryTwo also fetches `[A]` and `[B]`. What I'm doing then is I:
 - watch `QueryOne`
 - clear caches (HTTP cache and normalized cache)
 - watch `QueryTwo`

On the third step, when watching `QueryTwo`, the first watcher for `QueryOne` emits again, with the previous data. This is pretty, because it happens even when there are no common items between both queries. In particular it happens when all the returned lists in both queries are empty. 

Is this simply lack of optimization, or a bug? Previously I expected that `QueryOne` might emit again if `QueryTwo` actually updates some object returned by `QueryOne`, but with empty lists I'm not so sure. Might it be because both queries return the same root, under which `[A]` and `[B]` are then nested?

Sorry for unclear example, but as usual this happens with our production schema which I'm not sure I can share :\ 

This is with Apollo 1.2.3

sav007:
Each query watchers internally builds set that called `dependentKeys` (com.apollographql.apollo.api.Response#dependentKeys). These are keys that will trigger any query watcher when some other query touched them

The behaviour you see is totally legit as seems like in your example `QueryOne ` and `QueryTwo ` have intersection of `dependentKeys`. Means when you trigger second query it will first fetch data from server, put to cache, and that will trigger `QueryOne` watcher to refetch (when query writes to cache it notifies what cache keys were modified to trigger watchers to referesh)

lwasyl:
I expected something like this. So is this also what's happening when both queries return empty lists? Since in that case none of the returned objects (in both queries) even have an ID for the normalized cache, so I'd expect the first query to actually never emit again. Unless the top-level common object (also without an ID) triggers this new emission as well. Only if this happens, we'd be emitting unnecessarily all the time :/

sav007:
https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloQueryWatcher.java#L40

Yeah and this behaviour was added after fixing this issue  https://github.com/apollographql/apollo-android/issues/822

PR #2175, opened 2020-04-15
Fix cloned watcher leaking (#2164)

lwasyl:
I wanted to add a test case, but couldn't reproduce it without a schema that has a root object, since #2033 wouldn't be triggered. I'm aware the names there aren't ideal, but since I'm not sure if there's a more straightforward way to test I didn't invest time into them. The problem with testing I see is that I don't have a reference to the cloned watcher, so the only observable effect is an additional API call. Is there some other way to trigger refetching of open calls? Anyway, suggestions for testing and/or names are welcome.

I didn't add test for all `toFlow()` variants but applied the same fix. I could try to fit them in a single test case or add a separate test case for each, if needed.

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -23,7 +23,8 @@ import kotlinx.coroutines.flow.callbackFlow
  */
 @ExperimentalCoroutinesApi
 fun <T> ApolloCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
-  clone().enqueue(
+  val clone = clone()
+  clone.enqueue(
       object : ApolloCall.Callback<T>() {
         override fun onResponse(response: Response<T>) {
           runCatching {
@@ -42,7 +43,7 @@ fun <T> ApolloCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
         }
       }
   )
-  awaitClose { this@toFlow.cancel() }
+  awaitClose { clone.cancel() }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sav007:
>I wanted to add a test case, but couldn't reproduce it without a schema that has a root object

Could you please explain what you need, I can make changes to our existing `schema.json` with whatever you need to test

lwasyl:
I need something that triggers behavior explained in https://github.com/apollographql/apollo-android/issues/2033, that is:
 - query A
 - clear cache
 - query B <- should trigger refetching for A

so if I read correctly, I need two different queries with intersecting `dependentKeys`. When I do the same query (so it'd have exactly the same `dependentKeys`) then refetching is _not_ triggered.

The way I reproduced that is by mimicking something we've been hitting internally, where we have a root object for most queries, so whenever we fetch
```
root { 
  a
  b
}
```
clear cache and then fetch a different query
```
root {
  a
}
```
then the first query gets refetched. I played with the existing queries and didn't find any pair that triggered this.

sav007:
hm, what about:
```
query Query1 {
  hero {
    id
    name
  }
}

query Query2 {
  hero {
    name
  }
}
```

lwasyl:
Yep, that works :) I updated the PR

sav007:
@martinbonnin @tasomaniac 
```
com.apollographql.apollo.gradle.MultiplatformTests > ios framework compiles FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in /Users/runner/runners/2.169.0/work/apollo-android/apollo-android/apollo-gradle-plugin/build/testProject with arguments [--stacktrace, :linkDebugFrameworkIosArm64]

    Output:
    Daemon will be stopped at the end of the build after running out of JVM memory

    FAILURE: Build failed with an exception.

    * What went wrong:
    Metaspace

    * Try:
    Run with --info or --debug option to get more log output. Run with --scan to get full insights.

    * Exception is:
    java.lang.OutOfMemoryError: Metaspace
```
We are running out of mem :(

martinbonnin:
This happens in the gradle plugin tests, right ? I just made it so that gradle is relaunched more often. Hopefully that'll help. In the long run we should investigate with VisualVM see if there's a ClassLoader leak somewhere. I'll create an issue for that.

tasomaniac:
We recently increased memory to fix CI issues. With Kotlin Multiplatform we use much more memory now. I'm afraid Gradle tests has their own Gradle instance and has a smaller memory. Adding CLI flag or adding `gradle.properties` file to tests with increased memory should help

PR #2177, opened 2020-04-15
Fix compat issues on iOS target

### LINKED ISSUE #2173: Migrate apollo-api tests to Kotlin Tests
For backward compatibility, we kept `apollo-api` tests in Java. Migrate them now to Kotlin and move them to `commonTest` source set in order to run them both for Jvm and iOS.

tasomaniac:
- Move tests to `commonTest`
- Fix some issues with iOS target:
    - BigDecimal was not implementing Number
    - BigDecimal didn't have equals/hashcode implemented.
    - In some cases, Kotlin would treat types as primitives. Even calling `Boolean::class.simpleName` would return just `boolean` instead of `Kotlin.boolean`
    - While doing unsafe case, iOS would just crash with sigfault instead of throwing ClassCastException.

Fixes #2173


apollo-api/src/iosMain/kotlin/com/apollographql/apollo/api/BigDecimal.kt
```diff
@@ -33,17 +33,41 @@ actual class BigDecimal internal constructor(private val raw: NSDecimalNumber) :
 
   override fun compareTo(other: BigDecimal): Int = raw.compare(other.raw).toInt()
 
+
+
```
sav007, line 37: nit: extra blank line

tasomaniac, line 37:
```suggestion
```

PR #2179, opened 2020-04-15
Add networking layer to KMP sample

sav007:


samples/multiplatform/kmp-lib-sample/src/iosMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloNetworkClient.kt
```diff
@@ -0,0 +1,145 @@
...
+  private fun NSData.toByteArray(): ByteArray {
+    val data: CPointer<ByteVar> = bytes!!.reinterpret()
+    return ByteArray(length.toInt()) { index -> data[index] }
+  }
+
+  @Suppress("NAME_SHADOWING")
+  private fun <T : Operation.Data> Result<T>.dispatchOnMain(continuationRef: COpaquePointer) {
+    dispatch_async_f(
+        queue = dispatch_get_main_queue(),
+        context = DetachedObjectGraph { (continuationRef to this).freeze() }.asCPointer(),
```
martinbonnin, line 131: @sav007 Is there any specific reason to use `DetachedObjectGraph` here compared to just `StableRef` ? I was under the impression that the  advantage of `DetachedObjectGraph` was that it could work with unfrozen references by making sure no one else accesses them but since everything is frozen here, a `StableRef` would work just as well ?

sav007: I think you are right here, have to double-check but looks like no need for detached object wrapper. Will check

sav007:
```
  @Suppress("NAME_SHADOWING")
  private fun <T : Operation.Data> Result<T>.dispatchOnMain(continuationRef: COpaquePointer) {
    val continuationWithResultRef = StableRef.create((continuationRef to this).freeze())
    dispatch_async_f(
        queue = dispatch_get_main_queue(),
        context = continuationWithResultRef.asCPointer(),
        work = staticCFunction { ref ->
          val continuationWithResultRef = ref!!.asStableRef<Pair<COpaquePointer, Result<T>>>()
          val (continuationPointer, result) = continuationWithResultRef.get()
          continuationWithResultRef.dispose()

          val continuationRef = continuationPointer.asStableRef<CancellableContinuation<Response<T>>>()
          val continuation = continuationRef.get()
          continuationRef.dispose()

          when (result) {
            is Result.Success -> continuation.resume(result.value)
            is Result.Failure -> continuation.resumeWithException(result.error)
          }
        }
    )
  }
```

This works perfectly. Will open PR.

PR #2183, opened 2020-04-16
Try to workaround the metaspace error

martinbonnin:
Fork a gradle daemon even more often to try to workaround a MetaSpace error

See https://github.com/apollographql/apollo-android/pull/2175#issuecomment-614393777


martinbonnin:
Looks like the band-aid is working at the moment 🤞

martinbonnin:
We already have a 2g heap space: https://github.com/apollographql/apollo-android/blob/master/gradle.properties#L18

We could add `-XX:MetaspaceSize=512m` there if this happens again and we don't find thee root cause.

tasomaniac:
I'm not sure if that 2gb is applied to Gradle test runs. Gradle tests are run within a separate Gradle build process.

Issue #2134, opened 2020-04-04
Make kotlin-sample project use generated Kotlin models


tasomaniac:
Current Kotlin sample uses Java code generation. Enable Kotlin code generation and update usages accordingly.

SubhrajyotiSen:
Can I pick this up?

tasomaniac:
Please go ahead. That would be awesome.

tasomaniac:
Hey, just a heads-up. #2189 will migrate the samples to AndroidX. That's a minimal change and only imports change. You might have some conflicts.

SubhrajyotiSen:
Thanks for the heads up. I have done a rebase.
I will submit a PR shortly

PR #2196, opened 2020-04-18
Make kotlin-sample project use generated Kotlin models (#2134)

SubhrajyotiSen:


samples/kotlin-sample/build.gradle.kts
```diff
@@ -32,6 +32,7 @@ tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
 }
 
+
```
tasomaniac, line 35:
```suggestion
```

samples/kotlin-sample/src/main/java/com/apollographql/apollo/kotlinsample/data/ApolloCallbackService.kt
```diff
@@ -17,11 +17,11 @@ import com.apollographql.apollo.kotlinsample.type.RepositoryOrderField
  */
 class ApolloCallbackService(apolloClient: ApolloClient) : GitHubDataSource(apolloClient) {
   override fun fetchRepositories() {
-    val repositoriesQuery = GithubRepositoriesQuery.builder()
-        .repositoriesCount(50)
-        .orderBy(RepositoryOrderField.UPDATED_AT)
-        .orderDirection(OrderDirection.DESC)
-        .build()
+    val repositoriesQuery = GithubRepositoriesQuery(
+        50,
```
tasomaniac, line 21: Would it be better to use the named arguments here: Like `repositoriesCount = 50`

tasomaniac: And same in other places similar to this.

Issue #2144, opened 2020-04-07
Convert apollo-normalized-cache module to Kotlin


tasomaniac:
Convert apollo-normalized-cache module to Kotlin

tasomaniac:
Couple Pull Requests merged for this issue and allowed easier integration of SqlDelight. Feel free to pick this up and continue migrating more to Kotlin

SubhrajyotiSen:
Hi. Can I pick this up?

SubhrajyotiSen:
Also, would be better to do it in batches or all remaining java files at one go?

tasomaniac:
Hi, I've tackled part of this in couple of PRs. I think this is good to take over.

Please do not touch the `api` and `sqlite` modules of the normalized cache. I have a local branch converting them to multiplatform. 🎉

You can help by converting from `normalized-cache` module. Please keep changes small and make sure there are almost not changes in the test sources (to ensure backward compatibility)

martinbonnin:
Closed with https://github.com/apollographql/apollo-android/pull/2457

PR #2200, opened 2020-04-20
Fix failing checkstyle in apollo-normalized-cache

SubhrajyotiSen:
Checkstyle failed when running
./gradlew :apollo-normalized-cache:build

This will be helpful when #2144 is picked up


tasomaniac:
In CI and locally we run these with `-x checkstyleTest` to ignore test sources. I will merge this since it's a good change. 

Thanks for the contribution.

PR #2210, opened 2020-04-24
Add missing Java/Kotlin Docs generation for generated classes

sav007:
Use `description` from the GraphQL schema type definition (see `GraphQLDocumentParser.kt`) and use it to generate docs for model classes


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Operation.kt
```diff
@@ -46,6 +47,7 @@ data class Operation(
   }
 
   private fun normalizedOperationName(useSemanticNaming: Boolean, operationNameSuffix: String): String {
+    "".run {  }
```
tasomaniac, line 50:
```suggestion
```

sav007: :)

PR #2211, opened 2020-04-24
Convert cache-api module to multiplatform

tasomaniac:
Convert cache-api module to multiplatform

- More classes are converted to Kotlin `CacheReference`, `Record`, `RecordFieldJsonAdapter` and `RecordWeigher`
- Deprecated `nextCache()` method is removed since we cannot have `Optional` class
- `UUID` is made multiplatform by depending on an open-source library. Let me know if you have any concerns.
- `clone` function is not the same as what we have in Java. That's why it is deprecated. Users can use `toBuilder().build()` instead.

### Tests

I first kept the tests in Java and verified that they work. But then converted them to `commonTest` with the last commit to also test them with iOS implementation.


gradle/dependencies.gradle
```diff
@@ -101,6 +101,7 @@ ext.dep = [
         plugin : "com.squareup.sqldelight:gradle-plugin:$versions.sqldelight"
     ],
     truth                 : "com.google.truth:truth:$versions.truth",
+    uuid                  : "com.benasher44:uuid:0.1.0",
```
sav007, line 104: 👍 nice Ben :)

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.kt
```diff
@@ -0,0 +1,84 @@
...
+import okio.BufferedSource
+import okio.ByteString.Companion.encodeUtf8
+import okio.IOException
+import kotlin.jvm.JvmStatic
+
+/**
+ * An adapter used to serialize and deserialize Record fields. Record object types will be serialized to
+ * [CacheReference].
+ */
+class RecordFieldJsonAdapter {
```
sav007, line 18: should it be an object?

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/CacheKeyResolver.kt
```diff
@@ -2,6 +2,9 @@ package com.apollographql.apollo.cache.normalized
 
 import com.apollographql.apollo.api.Operation
 import com.apollographql.apollo.api.ResponseField
+import kotlin.jvm.JvmField
+import kotlin.jvm.JvmStatic
+import kotlin.jvm.JvmSuppressWildcards
```
sav007, line 7: unused ?

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/CacheKey.kt
```diff
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.cache.normalized
 
+import kotlin.jvm.JvmField
+import kotlin.jvm.JvmStatic
```
sav007, line 4: unused?

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.kt
tasomaniac: I could potentially make it an object but it is not under `internal`. I'm afraid that would break compatibilty.

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/CacheKeyResolver.kt
tasomaniac: These are used. Somehow in a regular `jvm` project, you can use these without import. But in Multiplatform, they need to be explicitly imported.

PR #2213, opened 2020-04-24
Take nullity into account to generate correct code

### LINKED ISSUE #2209: Android Kotlin codegen does not compile after 2.0.0 upgrade
Kotlin codegen generates  code which causes "unnecessary safe call on a non-null receiver of type <your query name>" for queries after 2.0.0 upgrade

Sample codegen where it throws error during build
>  data class _Backrefs(
>     val __typename: String = "CentreBackrefs",
>     val stores__via__centre: List<Stores__via__centre>?
>   ) {
>     fun marshaller(): ResponseFieldMarshaller = ResponseFieldMarshaller.invoke { writer ->
>       writer.writeString(RESPONSE_FIELDS[0], this@_Backrefs.__typename)
>       writer.writeList(RESPONSE_FIELDS[1], this@_Backrefs.stores__via__centre) { value,
>           listItemWriter ->
>         value?.forEach { value ->
>           listItemWriter.writeObject(value?.marshaller())}
>       }
>     }

stores__via__centre list item is expecting it to be nullable but its not and is causing build to fail. Have tried invalidating cache, rebuild, clean and try to re-generate without much sucess

Screenshot of issue
Screen Shot 2020-04-24 at 3 48 20 pm

Issue is similar in 12 other queries across the android app, and it was working fine with 1.4.5

Version info:
apolloGradlePlugin = "2.0.0"
androidGradlePlugin = "3.6.3"

tasomaniac:
Fixes #2209

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
@@ -10,8 +10,24 @@ import com.apollographql.apollo.compiler.ast.FieldType
 import com.apollographql.apollo.compiler.ast.InputType
 import com.apollographql.apollo.compiler.ast.ObjectType
 import com.apollographql.apollo.compiler.ast.TypeRef
-import com.squareup.kotlinpoet.*
+import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.BOOLEAN
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.CodeBlock
+import com.squareup.kotlinpoet.DOUBLE
+import com.squareup.kotlinpoet.FLOAT
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.INT
+import com.squareup.kotlinpoet.KModifier
+import com.squareup.kotlinpoet.LONG
+import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.ParameterizedTypeName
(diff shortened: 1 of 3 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
```diff
@@ -331,25 +347,31 @@ internal object KotlinCodeGen {
...
         is FieldType.Array -> {
           CodeBlock.builder()
               .beginControlFlow(
                   "listItemWriter.writeList(value)·{ value, listItemWriter ->",
-                  List::class.asClassName().parameterizedBy(rawType.asTypeName()))
-              .beginControlFlow("value?.forEach·{ value ->", List::class.asClassName().parameterizedBy(rawType.asTypeName()))
+                  List::class.asClassName().parameterizedBy(rawType.rawType.asTypeName())
+              )
+              .beginControlFlow(
+                  "value?.forEach·{ value ->", // value always nullable in ListItemWriter
```
sav007, line 372: hm, that strange why do we have this in `ResponseWriter`:
```
interface ListWriter<T> {
    fun write(items: List<T>?, listItemWriter: ListItemWriter)
  }
```
I have to double check, as it looks like it should be:
```
interface ListWriter<T> {
    fun write(items: List<T>, listItemWriter: ListItemWriter)
  }
```

tasomaniac: Would be good to merge this PR? I was wondering about this too. It would be great if you can take a look after merging this.

sav007: Yeah, pls merge will look after

PR #2215, opened 2020-04-24
Update OkHttp for important fixes

tasomaniac:
More info here: https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-450


gradle/dependencies.gradle
```diff
@@ -94,7 +92,6 @@ ext.dep = [
         java   : "io.reactivex.rxjava2:rxjava:$versions.rxjava",
         android: "io.reactivex.rxjava2:rxandroid:$versions.rxandroid",
     ],
-    spock                 : "org.spockframework:spock-core:$versions.spock",
```
tasomaniac, line 97: these seem to be unused.

PR #2219, opened 2020-04-25
Make sure generated code do not produce warnings

tasomaniac:
Enable Kotlin flag `allWarningsAsErrors` except for the tests where we expect warnings.

After doing this, discovered a small issue and fixed it.

This will make sure that problems like #2209 will not happen.


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/KotlinCodeGen.kt
```diff
@@ -251,20 +251,17 @@ internal object KotlinCodeGen {
     }
   }
 
-  private fun FieldType.readListItemCode(optional: Boolean): CodeBlock {
+  private fun FieldType.readListItemCode(): CodeBlock {
```
tasomaniac, line 254: We already check null value before this like, so they are never null after this point.

PR #2220, opened 2020-04-25
Migrate Transaction and EvictionPolicy to Kotlin (#2144)

SubhrajyotiSen:


apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/lru/EvictionPolicy.kt
```diff
@@ -0,0 +1,92 @@
...
+
+import com.apollographql.apollo.api.internal.Optional
+import java.util.concurrent.TimeUnit
+
+/**
+ * Controls how long a [com.apollographql.apollo.cache.normalized.Record] will
+ * stay in a [LruNormalizedCache].
+ */
+class EvictionPolicy internal constructor(
+    private val maxSizeBytes: Optional<Long>,
```
tasomaniac, line 11: Can you remove usages of Optional and replace them with Kotlin Nullable type? Of course we need to be careful in places where these are exposed as public API. But I think it is still possible to remove as much as possible.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/lru/EvictionPolicy.kt
```diff
@@ -0,0 +1,92 @@
...
+          expireAfterAccessTimeUnit,
+          expireAfterWrite,
+          expireAfterWriteTimeUnit
+      )
+    }
+  }
+
+  companion object {
+    val NO_EVICTION = builder().build()
+    fun builder(): Builder {
```
tasomaniac, line 87: JvmStatic?

SubhrajyotiSen: I shouldn't have missed that

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/lru/EvictionPolicy.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  companion object {
+
+    @JvmField
+    val NO_EVICTION = builder().build()
+
+    @JvmStatic
+    fun builder(): Builder {
```
SubhrajyotiSen, line 91: @tasomaniac Should I add imports for `@JvmStatic` and `@JvmField` since they are needed for Multiplatform like you mentioned on https://github.com/apollographql/apollo-android/pull/2211

tasomaniac: No need to do it manually. Right now, they're not needed. When we move classes to common code, IntelliJ /Android Studio does this automatically.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/lru/EvictionPolicy.kt
```diff
@@ -0,0 +1,96 @@
...
+
+import com.apollographql.apollo.api.internal.Optional
+import java.util.concurrent.TimeUnit
+
+/**
+ * Controls how long a [com.apollographql.apollo.cache.normalized.Record] will
+ * stay in a [LruNormalizedCache].
+ */
+class EvictionPolicy internal constructor(
+    private val maxSizeBytes: Optional<Long>,
```
tasomaniac, line 11: These are also private. Let's remove `Optional`s here too. You can take a look at this class for reference. Let's make these public and expose as `val maxSizeBytes: Long?`

Then we can keep the below functions with `Optional` for backward compatibility reasons and deprecate them. 

https://github.com/apollographql/apollo-android/blob/master/apollo-normalized-cache-api/src/main/java/com/apollographql/apollo/cache/normalized/NormalizedCache.kt#L129-L130

SubhrajyotiSen: Once I make the changes, should I updated the invocations in `LruNormalizedCache`?

tasomaniac: Especially if it already converted to Kotlin, why not.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/lru/LruNormalizedCache.kt
```diff
@@ -21,21 +21,21 @@ class LruNormalizedCache internal constructor(evictionPolicy: EvictionPolicy) :
...
         }
-        if (evictionPolicy.maxEntries().isPresent) {
-          maximumSize(evictionPolicy.maxEntries().get())
+        if (evictionPolicy.maxEntries != null) {
+          maximumSize(evictionPolicy.maxEntries)
         }
-        if (evictionPolicy.expireAfterAccess().isPresent) {
-          expireAfterAccess(evictionPolicy.expireAfterAccess().get(), evictionPolicy.expireAfterAccessTimeUnit().get())
+        if (evictionPolicy.expireAfterAccess != null) {
+          expireAfterAccess(evictionPolicy.expireAfterAccess, evictionPolicy.expireAfterAccessTimeUnit!!)
```
SubhrajyotiSen, line 35: @tasomaniac I had to use force unwrap here since `expireAfterAccess()` expects a `Nonnull`.

Given that the existing implementation does not check with `isPresent()` for `expireAfterAccessTimeUnit`, it the force unwrap here okay?

tasomaniac: Yeah, I think this is fine. It's the exact same behavior

SubhrajyotiSen:
Thanks for the quick reviews 😃

Issue #2188, opened 2020-04-16
Add UI layer to the iOS sample


tasomaniac:
Our sample app is a really small GitHub clients. It let's the user to display

Repositories, Repository Details and Commits.

The data layer is already done in ApolloiOSRepository class https://github.com/apollographql/apollo-android/blob/master/samples/multiplatform/kmp-lib-sample/src/iosMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloiOSRepository.kt

Need to use that to build the UI. 

Also don't forget to search for `change_me` in the codebase to replace with your personal GitHub token.

PR #2223, opened 2020-04-27
Swift UI funtimes for iOS Sample App on KMP

designatednerd:
Implementation of #2188 (repos pictured are mine since it's my github key I'm using while working on this). 

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-04-26 at 22 01 38
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-04-26 at 22 01 43

Notes: 
- Used SwiftUI instead of storyboards mostly for funsies and practice, also to prove it could be done. ⛰️ 
- Somewhat obnoxious side effect of K/N compiling down to Objective-C instead of Swift is that I needed to make an `@ObservableObject` to handle publishing changes of arrays instead of using `@State` for various properties since `@State` only works on `structs` (and arrays of structs - even if the Array itself is a `struct`, it has to contain `struct`s for the SwiftUI magic™ to work properly) and anything coming out of ObjC is going to be a `class`. 
- Due to stupidity with Objective-C runtime collision, I've had to typealias one of the `RepoFragment`s properties named `description`. On every `NSObject` subclass, `description` is essentially a debug description of an object, and whether the dynamic dispatch will resolve to that implementation or the implementation of a property _named_ `description` is basically a crapshoot, with the dice slightly weighted in favor of Apple's implementation. I don't know if there's a great way to add an iOS-only alias - I think an extension that, within Kotlin, still called into `.description`, would run into the same problem on iOS, but I didn't try that since aliasing was easier.

Questions: 
- ~~On the commits I'm running into the issue that the closure I get back from `fetchCommits(repositoryName:)` is returning `[Any]` - I'm guessing because this could return multiple possible types. Can you point me in the direction of what those possible types are, and how I'd find them in the future?~~ (Addressed)
- I noticed every K/N object is generating this `componentN` for all the properties - I don't think I've seen this in my prior work with K/N, so is this something we're doing (or causing it to do)? It still has the appropriate property names on top of this, but I thought this was weird:
Screen Shot 2020-04-26 at 9 56 02 PM

Now working all the way through and ready for review!


sav007:
Component fields are coming from the fact that our generated models are data classes. Not sure if there is a way to hide them from objc header file

sav007:
The other issue is related to `List` generics with nullable type:
```
fun fetchCommits(repositoryName: String, success: (List<GithubRepositoryCommitsQuery.Edge?>) -> Unit)
```

Looks like there is an issue with ObjectiveC interop https://kotlinlang.org/docs/reference/native/objc_interop.html#generics 

I will reach out to Kevin as far as I remember he was working on adding this feature to KN.

designatednerd:
That helps though, I can at least do some casting to get whatever edges are non-null and of the appropriate type. Thank you!

designatednerd:
@sav007 Still wondering about the `componentN` stuff - I'm definitely seeing that in the generated Obj-C headers here.

sav007:
@designatednerd sorry for the confusion, but that what I said. `componentN`  is generated by default by Kotlin compiler for all `data` classes (the `components` are used in destructuring declarations https://kotlinlang.org/docs/reference/multi-declarations.html). There is no way we can hide those fields from ObjC, (unless we make our models to be a regular classes that I think is not an option).

samples/multiplatform/kmp-ios-app/kmp-ios-app/RepositoryManager.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+    
+    func fetchCommits(for repo: RepositoryFragment) {
+        // NOTE: This comes in as `[Any]` due to some some issues with representing
+        // optional types in a generic array in Objective-C from K/N. The actual
+        // type coming back here is `GithubRepositoryCommitsQuery.Edge`, and the
+        // `Node` it contains is where actual information about the commit lives.
+        self.repository.fetchCommits(repositoryName: repo.name) { [weak self] commits in
+            let filteredCommits = commits
+                .compactMap { $0 as? GithubRepositoryCommitsQuery.Edge }
```
martinbonnin, line 41: Is there a way to force a crash if `$0` is not nil but not of type `GithubRepositoryCommitsQuery.Edge` either? Some swift equivalent of `compactMap { $0 as GithubRepositoryCommitsQuery.Edge? }` in Kotlin?

designatednerd: 🤔 Good question - this isn't really a pattern used much in iOS so I don't think it's a change worth making generally, but I'll mess around and see because now I'm curious :)

designatednerd: OK, survey says that's not a thing in Swift: 
Screen Shot 2020-04-30 at 5 23 42 PM

You also can't do`$0 as? GithubRepositoryCommitsQuery.Edge?` since the type is `[Any]` rather than `[Any?]`: 

Screen Shot 2020-04-30 at 5 30 19 PM

If the type were `[Any?]` then you could do that: 

Screen Shot 2020-04-30 at 5 29 03 PM

But yeah basically you can't do what causes a runtime crash in Kotlin because it causes a compile-time error in Swift.

PR #2232, opened 2020-05-01
Generate synthetic accessor for optional list properties

sav007:
Add generation of synthetic accessor `fun <optionalListField>FilterNotNull()` to mitigate the issue with KN - Obj-C interop when `List<T?>` is translated to Swift as `[Any]`.


sav007:
Case with `List<List<T?>?>` is a little complicated, for that case we would have to generate smth like `fun fooFiltetNotNullFilterNotNull():List<List<T>>`, we definitely can do this but, what do you think?

tasomaniac:
I somehow thought this PR already does that since I saw some recursive calls in there.

I think we should be fine. I'm not sure that is really necessary. We can do that if that request comes from a user.

PR #2234, opened 2020-05-01
Make sqlite module multiplatform

### LINKED ISSUE #2212: Convert http-normalized-cache-sqlite module to Kotlin Multiplatform
Convert http-normalized-cache-sqlite module to Kotlin Multiplatform. Need to integrate with SqlDelight's Native / JVM and Android Drivers.

tasomaniac:
`SqlNormalizedCacheFactory` have now 3 different versions for Jvm, Android and iOS.

It can be used in `common` code but cannot be initialized in common code since every platform specific implementation has its own constructor parameters (as per SqlDelight). Plus, Android's `Context` makes it impossible. 

**Follow-up idea:** We could use `ContentProvider` hack to get a global reference to application `Context` and use that. If we do that, we can keep the most basic constructor in common code.

Tests are moved to automatically test jvm and ios

Fixes #2212


apollo-normalized-cache-sqlite/src/commonTest/kotlin/com/apollographql/apollo/cache/normalized/sql/createDriver.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo.cache.normalized.sql
+
+import com.squareup.sqldelight.db.SqlDriver
+
+expect fun createDriver(): SqlDriver
```
tasomaniac, line 5: By having this in `commonTest`, I was able to keep all tests in common module and run them against multiple platforms.

Issue #2231, opened 2020-04-30
Cache not updating record for the same CacheKey


PierLucaLW:
After migrating to 2.0.0 the cache seems to not override the record for the same key anymore.

I'm using the `SqlNormalizedCacheFactory` with a null database name so that it clears everytime the app is killed. I did update the factory instance creation and the dependency as per the migration guide.

What I'm expecting, and this was working up to 1.4.5, is that when returning the same CacheKey in `fromFieldRecordSet` of the `CacheKeyResolver` this will update the record for that key with the current recordSet.
What's happening instead is that when I try to retrieve it from the cache it always picks the first record saved.

I manually trigger the cache with a call to `apolloStore.write(operation, operationData).execute()` as I'm getting the data I want to use to update it as a response of a mutation. I tried with `writeAndPublish` but the result is the same.

The only workaround I found is to always clear all the cache before with `apolloStore.clearAll().execute()` but this is not ideal. Seems also that `apolloStore.remove(CacheKey.from(id)).execute()` doesn't work to remove a single key. It actually seems that the call to remove is performed after the one to write.

tasomaniac:
Hi, would be possible to share some of you code. Especially the CacheKeyResolver would be useful. I have some ideas but it would be good to reproduce this with a test case.

sav007:
Also is it happening to SQL cache only? Have you tried just in memory cache such as `LruNormalizedCache`?

PierLucaLW:
@tasomaniac here it is
```
class ApolloCacheKeyResolver : CacheKeyResolver() {
    override fun fromFieldRecordSet(field: ResponseField, recordSet: Map<String, Any>): CacheKey {
        if (recordSet.containsKey("my_id") && field.responseName == "my_response") {
            val id = recordSet["my_id"] as String
            return CacheKey.from(id)
        }
        return CacheKey.NO_KEY
    }

    override fun fromFieldArguments(field: ResponseField, variables: Operation.Variables): CacheKey {
        return when (field.responseName) {
            "my_response" ->
                if ((variables.valueMap()["my_id"]?.toString()).isNotNullOrEmpty()) {
                    CacheKey.from(variables.valueMap()["my_id"].toString())
                } else CacheKey.NO_KEY
            else -> CacheKey.NO_KEY
        }
    }
} 

[...]

PR #2236, opened 2020-05-01
Sql Cache record update fix + test (#2231)

PierLucaLW:
#2231

Not sure if you wanna still keep the _id field, in that case we should retreive it from the old record and pass it when updating it

### DIFF
--- a/apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
+++ b/apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
@@ -9,90 +9,91 @@ import com.apollographql.apollo.cache.normalized.RecordFieldJsonAdapter
 import okio.IOException
 
 class SqlNormalizedCache internal constructor(
-        private val recordFieldAdapter: RecordFieldJsonAdapter,
-        private val cacheQueries: CacheQueries
+    private val recordFieldAdapter: RecordFieldJsonAdapter,
+    private val cacheQueries: CacheQueries
 ) : NormalizedCache() {
 
-    override fun loadRecord(key: String, cacheHeaders: CacheHeaders): Record? {
-        val record = selectRecordForKey(key)
+  override fun loadRecord(key: String, cacheHeaders: CacheHeaders): Record? {
+    val record = selectRecordForKey(key)
 
-        if (record != null) {
-            if (cacheHeaders.hasHeader(EVICT_AFTER_READ)) {
-                deleteRecord(key)
-            }
-            return record
+    if (record != null) {
+      if (cacheHeaders.hasHeader(EVICT_AFTER_READ)) {
+        deleteRecord(key)
+      }
+      return record
+    }
+    return nextCache?.loadRecord(key, cacheHeaders)
+  }
+
+  override fun clearAll() {
+    nextCache?.clearAll()
+    cacheQueries.deleteAll()
+  }
+
+  override fun remove(cacheKey: CacheKey, cascade: Boolean): Boolean {
+    val result: Boolean = nextCache?.remove(cacheKey, cascade) ?: false
+    if (result) {
+      return true
+    }
+    return if (cascade) {
+      selectRecordForKey(cacheKey.key)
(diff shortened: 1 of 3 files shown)


tasomaniac:
This approach will need migration. You've mentioned another potential solution in the code. Should we go for that? Looks much safer.

PierLucaLW:
At first I thought the problem was there but that part was correct since oldRecord var gets its fields updated after `.mergeWith()`. The `UPDATE` instruction doesn't seem to work on it and `INSERT OR REPLACE` looks like it needs all the field, or at least the id since is a primary key. I can try getting the id from the old record and pass it when to update, would that be good?

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -9,89 +9,90 @@ import com.apollographql.apollo.cache.normalized.RecordFieldJsonAdapter
...
-      null
+    fun selectRecordForKey(key: String): Record? {
+        return try {
+            cacheQueries.recordForKey(key)
+                    .executeAsList()
+                    .firstOrNull()
+                    ?.let {
+                        Record.builder(it.key)
+                                .addFields(recordFieldAdapter.from(it.record)!!)
+                                .addField("_id", it._id)
```
PierLucaLW, line 70: you might want to pass this in a nicer way, just showing it works here

apollo-normalized-cache-sqlite/src/commonTest/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCacheTest.kt
```diff
@@ -54,7 +54,19 @@ class SqlNormalizedCacheTest {
   }
 
   @Test
-  fun testRecordMerge() {
+  fun testRecordMerge_noOldRecord() {
```
sav007, line 57: I think you missing `createRecord(STANDARD_KEY)` before `cache.merge(`

sav007:
Also could you please fix formatting

apollo-normalized-cache-sqlite/src/commonTest/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCacheTest.kt
PierLucaLW: I've split the test in 2 for when a record is already present or not to test both branches

tasomaniac:
I'm still not sure if this is the right way. 1.4.x versions or any older do not replace into existing. So, I'm not sure why we would need a behavior change here.

The solution to the bug must be in the code. There must be something wrong with Kotlin migration.

sav007:
The changes in this fix look legit, but I agree with @tasomaniac before merging we have to find why it was working up to 1.4.5 version. As we might be missing bigger issue with migrating SQL module to Kotlin.

Going to try 1.4.4 version to run the test.

sav007:
🤦 🤦 🤦 🤦 🤦 🤦 
The issue is 🥁  🥁  🥁  🥁  🥁  🥁 
https://github.com/apollographql/apollo-android/blob/master/apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt#L56

```
cacheQueries.update(oldRecord.key, recordFieldAdapter.toJson(oldRecord.fields))
```

but it should be
```
cacheQueries.update(recordFieldAdapter.toJson(oldRecord.fields), oldRecord.key)
or even safer
cacheQueries.update(record = recordFieldAdapter.toJson(oldRecord.fields), key = oldRecord.key)
```

as it generated from the 
```
update:
UPDATE records SET record=:record WHERE key=:key;
```

sav007:
So the issue is not in SQL DML but with the order of passed arguments, I opened new PR https://github.com/apollographql/apollo-android/pull/2245 that fixes it.

PierLucaLW:
@tasomaniac @sav007 ooh I see, haven't thought about that at all. Alright will just close this one then, thanks for looking into it.

Issue #2184, opened 2020-04-16
Out of Memory: Metaspace during apollo-gradle-plugin tests


martinbonnin:
See https://github.com/apollographql/apollo-android/pull/2175#issuecomment-614393777, we're having Metaspace issues again. 

It looks like this is a Classloader leak somewhere. Right now the workaround is to fork a new Gradle instance more often but we should investigate the root cause behind this.

tasomaniac:
Looks like this is not happening anymore. It could be that the test execution is longer now. Even with that, I'm not sure if this is high priority.

PR #2240, opened 2020-05-02
Give more metaspace to tests processes

martinbonnin:
This PR gives more MetaSpace to the test processes. 

fixes https://github.com/apollographql/apollo-android/issues/2184.

It also bumps Gradle to 6.4-rc-4. I was hoping that this could somehow help but doesn't look like it does. Still, I think it's worth upgrading for the nicer error messages and to stay on the Gradle bleeding edge. Happy to revert to 6.3 if you think this is risky.

More context on the Gradle community slack: https://gradle-community.slack.com/archives/CA745PZHN/p1588352513265400

### DIFF
--- a/apollo-gradle-plugin/build.gradle.kts
+++ b/apollo-gradle-plugin/build.gradle.kts
@@ -29,11 +29,6 @@ dependencies {
 }
 
 tasks.withType<Test> {
-  // Restart the daemon once in a while or we end up running out of MetaSpace
-  // It's not clear if it's a real ClassLoader leak or something else. The okio timeout thread does hold some ClassLoaders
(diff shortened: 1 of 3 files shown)


sav007:
I think it's fine to move to Gradle RC as long as it won't break dev experience. 
In case if we decide to make a new release does this mean consumer must migrate to RC as well?

martinbonnin:
> In case if we decide to make a new release does this mean consumer must migrate to RC as well?

It really shouldn't. As long as we don't use any new API, we can build the plugin with a given version of gradle and use it with older versions. We support Gradle 6.0+ at the moment and we have tests for that: https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/GradleVersionTests.kt

Issue #2205, opened 2020-04-22
Expose http response headers


lttlang:
I have a situation where we have information being returned from the server which is not part of the GQL response. Specifically, we add "upgrade_recommended" to the response header so our apps know to show the user a dialog to indicate "hey, you'll need to update your app version soon". We don't block the user from using the app, so it's not an error. It's just a way for us to make sure users have enough warning in advance before we stop supporting that version of the app. 

I know that we can add an Interceptor to the OkHttpClient, which we do for logging, but this is a scenario where we would want a specific "screen" to handle the info and update UI for the user.

From what I can see, `Response` is passed from `InterceptorResponse` which also holds the http response. If the `Response` held and exposed it's http response, I'd be golden!
simple ex:
```
this.enqueue(object : ApolloCall.Callback<Data>() {
    override fun onResponse(response: Response<Data>) {
        val headers = response.rawResponse.headers()
        // do something on success
    }

    override fun onFailure(e: ApolloException) {
        // do something on error
    }
})
```

lttlang:
@sav007 Can you give me an idea of the timeline for this? I'm assuming that exposing the headers is something makes sense.

PR #2248, opened 2020-05-06
Introduce GraphQL execution context

sav007:
Provide type safe container to customize and extend `Response` with additional context, such as providing access to raw OkHttp response.

Example of getting access to the http response headers:
```
val response: Response<Query> = ...
val httpResponse: okhttp3.Response = response.executionContext[HttpExecutionContext.Key]?.response
httpResponse.headers
```

Closes https://github.com/apollographql/apollo-android/issues/2205


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ExecutionContext.kt
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A context of GraphQL operation execution, represented as a set of [Key] keys and corresponding [Element] values.
+ */
+class ExecutionContext private constructor(private val context: Map<Key<*>, Element>) {
+
+  /**
+   * Returns the element with the given [key] from this context or `null`.
+   */
+  @Suppress("UNCHECKED_CAST")
```
designatednerd, line 15: This gives an `UNCHECKED_CAST` warning even though you're using `as?`?

apollo-runtime/src/main/java/com/apollographql/apollo/http/HttpExecutionContext.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.apollographql.apollo.api.ExecutionContext;
+import com.apollographql.apollo.api.internal.Utils;
+import okhttp3.Response;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Http GraphQL execution context, provides access to the raw {@link okhttp3.Response} response.
+ */
+public class HttpExecutionContext implements ExecutionContext.Element {
```
designatednerd, line 11: Might be worth making this `OkHttpExecutionContext` to make it clearer it's tied to the specific `okhttp` implementation

apollo-runtime/src/main/java/com/apollographql/apollo/internal/util/HttpResponseUtils.java
```diff
@@ -0,0 +1,33 @@
+package com.apollographql.apollo.internal.util;
+
+import okhttp3.Response;
+import org.jetbrains.annotations.NotNull;
+
+import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
+
+public final class HttpResponseUtils {
```
designatednerd, line 8: Same thought of marking this as tied to `okhttp`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ExecutionContext.kt
sav007: Yep, for some reason it gives warning.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloParseInterceptor.java
```diff
@@ -81,12 +82,15 @@ public void interceptAsync(@NotNull final InterceptorRequest request, @NotNull A
...
-        OperationResponseParser parser = new OperationResponseParser(operation, responseFieldMapper, scalarTypeAdapters,
-            normalizer);
-        Response parsedResponse = parser.parse(httpResponse.body().source())
+        final OperationResponseParser parser = new OperationResponseParser(operation, responseFieldMapper, scalarTypeAdapters, normalizer);
+        final OkHttpExecutionContext httpExecutionContext = new OkHttpExecutionContext(httpResponse);
+        Response parsedResponse = parser.parse(httpResponse.body().source());
+        parsedResponse = parsedResponse
             .toBuilder()
             .fromCache(httpResponse.cacheResponse() != null)
+            .executionContext(parsedResponse.getExecutionContext().set(OkHttpExecutionContext.KEY, httpExecutionContext))
```
martinbonnin, line 91: What types of objects do we plan to put in the context?

CoroutinesContext has `Dispatcher`, `Job`, `Name`, maybe more
ExecutionContext will have `OkHttpExecutionContext` and what else ? Do we plan to put dispatchers/executors down there too ?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ExecutionContext.kt
designatednerd: that's...goofy, given that the `as?` returns `null` if the cast fails... 

`¯\_(ツ)_/¯`

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloParseInterceptor.java
sav007: For now we put just http context, but in future for new runtime we are going to put cache policy, cache headers, cache normalization records etc. Also any implementation of future request executor (now it's called Apollo interceptor) can put whatever it wants.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ExecutionContext.kt
tasomaniac: We could open a bug ticket to Jetbrains.

apollo-runtime/src/main/java/com/apollographql/apollo/http/OkHttpExecutionContext.java
```diff
@@ -0,0 +1,46 @@
...
+import com.apollographql.apollo.api.ExecutionContext;
+import okhttp3.Response;
+import org.jetbrains.annotations.NotNull;
+
+import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
+
+/**
+ * Http GraphQL execution context, provides access to the raw {@link okhttp3.Response} response.
+ */
+public class OkHttpExecutionContext implements ExecutionContext.Element {
```
tasomaniac, line 12: Just a small thing. Since all modules has Kotlin dependency transitively, it is also added to runtime module. So it is ok to write this now with Kotlin.

Issue #2222, opened 2020-04-26
MegaIssue: Kotlin Native Apollo client / runtime


sav007:
**Is your feature request related to a problem? Please describe.**
Apollo client moving towards KN, there is a need of having KN runtime / client module that is responsible for GraphQL operation execution.

**Describe the solution you'd like**
One possible and straightforward solution is to port existing Apollo runtime module to KN. Even if this looks like reasonable thing to do there are 2 major concerns regarding this:
- our existing runtime is so deeply coupled with OkHttp network library that limits us on adding new transport layers such as web socket, mocked network client. Community were asking us to add such features but with current arch it's almost impossible. On top of that with KN there is no multi-platform implementation of OkHttp.
- having public API in `ApolloClient` related to network / normalized cache,  always raised concerns from large companies (like Shopify, Netflix etc.) regarding been too opinionated, the feedback we got `Apollo is like all or nothing in solution `. Sometimes such companies have own network stack, infra to perform GraphQL operations. We made a huge step towards making our models been used without runtime but we need to redesign our runtime API and make cache layer is truly optional even from API perspective.

As a staring point of conversation regarding the vision of new design for our runtime module here is the high level diagram:
<img width="1324" alt="ApolloClient" src="https://user-images.githubusercontent.com/902816/80317091-cae1d800-87cf-11ea-8a1b-97135adbfcfe.png">

**Network Transport Layer:**

Responsible for network communication, establishing connection, sending and receiving GraphQL request / response

```
interface GraphQLNetworkTransport {
  fun <T> send(request: Request<T>): Flow<Response<T>>
}

class Request<T>(
  val operation: Operation<*, T, *>,
  val scalarTypeAdapters: ScalarTypeAdapters,
  val executionContext: ExecutionContext // similar idea as coroutine context
)

sealed class Response<T> {

  data class Success<T>(
    val operation: Operation<*, T, *>,
    val data: T?,
    val errors: List<Error>,
    val executionContext: ExecutionContext // similar idea as coroutine context
  ) : Response<T>()

  data class Failure<T>(
    val operation: Operation<*, T, *>,
    val error: ApolloError,
    val executionContext: ExecutionContext // similar idea as coroutine context
  ) : Response<T>() 
}
```

The reason why `send` returns `Flow` instead of being `suspended` is because of GraphQL subscriptions that can emit multiple responses.

[...]

PR #2256, opened 2020-05-08
Kotlin runtime initial structure

sav007:
Introduce new `apollo-runtime-kotlin` module with initial structure:
- `ApolloClient` similar to JVM version, a facade for user interface
- `RequestExecutor` is a business logic layer responsible for query execution. Allows to create a chain / pipeline of several executors. Each executor observes, modifies, and potentially short-circuits GraphQL operation requests. `NetworkExecutor` is a final / terminal request executor in the chain, responsible for network execution via configured `NetworkTransport`
- `NetworkTransport` is an abstraction for networking. Apollo is going to provide several `NetworkTransport` implementations: Http(JVM/iOS) and WebSocket(JVM/iOS)
- `DispatchersContext` provides configuration for coroutine dispatchers, that will be put into `ExecutionContext`

See [ApolloClientTest](https://github.com/apollographql/apollo-android/pull/2256/files#diff-855ea0b177f9e3728e24f96987a4e4af) as an idea how everything connected.

Here is the rough idea how normalized cache will be integrated:
- it's going to be completely isolated from runtime
- it will copy `ApolloClient` with injected executor into `ApolloClient#executors` to provide cache functionality via extension function `fun ApolloClient.withCache(<some config>) -> ApolloClient`
- it will add own execution context into `Apollo#executionContext` to provide default configuration such as default cache policy
- it will provide extension functions for `ApolloQueryCall` that will make copy of `RealApolloCall` with and injected request execution context (in such way we can provide request cache policy that will override default one) via `fun ApolloQueryCall.cachePolicy(CachePolicy.CacheFirst) -> ApolloQueryCall`
- cache executor will handle all cache logic by intercepting operation execution before `NetworkExecutor`

Part of #2222


apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/mock/TestLoggerExecutor.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.apollographql.apollo.executor.ExecutionRequest
+import com.apollographql.apollo.api.Response
+import com.apollographql.apollo.executor.RequestExecutor
+import com.apollographql.apollo.executor.RequestExecutorChain
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.onStart
+
+object TestLoggerExecutor : RequestExecutor {
```
sav007, line 12: Sample of custom user `RequestExecutor `

apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/TestUtils.kt
```diff
@@ -0,0 +1,7 @@
+package com.apollographql.apollo
+
+import kotlinx.coroutines.CoroutineScope
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+expect fun <T> runBlocking(context: CoroutineContext = EmptyCoroutineContext, block: suspend CoroutineScope.() -> T): T
```
sav007, line 7: KN doesn't provide `runBlocking` in common coroutines artifact, only in JVM and KN.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/NetworkRequest.kt
```diff
@@ -0,0 +1,8 @@
+package com.apollographql.apollo.network
+
+class NetworkRequest(
+    val operationName: String,
+    val document: String,
+    val variables: String,
+    val extensions: String = ""
+)
```
designatednerd, line 8: silly question: Why is this a `class` and not a `data class`?

apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/ApolloClientTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            body = Buffer().write("{\"data\":{\"name\":\"MockQuery\"}}".encodeUtf8()),
+            executionContext = ExecutionContext.Empty
+        )
+    )
+
+    val response = runBlocking {
+      apolloClient
+          .query(MockQuery())
+          .execute()
+          .retryWhen { cause, attempt -> cause is ApolloException && attempt == 0L }
```
designatednerd, line 101: omg i want the `retryWhen` API for everything I do now

apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/mock/MockQuery.kt
```diff
@@ -0,0 +1,76 @@
...
+  override fun parse(source: BufferedSource): Response<Data> {
+    return parse(source, ScalarTypeAdapters.DEFAULT)
+  }
+
+  override fun parse(byteString: ByteString): Response<Data> {
+    return parse(byteString, ScalarTypeAdapters.DEFAULT)
+  }
+
+  override fun composeRequestBody(scalarTypeAdapters: ScalarTypeAdapters): ByteString {
+    TODO("Unsupported")
```
designatednerd, line 64: Do these need tot be `TODO` or should they actively throw an exception?

apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/ApolloClientTest.kt
sav007: yeah nice that we get this out of the box with Flow

apollo-runtime-kotlin/src/commonTest/kotlin/com/apollographql/apollo/mock/MockQuery.kt
sav007: technically they throw exceptions, but I will replace with explicit ones

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.apollographql.apollo.internal.RealApolloCall
+import com.apollographql.apollo.network.NetworkTransport
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+
+@ApolloExperimental
+@ExperimentalCoroutinesApi
+class ApolloClient constructor(
+    private val networkTransport: NetworkTransport,
+    private val scalarTypeAdapters: ScalarTypeAdapters = ScalarTypeAdapters.DEFAULT,
+    private val executors: List<RequestExecutor> = emptyList(),
```
martinbonnin, line 20: why not use the `interceptors` wording here instead ? `Executors` always remind me of https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html but maybe I've done too much java.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/NetworkRequest.kt
```diff
@@ -0,0 +1,8 @@
+package com.apollographql.apollo.network
+
+class NetworkRequest(
+    val operationName: String,
+    val document: String,
+    val variables: String,
+    val extensions: String = ""
```
martinbonnin, line 7: I guess `operationId` will need to be in there as well ?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/internal/RealApolloCall.kt
```diff
@@ -0,0 +1,39 @@
...
+    val request = ExecutionRequest(
+        operation = operation,
+        scalarTypeAdapters = scalarTypeAdapters,
+        executionContext = executionContext
+    )
+    return flow {
+      emit(RealRequestExecutorChain(executors))
+    }.flatMapLatest { executorChain ->
+      executorChain.proceed(request)
+    }
```
martinbonnin, line 37: Why not just:
```
return RealRequestExecutorChain(executors).proceed(request = request)
```
?

sav007: Because we want make returned flow to be re-tryable, means when user uses `.retry()` flow API we should re-try whole chain of executors not just a final version of built flow.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
sav007: Sure we can name it as interceptors, no objection.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/executor/NetworkExecutor.kt
```diff
@@ -0,0 +1,90 @@
...
+@ExperimentalCoroutinesApi
+class NetworkExecutor(
+    private val networkTransport: NetworkTransport
+) : RequestExecutor {
+
+  override fun <T> execute(request: ExecutionRequest<T>, executorChain: RequestExecutorChain): Flow<Response<T>> {
+    return flow { emit(request.toNetworkRequest()) }
+        .flatMapLatest { networkRequest -> networkRequest.execute(request.executionContext) }
+        .map { networkResponse -> networkResponse.parse(request) }
+        .flowOn(request.executionContext.ioDispatcher)
```
martinbonnin, line 31: Do we need the dispatcher to be configurable per-request ? I was thinking this could be a property of the `NetworkTransport` instead ?

sav007: `ApolloClient` configures `RealApolloCall`:
```
class ApolloClient (...) {
  ...
  fun <D : Operation.Data, T, V : Operation.Variables> query(query: Query<D, T, V>): ApolloQueryCall<T> {
    return RealApolloCall(
        operation = query,
        scalarTypeAdapters = scalarTypeAdapters,
        executors = executors + NetworkExecutor(networkTransport),
        executionContext = executionContext
    )
  }
```

Then `RealApolloCall` configures `ExecutionRequest`:
```
val request = ExecutionRequest(
        operation = operation,
        scalarTypeAdapters = scalarTypeAdapters,
        executionContext = executionContext
    )
```

So dispatchers will be set only once in `ApolloClient` and will carry over to the request and every executor can get access to them. 

In this way we have one place where we can set dispatchers per `ApolloClient` instance.

martinbonnin: > So dispatchers will be set only once in ApolloClient and will carry over to the request and every executor can get access to them.

What other executor than the NetworkExecutor would require a dispatcher ? 

> In this way we have one place where we can set dispatchers per ApolloClient instance.

I was thinking that setting it on the `NetworkTransport` would make it more explicit to callers where the dispatcher is used. Obviously that doesn't work if other executors need to access it.

sav007: After working on network layer I think we can drop passing any executor in the execution context for now. We can leave it to the client to decide, for example:
```
apolloClient
  .query(repositoriesQuery)
  .execute()
  .flowOn(Dispatchers.IO) // up to the client to provide this
  .single()
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ExecutionContext.kt
```diff
@@ -8,46 +8,110 @@ import kotlin.jvm.JvmField
  * A context of GraphQL operation execution, represented as a set of [Key] keys and corresponding [Element] values.
  */
 @ApolloExperimental
-class ExecutionContext private constructor(private val context: Map<Key<*>, Element>) {
+interface ExecutionContext {
```
sav007, line 11: will be rebased after merge this PR: https://github.com/apollographql/apollo-android/pull/2258

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/executor/NetworkExecutor.kt
martinbonnin: Even better :+1:

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.apollographql.apollo.api.Query
+import com.apollographql.apollo.api.ScalarTypeAdapters
+import com.apollographql.apollo.executor.NetworkExecutor
+import com.apollographql.apollo.executor.RequestExecutor
+import com.apollographql.apollo.internal.RealApolloCall
+import com.apollographql.apollo.network.NetworkTransport
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+
+@ApolloExperimental
+@ExperimentalCoroutinesApi
```
martinbonnin, line 16: Some users might not want to enable `ExperimentalCoroutinesApi` in their repo. As long as it's only for multiplatform I guess it's fine but the day `apollo-runtime-kotlin` becomes the default JVM implementation, it might be an issue (if the api hasn't been promoted to stable before that)

samples/multiplatform/kmp-lib-sample/src/commonMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloCoroutinesRepository.kt
```diff
@@ -1,26 +1,39 @@
...
  */
-class ApolloCoroutinesRepository(private val service: ApolloCoroutinesService) {
+class ApolloCoroutinesRepository {
+  private val apolloClient = ApolloClient(
+      networkTransport = ApolloHttpNetworkTransport(
+          serverUrl = "https://api.github.com/graphql",
+          httpHeaders = mapOf(
+              "Accept" to "application/json",
+              "Content-Type" to "application/json",
+              "Authorization" to "bearer $GITHUB_KEY"
```
martinbonnin, line 25: A common use case is to have an interceptor that can refresh the token. Should we add a concept of interceptors there as well ?

samples/multiplatform/kmp-lib-sample/src/iosMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloNetworkClient.kt
```diff
@@ -1,138 +0,0 @@
...
-import platform.Foundation.setHTTPBody
-import platform.Foundation.setHTTPMethod
-import platform.Foundation.setValue
-import platform.darwin.dispatch_async_f
-import platform.darwin.dispatch_get_main_queue
-import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.native.concurrent.freeze
-
-internal class ApolloNetworkClient(
```
martinbonnin, line 38: :sunglasses:

samples/multiplatform/kmp-lib-sample/src/commonMain/kotlin/com/apollographql/apollo/kmpsample/data/ApolloCoroutinesRepository.kt
sav007: Sure, we can. I guess we can do this in next PR. As we have to make abstraction that works for both OkHttp and NSURLSession.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
sav007: Well, flow operators are still marked with `flatMapLatest`, so no sure what we do with this?

martinbonnin: I guess it means keeping the JVM runtime around long enough for these flow operators to become stable?

sav007: yeah, we should keep JVM for some time.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
```diff
@@ -0,0 +1,12 @@
...
+import com.apollographql.apollo.api.Response
+import kotlinx.coroutines.flow.Flow
+
+interface ApolloCall<T> {
+  fun execute(): Flow<Response<T>>
+}
+
+interface ApolloQueryCall<T> : ApolloCall<T>
+
+interface ApolloMutationCall<T> : ApolloCall<T>
```
designatednerd, line 12: I have two somewhat contradictory questions: 
- Why are we using multiple interfaces here if they're not adding any functionality?
- Why isn't there an `ApolloSubscriptionCall<T>`?

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  @Suppress("NAME_SHADOWING")
+  private fun Result.dispatchOnMain(producerRef: COpaquePointer) {
+    val producerWithResultRef = StableRef.create((producerRef to this).freeze())
+    dispatch_async_f(
+        queue = dispatch_get_main_queue(),
+        context = producerWithResultRef.asCPointer(),
+        work = staticCFunction { ref ->
+          val producerWithResultRef = ref!!.asStableRef<Pair<COpaquePointer, Result>>()
```
designatednerd, line 180: is it such a rare occurrence that `ref` would be `null` that the force-unwrap is appropriate here?

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
```diff
@@ -0,0 +1,208 @@
...
+      appendBytes(it.addressOf(0), size.convert())
+    }
+  }
+
+  @Suppress("NAME_SHADOWING")
+  private fun Result.dispatchOnMain(producerRef: COpaquePointer) {
+    val producerWithResultRef = StableRef.create((producerRef to this).freeze())
+    dispatch_async_f(
+        queue = dispatch_get_main_queue(),
+        context = producerWithResultRef.asCPointer(),
```
designatednerd, line 178: `context` and `queue` here are properties of `NetworkTransport`, right? I *think* you're OK in terms of automatic reference counting here, but it might be worth doing some profiling to make sure you're not creating retain cycles.

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
```diff
@@ -0,0 +1,208 @@
...
+          val (producerPointer, result) = producerWithResultRef.get()
+          producerWithResultRef.dispose()
+
+          val producerRef = producerPointer.asStableRef<ProducerScope<GraphQLResponse>>()
+          val producer = producerRef.get()
+          producerRef.dispose()
+
+          when (result) {
+            is Result.Success -> {
+              producer.offer(
```
designatednerd, line 190: Again if `producer` is a property of `NetworkTransport`, it's worth looking into what happens if the instance gets deallocated before this gets called. This is where you see `[weak self]` sprinkled liberally around Swift code, I'm not entirely sure what the equivalent is with coroutines and flows.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
sav007: 1. it's for the future like normalized cache, any API for normalized cache will be defined as an extension functions, but we need to have 2 separate interfaces as only for cache is supported for `ApolloQueryCall`

2. `ApolloSubscriptionCall`  will be added when we start working on subscription KN version.

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
sav007: `staticCFunction` signature defines argument as nullable, but we pass `context` as non nullable value, `ref` should never be `null` that's why force unwrap

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
sav007: >context and queue here are properties of NetworkTransport, right

not sure I follow, `queue` and `context ` are named arguments for `dispatch_async_f ` C function. 
`dispatch_get_main_queue()` - gives access to main queue
`producerWithResultRef.asCPointer()` - just a pointer

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
sav007: We are safe here as `producer ` is not part of `NetworkTransport`, I mean `NetworkTransport ` doesn't hold reference.

It's a flow that holds reference, that returned to the user when he calls `ApolloClient.query`. KN doesn't have such concept as weak ref, but even if we keep reference for the producer, it's ok flow cancelation should take care of cleaning up.

And any `collecting` is tied to the coroutine scope, so if user doesn't leak scope and proper cancel flow we are safe here.

The only I guess improvements we missing here is to check if producer still alive before offer:
```
producer.isActive
producer.offer(...
```

martinbonnin: > The only I guess improvements we missing here is to check if producer still alive before offer

Isn't this the case already a few lines above ? (https://github.com/apollographql/apollo-android/blob/d464cf4fb0450f1f39716e2c4a41730dcf26f49b/apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt#L204)

All in all, I think we should wrap all `.offer()` with `runCaching {}`. We've been hit by  https://github.com/Kotlin/kotlinx.coroutines/issues/974 before.

sav007: >Isn't this the case already a few lines above 

I've just added this right away.

>All in all, I think we should wrap all .offer() with runCaching {}

Right, I was thinking that even checking for isActive won't enough as it could be RC. Will wrap into runCatching

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
designatednerd: I suspect people will want mutation results to update the cache as well, but that makes sense.

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/ApolloHttpNetworkTransport.kt
designatednerd: ah ok-  I wasn't quite clear on the fact that they were named arguments for the function rather than vars declared elsewhere. Thanks!

designatednerd:
CI seems to be unhappy with some changes affecting `kmp-android-app`

RoryKelly:
I can't seem to compile the example when isolated from the project. Is the kotlin runtime available on jCenter? Gradle cant resolve it. The kotlin multiplatform docs link to the example but I can't get it to compile.

martinbonnin:
@RoryKelly it's not released yet but you shuold be able to use the snapshots.
See https://github.com/apollographql/apollo-android#releases for more details.

RoryKelly:
@martinbonnin works great! it looks like there's no normalized cache (yet). Is this correct should I be following #2222 for updates?

martinbonnin:
@RoryKelly yup, that's a great ticket to follow. There are a lot of moving parts at the moment. If you have ideas how to structure the cache or features you'd like to have, feel free to comment down there.

Issue #2243, opened 2020-05-04
generateApolloSources takes too long time to generate Kotlin model classes


developer--:
**Is your feature request related to a problem? Please describe.**

My problem is that when I'm generating Kotlin model classes from graphql schema.json (~5MB) it takes about 12 minutes from my machine 
processor: i9
RAM: 16GB
os: macOS Catalina 10.14.5

But if I ran this generation from windows pc RAM 8GB, i5, without ssd it takes ~1 hour.

I'm interested what can be done to improve apollo class generation build speed, cause when my teammate builds it for iOS it takes less than a  minute ( schema.json file is same ).

`apollo version 2.0.1`

martinbonnin:
On my 2018 Core i5 8GB MacBook pro, generating the sources for https://github.com/martinbonnin/github-graphql-sample takes ~400ms (build scan there: https://scans.gradle.com/s/423mgtaloumce).
There are not a lot of queries but the schema is the 3.5MB github schema so that's already something. 12min sounds really high. Is there any chance you can share a build scan ?

martinbonnin:
Also maybe it's a specific query that takes some time. If you could share them, that'll help a lot. 

If not, can you bisect and remove queries progressively to see if the time spent depends linearly on the queries if some specific query take more time.

developer--:
I can ensure that removing some of graphql files decreases build time from 15m to 2-3 m, but I'm interested why this problem doesn’t have for iOS

martinbonnin:
Any chance you can share your schema ?

developer--:
Here is the schema.json
https://drive.google.com/file/d/1X6X2HTY0Z37EVh7Od-oGrse5GdN_nTXQ/view?usp=sharing

martinbonnin:
Thanks ! I started a small repo there to investigate the issue: https://github.com/martinbonnin/testCodegenPerf

So far it contains 2 small queries and it's working well on my laptop. Can you try on your machine ? And possibly add more queries to see which onees take a long time to generate ?

developer--:
Thanks, I will try, but also I have to tell you that this problem only acquired about a week ago, when some query files added, and this problem is not only for me, every teammate has same problem, but iOS team members doesn't have any problem with build time

developer--:
I'm testing your repo:
without added queries it took ~200 ms
Than I started to add queries one by one, I found one query, which took build 2m 52s to complete 
```

mutation createAddress($address:AddressInput!) {
[...]

PR #2257, opened 2020-05-08
Fix issue with GraphQL document parse perf.

sav007:
TL;DR: 
replaced recursion for traversing tree of a input types and their children with plain loop.

Closes https://github.com/apollographql/apollo-android/issues/2243

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
@@ -645,29 +645,48 @@ class GraphQLDocumentParser(val schema: Schema, private val packageNameProvider:
     }
   }
 
-  private fun Set<String>.usedSchemaTypes(exclude: Set<String> = emptySet()): Set<Schema.Type> {
+  private fun Set<String>.usedSchemaTypes(): Set<Schema.Type> {
     if (isEmpty()) {
       return emptySet()
     }
 
-    val usedSchemaTypes = filter { ScalarType.forName(it) == null }
+    val (scalarTypes, inputObjectTypes) = filter { ScalarType.forName(it) == null }
         .map { schema[it] ?: throw GraphQLParseException(message = "Undefined schema type `$it`") }
         .filter { type -> type.kind == Schema.Kind.SCALAR || type.kind == Schema.Kind.ENUM || type.kind == Schema.Kind.INPUT_OBJECT }
-        .toSet()
+        .partition { type -> type.kind == Schema.Kind.SCALAR || type.kind == Schema.Kind.ENUM }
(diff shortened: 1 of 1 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
```diff
@@ -645,29 +645,48 @@ class GraphQLDocumentParser(val schema: Schema, private val packageNameProvider:
...
-  }
+    val inputTypesToVisit = inputObjectTypes.toMutableList()
+    while (inputTypesToVisit.isNotEmpty()) {
+      val inputType = inputTypesToVisit.removeAt(inputTypesToVisit.lastIndex).also {
+        usedTypes.add(it)
+        visitedTypeNames.add(it.name)
+      }
+      val (nestedScalarTypes, nestedInputTypes) = inputType
+          .inputFields
+          .asSequence()
```
martinbonnin, line 673: Just curious: does this actually save time ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
```diff
@@ -645,29 +645,48 @@ class GraphQLDocumentParser(val schema: Schema, private val packageNameProvider:
...
+          .map { field -> field.type.rawType.name!! }
+          .filterNot { type -> visitedTypeNames.contains(type) }
+          .map { schema[it] ?: throw GraphQLParseException(message = "Undefined schema type `$it`") }
+          .filter { type -> type.kind == Schema.Kind.SCALAR || type.kind == Schema.Kind.ENUM || type.kind == Schema.Kind.INPUT_OBJECT }
+          .partition { type -> type.kind == Schema.Kind.SCALAR || type.kind == Schema.Kind.ENUM }
+          .let { (scalaTypes, inputTypes) ->
+            @Suppress("UNCHECKED_CAST")
+            scalaTypes.filter { ScalarType.forName(it.name) == null } to (inputTypes as List<Schema.Type.InputObject>)
```
martinbonnin, line 681:
```suggestion
          .let { (scalarTypes, inputTypes) ->
            @Suppress("UNCHECKED_CAST")
            scalarTypes.filter { ScalarType.forName(it.name) == null } to (inputTypes as List<Schema.Type.InputObject>)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
sav007: No, it's just a small optimization probably even not noticeable.

What really saves time is:
1) recursion replaced with loop
2) we were visiting the same input types over and over when traverse deep nested tree.

PR #2258, opened 2020-05-08
Make Execution context more Kotlin coroutine context idiomatic

sav007:
After working on new version of kotlin runtime, I realized why coroutine context `Element` confirms to the `ExecutionContext` protocol, it's much easier to work with this when you need to pass only one element as a context.

Refactor our `ExecutionContext` implementation to be more similar to coroutines ones.


apollo-runtime/src/main/java/com/apollographql/apollo/http/OkHttpExecutionContext.kt
```diff
@@ -0,0 +1,42 @@
...
+class OkHttpExecutionContext(
+    response: Response
+) : ExecutionContext.Element {
+  val response: Response = response.strip()
+
+  override val key: ExecutionContext.Key<*> = Key
+
+  companion object Key : ExecutionContext.Key<OkHttpExecutionContext> {
+    @JvmField
+    val KEY: Key = Key
```
sav007, line 20: for JVM users

PR #2260, opened 2020-05-11
Update codegen tests

martinbonnin:
Test now match the files both ways and not just verify that expected contents are generated.

They will now fail if the compiler generates a file that was not added as expected.

Added missing expected files for inline_fragment_type_coercion


sav007:
Some other potential improvements to code gen tests:
1. Generate diff when test fails to check the content
2. When Kotlin compilation fails for generated models it doesn't provide much info why and where

PR #2275, opened 2020-05-15
Apply the annotations to the Parameter/Field/Method instead of applying it to the Type

### LINKED ISSUE #2246: annotation type not applicable to this kind of declaration
**Summary**
While running ./gradlew generateApolloSources the generated files for one of my mutations seems to a type class incorrectly.  It builds with ./gradlew succesfully, but when Making Project it fails.
Build fails with error: annotation type not applicable to this kind of declaration
**Description**
I build with four queries.
```
mutation SignUpBarber($input: NewBarber!) {
    signUpBarber(input: $input) {
        response
        error
    }
}

query GetAllBarbers{
    getAllBarbers{
        barberID
        firstName
        lastName
    }
}

mutation RefreshToken($input: RefreshTokenInput!) {
    refreshToken(input: $input) {
        response
        error
    }
}

mutation LoginBarber($input: Login!) {
    login(input: $input) {
        response
        error
    }
}
```
The mutation that throws the error is LoginBarber.
RefreshToken mutationis quite similar, but it does not throw the error in question.

The error lies in this code 
```
 public LoginBarberMutation(com.snipsnap.android.barbershop.type. @NotNull Login input) {
    Utils.checkNotNull(input, "input == null");
    variables = new LoginBarberMutation.Variables(input);
  }
```
While the mutation for RefreshToken looks like this and produces no error
```
 public RefreshTokenMutation(@NotNull RefreshTokenInput input) {
    Utils.checkNotNull(input, "input == null");
    variables = new RefreshTokenMutation.Variables(input);
  }
```

public LoginBarberMutation(com.snipsnap.android.barbershop.type. @NotNull Login input)
throws the annotation type not applicable to this kind of declaration error.

**Version**
Provide a version of library on which the problem occurred.

Android Studio 3.6.3
classpath("com.apollographql.apollo:apollo-gradle-plugin:1.3.3")
implementation("com.apollographql.apollo:apollo-runtime:1.3.3")

Let me know if y'all need additional information. Ty in advanced!

### LINKED ISSUE #2265: Code Generation throws "error: method asList in class Arrays cannot be applied to given types"
**Summary**
When building the project and the codes gets generated the following two error came up

`error: method asList in class Arrays cannot be applied to given types;
    private Input<List<ProfileWhereUniqueInput>> receivers = Input.fromNullable(Arrays.<ProfileWhereUniqueInput>asList("null"));`

`error: incompatible types: Input<String> cannot be converted to Input<ConversationWhereUniqueInput>
    private Input<ConversationWhereUniqueInput> conversation = Input.fromNullable("null");`

**Description**
schema
```
input MessageInput {
    conversation: ConversationWhereUniqueInput
    receivers: [ProfileWhereUniqueInput!]
    text: String!
}

input ConversationWhereUniqueInput {
    id: String
}

input ProfileWhereUniqueInput {
    id: String
}
```

**Version**
2.0.3

martinbonnin:
closes #2246

JavaPoet introduced Java 8 type annotations in https://github.com/square/javapoet/pull/618 (versions 1.11.+). Apollo-android used to depend on javapoet 1.9.0 but if javapoet 1.11.+ is pulled by another dependency such as `androidx.navigation:navigation-safe-args-gradle-plugin`, the annotations will be properly generated as type annotations and fail to compile because `Nullable`, etc... cannot be applied to types.

This patch:
* adds a test case
* applies the annotation to the Parameter/Field/Method instead of the type when needed
* bumps javapoet to 1.12.1


apollo-compiler/src/test/graphql/com/example/input_object_object_nameclash/type/Login.java
```diff
@@ -0,0 +1,88 @@
...
+import com.apollographql.apollo.api.internal.InputFieldWriter;
+import com.apollographql.apollo.api.internal.Utils;
+import java.io.IOException;
+import java.lang.Object;
+import java.lang.Override;
+import java.lang.String;
+import org.jetbrains.annotations.NotNull;
+
+public final class Login implements InputType {
+  private final @NotNull String name;
```
sav007, line 19: Did we miss this case? Shouldn't it generate:
```
@NotNull
private final String name;
```
instead?

martinbonnin: It looks like the problem only happens right now because there can be `Objects` and `InputObjects` with the same name and javapoet has to pull the package name in front of the annotation to avoid a nameclash. 
Since this is potentially a big change to change annotations everywhere,  I tried to be conservative and made the minimal patch to make the test case pass but it's certainly better to use the same logic everywhere. I'll look into that.

martinbonnin:
Converted as a draft as I feel like this PR complexifies the Java codegen code instead of simplifying it. I'll try to revisit it.

martinbonnin:
Closing for now, will reopen later

Issue #2065, opened 2020-03-13
Custom code for enums / open enums


enriquedacostacambio:
**Is your feature request related to a problem? Please describe.**
I have some finite set of values for which graphql enums and java enums work perfectly, but in some instances i have enums that are finite but just for now, new values are likely to be added.
And in the consumer i rather keep the original value and not i don't want the caller to end up with "$UNKNOWN", specially when the value is going to be stored or returned without further inspection to an upstream service.

**Describe the solution you'd like**
One option is to declare them as Scalars, but that's not ideal, i the constants are still useful.

I'd like the ability to tag some schema enums with a directive and map those graphql enums not to java enums but to regular java classes.

i'd like the ability to turn something this:
```
@openEnum
enum UserType {
  CUSTOMER,
  EMPLOYEE
}
```
into something like this:
```
public class UserType {
  private static final CUSTOMER("CUSTOMER");

  private static final EMPLOYEE("EMPLOYEE");
  
  private final String name;

  UserType(String name) {
    this.name = name;
  }

  public String name() {
    return name;
  }

  public static UserType[] values() {
    return new UserType[] {
      CUSTOMER,
      EMPLOYEE
    }
  }
 
  public static UserType valueOf(String name) {
    for (UserType name : values()) {
      if (enumValue.name.equals(name)) {
        return enumValue;
      }
    }
    return new UserType(name);
  }
}
```

tasomaniac:
I think we should re-evaluate the suggested solution since that would require a lot of maintenance in Apollo Compiler.

But I agree that we also sometimes need to access the `rawValue` of the new enums provided. With the current approach, the values are just lost.

In your case, having them as scalars really makes sense since you are not going to process them and just pass them.

metalpalo:
Hi, 
sorry for joining here
I would like to if possible to disable puting $UNKNOWN("$UNKNOWN") from enum generally. 
We have generated enum classes (probably generated by older version) without unknown value and we would like to stay by this way.

thanks
brpalo

tasomaniac:
@metalpalo in that scenario how would you handle the situation where your backend adds a new value to the enum? How would already released old versions work in that case?

metalpalo:
Hi
Firstly – is this value $UNKNOWN("$UNKNOWN") required by apollo runtime itself in order to handle unknown enum values on properly. ?
We only read data a it would be just fine for us if we received null values in case of unknown values.
We read data from one backend e.g.
Enum1Type {
VALUE1
[...]

PR #2279, opened 2020-05-17
Feature to generate GraphQL enum types as sealed classes

sav007:
Update Kotlin compiler to generate GraphQL enum type as sealed class.

Closes https://github.com/apollographql/apollo-android/issues/2065

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
@@ -4,19 +4,32 @@ import com.apollographql.apollo.compiler.applyIf
 import com.apollographql.apollo.compiler.ast.EnumType
 import com.squareup.kotlinpoet.*
 
-internal fun EnumType.typeSpec(generateAsInternal: Boolean = false) =
-    TypeSpec
-        .enumBuilder(name)
-        .applyIf(description.isNotBlank()) { addKdoc("%L\n", description) }
-        .applyIf(generateAsInternal) { addModifiers(KModifier.INTERNAL) }
-        .primaryConstructor(primaryConstructorSpec)
-        .addProperty(rawValuePropertySpec)
-        .apply {
-          values.forEach { value -> addEnumConstant(value.constName, value.enumConstTypeSpec) }
-          addEnumConstant("UNKNOWN__", unknownEnumConstTypeSpec)
-        }
-        .addType(companionObjectSpec)
-        .build()
+internal fun EnumType.typeSpec(
+    generateAsInternal: Boolean = false,
+    enumAsSealedClassPatternFilters: List<Regex>
+): TypeSpec {
+  val asSealedClass = enumAsSealedClassPatternFilters.isNotEmpty() && enumAsSealedClassPatternFilters.any { pattern ->
+    name.matches(pattern)
+  }
+
+  return if (asSealedClass) toSealedClassTypeSpec(generateAsInternal)
+  else toEnumTypeSpec(generateAsInternal)
+}
+
+private fun EnumType.toEnumTypeSpec(generateAsInternal: Boolean): TypeSpec {
+  return TypeSpec
+      .enumBuilder(name)
+      .applyIf(description.isNotBlank()) { addKdoc("%L\n", description) }
+      .applyIf(generateAsInternal) { addModifiers(KModifier.INTERNAL) }
+      .primaryConstructor(primaryConstructorSpec)
+      .addProperty(rawValuePropertySpec)
+      .apply {
+        values.forEach { value -> addEnumConstant(value.constName, value.enumConstTypeSpec) }
+        addEnumConstant("UNKNOWN__", unknownEnumConstTypeSpec)
+      }
+      .addType(enumCompanionObjectSpec)
+      .build()
+}
 
 private val primaryConstructorSpec =
     FunSpec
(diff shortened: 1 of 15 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
```diff
@@ -154,4 +154,10 @@ interface CompilerParams {
    * Default value: false
    */
   val generateAsInternal: Property<Boolean>
+
+  /**
+   * Whether to generate GraphQL enum types by matching as Kotlin sealed classes if they match one of provided regex patterns.
+   * By default any GraphQL enum type is generated as `enum class`.
+   */
```
martinbonnin, line 161:
```suggestion
  /**
   * A list of [Regex] patterns for GraphQL enums that should be generated as Kotlin sealed classes instead
   * of the default Kotlin enums.
   * Use this if you want your client to have access to the rawValue of the enum. This can be useful if new
   * GraphQL enums are added but the client was compiled against an older schema that doesn't have knowledge
   * of the new enums.
   */
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
```diff
@@ -154,4 +154,10 @@ interface CompilerParams {
...
+  val generateEnumAsSealedClass: ListProperty<String>
```
martinbonnin, line 162: Nit: Maybe name this `enumAsSealedClassPatterns`? `generate` is used above for Boolean properties and making the property plural would also hint that it is a List.

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
```diff
@@ -96,13 +96,19 @@ class CodeGenTest(private val folder: File) {
...
         messageOutputStream = System.out // see diagnostics in real time
       }.compile()
 
-      assertThat(result.exitCode).isEqualTo(KotlinCompilation.ExitCode.OK)
+      if (result.exitCode != KotlinCompilation.ExitCode.OK) {
+        val compilationErrorMessages = "\\ne: .*\\n+".toRegex().find(result.messages)?.groupValues ?: emptyList()
+        val errorMessages = compilationErrorMessages.joinToString(prefix = "\n", separator = "\n", postfix = "\n") {
+          "`${it.replace("\n", "")}`"
+        }
+        fail("Failed to compile generated Kotlin files due to compiler errors: $errorMessages")
```
martinbonnin, line 110: Nice 😃

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
```diff
@@ -4,19 +4,32 @@ import com.apollographql.apollo.compiler.applyIf
...
-          values.forEach { value -> addEnumConstant(value.constName, value.enumConstTypeSpec) }
-          addEnumConstant("UNKNOWN__", unknownEnumConstTypeSpec)
-        }
-        .addType(companionObjectSpec)
-        .build()
+internal fun EnumType.typeSpec(
+    generateAsInternal: Boolean = false,
+    enumAsSealedClassPatternFilters: List<Regex>
+): TypeSpec {
+  val asSealedClass = enumAsSealedClassPatternFilters.isNotEmpty() && enumAsSealedClassPatternFilters.any { pattern ->
```
martinbonnin, line 11: Is the check for `isNotEmpty()` needed here ? I would expect `any` to return false on empty lists.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
sav007: Yeah, hard to come up with proper name my thought `enumAsSealedClassPatternFilters`   is ok for internal usage but was not great for public API. 
`generateSealedClassForEnums`, `sealedClassForEnums`, `enumsAsClasses`, `enumAsSealedClassPatterns`?

Or do you think `enumAsSealedClassPatternFilters` should be fine?

tasomaniac: I also think that the one used in the code internally is not super clear for external facing extension. We can make it plural easily to indicate that. Since they're `enum` in GraphQL convention, I like starting with `generateEnums`

What about

```
generateEnumsAsSealedClassesFor
```

or 

```
generateSealedClassesForEnums
```

martinbonnin: Maybe something that describes the end functionality instead of the codegen details ? 

```
exposeRawValueForUnknownEnums
```
or
```
exposeRawValueForEnumsMatching
```

sav007: but then it's not clear that the end result is going to be a sealed class generated instead of regular enum, I think it's important to have `sealedClass` in there.
1 - `enumsAsSealedClassesMatching`
2 - `enumsAsClassesMatching`
3 - `sealedClassesForEnums`
4 - `sealedClassesForEnumsMatching`

Another option if this make sense make it lambda:
```
apollo {
  enumsAsSealedClasses {
    matching = [".*"]
  }
}
```

martinbonnin: I like `sealedClassesForEnumsMatching`

tasomaniac: I definitely like that too. 👍

martinbonnin: We got a winner 🎉 😃

Issue #2280, opened 2020-05-18
Caching partial responses upon errors


podarsmarty:
**Summary**
Incorrect caching behavior with errors.

**Description**
I think this issue is super similar to https://github.com/apollographql/apollo-android/issues/2202 with the same root cause. I am utilizing CACHE_FIRST to hit my api with an aliased request in order to hit the same api twice with different request params. If one of these requests succeeds but the other one results in a 500 error both data and errors are populated correctly. However, this partial data response is cached as the total response with no indication of the error. If/when a retry is triggered we will now fetch the partial response from the cache.

I would expect the partial response to not be stored in the cache, or that this partial response due to an error is stored with the error and logic results in it being a cache miss. Not entirely convinced that caching an error response and returning it should be expected behavior.

**Version**
1.45

sav007:
So just to clarify, when you hit network you get partial data along with the GraphQL errors in the response, something similar to this:
```
{
  "data": {
     "query" : {
       .....
     }
  },
  "errors": [
    {
      "message": null,
      "locations": null,
      "code":"userNotFound",
      "path":"loginWithPassword"
    }
  ]
}
```
And you don't want to store partial data into the cache?

podarsmarty:
@sav007 yup that is correct

PR #2281, opened 2020-05-18
Prevent incorrect caching of partial responses with errors.

podarsmarty:
- Fixes https://github.com/apollographql/apollo-android/issues/2280 by preventing the response from being cached with errors are present in the response
- Added unit tests for prevention of caching response behavior
- Updated files so that ./gradlew check passes lint checks

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
@@ -123,6 +123,9 @@ public final class ApolloCacheInterceptor implements ApolloInterceptor {
 
   Set<String> cacheResponse(final InterceptorResponse networkResponse,
       final InterceptorRequest request) {
+    if (networkResponse.parsedResponse.isPresent() && networkResponse.parsedResponse.get().hasErrors()) {
+      return Collections.emptySet();
+    }
     final Optional<List<Record>> records = networkResponse.cacheRecords.map(
         new Function<Collection<Record>, List<Record>>() {
           @NotNull @Override public List<Record> apply(@NotNull Collection<Record> records) {
--- a/apollo-runtime/src/test/java/com/apollographql/apollo/ApolloExceptionTest.java
+++ b/apollo-runtime/src/test/java/com/apollographql/apollo/ApolloExceptionTest.java
@@ -31,7 +31,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import static com.google.common.truth.Truth.assertThat;
 
 @SuppressWarnings("unchecked") public class ApolloExceptionTest {
-  private static long TIMEOUT_SECONDS = 2;
+  private static long timeoutSeconds = 2;
 
   @Rule public final MockWebServer server = new MockWebServer();
   private ApolloClient apolloClient;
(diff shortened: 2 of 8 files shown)


sav007:
I would recommend to add new cache header to `ApolloCacheHeaders` that will signal to no store partial response data.

By default we keep existing behaviour to avoid breaking changes.

Does this make sense?

podarsmarty:
@sav007 hmm im not sure why you would want to keep the existing behavior. we are incorrectly caching bad responses. if we were to cache the actual errors along with the data it would make more sense but then even then im not sure you want to cache errors.

sav007:
Yeah, my bad you are right. We shouldn't cache partial response data at first place.

apollo-runtime/src/test/java/com/apollographql/apollo/ApolloExceptionTest.java
```diff
@@ -31,7 +31,7 @@
 import static com.google.common.truth.Truth.assertThat;
 
 @SuppressWarnings("unchecked") public class ApolloExceptionTest {
-  private static long TIMEOUT_SECONDS = 2;
+  private static long timeoutSeconds = 2;
```
sav007, line 34: is there a reason this file updated?

sav007:
Could you please revert changes that not related to this PR

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
```diff
@@ -135,7 +135,7 @@ InterceptorResponse resolveFromCache(InterceptorRequest request) throws ApolloEx
         }
     );
 
-    if (!records.isPresent()) {
+    if (!records.isPresent() || (networkResponse.parsedResponse.isPresent() && networkResponse.parsedResponse.get().hasErrors())) {
```
sav007, line 138: I think we can move this to the top of `cacheResponse` function
```
    if (networkResponse.parsedResponse.isPresent() && networkResponse.parsedResponse.get().hasErrors()) {
      return Collections.emptySet();
    }
    final Optional<List<Record>> records = networkResponse.cacheRecords.map(
    ....
```

martinbonnin:
Not caching data that contains errors looks good to me, thanks 👍! 

Regarding the style changes, we disabled the `checkstyle` task in CI 🙈: https://github.com/apollographql/apollo-android/blob/6504f05a16a9c090fa1b98eacf68e09d64752203/scripts/ci.sh#L8

```
./gradlew clean build -x checkstyleTest --stacktrace --max-workers=2
```

We might want to fine tune this. 
@podarsmarty can you open a separate PR for this discussion?

podarsmarty:
@sav007 @martinbonnin ahh yea all the unrelated changes were just things to make the lint checks pass. if it has been updated in CI should be safe. i will revert all of those changes

apollo-runtime/src/test/java/com/apollographql/apollo/ApolloExceptionTest.java
podarsmarty: had to make unrelated changes to make lint checks pass. sounds like they have been updated in CI though so i will revert all the unrelated changes

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
podarsmarty: hmm im down with short circuiting. will update

tasomaniac:
I also agree that this behavior change is fine since it is considered as bugs. But it might accidentally affect people depending on this. We should make it clear on the changelog.

tasomaniac:
I'm actually ok with unrelated changes since this has been an issue with other contributors that running checkstyle would fail locally on these files. It is good to fix these small warnings.

podarsmarty:
@sav007 @martinbonnin updated with the short circuit to emptySet and based on @tasomaniac's comment left in the lint fixes

PR #2282, opened 2020-05-18
make master 2.1.0-SNAPSHOT

martinbonnin:
master actually contains new features in new features in `apollo-runtime-kotlin` so we could do a `2.1.0` but these features are really early and not really ready for prime time so I think 2.0.4 makes more sense.

**Edit**: it will be a 2.1.0 in the end

2.1.0 will have: 
* okhttp 3.12
* performance improvements
* multiplatform `apollo-normalized-cache-sqlite`
* sealedClassesForEnumsMatching
* enforced same version for all modules, including rootProject

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 GROUP=com.apollographql.apollo
-VERSION_NAME=2.1.0-SNAPSHOT
+VERSION_NAME=2.0.4-SNAPSHOT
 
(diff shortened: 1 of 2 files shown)


martinbonnin:
I forgot we also had `sealedClassesForEnumsMatching ` in the new version too so `2.1.0` might make more sense after all.

sav007:
I think we should make `2.1.0` release

PR #2284, opened 2020-05-19
Introduce Http execution context in kotlin runtime module

sav007:
- Provide API to set Http headers per request via `HttpExecutionContext.Request`
- Provide access to raw Http execution response status code and headers via `HttpExecutionContext.Response`


martinbonnin:
Do we really need `executionContext` in `ApolloException` ? I understand one of the big advantage of the `ExecutionContext` API is that it allows each interceptor in the chain to contribute to the context. 
My understanding is that any `ApolloException` will terminate the chain so it will come from a single interceptor and the caller could then check based on the instance of the exception where it came from?

```
when (apolloException) {
     is ApolloNetworkException -> {}
     is ApolloTransportException -> {}
     is ApolloParseException -> {}
     ...
}
```

Or is there a use case where the exception should contain several keys in its executionContext ?

sav007:
The idea of having `executionContext` in `ApolloException` is to provide additional context what went wrong. For example right now we put there raw http status code & headers. But I agree that for now we should introduce appropriate error like `com.apollographql.apollo.ApolloError.Http` with `statusCode` and `headers` fields instead.

martinbonnin:
What about subclassing `ApolloException` instead of using `apolloException.error` ? That would allow to put the additional fields directly in the root of the `Exception` ?

```
class ApolloHttpException(val message: String,
                                              val statusCode: String,
                                              val headers: Map<String, String>
                                              ): ApolloException 
```

martinbonnin:
@sav007 as an exercise, I added a Oauth interceptor there: https://github.com/sav007/apollo-android/pull/7/files

sav007:
@martinbonnin 
>What about subclassing ApolloException instead of using apolloException.error ? That would allow to put the additional fields directly in the root of the Exception ?

The idea was to have an algebraic type for error type, so make sure user handles them all by pattern matching, also it helps with discoverability. Current version of subclassing `ApolloException` in JVM runtime doesn't really help to discover all possible errors user should expect.

With sealed class that models error user can easily do:
```
when(error) {
 is ApolloError.SerializationError -> 
 is ApolloError.ParseError ->
 else -> 
```

Also having only one  `ApolloException` makes API cleaner, in terms of `try / catch`, user will expect only one exception. If there is a need of fine-grained error handling, it can be done via pattern matching over sealed class `ApolloError`

martinbonnin:
> Also having only one ApolloException makes API cleaner, in terms of try / catch, user will expect only one exception. If there is a need of fine-grained error handling, it can be done via pattern matching over sealed class ApolloError.

We could make ApolloException a sealed class itself:

```
@ApolloExperimental
sealed class ApolloException(
    message: String,
    cause: Throwable? = null
) : RuntimeException(message, cause)

class ApolloHttpException(
    message: String,
    cause: Throwable? = null,
    val statusCode: Int,
    val headers: Map<String, String>
    ): ApolloException(message, cause)

class ApolloParseException(
    message: String,
    cause: Throwable? = null,
    val line: Int,
    val column: Int
): ApolloException(message, cause) 
```

Users could still catch all the exceptions in one `try/catch` block by catching the base `ApolloExceptionClass`. And if they want something more fine grained, they can do it too:

```kotlin
try {
    apolloClient.query(query).execute().collect {}
} catch(e: ApolloHttpException) {
    // do something specific for Http exceptions
} catch(e: ApolloException) {
    // generic error case
}
```

It *flattens* a bit the data structure but would be functionally similar ?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloError.kt
```diff
@@ -2,6 +2,10 @@ package com.apollographql.apollo
 
 sealed class ApolloError {
   object SerializationError : ApolloError()
+
   object ParseError : ApolloError()
+
   object Network : ApolloError()
+
+  data class Http(val statusCode: Int, val headers: Map<String, String>) : ApolloError()
```
tasomaniac, line 10: Should this be renamed to `HttpError` instead?

PR #2294, opened 2020-05-21
[Gradle Plugin] Use relative strings for the rootFolder input property so as not to break cacheability

### LINKED ISSUE #2293: Use of absolute string path breaks generateSources task remote cacheability
**Summary**

Input property [rootFolders](https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt#L84) is referenced as an [absolute path string ](https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloPlugin.kt#L129).

If you use a gradle remote cache absolute paths may differ based on the machine used for the build.

**Description**
I believe we could change this to a `ConfigurableFileCollection` similar to [graphqlFiles](https://github.com/apollographql/apollo-android/blob/master/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloGenerateSourcesTask.kt#L78) fix the issue

**Version**
1.4.5, but pretty sure this is still on master.

martinbonnin:
Closes #2293

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/DefaultPackageNameProvider.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/DefaultPackageNameProvider.kt
@@ -2,8 +2,8 @@ package com.apollographql.apollo.compiler
 
 import java.io.File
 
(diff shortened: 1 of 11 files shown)


tasomaniac:
Should we merge this too?

PR #2298, opened 2020-05-22
[New Runtime] Add BearerTokenInterceptor

martinbonnin:
Continuation of https://github.com/sav007/apollo-android/pull/7

A BearerTokenInterceptor can be added to the list of interceptors to provide/refresh an accessToken.


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/BearerTokenInterceptor.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlinx.coroutines.sync.withLock
+
+@ApolloExperimental
+class BearerTokenInterceptor(private val tokenProvider: TokenProvider) : ApolloRequestInterceptor {
+  val mutex = Mutex()
+
+  private fun <T> ApolloRequest<T>.withHeader(name: String, value: String): ApolloRequest<T> {
+    val httpRequestContext = executionContext[HttpExecutionContext.Request]
+        ?: HttpExecutionContext.Request(emptyMap())
+            .let {
```
tasomaniac, line 24: I feel like this `let` only add to the default branch with `emptyMap`. It should add to the existing context, shouldn't it?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloException.kt
```diff
@@ -14,3 +14,5 @@ class ApolloHttpException(
     message: String,
     cause: Throwable? = null
 ) : ApolloException(message = message, cause = cause)
+
+class BearerTokenException(message: String, cause: Throwable? = null, val token: String): ApolloException(message = message, cause = cause)
```
tasomaniac, line 18: Would it make sense to somehow extend `ApolloHttpException` here since this is essentially also an `HttpException`

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/TokenProvider.kt
```diff
@@ -0,0 +1,6 @@
+package com.apollographql.apollo.interceptor
+
+interface TokenProvider {
+  suspend fun currentToken(): String
+  suspend fun renewToken(previousToken: String): String
```
tasomaniac, line 5: I wonder how much we should follow OAuth standards here? I was thinking about refreshToken for example. That is usually used for getting a new token in such scenario.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/BearerTokenInterceptor.kt
martinbonnin: Oh my. Good catch, thanks !

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloException.kt
martinbonnin: The reasoning was to keep the inheritance tree relatively flat. Also now that ApolloException is a sealed class, inheriting from sealed class wouldn't be working.
If the user wants statusCode and headers, they are still readable in `BearerTokenException.cause`

PS: I prefixed with `Apollo`: `ApolloBearerTokenException` to make it consistent with the other classes above.

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/TokenProvider.kt
martinbonnin: Renamed. Also, that aligns perfectly with `currentToken` 👌

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloException.kt
tasomaniac: Sounds good. Maybe the type of `cause` can be the concrete type rather than Throwable.

tasomaniac: But tbh, I'm not sure if that's even necessary since this is not going to be used directly, right?

PR #2302, opened 2020-05-25
[Runtime] add ApolloClient.applicationInterceptorFactories()

### LINKED ISSUE #2285: Application interceptor only created once
I created a custom interceptor like this: 

```kotlin
class OperationRegistryInterceptor(private val logger: ApolloLogger) : ApolloInterceptor {
    @Volatile
    private var disposed = false

    override fun interceptAsync(
        originalRequest: InterceptorRequest,
        dispatcher: Executor,
        callBack: ApolloInterceptor.CallBack
    ) {
        ...
        chain.proceedAsync(newRequest, dispatcher, object : ApolloInterceptor.CallBack {
            override fun onResponse(response: InterceptorResponse) {
                if (disposed) return
                callBack.onResponse(response)
                callBack.onCompleted()
            }
       ...
        })
    }

    override fun dispose() {
        disposed = true
    }
}
```

Then I created this interceptor when creating ApolloClient:
We have one ApolloClient for the whole app. 

```kotlin
fun apolloClient(): ApolloClient {
        return with(ApolloClient.builder()) {
            addApplicationInterceptor(OperationRegistryInterceptor(logger))
            ...
            build()
}
```

We have a use case that we send requests when user drags a map. When user moves the map to a new location, we need to cancel the old request and issue a new request with new location data. Thus we need to call `dispose()`manually in the code. 

For this OperationRegistryInterceptor, I found that `disposed` is set to be true for the first request. Then the following requests do not come back with any data because there seems to be no way that this is reset to be false. 

I checked the build-in interceptors such as ApolloServerInterceptor. All of them are created as new objects in prepareInterceptorChain() of RealApolloCall thus when a new request is issued, disposed is set to be false.

I think the ideal way to solve this is to be able to create a new object of OperationRegistryInterceptor when there is a new request, the same way as the built-in interceptors. I did not see a way to do this cleanly.

Can the ApplicationInterceptor behave the same as build-in interceptors? Basically a new object is created per request.

martinbonnin:
closes #2285

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -22,6 +22,7 @@ import com.apollographql.apollo.exception.ApolloParseException;
 import com.apollographql.apollo.fetcher.ResponseFetcher;
 import com.apollographql.apollo.interceptor.ApolloInterceptor;
 import com.apollographql.apollo.interceptor.ApolloInterceptorChain;
+import com.apollographql.apollo.interceptor.ApolloInterceptorFactory;
 import com.apollographql.apollo.internal.interceptor.ApolloAutoPersistedQueryInterceptor;
 import com.apollographql.apollo.internal.interceptor.ApolloCacheInterceptor;
 import com.apollographql.apollo.internal.interceptor.ApolloParseInterceptor;
@@ -65,6 +66,7 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
   final ApolloLogger logger;
   final ApolloCallTracker tracker;
   final List<ApolloInterceptor> applicationInterceptors;
+  final List<ApolloInterceptorFactory> applicationInterceptorFactories;
   final List<OperationName> refetchQueryNames;
   final List<Query> refetchQueries;
   final Optional<QueryReFetcher> queryReFetcher;
@@ -94,6 +96,7 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
     dispatcher = builder.dispatcher;
     logger = builder.logger;
     applicationInterceptors = builder.applicationInterceptors;
+    applicationInterceptorFactories = builder.applicationInterceptorFactories;
     refetchQueryNames = builder.refetchQueryNames;
     refetchQueries = builder.refetchQueries;
     tracker = builder.tracker;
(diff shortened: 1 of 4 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -375,7 +380,18 @@ private ApolloInterceptorChain prepareInterceptorChain(Operation operation) {
     HttpCachePolicy.Policy httpCachePolicy = operation instanceof Query ? this.httpCachePolicy : null;
     ResponseFieldMapper responseFieldMapper = responseFieldMapperFactory.create(operation);
 
-    List<ApolloInterceptor> interceptors = new ArrayList<>(applicationInterceptors);
+    List<ApolloInterceptor> interceptors = new ArrayList<>();
+
+    if (!applicationInterceptorFactories.isEmpty() && !applicationInterceptors.isEmpty()) {
+      throw new IllegalArgumentException("You can either use applicationInterceptors or applicationInterceptorFactories "
```
tasomaniac, line 386: I wonder if this check is something we should do earlier in `ApolloClient.Builder.build()` method.

tasomaniac:
This essentially makes the public API more complex with a little additional benefit. I wonder why we cannot really do this with the existing Interceptor API. What is the limitation there?

martinbonnin:
> This essentially makes the public API more complex with a little additional benefit. I wonder why we cannot really do this with the existing Interceptor API. What is the limitation there?

In the current API, the interceptors are only created once. Any interceptor that needs to have a lifecycle bound to the request will not work.

tasomaniac:
I'm comparing them with what OkHttp has in my mind. They also create them once. I really don't know in detail why do we have lifecycle concerns in there. That's why I don't understand completely but I feel like it would be simpler otherwise.

In OkHttp, are they throwing an exception when an operation is cancelled?

martinbonnin:
I *think* OkHttp is a mix of:
1. Closing the socket synchronously
2. Interceptors, etc... check `call.isCancelled()` to cancel cooperatively

In ApolloInterceptor, there is a `dispose` method but since the application interceptors are client-bound and not request-bound, application interceptors do not have a chance to start with a clean their state.

martinbonnin:
>  Please have a look at the other comment I had.

Sorry I missed it. It should be better now.

> Also, is this something we want to bring to the new runtime too? I saw that the new runtime interceptor does not have dispose callback.

I would say we should use coroutines cancellation for this in the new api? I don't think we should port this.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
martinbonnin: Yep, definitely. I put it in the `ApolloClient` constructor. I don't think anyone uses that directly but better safe than sorry.

PR #2304, opened 2020-05-26
Rx3 support

Jawnnypoo:
Just like the Rx2 support, but with the RxJava 3 imports in place.


apollo-rx3-support/src/main/java/com/apollographql/apollo/rx3/Rx3Apollo.java
```diff
@@ -0,0 +1,250 @@
...
+import io.reactivex.rxjava3.core.SingleOnSubscribe;
+import io.reactivex.rxjava3.disposables.Disposable;
+import io.reactivex.rxjava3.exceptions.Exceptions;
+import static com.apollographql.apollo.api.internal.Utils.checkNotNull;
+
+/**
+ * The Rx3Apollo class provides methods for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to RxJava 3
+ * sources.
+ */
+public class Rx3Apollo {
```
tasomaniac, line 35: Now that we have Kotlin in the repository, I think it would be much better to have this implemented fully in Kotlin.

That means, maybe we should merge it with RxJavaExtensions below and have everything as top-level functions.

Let me know what you think.

Jawnnypoo: Totally agree! I would love to follow up with another PR doing this - I think adding that to this PR would be a little bit out of scope since it would be done for both Rx2 and Rx3 support.

Jawnnypoo:
Integration tests added 👍

apollo-rx3-support/src/main/java/com/apollographql/apollo/rx3/Rx3Apollo.java
martinbonnin: > That means, maybe we should merge it with RxJavaExtensions below and have everything as top-level functions.

Does that mean RxJava3 support will not be usable for Java callers ?

Jawnnypoo: Ah no, the changes could be made in a way that makes it fully compatible and callable from Java.

tasomaniac: Yep, that can be done via JvmStatic and JvmName annotations

Jawnnypoo: @tasomaniac Would you prefer me to change this PR to include those changes as well? Or can I do a follow up PR?

tasomaniac: If you want to do a follow-up that would be awesome. Let's merge this then.

laomo:
Exactly what i need, i want to know when to release?

tasomaniac:
Hey, it will be in the next release. I'm not sure when would that be exactly

PR #2307, opened 2020-05-26
Build pushes only to master

tasomaniac:
With every pull request we are building both the branch and the PR. Building the branches is just redundant.

This PR changes so that they run only on `master` branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: CI
 
 on:
   push:
+    branches: [ master ]
   pull_request:
     paths-ignore:
       - 'docs/**'


martinbonnin:
Does that scenario work?

1. I open a pull request to merge `featureBranch` into `master`
2. The CI will receive a `pull_request` event
3. I push more commits to `featureBranch`
4. Will the CI receive another `pull_request` event ? 

I would be tempted to say yes or the checks wouldn't be valid but I didn't find a good documentation of when `pull_request` is triggered, i.e. at pull request creation or every time it is updated.

tasomaniac:
It is definitely every time it is updated. I can actually try it with this PR already.

Specifically on this PR, push checks did not run. Pushing a commit should trigger a build.

Also builds from forks never trigger the "push" event.

tasomaniac:
It is definitely every time it is updated. I can actually try it with this PR already.

Specifically on this PR, push checks did not run. Pushing a commit should trigger a build.

Also builds from forks never trigger the "push" event.

tasomaniac:
Yep, triggered

PR #2311, opened 2020-05-27
Provide base websocket abstraction

sav007:
Introduce websocket abstraction targeting both Android and iOS for later use in `ApolloWebSocketNetworkTransport`.

Use `OkHttpClient` and `NSURLSession` for Android and iOS targets accordingly as web socket implementation under the hood.

Provide `ApolloWebSocketNetworkTransport` implementation responsible for open and share web socket connection between query requests, establish and configure GraphQL web socket connection.

### DIFF
--- a/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/ApolloWebSocketNetworkTransport.kt
+++ b/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/ApolloWebSocketNetworkTransport.kt
@@ -11,95 +11,80 @@ import com.apollographql.apollo.network.GraphQLResponse
 import com.apollographql.apollo.network.NetworkTransport
 import com.apollographql.apollo.network.websocket.GraphQLServerMessage.Companion.parse
 import com.benasher44.uuid.Uuid
-import com.benasher44.uuid.uuid4
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.cancel
+import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.channels.BroadcastChannel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.broadcast
-import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.filter
-import kotlinx.coroutines.flow.filterIsInstance
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.firstOrNull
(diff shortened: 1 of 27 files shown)


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/ApolloWebSocketNetworkTransport.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.apollographql.apollo.api.Response
+import com.apollographql.apollo.interceptor.ApolloInterceptorChain
+import com.apollographql.apollo.interceptor.ApolloRequest
+import com.apollographql.apollo.interceptor.ApolloRequestInterceptor
+import kotlinx.coroutines.flow.Flow
+
+@ApolloExperimental
+class ApolloWebSocketNetworkTransport(
+    private val apolloWebSocketFactory: ApolloWebSocketFactory
+) : ApolloRequestInterceptor {
```
martinbonnin, line 13: Feels weird to name it `NetworkTransport` if it inherits from `Interceptor`

sav007: Sorry it was a mistake, it supposed to inherit `NetworkTransport`

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/websocket/ApolloWebSocket.kt
```diff
@@ -0,0 +1,160 @@
...
+import platform.Foundation.NSURLSessionWebSocketMessage
+import platform.Foundation.NSURLSessionWebSocketMessageTypeData
+import platform.Foundation.NSURLSessionWebSocketMessageTypeString
+import platform.Foundation.NSURLSessionWebSocketTask
+import platform.Foundation.setValue
+import kotlin.native.concurrent.AtomicReference
+
+typealias WebSocketFactory = (NSURLRequest) -> NSURLSessionWebSocketTask
+
+actual class ApolloWebSocketFactory(
```
designatednerd, line 29: This is a big one: We need to make it clear that this will only work with iOS 13+ - I don't think this will even compile on older versions of iOS since it leverages web socket stuff introduced in iOS 13.

If we're not making iOS 13 a requirement for KMP (which we could, but I think might be a smidge premature), then there needs to be a way to prevent this from compiling or being available on iOS 12 and below.

sav007: I was thinking on adding this as part of documentation later as we don't really promote new kotlin runtime module.

It's interesting is there any way to put restriction on iOS version in KN configuration? We should check that.

designatednerd: I have to imagine there's a way to do it somehow - Or I wonder if maybe the K/N shims are doing the work of no-op'ing new APIs on older versions under the hood? I haven't looked into it tbh.

PR #2321, opened 2020-06-01
[Plugin] simplify the downloadApolloSchema task

martinbonnin:
Looking at the documentation, I found this Gradle [@Option annotation](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/options/Option.html) that allows to pass task inputs directly from the commandline. That will simplify downloading the schema:

**before:**

```
./gradlew downloadApolloSchema -Pcom.apollographql.apollo.endpoint="https://..." -Pcom.apollographql.apollo.schema="src/main/graphql/..."
```

**after:**

```
./gradlew downloadApolloSchema --endpoint="https://..." --schema="src/main/graphql/..."
```

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
@@ -1,36 +1,32 @@
 package com.apollographql.apollo.gradle.internal
 
-import com.squareup.moshi.JsonWriter
-import okhttp3.HttpUrl.Companion.toHttpUrl
-import okhttp3.MediaType.Companion.toMediaTypeOrNull
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okhttp3.RequestBody.Companion.toRequestBody
-import okio.buffer
-import okio.sink
 import org.gradle.api.DefaultTask
-import org.gradle.api.provider.MapProperty
+import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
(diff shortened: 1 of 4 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
```diff
@@ -1,36 +1,32 @@
...
   @get:Input
-  abstract val headers: MapProperty<String, String>
+  @set:Option(option = "header", description = "headers in the form 'Name: Value'")
+  var header = emptyList<String>() // cannot be lazy for @Option to work
 
-  @get:Optional
   @get:Input
-  abstract val queryParameters: MapProperty<String, String>
+  @get:Optional
+  abstract val schemaRelativeToProject: Property<String>
```
francescocervone, line 29: @martinbonnin is there a particular reason why this property doesn't have the `@Option` annotation?

martinbonnin: Not really. The reasoning was that when run from the command line, most tools assume parameters to be relative the current working directory. But adding `@Option` here works as well. Feel free to make a pull request. Or I can do it if you think that'll be useful.

francescocervone: I just wanted to make sure that I wasn't missing something.
I can do it, thanks!

martinbonnin: I was looking into this this morning so I made a PR there: https://github.com/apollographql/apollo-android/pull/2378

PR #2324, opened 2020-06-01
[Doc] use Kotlin in README.md and other updates

martinbonnin:
Simplify README.md:

* sample code is now Kotlin
* simplify the schema download doc (see #2321)
* use the plugins {} block now that the plugin has a plugin Marker and is available from the gradle portal
* update index.mdx, which was pretty obsolete
* other small fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 ![CI](https://github.com/apollographql/apollo-android/workflows/CI/badge.svg)
 [![GitHub release](https://img.shields.io/github/release/apollographql/apollo-android.svg)](https://github.com/apollographql/apollo-android/releases/latest)
 
-Apollo-Android is a GraphQL compliant client that generates Java and Kotlin models from standard GraphQL queries. These models give you a typesafe API to work with GraphQL servers.  Apollo will help you keep your GraphQL query statements together, organized, and easy to access. Change a query and recompile your project - Apollo code gen will rebuild your data model.  Code generation also allows Apollo to read and unmarshal responses from the network without the need of any reflection.
+Apollo-Android is a GraphQL client that generates Java and Kotlin models from GraphQL queries. These models give you a typesafe API to work with GraphQL servers.  Apollo will help you keep your GraphQL query statements together, organized, and easy to access. Change a query and recompile your project - Apollo code gen will rebuild your data model.  Code generation also allows Apollo to read and unmarshal responses from the network without the need of any reflection.
 
 Apollo-Android is designed primarily with Android in mind but you can use it in any Java/Kotlin app. The android-only parts are in `apollo-android-support` and are only needed to use SQLite as a cache or the android main thread for callbacks.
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -25,122 +25,95 @@ Apollo-android features:
 
 The latest Gradle plugin version is [ ![Download](https://api.bintray.com/packages/apollographql/android/apollo/images/download.svg) ](https://bintray.com/apollographql/android/apollo-gradle-plugin/_latestVersion)
 
-To use this plugin, add the dependency to your project's root build.gradle file:
+In your app Gradle file, apply the `com.apollographql.apollo` plugin and add the Apollo dependencies:
```
tasomaniac, line 28: I don't think it needs to be app. What about "In your Gradle module's build.gradle(.kts) file"

README.md
```diff
@@ -25,122 +25,95 @@ Apollo-android features:
...
+  jcenter()
 }
 
 dependencies {
   implementation("com.apollographql.apollo:apollo-runtime:x.y.z")
   
-  // If not already on your classpath, you might need the jetbrains annotations
-  compileOnly("org.jetbrains:annotations:13.0")
-  testCompileOnly("org.jetbrains:annotations:13.0")
+  // optional: if you want to use the normalized cache
```
tasomaniac, line 43: Similar to these, what about adding the api too? We can have a comment like: 

```
// optional: if you do not want to depend on runtime, add apollo-api for generated code.
```

It would be really good for awareness

README.md
```diff
@@ -25,122 +25,95 @@ Apollo-android features:
...
 ```groovy
-// build.gradle or build.gradle.kts
+// app/build.gradle or app/build.gradle.kts
 apollo {
   generateKotlinModels.set(true) // or false for Java models
 }
 ```
 
-5) Execute `./gradlew generateApolloSources` to generate the models from your queries. This will create a generated `FeedQuery` Java or Kotlin source file for your query.
+5) Execute `./gradlew generateApolloSources` to generate the models from your queries. This will create a generated `LaunchDetailsQuery` Java or Kotlin source file for your query.
```
tasomaniac, line 94: This always confused me. Explicit running of this task is not necessary. Building/assembling also works.

README.md
```diff
@@ -250,20 +194,14 @@ Advanced topics are available in [the official docs](https://www.apollographql.c
 * [subscriptions.md](https://www.apollographql.com/docs/android/advanced/subscriptions/)
 * [migrations.md](https://www.apollographql.com/docs/android/essentials/migration/)
 
-
-## Changelog
```
tasomaniac, line 254: I think this is nice since it gets picked up and be available in getting started page on the docs website

README.md
martinbonnin: Yep. I chose app because I think there were other references but more generic works for me too.

README.md
martinbonnin: Sounds good, I'll remove this.

README.md
martinbonnin: It's just above, in the `Release` section.

tasomaniac: Nope, I just didn't see it. I think this is all good.

Issue #2327, opened 2020-06-02
Possibly an error in code generation


dmoklyakov:
**Summary**
Apollo generates incorrect code for object field if it recursively refers to itself in arguments.

**Description**
In my `schema.json` I have a type named `Poi` which has arguments `city` and `country` with the same type `Poi`:
![image](https://user-images.githubusercontent.com/13951438/83533306-15a1ee80-a4f8-11ea-8bdc-f399a90fc2d3.png)

When I run `gradlew generateApolloSources` task I get following data class Poi inside generated Query class:
![image](https://user-images.githubusercontent.com/13951438/83533101-d07dbc80-a4f7-11ea-965f-8be88316b2b7.png)

which is leads to following error while compiling project:
```Unresolved reference: Poi```

This problem is reproducing with Kotlin and Java models.

**Version**
Apollo GraphQL Android 2.1.0

**Comment by sav007** [comment]
There is no issue with the code generation itself, but rather issue with error is not reported properly.

I guess (from my test fixtures) `city` is missing selection set in you query:
```
 ....
 poi {
   city {
      id
  }
}
```

For any non scalar field type (object / interface / union ) you must provide selection set.

PR #2343, opened 2020-06-08
Add check for missing selection set for object types.

sav007:
Validate that any field that of `Object / Interface / Union` type must have selection set.

Also add test fixture for recursive data structure for future.

Closes https://github.com/apollographql/apollo-android/issues/2327


tasomaniac:
Good one.

PR #2345, opened 2020-06-09
Formatting of code snippets in README.md.

jamesonwilliams:
1. Use a standard indentation of 2 spaces, per Contributing guide, in all snippets.
2. Add language cues to code segments, to improve syntax highlighting when rendered on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,26 @@ dependencies {
 }
 ```
 
-## Downloading a schema.json file
+## Downloading a `schema.json` file
 
 Apollo Android requires your GraphQL server's schema as a `schema.json` file. You can obtain the contents of this file by running an introspection query on your server.
 
 The Apollo Gradle plugin exposes a `downloadApolloSchema` task to help you obtain your schema. Provide this task your server's GraphQL endpoint and the output location for the `schema.json` file:
 
 ```bash
-./gradlew downloadApolloSchema -Pcom.apollographql.apollo.endpoint=https://your.graphql.endpoint -Pcom.apollographql.apollo.schema=src/main/graphql/com/example/schema.json
+./gradlew downloadApolloSchema \
+  -Pcom.apollographql.apollo.endpoint=https://your.graphql.endpoint \
+  -Pcom.apollographql.apollo.schema=src/main/graphql/com/example/schema.json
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -176,10 +182,10 @@ Releases are hosted on [jcenter](https://jcenter.bintray.com/com/apollographql/a
 
 Latest development changes are available in Sonatype's snapshots repository:
 
-```
-  repositories {
-    maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
-  }
+```gradle
```
martinbonnin, line 185: I never know whether I should use `groovy` or `gradle` here. It looks like Github support both. Do you know any reason to use one over the other?

tasomaniac: Github supports both not everywhere. I think our docs site may not support it actually

tasomaniac: For Gradle stuff, especially for simple things like this, I started to put only Kotlin since it also is valid in Groovy

jamesonwilliams:
> Thanks for the contribution 👍 . Can you sign the CLA so that it can be merged?

You bet!

I wasn't expecting the CLA. In hind sight, I could have raised this PR from a different account. However, I do use this account daily for work at my company. I pinged our open-source group to see about getting approval for that. Will sign it when I can!

README.md
jamesonwilliams: If I understand your meaning, I could use `kotlin` as the format flavor?

I found [an example docs page](https://raw.githubusercontent.com/apollographql/apollo-android/master/docs/source/essentials/migration.md) which includes this block:
~~~
```kotlin:title=build.gradle
// Replace:
implementation("com.apollographql.apollo:apollo-android-support:x.y.z")

// With:
implementation("com.apollographql.apollo:apollo-normalized-cache-sqlite:x.y.z") // highlight-line
```
~~~

The rest are ```groovy.

martinbonnin: Yep, trying to find something that works in both Kotlin and Groovy would be nice. In this case, I believe the below would do?

```
    repositories {
        maven {
            url = uri("https://oss.sonatype.org/content/repositories/snapshots/")
        }
    }
```

PR #2349, opened 2020-06-09
[Runtime] fix AbstractMethodError due to @JvmSynthetic

### LINKED ISSUE #2333: AbstractMethodError exception from RealCacheKeyBuilder
**Description**
Setting up a query which looks like this :
```
mutation TrackUserActivity($input:TrackEventsInput!) {
  trackEvents(input:$input){
      __typename,
      completed
  }
}
```
And the expect output is something like this
```
{
  "data": {
    "trackEvents": {
      "__typename": "TrackEventsPayload",
      "completed": true
    }
  }
}
```
And my apollo client is setup like this 

```
ApolloClient.builder()
                .serverUrl(url)
                .okHttpClient(okHttpClient)
                .normalizedCache(LruNormalizedCacheFactory(EvictionPolicy.NO_EVICTION), object : CacheKeyResolver() {
                    override fun fromFieldRecordSet(field: ResponseField, recordSet: Map<String, Any>): CacheKey {
                        val id = (recordSet["id"] as? String)
                        return if (id != null) {
                            CacheKey.from(id)
                        } else {
                            CacheKey.NO_KEY
                        }
                    }

                    override fun fromFieldArguments(field: ResponseField, variables: Operation.Variables): CacheKey {
                        return CacheKey.NO_KEY
                    }
                })
```
With the above setup everytime I run it, I am seeing the error below:
Screen Shot 2020-06-04 at 2 57 17 PM

Any ideas what is wrong with this query.

**Version**
2.1.0

martinbonnin:
Convert `SortedInputFieldMapWriter` to Kotlin so that we can remove `@JvmSynthetic` and the `writeList` method gets generated correctly. 

Decompiling the Kotlin .class file shows

```
    public void writeList(@NotNull String fieldName, @NotNull Function1<? super InputFieldWriter.ListItemWriter, Unit> block) {
        Intrinsics.checkParameterIsNotNull((Object)fieldName, (String)"fieldName");
        Intrinsics.checkParameterIsNotNull(block, (String)"block");
        InputFieldWriter.DefaultImpls.writeList((InputFieldWriter)this, (String)fieldName, block);
    }
```

This method doesn't get generated in Java due to the `@JvmSynthetic`. 

closes #2333


apollo-normalized-cache/src/test/java/com/apollographql/apollo/cache/normalized/internal/CacheKeyBuilderTest.kt
```diff
@@ -0,0 +1,369 @@
...
+import com.apollographql.apollo.api.internal.InputFieldMarshaller
+import com.apollographql.apollo.api.internal.InputFieldWriter
+import com.apollographql.apollo.api.internal.UnmodifiableMapBuilder
+import com.google.common.truth.Truth
+import org.junit.Test
+import java.io.IOException
+import java.math.BigDecimal
+import java.util.HashMap
+
+class CacheKeyBuilderTest {
```
sav007, line 17: Just curious why this test was converted to Kotlin, is it required by this PR?

martinbonnin: In Kotlin, this test can make use of `InputFieldWriter.writeList(String, (ListItemWriter) -> Unit)` (the version that takes a block as parameter) and therefore reproduce the problem.

tasomaniac:
Damn it :D I just opened a PR but it was already there

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/SortedInputFieldMapWriter.kt
```diff
@@ -0,0 +1,148 @@
...
+    override fun writeMap(value: Map<String, Any?>?) {
+      if (value != null) {
+        list.add(value)
+      }
+    }
+
+  }
+
+  init {
+    this.fieldNameComparator = __checkNotNull(fieldNameComparator, "fieldNameComparator == null")
```
tasomaniac, line 145: Instead of `init`, can you move this to constructor. We also shouldn't use `checkNotNull` in kotlin.

martinbonnin: Sorry I forgot to check the automatic code conversion there 😅. Thank for the heads up!

Issue #2351, opened 2020-06-10
Migrating from 1.0.0 to 2.1.0 ignores @include annotation


GauravP123:
**Summary**
I am upgrading apollo SDK from 1.0.0 to 2.1.0. In the latest version generated model ignores the @include annotation form the fragment.

**Version**
`2.1.0`

**Description**
Query

```
query ProductDetailsQuery($frontPage : Boolean = false,$id: int) {
  details(id: $id) {
    ... on Product @include(if: $frontPage) {
      size
      name
      link
    }
    id
    description
}
```

We are using the above query on the Home screen for various scenarios.
1st scenario: we are passing `frontPage=true` 
2nd scenario: we are sending `frontPage=false`

The previous version(1.0.0) generates the following model 

` final @Nullable Integer size;`

But after migration, it's generating 

` final Integer size;`

Because of this in the 2nd scenario we are getting the following exception `cause: java.lang.IllegalStateException: corrupted response reader, expected nonnull value for size`. 

it was working fine with `1.0.0`.

Any changes in a model generation logic when we use `@include`?

GauravP123:
Any plan to release this fix?

martinbonnin:
Releases typically happen every couple of weeks. If you want to try before the next one, you can use the snapshots. See [the releases section of the README](https://github.com/apollographql/apollo-android#releases) for more details on how to use SNAPSHOTS.

GauravP123:
Hi,
Have we changed a model generation logic?
`v2.2.1` creates the following class

      public interface Product {
    @NotNull String __typename();

    @NotNull String description();

    ResponseFieldMarshaller marshaller();

      final AsProduct.Mapper asProductFieldMapper = new AsProduct.Mapper();

      final AsProduct1.Mapper asProduct1FieldMapper = new AsProduct1.Mapper();
    
  }

to access, size have to do
`((ProductDetailsQuery.AsProduct)ProductDetailsQuery.Data.product()).size()`

and size is 
`    final int size;
`
its not `@Nullable`

 why are we creating 2 FieldMapper `AsProduct,AsProduct1`

Whereas on the older version 1.0.0
`ProductDetailsQuery.Data.product().size`
and size is nullable

GauravP123:
@martinbonnin @sav007 may I know what is the best practice here

martinbonnin:
@GauravP123 are you still using that same query?

```graphql
query ProductDetailsQuery($frontPage : Boolean = false,$id: int) {
[...]

PR #2352, opened 2020-06-10
Don't merge inline fragment with directives (conditions)

sav007:
Don't merge inline fragment with directives (conditions)  when it's defined on the same type (condition type) as field type.

Closes https://github.com/apollographql/apollo-android/issues/2351

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/GraphQLDocumentParser.kt
@@ -297,18 +297,17 @@ class GraphQLDocumentParser(val schema: Schema, private val packageNameProvider:
     }
 
     val fragmentRefs = selectionSet().fragmentRefs()
-    val inlineFragments = selectionSet()?.selection()?.mapNotNull { ctx ->
+    val inlineFragmentsResult = selectionSet()?.selection()?.mapNotNull { ctx ->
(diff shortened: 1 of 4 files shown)


sav007:
>For users not using include, code that was working before will fail to compile. Unless it's backward compatible but I don't think it will since hero.name will become hero.asCharacter.name

There is no any breaking changes as we are going to generate absolutely the same code as was before. We do have test fixture `inline_fragment_merge_fields` that wasn't touched by this PR.

See https://github.com/apollographql/apollo-android/blob/master/apollo-compiler/src/test/graphql/com/example/inline_fragment_merge_fields/TestOperation.graphql#L11

martinbonnin:
Got it, I missed the part where the merge is done only if conditional directives are present. If I understand currently, the behaviour with this pull request is:

Without directives (unchanged):
```
query {
   hero {
     ... on Character {
       name
      }
   }  
}

// can be used with:
data.hero.name
```

With directives (fixed):
```
query HeroQuery($withDetails: Boolean!) {
   hero {
     ... on Character @include(if: $withDetails) {
       name
      }
   }  
}

// can be used with:
data.hero.asCharacter?.name
```

Would it make sense to always generate `asCharacter` to make the generated code more consistent?

tasomaniac:
In the first one, `asCharacter` is not needed because we know the type is `Character` already. Making them consistent would definitely be nice if it is possible.

sav007:
Sure we can do this, only it's going to be a breaking changes though.

PR #2355, opened 2020-06-11
[Build] also run Github actions on tag pushes

martinbonnin:
We need tags to be built on Github Actions for releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches: [ master ]
-    tags
+    tags: '*'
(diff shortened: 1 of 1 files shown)


tasomaniac:
I am now wondering if there will be a race condition. In this case both master and tag push will be triggered at the similar times. master push should not try to publish.

martinbonnin:
The releases are only done if `$GITHUB_REF` is a tag: https://github.com/martinbonnin/apollo-android/blob/e57f3a68841e9d4b3380df44c0946a0719682def/build.gradle.kts#L358

This ensures that only proper tags can push releases

PR #2356, opened 2020-06-11
Introduce KN subscriptions

sav007:
- Introduce new API `ApolloClient.subscribe` to execute subscriptions
- Replace `GraphQLRequest.kt` / `GraphQLResponse.kt` with new `ApolloRequest.kt` / `ApolloResponse.kt`
- Refactor `NetworkTransport` to execute `ApolloRequest` and return `ApolloResponse`
- Move `ApolloHttpNetworkTransport` and everything related to own package `http`

As side effect of this refactoring is that on iOS parsing GraphQL response is done on background thread.


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/ApolloRequest.kt
```diff
@@ -4,10 +4,13 @@ import com.apollographql.apollo.api.ApolloExperimental
...
-class ApolloRequest<T> constructor(
-    val operation: Operation<*, T, *>,
+class ApolloRequest<D : Operation.Data>(
+    val operation: Operation<D, D, *>,
     val scalarTypeAdapters: ScalarTypeAdapters,
-    val executionContext: ExecutionContext = ExecutionContext.Empty
-)
+    val executionContext: ExecutionContext
+) {
+  val requestUuid = uuid4()
```
martinbonnin, line 15: Looks like `requestUuid` is specific to WebSockets, a bit similar to how headers are specific to Http. Should it be in an opaque `ExecutionContext.Element`

sav007: I would leave it to be a pat of the request `ApolloRequest` object for the future as it gives a way to match request with the corresponding response by `requestUuid`. I feel it will be helpful if someone requires it in own custom interceptor implementation.

PR #2363, opened 2020-06-15
Add opt-in to cache responses with errors

### LINKED ISSUE #2358: Allow caching partial responses with error
I hate to be the one to bring up _[it broke my workflow](https://xkcd.com/1172/)_ thing, but #2281 broke our workflow and I'm not sure how to deal with it. Basically we do make requests that return errors, because we have common fragments that we reuse for various queries. Also because of some backend logic, it's expected that some of those queries return errors due to fields being unavailable (as the user doesn't have permissions to view them). Until now everything worked, and we're fine with caching of these _partial_ responses, as the permissions level doesn't change between queries and we don't seem to end up with bad state.

Now we could refrain from updating Apollo until we figure it out, but we do need #2349 to fix another issue 🙂  I was looking for a way to work around this by e.g. substituting `errors` array with an empty list if all the errors are _expected_, but didn't find any. For example this:

```kotlin
override fun onResponse(response: ApolloInterceptor.InterceptorResponse) {
    callBack.onResponse(
        ApolloInterceptor.InterceptorResponse(
            response.httpResponse.orNull(),
            response.parsedResponse.map { it ->
                it.copy(errors = emptyList())
            }.orNull(),
            response.cacheRecords.orNull()
        )
    )
}
```
in an `ApolloInterceptor` added with `addApplicationInterceptor` doesn't seem to do the trick. 

All in all it's kind of difficult to just _Make sure that your code doesn't rely on this behaviour_. We'd appreciate an optional, even if deprecated, fallback to the original behavior, or at least a suggestion on how to work around this without having to adjust the queries that might be returning errors.

lwasyl:
Most straightforward way to fix #2358 by implementing https://github.com/apollographql/apollo-android/pull/2281#issuecomment-631102371.

I could make `STORE_PARTIAL_RESPONSES` deprecated right away, if there's another suggestion to enable old behavior. Overall I consider this a quick workaround for people affected by the breaking change introduced in #2281

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
@@ -6,6 +6,7 @@ import com.apollographql.apollo.api.internal.ApolloLogger;
 import com.apollographql.apollo.api.internal.Function;
 import com.apollographql.apollo.api.internal.Optional;
 import com.apollographql.apollo.api.internal.ResponseFieldMapper;
+import com.apollographql.apollo.cache.ApolloCacheHeaders;
 import com.apollographql.apollo.cache.normalized.ApolloStore;
 import com.apollographql.apollo.cache.normalized.ApolloStoreOperation;
 import com.apollographql.apollo.cache.normalized.Record;
@@ -123,8 +124,11 @@ public final class ApolloCacheInterceptor implements ApolloInterceptor {
 
   Set<String> cacheResponse(final InterceptorResponse networkResponse,
       final InterceptorRequest request) {
-    if (networkResponse.parsedResponse.isPresent() && networkResponse.parsedResponse.get().hasErrors()) {
-      return Collections.emptySet();
+    if (networkResponse.parsedResponse.isPresent()
+        && networkResponse.parsedResponse.get().hasErrors()
+        && !request.cacheHeaders.hasHeader(ApolloCacheHeaders.STORE_PARTIAL_RESPONSES)
+    ) {
+        return Collections.emptySet();
     }
     final Optional<List<Record>> records = networkResponse.cacheRecords.map(
         new Function<Collection<Record>, List<Record>>() {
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
```diff
@@ -124,7 +125,9 @@ InterceptorResponse resolveFromCache(InterceptorRequest request) throws ApolloEx
   Set<String> cacheResponse(final InterceptorResponse networkResponse,
       final InterceptorRequest request) {
     if (networkResponse.parsedResponse.isPresent() && networkResponse.parsedResponse.get().hasErrors()) {
-      return Collections.emptySet();
+      if (!request.cacheHeaders.hasHeader(ApolloCacheHeaders.STORE_PARTIAL_RESPONSES)) {
```
martinbonnin, line 128: Nit: maybe merge this condition with the one above? A second `if` isn't really required here.

lwasyl: Done. I hope formatting is fine, IntelliJ doesn't seem to format the file for me at all 😕

martinbonnin: Looks good, thanks!

martinbonnin:
I too like the flexibility of this solution. I also like that the default is to not cache partial responses.  

I'm not sure we want to want to advertise the new header too much though. It's an advanced use case and I'd prefer users to not use errors for normal operation. Having to dig a bit for the header would ensure that users are aware of what they are doing ?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java
martinbonnin: Damn, checkstyle is now complaining with: `'&&' should be on a new line.`. Can you move that `&&`?

tasomaniac:
If they're aware of the custom header support, then these docs on the objects are all good. I'm just afraid that not a lot of people know that this even exists.

martinbonnin:
@tasomaniac I can write some doc about this. I think we need to elaborate on the `CacheKeyResolver` documentation. I was thinking about doing a separate `Normalized Cache` page and there could be a section about partial responses. Would that work?

@sav007 any objection to merging this?

tasomaniac:
@martinbonnin up to you. You know it the best. I just wanted to bring it up. If we don't have the right context to document this particular behavior we can just mention the fact that we have options to control caching behavior.

PR #2365, opened 2020-06-15
[Build scripts] publish on maven central

martinbonnin:
Added `./gradlew publishAllPublicationsToOssStaging`. `OssStaging` is the sonatype staging repo, which is used to publish on mavenCentral.

I just published 2.2.0 there, ~the artifacts should appear soon~ the artifacts are on https://repo1.maven.org/maven2/com/apollographql/apollo/

The process is still manual for now, I'll automate it for 2.2.1 if 2.2.0 worked well.


apollo-gradle-plugin/gradle.properties
```diff
@@ -1,4 +1,3 @@
 POM_ARTIFACT_ID=apollo-gradle-plugin
 POM_NAME=Apollo Gradle Plugin
 POM_DESCRIPTION=Gradle plugin for generating java/kotlin classes for graphql files
-POM_PACKAGING=jar
```
martinbonnin, line 4: This is required to not override the default `<packaging>pom</packaging>` for the plugin marker

apollo-normalized-cache-sqlite/gradle.properties
```diff
@@ -1,4 +1,3 @@
 POM_ARTIFACT_ID=apollo-normalized-cache-sqlite
 POM_NAME=Apollo GraphQL SQLite Cache
 POM_DESCRIPTION=Apollo GraphQL SQLite Cache
-POM_PACKAGING=jar
```
martinbonnin, line 4: Same as above, this is required to not override the default `<packaging>klib</packaging>` for the plugin marker

build.gradle.kts
```diff
@@ -255,8 +249,8 @@ fun Project.configurePublishing() {
         }
       }
 
-      withType<MavenPublication>() {
-        setDefaultPomFields()
+      withType<MavenPublication> {
+        setDefaultPomFields(this)
```
martinbonnin, line 253: The scoping was wrong here and the pom properties were always read from the root `gradle.properties`

Issue #2376, opened 2020-06-18
Not needed request variables escaping 


Edwhis:
**Summary**
Request input JSON is escaped and this leads to response error

**Version**
2.2.0, kotlin-multiplatform

**Description**
I'm trying to do request like this: 
```
mutation LoginUser($user: UserLoginInput!) {
    loginUser(user: $user) {
        accessToken {
            expiresIn
            token
            tokenType
        }
        refreshToken
    }
}

```
But request payload goes to server like this:
```
{
  "operationName": "LoginUser",
  "query": "mutation LoginUser($user: UserLoginInput!) { loginUser(user: $user) { __typename accessToken { __typename expiresIn token tokenType } refreshToken } }",
  "variables": "{\"user\":{\"emailAddress\":\"email@gmail.com\",\"password\":\"pass\"}}"
}
```
instead of this:
```
{
  "operationName": "LoginUser",
  "query": "mutation LoginUser($user: UserLoginInput!) { loginUser(user: $user) { __typename accessToken { __typename expiresIn token tokenType } refreshToken } }",
  "variables": {
    "user": {
      "emailAddress": "email@gmail.com",
      "password": "pass"
    }
  }
}
```

User login input from scheme:
 ```
   {
        "kind": "INPUT_OBJECT",
        "name": "UserLoginInput",
        "description": null,
        "fields": null,
        "inputFields": [
          {
            "name": "emailAddress",
            "description": null,
            "type": {
              "kind": "NON_NULL",
              "name": null,
              "ofType": {
                "kind": "SCALAR",
                "name": "String",
                "ofType": null
              }
            },
            "defaultValue": null
          },
          {
            "name": "password",
            "description": null,
            "type": {
              "kind": "NON_NULL",
              "name": null,
              "ofType": {
                "kind": "SCALAR",
                "name": "String",
                "ofType": null
              }
            },
            "defaultValue": null
          }
        ],
        "interfaces": null,
        "enumValues": null,
        "possibleTypes": null
      },
```
Mutation.composeRequestBody() prints this, which seems to be correct:
`[size=293 text={"operationName":"LoginUser","variables":{"user":{"emailAddress"…]`

pmoldenhauer:
Experiencing same issue - any help highly appreciated!

martinbonnin:
Closed with #2377

PR #2377, opened 2020-06-18
[Multiplatform] do not escape the variables json

martinbonnin:
See #2376


Edwhis:
Is it possible to get snapshot of this?

martinbonnin:
Are the snpashot not there already ? The CI build passed (https://github.com/apollographql/apollo-android/runs/784620333?check_suite_focus=true) so the artifacts should be uploaded.

martinbonnin:
@Edwhis and 2.2.1 is now released

Issue #2364, opened 2020-06-15
Automatically resolve schema location for downloadApolloSchema Task


tasomaniac:
The `downloadApolloSchema` task is great but it requires us to publish the `schema` location manually.

If it can figure out the path from the configuration, that would be awesome.

- If there are no custom `service` defined, the path would be completely optional.
- If there are one or more `service`s present, `--service` parameter can be passed.

martinbonnin:
Should the default behaviour be to put `schema.json` at the root of the sourceSet?

```
// If no service is present:
src/main/graphql/schema.json
// If a service is present:
src/main/graphql/service/schema.json
```

I'm asking because most of the times, files are in a subdirectory that defines the package:

```
src/main/graphql/com/example/schema.json
```

The schema package is used for types and fragments and there is no way to guess it from the configuration.

tasomaniac:
That's only true for new projects where the schema does not even exist. In such a case, picking a logical default makes sense. 

I believe in the majority of the cases, users already have it and wants to update. In that case, our configuration knows the location of the schemas files perfectly. That includes all cases whether if the user defines a service or custom location of schema etc.

martinbonnin:
I see. Yep, for updating the schema, we can just lookup the existing location. I'll look into this.

PR #2378, opened 2020-06-18
[Plugin] expose --schemaRelativeToProject to the command-line

martinbonnin:
This PR makes the option available on the command line. See https://github.com/apollographql/apollo-android/pull/2321#pullrequestreview-430412707

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
@@ -26,6 +26,7 @@ abstract class ApolloDownloadSchemaTask : DefaultTask() {
 
   @get:Input
   @get:Optional
+  @get:Option(option = "schemaRelativeToProject", description = "path where the schema will be downloaded, relative to the current working directory")
   abstract val schemaRelativeToProject: Property<String>
 
   init {


tasomaniac:
I saw the earlier discussion in the review but I'm really not sure the benefit of this. It just adds more complexity without a big benefit. What do you think?

martinbonnin:
No real strong opinion here. I initially left it out for this exact reason but on the other hand it doesn't cost much. @francescocervone can you detail your use case? Is there something you can achieve with `--schemaRelativeToProject` that doesn't work with `--schema`?

francescocervone:
Before that PR, we used the `schema` option which generated the json relative to the project it was invoked.
After that PR, just by replacing `-Pcom.apollographql.apollo.schema` with `--schema`, the behavior changed. Then I saw the `schemaRelativeToProject` option and I realized. I just wondered why excluding this option.

I'll describe our use case.
We have a shell script that we invoke from outside the module that applies the Apollo plugin, so it would be helpful also having the `--schemaRelativeToProject` property.
Also, passing `-Pcom.apollographql.apollo.schemaRelativeToProject` to `downloadApolloSchema` task, doesn't seem to work.
We currently configure the `downloadApolloSchema` task in our `build.gradle` script.

francescocervone:
It's not a big deal. You can close this PR if you believe it's too hard to maintain. As long as we are able to generate our schema, we have everything we need. This is just a small improvement.

martinbonnin:
Would https://github.com/apollographql/apollo-android/issues/2364 work in your case? The idea is to lookup the location of `schema.json` automatically. You would only have to specify `--schema` for the initial download but after that, updates will overwrite the existing `schema.json`.

Unless there's a strong argument against, I'll also deprecate the `com.apollographql.apollo.XYZ` properties as the same functionality should be achievable with `--schema` and `--endpoint`.

francescocervone:
@martinbonnin I hope this will not depend upon the working directory the task is invoked. In that case, what if I have two modules with two `schema.json` files?

martinbonnin:
The automatic resolution will not depend on the working directory, just the apollo configuration in the given module.

francescocervone:
That's perfect, then. Thanks!

martinbonnin:
Closing in favor of https://github.com/apollographql/apollo-android/pull/2380
@francescocervone can you take a look?

PR #2380, opened 2020-06-19
Make `./gradlew downloadApolloSchema` resolve the schema path automatically

### LINKED ISSUE #2364: Automatically resolve schema location for downloadApolloSchema Task

martinbonnin:
The `--schema` parameter is now optional and it's possible to specify `--variant` and `--service` parameters:

```
./gradlew updateApolloSchema --endpoint https://... --variant release --service starwars
```

This uses a new gradle task called `DownloadApolloSchemaCliTask`. This task doesn't feel very "gradle-y" because not only does it have no output but it also takes `compilationUnit` as internal parameter. I'm not sure how that will play with the configuration cache. Since this task will mainly be used outside of the graph, I hope there will be no consequences. If it is a problem, I guess the `compilationUnit` can be somehow serialized.

closes #2364

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaCliTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaCliTask.kt
@@ -1,0 +1,150 @@
+package com.apollographql.apollo.gradle.internal
+
+import com.apollographql.apollo.gradle.api.CompilationUnit
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.gradle.api.DefaultTask
+import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.MapProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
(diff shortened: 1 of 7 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaCliTask.kt
```diff
@@ -0,0 +1,150 @@
...
+      headersProp != null -> {
+        logger.lifecycle("Using the com.apollographql.apollo.headers property is deprecated. Use --header instead.")
+        ApolloPlugin.toMap(headersProp)
+      }
+      else -> header.toMap()
+    }
+
+    val queryParamsProp = project.findProperty("com.apollographql.apollo.query_params") as? String
+    if (queryParamsProp != null) {
+      logger.lifecycle("Using the com.apollographql.apollo.headers property is deprecated. Add parameters to the endpoint instead.")
```
francescocervone, line 120:
```suggestion
logger.lifecycle("Using the com.apollographql.apollo.query_params property is deprecated. Add parameters to the endpoint instead.")
```

PR #2388, opened 2020-06-23
Subscription connection keep alive

sav007:
Add support for heart beat feature, see https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_keep_alive

Auto close web socket connection if `GQL_CONNECTION_KEEP_ALIVE` has not been received in configured timeout.

### DIFF
--- a/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/ws/ApolloWebSocketNetworkTransport.kt
+++ b/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/ws/ApolloWebSocketNetworkTransport.kt
@@ -241,7 +241,6 @@ class ApolloWebSocketNetworkTransport(
 
     private suspend fun close() {
       mutex.withLock {
-        yield()
         webSocketConnection.close()
       }
     }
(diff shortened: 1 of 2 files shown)


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/ws/ApolloWebSocketNetworkTransport.kt
```diff
@@ -205,22 +214,34 @@ class ApolloWebSocketNetworkTransport(
...
+            delay(idleTimeoutMs)
+            close()
+          }
         }
       }
     }
 
     private suspend fun close() {
       mutex.withLock {
+        yield()
```
martinbonnin, line 244: Why is this `yield` required?

sav007: After second thought seems there is no need for it here.

My initial thinking was that the coroutine might have been canceled before we acquire mutex, so to handle this case we call `yield ` that will fail if coroutine is canceled. But `mutex.withLock ` is suspended function so it will be handled anyway, if coroutine is canceled then `mutex.withLock` will be canceled.

PR #2399, opened 2020-06-26
[Plugin] add `com.apollographql.apollo-kotlin` and `com.apollographql.apollo-java`

martinbonnin:
Having separate plugins simplifies onboarding and allows to add apollo-android by just adding:

```
plugins {
    id("com.apollographql.apollo-kotlin").version("2.2.2")
}
```

This snippet will ~add `apollo-runtime-kotlin` to the list of dependencies automatically and also~ set `generateKotlinModels=true`.

PS: Sorry for the +12k lines diff, most of it is the schema.json for tests. If git works as I think it does, it shouldn't make the size of the repo grow since the git object will be stored only once.


apollo-compiler/build.gradle.kts
```diff
@@ -25,27 +25,38 @@ dependencies {
...
     versionFile.parentFile.mkdirs()
     versionFile.writeText("""// Generated file. Do not edit!
-package com.apollographql.android
+package com.apollographql.apollo.compiler
 val VERSION = "${project.version}"
 """)
   }
 }
 
-tasks.getByName("compileKotlin").dependsOn("pluginVersion")
```
martinbonnin, line 44: Looks like this wasn't working before since the generated `Versions.kt` file was not added to the list of sources.

sav007:
>This snippet will add apollo-runtime-kotlin to the list of dependencies automatically 

Not really sure that this is good idea, we have burned in the past to be too opinionated about Apollo as all in or nothing. Some folks might want to use Apollo for code generation only

martinbonnin:
> Not really sure that this is good idea, we have burned in the past to be too opinionated about Apollo as all in or nothing. Some folks might want to use Apollo for code generation only

There was a `apollo.addRuntimeDependency` option to opt-out runtime support but I can see how keeping it opt-in is more consistent and makes it explicit what is included in the build. The main inspiration for this was SQLDelight but they don't have the same modularity. 

Now, the benefit of having `com.apollographql.apollo-kotlin` is to skip the `apollo.generateKotlinModels.set(true)`. It might be a bit overkill to add new pluginIds just for that. On the other hand, it feels a bit weird, having to opt-in Kotlin support.

sav007:
>Now, the benefit of having a separatate plugin ID com.apollographql.apollo-kotlin is mainly to skip the apollo.generateKotlinModels.set(true).

Is it possible to infer this by looking into applied plugins for user's project? 
For instance if we see that kotlin plugin is applied and `generateKotlinModels` is not set (we give option for user to override default behaviour) we set it to `true` by default.

martinbonnin:
Yes! That'd be a breaking change but I guess there aren't that many users using java models with the kotlin pluginOn Jun 30, 2020 7:37 PM, Ivan Savytskyi <notifications@github.com> wrote:

Now, the benefit of having a separatate plugin ID com.apollographql.apollo-kotlin is mainly to skip the apollo.generateKotlinModels.set(true).

Is it possible to infer this by looking into applied plugins for user's project?
For instance if we see that kotlin plugin is applied and generateKotlinModels is not set (we give option for user to override default behaviour) we set it to true by default.

martinbonnin:
Closed in favor of https://github.com/apollographql/apollo-android/pull/2424/files

Issue #2261, opened 2020-05-11
Support SDL schemas in addition to introspection results


martinbonnin:
Taken from https://spectrum.chat/apollo/apollo-android/can-i-create-a-graphql-sdl-instead-of-using-a-schema-json~e893df98-765f-41d8-8692-583555960ce8

It would be nice to be able to drop in a SDL file instead of using the schema.json format.

FernandoH-G:
It would be a nice feature to have, especially when starting out with Apollo. I was initially put off by the introspection tool. 
Also, if this feature would get implemented, one would be able to use the GraphQL plugin for android studio. JS GraphQL.

The plugin works when you create a schema.graphql using SDL, but the schema needs to be deleted in order for the apollo sources to be generated.

martinbonnin:
@FernandoH-G I have seen tthee apollo js plugin work with json introspection results. You might have to tweak a bit your `.graphqlconfig`:

Something like:
```
{
  "schema": "src/main/graphql/schema.json"
}
```

FernandoH-G:
That did it. Thank you.

sav007:
That would be great future, at least SDL more concise than JSON format.

Unfortunately I couldn't find existing Antlr grammar for this, we should write own one, to be able to parse it.

PR #2406, opened 2020-06-29
Map SDL schema to introspection

sav007:
As codegen works with introspection schema only, provide API to map SDL schema to introspection.

Fix some missed issues with SDL parsing.

Part of https://github.com/apollographql/apollo-android/issues/2261

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSdlSchema.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSdlSchema.kt
@@ -1,6 +1,12 @@
 package com.apollographql.apollo.compiler.parser.sdl
 
-data class GraphSdlSchema(
+import com.apollographql.apollo.compiler.ir.SourceLocation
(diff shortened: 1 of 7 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSdlSchema.kt
```diff
@@ -1,6 +1,12 @@
 package com.apollographql.apollo.compiler.parser.sdl
 
-data class GraphSdlSchema(
+import com.apollographql.apollo.compiler.ir.SourceLocation
+import com.apollographql.apollo.compiler.parser.sdl.GraphSDLSchemaParser.parse
+import java.io.File
+import java.util.Locale
+import com.apollographql.apollo.compiler.parser.Schema as IntrospectionSchema
```
martinbonnin, line 7: Could we rename `Schema` to `IntropspectionSchema` instead to avoid the aliased import?

sav007: For sure, it's in my list for the next PR

PR #2408, opened 2020-06-29
[Tooling] 🛠️  Add tasks to check if the API is backward compatible

### LINKED ISSUE #1814: [Build Scripts] Automate japicmp
Following https://github.com/apollographql/apollo-android/pull/1795, it would be nice to automate japicmp, maybe with custom rules: https://github.com/melix/japicmp-gradle-plugin#custom-rules

martinbonnin:
Closes #1814 
This pull requests adds 4 new tasks in each module:

* **generateJapicmp**:  will compare against the latest released jar (or the `japicmpBaselineVersion` property) and fail if incompatible changes are found generate a report in $module/build/reports/japicmp/japicmp.html
* **checkJapicmp**: will compare against the latest released jar (or the `japicmpBaselineVersion` property) and fail if incompatible changes are found
* **generateMetalava**: generates a listing of the API in $module/api.txt
* **checkMetalava**: compares the current source against $module/api.txt. The nice thing with this is that contrary to japicmp, it can easily be source controled. Breaking an api will make the build fail until the author either fixes the api or commits a new `api.txt` if the breakage is intended

A lot of the inspiration comes from the [firebase-android-sdk repo](https://github.com/firebase/firebase-android-sdk/tree/631484a1407da3dde3cfce23324a466ffd3f01b3/buildSrc/src/main/java/com/google/firebase/gradle/plugins/apiinfo). [Metalava](https://android.googlesource.com/platform/tools/metalava/) is the tool Google uses to diff the API of the Android source code. I _think_ metalava isn't as precise as japicmp but the fact that it works on readable text files and not binary jars like japicmp makes it way easier to automate without having to store jars in the repo. 

`checkMetalava` is now called in CI so we know when the API moves. It won't consider anything under a `*.internal` package so it's safe to make breaking changes there. It might trigger a few false positive at the beginning. If that happens, calling `./gradle generateMetalava` and committing the resulting `api.txt` files should make the build pass again

### DIFF
--- a/apollo-android-support/api.txt
+++ b/apollo-android-support/api.txt
@@ -1,0 +1,24 @@
+package com.apollographql.apollo {
+
+  public final class ApolloCallback<T> {
+    ctor public ApolloCallback(ApolloCall.Callback<T>, Handler);
+    method public void onFailure(ApolloException);
+    method public void onHttpError(ApolloHttpException);
+    method public void onNetworkError(ApolloNetworkException);
+    method public void onParseError(ApolloParseException);
+    method public void onResponse(Response<T>);
+    method public void onStatusEvent(ApolloCall.StatusEvent);
+    method public static <T> com.apollographql.apollo.ApolloCallback<T> wrap(ApolloCall.Callback<T>, Handler);
+  }
+
+  public final class ApolloPrefetchCallback {
+    ctor public ApolloPrefetchCallback(ApolloPrefetch.Callback, Handler);
+    method public void onFailure(ApolloException);
+    method public void onHttpError(ApolloHttpException);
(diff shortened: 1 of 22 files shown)


sav007:
Is `japicmp` still considered to be used alongside with Metalava or should we completely switched to Metalava?

martinbonnin:
I'm not sure yet. Metalava is integrated with the CI, which is useful to detect breaking changes as soon as possible.But I feel like Japicmp is more widespread and will detect more corner cases so I'd like to keep the option to run it manually or in the release.kts script.On Jun 30, 2020 7:40 PM, Ivan Savytskyi <notifications@github.com> wrote:
Is japicmp still considered to be used alongside with Metalava or should we completely switched to Metalava?

Issue #2402, opened 2020-06-26
[Performance] Move normalized ApolloStore write transactions to background thread.


BenSchwab:
**Question.**

We noticed a significant performance regression (particularly on large responses) after the Apollo 2.x upgrade, and the switch  [androidx/sqldelight](https://github.com/apollographql/apollo-android/pull/2158) database while using the normalized sql cache.

Part of this can be mitigated by running the database actions in a [transaction](https://github.com/apollographql/apollo-android/pull/2158/files#diff-626cc33cc6e124578a410f489be79d86L93) like the old sql cache, but even with that we noticed a perf hit with the new db.

A more complete solution is to make the `ApolloCacheInterceptor` perform the [response saving](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java#L73) on a new thread - there may be a minute risk that a new request that is executed after a response of a previous request will read before the first request has dispatched it's save, but that seems quite unlikely, and the worst case is that you get a cache miss.

Overall it seems pretty safe as:

1) The [Store is already thread safe ](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.java#L179)

2) All side effects of saving a response are [already dispatched on a new thread](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java#L178)

Does anyone see any issues with making this change?

martinbonnin:
We should be able to reach the same performance without adding new threads, right? 

Or is there a more fundamental difference between SQLDelight and the Android DB? I'm always a bit reluctant to adding threads if not 100% necessary as it's another source of race conditions and sneaky bugs.

BenSchwab:
I know the lack of the transaction is one reason of the slowdown, I need to do a bit more research into what else changed. Agree that we should optimize performance of the database regardless.

However, from a design perspective, it doesn't seem correct that a database write blocks an already parsed response from being used. We have been searching for a while why migrating to GraphQL generally causes performance to regress on a page, and I am pretty sure this is the cause.

martinbonnin:
[...]

PR #2412, opened 2020-06-30
Run SqlNormalizedCache merge in transaction

BenSchwab:
See https://github.com/apollographql/apollo-android/issues/2402 for benchmarks.

This improves performance of writes by roughly 300% whenever the writes contain records with new content.


martinbonnin:
Thanks!

PR #2416, opened 2020-07-01
Run ApolloCacheInterceptor store write on dispatcher thread

BenSchwab:
Discussed in https://github.com/apollographql/apollo-android/issues/2402

This allows cache writes to not block the response, as with the NormalizedCache the cache write can be relatively expensive (>100ms on larger responses).

I believe this is pretty safe:

1) The [Store is already thread safe ](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.java#L179)

2) All side effects of saving a response are [already dispatched on a new thread](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java#L178)

3) Because the entire cache interceptor is first [placed on the dispatcher](https://github.com/apollographql/apollo-android/blame/master/apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.java#L54)
A new operation that is enqueued after a previous operation returns should have it's read always run after a previous operations write.

4) The [exception which is potentially thrown](https://github.com/apollographql/apollo-android/pull/2416/files#diff-fde3ac9879953bc292263db8c4be85c6R84) does not seem to be handled by any sync code, so moving it to a dispatcher shouldn't change behavior

### DIFF
--- a/apollo-runtime/api.txt
+++ b/apollo-runtime/api.txt
@@ -74,6 +74,7 @@ package com.apollographql.apollo {
     method public com.apollographql.apollo.ApolloClient.Builder logger(Logger);
     method public com.apollographql.apollo.ApolloClient.Builder normalizedCache(NormalizedCacheFactory);
     method public com.apollographql.apollo.ApolloClient.Builder normalizedCache(NormalizedCacheFactory, CacheKeyResolver);
+    method public com.apollographql.apollo.ApolloClient.Builder normalizedCache(NormalizedCacheFactory, CacheKeyResolver, boolean);
     method public com.apollographql.apollo.ApolloClient.Builder okHttpClient(OkHttpClient);
     method public com.apollographql.apollo.ApolloClient.Builder serverUrl(HttpUrl);
     method public com.apollographql.apollo.ApolloClient.Builder serverUrl(String);
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -98,6 +98,7 @@ public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutati
   private final SubscriptionManager subscriptionManager;
   private final boolean useHttpGetMethodForQueries;
   private final boolean useHttpGetMethodForPersistedQueries;
+  private final boolean writeToNormalizedCacheAsynchronously;
 
   ApolloClient(HttpUrl serverUrl,
       Call.Factory httpCallFactory,
@@ -114,7 +115,8 @@ public final class ApolloClient implements ApolloQueryCall.Factory, ApolloMutati
       boolean enableAutoPersistedQueries,
       SubscriptionManager subscriptionManager,
       boolean useHttpGetMethodForQueries,
-      boolean useHttpGetMethodForPersistedQueries) {
+      boolean useHttpGetMethodForPersistedQueries,
+      boolean writeToNormalizedCacheAsynchronously) {
     this.serverUrl = serverUrl;
     this.httpCallFactory = httpCallFactory;
     this.httpCache = httpCache;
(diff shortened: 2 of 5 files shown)


martinbonnin:
Thanks for diving into this! 

Looking at the [default dispatcher](https://github.com/apollographql/apollo-android/blob/d354c69cd374ab48001b92bbd3641c3c9d164bd6/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java#L761), it's a `ThreadPoolExecutor`. Does that guarantee a FIFO order of execution? More specifically, from [`SynchronousQueue`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/SynchronousQueue.html#SynchronousQueue()): 

```
public SynchronousQueue()
Creates a SynchronousQueue with nonfair access policy
```

So the queue itself doesn't seem to guarantee order. Even if it did, wouldn't there be another race condition with multiple threads reading/writing the cache?

BenSchwab:
Good find on the queue documentation. It seems like we could switch it to use a `fair` implementation to get ordering gaurentee? 

>  Even if it did, wouldn't there be another race condition with multiple threads reading/writing the cache?

I don't think so - or at least the main case we want to guarantee order is that once you receive a response, if you issue a request _after_ the response you should be guaranteed to hit the cache. If the original situation is multi-threaded we currently wouldn't have any ordering guarantees besides those imposed by the read-write lock.

martinbonnin:
I was thinking of such a scenario:

```
MainThread                       DispatchQueue                     Thread1                    Thread2
    +                                  +                              +                           +
    |                                  |                              |                           |
    |                                  |                              |                           |
    |    request1.execute              |                              |                           |
    | +------------------------------> |      writeCache              |                           |
    |                                  | +--------------------------> +--+                        |
    |    request1.response             |                              |  |                        |
    | <------------------------------+ |                              |  |                        |
    |                                  |                              |  |                        |
    |                                  |                              |  |Thread1 takes           |
    |                                  |                              |  |a long time             |
    |    request2.execute              |                              |  |to schedule             |
    | +------------------------------> |      readCache               |  |the write               |
    |                                  | +------------------------------------------------------> |
    |                                  |                              |  |                        |
    |                                  |      cacheRead               |  |                        |
    |    request2.response             | <------------------------------------------------------+ |
    | <------------------------------+ |                              |  |                        |
    |                                  |                              |  |                        |
    |                                  |                              |  |                        |
    |                                  |                              |  |                        |
    |                  Cache is written|      cacheWritten            |  |                        |
    |                  after it is read| <--------------------------+ +<-+                        |
    |                                  |                              |                           |
    +                                  +                              +                           +
```

Even though the queued jobs are dispatched in a fifo order, the scheduling of threads could still inverse the priority.

BenSchwab:
That does seem valid, but given the light-weight nature of the enqueue action, I would expect that be exceedingly rare - I don't think we would care about this edge case at Airbnb, but understand that this ordering guarantee may be something you wish to provide for Apollo

I suppose we could block returning the request until we receive some signal from that the write has started? Maybe we could pass a sempahore which is unlocked when the write is started...

BenSchwab:
Put up a proof of concept, open to other ideas for a cleaner approach!

martinbonnin:
The latch might reduce the probability of the race condition happening but I don't think it removes it completely since the consumer might schedule just after the `countDown()`:

```
      return apolloStore.writeTransaction(new Transaction<WriteableStore, Set<String>>() {
        @Nullable @Override public Set<String> execute(WriteableStore cache) {
          cacheWriteStarted.countDown();
          // request1 can resume here and (theorically at least), nothing would prevent a `request2` to run before the merge starts.
          return cache.merge(records.get(), request.cacheHeaders);
        }
      });
```

What about putting this behaviour behind some kind of "feature-flip"? Something like `ApolloClient.writeToCacheAsynchronously`? 

That would allow performance-savvy users to enable it and monitor it knowing in full consciousness without risking weird issue for other users?

BenSchwab:
@martinbonnin Updated to make this configurable when setting the normalized cache on the apollo builder

martinbonnin:
The CI is failing for a single char 😅 . Can you wrap the line?

```
[ant:checkstyle] [ERROR] /Users/runner/work/apollo-android/apollo-android/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java:393: Line is longer than 140 characters (found 141). [LineLength]
```

martinbonnin:
> The CI is failing for a single char 😅 . Can you wrap the line?

I just realized I had write access to your branch so I went ahead and wrapped all the lines!

PR #2417, opened 2020-07-02
Enable SDL schema Apollo Plugin support

sav007:
Final PR to enable SDL schema support

- Resolve and call schema parser from schema file extension.
- Update implicit schema resolution to support `sdl` extension
- Add hardcoded built in scalar types into SDL schema to be more aligned with introspection query result.

Closes https://github.com/apollographql/apollo-android/issues/2261


martinbonnin:
Nice 😃

PR #2419, opened 2020-07-02
Allow calls to WebSocketSubscriptionTransport#send when not connected.

baconz:
Failures from the transport are reported via the dispatcher, and are not synchronous, so it is possible (and quite common) for callers to attempt to send messages after the transport has moved into an errored or closed state, but before the subscription manager has received the message that the transport is closed.

I realize that `onFailure()` is probably not the right callback here. Other options are:

 * Fail silently -- I don't see any issues with this?
 * Return boolean from `send` that tells the caller whether the send was successful
 * Change the failure/close message passing to be synchronous up to the subscription manager


martinbonnin:
Thanks! 👍 

For the record, this reminds me a lot of a [similar coroutines issue about Channel.send()]( https://github.com/Kotlin/kotlinx.coroutines/issues/974#issuecomment-625694444). We might want to add a `trySend()` at some point. This can be done in another PR though. Agreed that not crashing is better here.

PR #2420, opened 2020-07-03
[Runtime] add `ApolloClient.autoPersistedOperationsInterceptorFactory`

### LINKED ISSUE #2330: Make ApolloAutoPersistedQueryInterceptor configurable

The current implementation of `ApolloAutoPersistedQueryInterceptor` has the expectation that a missing registered query will surface in the graphql errors, however some servers may choose to surface this error through an http status code, such as 400.

For the retry behavior this interceptor is easiest to manage if it is placed near the end of the chain. If one used a standard interceptor to enable this, then it would run before response fetcher interceptors, and the retry logic would trigger these interceptors to run twice.

Allow a user to specify an override to when building an Apollo client to have a custom interceptor be used for the `ApolloAutoPersistedQueryInterceptor` . This allows users to change behavior if necessary, but still place the interceptor in the proper location for retries.

```
public Builder persistedQueryInterceptor(@NotNull ApolloInterceptor interceptor) {
```

martinbonnin:
closes #2330 

This PR allows to override the APQ (Auto Persisted Queries) interceptor and also enables APQs for mutations. I'm not sure if there was a specific reason to disable APQs for mutations. I tested against https://apollo-fullstack-tutorial.herokuapp.com/ and it's working well but maybe there is something else?

`ApolloClient.Builder.setAutoPersistedOperationsInterceptorFactory()` now takes precedence over `enableAutoPersistedQueries`. This allows to keep the existing behavior. There is one breaking change: the `ApolloInterceptorFactory` signature changes from 

```kotlin
fun newInterceptor(): ApolloInterceptor
```  

to 

```kotlin
fun newInterceptor(logger: ApolloLogger, operation: Operation<*, *, *>): Operation?
```

Since `ApolloInterceptorFactory` is pretty new and it's mainly adding new parameters, I'm hoping this change will be mostly pain-free.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
@@ -74,6 +74,36 @@ interface Operation<D : Operation.Data, T, V : Operation.Variables> {
   fun parse(byteString: ByteString): Response<T>
 
   /**
+   * Composes POST JSON-encoded request body to be sent to the GraphQL server.
+   *
+   * In case when [autoPersistQueries] is set to `true` special `extension` attributes, required by query auto persistence,
+   * will be encoded along with regular GraphQL request body. If query was previously persisted on the GraphQL server
+   * set [withQueryDocument] to `false` to skip query document be sent in the request.
+   *
+   * Optional [scalarTypeAdapters] must be provided in case when this operation defines variables with custom GraphQL scalar type.
+   *
+   * *Example*:
+   * ```
+   * {
+   *    "query": "query TestQuery($episode: Episode) { hero(episode: $episode) { name } }",
+   *    "operationName": "TestQuery",
+   *    "variables": { "episode": "JEDI" }
+   *    "extensions": {
+   *      "persistedQuery": {
+   *        "version": 1,
+   *        "sha256Hash": "32637895609e6c51a2593f5cfb49244fd79358d327ff670b3e930e024c3db8f6"
+   *      }
+   *    }
+   * }
+   * ```
+   */
+  fun composeRequestBody(
+      autoPersistQueries: Boolean,
+      withQueryDocument: Boolean,
+      scalarTypeAdapters: ScalarTypeAdapters
+  ): ByteString
+
+  /**
    * Composes POST JSON-encoded request body with provided [scalarTypeAdapters] to be sent to the GraphQL server.
    *
    * *Example*:
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/interceptor/ApolloAutoPersistedOperationInterceptor.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/interceptor/ApolloAutoPersistedOperationInterceptor.java
@@ -77,7 +77,11 @@ public class ApolloAutoPersistedOperationInterceptor implements ApolloIntercepto
             logger.w("GraphQL server couldn't find Automatic Persisted Query for operation name: "
                 + request.operation.name().name() + " id: " + request.operation.operationId());
 
-            return Optional.of(request);
+            InterceptorRequest retryRequest = request.toBuilder()
+                .autoPersistQueries(true)
+                .sendQueryDocument(true)
+                .build();
+            return Optional.of(retryRequest);
           }
 
           if (isPersistedQueryNotSupported(response.getErrors())) {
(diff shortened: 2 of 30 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Operation.kt
```diff
@@ -73,6 +73,36 @@ interface Operation<D : Operation.Data, T, V : Operation.Variables> {
...
+   *    }
+   * }
+   * ```
+   */
+  fun composeRequestBody(
+      autoPersistQueries: Boolean,
+      withQueryDocument: Boolean,
+      scalarTypeAdapters: ScalarTypeAdapters
+  ): ByteString
+
```
martinbonnin, line 105: I think this isn't used by subscriptions at the moment but looking at the json, it looks like it it could be factored in.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
```diff
@@ -383,14 +386,24 @@ private ApolloInterceptorChain prepareInterceptorChain(Operation operation) {
...
     interceptors.add(new ApolloCacheInterceptor(apolloStore, responseFieldMapper, dispatcher, logger));
-    if (operation instanceof Query && enableAutoPersistedQueries) {
-      interceptors.add(new ApolloAutoPersistedQueryInterceptor(logger, useHttpGetMethodForPersistedQueries));
+    if (autoPersistedOperationsInterceptorFactory != null) {
+      ApolloInterceptor interceptor = autoPersistedOperationsInterceptorFactory.newInterceptor(logger, operation);
+      if (interceptor != null) {
+        interceptors.add(interceptor);
+      }
+    } else {
+      if (operation instanceof Query && enableAutoPersistedQueries) {
```
BenSchwab, line 404: So it seems like this choose to preserve the old behavior of not allowing persisted mutations if you don't use a custom interceptor? Fine with me since we plan to use  a custom one :) but is there plan to follow-up to investigate enabling this for everyone?

martinbonnin: The plan would be to encourage using the custom interceptor approach for more advanced use cases. Maybe even deprecate `enableAutoPersistedQueries` and `useHttpGetMethodForPersistedQueries`. There's a lot of boilerplate involved right now in passing these variables over from the `ApolloClientBuilder` to `ApolloClient` to `RealApolloCall` (and even `QueryRefetcher`) and I didn't really fell like adding a new `enableAutoPersistedMutations` boolean. It _feels_ like most of this state can be local to the interceptor. Also, `ApolloAutoPersistedOperationInterceptor` is public so that users won't have to reimplement a full interceptor but can do something like:

```kotlin
builder.setAutoPersistedOperationsInterceptorFactory(ApolloAutoPersistedOperationInterceptor.Factory())
```

apollo-runtime/src/main/java/com/apollographql/apollo/interceptor/ApolloAutoPersistedOperationInterceptor.java
```diff
@@ -108,4 +112,21 @@ boolean isPersistedQueryNotSupported(List<Error> errors) {
     }
     return false;
   }
+
+  public static class Factory implements ApolloInterceptorFactory {
+
+    final boolean useHttpGet;
```
sav007, line 118: should we make it configurable to enable / disable autopersistence for queries / mutations?

martinbonnin: It's relatively easy to do so by implementing a custom `ApolloInterceptorFactory`:

```kotlin
class OnlyPersistQueriesInterceptor: ApolloInterceptorFactory {
    override fun newInterceptor(logger: Logger, operation: Operation<?, ?, ?>) = if (operation is Query) {
         ApolloAutoPersistedOperationInterceptor(logger, false);
    } else {
       null
    }
}
```

But it might be easier to add more params to the `Factory`. Something like this:

```java
Factory(boolean useHttpGetForQueries, boolean persistQueries, boolean persistMutations)
```
?

The `useHttpGetForQueries` feels a bit asymmetrical though. Maybe it is the one that should be left out instead?

sav007: I was thinking of `Factory(boolean useHttpGetForQueries, boolean persistQueries, boolean persistMutations)`, so to keep everything in one place.

martinbonnin: Done. Let me know what you think.

PR #2424, opened 2020-07-06
[Build] unbreak the generation of `com.apollo.compiler.VERSION`

martinbonnin:
~Supersedes https://github.com/apollographql/apollo-android/pull/2399~

~The aim of this pull request is to remove the~

```
apollo {
  generateKotlinModels.set(true)
}
```

~If the Kotlin plugin is applied, the Apollo plugin will default to generating Kotlin models.~

This pull request fixes generating the version source file.

### DIFF
--- a/apollo-compiler/build.gradle.kts
+++ b/apollo-compiler/build.gradle.kts
@@ -25,23 +25,33 @@ dependencies {
   add("testImplementation", groovy.util.Eval.x(project, "x.dep.truth"))
 }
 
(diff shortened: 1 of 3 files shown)


martinbonnin:
Agreed, let's not break backward compatibility. I renamed this pull request and started a wishlist/brainstorming list for 3.0.0 there: https://github.com/apollographql/apollo-android/issues/2425

PR #2427, opened 2020-07-07
Unbreak metalava

martinbonnin:
First "Metalava fixing" commit. This error happened mainly because #2408 wasn't rebased before being merged and therefore did not pick up the API changes from #2412 and #2422 but these kind of things should be picked up earlier in future PRs.

I also removed `apollo-compiler` and `apollo-runtime-kotlin` from the checks as they will most likely witness API changes in the near future.


martinbonnin:
> What is the reason why we have NonExistentClass in the api.txt? Is that expected.

Yep, that's unfortunate. It comes from the missing `--classpath` option. I'm not sure how much of this is JVM specific. I was under the impression that metalava could work on KMP but maybe it's not that simple.

martinbonnin:
Merging for now as I want to rebase the other PRs. Let's see in the long run if we need to refine the classpath.

PR #2428, opened 2020-07-07
[LruNormalizedCache] Fix a most likely Java -> Kotlin conversion error

martinbonnin:


martinbonnin:
There was already a test. It was testing the result was != -1. I included the actual value to check. This computation is deterministic so I don't see a reason not to include the actual value. If the measurement method changes, the test will break but that's expected at this point.

Issue #2397, opened 2020-06-25
Debugging why and observed query is refetched


lwasyl:
In UI tests we're enqueueing responses for the network requests that we expect. However, sometimes Apollo makes additional requests that we don't expect (e.g. because a query doesn't fetch ID). What's the best way to monitor _why_ Apollo is considering a query stale and in need for a network request to refetch it? I'm looking for something where I can e.g. log that a certain query is being observed and Apollo will now make a request for it, because some specific event happend/cache keys got modified

eyalw:
So how can one debug this? I have this problem...

martinbonnin:
Hi 👋 that's unfortunately not easy and definitely something we want to improve. 

There's some information there: https://www.apollographql.com/docs/android/essentials/normalized-cache/#troubleshooting

Also latest version (`2.4.5`) included https://github.com/apollographql/apollo-android/pull/2751. I'm not sure this would be reachable from `apolloClient.query` but you can try to read the cache directly with your query (`apolloClient.apolloStore.read`) and get the missing key from the exception. 

If nothing helps, feel free to open a separate issue with your schema and queries and we investigate together.

PR #2433, opened 2020-07-08
📖  Improve cache documentation

### LINKED ISSUE #2429: Improve documentation on normalized cache
Currently, [iOS](https://www.apollographql.com/docs/ios/caching/) and [React](https://www.apollographql.com/docs/react/caching/cache-interaction/) have a lot more detailed information about how the normalized cache works and how one can interact with it. The Android documentation should be updated to include this.

Android documentation includes:
* How the default cache key is defined
* How to configure the cache keys
* How to interact with cached data

martinbonnin:
Separate HTTP and normalized cache and give more details for the normalized cache.

Closes #2429

### DIFF
--- a/docs/source/essentials/http-cache.mdx
+++ b/docs/source/essentials/http-cache.mdx
@@ -1,0 +1,101 @@
+---
+title: HTTP Cache
+---
+
+import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
+
+Apollo Android has 2 different kinds of caches: HTTP cache and normalized cache. The HTTP cache is easier to setup but also has more limitations. This page focuses on the HTTP cache. If you want to deduplicate the storage of your objects and/or notify your UI when your data changes, take a look at the [normalized cache](./normalized-cache) instead.
+
+## Http Cache
+
+To enable HTTP Cache support, add the dependency to your project's build.gradle file. The latest version is
+[![Download](https://api.bintray.com/packages/apollographql/android/apollo/images/download.svg)](https://bintray.com/apollographql/android/apollo-http-cache/_latestVersion)
+
+```kotlin:title=build.gradle
+dependencies {
+  implementation("com.apollographql.apollo:apollo-http-cache:x.y.z")
+}
+```
+
+### Raw HTTP Response Cache
+
+<MultiCodeBlock>
+
+```kotlin:title=Kotlin
+// Directory where cached responses will be stored
+val file = File(cacheDir, "apolloCache")
+
+// Size in bytes of the cache
+val size: Long = 1024 * 1024
+
+// Create the http response cache store
+val cacheStore = DiskLruHttpCacheStore(file, size)
+
+// Build the ApolloClient
+val apolloClient = ApolloClient.builder()
+    .serverUrl("/")
+    .httpCache(ApolloHttpCache(cacheStore))
+    .okHttpClient(okHttpClient)
+    .build()
+                   
+// Control the cache policy  
+val query = FeedQuery(limit = 10, type = FeedType.HOT)
+val dataResponse = apolloClient.query(query)
+    .httpCachePolicy(HttpCachePolicy.CACHE_FIRST)
+    .toDeferred()
+    .await()
+```
+
+```java:title=Java
+//Directory where cached responses will be stored
+File file = new File(cacheDir, "apolloCache");
+
+//Size in bytes of the cache
+long size = 1024*1024;
+
+//Create the http response cache store
+DiskLruHttpCacheStore cacheStore = new DiskLruHttpCacheStore(file, size); 
+
+//Build the ApolloClient
+ApolloClient apolloClient = ApolloClient.builder()
+  .serverUrl("/")
+  .httpCache(new ApolloHttpCache(cacheStore))
+  .okHttpClient(okHttpClient)
+  .build();
+
+apolloClient
+  .query(
+    FeedQuery.builder()
+      .limit(10)
+      .type(FeedType.HOT)
+      .build()
+  )
+  .httpCachePolicy(HttpCachePolicy.CACHE_FIRST)
+  .enqueue(new ApolloCall.Callback<FeedQuery.Data>() {
+    @Override public void onResponse(@NotNull Response<FeedQuery.Data> dataResponse) {
+      ...
+    }
+
+    @Override public void onFailure(@NotNull Throwable t) {
+      ...
+    }
+  }); 
+```
+
+</MultiCodeBlock>
+
+**IMPORTANT:** Caching is provided only for `query` operations. It isn't available for `mutation` operations.
+
+There are four available cache policies `HttpCachePolicy`:
+
+- `CACHE_ONLY` - Fetch a response from the cache only, ignoring the network. If the cached response doesn't exist or is expired, then
+return an error.
+- `NETWORK_ONLY` - Fetch a response from the network only, ignoring any cached responses.
+- `CACHE_FIRST` - Fetch a response from the cache first. If the response doesn't exist or is expired, then fetch a response from the
+network.
+- `NETWORK_FIRST` - Fetch a response from the network first. If the network fails and the cached response isn't expired, then return cached
+data instead.
+
+For `CACHE_ONLY`, `CACHE_FIRST` and `NETWORK_FIRST` policies you can define the timeout after what cached response is treated as expired
+and will be evicted from the http cache, `expireAfter(expireTimeout, timeUnit)`.`
+
--- a/docs/source/essentials/http-cache.mdx
+++ b/docs/source/essentials/http-cache.mdx
@@ -1,14 +1,14 @@
 ---
-title: HTTP Cache
+title: HTTP cache
 ---
 
 import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
 
-Apollo Android has 2 different kinds of caches: HTTP cache and normalized cache. The HTTP cache is easier to setup but also has more limitations. This page focuses on the HTTP cache. If you want to deduplicate the storage of your objects and/or notify your UI when your data changes, take a look at the [normalized cache](./normalized-cache) instead.
+Apollo Android provides two different kinds of caches: an HTTP cache and a normalized cache. The HTTP cache is easier to set up but also has more limitations. This page focuses on the HTTP cache. If you want to deduplicate the storage of your objects and/or notify your UI when your data changes, take a look at the [normalized cache](./normalized-cache) instead.
 
-## Http Cache
+## HTTP cache
 
-To enable HTTP Cache support, add the dependency to your project's build.gradle file. The latest version is
+To enable HTTP cache support, add the dependency to your project's `build.gradle file. The latest version is
 [![Download](https://api.bintray.com/packages/apollographql/android/apollo/images/download.svg)](https://bintray.com/apollographql/android/apollo-http-cache/_latestVersion)
 
 ```kotlin:title=build.gradle
@@ -17,7 +17,7 @@ dependencies {
 }
 ```
 
-### Raw HTTP Response Cache
+### Raw HTTP reesponse cache
 
 <MultiCodeBlock>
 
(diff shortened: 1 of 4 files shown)


tasomaniac:
Looks awesome. I've read it all and do not have any comment.

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,325 @@
...
+      // your code...
+      matchingFallbacks = ["debug"]
+    }
+  }
+}
+```
+
+</MultiCodeBlock>
+
+Once the dependency added, you can create the `SqlNormalizedCacheFactory`:
```
designatednerd, line 104: english nit: Once the dependency *is* added

martinbonnin: Done, I also took this opportunity to switch to imperative voice, that felt more positive:

```Once the dependency is added, create the `SqlNormalizedCacheFactory`:```

File: docs/source/essentials/normalized-cache.mdx [hunk 2]
```diff
@@ -0,0 +1,325 @@
...
+query AuthorById($id: String!) {
+  author(id: $id) {
+      id
+      name
+    }
+  }
+}
+```
+
+In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This is because the author is stored only once in the cache.
```
sav007, line 34: What do you think to mention that the selection set of fields should be the same or less than in cache? Otherwise we won;t be able to resolve it.

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,325 @@
...
+    id = "book1",
+    title = "Les guerriers du silence",
+    author = BookWithAuthorName.Author(
+      id = "author1",
+      name = "Pierre Bordage"
+    )
+  )
+  // Write to the store. All watchers will be notified
+  apolloClient.apolloStore.writeAndPublish(BookWithAuthorName(), data).execute()
+```
```
sav007, line 325: Do you think we should add section like trouble shooting, what user should check first / pay attention to in case of cache miss or other issues?

Also we can mentioned here how to dump the cache content:
```
Map<KClass<?>, Map<String, Record>> dump = apolloClient.getApolloStore().normalizedCache().dump();
NormalizedCache.prettifyDump(dump)
```

docs/source/essentials/normalized-cache.mdx
martinbonnin: I added a small explanation and a link to https://www.apollographql.com/blog/the-concepts-of-graphql-bc68bd819be3. Since the process of normalization is the same for all platforms, I'd like to focus on the Android specific parts in this page.

docs/source/essentials/normalized-cache.mdx
martinbonnin: > Do you think we should add section like trouble shooting

That's definitely something that could be useful (See https://github.com/apollographql/apollo-android/issues/2397). Not sure what to recommend though, did you have something specific in mind?

> Also we can mentioned here how to dump the cache content

👍

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,343 @@
...
+query AuthorById($id: String!) {
+  author(id: $id) {
+      id
+      name
+    }
+  }
+}
+```
+
+In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This works because the author is stored only once in the cache and all the fields where retrieved in the initial BookWithAuthorName query. If you were to request more fields, like `birthdate` for an example, that wouldn't work anymore.BookWithAuthorName
```
sav007, line 34: `.BookWithAuthorName` should be deleted?

docs/source/essentials/normalized-cache.mdx
sav007: I think we should start with dump cache. Second we should mention to double check how cache key generated (make sure that there is stable ids). Third mention again about selection set should be the same or less. At least to start with these 3 points.

Maybe later we can expand more.

docs/source/essentials/normalized-cache.mdx
martinbonnin: Urg, thanks

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
+---
+title: HTTP Cache
```
StephenBarlow, line 2: Lowercase 'C'

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
...
+---
+
+import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
+
+Apollo Android has 2 different kinds of caches: HTTP cache and normalized cache. The HTTP cache is easier to setup but also has more limitations. This page focuses on the HTTP cache. If you want to deduplicate the storage of your objects and/or notify your UI when your data changes, take a look at the [normalized cache](./normalized-cache) instead.
```
StephenBarlow, line 7: has->provides
2->two
"an HTTP cache and a normalized cache"
setup->set up

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+## Http Cache
```
StephenBarlow, line 9: "HTTP cache"

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+To enable HTTP Cache support, add the dependency to your project's build.gradle file. The latest version is
```
StephenBarlow, line 11: Lowercase 'C'
Code font on `build.gradle`

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
...
+[![Download](https://api.bintray.com/packages/apollographql/android/apollo/images/download.svg)](https://bintray.com/apollographql/android/apollo-http-cache/_latestVersion)
+
+```kotlin:title=build.gradle
+dependencies {
+  implementation("com.apollographql.apollo:apollo-http-cache:x.y.z")
+}
+```
+
+### Raw HTTP Response Cache
```
StephenBarlow, line 20: "response cache" (we use sentence case everywhere for article titles and headings)

docs/source/essentials/http-cache.mdx
```diff
@@ -0,0 +1,101 @@
...
+  }); 
+```
+
+</MultiCodeBlock>
+
+**IMPORTANT:** Caching is provided only for `query` operations. It isn't available for `mutation` operations.
+
+There are four available cache policies `HttpCachePolicy`:
+
+- `CACHE_ONLY` - Fetch a response from the cache only, ignoring the network. If the cached response doesn't exist or is expired, then
```
StephenBarlow, line 91: Is one of these the default? Let's specify if so.

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
+---
+title: Normalized cache
+---
+
+import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
+
+Apollo Android has 2 different kinds of caches: HTTP cache and normalized cache. The HTTP cache is easier to setup but also has more limitations. This page focuses on the normalized cache. If you're looking for a simpler albeit coarser cache, take a look at the [HTTP cache](http-cache).
```
StephenBarlow, line 7: See comment at the top of the other cache article

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+}
+```
+
+In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This works because the author is stored only once in the cache and all the fields where retrieved in the initial BookWithAuthorName query. If you were to request more fields, like `birthdate` for an example, that wouldn't work anymore.
+
+To learn more about the process of normalization, check this [blog post](https://www.apollographql.com/blog/the-concepts-of-graphql-bc68bd819be3)
+
+## Storing your data in memory
+
+Apollo Android comes with a `LruNormalizedCache` that will store your data in memory:
```
StephenBarlow, line 40: a -> an

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+  .serverUrl("https://...")
+  .normalizedCache(sqlNormalizedCacheFactory)
+  .build();
+```
+
+</MultiCodeBlock>
+
+## Chaining caches
+
+To get the best of both caches, you can chain a `LruNormalizedCacheFactory` with a `SqlNormalizedCacheFactory`:
```
StephenBarlow, line 142: a -> an before `Lru`

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+```
+
+This will store the following records:
+
+* `"favoriteBook"`: `{"id": "book1", "title": "Les guerriers du silence", "author": "ApolloCacheReference{favoriteBook.author}"}`
+* `"favoriteBook.author"`: `{"id": "author1", name": "Pierre Bordage"}`
+* `"author("id": "author1")"`: `{"id": "author1", "name": "Pierre Bordage"}`
+* `"QUERY_ROOT"`: `{"favoriteBook": "ApolloCacheReference{favoriteBook}", "author(\"id\": \"author1\")": "ApolloCacheReference{author(\"id\": \"author1\")}"}`
+
+As you can see, the author is stored twice. This is unfortunate because it takes more space but also because changing data in one place will not notify other watchers. What we want instead is this:
```
StephenBarlow, line 199: "This is undesirable, both because it takes more space, and because modifying one of those objects will not notify the watchers of the other."

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+//Build the ApolloClient
+ApolloClient apolloClient = ApolloClient.builder()
+  .serverUrl("https://...")
+  .normalizedCache(cacheFactory, resolver)
+  .build();
+```
+
+</MultiCodeBlock>
+
+For this resolver to work, every object in your graph needs to have a globally unique ID. If some of them don't have one, you can fallback to using the path as cache key by returning `CacheKey.NO_KEY`.
```
StephenBarlow, line 258: "fallback" -> "fall back"

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+
+```java
+// Get a response from the cache if possible. Else, get a response from the network
+// This is the default behavior
+ApolloCall apolloCall = apolloClient().query(new BookWithAuthorName()).responseFetcher(ApolloResponseFetchers.CACHE_FIRST)
+```
+
+</MultiCodeBlock>
+
+Other possibilities are CACHE_ONLY, NETWORK_ONLY, CACHE_AND_NETWORK_ONLY, NETWORK_FIST. See to the [`ResponseFetchers` class](https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/fetcher/ApolloResponseFetchers.java) for more details.
```
StephenBarlow, line 280: `code voice` for each constant

add 'and' between the last two constant values

docs/source/essentials/normalized-cache.mdx
```diff
@@ -0,0 +1,350 @@
...
+```
+
+```java
+Map<KClass<?>, Map<String, Record>> dump = apolloClient.getApolloStore().normalizedCache().dump();
+NormalizedCache.prettifyDump(dump)
+```
+
+</MultiCodeBlock>
+
+Make sure that no data is duplicated in the dump. If it is the case, it probably means that some objects have a wrong CacheKey. Make sure to provide a CacheKeyResolver that can work with your graph. All objects should have a unique and stable ID. That means that the ID should be the same no matter what path the object is in the graph. That also mean you have to include the identifier field in your queries to be able to use in from the `CacheKeyResolver`.
```
StephenBarlow, line 347: Code voice for `CacheKey` and `CacheKeyResolver`

docs/source/essentials/normalized-cache.mdx
martinbonnin: Today [I learnt](https://english.stackexchange.com/questions/378279/why-use-an-before-a-word-that-starts-with-a-l) I should use `an` before `l` if `el` is pronounced literally ("el")

Issue #2338, opened 2020-06-08
Count 'n' must be non-negative, but was -2.


wooyukit:
**Description**
After upgraded the apollo version from 1.0 to 2.1, built the app and it showed "Count 'n' must be non-negative, but was -2." error message. It is related to `GraphQLDocumentSourceBuilder.kt` line 93 string with indents problem.

wooyukit:
I don't know which query caused the issue. May be I can give you one sample.

```
query ArticleListQuery($name: String!, $limit: Int!, $offset: Int!, $excludeEntityIds: [String]!) {
    queue(filter: {name:$name}){
        entityId
        entityUuid
        name
        items(limit: $limit, offset: $offset, exclude: {entityIds: $excludeEntityIds}) {
            ... on Article {
                ...ArticleLite
            }
            ... on Video {
                ...VideoLite
            }
        }
    }
}
```

sav007:
hi @wooyukit sorry for the delay, but we couldn't reproduce this issue on provided sample.
Is it still a valid issue?

sav007:
I put a potential fix, but TBH if you see this issue looks like your GraphQL query is no valid.

martinbonnin:
There is a [workaround for this](https://github.com/apollographql/apollo-android/commit/6c7e785551924916a1bc9de3403975920ac23222) in 2.x and 3.x uses a [SDLWriter](https://github.com/apollographql/apollo-android/blob/d77b73f6f387ab474045f0db3c568d251ef55932/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SDLWriter.kt#L9) to avoid this.

PR #2441, opened 2020-07-09
Fix potential issue with GraphQL source builder indents

sav007:
Related to https://github.com/apollographql/apollo-android/issues/2338


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentSourceBuilder.kt
```diff
@@ -89,7 +90,7 @@ object GraphQLDocumentSourceBuilder {
     get() {
       var indent = 0
       return lines().joinToString(separator = "\n") { line ->
-        if (line.endsWith("}")) indent -= 2
+        if (line.endsWith("}")) indent = max(0, indent - 2)
```
martinbonnin, line 93: Isn't this going to potentially hide errors? What about throwing the input string instead? That'll make troubleshooting easier next time this happens?

sav007: Yeah, not great. Going to close this PR

Issue #2331, opened 2020-06-04
[Normalized Cache] Allow to set a max size


martinbonnin:
It would be great to be able to pass a max size to the Sqlite Normalized cache

ychescale9:
I assume this requires [weigher API](https://github.com/dropbox/Store/pull/156) if you're still planning to use Dropbox/Store/cache? Are there any other features / API changes you'd like to see from cache to enable these improvements?

martinbonnin:
@ychescale9 I think multiplatform is a big one. I don't think [Store](https://github.com/dropbox/Store) supports multiplatform yet?

Multiplatform would be required for the `LruNormalizedCache` to be used from [apollo-runtime-kotlin](https://github.com/apollographql/apollo-android/tree/master/apollo-runtime-kotlin)

ychescale9:
@martinbonnin multiplatform support for Store itself has been deprioritized. I’m keen to push for KMP support for the cache module first which currently is JVM only and does not depend on coroutines, so we can unblock apollo-android.

martinbonnin:
That's be great! I think `cache4` is all we need. I don't think we use store at the moment.

ahetawal-p:
@martinbonnin  Could this be also part of these improvements related to cache eviction: 
https://github.com/apollographql/apollo-android/issues/2507

PR #2446, opened 2020-07-15
[apollo-normalized-cache] Convert remaining Java classes to Kotlin

martinbonnin:
This is a plain Java -> Kotlin conversion. There should be (almost) no behaviour change (see comments in the diff for more details).

Next steps:

* use the Kotlin `multiplatform` plugin
* move `ApolloStore` and `ApolloStoreOperation` to `jvmMain`. These two classes are based on `Executors` and `AtomicReferences` so they will be hard to convert to KMP. I think we can use another API  for KMP, based on coroutines.
* move `LruNormalizedCache` to `jvmMain` as well as it depends on `com.nytimes.android:cache`, which is not KMP
* when `com.dropbox.mobile.store:cache4` becomes KMP (https://github.com/apollographql/apollo-android/issues/2331#issuecomment-658621856), move everything to `commonMain`, potentially in another module?


apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/NoOpApolloStore.kt
```diff
@@ -0,0 +1,133 @@
...
+  override fun <R> readTransaction(transaction: Transaction<ReadableStore, R>): R {
+    return transaction.execute(this)!!
+  }
+
+  override fun <R> writeTransaction(transaction: Transaction<WriteableStore, R>): R {
+    return transaction.execute(this)!!
+  }
+
+  override fun normalizedCache(): NormalizedCache {
+    error("Cannot get normalizedCache: no cache configured")
```
martinbonnin, line 74: This is a behaviour change. It used to return `null` here, which I believe would have crashed too.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/NoOpApolloStore.kt
```diff
@@ -0,0 +1,133 @@
...
+
+  override fun <D : Operation.Data, T, V : Operation.Variables> read(
+      operation: Operation<D, T, V>): ApolloStoreOperation<T> {
+    error("Cannot read operation: no cache configured")
+  }
+
+  override fun <D : Operation.Data, T, V : Operation.Variables> read(
+      operation: Operation<D, T, V>, responseFieldMapper: ResponseFieldMapper<D>,
+      responseNormalizer: ResponseNormalizer<Record>, cacheHeaders: CacheHeaders): ApolloStoreOperation<Response<T>> {
+    error("Cannot read operation: no cache configured")
```
martinbonnin, line 89: Same as above, behaviour change. It used to return emptyOperation(null), which I believe would also crash. The rationale here is that calling these do not make any sense if no cache is configured.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/NoOpApolloStore.kt
```diff
@@ -0,0 +1,133 @@
...
+    error("Cannot write fragment: no cache configured")
+  }
+
+  override fun writeAndPublish(fragment: GraphqlFragment, cacheKey: CacheKey,
+                               variables: Operation.Variables): ApolloStoreOperation<Boolean> {
+    error("Cannot writeAndPublish fragment: no cache configured")
+  }
+
+  override fun <D : Operation.Data, T, V : Operation.Variables> writeOptimisticUpdates(operation: Operation<D, T, V>, operationData: D, mutationId: UUID): ApolloStoreOperation<Set<String>> {
+    return emptyOperation(emptySet())
```
martinbonnin, line 118: For the optimistic methods, I kept `emptyOperation` as I was unsure if they could be called.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/NoOpApolloStore.kt
martinbonnin: Ok, that didn't work. The NO_OP cache is always called, looking into it...

martinbonnin: Looks like only this method was called. I still made the other calls safer when it was possible.

apollo-normalized-cache/src/main/java/com/apollographql/apollo/cache/normalized/internal/SimpleStack.kt
```diff
@@ -0,0 +1,31 @@
+package com.apollographql.apollo.cache.normalized.internal
+
+import java.util.ArrayList
+
+/**
+ * Simple stack data structure which accepts null elements. Backed by list.
+ * @param <E>
+</E> */
+class SimpleStack<E> {
+  private var backing: MutableList<E>
```
sav007, line 10: nit: I guess can be `val`?

martinbonnin:
Superseded by https://github.com/apollographql/apollo-android/pull/2457 (same commits but using a local branch and not from my fork so that I could merge the dependant PR first)

Issue #2320, opened 2020-06-01
[Infra] Make the project compile with java 13 (and most likely 11)


martinbonnin:
Errorprone currently fails with `unable to find tools.jar`. Once we have the runtime moved to Kotlin, we should remove errorprone.

PR #2449, opened 2020-07-16
[Build scripts] make the project compile with jdk13

martinbonnin:
closes https://github.com/apollographql/apollo-android/issues/2320

This pull request:
* bumps the Gradle wrapper version to 6.5.1
* removes errorprone
* fixes javadoc compilation with Java 13

The main issue preventing compilation with jdk13 was errorprone. As less and less new Java code is written and more and more users have jdk >8 installed, I think it makes sense to remove it. Let me know otherwise.


apollo-android-support/build.gradle.kts
```diff
@@ -22,7 +22,6 @@ android {
 dependencies {
   api(project(":apollo-normalized-cache-sqlite")) // TODO remove with removal of deprecated ApolloSqlHelper
   implementation(kotlin("stdlib"))
-  compileOnly(project(":apollo-runtime"))
-  compileOnly(project(":apollo-api"))
+  implementation(project(":apollo-runtime"))
```
martinbonnin, line 25: Not sure why this was `compileOnly` before

tasomaniac: This module has bunch of different stuff before. It included the Sqlite cache. Only reason that I can think of would be that some features work without runtime. But tbh, I don't think so. It is probably unnecessary.

sav007: I think it was a mistake from day 0, might was thinking that without runtime you won't be able to use this module so dependency to runtime already was present in user's build configuration, but yeah no reason.

PR #2458, opened 2020-07-20
Add version to the runtime dependency

jeffnyauke:


jeffnyauke:
I was just getting started with this in a multiplatform project, so more feedback to come! But for now, I noticed that the details of how to add the gradle plugin are missing but the title is there.

martinbonnin:
Oh my, you're completely right 👍 , I just pushed a fix. I'll merge if the preview looks good.

PR #2462, opened 2020-07-22
Fix issue with using inline fragment inside another inline fragment

sav007:
For this query we completely ignored `on Human`, `on Droid` nested inline fragments (oops):
```
query TestQuery {
  search(text: "bla-bla") {
    ... on Character {
      name
      ...on Human {
        homePlanet
      }
      ...on Droid {
        primaryFunction
      }
    }
  }
}
```


apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/fragment/HeroDetails.java
```diff
@@ -279,30 +279,27 @@ public static Builder builder() {
     }
 
     public static class Fragments {
-      final Optional<DroidDetails> droidDetails;
+      final @NotNull DroidDetails droidDetails;
```
sav007, line 282: as side effect of this PR, it is another issue addressed here. Named fragment `DroidDetails ` is defined on concrete type `Droid` that is type condition of inline fragment. So it's correct no `droidDetails ` is not optional

PR #2465, opened 2020-07-24
[Publication] Fix packaging of apollo-normalized-cache

martinbonnin:
Using a `jar` packaging confuses Gradle when artifacts are deployed in `mavenLocal`. (Somehow it works with OSS snapshots). 
Use a `pom` packaging instead since `apollo-normalized-cache` does not contain the actual jar anymore. (It is in `apollo-normalized-cache-jvm`)


tasomaniac:
LGTM. Not 100% sure but I saw people removing this completely. I wonder what would happen then 🤔

martinbonnin:
> Not 100% sure but I saw people removing this completely. I wonder what would happen then 🤔

Indeed, this can be removed from `gradle.properties`. The multiplatform plugin will set it properly.

PR #2467, opened 2020-07-24
JS support in apollo-api

andersio:
Hi, this is an attempt to make `apollo-api` available on Kotlin/JS.

### BigDecimal
`BigDecimal` is implemented using [big.js](https://github.com/MikeMcl/big.js/) via npm.

However, for `BigDecimal` to work on Kotlin/JS, it can no longer be a `Number` subclass (across all platforms), which appears to have special behavior in Kotlin/JS (denoted as a "primitive class" in JS vs typical stdlib classes), causing the compiled JS subclass failing to initialize. 

The solution introduces `BigDecimal.toNumber()` to support `CustomTypeValue.fromRawValue`. On JVM/iOS, `BigDecimal` returns itself (i.e., the inheritance remains). On JS, `toNumber()` returns either `Double` or `Long`, depending on whether there is a fractional part.

### @JsName
Annotated plenty of deprecated methods with `@JsName`, because their default JS name would otherwise clash with their property equivalents.

~~Removed deprecated methods that has a default JS name clashing with their property counterpart.~~


tasomaniac:
Thanks for your time. Do you have experience with the official JS tools from Apollo? I guess one of the reasons why we haven't done this is that Apollo is pretty advanced in js world.

Having said that, our use case is for Kotlin projects that are targeting multiple platforms. 👍

Regarding BigDecimal, we have plenty of places in the code that does instance check  against `Number`. I wonder if they would be a problem or not? Also btw, we were thinking to replace BigDecimal with just Number in the API to simplify things. We don't necessarily need BigDecimal in majority of the cases. This is being considered for version 3.

andersio:
Haven't used Apollo JS personally. This is indeed intended for a Multiplatform project. We'd love to have at least the generated query & models working (hence the focus on `apollo-api`), so that we can use Ktor to do the rest.

> Regarding BigDecimal, we have plenty of places in the code that does instance check against Number. I wonder if they would be a problem or not? 

I haven't checked except for `CustomTypeValue.fromRawValue`, which now tries to match `BigDecimal` before matching `Number`. I will take a look.

martinbonnin:
Thanks! 

Looks like now is a good time to start the `dev-3.x` branch. I've been thinking about that for some time now. That would allow (amongst other things) to:

* Remove BigDecimal from `apollo-api` and let users implement their own CustomAdapters if they need bigger precision that Int32 or Double. Maybe we don't even need `Number` and can do everything with `Double`
* Remove the Deprecated fields that trigger the nameclashes

I just pushed the `dev-3.x` branch, feel free to rebase on top of it and remove the deprecated fields. For BigDecimal support, I added an issue there: https://github.com/apollographql/apollo-android/issues/2468

apollo-api/build.gradle.kts
```diff
@@ -36,6 +44,7 @@ kotlin {
       dependsOn(commonMain)
       dependencies {
         implementation(kotlin("stdlib"))
+        implementation(groovy.util.Eval.x(project, "x.dep.kotlin.reflect"))
```
martinbonnin, line 47: I'm super reluctant to adding `kotlin-reflect` as a dependency as it's known to be a laaaaaarge dependency. See https://youtrack.jetbrains.com/issue/KT-20793 for an exemple.

`apollo-api` is pretty lightweight at the moment and given that it's imported by all other modules, I'd like to keep it that way. 

Can you elaborate why keeping a String for scalar types doesn't work ? I didn't find that many places using `KClass.qualifiedName()` and most of them are for debug/troubleshooting.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Response.kt
```diff
@@ -1,6 +1,7 @@
 package com.apollographql.apollo.api
 
 import kotlin.jvm.JvmStatic
+import kotlin.js.JsName
```
martinbonnin, line 4: Nit: unused import?

apollo-api/build.gradle.kts
andersio: [The dependency shouldn't be required since 1.3.70](https://blog.jetbrains.com/kotlin/2020/03/kotlin-1-3-70-released/), since KClass is moved out of `kotlin-reflect` jar. I will remove it.

> Can you elaborate why keeping a String for scalar types doesn't work ? I didn't find that many places using KClass.qualifiedName() and most of them are for debug/troubleshooting.

It is used by `apollo-api` to select the type adapters to use (among the set of default ones for Kotlin/Java stdlib types, and the user-supplied ones via `customTypeMapping` in Gradle).

martinbonnin: >The dependency shouldn't be required since 1.3.70, since KClass is moved out of kotlin-reflect jar. I will remove it.

Nice 👍 

> The lookup tables are currently keyed by KClass.qualifiedName(), which does not work at Kotlin/JS runtime.

That's what I'm trying to understand. 

Looking at `ScalarTypeAdapters`:

```kotlin
  fun <T : Any> adapterFor(scalarType: ScalarType): CustomTypeAdapter<T> {
    var customTypeAdapter: CustomTypeAdapter<*>? = customTypeAdapters[scalarType.typeName()]
    if (customTypeAdapter == null) {
      customTypeAdapter = DEFAULT_ADAPTERS[scalarType.className()]
    }
    return requireNotNull(customTypeAdapter) {
      "Can't map GraphQL type: `${scalarType.typeName()}` to: `${scalarType.className()}`. Did you forget to add a custom type adapter?"
    } as CustomTypeAdapter<T>
  }
```

This doesn't use `KClass.qualifiedName()`, does it ? both `typeName()` and `className()`are generated by the compiler base on either default hardcoded value (`java.lang.String`, etc...) or user input in case of a custom adapter:

```kotlin
enum class CustomType : ScalarType {
  ID {
    override fun typeName(): String = "ID"

    override fun className(): String = "kotlin.String"
  },

  URL {
    override fun typeName(): String = "URL"

    override fun className(): String = "kotlin.Any"
  }
}
```

andersio: That's a good point. I probably had made the decision to change it simply because Gradle wasn't building, ignoring the fact that `.qualifiedName()` is used only in Tests. I should revert it.

andersio:
@martinbonnin 

I have reverted the removals of deprecated methods needing `@JsName`. It has a knock-on effect in `apollo-integration` Java sources. I think it is better to be dealt with separately.

For example, removing `fromCache()` means Java users need to switch to Kotlin synthesized `getFromCache()`, but the Java convention appears to suggest naming it `isFromCache()`. 

Perhaps with all the unnecessary bits reverted, now it makes sense to target 2.x/master instead?

martinbonnin:
> the Java convention appears to suggest naming it isFromCache()

I see. We should rename that Kotlin field to `isFromCache` 

> with all the unnecessary bits reverted, perhaps it makes sense to target 2.x/master instead?

As long as there's no big breaking change it's fine. The only one I can see right now is that `BigDecimal` does not inherit `Number` anymore but that's acceptable.

PR #2473, opened 2020-07-25
[apollo-api] Parse Longs only once

martinbonnin:
~First (baby) step toward removing `BigDecimal` as a mandatory intermediate when converting from json to GraphQL `Int` and `Double`: Remove the default adapters for Float and Long.~

Edit: After some investigation, this pull request keeps the `DEFAULT_ADAPTERS` for now as that would be a breaking change. It still optimizes the parsing of `Longs`. There is no need to parse from String again since `BufferedSourceJsonReader` already parses internally. 

Follow up issue about removing `DEFAULT_ADAPTERS` there: https://github.com/apollographql/apollo-android/issues/2486

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/performance/NumberParsingTest.kt
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/performance/NumberParsingTest.kt
@@ -1,0 +1,58 @@
+package com.apollographql.apollo.performance
+
+import com.apollographql.apollo.ApolloClient
+import com.apollographql.apollo.Utils.immediateExecutor
+import com.apollographql.apollo.Utils.immediateExecutorService
+import com.apollographql.apollo.api.internal.SimpleOperationResponseParser
+import com.apollographql.apollo.fetcher.ApolloResponseFetchers
+import com.apollographql.apollo.integration.performance.GetIntsQuery
+import kotlinx.coroutines.runBlocking
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import okhttp3.Dispatcher
+import okhttp3.OkHttpClient
+import okio.ByteString
+import okio.ByteString.Companion.encodeUtf8
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.time.ExperimentalTime
+import kotlin.time.measureTime
+
+data class Data(
+    val randomInts: List<Int>
+)
+
+class NumberParsingTest {
+  @OptIn(ExperimentalTime::class)
+  @Test
+  fun parseInts() {
+
+    val random = Random.Default
+
+    val data = JsonObject(
+        mapOf(
+            "data" to JsonObject(
+                mapOf(
+                    "randomInts" to JsonArray(
+                        0.until(10000).map {
+                          JsonPrimitive(random.nextInt())
+                        }
+                    )
+                )
+            )
+        )
+    )
+
+    val json = data.toString().encodeUtf8()
+
+    val time = measureTime {
+      val operation = GetIntsQuery()
+      repeat(10000) {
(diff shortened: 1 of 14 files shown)


sav007:
I have a question regarding `Long`, looking into the code of our json reader seems like we read long values `https://github.com/apollographql/apollo-android/blob/master/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/BufferedSourceJsonReader.kt#L545`

martinbonnin:
> I have a question regarding Long, looking into the code of our json reader seems like we can read long values

It can indeed but it's unused in the default GraphQL type system. It could be useful in custom scalar type adapters but they should be registered manually and not put in `DEFAULT_ADAPTERS`? 
More generally, what do we need `DEFAULT_ADAPTERS` for? All the builtin GraphQL types are handled in a different path (`readString`, `readInt`, etc...). The only case I could find for `DEFAULT_ADAPTERS` is to handle the `ID`
 scalar type but that's one in many default adapters.

martinbonnin:
I added a couple of benchmarks. It turns out that going from `Int` -> `Double` -> `Int` is slower than `Int` -> `BigDecimal` -> `Int`. So ideally, we should avoid using intermediate representations, which make sense.

martinbonnin:
And I just pushed an optimization that divides by ~2 the time spent on `Int` parsing. The code that we had from Moshi already parses the `Longs` as it scans the Json and we were doing it a second time just after.

sav007:
>More generally, what do we need DEFAULT_ADAPTERS for?

They provide more convenience for user, when for instance some custom scalar type mapped to `java.lang.Long` there is no need to provide adapter for it. Same for `Float`.

Is there any reason to remove them?

apollo-integration/src/test/java/com/apollographql/apollo/performance/NumberParsingTest.kt
```diff
@@ -0,0 +1,103 @@
...
+                        0.until(10000).map {
+                          JsonPrimitive(random.nextInt())
+                        }
+                    )
+                )
+            )
+        )
+    )
+
+    val json = data.toString().encodeUtf8()
```
sav007, line 55: Is there any reason to not use our `JsonWriter` to compose thus test json? I mean at least we can avoid extra dependency to `kotlinx-serialization` that is used as far as I understood only for this test?

martinbonnin:
> They provide more convenience for user, when for instance some custom scalar type mapped to java.lang.Long there is no need to provide adapter for it. Same for Float.

Got it 👍 . Yea, removing these adapters will be a breaking change for users that rely on them.

> Is there any reason to remove them?

There's an argument to be made for API consistency. Some adapters like `java.util.Date` currently have to be manually added using `addCustomTypeAdapter` while others like `com.apollographql.apollo.api.FileUpload` are registered automatically. 

We could instead declare the implementations and let the user register them as needed:

```kotlin
// build.gradle.kts
apollo {
   customTypeMapping.set(mapOf("TimeStamp" to "java.lang.Long"))
}

// Apollo.kt
apolloClient = ApolloClient.builder().addCustomTypeAdapter(
                                   CustomType.TimeStamp, 
                                   ApolloCustomTypeAdapter.LongAdapter
                             )
```

We could also add `Iso8601DateAdapter`, I think this is one of the first custom scalars that users encounter.

apollo-integration/src/test/java/com/apollographql/apollo/performance/NumberParsingTest.kt
martinbonnin: It was mostly me being lazy to count the number of `beginObject` and `endObjects` using the `JsonWriter` API. I just did it.

martinbonnin:
@sav007 I restored the default adapters for now.  Follow up issue there: https://github.com/apollographql/apollo-android/issues/2486

PR #2476, opened 2020-07-27
[Gradle Plugin] add a way to generate operation ids as batch

### LINKED ISSUE #2455: [Gradle Plugin] make `operationIdGenerator` part of the task graph
Some servers require the GraphQL queries to be registered in advance for security/performance purposes. During compilation, the dev machine would send the list of queries and retrieve a list of matching `operationIds` in return. That can be done in a custom `generateOperationIds` task.

This process is hard to connect to the codegen at the moment because `generateApolloSources` dependsOn `generateOperationIds` to compute the ids but `generateOperationIds` dependsOn `generateApolloSources` for `operationOutput.json`, making it a circular dependency.

One option would be to break the ApolloGenerateSourcesTask in separate subTasks:
* `ApolloGenerateIR` -> Generates the IR and operationOutput.json
* `ApolloGenerateOperationIds` -> Generates the operationIds from operationOuput.json
* `ApolloGenerateSources`  -> Generate the models

martinbonnin:
Closes #2455 

This pull requests adds `OperationOutputGenerator`, like we had `OperationIdGenerator` before. `OperationOutputGenerator`
 works on lists of operations. This is useful for an exemple to register all operations on a backend in one shot.

Exemple:

```kotlin
apollo {
  val customOperationOutputGenerator = object: OperationOutputGenerator {
    override fun generate(operationDescriptorList: List<OperationDescriptor>): OperationOutput {
      return operationDescriptorList.map {
        it.source.md5() to it
      }.toMap()
    }

    override val version = "md5-v1"
  }

  operationOutputGenerator.set(customOperationOutputGenerator)
} 
```


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
```diff
@@ -10,10 +10,9 @@ interface OperationIdGenerator {
        */
       operationDocument: String,
 
-      /**
-       * The path to the GraphQL file
-       */
-      operationFilepath: String
+      operationName: String,
+      operationPackageName: String
```
martinbonnin, line 14: That's a breaking change. In order to make the task cacheable, I had to remove `filePath` from the generated files which means that it's better if identification is made based on name and packageName. If needed, we could add `file.name` (without path) but I don't think this would help much compared to qualified name.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Argument.kt
```diff
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.compiler.ir
 
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
 data class Argument(
```
martinbonnin, line 6: One main side effect of this pull request is that the IR class have been made Json serializable. I think that's ok but let me know if there's another way to look at this.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Fragment.kt
```diff
@@ -5,9 +5,10 @@ import com.apollographql.apollo.compiler.VisitorSpec.VISITOR_CLASSNAME
 import com.squareup.javapoet.ClassName
 import com.squareup.javapoet.FieldSpec
 import com.squareup.javapoet.TypeSpec
+import com.squareup.moshi.JsonClass
 import javax.lang.model.element.Modifier
 
-data class Fragment(
+data class ParsedFragment(
```
martinbonnin, line 11: As the `Fragment` and `Operation` IR classes contained absolute file paths, I had to duplicate the struct:

* `ParsedFragment` is used during parsing and contains the absolute file path for diagnostics purposes
* `ParsedFragment` is serialized and later used by codegen. It contains package name and name but no absolute file path.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Operation.kt
```diff
@@ -4,17 +4,31 @@ import com.apollographql.apollo.api.Operation
...
+    val operationName: String,
+    val packageName: String,
+    val operationType: String,
+    val description: String,
+    val variables: List<Variable>,
+    val source: String,
+    val fields: List<Field>,
+    val filePath: String,
+    val fragments: List<FragmentRef>
+)
```
martinbonnin, line 20: Same as for `Fragment` above

apollo-gradle-plugin/testProjects/operationIds/src/main/graphql/com/example/schema.sdl
```diff
@@ -0,0 +1,4 @@
+
+type Query {
+  greeting: String
+}
```
martinbonnin, line 4: Being able to use SDL in unit tests is lovely. Thanks @sav007 ! 🤗

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/ApolloGenerateOperationIdsTask.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+
+@CacheableTask
+abstract class ApolloGenerateOperationIdsTask : DefaultTask() {
```
DSteve595, line 19: Does this class actually need to be extended by users? Couldn't it work with just an interface containing `generateOperationOutput`, which gets invoked by a task managed by the Apollo plugin?

martinbonnin: I thought you preferred defining your own Task. Having an interface, like `OperationIdGenerator` is of course doable too. It just feels a bit less "Gradle-y" because users will have to remember to bump `operationIdGenerator.version` every time the implementation changes.

With a user-defined task, Gradle will magically determine the classpath and invalidate when needed.

DSteve595: The classpath invalidation is a good point.

martinbonnin: At this stage, I think you'll be the first user of this API so I'm happy to modify OperationIdgenerator with something like below instead:

```kotlin
abstract class OperationIdGenerator {
  abstract fun apply(operationDocument: String, name: String, packageName: String)

  // overridee for batch handling
  open fun generateOperationOutput(operationList: OperationList): OperationOutput {
    operationList.map {
      apply(it.source, it.name, it.packageName) to it
    }
  }

  // change version here everytime the implementation of this class changes
  val version: String
}
```

martinbonnin: Or if we want an interface:

```kotlin
interface BatchOperationIdGenerator {
  fun generateOperationOutput(operationList: OperationList): OperationOutput

  // change version here everytime the implementation of this class changes
  val version: String
}
```

DSteve595: I do think the automatic classpath invalidation is worth extending the custom task. Less mental load.

martinbonnin: 👍 sounds good

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Argument.kt
sav007: Could you please elaborate more on this, why it's serializable?

martinbonnin: The IR is now used as a task input/output so Gradle will need it be serialized for `UP-TO-DATE` checks. 

The `generateOperationIds` tasks needs access to the `sourceWithFragments` and most of that is the result of parsing the `.graphql` files. 

The other option is to keep everything bundled in one single task and keep relying on the `OperationIdGenerator.version` for invalidation, which feels less "Gradle-ly".

Something else to consider is multi-modules. I've been thinking about that lately and the IR would be a convenient point to inject fragments and types from another module.

martinbonnin:
@sav007 Regarding performance, below are a few build scans using [24k lines of Github queries](https://github.com/martinbonnin/apollo-android-samples/blob/master/github-schema/src/main/graphql/queries.graphql) resulting in a 10MB ir.json (and doing `rm -rf build` and changing the plugin version in between builds):

With the IR serialisation (this branch):
* https://scans.gradle.com/s/rmfibi3vrpeog -> 6.9s
* https://gradle.com/s/fdrw7h7rzg6im -> 7.1s
* https://gradle.com/s/sba6i4iaswbqw -> 7.2s

Without (master):
* https://scans.gradle.com/s/a2n5pyv5jgufe -> 6.5s
* https://scans.gradle.com/s/g4ciudzqg5zjw -> 5.4s
* https://scans.gradle.com/s/y73mhrkr3k3h4 -> 6.5s

There is a performance impact in the of order of 7-8%. Of course it depends a lot of factors like the warmness of the gradle daemon and stuff but the above should be a good approximation.
Serializing `ir.json` takes ~100ms (same for deserializing) so they account for ~200ms in the ~500ms slowdown, the rest being the bigger task graph I guess, time checking the up-to-date-ness of the inputs and computing checksum?

0.5s might not be a huge thing in the grand scheme of Android builds but it might be worth keeping everything as a single task and relying on bumping `version` as we currently do. Let me know what you think. (Also ping @DSteve595 for awareness.)

martinbonnin:
I updated this PR to keep everything in a single task and skip serialization. Sample usage is in the (edited) original post. Let me know what you think.

DSteve595:
Does `operationDescriptorList` need to be a `List`? Does a `Set` make more sense, since there's no inherent ordering?

martinbonnin:
> Does operationDescriptorList need to be a List? Does a Set make more sense, since there's no inherent ordering?

I guess it can even be a `Collection`. Let me try

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationIdGenerator.kt
martinbonnin: I reverted that since it's not serialized anymore.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Argument.kt
martinbonnin: Serialization is now removed

martinbonnin:
I'd like to make a release this week. If there's no additional comments, I'll merge this tomorrow so we have some time test the snapshots until a release at the end of the week.

PR #2478, opened 2020-07-27
Introduce new codegen AST (frontend) to support fragment as interfaces

sav007:
Introduce new codegen AST to make it possible generate named / inline fragments as interfaces, following the similar approach described by Airbnb.

This is the first phase (we can call it as frontend) of new codegen, build AST structure that very closely represents Kotlin / JVM code to be generated.

Next phase (backend) is actually take AST and transform it to the Kotlin / JVM code.

There are no tests yet as this is just a first step of new codegen, tests will be added with next PR for backend.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/ast/CodeGenerationAstBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/ast/CodeGenerationAstBuilder.kt
@@ -1,0 +1,286 @@
+package com.apollographql.apollo.compiler.next.ast
+
+import com.apollographql.apollo.api.internal.QueryDocumentMinifier
+import com.apollographql.apollo.compiler.OperationIdGenerator
+import com.apollographql.apollo.compiler.ast.CustomTypes
+import com.apollographql.apollo.compiler.escapeKotlinReservedWord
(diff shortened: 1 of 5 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/ast/CodeGenerationAstBuilder.kt
```diff
@@ -0,0 +1,286 @@
...
+import com.apollographql.apollo.compiler.parser.introspection.resolveType
+import com.apollographql.apollo.compiler.ir.Fragment as IrFragment
+
+internal fun CodeGenerationIR.buildCodeGenerationAst(
+    schema: IntrospectionSchema,
+    customTypeMap: CustomTypes,
+    typesPackageName: String,
+    fragmentsPackage: String,
+    useSemanticNaming: Boolean,
+    operationIdGenerator: OperationIdGenerator
```
martinbonnin, line 20: That will most likely conflict with https://github.com/apollographql/apollo-android/pull/2476 since it takes a `operationOutput` instead of an `operationIdGenerator` but I can rebase once this is merged.

martinbonnin:
Awesome 😃 . I just merged master into dev-3-x to make sure we don't diverge too much.

PR #2482, opened 2020-07-28
Add JS target to `apollo-api` module

omainegra:
- First attempt to bring JS support in Kotlin Multiplatform.
- One test is still failing due to some bugs in `Kolin.Number` in JS, I fixed some, and I'm still working on it, but I wanted to have early feedback.
- Maybe the PR shouldn't be merged in `master`, but in some kind of epic/feature branch, like `js-support`

### DIFF
--- a/apollo-api/build.gradle.kts
+++ b/apollo-api/build.gradle.kts
@@ -50,10 +50,9 @@ kotlin {
     }
 
     val jsMain by getting {
-      dependsOn(commonMain)
       dependencies {
         implementation(kotlin("stdlib-js"))
-        implementation("com.ionspin.kotlin:bignum-js:0.1.5")
+        implementation(groovy.util.Eval.x(project, "x.dep.bignum"))
       }
     }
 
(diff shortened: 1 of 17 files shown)


martinbonnin:
Hi and thanks for looking into this 👍 . Looks like there's a lot of JS momentum 🙌  🙌 . 

I hate to be the one saying that but another pull request was opened a few days ago about the same thing: https://github.com/apollographql/apollo-android/pull/2467. 

Do you mind looking into it and see if it matches what you had in mind?

omainegra:
🤦 That's right, many of my changes are already there. One thing different is that in my case `BigDecimal` extends `Number`, but then I had to patch the generated js to fix the issue

omainegra:
FWIW I also added a workaround to the `KClass<*>.qualifiedName` in JS, then I did had to change so much code

martinbonnin:
Regarding `qualifiedName2`, I'd prefer to avoid adding the method in `apollo-api` since it's not strictly needed at runtime, it's only used for tests where we know in advance what the qualified name will be.

If that's ok with you, I'd like to merge #2467 in master since there's been some discussion there already. And then work in `3.x` to remove `BigDecimal`. What do you think, any objection to merging #2467 ?

omainegra:
Please go ahead, it'd be nice to coordinate future efforts tho, there are still some modules to add JS support

PR #2485, opened 2020-07-28
[Doc] add a section about models naming

martinbonnin:
A simple page to explain how Apollo Android models are named and how to customize them with field aliases.


docs/source/essentials/using-aliases.mdx
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+}
+```
+
+## Using aliases to fix singularization rules
+
+The [singularization code](https://github.com/martinbonnin/apollo-android/blob/master/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt) works most of time but does not contain all the exceptions in the english dictionnary.
+
+If you encouter one, please let me know so I can put a valid example here (we usually fix those issues as we go but there certainly are a few left in the wild). Until we fix it in the singularization code, you can use an alias to control the name of the model in your code.
```
tasomaniac, line 84: The second half of the sentence uses we/us. The first half can do that too.

```suggestion
If you encouter one, please let us know so we can put a valid example here (we usually fix those issues as we go but there certainly are a few left in the wild). Until we fix it in the singularization code, you can use an alias to control the name of the model in your code.
```

docs/source/essentials/using-aliases.mdx
```diff
@@ -0,0 +1,85 @@
...
+    emergencyAddress: homeAddress {
+      street
+    }
+  }
+}
+```
+
+## Using aliases to fix singularization rules
+
+The [singularization code](https://github.com/martinbonnin/apollo-android/blob/master/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt) works most of time but does not contain all the exceptions in the english dictionnary.
```
tasomaniac, line 82:
```suggestion
The [singularization code](https://github.com/martinbonnin/apollo-android/blob/master/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt) works most of time but does not contain all the exceptions in the English dictionnary.
```

docs/source/essentials/using-aliases.mdx
martinbonnin: Good catch, thanks!

PR #2488, opened 2020-07-30
[Plugin] better error message for multiple schemas

martinbonnin:
* do not put absolute paths in the error message
* add a rootPackageName
* use `.set` instead of `=` as it works for both Groovy and Kotlin build scripts

**before**
```
  ApolloGraphQL: By default only one schema.[json | sdl] file is supported.
  Please use multiple services instead:
  apollo {
  
    service("github") {
      sourceFolder = "/Users/mbonnin/git/apollo-android-samples/multiple-services/src/main/graphql/com/github"
    }
  
    service("fullstack") {
      sourceFolder = "/Users/mbonnin/git/apollo-android-samples/multiple-services/src/main/graphql/com/apollographql/fullstack"
    }
  }
```

**after**
```
  ApolloGraphQL: By default only one schema.[json | sdl] file is supported.
  Please use multiple services instead:
  apollo {
  
    service("service1") {
      sourceFolder.set("com/service1)"
      rootPackageName.set("com.service1)
    }
  
    service("service2") {
      sourceFolder.set("com/service2)"
      rootPackageName.set("com.service2)
    }
  }
```

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
@@ -100,17 +100,20 @@ abstract class DefaultCompilationUnit @Inject constructor(
     }
 
     private fun multipleSchemaError(project: Project, schemaList: List<File>): String {
-      val services = schemaList.joinToString("\n") {
-        val name = it.parentFile.name
-        """|
-          |  service("$name") {
-          |    sourceFolder.set("com/$name)"
-          |    rootPackageName.set("com.$name)
+      val services = """|
+          |  service("service1") {
+          |    sourceFolder.set("com/service1)"
+          |    rootPackageName.set("com.service1)
+          |  }
+          |  
+          |  service("service2") {
+          |    sourceFolder.set("com/service2)"
+          |    rootPackageName.set("com.service2)
           |  }
         """.trimMargin()
-      }
       return "ApolloGraphQL: By default only one schema.[json | sdl] file is supported.\n" +
-          "Please use multiple services instead:\napollo {\n$services\n}"
+          "Please use multiple services instead:\napollo {\n$services\n}" +
+          "See https://www.apollographql.com/docs/android/essentials/plugin-configuration/#apolloextension-services-and-compilationunit for more details."
     }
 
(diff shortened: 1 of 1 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -99,11 +99,13 @@ abstract class DefaultCompilationUnit @Inject constructor(
       }
     }
 
-    private fun multipleSchemaError(schemaList: List<File>): String {
+    private fun multipleSchemaError(project: Project, schemaList: List<File>): String {
       val services = schemaList.joinToString("\n") {
+        val name = it.parentFile.name
```
tasomaniac, line 104: Are you sure the parentFile is enough? What happens if there are multiple nested folders? It might be better to `relativize` to the project folder?

martinbonnin: This is mostly informative and I wanted to keep it simple for most users where I expect `parentFile.name` to be different accross services. Another option would just be to use placeholders?

```
apollo {
  service("service1") {
    ...
  }
  service("service2") {
    ...
  }
}
```

tasomaniac: When you first did these warnings, they were really close to be something that the user can copy paste and fix their issue. Although I like that idea, it is also too much maintainance to keep that.

That's why maybe placeholders is a good idea. Another idea is to also link to the documentation??

martinbonnin: Link to the documentation is a great idea 👍 . I don't think we have a page talking about multiple services yet. Let me see if I can add one quickly

tasomaniac: 👍Just the page talking about service would be enough actually. Up to you.

martinbonnin: Ah yes, this looks enough: https://www.apollographql.com/docs/android/essentials/plugin-configuration/#apolloextension-services-and-compilationunit 👍

PR #2493, opened 2020-07-31
[SDL] add a basic validation test

martinbonnin:
I've just been bitten by this:

```
type Animal {
  species: String
}

type Cat implements Animal {
  mustaches: Int
}

type Query {
  randomAnimal: Animal
}
```

`Cat` cannot implement Animal since it's not an interface. Throw an error in that case.


sav007:
Technically server side validation should catch this error, as it's invalid schema definition? So user shouldn't have invalid schema at first place. Sure we can add this sort of validation, it's just I'm not really sure how extreme are we going to go with schema validation, it's probably not our responsibility of schema validation. At the same time we need to handle somehow invalid schema.

martinbonnin:
> Technically server side validation should catch this error, as it's invalid schema definition?

Yep

> So user shouldn't have invalid schema at first place. 

I'm writing SDL schemas "by hand" all the time now for unit tests, debugging, etc.. 😀 . I agree it's a niche use case but it's super convenient and way easier than the json. The drawback is that I sometimes write invalid schemas 😬 . 

> it's just I'm not really sure how extreme are we going to go with schema validation, it's probably not our responsibility of schema validation

Agreed. I really see it as a best effort thing. It's isolated enough that it shouldn't make it more complicated to change other parts of the code. And if it allows to save a bit of debugging time when someone copy pastes the wrong version of the schema, it's already something. 

> At the same time we need to handle somehow invalid schema.

We can see how it goes. If we have issues about this we can add more validation. If there's no problem then, it's no problem 😃

tasomaniac:
Very early days when Apollo was using the npm package for code generation, it also had validation. It would validate the whole schema. It was annoying for us since it would fail for stuff that we don't even query/use.

I think the validation is nice. One point would be to make sure  it is consistent with backend focused Apollo tools wich is not really easy.

designatednerd:
@tasomaniac We're hoping to port some of this stuff into the forthcoming [Rust CLI](https://github.com/apollographql/apollo-cli), which is going to replace our node tooling. In _theory_, that should create a binary that can be embedded with the IntelliJ plugin or easily downloaded onto any platform.

Issue #2500, opened 2020-08-05
requestHeaders is deprecated


zhchen-zhchen:
Recently I upgraded ApolloClient from the old version to v2.2.3. It seems that the `requestHeaders` is deprecated.

Which one should I use if I want to request with some customized headers?
```kotlin
apolloClient
            .query(**Query(
                para = paras
            ))
            .requestHeaders(RequestHeaders
                .builder()
                .addHeader(CUSTOM_HEADER, requestId)
                .build()
            )
```

martinbonnin:
`apolloCall.requestHeaders(..)` has been deprecated in favor of `apolloCall.toBuilder().requestHeaders(..).build()`

This was made because it was easy to ignore the return value of `requestHeaders` in RxJava extensions and therefore have the headers ignored. See https://github.com/apollographql/apollo-android/issues/2430 for more background. 
I can see how it makes chaining more verbose though, I'll try to think about something else for a future version.

zhchen-zhchen:
> `apolloCall.requestHeaders(..)` has been deprecated in favor of `apolloCall.toBuilder().requestHeaders(..).build()`
> 

Thanks for your prompt reply.
Just for confirmation,  `apolloCall.requestHeaders(..)` is only deprecated for `ApolloQueryCall`?

martinbonnin:
It should be deprecated for `ApolloMutationCall` as well. Sorry about that, a `deprecated` annotation slipped through the cracks there, I'll add it.

zhchen-zhchen:
Thanks!

tmtrademarked:
Unfortunately, `ApolloMutationCall` doesn't override `toBuilder` - which means that when you try to invoke `mutation.toBuilder()`, you get the behavior of `ApolloCall.toBuilder()`, which does *not* support using the request headers correctly. How difficult would it be to add the correct `toBuilder` override? (I'm not sure how deep that ends up reaching into the stack)

Alternately, since both `ApolloQueryCall` and `ApolloMutationCall` support the `requestHeaders` concept, this could be moved into `ApolloCall` potentially, although I'm sure that's more work. :-)

martinbonnin:
Yikes, good catch 👍 . 

[...]

Issue #2430, opened 2020-07-08
ApolloClient.rxQuery configure lambda confusing and error prone


hameno:
**Summary**
It is not clear from documentation and API usage, that you need to chain you calls inside the configure lambda. Normally with this kind of API you only configure the builder and it returns the same instance of the builder.
This lead to at least one bug in our code, becuase we were not using the results of the methods (e.g. responseFetcher)

**Version**
2.2.2

**Description**
see above

hameno:
It looks like internally it uses a toBuilder() and then build()
I think it would be better to provide the builder instance to the lambda

martinbonnin:
I feel like the consistency of the `builder()/configure` patterns can definitely be improved. Definitely something to keep in mind for `apollo-runtime-kotlin`, I'm not sure how much of it can be changed in the RX extensions without breaking the API. Maybe with some overloads/deprecation...

To make things explicit, can you give an exemple of code that doesn't work as expected and how you'd expect it to be written instead?

hameno:
I had the following (wrong) code:

```kotlin
apolloClient.rxQuery(this) {
    responseFetcher(responseFetcher)
    if (cacheHeaders != null) {
        cacheHeaders(cacheHeaders)
    } else this
}
```

I think this style would have been easier to understand:

```kotlin
apolloClient.rxQuery(this) { builder ->
    builder.responseFetcher(responseFetcher)
    if (cacheHeaders != null) {
        builder.cacheHeaders(cacheHeaders)
    }
}
```

or also possible:

```kotlin
apolloClient.rxQuery(this) { // this: Builder
    responseFetcher(responseFetcher)
    if (cacheHeaders != null) {
        cacheHeaders(cacheHeaders)
    }
}
```

martinbonnin:
Tentative improvement there: https://github.com/apollographql/apollo-android/pull/2434

It won't directly fix the `rxQuery` usages but will encourage to use `toBuilder()` to mutate values of `ApolloCall`

martinbonnin:
#2434 has been merged, I'm not sure how to follow up and make the `rxQuery` API much better. We _could_ find a new name and deprecate `rxQuery`:

```kotlin
@Deprecated("use rx3Query instead")
inline fun <D : Operation.Data, T, V : Operation.Variables> ApolloClient.rxQuery(
    query: Query<D, T, V>,
    configure: ApolloQueryCall<T>.() -> ApolloQueryCall<T> = { this }
)

[...]

PR #2501, opened 2020-08-05
[Runtime] add missing @Deprecated and `toBuilder` for ApolloMutationCall

martinbonnin:
This PR adds the `Deprecated` annotation on `ApolloMutationCall` so that it is consistent with `ApolloQueryCall`.

See https://github.com/apollographql/apollo-android/issues/2500#issuecomment-669092437

See also https://github.com/apollographql/apollo-android/issues/2430 for the original issue why deprecation was needed.

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
@@ -56,6 +56,8 @@ public interface ApolloMutationCall<T> extends ApolloCall<T> {
 
   @Deprecated @NotNull @Override ApolloMutationCall<T> clone();
 
+  @NotNull @Override ApolloMutationCall.Builder<T> toBuilder();
+
   interface Builder<T> extends ApolloCall.Builder<T> {
     @NotNull @Override ApolloMutationCall<T> build();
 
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloMutationCall.java
```diff
@@ -40,9 +52,11 @@
    * @param requestHeaders The {@link RequestHeaders} to use for this request.
    * @return The ApolloCall object with the provided {@link RequestHeaders}.
    */
-  @NotNull ApolloMutationCall<T> requestHeaders(@NotNull RequestHeaders requestHeaders);
+  @Deprecated @NotNull ApolloMutationCall<T> requestHeaders(@NotNull RequestHeaders requestHeaders);
+
+  @Deprecated @NotNull @Override ApolloMutationCall<T> clone();
 
-  @NotNull @Override ApolloMutationCall<T> clone();
+  @NotNull @Override ApolloMutationCall.Builder<T> toBuilder();
```
tasomaniac, line 59: While reviewing I remembered I also added similar capability to `ApolloClient`. There I used `newBuilder` naming since OkHttp and Retrofit use that. I don't have a strong feeling but we should align them. 

https://github.com/apollographql/apollo-android/blob/master/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java#L352

martinbonnin: I like `newBuilder()` better, it's more consistent with other libs. The question is whether we want to invest in this API if end up rebuilding it in `apollo-runtime-kotlin`. If we end up deprecating `apollo-runtime` in favor of `apollo-runtime-kotlin`, there's no need to deprecate too much.

Issue #2504, opened 2020-08-06
ApolloStore throws error when removing cacheKeys


piotr-pawlowski:
**Summary**
Exception thrown when calling `ApolloStore#remove(cacheKeys: List<CacheKey>): ApolloStoreOperation<Int>`. CacheKeys are valid and exist in normalized cache.

**Version**
2.2.2

**Description**
Error:
```
com.apollographql.apollo.exception.ApolloException: Failed to perform store operation
cause: java.lang.ClassCastException: com.squareup.sqldelight.TransactionWrapper cannot be cast to com.squareup.sqldelight.Transacter$Transaction

stactTrace:
com.apollographql.apollo.cache.normalized.sql.SqlNormalizedCache$deleteRecord$1.invoke(SqlNormalizedCache.kt:11)
com.squareup.sqldelight.TransacterImpl.transactionWithWrapper(Transacter.kt:218)
com.squareup.sqldelight.TransacterImpl.transaction(Transacter.kt:197)
com.squareup.sqldelight.Transacter$DefaultImpls.transaction$default(Transacter.kt:82)
com.apollographql.apollo.cache.normalized.sql.SqlNormalizedCache.deleteRecord(SqlNormalizedCache.kt:96)
com.apollographql.apollo.cache.normalized.sql.SqlNormalizedCache.remove(SqlNormalizedCache.kt:45)
com.apollographql.apollo.cache.normalized.lru.LruNormalizedCache.remove(LruNormalizedCache.kt:63)
com.apollographql.apollo.cache.normalized.OptimisticNormalizedCache.remove(OptimisticNormalizedCache.kt:35)
com.apollographql.apollo.cache.normalized.NormalizedCache.remove(NormalizedCache.kt:109)
com.apollographql.apollo.internal.RealApolloStore$5$1.execute(RealApolloStore.java:159)
com.apollographql.apollo.internal.RealApolloStore$5$1.execute(RealApolloStore.java:155)
com.apollographql.apollo.internal.RealApolloStore.writeTransaction(RealApolloStore.java:182)
com.apollographql.apollo.internal.RealApolloStore$5.perform(RealApolloStore.java:155)
com.apollographql.apollo.internal.RealApolloStore$5.perform(RealApolloStore.java:153)
com.apollographql.apollo.cache.normalized.ApolloStoreOperation$enqueue$1.run(ApolloStoreOperation.kt:55)
```

martinbonnin:
[...]

PR #2508, opened 2020-08-07
[Normalized Cache] bump SqlDelight version

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/2504

SqlDelight 1.4.0 introduced an incompatible change with 1.3.0. Bumping the version at least gives a path forward. It doesn't solve issues if another dependency is compiled against 1.3.0 but if all dependencies update to 1.4.0, the problem should disappear.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -22,7 +22,7 @@ def versions = [
     rxjava                : '2.0.5',
     rxjava3               : '3.0.4',
     rxandroid             : '2.0.1',
-    sqldelight            : '1.3.0',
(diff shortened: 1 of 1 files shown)


martinbonnin:
Corresponding SQLDelight issue: https://github.com/cashapp/sqldelight/issues/1865
I'm not sure a hotfix is worth the hassle given that it won't retroactively fix any lib compiled against `1.3.0`. What I'm hoping is that most SQLDelight usages are from app modules and not through transitive dependencies (like is the case here)... Then this issue should not stick too long.

PR #2511, opened 2020-08-07
A small change to update a file name in the query tutorial

nymerias:
I was working through this tutorial and noticed that the generated file name is actually `LaunchListQuery.kt` and thought it might confuse people.


martinbonnin:
Thanks a lot 🤗 !

Issue #1973, opened 2020-02-06
Manage importing fragments from different modules into a single query


juliafu1:
**Is your feature request related to a problem? Please describe.**
We have a query **(in module 3)** that has fragments from different modules and want to be able to combine into one query. 
```
query {
  User {
    section {
      ...SharedUserFragment #(from module 1)
      ...PaidUserFragment #(from module 2)
    }
  }
}
```
**Describe the solution you'd like**
I hope to be able to import the fragments from different modules into a single query. This is related to  https://github.com/apollographql/apollo-android/issues/593 and https://github.com/apollographql/apollo-android/issues/1008

Jawnnypoo:
Do we know if there have been any attempts at this? I am potentially going to look into this and was curious if there is any additional info on how it can be done.

martinbonnin:
@Jawnnypoo thanks for looking into this! 

There is [a great discussion](https://github.com/apollographql/apollo-ios/issues/1223) in the iOS repo. I'm not sure how much of that would apply to Android. Especially, I'm not sure Android needs a separate `ApolloConfig.json` since a lot of that logic could be added directly in the Gradle plugin. 

Ideally, I think this could even be a "no configuration" thing, just adding another module as a dependency with `implementation(project(":commonModule"))` could pull the appropriate fragments and custom types, adapters, configuration, etc... automagically. But I definitely didn't look very close there so there might be a lot of caveats !

martinbonnin:
The branch I'm using to investigate this: https://github.com/apollographql/apollo-android/pull/2514
It's working modulo a lot of edge cases but the general direction should be there, let me know what you think.

martinbonnin:
`2.4.0` is now released with experimental multi-modules support. Please try it out and let open new issues if needed before we remove the `@ApolloExperimental` annotation down there, most likely by the end of this year.

PR #2514, opened 2020-08-10
Multi-module support

martinbonnin:
See #1973 

## How it works:
* The parent module has a flag to enable generation of additional metadata for each compilation unit:

```kotlin
// parent/build.gradle.kts
apollo {
}
```
will generate:
```
metadata.json that contains:
-the introspection schema
-the IR fragments after parsing
-the names of generated input/enum/scalar types (see more below)
-various configuration options used by the compilationUnit: schemaPackageName,  generateKotlinModels, etc...
```

* The child module defines a dependency on the parent module:
```kotlin
dependencies {
  implementation("com.apollographql.apollo:apollo-api:2.2.3")
  // ...
  apollo(project(":parent"))
}
```
* The child module will use the parent schema (an error is thrown if the child also defines a schema) and inject the parent fragments in CodeGenerationIR before generating code.

## Handling different types
* Operations and Fragments are always generated in the module where they are defined and can only be consumed by downstream modules. 

For input/scalar/enum types, one solution is to generate everything in the root module since we don't know in advance which ones are going to be used by downstream modules. Unfortunately, [that scales poorly for big schemas](https://github.com/apollographql/apollo-android/pull/2514#issuecomment-673417360). Instead, the following is proposed:

* Keep this behaviour the default in the hope that it'll work for most schemas (Github for an exemple is ~280 input types and compilation time is acceptable) 
* For big modules, users can opt-in `alwaysGenerateTypesMatching = emptyList()`. In this case, the input/enums will be generated in the modules where they are used. If two sibling modules use the same type, an error is raised and users have to manually generated the type in the common upstream module:

```kotlin
apollo {
  // Declare the types that are used in several modules so they will be generated in this module
  alwaysGenerateTypesMatching("InputType1", "Enum1", ...)
  // Can accept regexes
  alwaysGenerateTypesMatching(".*Input")
  // By default, it will include everything so smaller schemas will not have any configuration to do
  // alwaysGenerateTypesMatching(".*")
}
```

## Questions
1. ~What options should be inherited?~  `generateKotlinModels`, `customTypeMappings`, `rootPackageName` are inherited.
2. ~The current validation errors read the fragments files to show a preview of where the error is happening. That won't work with this approach unless we also add the graphql files in the metadata which feels a bit overkill. An exemple is if an inherited fragment uses a variable that is not defined in a child query. How to handle this?~ The compiler now takes the rootProjectDir as parameter so as long as all compilations are done in the same root project, this should work.
3. ~Should we handle publication? Is there a use case to ship the metadata as a maven artifact?~ out of scope for now
4. ~I haven't looked too much at how this will behave with transitive dependencies. I feel like as long as each module only contributes it's own fragments, it _should_ be ok but not sure...~ There is a test for transitive dependencies there: https://github.com/apollographql/apollo-android/blob/0757f8878f4011146cf3ef88429ecc94c1d4f62c/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/MultiModulesTests.kt#L26
5. ~The root module, containing the schema currently generates all possible types, which can be a lot. Is this acceptable? If not what can we make about it.~ That was not acceptable, hence `generateAllTypesMatching`.
6. ~Error handling, what error is happening when when the same input type is defined in two sibling modules~ There is a test there: https://github.com/apollographql/apollo-android/blob/0757f8878f4011146cf3ef88429ecc94c1d4f62c/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/MultiModulesTests.kt#L45


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Fragment.kt
```diff
@@ -20,9 +22,9 @@ data class Fragment(
     val fields: List<Field>,
     val fragmentRefs: List<FragmentRef>,
     val inlineFragments: List<InlineFragment>,
-    val filePath: String,
+    val filePath: String?, // might be null if coming from another module
     val sourceLocation: SourceLocation
-) : CodeGenerator {
+  ) : CodeGenerator {
```
Jawnnypoo, line 27: Nit: unneeded space

Jawnnypoo:
This is looking great! I'll give it a try with our project to see if it all works in the way we need. Am I able to use jitpack.io or something similar to depend on this PR?

martinbonnin:
@Jawnnypoo Thanks!! Re: testing, I'm not sure jitpack.io works but using `mavenLocal()` is easy:

```
git clone https://github.com/apollographql/apollo-android
cd apollo-android
git checkout feature-1973/multi-module
./gradlew publishToMavenLocal
``` 

Then use that in your app:

```
buildscript { 
  repositories {
    // mavenCentral, jcenter, etc...
    mavenLocal()
  }
  classpath("com.apollographql.apollo:apollo-gradle-plugin:2.2.3-SNAPSHOT")
}

dependencies {
  // ...
  implementation("com.apollographql.apollo:apollo-runtime:2.2.3-SNAPSHOT")
}

```

Depending how you apply your plugins, that might look slightly different but the process of publishing to mavenLocal should be the same. 

There's a small sample project that I use there for testing if it can help: https://github.com/martinbonnin/apollo-android-samples/tree/master/multi-modules

BenSchwab:
(Disclaimer haven't read the code) - But how does this handle input/enum types. Does each module still generate their own versions? From the description I don't see anything about this. It would be confusing if an inherited fragment contains a different enum type than the containing query.

martinbonnin:
> (Disclaimer haven't read the code) - But how does this handle input/enum types?

Yep, I should have mentioned that, as of now, the base module generates all possible enum/input/scalar types.

>> Writing types and fragments is skipped in the child module so that we don't end up with duplicates.
> What are the conditions of skipping? Only the types used in inherited fragments, or all types/fragments?

All enum/input/scalar types are skipped in the child module. Only fragments defined in the child module are written.

What this all means is that the base module will write all possible input types, which can generate **a lot** of classes. Not sure yet if this is an issue but didn't find another way to do it since obviously we don't know in the parent module which types are going to be generated ultimately.

BenSchwab:
> The current validation errors read the fragments files to show a preview of where the error is happening. That won't work with this approach unless we also add the graphql files in the metadata which feels a bit overkill. An exemple is if an inherited fragment uses a variable that is not defined in a child query. How to handle this?

I am unfamiliar with how the apollo codegen does validation, but we marked the dependent module's graphql source as input, so that we could read the fragment contents and perform full validation.

> Should we handle publication? Is there a use case to ship the metadata as a maven artifact?
This was not a requirement for us. It makes this much more tricky. It is theoretically possible if you wanted to depend on a published module using graphql - but even keeping all the modules on the same schema version seems like a can of worms. I feel like it's a fair restriction that multi-module support only works within a single project.

> I haven't looked too much at how this will behave with transitive dependencies. I feel like as long as each module only contributes it's own fragments, it should be ok but not sure...

Fragments can reference fragments (which could be in another module), so some sort of transitive system is needed. We choose to merge transitive at each level (`api` like in gradle world)

martinbonnin:
> This was not a requirement for us. It makes this much more tricky. 

I think with the approach outlined here it'll be easier to support that use case since modules are independant. They can:

1. contribute the contain the schema and all input/enum/scalar types (schema.json)
2. contribute fragments (fragments.json)
3. contribute models (generated classes)

We could ensure that there is a single schema in the dependencies (~classpath) without too much hassle by checking inside each zip if a schema.json is present and make sure there's only one.

> for our project with a massive schema this would be prohibitively expensive for compile time. Our solution was to publish a similar registry of used input/enum types similar to the fragment registry you proposed here.

Good point, if that ends up being too prohibitive then I guess there's no other choice than compile input types on demand. 2 questions:
* What size is your schema? I'll try to run some benchmark with the github schema to get an idea of the time taken there.
* What if 2 child modules use the same input type? Where does it get generated? In the base module I guess? Which means the base module would have to be compiled last? Or make a 2-pass compilation?

BenSchwab:
> What size is your schema? 
On the order of magnitude of 10k enum/input types

> What if 2 child modules use the same input type?
Good question - then we would have a compile error, and the solution would be to tell plugin to manually specify in a common ancestor module to generate the type (we have a plugin option to specify additional types to generate, even if not used). We want to strictly use normal gradle compilation order and tasks - i.e. a contribution to a downstream module should never affect the output of an upstream module.

One other aside - the not inspecting the contents of the fragment file only works when fragments are generated as standalone data classes. So something to consider if that's on the roadmap .

martinbonnin:
> One other aside - the not inspecting the contents of the fragment file only works when fragments are generated as standalone data classes. So something to consider if that's on the roadmap .

Not sure I follow here. Are you referring to generating "Fragment as Interfaces" (#2478 )? What would be different in that case?

BenSchwab:
Yes - you need to know selection of the fragment to be able to generate the override vals in the implementing operation

martinbonnin:
> Yes - you need to know selection of the fragment to be able to generate the override vals in the implementing operation

👍  I _think_ that'd be ok because the current `fragments.json` file contains the IR so it has the fields. The possible types are also needed in all cases for validation. Definitely something to keep in mind though.

martinbonnin:
A few data points (2020 i7 13" MacBook):

```
Github Schema   : input types: 283, generateApolloSources: 0.3s, compileKotlin:   5.6s, memory usage:  830M
RandomSchema 1k : input types: 1k , generateApolloSources: 0.6s, compileKotlin:  10.1s, memory usage:  900M
RandomSchema 10k: input types: 10k, generateApolloSources: 7.4s, compileKotlin: 151.8s, memory usage: 1520M
RandomSchema 20k: input types: 20k, generateApolloSources: 9.9s, compileKotlin: 640.8s, memory usage: 2780M
```

(RandomSchema generated using [this script](https://github.com/martinbonnin/apollo-android-samples/blob/283122deb8981bd0bd1640ee538b38094d27a866/multi-modules/generateSchema.main.kts#L35))

Having 10k Kotlin classes to compile doesn't sound great, not talking about the space occupied by all these files, build cache issues, etc...

Still, I'd like to make this process as automatic and error proof and if possible avoid asking users to manually specify the extra input types needed (and also track when not required anymore, etc...)

I'm contemplating the possibility of breaking the ApolloGenerateSources task in 2 tasks again. This is something we didn't want to do for performance reasons in https://github.com/apollographql/apollo-android/pull/2476 but I think could be acceptable in this case (would have to be confirmed with cold hard numbers). Especially since some Json is already written (registry, or in this case the Fragments IR) and it brings some features (automatic detection of all input classes across all modules).

The task graph would look like this which, if I'm not mistaken would break the circular dependency between tasks:

chart

martinbonnin:
Marking this as ready to review before it grows too much (it's already quite big). I updated the initial description, I'll put more comments in the diff. 

The big change is in `CodeGenerationIR` that now contains `Sets` for Inputs, Enums, Fragments, Scalars to generate. There will most likely be conflict with the new codegen but they shouldn't be too hard to resolve.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -1,52 +1,174 @@
...
+import com.apollographql.apollo.compiler.parser.introspection.IntrospectionSchema.Companion.toIntrospectionSchema
+import com.apollographql.apollo.compiler.parser.introspection.IntrospectionSchema.Companion.wrap
+import com.apollographql.apollo.compiler.parser.sdl.GraphSdlSchema
+import com.apollographql.apollo.compiler.parser.sdl.toIntrospectionSchema
 import com.squareup.javapoet.JavaFile
+import com.squareup.kotlinpoet.asClassName
 import java.io.File
 
 class GraphQLCompiler {
   fun write(args: Arguments) {
```
martinbonnin, line 25: There are a lot of changes in this file because I moved some of the IR logic that was in the Gradle plugin here.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -1,52 +1,174 @@
...
+          schemaPackageName = schemaPackageName,
+          moduleName = args.moduleName,
+          types = ir.enumsToGenerate + ir.inputObjectsToGenerate,
+          fragments = ir.fragments.filter { ir.fragmentsToGenerate.contains(it.fragmentName) },
+          generateKotlinModels = generateKotlinModels,
+          customTypesMap = args.customTypeMap,
+          pluginVersion = com.apollographql.apollo.compiler.VERSION
+      ).let {
+        if (args.rootProjectDir != null) {
+          it.withRelativeFragments(args.rootProjectDir)
```
martinbonnin, line 128: That's the workaround to not break Gradle build cache and still be able to read back into the file to display nice parse errors.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -1,52 +1,174 @@
...
+    }
+
+    if (args.operationOutputFile != null) {
+      args.operationOutputFile.writeText(operationOutput.toJson("  "))
+    }
+
+    val customTypeMap = (introspectionSchema.types.values.filter {
+      it is IntrospectionSchema.Type.Scalar && ScalarType.forName(it.name) == null
+    }.map { it.name } + ScalarType.ID.name)
+        .supportedTypeMap(userCustomTypesMap, generateKotlinModels)
```
martinbonnin, line 82: The code around custom scalars is more complex now, I'll most likely try to refactor in a subsequent pull request. `customTypeMap` here contains all the scalar types from the schema and their matching className either defined by the user or by us (for ID and Any).

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/SchemaBuilder.kt
```diff
@@ -19,16 +18,26 @@ internal fun CodeGenerationIR.ast(
...
-  val inputTypes = typesUsed.filter { it.kind == TypeDeclaration.KIND_INPUT_OBJECT_TYPE }.map {
+  val enums = typeDeclarations.filter {
+    it.kind == TypeDeclaration.KIND_ENUM
+        // && enumsToGenerate.contains(it.name) // filtering is done later as the rest of the codegen needs all enums
+  }.map {
+    it.ast()
+  }
+  val inputTypes = typeDeclarations.filter {
+    it.kind == TypeDeclaration.KIND_INPUT_OBJECT_TYPE
+        && inputObjectsToGenerate.contains(it.name)
```
martinbonnin, line 29: That's where we tell the compiler to only generate a subset of input objects

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/builder/SchemaBuilder.kt
```diff
@@ -19,16 +18,26 @@ internal fun CodeGenerationIR.ast(
...
     it.ast(
         enums = enums,
         customTypeMap = customTypeMap,
         typesPackageName = typesPackageName
     )
   }
   val irFragments = fragments.associateBy { it.fragmentName }
-  val fragments = fragments.map {
+  val fragments = fragments.filter {
+    fragmentsToGenerate.contains(it.fragmentName)
```
martinbonnin, line 39: same for fragments (all fragments are sill needed as input to validate arguments, type conditions, etc...)

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/Argument.kt
```diff
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.compiler.ir
 
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
```
martinbonnin, line 5: The IR is made serializable (to save the Fragment in the metadata). The serialization is only done if `generateMetadata` is true so as not to slow down compilation in the general case.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    return schema.types.values.filter { type ->
+      (type.kind == IntrospectionSchema.Kind.ENUM
+          || type.kind == IntrospectionSchema.Kind.INPUT_OBJECT)
+          && regexes.indexOfFirst { it.matches(type.name) } >= 0
+    }.map { it.name }
+        .toSet()
+  }
+
+  fun build(documentParseResult: DocumentParseResult): CodeGenerationIR {
```
martinbonnin, line 38: This method decides what types to generate/not generate.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
```diff
@@ -0,0 +1,258 @@
...
+    val enumsToGenerate = typeDeclarations.filter { it.kind == KIND_ENUM }
+        .map { it.name }
+        .filter { !incomingTypes.contains(it) }
+
+    val inputObjectsToGenerate = typeDeclarations.filter { it.kind == KIND_INPUT_OBJECT_TYPE }
+        .map { it.name }
+        .filter { !incomingTypes.contains(it) }
+
+    // Always generate the extra "ID" scalar type
+    // I'm not 100% sure why this is required but keep this for backward compatibility
```
martinbonnin, line 60: Do we need to generate "ID" in the `CustomType` enum? Is there a use case where users can register a custom adapter for "ID" ?

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
```diff
@@ -218,16 +175,21 @@ class CodeGenTest(private val folder: File) {
...
-    fun data(): Collection<File> {
+    @Parameterized.Parameters(name = "{0}-{1}")
+    fun data(): Collection<Array<Any>> {
       return File("src/test/graphql/com/example/")
           .listFiles()!!
           .filter { it.isDirectory }
+          .flatMap { listOf(
+              arrayOf(it, TestLanguage.Java),
+              arrayOf(it, TestLanguage.Kotlin)
+          ) }
```
martinbonnin, line 192: Refactored these tests slightly to have one test for Java and another for Kotlin

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/CompilerParams.kt
```diff
@@ -193,4 +194,29 @@ interface CompilerParams {
...
+  /**
+   * Whether or not to generate Apollo metadata. Apollo metadata is used for multi-module support. Set this to true if you want other
+   * modules to be able to re-use fragments and types from this module.
+   *
+   * This is currently experimental and this API might change in the future.
+   *
+   * Default value: false
+   */
+  @ApolloExperimental
+  val generateApolloMetadata: Property<Boolean>
```
martinbonnin, line 207: I expect multi-module to stay in experimental for some time so that we can tweak the APIs if needed.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -45,17 +45,29 @@ abstract class DefaultCompilationUnit @Inject constructor(
     }
 
     if (!compilerParams.schemaFile.isPresent) {
-      compilerParams.schemaFile.set {
-        project.file(
+      // This needs to be lazy for:
+      // - test variants that don't have a schema
```
martinbonnin, line 49: This is a pain. For 3.0, I'd like to do something about test variants. Most likely remove them... Do we know if anyone uses that feature?

sav007: Yeah we can remove this. not really sure what is the benefit of having unique graphql queries for test variants

sav007 (approved):

Issue #2516, opened 2020-08-10
Query request doesn't work on Android 6 and 5


Edwhis:
**Summary**
As title says - query request doesn't work on Android 6 and Android 5

**Version**
2.2.3

**Description**
I'm using kotlin multiplatform library for my project. Everything seems to be working fine on Android 7 and up, but query requests doesn't work on Android 6 and Android 5.  Mutation requests works as expected!
App is not crashing, but before every query logcat prints:
`Rejecting re-init on previously-failed class java.lang.Class<com.apollographql.apollo.network.http.ApolloHttpNetworkTransport$toHttpPostRequest$1$1>
`
I checked profiler and tried RequestBin - requests are not fired from phone. Any ideas?

martinbonnin:
Thanks for reaching out! 

This is most likely coming from [this usage of `Map.forEach()`](https://github.com/apollographql/apollo-android/blob/1465640c4bd8920a3755d0df62b09e6ae568c262/apollo-runtime-kotlin/src/jvmMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt#L174) that is a JDK8-only API, most likely not supported by older versions of Android. 

It's easy to replace that call and we should do it. I'm also curious if there are ways to detect at build time such usages to make sure this does not happen again.

In the meantime, you can enable these API by using desugaring like described there: https://developer.android.com/studio/write/java8-support#library-desugaring

Edwhis:
Thanks, it worked!

PR #2517, opened 2020-08-10
[Kotlin Runtime] do not use Java8 forEach

martinbonnin:
Do not use `Map.forEach {}`, it doesn't work on some old Android devices without desugaring

fixes https://github.com/apollographql/apollo-android/issues/2516

I'm wondering whether we can:
1. force users to desugar
or
2. break at compile time for Java8 API usages.

### DIFF
--- a/apollo-runtime-kotlin/build.gradle.kts
+++ b/apollo-runtime-kotlin/build.gradle.kts
@@ -20,9 +20,7 @@ kotlin {
     }
   }
 
-  jvm {
-    withJava()
-  }
+  jvm()
 
   sourceSets {
     val commonMain by getting {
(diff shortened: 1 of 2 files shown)


apollo-runtime-kotlin/build.gradle.kts
```diff
@@ -20,9 +20,7 @@ kotlin {
     }
   }
 
-  jvm {
-    withJava()
-  }
+  jvm()
```
martinbonnin, line 23: `withJava()` is not required since there is no java source in this module.

martinbonnin:
> It is possible to add Android lint as a separate plugin to Kotili only projects. 
Ah, maybe that can do the trick! Do you have a link to the plugin? My googling always points me to applying the android plugin, which I don't really want to do.

Also, I modified a bit the PR to use Kotlin's `forEach`, that shouldn't use Java8 apis and is closer to the original intent.

tasomaniac:
Yeah, the destructuring of the arguments is nice. Tbh, it is too easy to make the mistake since it is only the parentheses. Ive seen it on ASG slack. It might be `com.android.lint`. unfortunately it is not documented.

PR #2535, opened 2020-08-23
Part 2 (backend): New Kotlin codegen

sav007:
Kotlin backend, transform generated AST to Kotlin code. Modified copy of old Kotlin code gen except generating code for fields with fragments.

Named fragment now generated as interfaces with default implementation.

Right now there is no tests as the PR is going to be insane. Tests for new generated models is going in next PR.


tasomaniac:
Maybe you could open the test PR already against this one to see if anything is wrong on the CI. It's not going to be easy to review 😃

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/codegen/Common.kt
```diff
@@ -0,0 +1,409 @@
...
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.asClassName
+import com.squareup.kotlinpoet.asTypeName
+import com.squareup.kotlinpoet.joinToCode
+
+internal fun CodeGenerationAst.FieldType.asTypeName(nullable: Boolean = false): TypeName {
+  return when (this) {
+    is CodeGenerationAst.FieldType.Scalar -> when (this) {
+      is CodeGenerationAst.FieldType.Scalar.ID -> ClassName.bestGuess(type)
```
tasomaniac, line 31: Importing these classes would make the code easier to read

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/codegen/Common.kt
```diff
@@ -0,0 +1,409 @@
...
+      .add(fields.map { field -> field.responseFieldInitializerCode }.joinToCode(separator = ",\n"))
+      .unindent()
+      .addStatement("")
+      .add(")")
+      .build()
+  return PropertySpec
+      .builder(
+          name = "RESPONSE_FIELDS",
+          type = Array<ResponseField>::class.asClassName().parameterizedBy(ResponseField::class.asClassName()),
+          modifiers = *arrayOf(KModifier.PRIVATE)
```
tasomaniac, line 91: Can this be just KModifier.PRIVATE

tasomaniac (approved):
I have just small comments. I reviewed it but not thoroughly.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/next/codegen/Common.kt
martinbonnin: If we do that, I wouldn't be against some kind of prefixing like `ASTFieldType`, `IRFieldType`, etc... Maybe `FieldType` isn't too much used in other places but we have 7 different `Field` classes and it's sometimes hard to get which layer is used when opening a new file.

tasomaniac: I now realized that this is in Common.kt. So my suggestion is not a good idea.

PR #2538, opened 2020-08-24
[SQLNormalizedCache] fix watcher not called on initial cache write

### LINKED ISSUE #2537: Query watcher with SQLNormalizedCache not triggering updates
**Summary**
The query watcher is not triggering any updates (institutesGraphQL) when I issue a query fetch (updateInstitutes). However if I restart my activity after a successful updateInstitutes the watcher will initially emit the previously fetched query, but again wont notify future changes for the remaining lifecycle of the activity. Also, shouldn't I be able to see the contents of the cache with the dump inside updateInstitutes?! Thanks in advance.

**Version**
Tested with 2.2.0 and 2.3.0

**Description**
I have the following in my viewmodel class
```kotlin
     private val institutesGraphQL =
        apolloClient.query(GetSupportedInstitutesQuery())
            .responseFetcher(ApolloResponseFetchers.CACHE_ONLY)
            .watcher()
            .toFlow().asLiveData()

     val institutesEntries = institutesGraphQL.switchMap {
        liveData {
            val tmpInstitutesEntries = ArrayList<String>()
            it.data?.let { data ->
                data.institutes!!.forEach { instGraphQl ->
                    tmpInstitutesEntries.add(instGraphQl.name)
                }
            }
            emit(tmpInstitutesEntries)
        }
    }
    
    fun updateInstitutes() {
        launchDataLoad {
            val response = apolloClient.query(GetSupportedInstitutesQuery())
                .responseFetcher(ApolloResponseFetchers.NETWORK_ONLY) // tested also with CACHE_AND_NETWORK and NETWORK_FIRST
                .toDeferred()
                .await()

//          Interestingly the following dump reports these in my console "OptimisticNormalizedCache {}, SqlNormalizedCache {}"
//          val dump = apolloClient.apolloStore.normalizedCache().dump()
//          println(NormalizedCache.prettifyDump(dump))

            println(response)
        }
    }

    private fun launchDataLoad(block: suspend () -> Unit): Job {
        return viewModelScope.launch {
            try {
                block()
            } catch (t: Throwable) {
                println(t)
            } finally {

            }
        }
    }
```
and tested with the following key resolver
```kotlin

      val resolver: CacheKeyResolver = object : CacheKeyResolver() {
        override fun fromFieldRecordSet(
            field: ResponseField,
            recordSet: Map<String, Any>
        ): CacheKey {
            // Retrieve the id from the object itself
            when (val type = recordSet["__typename"].toString()) {
                "Institute" -> {
                    return CacheKey.NO_KEY
                }
//             Tested also with the following
//              "Institute" -> {
//                 return CacheKey.from(type + "_" + recordSet["id"].toString())
//              }
            }
            throw Exception()
        }

        override fun fromFieldArguments(
            field: ResponseField,
            variables: Operation.Variables
        ): CacheKey {
            println("fromFieldArguments")
            return CacheKey.from(field.resolveArgument("id", variables) as String)
        }
    }
```

martinbonnin:
When no record was present in the cache, return all changed keys. 

fixes #2537

### DIFF
--- a/apollo-integration/build.gradle.kts
+++ b/apollo-integration/build.gradle.kts
@@ -11,6 +11,7 @@ dependencies {
   add("implementation", "com.apollographql.apollo:apollo-rx3-support")
   add("implementation", "com.apollographql.apollo:apollo-coroutines-support")
   add("implementation", "com.apollographql.apollo:apollo-http-cache")
+  add("implementation", "com.apollographql.apollo:apollo-normalized-cache-sqlite")
   add("implementation", "com.apollographql.apollo:apollo-compiler")
 
   add("testImplementation", kotlin("test-junit"))
@@ -50,6 +51,12 @@ configure<ApolloExtension> {
   service("performance") {
     sourceFolder.set("com/apollographql/apollo/integration/performance")
     rootPackageName.set("com.apollographql.apollo.integration.performance")
+  }
+}
+
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile::class.java) {
+  kotlinOptions {
+    freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
   }
 }
 
--- a/apollo-integration/src/test/java/com/apollographql/apollo/SqliteNormalizedCacheTest.kt
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/SqliteNormalizedCacheTest.kt
@@ -68,7 +68,7 @@ class SqliteNormalizedCacheTest {
       Truth.assertThat(responses[0].isFromCache).isEqualTo(false)
 
       withTimeout(1000) {
-        while(watchedResponses.isEmpty()) {
+        while(watchedResponses.size < 2) {
           delay(100)
         }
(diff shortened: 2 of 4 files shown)


martinbonnin:
> Should we have test for this?

Yes, 100%! I'll add one.

tasomaniac (approved):

apollo-integration/build.gradle.kts
```diff
@@ -11,6 +11,7 @@ dependencies {
   add("implementation", "com.apollographql.apollo:apollo-rx3-support")
   add("implementation", "com.apollographql.apollo:apollo-coroutines-support")
   add("implementation", "com.apollographql.apollo:apollo-http-cache")
+  add("implementation", "com.apollographql.apollo:apollo-normalized-cache-sqlite")
```
martinbonnin, line 14: It's nice to be able to test the SqliteCache without Android 😃

tasomaniac (approved):
Awesome tests.

apollo-integration/build.gradle.kts
tasomaniac: Definitely. It's a big win.

PR #2543, opened 2020-08-26
[Compiler] fix optional Int properties with an uppercase name

### LINKED ISSUE #2541: Optional QL properties in schema result in generated code that references the property incorrectly
**Note**: I am using Apollo Gradle Plugin 2.3.1

**Example 1:** Suppose I have a type called `VisitorCardIssueInput`. That type has an **optional** property called `workerIdVisiting`. The Kotlin class generated for it by Apollo looks like below. Note that the property `workerIdVisiting` starts with a lower case `w`. The reference to it that begins with `this@VisitorCardIssueInput` calls it `WorkerIdVisiting` - with an upper case 'w'. This code won't compile.

```
data class VisitorCardIssueInput(
  ...
  val workerIdVisiting: Input<Int> = Input.absent()
) : InputType {
    if (this@VisitorCardIssueInput.WorkerIdVisiting.defined) {
  ---------------------------------^ capital W - should be lower case w    
```

**Example 2:** Suppose I have the type `VisitorInput`. That type has an **optional** property called `visitorRawId` The Kotlin class generated by Apollo for `VisitorInput` looks like below. Note that the property `visitorRawId` starts with a lower case `v`. The reference to it that begins with `this@VisitorInput` calls it `VisitorRawId` - with an upper case `v`. This code won't compile.

```
  data class VisitorInput(
   ...
  val visitorRawId: Input<Int> = Input.absent()
) : InputType {
    if (this@VisitorInput.VisitorRawId.defined) {
 -------------------------^ capital V - should be lower case v   
```
Both generated classes are incredibly close to compiling. It just requires a change in capitalisation when referring to an optional field.

martinbonnin:
Fixes #2541 

The actual fix is really small. The reason the PR is long it that while adding tests, I moved the test schema to SDL (it's easier to edit and removes 10k lines 🤩  from the repo) and this took me to a rabbit hole of small SDL fixes (see comments inline).


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/InputType.kt
```diff
@@ -89,7 +89,7 @@ internal fun InputType.Field.writeCodeBlock(thisRef: String): CodeBlock {
       is FieldType.Scalar.Int -> {
         if (isOptional) {
           CodeBlock.builder()
-              .addStatement("if·(this@%L.%L.defined)·{", thisRef, schemaName)
+              .addStatement("if·(this@%L.%L.defined)·{", thisRef, name)
```
martinbonnin, line 92: This is the actual fix

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
```diff
@@ -17,38 +18,207 @@ import java.io.IOException
...
+  ""${'"'}
+  Indicates this type is an input object. `inputFields` is a valid field.
+  ""${'"'}
+  INPUT_OBJECT
+  ""${'"'}Indicates this type is a list. `ofType` is a valid field.""${'"'}
+  LIST
+  ""${'"'}Indicates this type is a non-null. `ofType` is a valid field.""${'"'}
+  NON_NULL
+}
+  """.trimIndent()
```
martinbonnin, line 221: Adding the built-in introspection types was required for the `introspection_query` codegen test to pass. More generally, users should be able to generate models for introspection types if needed.

I started adding these types programmatically but it becomes verbose real fast and we have a parser for that already so I put the builtin types as inline SDL

apollo-compiler/src/test/graphql/com/example/directive_with_inline_fragment/TestQuery.java
```diff
@@ -340,7 +340,7 @@ public Hero read(ResponseReader reader) {
         )),
         ResponseField.forFragment("__typename", "__typename", Arrays.<ResponseField.Condition>asList(
           ResponseField.Condition.booleanCondition("withDetails", false),
-          ResponseField.Condition.typeCondition(new String[] {"Human", "Droid"})
+          ResponseField.Condition.typeCondition(new String[] {"Droid", "Human"})
```
martinbonnin, line 343: The order change is unfortunate but I don't think it matters a lot. I think it is due to moving to SDL and the order types are parsed. Where would be the good place to enforce iteration order? When parsing the schemas? Generating code? Somewhere else?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
```diff
@@ -271,7 +445,7 @@ internal object GraphSDLSchemaParser {
       enumValue() != null -> enumValue().name().text
       listValue() != null -> listValue().value().map { it.parse() }
       objectValue() != null -> text
-      stringValue() != null -> text
+      stringValue() != null -> text.removePrefix("\"").removeSuffix("\"")
```
martinbonnin, line 448: I'm still not 100% sure about coercing rules for default values but I think the quotes are not needed here.

apollo-compiler/src/test/graphql/com/example/directive_with_inline_fragment/TestQuery.java
sav007: It really doesn't matter in the behaviour. We could sort them alphabetically to maintain the order but not sure if it worth doing

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
sav007: Hm, yeah `text` in this context will return string with quotes, cos:
```
stringValue
  : STRING
  | BLOCK_STRING
  ;

STRING: '"' CHARACTER* '"';
BLOCK_STRING:   '"""' .*? '"""';
```

sav007:
Oh, we are going in hell when will rebase `dev-3.x`. Will have conflicts a lot :(

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
sav007: Ha, very smart :) Nice move. Do you think if it worth to move this to separate file?

apollo-compiler/src/test/graphql/com/example/directive_with_inline_fragment/TestQuery.java
martinbonnin: Ok let's keep this for now.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
martinbonnin: Yes, I can make it a resource, this should work as long as we keep `apollo-compiler` JVM

martinbonnin:
> Oh, we are going to hell when rebase dev-3.x. Will have conflicts a lot :(

Currently I merge master into `dev-3.x` to avoid the rebases. But that won't change the fact that we will have conflicts. It's mainly going to be in tests outputs though, right?

PR #2545, opened 2020-08-27
Enable new code gen

sav007:
Switch `GraphQLCompiler` to use new code gen for Kotlin models that generates interfaces for fragments.

See test fixtures ([2nd commit](https://github.com/apollographql/apollo-android/pull/2545/commits/be165df26676e4a5fbf427e3d931070128883798)) with inline / named fragments for reference.


apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
```diff
@@ -126,19 +126,21 @@ data class TestQuery(
...
   ) {
-    fun marshaller(): ResponseFieldMarshaller = ResponseFieldMarshaller.invoke { writer ->
-      writer.writeString(RESPONSE_FIELDS[0], this@HeroWithReview.__typename)
-      writer.writeString(RESPONSE_FIELDS[1], this@HeroWithReview.name)
-      writer.writeDouble(RESPONSE_FIELDS[2], this@HeroWithReview.height)
+    fun marshaller(): ResponseFieldMarshaller {
+      return ResponseFieldMarshaller.invoke { writer ->
+        writer.writeString(RESPONSE_FIELDS[0], this@HeroWithReview.__typename)
+        writer.writeString(RESPONSE_FIELDS[1], this@HeroWithReview.name)
+        writer.writeDouble(RESPONSE_FIELDS[2], this@HeroWithReview.height)
```
martinbonnin, line 133: Not strictly related to this PR but the `ResponseFieldMarshaller` would be a good candidate to try out the shiny new `fun interface` in Kotlin 1.4

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
```diff
@@ -86,57 +89,381 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
...
+
+    companion object {
+      private val RESPONSE_FIELDS: Array<ResponseField> = arrayOf(
+        ResponseField.forString("__typename", "__typename", null, false, null)
+      )
+
+      operator fun invoke(reader: ResponseReader): Node1 {
+        val typename = reader.readString(RESPONSE_FIELDS[0])
+        return when(typename) {
+          "Person" -> PilotFragmentImpl(reader)
```
martinbonnin, line 302: Feature Request: if we know the type condition is always true, could we implement the interface by default? 

```
interface Node1: StarshipFragment.Node, PilotFragment
```

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
```diff
@@ -86,57 +89,381 @@ class AllStarships : Query<AllStarships.Data, AllStarships.Data, Operation.Varia
...
+    /**
+     * The name of this person.
+     */
+    override val name: String?,
+    /**
+     * A planet that this person was born on or inhabits.
+     */
+    override val homeworld: Homeworld?,
+    override val __typename: String = "Person"
+  ) : Node1, PilotFragment, StarshipFragment.Node {
```
martinbonnin, line 252: I'm still trying to wrap my head around this but here inheriting from `StarshipFragment.Node` is not strictly required as it's already a super interface of `Node1`.

martinbonnin:
This looks swift 🤩 . I left a few comments. Should we document the codegen principles (when do we generate interfaces vs data classes, how we handle name clashes, etc...)? I can work on something like this.

apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
sav007: yeah for sure, we can tune it later

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
sav007: Yeah makes sense such optimization

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
sav007: Could you pls elaborate on this, not sure I understand this feature request.

sav007:
@martinbonnin regarding the documentation, for sure we must provide extensive documentation for new codegen (we can take Airbnb doc provided by Ben as an example) as new codegen rules aren't trivial.

I think we should start working on doc after we stabilize and validate new codegen.

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarships.kt
martinbonnin: Yes, here `starship.pilotConnection.edges[0].node` is of GraphQL type `Person` but the generated type is `Node1`, meaning that users will always have to cast to access pilot fields while we know at compile time that they should always be there.

I'm not 100% sure how to generate this code but I feel like making "Node1" implement "PilotFragment" when Node1 is always of type "Person" would be handy. Of course if there are more possible types it wouldn't work anymore but I feel like the usage could be simplified for this specific case.

sav007: Agh right, nice catch. Now I have to think how to formalize this optimization in terms of codegen.

sav007: Done

sav007:
@martinbonnin could you please do another 👁️

martinbonnin (approved):
Looks good to me 👍 . I think we should merge it.

samples/kotlin-sample/src/main/java/com/apollographql/apollo/kotlinsample/data/ApolloCoroutinesService.kt
```diff
@@ -67,7 +67,7 @@ class ApolloCoroutinesService(
     job = CoroutineScope(processContext).launch {
       try {
         val response = apolloClient.query(commitsQuery).toDeferred().await()
-        val headCommit = response.data?.viewer?.repository?.ref?.target?.asCommit
+        val headCommit = response.data?.viewer?.repository?.ref?.target as GithubRepositoryCommitsQuery.Commit?
```
martinbonnin, line 70: Bumped into this while merging master. This feels more verbose than the previous solution. Maybe for inline fragments we can keep the `asXyz` helpers?

PR #2554, opened 2020-08-31
Kotlin Multiplatform Mobile integration for iOS targer

tasomaniac:
Put the location of the .xcodeproj file for Kotlin Multiplatform Mobile integration in Android Studio.

Once you Gradle Sync, you will have the following run configuration. I hit the run button and everything worked as expected. 

Note: You of-course need to run `./gradlew multiplatform:kmp-lib-sample:copyFramework` manually. Maybe we can automate that. 

Screen Shot 2020-08-31 at 19 30 02

### DIFF
--- a/composite/gradle.properties
+++ b/composite/gradle.properties
@@ -1,2 +1,4 @@
 org.gradle.jvmargs=-Xmx2g
 android.useAndroidX=true
+
+xcodeproj=../samples/multiplatform/kmp-ios-app/kmp-ios-app.xcodeproj


martinbonnin (commented):
Nice 🤩

martinbonnin (approved):

martinbonnin:
> You of-course need to run ./gradlew multiplatform:kmp-lib-sample:copyFramework manually. Maybe we can automate that.

Looks like KMM has a custom build phase that calls:

```
shellScript = "cd \"$SRCROOT/..\"\n./gradlew :shared:packForXCode -PXCODE_CONFIGURATION=${CONFIGURATION}\n";
```

Any idea why we have a custom gradle wrapper? It looks like it passes the java path to Xcode but KMM just assumes `java` is already present in the path?

designatednerd:
I think the custom gradle wrapper was the previously recommended setup - I haven't had time to dive into the new stuff but they definitely used to recommend that this get set up within an Xcode build phase.

PR #2569, opened 2020-09-10
[FileUpload] allow to sub class FileUpload 

### LINKED ISSUE #2392: Upload file with using content uri
Hello!

Starts with 29 API google deprecate work with the absolute path of files and insist to use content resolver with content uri. Current apollo upload files API give only one way - use files. So in our application for upload files, we should create a copy of the file in the cache directory and only then pass file path to FileUpload object (Only this way guarantee that file path will be correct).

Will be nice if for upload may be used contentUri, FileDescriptor, or InputStream.

Cheers!

martinbonnin:
This allows callers to use the `okio.Source` they want to upload files. With Android11 removing filestorage access, this will become mandatory.

Also took this opportunity to move some files to Kotlin.

fixes #2392


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -1,19 +1,37 @@
 package com.apollographql.apollo.api
 
-class FileUpload(val mimetype: String, val filePath: String) {
-  override fun equals(other: Any?): Boolean {
```
martinbonnin, line 4: I didn't find a need to implement equals/hashCode on this class. Am I missing something?

martinbonnin:
@darjke can you take a look and let me know if it works for you? I didn't explicitely use `contentUri` but the `FileUpload` now has a `writeTo(sink: BufferedSink)` so that implementation can decide where the data comes from. I feel it's more flexible and also it doesn't pull an Android dependency. Let me know your thoughts!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/FileUpload.kt
```diff
@@ -1,19 +1,37 @@
...
+  open fun writeTo(sink: BufferedSink) {
+    throw UnsupportedOperationException("ApolloGraphQL: if you're not passing a `filePath` parameter, you must override `FileUpload.writeTo`")
+  }
 
-    return true
+  /**
+   * The fileName to send to the server. Might be null
+   */
+  open fun fileName(): String? {
+    throw UnsupportedOperationException("ApolloGraphQL: if you're not passing a `filePath` parameter, you must override `FileUpload.fileName`")
```
sav007, line 31: Should we just remove `filePath` field and make this class abstract? Also make `filename` to be non null abstrac property?

martinbonnin: That would be much nicer but also a breaking change which I'd like to avoid if possible. I'll do it in dev-3x but given Android 11 is just out, I think it makes sense to release this faster.

sav007 (approved):

PR #2574, opened 2020-09-12
Deprecate toDeferred and toJob in favor of await.

### LINKED ISSUE #2571: Coroutines toDeferred extension doesn't respect structured concurrency
**Summary**
Apollo's client Coroutines Extensions doesn't respect structured concurrency when the parent's job is cancelled. 

**Version**
2.3.1

**Description**

I'm having a flow like this 

```
buttonClicks.mapLatest { buttonType ->
    val queryButtonTypeQuery = getQueryFromButtonType(buttonType)
    apolloClient.query(queryButtonTypeQuery).toDeferred().await()
}.onEach { response ->
    // Do something with response.
}.collect()
```

Let's assume I'm clicking buttons quickly, I'd expected OkHttp cancelling requests in the logs since `mapLatest` cancels the old emission in favor of the new emission. However that never happens. 

On the contrary when using `toFlow().first()` rather than toDeferred() I can see OkHttp cancelling requests in the logs.
`I/okhttp.OkHttpClient: <-- HTTP FAILED: java.io.IOException: Canceled`

On Kotlin's Slack channel @streetsofboston has suggested to remove `toDeferred` and introduce an `await()` function that respects structural concurrency through a `suspendCancellableCoroutine`

```
suspend fun <T> ApolloCall<T>.await(): Response<T> =
        suspendCancellableCoroutine { cont ->
            enqueue(
                object : Callback<T>() {
                        cont.resume(response)
                    }
                    override fun onFailure(e: ApolloException) {
                        cont.resumeWithException(e)
                    }
                }
            )
        }
```

R4md4c:
Closes #2571

I've removed the `toDeferred` and `toJob` APIs since they weren't respecting structured concurrency properly. I've introduced new `await` calls as discussed in the ticket.

~~I wanted to provide some tests, but I found out the necessary infrastructure isn't there yet, and it's out of scope of this PR.~~
EDIT: Found where the tests are, will add a few of them.

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -13,6 +13,10 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.suspendCancellableCoroutine
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.resume
 
 
 /**
@@ -73,12 +77,48 @@ fun <T> ApolloQueryWatcher<T>.toFlow(): Flow<Response<T>> = callbackFlow {
 
 
 /**
+ * Suspends the [ApolloCall] until it completes and returns the value on success or throws an exception on failure.
+ * The [ApolloCall] is cancelled when the coroutine running the operation is cancelled as well.
+ *
+ * This is a convenience method that will only return the first value emitted. If more than one
+ * response is required, for an example to retrieve cached and network response, use [toFlow] instead.
+ *
+ * @param <T>  the value type.
+ * @return the response on success.
+ * @throws ApolloException on failure.
+ */
+suspend fun <T> ApolloCall<T>.await(): Response<T> = suspendCancellableCoroutine { cont ->
+
+  cont.invokeOnCancellation {
+    cancel()
+  }
+
+  enqueue(object : ApolloCall.Callback<T>() {
+
+    private val wasCalled = AtomicBoolean(false)
+
+    override fun onResponse(response: Response<T>) {
+      if (!wasCalled.getAndSet(true)) {
+        cont.resume(response)
+      }
+    }
+
+    override fun onFailure(e: ApolloException) {
+      if (!wasCalled.getAndSet(true)) {
+        cont.resumeWithException(e)
+      }
+    }
+  })
+}
+
+/**
  * Converts an [ApolloCall] to an [Deferred]. This is a convenience method that will only return the first value emitted.
  * If the more than one response is required, for an example to retrieve cached and network response, use [toFlow] instead.
  *
  * @param <T>  the value type.
  * @return the deferred
  */
+@Deprecated("Use await() instead.")
 fun <T> ApolloCall<T>.toDeferred(): Deferred<Response<T>> {
   val deferred = CompletableDeferred<Response<T>>()
 
@@ -141,11 +181,44 @@ fun <T> ApolloSubscriptionCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
 }
 
 /**
+ * Suspends the [ApolloPrefetch] until it completes and returns the value on success or throws an exception on failure.
+ * The [ApolloPrefetch] is cancelled when the coroutine running the operation is cancelled as well.
+ *
+ * @param <T>  the value type.
+ * @return the response on success.
+ * @throws ApolloException on failure.
+ */
+suspend fun ApolloPrefetch.await(): Unit = suspendCancellableCoroutine { cont ->
+
+  cont.invokeOnCancellation {
+    cancel()
+  }
+
+  enqueue(object : ApolloPrefetch.Callback() {
+
+    private val wasCalled = AtomicBoolean(false)
+
+    override fun onSuccess() {
+      if (!wasCalled.getAndSet(true)) {
+        cont.resume(Unit)
+      }
+    }
+
+    override fun onFailure(e: ApolloException) {
+      if (!wasCalled.getAndSet(true)) {
+        cont.resumeWithException(e)
+      }
+    }
+  })
+}
+
+/**
  * Converts an [ApolloPrefetch] to [Job].
  *
  * @param <T>  the value type.
  * @return the converted job
  */
+@Deprecated("Use await() instead.")
 fun ApolloPrefetch.toJob(): Job {
   val deferred = CompletableDeferred<Unit>()
 
@@ -171,5 +244,4 @@ fun ApolloPrefetch.toJob(): Job {
 
   return deferred
 }
-
 
--- a/apollo-coroutines-support/api.txt
+++ b/apollo-coroutines-support/api.txt
@@ -2,6 +2,8 @@ package com.apollographql.apollo.coroutines {
 
   public final class CoroutinesExtensionsKt {
     ctor public CoroutinesExtensionsKt();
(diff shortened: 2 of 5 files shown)


recheej:
haha you beat me to it. I was just about to push my branch for this :)

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -73,35 +73,27 @@ fun <T> ApolloQueryWatcher<T>.toFlow(): Flow<Response<T>> = callbackFlow {
...
-      cancel()
-    }
-  }
+suspend fun <T> ApolloCall<T>.await(): Response<T> = suspendCancellableCoroutine { cont ->
   enqueue(object : ApolloCall.Callback<T>() {
     override fun onResponse(response: Response<T>) {
-      if (deferred.isActive) {
-        deferred.complete(response)
-      }
+      cont.resume(response)
```
recheej, line 86: You should have some kind of boolean to check if you already resumed once, otherwise an exception will be thrown second time the callback is called (I assume it's possible)

recheej (changes requested):

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
R4md4c: Thanks for your notice, I've adapted the code a bit, hope it's now

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -73,35 +73,33 @@ fun <T> ApolloQueryWatcher<T>.toFlow(): Flow<Response<T>> = callbackFlow {
...
+        cont.resume(response)
+        resumeIsCalled = true
       }
     }
 
     override fun onFailure(e: ApolloException) {
-      if (deferred.isActive) {
-        deferred.completeExceptionally(e)
-      }
+      cont.resumeWithException(e)
```
recheej, line 96: You should also guard here too I guess since you can possibly get multiple failures.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -141,35 +139,34 @@ fun <T> ApolloSubscriptionCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
...
-  deferred.invokeOnCompletion {
-    if (deferred.isCancelled) {
-      cancel()
-    }
-  }
+suspend fun ApolloPrefetch.await(): Unit = suspendCancellableCoroutine { cont ->
 
   enqueue(object : ApolloPrefetch.Callback() {
+
+    private var resumedIsCalled = false
```
recheej, line 153: I'm not familiar with this library tbh, but reading the documentation for `ApolloPrefetch` says it should only be called back once so you shouldn't need this. Then again adding the check doesn't hurt. I'll let library maintainers say whether this can be called multiple times or not.

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
recheej: (or one success and one failure)

R4md4c: From the documentation it says `onFailure` can be called at most one time, so I don't think another guard is needed here. 
https://github.com/apollographql/apollo-android/blob/9513025a9640517f0fa448c9859938106c50606a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java#L103

recheej: Ah ok got it. What about the one failure and n successes? Can that happen?

R4md4c: The documentation says, that onFailure is a terminal that means if it happens, onSuccess will never be called, however I'm not sure if we can get onSuccess prior the onFailure or not. Better to hear from the maintainers in that case.

recheej: Got it thanks for clarification! I'll leave up to to you guys who know more

recheej (approved):

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
martinbonnin: `onSuccess` will be called several time if using the cache (one for the cache response, another for the network response for an exemple). If fetching from the network fails, there will be one `onSuccess` and one `onFailure` call. 

Users needing to use cache responses should really use`toFlow` instead but `await()` should not crash if that ever happens.

Also, I think there's no guarantee what thread the callbacks are called in so `resumeIsCalled` would have to be thread-safe. Maybe something like:

```kotlin
resumeIsCalled = AtomicBoolean()
if (!resumeIsCalled.getAndSet(true)) {
   cont.resume(response)
}
```

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
martinbonnin: I'm not 100% sure either. I'd vote to play it safe and keep the guard there.

martinbonnin:
Thanks a lot for the contribution! I made a few comments up there, let me know what you think.

Also, removing `toDeferred()` is unfortunately not possible at this point since a lot of users might still be relying on it and we want to provide proper deprecation periods. Can you add it back? We'll remove it for the next major version.

R4md4c:
@martinbonnin I've readded the old APIs, and put a deprecation notice on them, please let me know if I need to do additional stuff.

martinbonnin (approved):
Thanks!

apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
```diff
@@ -72,13 +76,49 @@ fun <T> ApolloQueryWatcher<T>.toFlow(): Flow<Response<T>> = callbackFlow {
...
+ */
+suspend fun <T> ApolloCall<T>.await(): Response<T> = suspendCancellableCoroutine { cont ->
+
+  cont.invokeOnCancellation {
+    cancel()
+  }
+
+  enqueue(object : ApolloCall.Callback<T>() {
+
+    private val wasCalled = AtomicBoolean(false)
```
sav007, line 98: is there any reason to not use `cont.isCancelled`? There was an issue with RC but looks like it was fixed https://github.com/Kotlin/kotlinx.coroutines/issues/830

R4md4c: This boolean was introduced to guard against calling `resume` multiple times to prevent an exception getting thrown for resuming with multiple values instead of one. Same below as. So I'm not sure how `isCancelled ` would solve this particular case.

sav007: Sorry, my bad. I misinterpreted the description of the issue.

sav007 (approved):

PR #2575, opened 2020-09-14
[Plugin] Add an option to fail on deprecated usages

### LINKED ISSUE #2489: [Gradle plugin] add an option to fail compilation if a deprecated field is used
To improve awareness of when deprecated schema fields are used, add a configuration option to the plugin to make the build fail:

```kotlin
apollo {
  failOnDeprecatedFieldUsage.set(true) // defaults to false
}
```

On failure, display the file and line where the field is used.

martinbonnin:
Add 

```
apollo {
  warnOnDeprecatedUsages.set(true) // true by default
  failOnWarnings.set(true) // false by default
}
```

Fixes #2489 

Using the `w: $file: ($line, $column): $message` format, intelliJ will display it nicely in the "Run" panel:

Screenshot 2020-09-14 at 11 57 09

Somehow the warnings are not clickable in the "Run" panel but are in the build output, not really sure how to make IntelliJ pick them up. 

Also, this PR only checks deprecated fields usages. Checking for deprecated enums could be done too but that's more complex so will be done in a future PR.

Ping @kenyee, sorry it took so long, I wanted to get the multi-module PR in first to avoid too many conflicts.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
@@ -63,7 +63,7 @@ class GraphQLCompiler(val logger: Logger = NoOpLogger) {
         // antlr is 0-indexed but IntelliJ is 1-indexed. Add 1 so that clicking the link will land on the correct location
         val column = it.sourceLocation.position + 1
         // Using this format, IntelliJ will parse the warning and display it in the 'run' panel
-        logger.warning("w: ${it.filePath}:${it.sourceLocation.line}:${column}: ApolloGraphQL: Use of deprecated field '${it.field.responseName}'")
+        logger.warning("w: ${it.filePath}:${it.sourceLocation.line}:${column}: ApolloGraphQL: Use of deprecated field '${it.field.fieldName}'")
       }
       if (args.failOnWarnings && deprecatedUsages.isNotEmpty()) {
         throw IllegalStateException("ApolloGraphQL: Warnings found and 'failOnWarnings' is true, aborting.")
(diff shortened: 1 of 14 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -139,6 +160,31 @@ class GraphQLCompiler {
...
+  private fun DocumentParseResult.collectDeprecatedUsages(): List<DeprecatedUsage> {
+    return operations.flatMap { it.fields.collectDeprecatedUsages(it.filePath) } +
+        fragments.flatMap { it.fields.collectDeprecatedUsages(it.filePath) }
+  }
+
+  /**
+   * walk the list and return any deprecated fields
+   * TODO: add support for deprecated enums
+   */
+  private fun List<Field>.collectDeprecatedUsages(filePath: String): List<DeprecatedUsage> {
```
tasomaniac, line 174: I wonder if this can be done via `tailrec fun` in Kotlin.  If possible, Kotlin allows recursion without the possibility of stack overflow since it optimizes the bytecode.

Im just curious. And the end, it might look slightly better. That's all 😊

tasomaniac (approved):
Thanks for working on this. It's a great addition.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
martinbonnin: I'm not sure this maps well as a tail rec function. Each Field node has multiple Field children so there would have to be several recursive calls at each depth. The only way I can see to make it tailrec is to pass the `fieldsToVisit` as some state:

```
tailrec fun collectDeprecatedUsages(fieldsToVisit: MutableList<Field>, filePath: String, deprecatedUsages: MutableList<DeprecatedUsage>)
```

But that would add an additional method, plus we don't use tailrec anywhere else in the codebase so not sure it's worth it?

tasomaniac: Make sense. I didn't really see that it is not possible with a single fun. Don't mind me on this one 😊

tasomaniac (approved):

PR #2586, opened 2020-09-16
[Plugin] Allow to download SDL schemas

### LINKED ISSUE #2582: [Gradle Plugin] Allow to download the schema as SDL
Currently, the `downloadApolloSchema` task download the schema with introspection and saves it as a Json. Since we now support SDL schemas, we could transform them to SDL before saving them to a file. Since the schema is checked in source control, that will make smaller repositories. It will also make it easier to read the schema for troubleshooting.

martinbonnin:
closes #2582 

Initial support for downloading SDL schemas from an introspection endpoint.

The code generator is relatively barebones, maybe we could make it smarter to handle better indentation, line breaks, etc...

Usage:

```
#the plugin will detect the `.sdl` extension and convert on the fly if that's the case
./gradlew downloadApolloSchema 
     --endpoint=https://apollo-fullstack-tutorial.herokuapp.com/graphql 
     --schema=src/main/graphql/com/library/schema.sdl 
```

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/moshi.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/moshi.kt
@@ -74,3 +74,9 @@ inline fun <reified T> T.toJson(file: File) {
     return getJsonAdapter<T>().toJson(it, this)
   }
 }
+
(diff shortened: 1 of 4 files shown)


sav007 (approved):
wow so simple.

PR #2592, opened 2020-09-21
[Compiler] Handle lists in arguments

### LINKED ISSUE #2589: [Bug] Incomplete query cache key
**Summary**
"RESPONSE_FIELDS" in "Content.Data" class of generated GetContentQuery.kt file contains wrong value for "_or" and "_and" expressions.
As a result, queries with different parameters get wrong cache values and overwrite each other

**Version**
2.3.1

**Description**
Query request should contain "_and" or "_or" expressions with input parameters in it.

Example query:
```graphql
query getContent(
  $vendors: [Int!],
  $tags_filter_1: [Int!],
  $tags_filter_2: [Int!]
) {
  content(
    where:{
      state: {_eq: "available"},
      vendor_id: {_in: $vendors},
      _and : [
        { recipe_tags : { tag_id: {_in: $tags_filter_1}}},
        { recipe_tags : { tag_id: {_in: $tags_filter_2}}}
      ]
    }
  ) {id}
}
```

Generated code:
```kotlin
    ResponseField.forList("content", "content", mapOf<String, Any>(
        "where" to mapOf<String, Any>(
            "_and" to
            "[{recipe_tags={tag_id={_in={kind=Variable, variableName=tags_filter_1}}}}, {recipe_tags={tag_id={_in={kind=Variable, variableName=tags_filter_2}}}}]",            
            "vendor_id" to mapOf<String, Any>(
                "_in" to mapOf<String, Any>(
                    "kind" to "Variable",
                    "variableName" to "vendors"))
        )
    )))
```

The content of "_and" tag is not parsed.

This data is used by [RealCacheKeyBuilder.kt ](https://github.com/apollographql/apollo-android/blob/8e60deee1ce4bb6a54f561921e78579fa2ccaa2f/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt#L34)
for generating unique query key to get valid data from cache.

In current case, cache key generator skip variables in  "_and" block, and queries with different params (tags_filter_1, tags_filter_2) generate the same cache key.

**Solution**
Code generator should expand "_and" field value.
Temporary solution is to edit generated code in project this way:
```
"_and" to mapOf<String, Any>(
    "content_tags1" to mapOf<String, Any>(
        "_in" to mapOf<String, Any>(
        "kind" to "Variable",
        "variableName" to "tags_filter_1")),
    "content_tags2" to mapOf<String, Any>(
        "_in" to mapOf<String, Any>(
        "kind" to "Variable",
        "variableName" to "tags_filter_2"))
)
```

In future versions, should be added possibility to use List<Pair<String,Any>>
Plugin should generate code like this:
```
"_and" to listOf<Pair<String, Any>>(
    "content_tags" to mapOf<String, Any>(
        "_in" to mapOf<String, Any>(
        "kind" to "Variable",
        "variableName" to "tags_filter_1")),
    "content_tags" to mapOf<String, Any>(
        "_in" to mapOf<String, Any>(
        "kind" to "Variable",
        "variableName" to "tags_filter_2"))
)
```

martinbonnin:
Generate Kotlin `Lists` for GraphQL lists arguments so that `RealCacheKeyBuilder` can traverse them and replace variables when needed.

For an exemple, the below query would not replace the `tags_filter_1` and `tags_filter_2` variables:

```graphql
query getContent(
  $vendors: [Int!],
  $tags_filter_1: [Int!],
  $tags_filter_2: [Int!]
) {
  content(
    where:{
      state: {_eq: "available"},
      vendor_id: {_in: $vendors},
      _and : [
        { recipe_tags : { tag_id: {_in: $tags_filter_1}}},
        { recipe_tags : { tag_id: {_in: $tags_filter_2}}}
      ]
    }
  ) {id}
}
```

closes #2589


apollo-compiler/src/test/graphql/com/example/arguments_complex/TestQuery.kt
```diff
@@ -170,21 +171,34 @@ data class TestQuery(
...
+            ),
             "review" to mapOf<String, Any>(
               "stars" to mapOf<String, Any>(
                 "kind" to "Variable",
-                "variableName" to "stars"),
+                "variableName" to "stars"
+              ),
               "favoriteColor" to mapOf<String, Any>(
-                "red" to "0",
+                "red" to 0,
```
martinbonnin, line 182: I also took this opportunity to map GraphQL's Int and Float to Kotlin's Int and Double for better typing. While I believe it's more correct, it's a behaviour change so I'm happy to revert if we think this is risky.

martinbonnin: Thinking more about this, this shouldn't change what's sent to the server (which is in `QUERY_DOCUMENT`) so the impact would only be for the cache. So impact is still limited but that could trigger some unwanted cache miss so I'm going to revert that.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
```diff
@@ -29,34 +29,39 @@ class RealCacheKeyBuilder : CacheKeyBuilder {
...
+      is Map<*, *> -> {
+        value as Map<String, Any?>
+        if (isArgumentValueVariableType(value)) {
+          resolveVariable(value, variables)
         } else {
-          resolveArguments(nestedObjectMap, variables)
+          value.mapValues {
+            resolveVariables(it.value, variables)
+          }.toList()
+              .sortedBy { it.first }
```
moshkit, line 43: toSortedMap() fits also
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-sorted-map.html

moshkit (approved):

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
martinbonnin: I wish it worked but unfortunately `toSortedMap()` is JVM only. At this point this is mainly rhetorical because there's no real way to consume the normalized cache from anything else than JVM but there will be a native cache in the future.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
```diff
@@ -29,34 +29,39 @@ class RealCacheKeyBuilder : CacheKeyBuilder {
...
   }
 
   @Suppress("UNCHECKED_CAST")
-  private fun resolveVariableArgument(objectMap: Map<String, Any?>, variables: Operation.Variables): Any? {
+  private fun resolveVariable(objectMap: Map<String, Any?>, variables: Operation.Variables): Any? {
     val variable = objectMap[ResponseField.VARIABLE_NAME_KEY]
 
     return when (val resolvedVariable = variables.valueMap()[variable]) {
-      null -> null
-      is Map<*, *> -> resolveArguments(resolvedVariable as Map<String, Any?>, variables)
```
sav007, line 55: sanity check: why these lines were removed?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
```diff
@@ -29,34 +29,39 @@ class RealCacheKeyBuilder : CacheKeyBuilder {
...
       is InputType -> {
         val inputFieldMapWriter = SortedInputFieldMapWriter()
         resolvedVariable.marshaller().marshal(inputFieldMapWriter)
-        resolveArguments(inputFieldMapWriter.map(), variables)
```
sav007, line 59: same here: don't really remember why but looks like this launch recursion for nested map arguments resolution

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
sav007: nvm, I found that it's handled above

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
sav007: Looks like it doesn't make any sense for recursion here at first place. If `resolvedVariable` is `InputType` there is no way it has reference to other variables.

sav007 (approved):

sav007:
We need to carry this fix over to our `dev-3.x` branch.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/RealCacheKeyBuilder.kt
martinbonnin: Yes, exactly !

martinbonnin:
> We need to carry this fix over to our dev-3.x branch.

I'll merge once https://github.com/apollographql/apollo-android/pull/2604 is in

martinbonnin:
Released in [2.4.1](https://github.com/apollographql/apollo-android/releases/tag/v2.4.1) (also merged in `dev-3.x` as part of https://github.com/apollographql/apollo-android/pull/2625)

PR #2603, opened 2020-09-23
[Runtime] add a client awareness OkHttp interceptor

martinbonnin:
Add `ApolloClientAwarenessInterceptor` to implement [Client Awareness](https://www.apollographql.com/docs/studio/client-awareness/#advanced-apollo-server-configuration).

Usage:

```
    val okHttpClient = OkHttpClient.Builder()
        .addInterceptor(ApolloClientAwarenessInterceptor(BuildConfig.VERSION_NAME, BuildConfig.APPLICATION_ID))
        .build()

    instance = ApolloClient.builder()
        .serverUrl("https://apollo-fullstack-tutorial.herokuapp.com/graphql")
        .okHttpClient(okHttpClient)
        .build()
``` 

I'd like to make the setup even easier but all approaches I can think of have major drawbacks:

* `fun ApolloClient.Builder.clientAwareness(context: Context): ApolloClient.Builder` to get name/version automagically from context is not possible because `apollo-runtime` doesn't know about Android `Context`
* `fun ApolloClient.Builder.clientAwareness(clientName: String, clientVersion: String): ApolloClient.Builder` is dangerous because adding an Interceptor for authentication is a super widely used use case and I prefer the calling code to be in control of the interceptors (for order and more generally it's more consistent than having your okHttpInstance modified under your feet).

Ultimately, I think it's ok to add the interceptor manually if we add the appropriate documentation page. If anyone has any ideas how to make this process easier, I'll take them

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClientAwarenessInterceptor.kt
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClientAwarenessInterceptor.kt
@@ -1,0 +1,20 @@
+package com.apollographql.apollo
+
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+ * An [Interceptor] to add [Client Awareness](https://www.apollographql.com/docs/studio/client-awareness/).
+ *
+ * Add this interceptor to your [okhttp3.OkHttpClient] instance when creating your [ApolloClient]
+ */
+class ApolloClientAwarenessInterceptor(val clientName: String, val clientVersion: String): Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val newRequest = chain.request().newBuilder()
+        .addHeader("apollographql-client-name", clientName)
+        .addHeader("apollographql-client-version", clientVersion)
+        .build()
+
+    return chain.proceed(newRequest)
+  }
+}
--- a/apollo-runtime/api.txt
+++ b/apollo-runtime/api.txt
@@ -94,6 +94,13 @@ package com.apollographql.apollo {
     method public com.apollographql.apollo.ApolloClient.Builder useHttpGetMethodForQueries(boolean);
   }
 
+  public final class ApolloClientAwarenessInterceptor {
+    ctor public ApolloClientAwarenessInterceptor(error.NonExistentClass, error.NonExistentClass);
+    method public error.NonExistentClass getClientName();
+    method public error.NonExistentClass getClientVersion();
+    method public error.NonExistentClass intercept(error.NonExistentClass);
+  }
+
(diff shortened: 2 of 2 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClientAwarenessInterceptor.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+ * An [Interceptor] to add [Client Awareness](https://www.apollographql.com/docs/studio/client-awareness/).
+ *
+ * Add this interceptor to your [okhttp3.OkHttpClient] instance when creating your [ApolloClient]
+ */
+class ApolloClientAwarenessInterceptor(val clientName: String, val clientVersion: String): Interceptor {
```
tasomaniac, line 11: If we have Context somehow, `Context.packageName` is the client name I guess. It is the same as Build config.APLICATION_ID.

We can then use the same packageName to retrieve the version from packageManager.

Got this from [here](https://stackoverflow.com/questions/4616095/how-to-get-the-build-version-number-of-your-android-application/6593822#6593822)

```
try {
    PackageInfo pInfo = context.getPackageManager().getPackageInfo(getPackageName(), 0);
    String version = pInfo.versionName;
} catch (PackageManager.NameNotFoundException e) {
    e.printStackTrace();
}
```

tasomaniac: Can we not put this via android-support module? Or something like this so that we access Context.

Or I had the idea to use AndroidX startup library to retrieve the Application context and store it statically. It's pretty safe because there is only 1 application Context.

martinbonnin: > Can we not put this via android-support module? 

That would require adding a new dependency in `build.gradle[.kts]`, not sure this is easier than adding an interceptor. I suspect it's a common usage to have an interceptor already setup for authentication so adding the ApolloClientAwerenessInterceptor should be a single line in that case.

Also the functionnality should be available for non-Android users as well, which would mean splitting it between `apollo-android-support` (for default Android values) and `apollo-runtime` (for the general case). All in all it felt like a lot of hassle for something that's easy to explain with a bit of documentation.

sav007:
I think this manual process make a lot of sense, as `clientName` doesn't mean it must be `Context.packageName` and it's up to user to provide whatever name and version. Also are these `ClientAwareness` headers optional from apollo server perspective?

martinbonnin:
>  are these ClientAwareness headers optional from apollo server perspective?
Yes, they are completely optional

sav007 (approved):

PR #2604, opened 2020-09-23
Fragment codegen optimizations

sav007:
- Fix issue with named fragment interface generation, when we generate bunch of unnecessary / unused class implementations see [fragment_in_fragment test fixture](https://github.com/apollographql/apollo-android/pull/2604/files#diff-c1defd180631339ff8ffa20d3ccef9f7)
- In case when there is only one fragment and its type condition aligned with the type of the field it's spread on, don't generate implementation classes but rather merge fields see [root_query_fragment test fixture](https://github.com/apollographql/apollo-android/pull/2604/files#diff-b738620bba0e8bca52e165f2aca603ed)
- For fragments defined on concrete type that spread in selection set without additional fields except `__typename` generate class that delegated to that fragment default implementation see [fragments_with_type_condition test fixture](https://github.com/apollographql/apollo-android/pull/2604/files#diff-835ec7b5ec2dbc6dd046f689bf05c3d4)


martinbonnin:
This is looking great 🤩 . A few comments: 

> Fix issue with named fragment interface generation, when we generate bunch of unnecessary / unused class implementations

👍 

> In case when there is only one fragment and its type condition aligned with the type of the field it's spread on, don't generate implementation classes but rather merge fields

What is the reason to limit this case to to only one fragment? I believe the two fragments below could be merged as well?

```
query TestQuery {
  ...QueryFragment1
  ...QueryFragment2
}
```

Somewhat related I think there's something wrong if multiple fragments are defined on the same type:

https://github.com/martinbonnin/apollo-android/blob/f1105a2c9a65becc2217d87abfef6bc10900d62e/apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/TestQuery.kt#L236

For this to work with multiple `Human` fragments for an example, it should be something like:

```
        val typename = reader.readString(RESPONSE_FIELDS[0])
        return when(typename) {
          "Human" -> HumanLukeImpl(reader) // not HumanDetailsImpl1, which is bound to a single fragment
          "Droid" -> DroidLukeImpl(reader) // not DroidDetailsImpl1, which is bound to a single fragment
          else -> LukeImpl(reader)
        }
```

> For fragments defined on concrete type that spread in selection set without additional fields except __typename generate class that delegated to that fragment default implementation

What is the rationale behind that? From what I can see the number of classes will be the same? But it factors out the parsing code leading to less bytecode? Am I right?

martinbonnin:
Also, I pushed a small improvement to speed up testing. I hope you don't mind. We can now type:

```
# runs a single codegen test
./gradlew :apollo-compiler:test -DcodegenTests=".*fragments_with_type_condition.*Kotlin" --tests '*Codegen*'
```

sav007:
>What is the reason to limit this case to to only one fragment? I believe the two fragments below could be merged as well?

Yeah you are right, will make changes to support this case.

>>For fragments defined on concrete type that spread in selection set without additional fields except __typename generate class that delegated to that fragment default implementation
>What is the rationale behind that? From what I can see the number of classes will be the same? But it factors out the parsing code leading to less bytecode? Am I right?

The number of classes will be the same only if fragment doesn't have nested classes. But if named fragment has a deep structure of nested classes, we won't generate implementations for those nested classes. I've added test fixture that demonstrates this [see](https://github.com/apollographql/apollo-android/pull/2604/files#diff-01d48f2e83e7207396935fdfdf458ffeR89-R110). Notice that not in query we don't have `Friend`, `FriendsConnection`, `Edges`, `Node`. We just delegate to default impl.

>https://github.com/martinbonnin/apollo-android/blob/f1105a2c9a65becc2217d87abfef6bc10900d62e/apollo-compiler/src/test/graphql/com/example/fragments_with_type_condition/TestQuery.kt#L236

Not sure that I understand what do you mean we have this:
```
interface Luke {
      operator fun invoke(reader: ResponseReader): Luke {
        val typename = reader.readString(RESPONSE_FIELDS[0])
        return when(typename) {
          "Human" -> HumanDetailsImpl1(reader)
          "Droid" -> DroidDetailsImpl1(reader)
          else -> LukeImpl(reader)
        }
      }
    }
  }

where:
 - HumanDetailsImpl1: Luke, HumanDetails
 - DroidDetailsImpl1: Luke, DroidDetails


interface R2 {
      operator fun invoke(reader: ResponseReader): R2 {
        val typename = reader.readString(RESPONSE_FIELDS[0])
        return when(typename) {
          "Human" -> HumanDetailsImpl(reader)
          "Droid" -> DroidDetailsImpl(reader)
          else -> R2Impl(reader)
        }
      }
    }
  }

where:
 - HumanDetailsImpl1: R2, HumanDetails
 - DroidDetailsImpl1: R2, DroidDetails

```

Looks like legit to me.

martinbonnin:
> The number of classes will be the same only if fragment doesn't have nested classes. But if named fragment has a deep structure of nested classes, we won't generate implementations for those nested classes in query class

Got it, Thanks👍 

> Not sure that I understand what do you mean we have this:

Imagine the following case:

```
query TestQuery {
  luke: hero {
    ...HumanDetails
    ...Appearances
    ...DroidDetails
  }
}

fragment HumanDetails on Human {
  name
  height
}

fragment Appearances on Human {
  appearsIn
}

fragment DroidDetails on Droid {
  name
  primaryFunction
}

```

When the type is "Human", we want an implementation that implements `Luke`, `HumanDetails` and `Apparances` so it can't just be `HumanDetailsImpl`, it certainly has to be something more generic like `HumanLukeImple` ?

sav007:
Agh right, we did have this before optimization but now it's broken. Nice catch!!!! I will fix this.

sav007:
@martinbonnin  back to your question after some thinking:
> What is the reason to limit this case to to only one fragment? I believe the two fragments below could be merged as well?

The reason that even 2 fragments defined on the same type we can't just merge them as the selection set could be different, the condition on fields could be different.

martinbonnin (approved):
Looks good! 
I moved the integration tests to the new codegen in #2625 and it worked well 👍 . 
Something I was thinking while doing the migration is that we could add helpers to discover fragments instead of having to cast:

```kotlin
fun Friend.humanWithIdFragment() = this as? HumanWithIdFragment
```

This could:
* make it easier to discover the variant fragments with IntelliJ auto-complete
* avoid breaking the typing flow to go backwards to type the opening parenthesis when casting.

Can definitely be added later given the size of this PR already :)

sav007:
>fun Friend.humanWithIdFragment() = this as? HumanWithIdFragment

Yes that I was planning to start working on next

Issue #2612, opened 2020-09-28
The number of operation IDs (20) should match the number of operations (21). Check that all your IDs are unique.


piotr-pawlowski:
**Summary**
After bump version from 2.2.0 to 2.4.0 I receive that error while compilation: `The number of operation IDs (20) should match the number of operations (21).
Check that all your IDs are unique.`

**Version**
2.4.0

martinbonnin:
Thanks for the report! This would happen if there's somehow a duplicate `operationId` or if two queries have the same queryDocument. 

A few questions to help diagnose the issue:
Do you use a custom `OperationIdGenerator` ?
Can you check your queries, would two of them be the same?

piotr-pawlowski:
I do not use a custom `OperationIdGenerator`.

> Can you check your queries, would two of them be the same?

Do you mean the same name?

martinbonnin:
> Do you mean the same name?

Same query text, something like a duplicate .graphql file somewhere or something similar. The operationId is by default computed from the sha1 of the query text (query document, what is sent to the server)

https://github.com/apollographql/apollo-android/pull/2476 enforced that there's no duplicate. It's possible that this was ignored before but I don't thing having duplicate queries is desirable. Unless I'm missing a use case?

The other option would be a sha1 collision but I doubt that very much

piotr-pawlowski:
Ok, I found two the same query in project. Thanks for quick response!

martinbonnin:
Glad to hear that, I'll try to make that error message more explicit, I think we can make it better.

martinbonnin:
Been looking a bit into this and it turns out we already have [code to detect duplicates](https://github.com/apollographql/apollo-android/blob/d3cb1f89adfd5408c98f742df465d8ca02732868/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/Validation.kt#L16) but duplicate queries with different package names could be theoretically allowed, which is why I believe is why your case wasn't caught. Maybe we can relax the format of `operationOutput.json` to allow duplicate queries...

PR #2613, opened 2020-09-28
[Plugin] Write operationOutput.json even if there are duplicateIds

martinbonnin:
This will help debugging. See #2612


martinbonnin:
Closing for now as it turns out we can have duplicate queries in different package names. Will keep investigating

PR #2617, opened 2020-09-29
[Gradle Plugin] add registry download

martinbonnin:
Allow to download a schema from the Apollo registry with:

```
./gradlew downloadApolloSchema --graph $APOLLO_GRAPH --key $APOLLO_KEY --schema schema.[json | sdl]
```

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
@@ -94,7 +94,7 @@ object SchemaDownloader {
       }
     """.trimIndent()
 
-    val response = executeQuery(query, variables, "https://engine-graphql.apollographql.com/api/graphql", mapOf("x-api-key" to key))
+    val response = executeQuery(query, variables, "https://graphql.api.apollographql.com/api/graphql", mapOf("x-api-key" to key))
 
     val responseString = response.body.use { it?.string() }
 
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo.gradle.internal
 
+import com.apollographql.apollo.compiler.fromJson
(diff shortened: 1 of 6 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
```diff
@@ -39,30 +44,56 @@ object SchemaDownloader {
...
+  }
 
-    response.body.use { responseBody ->
-      if (schema.extension.toLowerCase() == "json") {
-        schema.writeText(responseBody!!.string())
-      } else {
-        IntrospectionSchema(responseBody!!.byteStream()).toSDL(schema)
+  fun downloadRegistry(graph: String, key: String, variant: String): String? {
+    val registryQuery = """
+    query {
```
zionts, line 75: Could we make sure to name the operation here? Additionally, using variables is a best practice, along with identifying your client traffic using the `apollographql-client-name` header.

Maybe something like this?

```
query DownloadSchema($graphID: ID!, $variant: String! = "current") {
  service(id: $graphID) {
    variant(name: $variant) {
      activeSchemaPublish {
        schema {
          document
        }
      }
    }
  }
}
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaDownloader.kt
```diff
@@ -39,30 +44,56 @@ object SchemaDownloader {
...
+      service(id: "$graph") {
+        schema(tag: "$variant") {
+          document
+        }
       }
     }
+  """.trimIndent()
+
+    val response = executeQuery(registryQuery, "https://engine-graphql.apollographql.com/api/graphql", mapOf("x-api-key" to key))
```
zionts, line 84: can we add in an `apollographql-client-name` header? This will help us make sure that we're tracking this operation back to this particular client so if there are any changes that might affect it down the road, we can reach out :) https://www.apollographql.com/docs/studio/client-awareness/

martinbonnin: Sounds good 👍 . I also added `apollographql-client-version` there just in case.

martinbonnin: Aslo, I set the client name to `apollo-gradle-plugin`, which is the name of the maven artifact. Let me know how that sounds.

Issue #2616, opened 2020-09-29
[Compiler] 3.x Java Codegen


martinbonnin:
The codegen is undergoing significant changes in 3.x to support fragments as interfaces (#1854) and a streaming parser (#2523). 3.x will have Kotlin-first API and therefore efforts for the new codegen focus on Kotlin code generation, leaving Java codegen behind.

This issue is to add Java codegen back once 3.x APIs are stabilized.

martinbonnin:
Fixed in `3.0.0-alpha04`

PR #2625, opened 2020-10-04
Massive Fall Cleanup 🍁🍃

martinbonnin:
Very sorry for the huge PR, there's no real other solution. This PR:
* removes the java codegen (see #2616)
* makes the Gradle plugin tests and integration tests use Kotlin models
* convert integration tests source code to Kotlin 
* disables metalava checks while we tweak the APIs

~Don't merge yet, it's based on https://github.com/apollographql/apollo-android/pull/2604 and should be merged after.~


sav007:
Should we rebase it as #2604 was merged?

martinbonnin:
> Should we rebase it as #2604 was merged?
Will do it as soon as I'm done with releasing 2.4.1

martinbonnin:
> Will do it as soon as I'm done with releasing 2.4.1

It's rebased

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
```diff
@@ -304,34 +304,6 @@ open class ResponseField internal constructor(
...
-     *
-     * @param responseName alias for the result of a field
-     * @param fieldName name of the field in the GraphQL operation
-     * @param arguments arguments to be passed along with the field
-     * @param optional whether the arguments passed along are optional or required
-     * @param conditions list of conditions for this field
-     * @return Field instance representing [Type.LONG]
-     */
-    @JvmStatic
-    fun forLong(
```
tasomaniac, line 318: Since these are in the api, is it fine to remove without deprecating them first?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -189,13 +163,13 @@ class GraphQLCompiler(val logger: Logger = NoOpLogger) {
   private fun idClassName(generateKotlinModels: Boolean) = if (generateKotlinModels) {
     String::class.asClassName().toString()
   } else {
-    ClassNames.STRING.toString()
+    TODO("ClassNames.STRING.toString()")
```
tasomaniac, line 166: Is this change here by mistake?

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
```diff
@@ -89,10 +89,7 @@ abstract class DefaultCompilationUnit @Inject constructor(
     }
   }
 
-  fun generateKotlinModels(): Boolean = when {
-    project.isKotlinMultiplatform -> true
-    else -> generateKotlinModels.orElse(service.generateKotlinModels).orElse(apolloExtension.generateKotlinModels).getOrElse(false)
-  }
+  fun generateKotlinModels(): Boolean = true
```
tasomaniac, line 92: Can we not inline this function too? Do we still need it?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
martinbonnin: 3.x will break a huge amount of things. I don't really see a nice deprecation path there. We could deprecate in `2.x` but deprecating without giving a clear path forward isn't great either.

composite/settings.gradle.kts
```diff
@@ -1,16 +1,5 @@
 rootProject.name = "apollo-composite"
 
-// Samples
```
tasomaniac, line 3: What is the reason to remove samples from composite project? Are we going to replace this with something else?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
martinbonnin: This is the java codegen path. It's currently broken in `dev-3.x` and will have to be enabled again.

composite/settings.gradle.kts
martinbonnin: Definitively an oversight, I only meant to delete the java sample, I'll add the kotlin samples back 👍

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilationUnit.kt
martinbonnin: I kept the Gradle plugin java codegen path intact for now since it can ultimately be reused. I'd really like to remove this function though and not having to build the java/kotlin task graphs based on user input (i.e. certainly have different plugin ids for java and kotlin). Not 100% sure yet how this is going to end so this is a very rough first step.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
tasomaniac: Makes sense :+1:

sav007 (approved):
🤞 it works and tests will catch errors.

PR #2627, opened 2020-10-06
Another cleanup pass, focused at build scripts

martinbonnin:
Continuation from https://github.com/apollographql/apollo-android/pull/2625, with a focus on build scripts:

* removes javadoc & checkstyle as most new is Kotlin and is not supported by these tools
* move apollo-android-support to Kotlin
* uniformize a lot of things in the build scripts so that they are more consistent (plugin names, plugin block, adding dependencies, remove `java-library`, etc...)
* added the [Ben Manes](https://github.com/ben-manes/gradle-versions-plugin) versions plugin to detect newer version of our dependencies
* update dependencies
* remove ApolloSqlHelper

Also, added a small hack to make the IDE happier. Since a lot of the problems seem to come from the Android plugin interacting with the MPP one, you can now skip the Android modules by doing:

```
echo "apollographql_skipAndroidModules=true" >> ~/.gradle/gradle.properties
```

For working on Android code, this flag has to be enabled again but there isn't that much Android code. So far, I haven't seen  any issues with autocomplete, I'll keep monitoring this.


tasomaniac:
When are you planning to merge the cleanup PR? It's big but it would be better to merge it soon to avoid conflicts.

martinbonnin:
> When are you planning to merge the cleanup PR? It's big but it would be better to merge it soon to avoid conflicts.

I just did it. For these PRs, I make them "sequentially", basing  PR(N+1) on PR(N). This means they have to be merged sequentially but it should avoid conflicts.

sav007 (approved):

PR #2640, opened 2020-10-08
Delete more java only code

tasomaniac:
I realized we can delete even more :)


martinbonnin (approved):
It currently fails because `Optional` is still used from `apollo-runtime`. If everything is moved to Kotlin, we can use Kotlin optionals but there might be some Java files to convert.

tasomaniac:
Yeah, I thought this would be easier. I moved Optional to the runtime module. And converted some more unit tests to kotlin in order to remove this unnecessary abstraction over map.

apollo-api/build.gradle.kts
```diff
@@ -20,9 +20,7 @@ kotlin {
     }
   }
 
-  jvm {
-    withJava()
```
martinbonnin, line 24: Nice :)

martinbonnin (approved):

sav007 (approved):
holly .... -1585 :)

PR #2641, opened 2020-10-09
[Compiler] Unify antlrs grammars

### LINKED ISSUE #2633: [Compiler] Merge GraphQL antlr grammars
We have two grammars:
* GraphQL.g4 for operations
* GraphSDL.g4 for type definitions

The [spec](https://spec.graphql.org/June2018/#sec-Appendix-Grammar-Summary) uses the same grammar for both. Merging them will make maintenance easier in the long run.

martinbonnin:
Merge GraphQL.g4 and GraphSDL.g4. This makes less code to parse and generate and also will make the grammar easier to maintain in the long run since a lot of rules are shared between operations and type definitions (see [spec](https://spec.graphql.org/June2018/#sec-Appendix-Grammar-Summary)).

 As a side effect, this fixes a couple of bugs (passing string blocks as arguments, field names clashing with some GraphQL keywords, stricter parsing for fields definitions). Hopefully it does not introduce too many new bugs 🙈 

Closes #2633


sav007:
Question, if we unify 2 grammars does this mean we can now mix grapqhl queries and sdl schema definitions?
Should we extract common grammar into separate file and then use antlr import?

In this case when we try to parse query file with SDL it should fail with ANTLR validation?

martinbonnin:
> Question, if we unify 2 grammars does this mean we can now mix grapqhl queries and sdl schema definitions?

Yes

> Should we extract common grammar into separate file and keep 2 grammars for parsing queries and sdl separately?
> My concern what will happen when instead of query user provide SDL and vice versa.

Good point. I added some guards against that: https://github.com/apollographql/apollo-android/pull/2641/commits/000c097e764f0267d88cdef82d52d80d17d9c907

sav007 (approved):

martinbonnin:
CI failed again 😒 . It looks like it stays stuck on `Task :apollo-integration:test`... Without more real time logs, it'll be hard to debug... Anyways, I'll relaunch it and we'll see....

PR #2647, opened 2020-10-10
Fix Missing Referenced Fragments in Query Document

### LINKED ISSUE #2644: Missing Fragment Definitions in QUERY_DOCUMENT for Root Query Fragment
**Summary**
When using a shared fragment on the root query, the compiler fails to contribute the child fragment definitions to the `QUERY_DOCUMENT` field resulting in incomplete request bodies.

**Version**
Apollo Android - 2.2.0, 2.4.1
OkHttp - 4.9.0

**Description**
https://github.com/tylerbwong/ApolloRootQueryFragmentBug

The above project is able to reproduce the problem given the following query (using https://swapi.apis.guru/):

```graphql
query PlanetRootQueryBug {
  ...rootPlanetQueryFragment
}

# Might want a common root query fragment for different directives per query
fragment rootPlanetQueryFragment on Root {
  planet(planetID: 1) {
    ...planetFragment
  }
}

fragment planetFragment on Planet {
  name
  diameter
  rotationPeriod
  orbitalPeriod
}
```

The following `QUERY_DOCUMENT` field is generated for `PlanetRootQueryBugQuery` with missing fragment definitions:

```java
public static final String QUERY_DOCUMENT = QueryDocumentMinifier.minify(
    "query PlanetRootQueryBug {\n"
        + "  ...rootPlanetQueryFragment\n"
        + "}"
  );
```
which causes the fragment definitions to go missing in the request itself

```
{"operationName":"PlanetRootQueryBug","variables":{},"query":"query PlanetRootQueryBug { ...rootPlanetQueryFragment }"}
```

Not sure if this is at all related to work done in #2309.

tylerbwong:
Fixes #2644.


martinbonnin (approved):
Nice catch, Thanks a lot!

tylerbwong:
@martinbonnin We'll have to hold off on merging this. The current patch doesn't walk every child fragment on `operation.fragments`.

martinbonnin:
Thanks for digging into this! 

Looking at this code made me realise two things:
1. looks like we're going to have the same issue with inline fragments, which are not even in the `Operation` IR at the moment
2. We should add validation on the root fragments so that querying a fragment on another type than the root (`schema.queryType`) fails:

```
query PlanetRootQueryBug {
  #This should fail with 
  # Fragment `planetFragment` can't be spread here as result can never be of type Root
  ...planetFragment
}

fragment planetFragment on Planet {
  name
  diameter
  rotationPeriod
  orbitalPeriod
}
```

Let me know if you want to look into this as part of this PR or I can look into it in a follow up PR.

tylerbwong:
I'll look into both issues as part of this PR!

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
```diff
@@ -188,6 +195,21 @@ class IRBuilder(private val schema: IntrospectionSchema,
     return usedTypes
   }
 
+  private fun List<FragmentRef>.referencedRootFragmentNames(operationType: String, fragments: List<Fragment>, filePath: String): Set<String> {
```
martinbonnin, line 198: Mostly nitpicking at this point but it feels a bit "off" that we need a special case for the root fragmentsRefs compared to `field.fragmentRefs`. What about moving the `schema.rootTypeForOperationType(operationType)` to the call site and re-using `List<FragmentRef>.findFragments` ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
```diff
@@ -237,22 +259,26 @@ class IRBuilder(private val schema: IntrospectionSchema,
...
-          )
-        }
-      }
+      fragment.validateTypeCondition(ref, typeCondition, filePath)
 
       fragment
     }
   }
+
+  private fun Fragment.validateTypeCondition(fragmentRef: FragmentRef, typeCondition: String, filePath: String) {
```
martinbonnin, line 268: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
tylerbwong: Ah nice catch. A lot of code here ended up being mostly duplicated from `List<FragmentRef>.findFragments` and I didn't even realize. Will refactor! I think I'll keep this function here though since we still need to do some extra lookups for all the nested fragments inside the root fragment.

tylerbwong:
So after looking into supporting root inline fragments, I was able to add `inlineFragments` to the `Operation` IR and even get the Kotlin codegen to work properly for that use-case, but I wasn't able to figure out the Java codegen. If it's alright I think maybe we can get these changes merged in for root fragment support and I can open a separate issue/PR for supporting root inline fragments in a follow-up?

martinbonnin:
>  If it's alright I think maybe we can get these changes merged in for root fragment support and I can open a separate issue/PR for supporting root inline fragments as a follow-up?

That sounds good 👍 , thanks again !

Re: inline fragment, I didn't realise this would require big changes to the codegen. That's an area that's changing considerably in version 3.0.0 so we might want to implement that in the `dev-3.x` branch directly. I'll leave it up to you whether to:
1. fix in `main` first and merge to `dev-3.x` later (but codegen will be most likely quite different, meaning a bunch of conflicts, I can help with merging if needed)
2. or fix directly in `dev-3.x` and backport in `main` if required (maybe it won't be required).

In all cases, feel free to ignore the java codegen for the time being as it's currently removed from `3.0.0` (see https://github.com/apollographql/apollo-android/issues/2616)

martinbonnin (approved):

Issue #2523, opened 2020-08-12
Feature Request: Streaming Json parser


martinbonnin:
This is a feature request to make the Json parsing logic use a single pass "streaming" (or "pull") parser, (like Moshi uses for an exemple). This would make Json parsing faster compared to the current approach.

Our current parser parses in two passes:
* a [first pass](https://github.com/apollographql/apollo-android/blob/905620cd98dfb5cbb07b5a540558c1cc761c45e1/apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java#L94) where the Json is converted to a `Map`
* a [second pass](https://github.com/apollographql/apollo-android/blob/905620cd98dfb5cbb07b5a540558c1cc761c45e1/apollo-runtime/src/main/java/com/apollographql/apollo/response/OperationResponseParser.java#L97) where the `Map` is converted to the generated models.

This was made because current fragments could contain the same field multiple times and it is impossible to "go backwards" and re-read the same field multiple times in a streaming parser. With the work from @sav007 for modelling fragments as interfaces  (#2478), switching to a streaming parser might become possible.

A benchmark comparing Apollo `SimpleOperationParser` to Moshi ([benchmark source](https://github.com/martinbonnin/apollo-android-samples/blob/e7f093ebae93e032f009a6b343230f3995f8b27d/android-perf/src/main/java/com/perf/MainActivity.kt#L22)) gives the following results:

``` 
Benchmark Apollo 3.63ms
Benchmark Moshi  1.60ms
```

That's the average time to parse a 52kB json on a Pixel 3XL averaged 100 times. A lot of factors come into play but there are reasons to think the performance could be greatly improved by switching to a streaming parser. Another benefit is that the type of the field will most likely be known at the time we parse the Json so we won't have to convert to/from `BigDecimal` or other intermediate types anymore (see #2468).

martinbonnin:
@sav007 can we close this one?

sav007:
Closing, I was thinking that we can migrate `RealResponseReader ` to stream parser but we blocked right now by `ResponseNormalizer` so we can't touch `RealResponseReader` now.

PR #2648, opened 2020-10-10
Introduce new response reader API to support reading from stream - part1

sav007:
- Introduce new API `selectField` to be used in feature implementation of stream reader
- Provide implementation to existing Map based readers
- Update code generation to parse response based on `selectField`

Part of #2523

The actual implementation of stream readers will be introduced in next PR.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
@@ -8,6 +8,8 @@ import com.apollographql.apollo.api.ScalarType
  */
 interface ResponseReader {
 
+  fun selectField(fields: Array<ResponseField>): Int
+
   fun readString(field: ResponseField): String?
 
   fun readInt(field: ResponseField): Int?
@@ -77,5 +79,25 @@ interface ResponseReader {
         }
       })
     }
+  }
+
+  fun ResponseField.shouldSkip(variableValues: Map<String, Any?>): Boolean {
+    for (condition in conditions) {
+      if (condition is ResponseField.BooleanCondition) {
+        val conditionValue = variableValues[condition.variableName] as Boolean
+        if (condition.isInverted) {
+          // means it's a skip directive
+          if (conditionValue) {
+            return true
+          }
+        } else {
+          // means it's an include directive
+          if (!conditionValue) {
+            return true
+          }
+        }
+      }
+    }
+    return false
   }
 }
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
@@ -80,24 +80,4 @@ interface ResponseReader {
       })
     }
   }
-
-  fun ResponseField.shouldSkip(variableValues: Map<String, Any?>): Boolean {
-    for (condition in conditions) {
-      if (condition is ResponseField.BooleanCondition) {
-        val conditionValue = variableValues[condition.variableName] as Boolean
-        if (condition.isInverted) {
-          // means it's a skip directive
-          if (conditionValue) {
-            return true
-          }
-        } else {
-          // means it's an include directive
-          if (!conditionValue) {
-            return true
-          }
-        }
-      }
-    }
-    return false
-  }
 }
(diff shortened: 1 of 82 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
```diff
@@ -78,4 +80,24 @@ interface ResponseReader {
       })
     }
   }
+
+  fun ResponseField.shouldSkip(variableValues: Map<String, Any?>): Boolean {
```
martinbonnin, line 84: Why do we even need `shouldSkip` ? Isn't this supposed to be handled server side? If a server returns a value for a skipped field, this is most likely against [the spec](https://spec.graphql.org/June2018/#sec-Field-Collection). Maybe we can avoid this logic client side and save a few more CPU cycles?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReader.kt
```diff
@@ -13,121 +13,64 @@ class SimpleResponseReader private constructor(
...
 
-  override fun readString(field: ResponseField): String? {
-    if (shouldSkip(field)) {
-      return null
-    }
+  override fun selectField(fields: Array<ResponseField>): Int {
+    while (true)
+      if (responseRecordSetIterator.hasNext()) {
+        val (nextFieldName, _) = responseRecordSetIterator.next()
+        val fieldIndex = fields.indexOfFirst { field -> field.responseName == nextFieldName }
```
martinbonnin, line 28: We could leverage the fact that field [order should be preserved](https://spec.graphql.org/June2018/#sec-Serialized-Map-Ordering) to remove the iteration here and have faster parsing performance than json. Maybe with a "slow" path for non-compliant server implementations.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
sav007: Yeah, I think you are right. This function is only needed in case of normalized cache. We can remove it.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReader.kt
sav007: Even this is in 90% cases true, I wouldn't trust the order of fields we have in our models vs fields sent in query, we already do couple optimizations / merging fields, the order might changed and it's very fragile in terms of our codegen, any changes that impact fields order in our models everything will fall down as house of cards.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ResponseReader.kt
sav007: I will leave it for now in `RealResponseReader` as it used when parse normalized cache, until we replace it with stream parses.

martinbonnin: Right 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReader.kt
martinbonnin: > we already do couple optimizations / merging fields

Do you have an exemple for that?

We'd have to do some benchmarking but if we started the string matching on an index that increments as new fields are read and that'd work in 90% of the case, I think we'd save some time and be effectively faster than json (that'd be a cool marketing claim ;)). If tje 10% case fallbacks to iterating all the possible fields, everything would still work, albeit with performance on average the same as the one we have now?

sav007: Inline fragments, in some conditions we merge fields. Also I'm not sure that we preserve 100% order when build IR or AST. If we want to do this we need to verify all over the place where we build IR and AST that we keep the original order of fields.

I would say let's first do perf measurement and see if it's a case, pretty sure we will get the same numbers as Moshi as it does the same field lookup.

martinbonnin: Damn inline fragments! But yea, I agree this all needs to be benchmarked.

sav007: Actually after thinking we can do small optimization here, we can keep previously selected field index, then during next `selectField` call we can try to guess next field to select by `fields[previousSelectedFieldIndex + 1]` and if we fail fallback to lookup. Should optimize in 90%.

martinbonnin: Yes, exactly 👍

sav007: I will address this in my next PR in `StreamResponseReader`.

martinbonnin (approved):

PR #2652, opened 2020-10-13
Remove operation generic type argument for wrap data API.

sav007:
As Java codegen is not a thing any more, we don't need API to wrap data and extra `Operation` generic type.


martinbonnin (approved):
This is something that got me very confused when I started working with the project. Removing it will make reading the generated models (and using them) way easier 👍

PR #2653, opened 2020-10-13
Add Operation Inline Fragment Support

tylerbwong:
This is a continuation of #2647 and adds inline fragment support to `Operation`s.


martinbonnin:
Thanks @tylerbwong ! I just [merged main in dev-3.x](https://github.com/apollographql/apollo-android/pull/2654). Can you rebase/cherry-pick this PR in the tip of`dev.3x` ?

tylerbwong:
@martinbonnin Should be up to date with the latest from `dev-3.x` now! Also, I hope it's alright but I removed some old leftover Java codegen for one of the test cases I recently added in fe2f01a.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
```diff
@@ -371,6 +374,20 @@ class GraphQLDocumentParser(
...
+  private fun GraphQLParser.SelectionSetContext?.inlineFragments(operationType: String): List<InlineFragment> {
+    val rawRootParentType = schema.rootTypeForOperationType(operationType) ?: throw ParseException(
+        message = "Invalid operation type `${operationType}`"
+    )
+    val rootParentType = schema.resolveType(schema.resolveType(rawRootParentType))
+
+    return this
+        ?.selection()
+        ?.mapNotNull { ctx ->
+          ctx.inlineFragment()?.parse(parentSchemaType = rootParentType, ParseResult(result = emptyList()))?.result
```
martinbonnin, line 386: Double check: passing `ParseResult(result = emptyList()))` as `parentFields` is ok _I think_ because there's no polymorphism on the root query and the parent fields will be merged in all cases.  @sav007 does that sound good to you?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
```diff
@@ -371,6 +374,20 @@ class GraphQLDocumentParser(
         ?: emptyList()
   }
 
+  private fun GraphQLParser.SelectionSetContext?.inlineFragments(operationType: String): List<InlineFragment> {
```
martinbonnin, line 377: I think we want the result to be a `List<ParseResult<InlineFragment>>` to correctly track the used types. If you do something like:

```
query TestQuery {
  ... on Query {
    hero {
      appearsIn
    }
  }
}
```

`appearsIn` is an enum type that needs to be generated but won't be if the "usedTypes" are not "bubbled" up.

tylerbwong: Whoops, should be fixed by 0ed96fb!

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
sav007: hm, not sure about this. I think we need to pass parent fields as we are merging them. For example:
```
query {
  hero {
    name
    ... on Human {
      height
   }
 }
}
```
we will merge  parent field `name` and `height` for inline fragment

apollo-compiler/src/test/graphql/com/example/root_query_inline_fragment/TestOperation.graphql
```diff
@@ -0,0 +1,26 @@
...
+}
+
+fragment heroFragment on Character {
+  name
+}
+
+fragment droidFragment on Droid {
+  name
+  primaryFunction
+}
```
sav007, line 23: should we omit named fragment to make this test fixture more simple? As I understood we testing just a case when inline fragment is used as root.

tylerbwong: Sure! Removed in 6fbb8af.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
tylerbwong: I think d665d93 addresses this? Hopefully I'm on the right track.

I simplified the test cases a bit, but I'm a little confused about the purpose of [`OtherHero` in the latest test result](https://github.com/apollographql/apollo-android/pull/2653/files#diff-abc4ce9f81b5693d137e13015d4d8e4833e0a45cfa70a9da4a1ea0c4b3b2aa01R161).

martinbonnin: https://github.com/apollographql/apollo-android/commit/d665d934749ccd33abe5219acd66d5a5fb999b6a looks good to me even though I don't think it's strictly required.

@sav007 your sample will take a different path as the inline fragment is on a field. This PR is about inline fragments on the root query:

```
query {
  ... on Query  {
    hero {
    ... on Human {
      height
   }
 }
}
```
I'd say in all cases, the root "query" type will always resolve to a single concrete type so all the fields will end up merged anyways? I'm good with either https://github.com/apollographql/apollo-android/commit/d665d934749ccd33abe5219acd66d5a5fb999b6a or without. @tylerbwong Maybe put a comment linking back here in case our future selves bump into something weird?

martinbonnin:
Thanks a lot!

Issue #2661, opened 2020-10-15
Support directives at query and mutation locations


jorgebay:
The spec allows directives at QUERY and MUTATION locations but code generation doesn't.

For example, the following valid graphql is not supported by the plugin:

```graphql
mutation @myModifier CreateReviewForEpisode($review: ReviewInput!) {
  val1: createReview(review: $review),
  val2: trackReview(review: $review),
}
```

martinbonnin:
Hi! Thanks for the bug report! Just to make sure: this fails at compile time, right? 

I'm asking this because I don't think we have code to handle any operation directives. The only directives we support are `@include` and `@skip`, which are on fields. So any operation directive would most likely be ignored. Or are you expecting `myModifier` to have a specific effect?

jorgebay:
Custom directives can be created and used in `graphql-java`: 

```java
    GraphQLSchema.Builder builder = new GraphQLSchema.Builder();
    // ...

    builder.additionalDirective(
        GraphQLDirective.newDirective()
            .validLocation(Introspection.DirectiveLocation.MUTATION)
            .name("myModifier")
            .build());
```

Then, those can be read from the operation definition during execution:

```java
val directives = environment.getOperationDefinition().getDirectives();
```

But code generation doesn't support it:

```
[ERROR] Failed to parse GraphQL file /path/to/Sample.graphql (1:21)
[ERROR] Unsupported token `InsertProducts`
[ERROR] ----------------------------------------------------
[ERROR] 
[ERROR] [1]:mutation @myModifier InsertProducts(
[ERROR] [2]:    $value: ProductsInput!
[ERROR] ----------------------------------------------------
```

martinbonnin:
> Then, those can be read from the operation definition during execution:

By execution, you mean server side, right? If I understand correctly, if apollo-android would just "pass these directives through", that'll work for your use case, right?

jorgebay:
yes, just blindly pass them through would be great.

martinbonnin:
Thanks for the confirmation 👍 , I'll look into this.

martinbonnin:
Turns out, operation directives need to come **after** the operation name: https://spec.graphql.org/June2018/#OperationDefinition

Everything should work fine if you write to queries like:

```graphql
mutation InsertProducts($value: ProductsInput!) @myModifier {
  id
}
```

Let me know if that worked

jorgebay:
oh my... 🤦 I've just double checked...it works as expected.

[...]

PR #2666, opened 2020-10-16
[SDL] add support for type extensions

### LINKED ISSUE #2656: Support for type extensions in schema
Hey there 👋 

In a project I'm working on, we are using type extension, like this:

```
type A {
    field
}

extend type A {
    newField
}
```

Right now, it doesn't work as the codegen doesn't detect any field added in the extension. Do you think it would be possible to handle this?

Many thanks!

martinbonnin:
Add support for SDL type extensions: https://spec.graphql.org/June2018/#sec-Type-System-Extensions

closes #2656

PS: This PR also adds a test case for operation directives, (see #2661)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
```diff
@@ -106,6 +109,218 @@ object GraphSDLSchemaParser {
...
+
+  private fun GraphSDLParser.ScalarTypeExtensionDefinitionContext.parse(typeDefinitions: MutableMap<String, GraphSdlSchema.TypeDefinition>) {
+    val name = name().text
+    val scalar = typeDefinitions.get(name)
+    if (scalar == null) {
+      throw ParseException(
+          message = "Cannot add scalar type extension on unknown scalar `$name`",
+          token = start
+      )
+    }
```
sav007, line 135: nit:  just a preference
```
val scalar = typeDefinitions.get(name) ?: throw ParseException(
          message = "Cannot add scalar type extension on unknown scalar `$name`",
          token = start
      )
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
```diff
@@ -106,6 +109,218 @@ object GraphSDLSchemaParser {
     )
   }
 
+  private fun List<GraphSDLParser.TypeSystemExtensionContext>.parse(
+      typeDefinitions: MutableMap<String, GraphSdlSchema.TypeDefinition>
```
sav007, line 113: not a big fan of passing mutable map and having side affect to mutate `typeDefinitions` map, but don't see easy way to make it pure. We might consider to rename this function to smth that describes mutation maybe like `MutableMap<String, GraphSdlSchema.TypeDefinition>.decorateWithTypeExtentantion()`?

apollo-compiler/src/test/graphql/com/example/operation_directive/TestOperation.graphql
```diff
@@ -0,0 +1,5 @@
+query TestQuery @myModifier {
```
sav007, line 1: that interesting am I right that this `myModifier` directive is not declared in the schema? I thought that GraphQL spec says that directive also must be declared in the schema?

martinbonnin: Yup, that's 100% correct. I added this test as part of checking https://github.com/apollographql/apollo-android/issues/2661. 

But since `myModifier` is not in the schema, this means we're missing some validation there. I added https://github.com/apollographql/apollo-android/issues/2673 as a follow up.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
martinbonnin: IntelliJ agrees with you and put yellow underlines. I replaced with them.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/sdl/GraphSDLSchemaParser.kt
martinbonnin: Yea, not a huge fan for mutable maps either. I just pushed a new version with a `fold`. That's a lot more copying Maps all over the place but I think that's the price to pay for immutability.

apollo-compiler/src/test/graphql/com/example/operation_directive/TestOperation.graphql
sav007: Yeah, I guess we missed that validation as we have tests for skip / include and I don't see them either in schema sdl.

sav007 (approved):

Issue #2563, opened 2020-09-08
[Plugin] Robustify `no-source` handling


martinbonnin:
`.graphql` files are not found if they are in a `graphQL` folder instead of `graphql`, that can be confusing when running the plugin on a linux CI vs a MacBook laptop (see https://spectrum.chat/apollo/apollo-android/ci-cd-checks-failing~56d4f4cb-ce6d-451e-a72d-7c41d8b2d160)

We could make the plugin fail early with a better error message if no `src/*/graphql` folder exists on the filesystem.

But that will most likely make 99% of the builds that have no such folder for test variants fail. 

I'm starting to think the solution is to make the test variant compilation unit opt-in. Especially since we have some custom logic to not inherit the main sourceSet from test.

martinbonnin:
Closed by #2668

PR #2668, opened 2020-10-16
🍂 3.0.0 fall cleanup of the Gradle Plugin 🍂

martinbonnin:
**At a high level**, the main change is that the plugin doesn't try to create a GraphQL source set per Kotlin source set anymore. Instead, there is now a 1:1 mapping between a service and a GraphQL source set. By default, files are are searched under `src/main/graphql` or `src/commonMain/graphql` but this is just a convention that can be overriden.

Trying to map GraphQL to Java/Kotlin source sets was confusing and created issues like https://github.com/apollographql/apollo-android/issues/1981 or special cases for the test source sets where we would end up with duplicate classes in the classpath. Also the schema resolution rules were becoming super complex and hard to maintain.

Having one GraphQL service per service is a simple mapping that allows the files/schema resolution rules to be way easier than they used to be, removing the need for `schemaPath` vs `schemaFile`, `CompilationUnit`, `ApolloVariant`, ... 

**In more details**, a few points worth noting:

### Android variants handling

The custom Android variants handling code has been removed. This means that the plugin will not look into the different `src/${flavor}/graphql` directories by default. I'm hoping that in the very vast majority of case this will not an issue because there is rarely a need for different queries. This will even save some compilation time by not generating sources for each variant. If per-variant queries are still required, this can still be achieved either with:
1. R8: define all the queries in the main variant and have R8 remove the ones that are unused in other variants.
2. Or define different services and wire them to the variants manually (see below)

### Custom task wiring

For the case where finer control over how the generated sources are used is required, the `Service` now offers a `withOutputDir` action that exposes a `Task` and a `Provider<Directory>` that you can forward to other parts of the build as needed. Coupled with `Service.outputDir`, this allows more complex scenarios like registering generated sources to specific Android variants or to a test source set:

```kotlin
apollo {
  service("starwars") {
    graphqlSourceDirectorySet.srcDir("src/test/graphql/starwars")
    withOutputDir {
      val kotlinProjectExtension = project.extensions.get("kotlin") as KotlinProjectExtension
      // add the starwars generated sources to the test source set
      kotlinProjectExtension.sourceSets.getByName("test").kotlin.srcDir(outputDir)
    }
  }
}
```

### Default service

This PR also gets rid of the whole `CompilerParams` inheritance/overriding because:

1. the behaviour wasn't always obvious
2. explicit is better than implicit
3. it didn't make any sense for some arguments like `customTypeMapping` that are specific to a schema to be shared by a common ancestor.

To keep things simple and keep some of the previous behaviour, for simple cases where there is only one schema, a default service is created at the root of the `apollo {}` block:

```kotlin
apollo {
  // this is still working and will generate generated/source/apollo/service/...
  sourceFolder.set("starwars") 
}
```

Mixing implicit and explicit services will fail to make this new behaviour explicit:

```
apollo {
  sourceFolder.set("starwars") 
  service("github") {
    // this will fail. You can either remove `sourceFolder.set("starwars")` above or create an explicit "starwars" service
    sourceFolder.set("github")
  }
}
```

### (Almost) no more `afterEvaluate {}`

The plugin is one step closer to removing `afterEvaluate` completely. Tasks are now registered 100% reactively. There are still 2 `afterEvaluate`: one for the defaultService and the other to make sure the Android and Kotlin plugins have settled. This last one is maybe not even required but it felt safer to keep it. 

closes https://github.com/apollographql/apollo-android/issues/2563


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/Registry.kt
```diff
@@ -0,0 +1,22 @@
+package com.apollographql.apollo.gradle.api
+
+import org.gradle.api.provider.Property
+
+interface Registry {
```
sav007, line 5: Is this class used for just grab GraphQL schema from Apollo registry? Should we add some class docs?

sav007 (approved):
That great clean up. TBH I was lost with compilation units and inheritance in the past.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/Registry.kt
martinbonnin: Yes, it's the equivalent of `apollo client:download-schema` in the npm cli.  👍  I'll add some class docs.

PR #2669, opened 2020-10-16
[bugfix] fragment code-gen respect module package.

Laimiux:


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/FragmentsResponseMapperBuilder.kt
```diff
@@ -69,7 +69,7 @@ class FragmentsResponseMapperBuilder(
   private fun mapperFields(fragmentFields: List<ResponseFieldSpec>): List<FieldSpec> {
     return fragmentFields.map { field ->
       val fieldType = field.normalizedFieldSpec.type as ClassName
-      val mapperClassName = ClassName.get(context.ir.fragmentsPackageName, fieldType.simpleName(),
+      val mapperClassName = ClassName.get(fieldType.packageName(), fieldType.simpleName(),
```
Laimiux, line 72: I think this should work but would love another pair of eyes.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
```diff
@@ -531,7 +531,8 @@ class GraphQLDocumentParser(
             fragmentRefs = fragmentRefs.union(mergeInlineFragmentRefs).toList(),
             inlineFragments = inlineFragments.result.filter { it.typeCondition != typeCondition },
             filePath = graphQLFilePath,
-            sourceLocation = SourceLocation(start)
+            sourceLocation = SourceLocation(start),
+            packageName = "${packageNameProvider.operationPackageName(graphQLFilePath)}.fragment".removePrefix(".")
```
martinbonnin, line 535: A few comments:

* What about dropping the ".fragment" here? As long as all fragments ended up in the same package but if we start using the file package name instead, it might be confusing to append ".fragment
* Since fragments are not operations, we might want to rename `operationPackageName` to `filePackageName`
* It's ok to change this behaviour for experimental multi-module things but not for the stable non-multi-module users so this will have to be configurable somewhere.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/SchemaTypeSpecBuilder.kt
```diff
@@ -190,7 +190,7 @@ class SchemaTypeSpecBuilder(
             val optional = fragmentRef.isOptional()
             val possibleTypes = fragment.takeIf { fragment.typeCondition != normalizeGraphQlType(schemaType) }?.possibleTypes ?: emptyList()
             val fieldSpec = FieldSpec.builder(
-                JavaTypeResolver(context = context, packageName = context.ir.fragmentsPackageName)
+                JavaTypeResolver(context = context, packageName = fragment.packageName)
```
Laimiux, line 193: I'm assuming `fragmentRef` and `fragment` are representing the same object.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
Laimiux: Thanks for the feedback:
- I'm for dropping the `fragment` package suffix. 
- Makes sense, however, if we want to make this behavior configurable, maybe having a separate `fragmentPackageName()` function that encapsulates this logical fork?
- Do you have a suggestion for a flag name?

martinbonnin: >I'm for dropping the fragment package suffix.

👍 

>Do you have a suggestion for a flag name?

Maybe something like: 

```
apollo {
  // Use the path to the file that defined the fragment as package name
  // i.e. com/example/Fragment.graphql will end up in `com.example` package instead of a global `*.fragment` package
  useFilePackageNameForFragments.set(true)
}
```

> maybe having a separate fragmentPackageName() function that encapsulates this logical fork?

Yup, that works too

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/FragmentsResponseMapperBuilder.kt
martinbonnin: Looks good to me 👀

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/AST.kt
```diff
@@ -92,6 +92,11 @@ internal data class ObjectType(
   }
 }
 
+internal class FragmentType(
+    val objectType: ObjectType,
+    val packageName: String
+)
```
martinbonnin, line 98: Can you add `packageName: String?` to `ObjectType` instead? It doesn't really make sense to have an AST type named "Fragment" as Kotlin doesn't know anything about a Fragment.

martinbonnin: Mmm but there's also an OperationType too so maybe it doesn't matter that much. @sav007 any opinion about this ?

Laimiux: I think your point that it doesn't map to a `Kotlin` concept is valid. The reason I didn't place it on `ObjectType` is that `ObjectType` is also used as an intermediary object in a couple of places so it feels like it is breaking this principle already. I used a separate type to more closely tie package name to class generation logic which allows the type system to ensure that packageName is provided. Let me know if you want me to change that, not a big deal to me just wanted to explain my position.

Laimiux:
Note: already ran `./gradlew :apollo-compiler:test -DupdateTestFixtures=true`

martinbonnin (approved):
Thanks again for diving into this!

The CI check just failed because the public API changed (new useFilePackageNameForFragments), I just updated the metalava files so it should be good now 🤞 . 

Something else we will want is a codegen test. Let me know if you want to add one or if you want me to add it before merging.

Laimiux:
> Thanks again for diving into this!
> 
> The CI check just failed because the public API changed (new useFilePackageNameForFragments), I just updated the metalava files so it should be good now 🤞 .
> 
> Something else we will want is a codegen test. Let me know if you want to add one or if you want me to add it before merging.

No problem. Regarding adding a codegen test, I can do it maybe towards the middle of the week at the earliest. Is there a specific class/example I can take a look at? Is `MetadataTest` the right file to look at?

martinbonnin:
>  Is there a specific class/example I can take a look at? 

I think the easiest place would be `CodegenTest.kt`: https://github.com/apollographql/apollo-android/blob/main/apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt

Adding a new folders in `src/test/graphql/com/example/fragment_package_name` and verify that the Fragments are not generated in the `fragment` folder.

Laimiux:
> > Is there a specific class/example I can take a look at?
> 
> I think the easiest place would be `CodegenTest.kt`: https://github.com/apollographql/apollo-android/blob/main/apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/CodegenTest.kt
> 
> Adding a new folders in `src/test/graphql/com/example/fragment_package_name` and verify that the Fragments are not generated in the `fragment` folder.

Let me know if this works, I ran it with `./gradlew :apollo-compiler:test -DupdateTestFixtures=true` to generate the fixtures.

martinbonnin:
> Let me know if this work

Looks good 👍 . The fragments aren't generated in the `fragment` package, which is what we wanted.

Laimiux:
> > Let me know if this work
> 
> Looks good 👍 . The fragments aren't generated in the `fragment` package, which is what we wanted.

It might be worth adding a test to imitate two modules (parent/child) generating fragments in different packages. However, I don't have a clear path for that.

martinbonnin:
Hi @Laimiux ! I've been thinking more about this as part of the [3.0 Gradle plugin cleanup](https://github.com/apollographql/apollo-android/pull/2668). Ultimately, I'd like to simplify the current way the package names are generated. I think the current way of generating package names from the folder hierarchy is complex and I was considering forcing all the `.graphql` files in a single "flat" directory. 
Would that work for your use case?

Also, even with this PR, input types/enums/scalars will continue being generated in the `schemaPackageName`. That feels weird. What do you think of adding a `packageName` option that will set the packageName explicitly without relying on the complex logic we have now?

Laimiux:
> Hi @Laimiux ! I've been thinking more about this as part of the [3.0 Gradle plugin cleanup](https://github.com/apollographql/apollo-android/pull/2668). Ultimately, I'd like to simplify the current way the package names are generated. I think the current way of generating package names from the folder hierarchy is complex and I was considering forcing all the `.graphql` files in a single "flat" directory.
> Would that work for your use case?
> 
> Also, even with this PR, input types/enums/scalars will continue being generated in the `schemaPackageName`. That feels weird. What do you think of adding a `packageName` option that will set the packageName explicitly without relying on the complex logic we have now?

I actually use a flat directory structure for `graphql` files and each module specifies `rootPackageName` where it wants the generated classes to be placed in.

martinbonnin:
> I actually use a flat directory structure for graphql files and each module specifies rootPackageName where it wants the generated classes to be placed in.

👍  If that's ok with you, I'll work on a follow up PR to replace `useFilePackageNameForFragments ` with a `packageName` option that will assume a flat directory and override the package name for everything: input types, fragments, operations, etc..

martinbonnin:
@Laimiux I ended up pushing the changes here as it was easier and keeps the discussion in one place. 

I replaced `useFilePackageNameForFragments` with `packageName`:

* flattens and uses `$packageName` and all operations
* flattens and uses `$packageName.fragment` for all fragments. That's a difference but I think it can be useful in the future to avoid name clashes and also to keep things organized.
* does not touch the input/enum/scalar package name which is still under `schemaPackageName` as it's a much bigger change

Let me know if that works for your use case. If yes I'll push a new release next Monday

Laimiux:
> @Laimiux I ended up pushing the changes here as it was easier and keeps the discussion in one place.
> 
> I replaced `useFilePackageNameForFragments` with `packageName`:
> 
> * flattens and uses `$packageName` and all operations
> * flattens and uses `$packageName.fragment` for all fragments. That's a difference but I think it can be useful in the future to avoid name clashes and also to keep things organized.
> * does not touch the input/enum/scalar package name which is still under `schemaPackageName` as it's a much bigger change
> 
> Let me know if that works for your use case. If yes I'll push a new release next Monday

Thanks for making this change. This will work for my use cases, though personally, I would vote for removing the `fragment` package name altogether.

martinbonnin:
> I would vote for removing the fragment package name altogether.

What makes it better for your use cases? One big reason to keep it is to avoid name collisions between Fragments and o

martinbonnin:
Thanks for investigating this! The `2.4.2-SNAPSHOT` version should be available from OSS snapshots in ~30min. New developments for input/scalar/enum package names will happen in `dev-3.x`.

PR #2674, opened 2020-10-19
[Performance] Add a Benchmark project

### LINKED ISSUE #2655: Add Performance Benchmarks
With `3.0.0` focusing on performance, we should add performance benchmarks to the repo so we can measure the progress. 

I think benchmark could be run on the host JVM in a first step since it's the easiest. Having them run on real Android devices with https://developer.android.com/studio/profile/benchmark would give more fine grained insights in a second step.

martinbonnin:
Add an Android benchmark project based on [Jetpack Benchmark](https://developer.android.com/studio/profile/benchmark).

It uses the same models as [zacsweers/json-serialization-benchmarking](https://github.com/ZacSweers/json-serialization-benchmarking/tree/main/models/src/main/java/dev/zacsweers/jsonserialization/models). They're not super deep but they should give a good first point of comparison. 

Going from `2.4.1` to `3.0.0-SNAPSHOT` ([commit](https://github.com/apollographql/apollo-android/commit/8dfc59804f57aa03a11c94fe7731dc21195ed6c8))

**2.4.1 Java codegen**:
```
benchmark:    23,697,764 ns Benchmark.apollo
benchmark:    12,750,782 ns Benchmark.moshi
```

**2.4.1 Kotlin codegen**:
```
benchmark:    23,854,430 ns Benchmark.apollo
benchmark:    12,649,481 ns Benchmark.moshi
```

**3.0.0-SNAPSHOT Kotlin codegen**:
```
benchmark:    14,700,418 ns Benchmark.apollo
benchmark:    12,862,032 ns Benchmark.moshi
```

**Edit**: this was comparing 2.4.1 Java codegen to 3.0.0-SNAPSHOT Kotlin codegen. I just added the test with 2.4.1 Kotlin codegen. Let me know if I'm missing something else.

closes #2655


benchmark/addTypename.main.kts
```diff
@@ -0,0 +1,45 @@
...
+@file:DependsOn("com.squareup.okio:okio:2.9.0")
+@file:DependsOn("com.squareup.moshi:moshi:1.11.0")
+
+import com.squareup.moshi.Moshi
+import okio.buffer
+import okio.sink
+import okio.source
+import java.io.File
+
+val adapter = Moshi.Builder().build().adapter(Any::class.java)
```
BenSchwab, line 11: what is this file? Is it might to be committed - it doesn't seem to be used in the actual benchmark from what I can tell?

benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Benchmark.kt
```diff
@@ -0,0 +1,43 @@
...
+      .source()
+      .buffer()
+
+  @Test
+  fun apollo() = benchmarkRule.measureRepeated {
+    val bufferedSource = runWithTimingDisabled {
+      bufferedSource()
+    }
+
+    operation.parse(bufferedSource)
```
BenSchwab, line 32: I would love to also see a benchmark that goes the actual apollo client with response normalization enabled

BenSchwab (approved):
one Q - but benchmark lgtm!

benchmark/addTypename.main.kts
martinbonnin: Good catch! It was a one-time script to add the `__typename` field to the [ZacSweers's largesample.json](https://github.com/ZacSweers/json-serialization-benchmarking/blob/main/models/src/main/resources/largesample.json) since GraphQL requires `__typename`. It's actually one area where we might gain a tiny bit of performance if we wanted to save this parsing for fields that are not polymorphic.
In all cases, this file isn't needed anymore, I removed it.

benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Benchmark.kt
martinbonnin: 👍 Good idea, I'll try to think of something

AdamMTGreenberg:
@martinbonnin will there be any ways to hook from the client end and check specific performance benchmarks of parsing for implementation purposes, perhaps via the ApolloParseInterceptor?

martinbonnin:
@AdamMTGreenberg can you elaborate a bit more? Right now I can see two scenarios:

1. Parsing benchmark (this PR): this can be made offline on test data.
2. Real User Monitoring: this requires actual production data and would also measure network speed.

Since 1. can be made offline, it can be unit tested. I don't think we need hooks in the runtime for this but maybe you have different requirements?
2. is a different story and we might need hooks to inspect the performance of the client in real life. But in this case, we'd need hooks to measure at least network, cache and parsing, maybe more.

AdamMTGreenberg:
@martinbonnin thank you - our use case would involve the Real User Monitoring scenario.

Ideally, we are searching for ways to include hooks in production for network, cache and parsing as you mentioned. The goal would be to ensure that in our application in production we could test for regressions and monitor app health. The hooks/callbacks would feed into our system and allow us to trace these benchmarks. Simple start/finish would be the best.

Thanks so much

martinbonnin:
@AdamMTGreenberg I'm assuming you're talking about the upcoming 3.x version ? If that's the case, let's follow up there: https://github.com/apollographql/apollo-android/issues/3221

If you're looking at 2.x, you can take a look at [`ApolloCall.Callback.onStatusEvent`](https://github.com/apollographql/apollo-android/blob/9513025a9640517f0fa448c9859938106c50606a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.java#L113) that triggers Cache/Network/Completed events.

Issue #2536, opened 2020-08-23
Is it possible to use apollo-android v2 with Gradle 5?


RyPope:
[The docs for v2 say that][1] 

> Gradle 6.x recommended. In 5.x, Gradle Metadata needs to be enabled by
> putting this into settings.gradle
> enableFeaturePreview("GRADLE_METADATA")

Which says that 6.x is simply recommended, but when I try to use Gradle 5 I get the error

    apollo-android requires Gradle version 6.0 or greater

Is there a workaround for this? Upgrading to Gradle 6 is not an option for my project right now.

[Here is the change][2] in apollo that bumps the minimum required Gradle version

  [1]: https://www.apollographql.com/docs/android/essentials/migration/#migrating-to-2x
  [2]: https://github.com/apollographql/apollo-android/commit/c5fee8e015ec18fd4e56510fe49c23cccf0add44#diff-bf25993eb69e5ccf9c0a98ed2a93eaf8R17

martinbonnin:
I don't think there's a workaround to use Gradle 5 with version 2.3.0. 
I can check tomorrow if there's a way to disable the check (by adding a special plugin option) but that would require a new version in all cases. 
Can you detail a bit more why upgrading your project to Gradle 6 is not an option? I was under the impression that the upgrade should be transparent most of the times but if it's not, we should plan for longuer support periods.

RyPope:
I work at a larger organization that has its own build system and manages dependencies internally so updating is just a little slower and for a major dependency like Gradle it requires some effort.

martinbonnin:
It turns out it should be possible to run the plugin with 5.6. See https://github.com/apollographql/apollo-android/pull/2539

Anything lower than 5.6 will fail and as Gradle moves forward, I would strongly advise to update to more recent versions. The more you wait, the more effort will be needed to upgrade.

RyPope:
I appreciate the prompt fix @martinbonnin ! Our Gradle 6 upgrade is most certainly underway but sometimes these things are easier said than done.

PR #2677, opened 2020-10-20
[Gradle Plugin] Add helpers for android and multi source sets

martinbonnin:
Follow up from https://github.com/apollographql/apollo-android/pull/2668

Add `createAllAndroidVariantServices` and `createAllKotlinJvmSourceSetSevices`. These are long method names for use cases where adding GraphQL files to non-main source sets/variants is needed:

```kotlin
apollo {
  // Will create
  // - mainExample
  // - testExample
  createAllKotlinJvmSourceSetServices("example") {
    schemaFile.set(file("src/main/graphql/com/example/schema.sdl"))
  }
}
```

```kotlin
apollo {
  // Will create the services below and collect all graphql files according to Android source sets rules
  // - debugExample
  // - debugUnitTestExample
  // - debugAndroidTestExample
  // - releaseExample
  // - releaseUnitTestExample
  createAllAndroidVariantServices("example") {
    // It's ok to omit this in which case there should be one and only one schema per source set
    schemaFile.set(file("src/main/graphql/com/example/schema.sdl"))
  }
}
```


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -45,7 +45,10 @@ class GraphQLCompiler(val logger: Logger = NoOpLogger) {
     )
 
     val files = args.graphqlFiles
-    checkDuplicateFiles(roots, files)
+    /**
+     * Android can have duplicate files but that's ok as long as they do not define the same operations (which is checked in IRBuilder)
+     */
+    // checkDuplicateFiles(roots, files)
```
tasomaniac, line 51: Shouldn't we just delete this completely now?

martinbonnin: I guess I got sentimentally attached to this method 😂 . You're right, I'll remove it.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
```diff
@@ -20,14 +21,22 @@ import javax.inject.Inject
 abstract class DefaultService @Inject constructor(val objects: ObjectFactory, override val name: String)
   : Service {
 
+  init {
+    // see https://github.com/gradle/gradle/issues/7485
+    // TODO replace with `convention(null)` when we can target Gradle 6.2
```
tasomaniac, line 26: Can we not target 6.2 now?

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
```diff
@@ -123,16 +128,7 @@ abstract class DefaultService @Inject constructor(val objects: ObjectFactory, ov
         """
 Multiple schemas found:
 ${candidates.joinToString(separator = "\n")}
```
tasomaniac, line 130: This error is now missing a suggestion. When we fail because of multiple schemas, what should they do?

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
martinbonnin: This was filed not so long ago: https://github.com/apollographql/apollo-android/issues/2536

Overall, I don't think this alone justifies bumping the Gradle version. We could do something like this:

```
if (gradleVersion >= 6.2) {
   customTypeMapping.convention(null)
} else {
   customTypeMapping.set(null as Map<String, String>?)
}
```

This would make it possible for 6.2+ users to do something like

```
apollo {
  customTypeMapping.put("Date", "java.util.Date")
}
```

(in addition to the always working)

```
apollo {
  customTypeMapping.set(mapOf("Date" to "java.util.Date"))
}
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
martinbonnin: The answer to this question depends whether you're using `service` or `createAllAndroidVariantServices`. It was actually wrong to suggest using multiple services before if the multiple schemas were the result of schemas being present in multiple flavor/buildType source sets. So I'd prefer a less precise error message over a wrong one.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/AndroidProject.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.file.Directory
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.TaskProvider
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+object AndroidProject {
+  fun onEachVariant(project: Project, withTestVariants: Boolean = false, block: (BaseVariant) -> Unit) {
```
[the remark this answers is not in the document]
martinbonnin: > project is not used in this function. What about making this an extension over androidExtension or making androidExtension a parameter?

I like that it's somehow symmetric with `KotlinJvmProject` and `KotlinMultiplatformProject`. These helpers are now public APIs so users can use them to do all kind of crazy things if they really want to. They all take a `project` as a first parameter because this is what's usually available from `build.gradle[.kts]`

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
tasomaniac: Hmm, I was thinking because of 3.x but you're right. If we don't need it, we shouldn't.

Your suggestion also looks like a good improvement. I didn't know that. It is up to you.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
tasomaniac: Can we say that multiple schema files are not supported?

tasomaniac (approved):

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
martinbonnin: I added it, it feels more natural to have proper `.put` support

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultService.kt
martinbonnin: I added this: `Multiple schemas are not supported. You can either define multiple services or specify the schema you want to use explicitely with `schemaFile`. I think this is technically not wrong :)

PR #2684, opened 2020-10-22
[SDL] allow to have schemas without mutations/subscriptions

### LINKED ISSUE #2683: downloadApolloSchema inserts a bogus `subscription` even when schema doesn't support subscriptions
**Summary**
Running the `downloadApolloSchema` in SDL mode against a schema without subscriptions results in an incorrect SDL.

**Version**
2.4.1

**Description**
Our schema currently doesn't support subscriptions. When I run the schema download tool to pull down our JSON schema, it works as expected, but the SDL version is outputting a malformed schema with a stub `subscription` type.

The following command correctly pulls down a JSON format schema: `./gradlew :service:downloadApolloSchema --endpoint=<endpoint>--schema=service/src/main/graphql/schema.json` In our output, I see the following, elided for brevity:
```
{
  "data":{
    "__schema":{
      "queryType":{"name":"QueryRoot"},
      "mutationType":{"name":"MutationRoot"},
      "subscriptionType":null
    }
  }
}
```

But when I run the same command to produce a `.graphql` file instead (like so: `./gradlew :service:downloadApolloSchema --endpoint=<endpoint>--schema=service/src/main/graphql/schema.graphql`), I get the following output in our schema file:
```
schema {
  subscription: subscription
  query: QueryRoot
  mutation: MutationRoot
}
```

It's not a huge deal, but this file is now considered invalid. This kind of breaks some of our tooling, like the GraphQL plugin for Android Studio, because `subscription` is an unknown type.

martinbonnin:
From [the spec](https://spec.graphql.org/June2018/#sec-Root-Operation-Types), it is allowed to have SDL schemas without subscriptions/mutations.

SDL will use the default of `Query`, `Mutation`, `Subscription` if no `schema {}` definition is found.
Introspection doesn't have default value so whatever is returned will be used and an error will be thrown if there's no "query" root type.

Fixes #2683

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/introspection/IntrospectionSchema.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/introspection/IntrospectionSchema.kt
@@ -10,9 +10,9 @@ import java.io.File
 import java.io.InputStream
 
 data class IntrospectionSchema(
-    val queryType: String = "query",
-    val mutationType: String = "mutation",
-    val subscriptionType: String = "subscription",
+    val queryType: String,
+    val mutationType: String?,
+    val subscriptionType: String?,
     val types: Map<String, Type>) : Map<String, IntrospectionSchema.Type> by types {
   sealed class Type(val kind: Kind) {
     abstract val name: String
@@ -162,9 +162,9 @@ data class IntrospectionSchema(
 
     fun IntrospectionQuery.Schema.toIntrospectionSchema(): IntrospectionSchema {
       return IntrospectionSchema(
-          queryType = queryType?.name ?: "query",
-          mutationType = mutationType?.name ?: "mutation",
-          subscriptionType = subscriptionType?.name ?: "subscription",
+          queryType = queryType?.name ?: "Query",
+          mutationType = mutationType?.name,
+          subscriptionType = subscriptionType?.name,
           types = types.associateBy { it.name }
       )
     }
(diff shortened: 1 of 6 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/introspection/IntrospectionSchema.kt
```diff
@@ -162,19 +162,19 @@ data class IntrospectionSchema(
 
     fun IntrospectionQuery.Schema.toIntrospectionSchema(): IntrospectionSchema {
       return IntrospectionSchema(
-          queryType = queryType?.name ?: "query",
-          mutationType = mutationType?.name ?: "mutation",
-          subscriptionType = subscriptionType?.name ?: "subscription",
+          queryType = queryType?.name ?: "Query",
```
tasomaniac, line 165: Just to double check Is `Query` supposed to be uppercase?

tasomaniac (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/introspection/IntrospectionSchema.kt
martinbonnin: That's my understanding: 

```
While any type can be the root operation type for a GraphQL operation, the type system definition language can omit the schema definition when the query, mutation, and subscription root types are named Query, Mutation, and Subscription respectively.
```

I do expect this to never happen in practice though. I don't see how an introspection result could not have a "query" root type since it was used to get the introspection result. I took the conservative option but maybe we should throw here

sav007 (approved):

Issue #2681, opened 2020-10-21
[Gradle Plugin] ApolloCheckDuplicatesTask is not marked as a CacheableTask


olivierg13:
**Summary**
When using the gradle build cache, since ApolloCheckDuplicatesTask is not marked as a CacheableTask (but has an output), the task won't be optimized against the same generated models (since ApolloGenerateSourcesTask is effectively marked as a CacheableTask).

**Version**
2.4.1

**Description**
How to reproduce: 
* Enable your gradle build cache
* assemble -> clean -> assemble
* You'll see the build messages: x actionable tasks: y executed, z from cache, xx up-to-date (xx being > 0 and ApolloCheckDuplicatesTask being one of them)

PS: I'll work on marking ApolloCheckDuplicatesTask as Cacheable, and the corresponding unit test :)

martinbonnin:
Fixed in https://github.com/apollographql/apollo-android/pull/2688. This can be checked in `2.4.2-SNAPSHOT` and will be released in `2.4.2`, probably next week.

PR #2688, opened 2020-10-22
[Gradle Plugin] Make CheckDuplicate task cacheable

olivierg13:
Fix for #2681

**Summary**
When enabling the gradle build cache, since ApolloCheckDuplicatesTask is not marked as a CacheableTask (but has an output), this task won't be optimized against the same generated models (ApolloGenerateSourcesTask is already marked as a CacheableTask).


martinbonnin (approved):
Thanks!

tasomaniac (approved):

PR #2690, opened 2020-10-23
[KMP] added a 'timeoutMillis' parameter to ApolloHttpNetworkTransport

### LINKED ISSUE #2680: Way to add request timeout in apollo kotlin multiplatform
**We don't have a way to set request timeout in Apollo client**
I'm using apollo kotlin multiplatform in my project and it's amazing but i have found that there is no way to set request timeout. Both OkHttp and NSURLConnection supports it currently we have to re-implement the whole NetworkTransport in both android and ios to achieve this.

**Adding requestTimeout parameter ApolloHttpNetworkTransport class**
Adding the requestTimeout parameter would be a good idea.

``` kotlin
val apolloClient = ApolloClient(
      networkTransport = ApolloHttpNetworkTransport(
          serverUrl = "https://api.example.com",
          requestTimeout = 10_000 //10 seconds
       )
 )
```

martinbonnin:
closes #2680

### DIFF
--- a/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
+++ b/apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
@@ -10,10 +10,17 @@ expect class ApolloHttpNetworkTransport constructor(
     headers: Map<String, String>,
     httpMethod: HttpMethod = HttpMethod.Post,
     /**
+     * The timeout interval to use when connecting
+     *
+     * - on iOS, it is used to set [NSMutableURLRequest.timeoutInterval]
+     * - on Android, it is used to set [OkHttpClient.connectTimeout]
+     */
+    connectTimeoutMillis: Long = 60_000,
+    /**
      * The timeout interval to use when waiting for additional data.
      *
-     * - on iOS, it is used to set NSMutableURLRequest.timeoutInterval
-     * - on Android, it is used to set both readTimeout and connectTimeout
+     * - on iOS, it is used to set [NSURLSessionConfiguration.timeoutIntervalForRequest]
+     * - on Android, it is used to set  [OkHttpClient.readTimeout]
      */
-    timeoutMillis: Long = 30_000
+    readTimeoutMillis: Long = 60_000,
 ) : NetworkTransport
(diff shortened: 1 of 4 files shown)


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
```diff
@@ -8,5 +8,12 @@ import com.apollographql.apollo.network.NetworkTransport
...
     headers: Map<String, String>,
-    httpMethod: HttpMethod = HttpMethod.Post
+    httpMethod: HttpMethod = HttpMethod.Post,
+    /**
+     * The timeout interval to use when waiting for additional data.
+     *
+     * - on iOS, it is used to set NSMutableURLRequest.timeoutInterval
+     * - on Android, it is used to set both readTimeout and connectTimeout
+     */
+    timeoutMillis: Long = 30_000
```
sav007, line 18: should we have 2 timeouts for connection and reading, looks like NSURLsession allows to set both:
```
https://developer.apple.com/documentation/foundation/nsmutableurlrequest/1414063-timeoutinterval?language=objc
https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408259-timeoutintervalforrequest?language=objc
```

martinbonnin: 👍 I missed `timeoutIntervalForRequest`, will add that

martinbonnin: Mmm, `timeoutIntervale` is on `NSMutableUrlRequest` but `timeoutIntervalForRequest` is for `NSUrlSession`. I'm not sure we can have both at the same time?

martinbonnin: Looks like we can! I just pushed a few changes

martinbonnin: i'm not 100% sure the relation between `timeoutInterval` and `timeoutIntervalForRequest`. It might make sense to map everything to the same timeout on iOS. But on the other hand, having both is more flexible in terms of API so we can certainly keep it as it is, even if there's a small chance the concept do not map 1:1

PR #2691, opened 2020-10-23
[KMP] set the Content-Type to `application/json; charset=utf-8` on iOS as well

### LINKED ISSUE #2665: Multiplatform - wrong content-type by default on iOS
**Summary**
The content-type header value on iOS by default is `application/x-www-form-urlencoded` instead of `application/json; charset=utf-8` on android. 

**Version**
2.4.1

**Description**

The content-type header value on iOS by default is `application/x-www-form-urlencoded` instead of `application/json; charset=utf-8` on android. 

This leads to server HTTP 400 Response.

If manually override header to `application/json` (like in example below) then request works on both platforms.

```
return ApolloClient(
            networkTransport = ApolloHttpNetworkTransport(
                serverUrl,
                mapOf("content-type" to "application/json; charset=utf-8") // Fix default content-type on iOS
            ),
            interceptors = listOf(bearerTokenInterceptor, loggingInterceptor, errorInterceptor),
            scalarTypeAdapters = ScalarTypeAdapters(getCustomTypeAdapters()),
        )
```

martinbonnin:
closes #2665

### DIFF
--- a/apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
+++ b/apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
@@ -152,6 +152,7 @@ actual class ApolloHttpNetworkTransport(
       val postBody = operation.composeRequestBody(scalarTypeAdapters).toByteArray().toNSData()
       setHTTPMethod("POST")
       headers
+          .plus("Content-Type" to "application/json; charset=utf-8")
           .plus(httpExecutionContext?.headers ?: emptyMap())
           .forEach { (key, value) -> setValue(value, forHTTPHeaderField = key) }
       setCachePolicy(NSURLRequestReloadIgnoringCacheData)


tasomaniac (approved):

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
```diff
@@ -152,6 +152,7 @@ actual class ApolloHttpNetworkTransport(
       val postBody = operation.composeRequestBody(scalarTypeAdapters).toByteArray().toNSData()
       setHTTPMethod("POST")
       headers
+          .plus("Content-Type" to "application/json; charset=utf-8")
```
tasomaniac, line 155: Is this something we can test?

sav007 (approved):

apollo-runtime-kotlin/src/iosMain/kotlin/com/apollographql/apollo/network/http/ApolloHttpNetworkTransport.kt
martinbonnin: I must admit I'm a bit lazy to add a test there. The only thing I can think of is to use a mock server and check the headers but that feels like a lot of code for something that most likely won't be changing too much?

tasomaniac: 👌

PR #2692, opened 2020-10-23
Add support for `@deprecated` directive without any deprecation reason

### LINKED ISSUE #2664: [Codegen] support @deprecated with default reason
The GraphQL spec [specifies](https://spec.graphql.org/June2018/#sec--deprecated) a default value of "No longer supported" for the `@deprecated` directive. Currently, if that happens, this generates `@Deprecated` in kotlin code without a message, which fails to compile.

martinbonnin:
deprecatedReason will fall back to 'No longer supported'

I also realised that we support `@deprecated` on input fields and arguments, making `apollo-android` ahead of the [spec](https://github.com/graphql/graphql-spec/pull/525) :)

Fixes #2664

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/AST.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/AST.kt
@@ -80,8 +80,7 @@ internal data class ObjectType(
       val type: FieldType,
       val description: String,
       val isOptional: Boolean,
-      val isDeprecated: Boolean,
-      val deprecationReason: String,
+      val deprecationReason: String?, // null if not deprecated
       val arguments: Map<String, Any?>,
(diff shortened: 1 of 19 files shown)


tasomaniac (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/AST.kt
```diff
@@ -80,8 +80,7 @@ internal data class ObjectType(
       val type: FieldType,
       val description: String,
       val isOptional: Boolean,
-      val isDeprecated: Boolean,
-      val deprecationReason: String,
+      val deprecationReason: String?, // null if not deprecated
```
sav007, line 83: 👍

sav007 (approved):

PR #2697, opened 2020-10-25
Change the way how name for fragment types generated

sav007:
- Use type condition as class / interface name with suffix of parent type for named fragments.
- Use root type with `Impl` suffix as name for default implementation of named fragment.


martinbonnin (approved):
👍

PR #2701, opened 2020-10-26
Rework backend compiler

sav007:
Rework backend compiler

- Use class nesting instead of flattening class and resolve name collision
- Address issue with nested fragments and merge fields
- Move code responsible for codegen to backend folder 

Notes: 
- any optimizations such as fragment delegation is switched off, will be addressed in next PR
- code generation for default named fragment implementation also commented out


olivierg13:
Checked out the dev-3.x branch to try it out and realized that a List of Fragment in a List of Fragment wasn't codegen-ing the second Fragment interface's Node. Not sure if this PR is fixing it?

For example, given a query:
```
query search(...) {
  id
  sections(...) {
    edges {
      node {
        ...VideoList
      }
    }
}
fragment VideoList on VideoListSection {
  id
  entities(...) {
    edges {
      node {
        ...VideoFields
      }
    }
  }
}
fragment VideoFields on Video {
  id
  displayString
}
```

The VideoList's Node will be generated as:
```
interface Node {
    val __typename: String

    fun marshaller(): ResponseFieldMarshaller
  }
```
without any reference to the VideoFields fragment. 

Any idea why? Or something I'm doing wrong? 😁

sav007:
@olivierg13 yeah we noticed that issue too, that was the reason why  we rewrote our codegen from scratch (again). We might fix this issue with this PR. Will double check.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/GraphQLCompiler.kt
```diff
@@ -98,8 +98,8 @@ class GraphQLCompiler(val logger: Logger = NoOpLogger) {
     }.map { it.name } + ScalarType.ID.name)
         .supportedTypeMap(userCustomTypesMap, generateKotlinModels)
 
-    GraphQLKompiler(
-        ir = ir,
+    GraphQLCodeGenerator(
```
martinbonnin, line 101: Yay for better distinction between compiler and code generator 🙌

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -1,6 +1,6 @@
-package com.apollographql.apollo.compiler.ast
+package com.apollographql.apollo.compiler.backend.ast
 
-internal typealias ObjectTypeContainer = Map<CodeGenerationAst.TypeRef, CodeGenerationAst.ObjectType>
+import com.apollographql.apollo.compiler.parser.introspection.IntrospectionSchema
 
 internal typealias CustomTypes = Map<String, CodeGenerationAst.CustomType>
```
martinbonnin, line 6: Can we add a top level Kdoc block to explain what the backend `Ast` is? I'd expect it to be very close to the Kotlin language but the `Operation` and `Fragment` below let me think it's not?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -27,37 +27,32 @@ internal data class CodeGenerationAst(
...
-      val deprecationReason: String,
+      val deprecationReason: String?,
       val fields: List<Field>,
+      val nestedObjects: List<ObjectType>,
       val implements: Set<TypeRef>,
-      val schemaType: String?,
       val kind: Kind,
-      val typeRef: TypeRef
+      val typeRef: TypeRef,
+      val introspectionSchemaType: IntrospectionSchema.TypeRef?,
```
martinbonnin, line 55: It looks like this is only used in `ObjectType.primaryConstructorSpec`, can we use `introspectionSchemaType.rawType.name` instead ? That'll make the Ast more independant of the introspection.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -82,8 +77,7 @@ internal data class CodeGenerationAst(
       val schemaName: String,
       val type: FieldType,
       val description: String,
-      val deprecated: Boolean,
-      val deprecationReason: String,
+      val deprecationReason: String?,
```
martinbonnin, line 80: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
```diff
@@ -0,0 +1,116 @@
...
+      val name: String,
+      val alias: String?,
+      val type: IntrospectionSchema.TypeRef,
+      val args: List<Argument>,
+      val fields: List<Field>,
+      val fragments: List<InlineFragment>,
+      val conditions: List<Condition>,
+      val description: String,
+      val deprecationReason: String?,
+      val selectionKeys: Set<SelectionKey>,
```
martinbonnin, line 54: Shouldn't there be a list of named fragments there as well? Or do you map the named fragments to inline fragments since they're quite similar?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/FieldMergeUtils.kt
```diff
@@ -0,0 +1,64 @@
...
+      } else {
+        field.merge(fieldToMerge)
+      }
+    } + fieldsToAdd
+  }
+
+  private fun BackendIr.Field.merge(otherField: BackendIr.Field): BackendIr.Field {
+    val mergedFields = this.fields.mergeFields(otherField.fields)
+    return this.copy(
+        args = (this.args + otherField.args).distinct(),
```
martinbonnin, line 21: This feels weird. I'd expect the arguments to be the same at this point. Should we throw if they're not?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -0,0 +1,658 @@
...
+    return this
+        .flatMap { fragment -> fragment.possibleTypes }
+        .toSet()
+        .map { possibleType ->
+          possibleType to this.filter { fragment ->
+            fragment.possibleTypes.contains(possibleType)
+          }
+        }.fold(emptyMap()) { acc, (possibleType, fragments) ->
+          acc + (fragments to (acc[fragments]?.plus(possibleType) ?: listOf(possibleType)))
+        }
```
martinbonnin, line 639: At some point we need to restrict the possible types to the type in scope. Imagine the following. schema:

```
interface Being
interface HasLegs

type Human implements Being, HasLegs
type Droid implements Being
type Table implements HasLegs

query {
  hero {
    ... on HasLegs
  }
}
```

We will end up taking `Table` in consideration in the `fromResponse.when` statement

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
sav007: I will think if we can remove AST and directly generate code from BackendIR

sav007: Actually I will take this back. We do some `BackendIr lowering` here so we should keep it.

apollo-compiler/src/test/graphql/com/example/inline_frgament_intersection/TestOperation.kt
```diff
@@ -0,0 +1,386 @@
+// AUTO-GENERATED FILE. DO NOT MODIFY.
+//
+// This class was automatically generated by Apollo GraphQL plugin from the GraphQL queries it found.
+// It should not be modified by hand.
+//
+package com.example.inline_frgament_intersection
```
martinbonnin, line 6: Typo: `inline_fragment_intersection`

martinbonnin (approved):

PR #2711, opened 2020-10-29
[Compiler] support multiple arguments in directives

### LINKED ISSUE #2708: Directive with multiple arguments throws DocumentParseException
**Summary**
It appears that apollo-android does not support directives with more than one argument.

**Version**
2.4.1
2.4.2-SNAPSHOT

**Description**
While this is not compliant, let's say a JSON schema contains something like this
```
"directives": [
  {
    "name": "customDirective",
    "locations": [
      "QUERY"
    ],
    "args": [
      {
        "name": "argument1",
      },
      {
        "name": "argument2",
      }
    ]
  }
]
```
The compiler fails with
```
Unsupported token `argument2`
----------------------------------------------------
[1]:query exampleQuery($someValue: String $argument1: String $argument2: String)
[2]:@customDirective(argument1: $argument1 argument2: $argument2) {
[3]:example(
----------------------------------------------------
```
While looking through the code to try to gain some insight, I noticed that `apollo-compiler/src/main/antlr/com/apollographql/apollo/compiler/parser/antlr/GraphQL.g4` contains
```
directive
   : '@' NAME ':' valueOrVariable | '@' NAME | '@' NAME '(' argument ')'
   ;
```
I believe `argument` should be plural since `args` in the schema is a JSON array, and 2.12 of the GraphQL specification shows that multiple arguments are allowed. 

I updated the above to be `arguments` and made some quick changes to `GraphQLParser.DirectivesContext?.parse()` function in `GraphQLDocumentParser.kt` so it would compile and test this idea.

I was able to compile everything, except the Gradle plugin, so I could not personally verify if this actually fixes the problem I am facing.

Hoping that someone can verify that this is a bug, and provide a fix, or provide some guidance on how to work around this.

Thanks!

martinbonnin:
closes #2708

### DIFF
--- a/apollo-compiler/src/main/antlr/com/apollographql/apollo/compiler/parser/antlr/GraphQL.g4
+++ b/apollo-compiler/src/main/antlr/com/apollographql/apollo/compiler/parser/antlr/GraphQL.g4
@@ -96,7 +96,7 @@ directives
    ;
 
 directive
-   : '@' NAME ':' valueOrVariable | '@' NAME | '@' NAME '(' argument ')'
+   : '@' NAME arguments?
    ;
 
 typeCondition
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
@@ -361,18 +361,30 @@ class GraphQLDocumentParser(
   private fun GraphQLParser.DirectivesContext?.parse(): List<Condition> {
     return this?.directive()?.mapNotNull { ctx ->
       val name = ctx.NAME().text
-      val argument = ctx.argument()
-      when {
-        argument == null -> null
-        name != "skip" && name != "include" -> null
-        argument.NAME()?.text != "if" || argument.valueOrVariable()?.variable() == null -> null
-        else -> Condition(
-            kind = "BooleanCondition",
-            variableName = argument.valueOrVariable().variable().NAME().text,
-            inverted = ctx.NAME().text == "skip",
-            sourceLocation = SourceLocation(argument.valueOrVariable()?.start ?: argument.NAME().symbol)
-        )
+      val arguments = ctx.arguments()
+
+      if (name != "skip" && name != "include" ) {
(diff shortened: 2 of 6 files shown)


apollo-compiler/src/main/antlr/com/apollographql/apollo/compiler/parser/antlr/GraphQL.g4
```diff
@@ -96,7 +96,7 @@ directives
    ;
 
 directive
-   : '@' NAME ':' valueOrVariable | '@' NAME | '@' NAME '(' argument ')'
+   : '@' NAME arguments?
```
martinbonnin, line 99: Not sure what the ':' was up there but I didn't find it in http://spec.graphql.org/draft/#Directives so I'm assuming it's safe to remove that case.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
```diff
@@ -361,18 +361,30 @@ class GraphQLDocumentParser(
...
+
+      if (arguments.argument().size != 1) {
+        // skip and include should only have a single argument
+        return@mapNotNull null
       }
+
+      val argument = arguments.argument().first()
+      if (argument.NAME().text != "if" || argument.valueOrVariable().variable() == null) {
+        // The argument should be named "if"
+        // Not 100% why we ignore non-variable arguments
```
martinbonnin, line 379: It looks like a field with `@include(if: false)` will still generate a non-nullable field. That sounds like a bug?

sav007: Yeah look like a bug to me. Just I guessed we missed it.

sav007 (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/graphql/GraphQLDocumentParser.kt
martinbonnin: 👍 will fix in a follow-up PR

PR #2723, opened 2020-11-03
Adds documentation whem using apollo with the -native-mt branch of co…

enyciaa:
…routines

### DIFF
--- a/docs/source/essentials/get-started-multiplatform.mdx
+++ b/docs/source/essentials/get-started-multiplatform.mdx
@@ -36,6 +36,17 @@ kotlin {
 }
 ```
 
+## Multithreaded coroutines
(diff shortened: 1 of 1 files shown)


docs/source/essentials/get-started-multiplatform.mdx
```diff
@@ -36,6 +36,17 @@ kotlin {
...
 ```
 
+## Multithreaded coroutines
+If using the x.y.z-native-mt branch of coroutines, gradle will replace the -native-mt version with the non-mt version as outlined [here](https://kotlinlang.org/docs/mobile/concurrency-and-coroutines.html#multithreaded-coroutines). To prevent this happening use the following:
+
+```
+implementation ("org.jetbrains.kotlinx:kotlinx-coroutines-core:x.y.z-native-mt"){
+    isForce = true
+}
+// Must come after the coroutines dependency
```
martinbonnin, line 46: I wouldn't expect the order of the declaration to be important. What happens if you swap the lines?

enyciaa: You're right! I was misunderstanding the gradle docs on isForce

martinbonnin (approved):
Thanks a lot for the contribution!

PR #2737, opened 2020-11-06
[Gradle plugin] Add schema upload

martinbonnin:
add `./gradlew uploadApolloSchema --key $key --schema schema.sdl` to upload to the apollo graph manager

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloUploadSchemaTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloUploadSchemaTask.kt
@@ -6,21 +6,59 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.options.Option
+import java.io.File
 
 abstract class ApolloUploadSchemaTask : DefaultTask() {
-  @get:Optional
-  @get:Input
-  @get:Option(option = "endpoint", description = "url of the GraphQL endpoint")
-  abstract val endpoint: Property<String>
-
   @get:Input
   @get:Optional
-  @get:Option(option = "schema", description = "path where the schema will be downloaded, relative to the current working directory")
+  @get:Option(option = "schema", description = "schema to upload as SDL")
   abstract val schema: Property<String>
+
+  @get:Optional
+  @get:Input
+  @get:Option(option = "key", description = "The Apollo API key. See https://www.apollographql.com/docs/studio/api-keys/ for more information on how to get your API key.")
+  abstract val key: Property<String>
+
+  @get:Optional
+  @get:Input
+  @get:Option(option = "graph", description = "The identifier of the Apollo graph used to download the schema.")
+  abstract val graph: Property<String>
+
+  @get:Optional
+  @get:Input
+  @get:Option(option = "graphVariant", description = "The variant of the Apollo graph used to download the schema.")
+  abstract val graphVariant: Property<String>
+
+
+  private fun Property<String>.orProperty(name: String) = orElse(project.provider {
+    (project.findProperty("com.apollographql.apollo.$name") as? String)?.also {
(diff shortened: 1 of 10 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaUploader.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.apollographql.apollo.compiler.fromJson
+import com.apollographql.apollo.gradle.internal.SchemaDownloader.cast
+import java.net.InetAddress
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+import java.util.UUID
+
+object SchemaUploader {
+  fun uploadSchema(key: String, graph: String, variant: String, sdl: String) {
```
zionts, line 11: I think `key`, `graph`, `variant`, and `sdl` should all be required String types here

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaHelper.kt
```diff
@@ -0,0 +1,60 @@
...
+import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.Response
+import okio.buffer
+import okio.sink
+import java.io.ByteArrayOutputStream
+import java.util.concurrent.TimeUnit
+
+internal object SchemaHelper {
+  private fun newOkHttpClient(): OkHttpClient {
+    val connectTimeoutSeconds = System.getProperty("okHttp.connectTimeout", "600").toLong()
```
zionts, line 17: it may be worth having some kind of automatic retry on disconnect, but 🤷

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloUploadSchemaTask.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    check (graph != null) {
+      "please define graph"
+    }
+
+    check (schema != null) {
+      "please define schema"
+    }
+
+    SchemaUploader.uploadSchema(key = key, graph = graph, variant = graphVariant ?: "current", File(schema).readText())
```
zionts, line 62: Could we call the action `pushSchema` instead of `uploadSchema`? Though we call it `uploadSchema` in our API, in our public documentation and existing CLI, this is called "publishing to the registry," "pushing to the registry" and `service:push`, and I want to make sure we're telling a consistent message to users since the `upload` will actually correspond the uploaded schema as **active** on the graph@variant in question.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaHelper.kt
martinbonnin: I'll keep it simple for now. Given that it's mostly used from the cli, it shouldn't be too hard to relaunch the command. But if disconnects happen too often, it shouldn't be too hard to do.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloUploadSchemaTask.kt
martinbonnin: Makes sense, I'll change that.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/SchemaUploader.kt
martinbonnin: I think that's the case already?

zionts: oh heh, I was reading this as GraphQL, not Kotlin 😂  my bad!

PR #2752, opened 2020-11-12
Fix potential crash on Android regarding forEach usage

tasomaniac:
Fix potential crash on Android regarding forEach usage


martinbonnin:
Oh my, Thanks! Did you catch it with lint or looking at the source? I tried setting up lint some time ago but that didn't go very far...

martinbonnin (approved):

tasomaniac:
There was a warning on Android Studio UI while browsing

PR #2755, opened 2020-11-16
Fix error.NonExistentClass in Metalava Signature Files

tylerbwong:
This fixes most of the `error.NonExistentClass` occurrences in each module's Metalava signature descriptors.

A [standalone Gradle plugin](https://github.com/tylerbwong/metalava-gradle) has been added which built on top of this project's build logic, but adds the proper `--classpath` flag arguments for Android, Multiplatform, and Java libraries and other configurations.

There are a few notable changes to signature descriptor files since the Gradle plugin uses an embedded Metalava artifact from [here](https://maven.google.com/web/index.html#com.android.tools.metalava:metalava:1.0.0-alpha01):

* Non-abstract functions seem to be omitted from public sub-class API.
* There is a [bug](https://issuetracker.google.com/issues/161153790) with `@JvmSynthetic` targets being omitted from the output. (Found in `rx2`/`rx3` artifacts)
* More nullability information is included.
* Empty Kotlin constructors from top-level only files are removed.
* Lots of annotation information is now included.


build.gradle.kts
```diff
@@ -102,8 +102,8 @@ fun Project.configurePublishing() {
   if (javadocTask == null && android != null) {
     // create the Android javadoc if needed
     javadocTask = tasks.create("javadoc", Javadoc::class.java) {
-      source = android.sourceSets["main"].java.sourceFiles
-      classpath += project.files(android.getBootClasspath().joinToString(File.pathSeparator))
+      source = android.sourceSets["main"].java.getSourceFiles()
```
martinbonnin, line 105: Any reason not to use the property syntax?

tylerbwong: I was also confused about this, but it seems to be a result of `AndroidSourceDirectorySet` being converted to Kotlin in `4.1.1` of the Android Gradle Plugin. [Source](https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/api/AndroidSourceDirectorySet.kt).

martinbonnin:
Thanks a lot for the contribution, that looks awesome ! I'll look into the details of the plugin source code later this week.

martinbonnin:
CI is a bit flaky at the moment, I'm not sure why but `:apollo-integration:jvmTest` sometimes stay stuck for 6h. I just relaunched it

tylerbwong:
There seems to be an issue with the new check compatibility task :disappointed:. It looks like the temporary generated API is incorrect. I'll look into fixing it.

gradle/dependencies.gradle
```diff
@@ -1,6 +1,6 @@
 def versions = [
     minAndroidPlugin      : '3.4.2',
-    androidPlugin         : '3.6.2',
+    androidPlugin         : '4.1.1',
```
martinbonnin, line 3: Can we keep AGP `3.6.2` for now as IntelliJ cannot open projects using `4.1.1` and IntelliJ deals with the multiplatform part better than Android Studio. 
I don't think there's any drawback keeping the old version of the AGP. It's mainly a `compile-only` dependency so it'll be substituted at runtime for the actual version.

Tyler-Wong-CK: Sounds good, I'll revert it back. The main reason why I upgraded it was because the Gradle plugin was transitively pulling it in anyway, but that's likely an issue on my part. I'll also probably need to make it a `compileOnly` dependency for the plugin itself.

tylerbwong: Sounds good, I'll revert it back. The main reason why I upgraded it was because the new Gradle plugin was transitively pulling it in anyway, but that's likely an issue on my part. I'll also probably need to make it a compileOnly dependency for the plugin itself.

tylerbwong:
Turns out the build was failing because I generated the signature files on a Linux machine and the multiplatform module `apollo-normalized-cache-sqlite/api.txt` was inconsistent with the macOS image CI was generating. It should be fixed by 88ac699!

tylerbwong:
@martinbonnin Just updated and released `0.1.4` of the Metalava Gradle plugin which makes the changes we discussed today (See https://github.com/tylerbwong/metalava-gradle/commit/55898039a1c345dc5815151659ef48a076d075ac). Let me know if there's anything else that needs to be done! :smile:

martinbonnin:
Awesome, looks good to me 👍 . I'll keep this open for the week end if anyone wants to look into it and merge on Monday. Thanks a lot for the contribution !

martinbonnin (approved):

Issue #2727, opened 2020-11-06
Default string values in mutation inputs are not escaped correctly


joeboyscout04:
**Summary**
When using `defaultValues` to generate Input<String> objects for mutations, the quotations are not un-escaped correctly, so the default value for the input contains quotations where there shouldn't be any. 

This affects the Kotlin code generation (possibly the Java one also, although I haven't tested that one). 

**Version**
Latest version (master)

**Description**
For example, in this schema (this is an Apollo example), there is a default value of ` "defaultValue": "\"Me\""`.

This server would be expecting a default value of `Me`.  However, when the schema is parsed into Kotlin, the default value is `Input.optional("\"Me\"")`, which would send a default value of `"Me"` (with quotations) to the server. 

```
 {
          "kind": "INPUT_OBJECT",
          "name": "BookInput",
          "description": null,
          "fields": null,
          "inputFields": [
            {
              "name": "title",
              "description": null,
              "type": {
                "kind": "NON_NULL",
                "name": null,
                "ofType": {
                  "kind": "SCALAR",
                  "name": "String",
                  "ofType": null
                }
              },
              "defaultValue": null
            },
            {
              "name": "author",
              "description": null,
              "type": {
                "kind": "NON_NULL",
                "name": null,
                "ofType": {
                  "kind": "SCALAR",
                  "name": "String",
                  "ofType": null
                }
              },
              "defaultValue": "\"Me\""
            }
          ],
          "interfaces": null,
          "enumValues": null,
          "possibleTypes": null
        },
```

I've reproduced this in the sample project here: https://github.com/joeboyscout04/apollo-android/tree/default-values-string

I've added a mutation like so:
```
mutation deleteBranch($input: DeleteBranchProtectionRuleInput!) {
	deleteBranchProtectionRule(input: $input) {
		clientMutationId
	}
}
```

And I've updated the schema (on `clientMutationId`...this was the easiest parameter to change) to reflect the default values as described above. 

The generated code resulting shows that the Input<String> for `clientMutationId` would incorrectly include quotations.

```
  val clientMutationId: Input<String> = Input.optional("\"Me\"")
```

I would expect that instead, the default value would be generated as:
```
val clientMutationId: Input<String> = Input.optional("Me")
```

martinbonnin:
Hi 👋  Thanks for sending this! 
I believe this is working as intended? Json can represent string values so there should be no need for additional escaping? Can you point to specifications or other implementation that do this un-escape?

[...]

Issue #2572, opened 2020-09-11
[Compiler] Improve validation of input objects


martinbonnin:
Given the following schema:

```
enum SomeEnum {
  ENUM1,
  ENUM2
}

input SomeInput {
   mandatoryInt: Int
   mandatoryEnum: SomeEnum
}

type Query {
  randomInt(input: SomeInput): Int
}
```

The following queries fail validation on the server 

```
query getRandomInt {
  # missing mandatoryEnum field
  randomInt({mandatoryInt: 0}) 
}

query getRandomInt {
  # mandatoryEnum is not a valid SomeEnum
  randomInt({mandatoryInt: 0, mandatoryEnum: FOO}) 
}
```

These errors could be caught client-side during compilation

martinbonnin:
Closing as it is fixed in `dev-3x`

PR #2756, opened 2020-11-16
Use GQLDocument for validation/IR building

martinbonnin:
Use GQLDocument for building the IR. GQLDocument maps 1:1 to the GraphQL (except for whitespaces) so it's easier to modify the source. Using it for validation also allows for a better separation of the different stages of the parsing:

```
Antlr parsing
⬇️ 
GQLDocument construction
⬇️ 
Validation
⬇️ 
Associate operation with their named fragments
⬇️ 
Add typenames when needed
⬇️ 
Generate document source
⬇️ 
Compute used types
⬇️ 
Generate IR
```

The produced IR is not 100% compatible yet because it does not attempt to merge fields (`hero_name` test is disabled for an example).

This PR adresses the following issues that were not merged from main in the [last merge](https://github.com/apollographql/apollo-android/pull/2724):

4fa0158f8c88d27069b55b7b06b7df6bcf501c1d default deprecationReason
60d7e1c662f92cc111c45dcfcf9a95e8505c69b7 optional mutations/subscriptions
f3dca0383980359ff453749a74748f140daf8eb3 type extensions
c330ab4decbaeca6bd6bfe8680ea5232599d7a66 @deprecated(reason: "")

Remaining features/commits that needs to be ported/decided in 3.x:
d9039bf023b48c0758ec448d47c0854dc8dabfbe default variable values
019edd04fdfe51f107061d0d74cb5025d58f2528 fragment package name

Fixes https://github.com/apollographql/apollo-android/issues/2727
Fixes https://github.com/apollographql/apollo-android/issues/2572


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/CodeGenerationAst.kt
```diff
@@ -43,6 +43,7 @@ internal data class CodeGenerationAst(
   data class FragmentType(
       val graphqlName: String,
       val rootType: TypeRef,
+      val description: String,
```
sav007, line 46: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/gql/Schema.kt
```diff
@@ -0,0 +1,56 @@
+package com.apollographql.apollo.compiler.parser.gql
+
+/**
+ * a very thin wrapper around a schema GQLDocument
+ *
+ * It serves as a common ground between GQLDocument and IntrospectionSchema
+ *
+ * Schema should always contain all types, including the builtin ones
+ */
+class Schema(
```
sav007, line 10: nit: should we call it `GQLSchema`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/gql/add_typename.kt
```diff
@@ -0,0 +1,57 @@
...
+      is GQLFragmentSpread -> it
+      is GQLField -> it.copy(
+          selectionSet = it.selectionSet?.withTypenameWhenNeeded(schema, true)
+      )
+    }
+  }
+
+  newSelections = if (needed) {
+    // remove the __typename if it exists
+    // and add it again at the top so we're guaranteed to have it at the beginning of json parsing
```
sav007, line 48: 👍 👍 👍  very important for fragment parsing

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ast/CodeGenerationAst.kt
martinbonnin: Once this is merged and used, I think we should try to upstream this inside the GraphQL spec so that other tools don't choke on it.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/gql/Schema.kt
martinbonnin: I'd like to keep the `GQL` prefix for everything that is in the `GQLNode` type hierarchy and `Schema` doesn't inherit from `GQLNode`.

There should be only 2 instances of `Schema`: one for introspection and this one so I think having `Schema` and `IntrospectionSchema` isn't too confusing but I'm happy to think of another prefix if needed.

apollo-compiler/src/test/typename/fragment_spread.graphql
```diff
@@ -0,0 +1,13 @@
+query TestQuery {
+    hero {
+        ...heroFragment
+    }
+}
+
+fragment heroFragment on Character {
+    # No typename should be appended here
```
sav007, line 8: it should because we have nested inline fragment here `Droid`

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/TypenameTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    @JvmStatic
+    @Parameterized.Parameters(name = "{0}")
+    fun data(): Collection<Array<Any>> {
+      return File("src/test/typename/")
+          .walk()
+          .toList()
+          .filter { it.isFile }
+          .filter { it.extension == "graphql" }
+          .sortedBy { it.name }
+          //.filter { it.name.contains("InputObjectFieldType") }
```
sav007, line 59: remove?

apollo-compiler/src/test/typename/fragment_spread.graphql
martinbonnin: I thought that as well but since the Fragment spread will always be queried on a parent field, the `__typename` will be added there. In that case, the transformed definitions will be:

```graphql
query TestQuery {
    hero {
        __typename
        ...heroFragment
    }
}

# Fragment did'nt change
fragment heroFragment on Character {
    name
    ... on Droid {
        primaryFunction
    }
} 
```

sav007 (approved):

PR #2759, opened 2020-11-17
[codegen] Honor packageName for schema types as well

martinbonnin:
Follow up on https://github.com/apollographql/apollo-android/pull/2669

Since we now have `packageName` for fragments, use it also for the `schemaPackageName` if it's not null (`schemaPackageName` is used for input/enums)


martinbonnin:
@Laimiux mind taking a look? As far as I know, you're one of the early users of `packageName`, does that look ok to you?

Laimiux:
> @Laimiux mind taking a look? As far as I know, you're one of the early users of `packageName`, does that look ok to you?

Hey, thanks for reaching out. I'm a little bit confused what is the difference between `rootPackageName` and `packageName` and how they are used together. I think moving `Input/Enum` types shouldn't be a problem for me.

martinbonnin:
> I'm a little bit confused what is the difference between rootPackageName and packageName and how they are used together.

* `rootPackageName` is the "old" way to do it. The package name would come from the file path and then be prepended with `rootPackageName` but doesn't work for fragments
* `packageName` is new. It works for fragments and forces `packageName` for every query, input type, enum and fragments. 

They're not really meant to be used together. Actually, it might be a good idea to throw an error if that happens since the user will most likely expect something that's not going to happen.

Laimiux:
That makes sense. Currently, I'm setting both values - will remove `rootPackageName`.

Laimiux:
Actually, I will only be able to remove it after this PR since `types` use it.

martinbonnin:
> Actually, I will only be able to remove it after this PR since types use it.

Yup makes a lot of sense 👍 . I hope things will be simpler with just `packageName`

Issue #2744, opened 2020-11-10
Cannot supply default value for Enum input in queries


ReeceLaF:
Starting in version 4.3.2, compilation fails when generating the query class for a query with a default value for an enum input.

eg-
```graphql
query MyQuery($myArg: MyEnum = MY_VALUE) {
  mySearch(myArg: $myArg) {
    myResult
  }
}
```

The compiler fails with the following error: `Cannot access '<init>': it is internal in 'Input'`. Apollo is generating the code incorrectly as
```kotlin
val myArg: Input<MyEnum> = Input.optional(Input<MyEnum>.safeValueOf("MY_VALUE"))
```
when it should be
```kotlin
val myArg: Input<MyEnum> = Input.optional(MyEnum.safeValueOf("MY_VALUE"))
```

martinbonnin:
Thanks for sending this. This was introduced by an attempt to use Kotlin default parameter to represent variable default GraphQL values (#2704). It turns out this was a [relatively bad idea](https://github.com/apollographql/apollo-android/pull/2741). Sorry about that. It's now disable in main, I'll do a version in a bit.

martinbonnin:
`2.4.4` is now released, can you check it's working again?

ReeceLaF:
It is working now on `2.4.4`. Thank you!

Issue #2742, opened 2020-11-10
generateMainServiceApolloSources fails when an Int has a defaultValue of null


maxgdn:
**Summary**

When generating the models from a directory of schema files the generateMainServiceApolloSources task fails with

```
Failed to parse GraphQL file /home/user/project1/src/main/graphql/FooBar.graphql (1:57)
  Can't parse `Int` value
```

**Version**
2.4.3

**Description**

Error does not appear on 2.4.1. Suspected that #2704 is related.

*Reproduction*

1. run ./gradlew downloadApolloSchema
2. Have a query file where an Int parameter is defaulted to null
3. Run generateKotlinModels.set(true) / generateMainServiceApolloSources

FooBar.graphql (getFoooo is a pseudo name)
```
query getFoooooBarrAbcde12345678($id: ID!, $first: Int = null, $after: String = null) {
   ...
}
```

- Changing $first: Int = null to  $first: Int = 0 resolves error.
- According to [spec](https://spec.graphql.org/October2016/#sec-Types.Non-Null) all types in GraphQL are nullable. If this is not the case for defaults could a specific error message be thrown?
- Using IntelliJ IDEA

*Stacktrace*

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':generateMainServiceApolloSources'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:205)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:203)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:184)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:114)
	at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
[...]

Issue #2686, opened 2020-10-22
Allow to omit non-nullable variables if they have a defaultValue


renclav:
**Is your feature request related to a problem? Please describe.**
I would like to define default values for my query variables inside the query definitions defined in .graphql file/s. Currently this is possible, but the default values are ignored in the generated models.

**Describe the solution you'd like**
The client library follows the spec defined [here](https://graphql.org/learn/queries/#default-variables) which would result in kotlin query models being generated which include default values.

Notes:

1. I use kotlin models and as such, have no awareness of how this may effect java based models.
2. I assume this would be desired for subscriptions and mutations as well (no experience there unfortunately ).

martinbonnin:
@renclav Most of use case should work with #2704. This has been merged and will be available in the `2.4.2-SNAPSHOT` in ~30 min ([ci build](https://github.com/apollographql/apollo-android/actions/runs/333476168)). Any chance you can try the snapshots when they're ready? If everything looks good, I'll make a release next Monday.

In all cases, I'm keeping this issue open (with an updated title) for input object default values, which will require more work.

renclav:
@martinbonnin 2.4.2-SNAPSHOT works well for me 🥳

martinbonnin:
Awesome :). Glad to hear that and perfect timing as I'm preparing the `2.4.2` release. It should be live in a few hours

martinbonnin:
@renclav after some tests, passing the default value as default argument in the model constructor might not be the best way to do it since it will send a GraphQL variable by default, which might be counter intuitive. 

Can you elaborate on why you need the default value in the Kotlin models? Passing `Input.absent()` as was the case before `2.4.2` should still send the default value to your server as part of the GraphQL query. Or did you need the defaultValue somewhere else?

martinbonnin:
I'm going to close this one as not overriding the GraphQL document default variables seems to be the way to go, even for input objects (see #2741). Please leave a message if you want me to reopen it.

martinbonnin:
Reopening this issues as there is more to it. While the solution of passing `Input.absent()` works for variables of nullable types, it doesn't work for variables of non-null type.

```graphql
query launches($id: String! = "83") {
  launch(id: $id) {site}
}
```

will generate 

```kotlin
data class LaunchesQuery(val id: String) { ... }
```

[...]

PR #2760, opened 2020-11-17
[Codegen] Allow default value for non nullable variables

martinbonnin:
See #2686. This allow to omit the variable value if it has a defaultValue (even if the variable is non-nullable)

```
query launches($id: String! = "83") {
  launch(id: $id) {site}
}
```

will generate

```
// by default, id will be omitted 
data class LaunchesQuery(val id: Input<String> = Input.absent()) { ... }
```

I'm opening this mainly for reference. That's a potentially relatively seriously invasive/breaking change that has bitten us in the past (https://github.com/apollographql/apollo-android/issues/2744, https://github.com/apollographql/apollo-android/issues/2742). Given that it can be worked around relatively easily by passing the defaultValue in Kotlin code, and unless there are other important use cases I'm missing, this PR will only be merged for the next major version


martinbonnin:
Closing for now. Will reopen for 3.0

PR #2771, opened 2020-11-25
Open some classes from the apollo-http-cache and apollo-http-cache-api modules

ludovicroland:
The aim of this very little pull request is just to open some classes from the apollo-http-cache and apollo-http-cache-api modules in order to be able to extends the current http cache system with a new custom one.

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloConvertSchemaTask.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloConvertSchemaTask.kt
@@ -9,6 +9,7 @@ import com.apollographql.apollo.compiler.toJson
 import org.gradle.api.DefaultTask
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
@@ -20,11 +21,24 @@ abstract class ApolloConvertSchemaTask: DefaultTask() {
   @get:Option(option = "from", description = "schema to convert from")
   abstract val from: Property<String>
 
-  @get:OutputFile
+  // Even if this is points to an output file, for the purpose of the task this is seen as an input
+  @get:Input
   @get:Option(option = "to", description = "schema to convert to")
   abstract val to: Property<String>
 
+  init {
+    /**
+     * This task is really meant to be called from the command line so don't do any up-to-date checks
+     * If someone wants to register its own conversion task, it should be easy to do using the
+     * compiler APIs directly (see [convert] below)
+     * This code actually redundant because the task has no output but adding it make it explicit.
+     */
+    outputs.upToDateWhen { false }
+    outputs.cacheIf { false }
+  }
+
   private fun File.isIntrospection() = extension == "json"
+
   fun convert(from: File, to: File) {
     check (from.isIntrospection() && !to.isIntrospection() || !from.isIntrospection() && to.isIntrospection()) {
       "Cannot convert from ${from.name} to ${to.name}, they are already the same format"
--- a/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/ConvertSchemaTests.kt
+++ b/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo/gradle/test/ConvertSchemaTests.kt
@@ -26,6 +26,31 @@ class ConvertSchemaTests {
   }
 
   @Test
+  fun `convert is never up-to-date`() {
+    TestUtils.withTestProject("convertSchema") { dir ->
+      val from = File(dir, "schemas/schema.sdl")
+      val to = File(dir, "schema.json")
+      var result = TestUtils.executeTask("convertApolloSchema",
+          dir,
+          "--from",
+          from.absolutePath,
+          "--to",
+          to.absolutePath
+      )
+      Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":convertApolloSchema")!!.outcome)
+      result = TestUtils.executeTask("convertApolloSchema",
+          dir,
+          "--from",
+          from.absolutePath,
+          "--to",
+          to.absolutePath
+      )
+      // even if inputs are the same,
+      Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":convertApolloSchema")!!.outcome)
+    }
+  }
+
+  @Test
   fun `convert from Json to SDL works`() {
     TestUtils.withTestProject("convertSchema") { dir ->
       val from = File(dir, "schemas/schema.json")


martinbonnin:
Looks good to me! Of course this increases the public API surface which means more API to maintain and keep backward compatible. Do you mind sharing your use cases? Maybe there's a way to do what you need without making everything open?

ludovicroland:
> Looks good to me! Of course this increases the public API surface which means more API to maintain and keep backward compatible. Do you mind sharing your use cases? Maybe there's a way to do what you need without making everything open?

The CI has failed but I am not sure that is because my changes.

Because I am in love with the http cache implemented on Apollo, I would like to wrap it in order to be able to use it directly on okhttp/retrofit with a new feature: a fake delete request that can delete all or part of the cache.

1. I order to implement this feature, I need to override the `intercept` method from the `HttpCacheInterceptor` class in order to catch/evaluate new headers.

2. I need to open `ExpirePolicy` the class and change the visibility of its constructor in order to implement a new Detele Http Cache Policy.

3. And finally, I need to extend the `ApolloHttpCache` in order to override the `interceptor` method to return my own `HttpCacheInterceptor` from the step 1.

martinbonnin:
> The CI has failed. I need to update the api.txt files too.

You can do that with `./gradlew generateMetalava`

martinbonnin:
Thanks for sending the use case! I'm guessing you need a fake delete request to avoid keeping a reference to `DiskLruCache`? Else you could always delete from the cache by doing `diskLruCache.remove(cacheKey)`.

ludovicroland:
> You can do that with `./gradlew generateMetalava`

Unfortunately the task is not found. Should I need to install external tools to execute it?

> Thanks for sending the use case! I'm guessing you need a fake delete request to avoid keeping a reference to `DiskLruCache`? Else you could always delete from the cache by doing `diskLruCache.remove(cacheKey)`.

Yes, that's the idea. I guess I can keep a reference on the `DiskLruCache` using the the `create` static method. But having the possibility to extends all these classes in order to implement my logic allows me to really have a centralized way to manage my http cache.

martinbonnin:
> Unfortunately the task is not found. Should I need to install external tools to execute it?

Arg, sorry about this, I forgot this changed recently (see https://github.com/apollographql/apollo-android/pull/2755). This should be `metalavaSignaturePublish`

> Yes, that's the idea. I guess I can keep a reference on the DiskLruCache using the the create static method. But having the possibility to extends all these classes in order to implement my logic allows me to really have a centralized way to manage my http cache.

Got it. That works for me 👍

martinbonnin (approved):

ludovicroland:
> Arg, sorry about this, I forgot this changed recently (see #2755). This should be `metalavaSignaturePublish`

I am not sure that the result on my computer is the expected one. When I execute the `metalavaSignaturePublish` gradle task, all the `api.txt` file are deleted.

martinbonnin:
> I am not sure that the result on my computer is the expected one. When I execute the metalavaSignaturePublish gradle task, all the api.txt file are deleted.

Well, that's unexpected... @tylerbwong any idea what could happen there?

ludovicroland:
@tylerbwong : 

For each module I have the following kind of message :

> Aborting: C:\Users\rolan\.gradle\caches\modules-2\files-2.1\io.reactivex.rxjava3\rxjava\3.0.7\c7fe5bf1d1d1088941f34cf4e37e1edf219582b5\rxjava-3.0.7.jar:C:\Users\rolan\Documents\workspace\apollo-android-fork\apollo-runtime\build\classes\java\main:C:\Users\rolan\Documents\workspace\apollo-android-fork\apollo-normalized-cache\build\libs\apollo-normalized-cache-jvm-2.4.6-SNAPSHOT.jar:C:\Users\rolan\Documents\workspace\apollo-android-fork\apollo-normalized-cache-api\build\libs\apollo-normalized-cache-api-jvm-2.4.6-SNAPSHOT.jar:C:\Users\rolan\Documents\workspace\apollo-android-fork\apollo-api\build\libs\apollo-api-jvm-2.4.6-SNAPSHOT.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk8\1.4.10\998caa30623f73223194a8b657abd2baec4880ea\kotlin-stdlib-jdk8-1.4.10.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.reactivestreams\reactive-streams\1.0.3\d9fb7a7926ffa635b3dcaa5049fb2bfa25b3e7d0\reactive-streams-1.0.3.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\com.squareup.okhttp3\okhttp\3.12.11\301adbdad6ff9d613a2ba4772443560d25df5538\okhttp-3.12.11.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-jdk7\1.4.10\30e46450b0bb3dbf43898d2f461be4a942784780\kotlin-stdlib-jdk7-1.4.10.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\com.squareup.okio\okio\2.9.0\dcc813b08ce5933f8bdfd1dfbab4ad4bd170e7a\okio-jvm-2.9.0.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib\1.4.10\ea29e063d2bbe695be13e9d044dcfb0c7add398e\kotlin-stdlib-1.4.10.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.jetbrains.kotlin\kotlin-stdlib-common\1.4.10\6229be3465805c99db1142ad75e6c6ddeac0b04c\kotlin-stdlib-common-1.4.10.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\org.jetbrains\annotations\13.0\919f0dfe192fb4e063e7dacadee7f8bb9a2672a9\annotations-13.0.jar:C:\Users\rolan\.gradle\caches\modules-2\files-2.1\com.benasher44\uuid-jvm\0.2.0\4f45ddff7626ca3342c80897afa235cdbf540ba2\uuid-jvm-0.2.0.jar is not a jar or directory

martinbonnin:
This seems to be a Windows thing indeed, it works on my Mac. I pushed the api.txt to https://github.com/hagergroup/apollo-android/pull/1 if you want to take them while we investigate, it should make CI pass for this time.

vkondrav:
If we're okay with opening up some classes in this module can I request we open up the [`LruNormalizedCache`](https://github.com/apollographql/apollo-android/blob/main/apollo-normalized-cache/src/jvmMain/kotlin/com/apollographql/apollo/cache/normalized/lru/LruNormalizedCache.kt) also? use case here in this [slack thread](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1605834318015400)

martinbonnin:
> If we're okay with opening up some classes in this module can I request we open up the LruNormalizedCache also? use case here in this slack thread

HI @vkondrav ! I'd argue that there's slightly more logic in this specific case (~600 lines to copy/paste vs ~100 for LruNormalizedCache) and that increasing the API surface always has a maintenance cost in the long run. 

It's definitely not a hill I want to die on though! Feel free to make a separate PR and we can follow up there.

tylerbwong:
@ludovicroland @martinbonnin so sorry for the inconvience! It looks like Metalava doesn't like the JAR paths being passed to it since it uses `:` to separate classpaths, but on Windows `C:` is part of each path. I'll investigate the issue to see if it's something we can fix.

tasomaniac:
In my opinion, if this provides a way to implement a custom caching, I'm all in for it. But the changes seem different (as explained here https://github.com/apollographql/apollo-android/pull/2771#issuecomment-733608771)

This looks like a hacky way to just subclass the official implementation to change the behavior slightly. This might create a lot of maintainability problems because the extension mechanism provided here is extremely coupled with implementation details.

ludovicroland:
> In my opinion, if this provides a way to implement a custom caching, I'm all in for it. But the changes seem different (as explained here [#2771 (comment)](https://github.com/apollographql/apollo-android/pull/2771#issuecomment-733608771))
> 
> This looks like a hacky way to just subclass the official implementation to change the behavior slightly. This might create a lot of maintainability problems because the extension mechanism provided here is extremely coupled with implementation details.

Thank you for your comment. 

Sure, since the `httpCache` attribut of the `ApolloClient` class is an instance of the `HttpCache` interface, it's already possible to use a full custom implementation.

Since the offical implementation through the `ApolloHttpCache` and the `HttpCacheInterceptor` classes is awesome, it's not convenient to not be able to use it in order to extend or change a behavior. Currently, we have to copy/paste the whole code juste to add a feature. It makes me sad :(

tasomaniac:
What about contributing that feature?

ludovicroland:
> What about contributing that feature?

Not sure that everything will fit to apollo, but I can try to draft something yes

PR #2773, opened 2020-11-25
Open lru normalized cache class

vkondrav:
I have a specific requirement that I would like to address using the LruNormalized Cache. I would be overriding the just the merge method to address a specific case. You can see the issue in this [slack thread](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1605834318015400). Currently we are going to pull the whole class into our code base and put in the feature, although it's a working solution it does mean we have to maintain the said class and might miss out on any bug fixes/changes on it from future releases. 

Not a show stopper by any means but I would love to keep our code clean as well.


martinbonnin (approved):

martinbonnin:
@vkondrav can you rebase and update the metalava signatures with `./gradlew metalavaGenerateSignature` ?

martinbonnin:
Closing as there doesn't seem to be a consensus on opening classes by default.  If having the class duplicated bugs you too much, please reopen with the metalava signatures updated and I'll merge. If not, we'll make sure this use case works in 3.0

Issue #2775, opened 2020-11-26
classes generated with `sealedClassesForEnumsMatching` break sqlite based caching


mateuszkwiecinski:
**Summary**
When using `sealedClassesForEnumsMatching` generated classes as query parameters, `apollo-android` library uses their hashcode when generating a cache keys which gets problematic when using cache provided by `com.apollographql.apollo:apollo-normalized-cache-sqlite` 

![image](https://user-images.githubusercontent.com/36954793/100370773-116a1000-3007-11eb-8a9c-5cdd6ffba311.png)
After app restart all generated classes have different hashcodes which result in cache misses. 

**Version**
2.4.5

**Description**
When implementing sqlite based cache I noticed some of the queries are always fetched from the network reporting 100% cache misses. I dug into the problem and noticed that apollo includes class hashcode when generating cache keys. 
This make app to properly cache responses in its memory cache (as long as generated objects are the same instance), but it breaks after app restart. 
I guess generated classes should have hashcode method overriden, or if that's a result of `toString()` maybe it would be worth to have it constant and also generated?

tl;dr
1. Have a graphql query that takes graphql `enum`
2. Use cache from `com.apollographql.apollo:apollo-normalized-cache-sqlite` on Android
3. When fetched library will report 100% cache misses

martinbonnin:
Oh wow, good catch! That's definitely not what we want. The `toString()` is coming from there: https://github.com/apollographql/apollo-android/blob/1fd45e073960ee12b73a4967e26b261ffe14443d/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/Utils.kt#L35

Overriding `toString()` might do the job, I wonder if there's another way to fix it though.

mateuszkwiecinski:
The plugin generates unique `override fun operationId(): String = OPERATION_ID`  for each query. Maybe it could similary generate identifiers for each enum option? 

btw. I just realized including a class name might also be problematic if generated class is obfuscated using R8 🤔 So even if I  (as a workaround) disable the `sealedClassesForEnumsMatching` and use enum instead of an `object` it would include the obfuscated name as a part of a cache key - which might also cause cache misses across different app versions. So id doesn't matter if I use `sealedClassesForEnumsMatching` or not 🤔

martinbonnin:
> The plugin generates unique override fun operationId(): String = OPERATION_ID for each query. Maybe it could similary generate identifiers for each enum option?

[...]

PR #2764, opened 2020-11-21
Increase Gradle verbosity during integration test to investigate Github Action hangs

martinbonnin:

PR #2776, opened 2020-11-26
Add interface for classes generated for enum types and fix sealed classes json representation

lwasyl:
Objects representing enums in sealed classes don't have stable toString() implementation: instead, they'll return the class name along with memory address for that object. This breaks SQLite-based caching for queries that accept enum parameters, as toString() is used to generate the query key, so it would change between app restarts. This commit fixes this behavior by adding an EnumValue interface to types representing enums, and using its `rawValue` property to build proper cache key.

Fixes https://github.com/apollographql/apollo-android/issues/2775

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo.compiler.codegen.kotlin
 
+import com.apollographql.apollo.api.EnumValue
 import com.apollographql.apollo.compiler.applyIf
 import com.apollographql.apollo.compiler.ast.EnumType
 import com.squareup.kotlinpoet.*
@@ -21,7 +22,8 @@ private fun EnumType.toEnumTypeSpec(generateAsInternal: Boolean): TypeSpec {
       .enumBuilder(name)
       .applyIf(description.isNotBlank()) { addKdoc("%L\n", description) }
       .applyIf(generateAsInternal) { addModifiers(KModifier.INTERNAL) }
-      .primaryConstructor(primaryConstructorSpec)
+      .primaryConstructor(primaryConstructorWithOverriddenParamSpec)
+      .addSuperinterface(EnumValue::class)
       .addProperty(rawValuePropertySpec)
       .apply {
         values.forEach { value -> addEnumConstant(value.constName, value.enumConstTypeSpec) }
@@ -35,6 +37,12 @@ private val primaryConstructorSpec =
     FunSpec
         .constructorBuilder()
         .addParameter("rawValue", String::class)
+        .build()
+
+private val primaryConstructorWithOverriddenParamSpec =
+    FunSpec
+        .constructorBuilder()
+        .addParameter("rawValue", String::class, KModifier.OVERRIDE)
         .build()
 
 private val rawValuePropertySpec =
@@ -86,7 +94,8 @@ private fun EnumType.toSealedClassTypeSpec(generateAsInternal: Boolean): TypeSpe
       .applyIf(description.isNotBlank()) { addKdoc("%L\n", description) }
       .applyIf(generateAsInternal) { addModifiers(KModifier.INTERNAL) }
       .addModifiers(KModifier.SEALED)
-      .primaryConstructor(primaryConstructorSpec)
+      .primaryConstructor(primaryConstructorWithOverriddenParamSpec)
+      .addSuperinterface(EnumValue::class)
       .addProperty(rawValuePropertySpec)
       .addTypes(values.map { value -> value.toObjectTypeSpec(ClassName("", name)) })
       .addType(unknownValueTypeSpec)
--- a/apollo-compiler/src/test/graphql/com/example/java_beans_semantic_naming/type/Episode.kt
+++ b/apollo-compiler/src/test/graphql/com/example/java_beans_semantic_naming/type/Episode.kt
@@ -5,6 +5,7 @@
 //
 package com.example.java_beans_semantic_naming.type
 
+import com.apollographql.apollo.api.EnumValue
 import kotlin.Deprecated
 import kotlin.String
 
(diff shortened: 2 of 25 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/SealedEnumType.kt
```diff
@@ -0,0 +1,11 @@
+package com.apollographql.apollo.api
+
+/**
+ * Base class for an enum type when generating sealed Kotlin class
+ */
+abstract class SealedEnumType {
```
lwasyl, line 6: I kind of wanted to follow `ScalarType` naming, but of course can change it to whatever is more appropriate.

Also, @martinbonnin suggested using an interface (https://github.com/apollographql/apollo-android/issues/2775#issuecomment-734399072), but I wasn't sure it's okay to have a default method with `toString` implementation

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
```diff
@@ -37,6 +38,12 @@ private val primaryConstructorSpec =
         .addParameter("rawValue", String::class)
         .build()
 
+private val sealedSubclassPrimaryConstructorSpec =
```
lwasyl, line 41: The `primaryConstructorSpec` is used also when generating `enum class` and for `Unknown` case, so we need a new one that overrides the raw value

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/SealedEnumType.kt
martinbonnin: > I kind of wanted to follow ScalarType naming, but of course can change it to whatever is more appropriate.

Naming is hard... I'm personnaly not a fan of `ScalarType` as it acts more like a type-safe `Binding` or `Mapping` between a GraphQL type and a Kotlin one. So I don't mind not following the logic here. 

What about the below:
```
interface EnumValue {
  val rawValue: String
}
```

? This will be closest to the spec: https://spec.graphql.org/draft/#EnumValue

> And if the idea was to add a case here then of course I can also do that, it's just overriding toString feels like a good solution

It was the original intent. I feel it's dangerous to rely on `toString()` because it typically exposes to these kind of bugs that we're seeing here. The compiler cannot catch if a new type is added a someone forgets to override `toString()`. 

Ultimately, I was thinking of moving this:
```
fun writeToJson(value: Any?, jsonWriter: JsonWriter) {
```
to something more like:
```
fun writeToJson(value: InputValue, jsonWriter: JsonWriter) {
```

where `InputValue` would be an instance of an InputType as defined by https://spec.graphql.org/draft/#IsInputType(), i.e. scalar, enum or InputObject

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/EnumType.kt
martinbonnin: Makes sense 👍

martinbonnin:
Thanks a lot for looking into this! I took the liberty of pushing a small integration test to your branch, I hope you don't mind. The test checks that the cache key is the one we expect. I didn't think of any other test to do since we can't really restart the JVM in the middle of the tests to make the cache key change...

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/SealedEnumType.kt
lwasyl: I changed the abstract class to `interface EnumValue`, following the GraphQL spec definitely makes sense! Also with that change I now could implement that interface for all enum classes, not only the ones generated as sealed classes.

`writeToJson(value: InputValue` also makes perfect sense. I originally thought it's more of a general-purpose utility, that's why I had doubts :)

lwasyl:
> I took the liberty of pushing a small integration test to your branch, I hope you don't mind. The test checks that the cache key is the one we expect. 

Not at all, now at least I know for sure the fix works ;)

> I didn't think of any other test to do since we can't really restart the JVM in the middle of the tests to make the cache key change...

Technically another `object` instance could be created using reflection, or by serializing and deserializing an object instance. I don't think that's necessary though, I don't see a scenario where such test would fail but yours would pass

apollo-compiler/src/test/graphql/com/example/java_beans_semantic_naming/type/Episode.kt
```diff
@@ -5,15 +5,16 @@
...
 import kotlin.String
 
 /**
  * The episodes in the Star Wars trilogy
  */
 enum class Episode(
-  val rawValue: String
-) {
+  override val rawValue: String
+) : EnumValue {
```
martinbonnin, line 17: Paranoid checks: I think this is a backward compatible change but since it touches regular enums and not only sealed classes this has the potentially to impact a lot more users. Are we missing possibly missing a weird edge case?

lwasyl: I don't see any 🤔 One edge case would be using jar with generated old classes (no interface) with new runtime, but it should work (just the check won't trigger). For enum classes behavior doesn't change at all, as from what I see both raw value and enum name always match (so cache won't break when updating app with new generated classes). 

I can of course revert this if you find some problem or just feel I should only fix the bug with sealed classes and let enum classes as is. It just felt weird leaving actual enum classes without having them implement `EnumValue` interface 🙂

martinbonnin: Yep, I think it's safe too 👍 and way more uniform like this 👍 .

martinbonnin (approved):

lwasyl:
@martinbonnin the CI seems to be acting up so I took this opportunity and fixed commit message (unless CI isn't supposed to pass for PRs from forks). So just fyi I didn't push any more changes, only changed the commit message

martinbonnin:
Yea, the CI randomly hangs for 6 hours 😒 . Sorry about that. It looks like it's happening in integration test but we haven't been able to find what's triggering the hang. Of course it's highly non reproducible...

I'm starting to log failures in this PR in the hopes of finding a common pattern: https://github.com/apollographql/apollo-android/pull/2764.

PR #2786, opened 2020-12-02
Address fragment accessors generation

sav007:
- Generate fragment accessor functions for named fragment interfaces
- Change the accessor naming: for inline fragment use `as*` prefix for named fragment accessor use just name of the fragment.
- Generate accessor functions as part of companion object


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/codegen/FragmentType.kt
```diff
@@ -71,12 +83,5 @@ internal fun CodeGenerationAst.FragmentType.implementationTypeSpec(generateAsInt
       .toBuilder()
       .addSuperinterface(GraphqlFragment::class)
       .applyIf(generateAsInternal) { addModifiers(KModifier.INTERNAL) }
-      .addType(TypeSpec.companionObjectBuilder()
-          .addProperty(PropertySpec.builder("FRAGMENT_DEFINITION", String::class)
```
martinbonnin, line 75: 👍 I guess we don't need that ?

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarship.kt
```diff
@@ -233,8 +229,33 @@ class AllStarship : Query<AllStarship.Data, Operation.Variables> {
...
                       }
+
+                      companion object {
+                        fun Homeworld.planetFragment(): PlanetFragment? = this as? PlanetFragment
+
+                        fun Homeworld.pilotFragmentHomeworldPlanet():
+                            PilotFragment.Homeworld.Planet? = this as?
+                            PilotFragment.Homeworld.Planet
+
+                        fun Homeworld.starshipFragmentPilotConnectionEdgesNodePersonHomeworldPlanet():
```
martinbonnin, line 240: That sounds like a little bit too much 🙃 . Feels like this shouldn't be generated

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/codegen/FragmentType.kt
sav007: Yep, for named fragment default implementation we don't need graphql definition as it already declared in interface

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/AllStarship.kt
sav007: Yeah, I guess we need to skip generating accessors for nested named fragments. Only accessors for top level fragments.

martinbonnin (approved):

PR #2788, opened 2020-12-04
Remove frontend IR

martinbonnin:
Since frontend ir was so close to GQLDocument, it's not super useful to keep it. This PR directly builds the backend IR from the GQLDocument. 

I believe there are a few steps that we could do outside `BackendIRBuilder` though. Stuff like:

* merging inline fragments
* coercining input type values
* computing the document source

I'll adress taht in a follow up PR


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -1,205 +1,298 @@
...
-    private val schema: IntrospectionSchema,
-    private val namedFragments: List<Fragment>,
+    private val schema: Schema,
+    private val fragmentDefinitions: Map<String, GQLFragmentDefinition>,
     private val useSemanticNaming: Boolean,
+    val packageNameProvider: PackageNameProvider?,
+    val typesPackageName: String,
+    val fragmentsPackageName: String
 ) {
+  data class BackendIrBuilderInput(
```
sav007, line 44: Looks like we can't avoid having some abstraction that will be an input for backend IR builder. What do you think of moving this class out of `BackendIrBuilder` and call it `FrontendIR`. Also we can rename `IRBuilder` to `FrontendIRBuilder`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/IRBuilder.kt
```diff
@@ -1,19 +1,19 @@
...
 import com.apollographql.apollo.compiler.frontend.gql.GQLInputObjectTypeDefinition
 import com.apollographql.apollo.compiler.frontend.gql.GQLOperationDefinition
 import com.apollographql.apollo.compiler.frontend.gql.GQLScalarTypeDefinition
 import com.apollographql.apollo.compiler.frontend.gql.Schema
-import com.apollographql.apollo.compiler.frontend.gql.toIR
 import com.apollographql.apollo.compiler.frontend.gql.usedTypeNames
 import com.apollographql.apollo.compiler.frontend.gql.withTypenameWhenNeeded
 
-class IRBuilder(private val schema: Schema,
+internal class IRBuilder(private val schema: Schema,
```
sav007, line 16: If we rename `BackendIrBuilder.BackendIrBuilderInput` to `FrontedIR` we can rename this builder to `FrontendIRBuilder`

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -1,205 +1,298 @@
...
+        description = "",
+        name = "data",
+        arguments = emptyList(),
+        directives = emptyList(),
+        type = GQLNamedType(name = rootTypeDefinition.name)
+    )
+    val dataField = GQLField(
+        name = "data",
+        alias = null,
+        sourceLocation = com.apollographql.apollo.compiler.frontend.gql.SourceLocation.UNKNOWN,
```
sav007, line 147: nit: should it be imported?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -1,205 +1,298 @@
...
+        arguments = null,
+        directives = emptyList(),
+        selectionSet = selectionSet
     ).buildBackendIrField(
         selectionKey = selectionKey,
         generateFragmentImplementations = true,
+        dataFieldDefinition
```
sav007, line 154: nit: `dataFieldDefinition = dataFieldDefinition,`

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -1,205 +1,298 @@
...
+        operationName = name!!,
+        targetPackageName = packageNameProvider?.operationPackageName(filePath = sourceLocation.filePath ?: "") ?: "",
+        operationType = IntrospectionSchema.TypeRef(
+            IntrospectionSchema.Kind.OBJECT,
+            rootTypeDefinition.name
+        ),
+        comment = this.description ?: "",
         variables = variables,
-        definition = this.sourceWithFragments,
+        definition = (toUtf8WithIndents() + "\n" + fragmentNames.map { fragmentDefinitions[it]!!.toUtf8WithIndents() }.joinToString("\n")).trimEnd('\n'),
```
sav007, line 173: nit: I think it can be shortent  
```
toUtf8WithIndents() + fragmentNames.joinToString(
  separator = "\n"
) { fragmentName ->
  fragmentDefinitions[fragmentName]!!.toUtf8WithIndents()
}
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -1,205 +1,298 @@
...
-  private fun Operation.normalizeOperationName(): String {
+  private fun GQLOperationDefinition.normalizeOperationName(): String {
     fun normalizeOperationName(
         useSemanticNaming: Boolean,
         operationNameSuffix: String,
     ): String {
-      return if (useSemanticNaming && !operationName.endsWith(operationNameSuffix)) {
-        operationName.capitalize() + operationNameSuffix
+      require(name != null) {
+        "anonymous operations are not supported. This should have been caught during validation"
```
sav007, line 184: if this should be caught by validation, do we need check it here again? If yes should we update this message to be user friendly?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/FrontendIrMergeUtils.kt
```diff
@@ -33,55 +36,61 @@ internal object FrontendIrMergeUtils {
...
+      when (selection) {
+        is GQLFragmentSpread -> {
+          val index = selectionsToAdd.indexOfFirst { (it as? GQLFragmentSpread)?.name == selection.name }
+          if (index >= 0) {
+            selectionsToAdd.removeAt(index)
+            // named fragments are easy to merge
+            selection
+          } else {
+            selection
+          }
```
sav007, line 67: nit:
```
selectionsToAdd.removeIf {
  (it as? GQLFragmentSpread)?.name == selection.name
}
selection
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: Yep, we won't save the intermediate abstraction there. I don't think "IR" matches well though because it will still contain the `GQLDocuments`. What this abstraction brings is: 

* `fragmentsToGenerate`, `enumToGenerate`, `inputTypeToGenerate`
* packageName stuff
* maybe `sourceWithFragment` 

So it's more about "augmenting" the GQLDocument than transforming it to an intermediate representation. But I don't really have good alternative suggestions either.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: I added the check mainly so that the compiler can infer the non-nullability and not force me to type `name!!` later on but really this should never happen™️. I'm happy to use `name!!` or anything else.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: Right, maybe IR is not good naming here and we should come up with some better naming, as we won't just pass it to BackendIR but also run some optimization like squasing inline fragments, merging fields etc.

sav007 (approved):
👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: See https://github.com/apollographql/apollo-android/pull/2789, I added a `TypesToGenerate` class because this is the main information actually computed by `IRBuilder`. 

squasing inline fragments, merging fields etc, are transformations on the GQLDocument so I don't think we need an intermediate data structure for this. GraphQLCompiler is growing a bit with https://github.com/apollographql/apollo-android/pull/2789 but I don't mind this too much. We need a central place where the compiler applies the different steps.

martinbonnin:
Merged as part of https://github.com/apollographql/apollo-android/pull/2789

PR #2792, opened 2020-12-08
Merge fields in backend ir

martinbonnin:
This fixes the `hero_name` test case.

Also handle `@include`/`@skip` directives with literal arguments.

`@include`/`@skip`handling is still not 100% correct in the case of merged fields, I'll look into this in a follow up PR


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -52,6 +54,42 @@ internal class BackendIrBuilder constructor(
...
+          it.toBoolean() != true
+        })
+      }
+    }
+    if (newFields.isEmpty()) {
+      return null
+    } else {
+      return copy(fields = newFields)
+    }
+  }
```
sav007, line 92: Can we move this logic out of the `BackendIrBuilder.kt` into separate file? We do have `FrontendIrMergeUtils`, do you think it will be better to move this logic over there?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -52,6 +54,42 @@ internal class BackendIrBuilder constructor(
     )
   }
 
+  /**
+   * a GroupedField is a list of fields with the same responseName and arguments but possibly different selectionSets and directives
+   */
+  private data class GroupedField(val fields: List<GQLField>)
```
sav007, line 60: Why not just merger all fields into one field?

martinbonnin: A `GQLField` cannot represent exactly a list of field because:
```
{
  name @include(if: $condition1)
  name @include(if: $condition2)
}
```

is not equivalent to 
```
  name @include(if: $condition1) @include(if: $condition2)
```

We could use another "IR" there that can represent these cases. That felt a bit overkill just for this use case but maybe there can be other uses too.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: This logic is about eliminating literal `@include`/`@skip` directives. I can move them to a separate `IncludeSkipUtils` file but it's not really about merging so I wouldn't use `FrontendIrMergeUtils`

apollo-compiler/src/test/graphql/com/example/merged_include/TestOperation.graphql
```diff
@@ -0,0 +1,8 @@
+query TestQuery {
+    hero {
+        name
+    }
+    hero @include(if: false) {
+        id
```
sav007, line 6: I'm curious what will happen if we add duplicate `name` field? I have assumption that it generator fails in this case

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: Yeah from GraphQL execution perspective, but from codegen we don't care right? we should treat this as one field as we going to generate only one `name` field.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: Yeah I would suggest to move it out from the builder as it keep growing fast, so might be good `IncludeSkipUtils `.

apollo-compiler/src/test/graphql/com/example/merged_include/TestOperation.graphql
martinbonnin: Well, for this specific case, it doesn't change anything to the codegen since the `@include(if: false)` prunes any selection in `hero {}`

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: The thing is... apollo-android is also an execution engine, kind of. When reading from the cache, we're effectively acting like a server. If we don't honor directives, we will over fetch from the cache. Maybe it's not a big deal, maybe it is, I don't know...

martinbonnin:
For the record, I finally ended up re-introducing `FrontendIr` because it can represent arbitrary `@include`/`@skip` conditions. It also makes a clearer separation so I think it's worth having it.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/FrontendIrMergeUtils.kt
```diff
@@ -7,45 +7,6 @@ import com.apollographql.apollo.compiler.frontend.GQLSelectionSet
...
-  fun List<GQLInlineFragment>.mergeInlineFragmentsWithSameTypeConditions(): List<GQLInlineFragment> {
-    return this
-        .groupBy { fragment -> fragment.typeCondition.name }
-        .map { (_, groupedFragments) ->
-          groupedFragments.drop(1).fold(groupedFragments.first()) { result, fragment ->
-            result.merge(fragment)
-          }
-        }
-  }
-
```
sav007, line 48: why this was removed? I actually need it in my upcoming PR.

apollo-compiler/src/test/graphql/com/example/arguments_simple/fragment/HeroDetail.kt
```diff
@@ -68,20 +68,6 @@ interface HeroDetail : GraphqlFragment {
...
-        |  friendsConnection(first: ${'$'}friendsCount) {
-        |    totalCount
-        |    edges {
-        |      node {
-        |        name @include(if: ${'$'}IncludeName)
-        |      }
-        |    }
-        |  }
-        |}
-        """.trimMargin()
```
sav007, line 83: hm, why it's removed?

martinbonnin: Might be an oversight. I think I remembered us saying we don' t need it but I might be confusing with something else. I'll add it back

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/FrontendIrMergeUtils.kt
martinbonnin: That's done in the FrontendIr: https://github.com/apollographql/apollo-android/blob/152b0c7d7c6a38c71ca56e1e6ebe00a052ef7e7a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/frontend/ir/gql_to_frontend_ir.kt#L161

apollo-compiler/src/test/graphql/com/example/arguments_simple/fragment/HeroDetail.kt
martinbonnin: It's back in! I'm still curious, what is the use case for `FRAGMENT_DEFINITION` ?

sav007 (approved):

PR #2795, opened 2020-12-12
Fix watcher clone

martinbonnin:
`refetchResponseFetcher` wasn't taken into account in `RealApolloQueryWatcher.clone`. Many thanks to @lwasyl for investigating this and helping find the issue.


lwasyl:
I noticed this change has some far reaching consequences. Consider two queries:

```graphql
Foo {
  heroes { // List of items
    id
    name
  }
}

Bar {
  heroes {
    id // Same items as `Foo` but different/incomplete data
  }
}
```

Now let's consider the following scenario:
 1) Observe `Foo` query with `CACHE_ONLY` strategy
 2) Fetch `Foo` with `NETWORK_ONLY` strategy and close the watcher after receiving first element
 3) Let's say `heroes` data on the server changes at this point and adds one element
 4) Observe `Bar` query with `NETWORK_ONLY` strategy

Current behavior:
 1) Nothing happens yet
 2) Both watchers will emit e.g. 4 items from the server. Second watcher is closed (first one stays active)
 3) Server data changes and now has 5 items on the server
 4) `Bar` watcher will emit data with 5 items. But `Foo` watcher with `CACHE_ONLY` strategy will see that the data has changed. It'll try to get data from the cache, but it will miss because `Bar` doesn't fetch `name` property. So the watcher will make an API call to fetch all five items from the server, this time with `name` as well. It will emit this new data

 Behavior after this patch is merged:
  Points 1, 2, 3 don't change

  Point 4) changes: the watcher will now try to update the data from the cache, but it will fail. However, no additional network call will be made. So now a query for `Foo` and `Bar` are not consistent wrt to the data they emit. This can happen for any queries that fetch the same items but with different data inside.

  I don't know what would be the expected behavior. The new one seems bad because the cache miss will be completely transparent to the users. The old one is more correct but causes additional network calls without clear information the library consumers. The old (current) behavior has also another flaw: if the refetching fails, then error is emitted from the watcher, in turn closing it. So a refetch failure in a watcher with `CACHE_ONLY` strategy will close the watcher such that subsequent cache updates won't trigger any emission from that watcher. The documentation suggests that `CACHE_ONLY` queries never fail so it seems like an issue (especially if someone wants to use `CACHE_ONLY` for observing data and only update the cache using one-off `NETWORK_ONLY` calls or `prefetch`ing).

martinbonnin:
Hi! Thanks again for diving into this! 

>  I observe that the CACHE_ONLY query actually emits null, because it tries to refetch the data, encounters a cache miss, and so emits an empty response. 

Indeed the semantics are surprising there. There will also be a response with `data = null` for the initial query, when `dependentKeys` is not known yet. Maybe these should be filtered out by default so that the watcher only returns successful responses or errors (but not empty responses)?

> if the refetching fails, then error is emitted from the watcher, in turn closing it

Maybe we should stop using exceptions for errors and use a sealed `Result` class instead? That'll allow to keep the flow open in this case (but will require most callers to write additional logic at call site)

lwasyl:
> Maybe these should be filtered out by default so that the watcher only returns successful responses or errors (but not empty responses)?

Then you'd possibly observe inconsistent data between the queries, which is not great. Personally I'd say adding `filterNotNull` is better if it'd mean keeping the ability to decide what to do with stale date

> Maybe we should stop using exceptions for errors and use a sealed Result class instead? That'll allow to keep the flow open in this case (but will require most callers to write additional logic at call site)

That would make sense, as I understand the only alternative right now is carefully written retry logic with specific fetchers. But even then a failed refetch would result in inconsistent data 😉 One other (I don't know how feasible) thought we had was to make a query or a prefetch that fetches from the network fail if any of the cache_only queries result in a cache miss. That's probably a whole lot different conversation though

martinbonnin:
> Then you'd possibly observe inconsistent data between the queries, which is not great.

Not sure I get why this would be inconsistent. We could by default filter out the cache misses and make the ability to observe them opt-in, maybe with a per-query parameter:

```
fun ApolloClient.query(query: Query, reportCacheMisses: Boolean = false)
``` 

This way, cache misses are ignored and all queries ignore them by default, which is somewhat consistent. I _think_ most users don't care about cache misses. Usually they just want data.

> One other (I don't know how feasible) thought we had was to make a query or a prefetch that fetches from the network fail if any of the cache_only queries result in a cache miss. That's probably a whole lot different conversation though

Right now there's a unidirectional dataflow where the cache is updated and triggers watchers. Having the watchers circle back to the original query will certainly introduce a new class of issues.

lwasyl:
>Not sure I get why this would be inconsistent.

After this patch, if two queries fetch some disjoint fields (`id`, `foo`, `bar` in the first one and `id`, `bar`, `baz` in the second one) then in the case I mentioned, with the list, we might end up with this flow:
 - query 1 fetches list of items, `[id=1, foo=foo, bar=bar]`. it's cache_only
 - query 2 fetches the same list but with different fields. It gets different data: `[id=1, bar=rab, baz=zab], [id=2, bar=...]`
 - query 1 should refetch (because it observes the same list as query 2 and the list has changed). But there's a cache miss as `foo` field is not fetched for `id=2`. Since the query is cache_only, it'll report a cache miss instead of reaching out to the network
 - if we filter out this cache miss response, we end up with latest value of query 1 which is `id=1, foo=foo, bar=bar`. But query 2 has the same object with id1 and different (more recent) `bar` field.

That's what I understand would happen, please clarify if I'm mistaken somewhere. And that's what I mean by inconsistency: a possibility that the same object is emitted by two queries, but some fields differ between those two emissions

martinbonnin:
> if we filter out this cache miss response, we end up with latest value of query 1 which is id=1, foo=foo, bar=bar. But query 2 has the same object with id1 and different (more recent) bar field.

Got it. Yep, that will happen.

> And that's what I mean by inconsistency: a possibility that the same object is emitted by two queries, but some fields differ between those two emissions

I see. I'm not sure there's a "good" solution there though. Emitting the cache misses signals that some data is needed but ultimately, a network request is required to get it.

lwasyl:
> I'm not sure there's a "good" solution there though. Emitting the cache misses signals that some data is needed but ultimately, a network request is required to get it.

I agree, and that's probably out of scope of this change anyway. If anything, I'd vote for keeping the current behavior (returning nulls) default - most users will likely use `cache_first` strategy and so not hit this, and those that do would explicitly decide what to do about lack of data

PR #2799, opened 2020-12-14
Fix the example arguments for downloadApolloSchema

ansman:
Using the `-P` parameters have been deprecated


martinbonnin (approved):
Thanks !

PR #2800, opened 2020-12-14
Replace all .toDeferred().await() with .await() in the tutorial

ansman:
`toDeferred` has been deprecated and should be replaces by just `await`.


martinbonnin (approved):
Thanks again!

PR #2801, opened 2020-12-14
Update 11-subscriptions.mdx

codeteo:
Changes Wikipedia's link from `fr` to `en`. (Since this is an English tutorial)


martinbonnin (approved):
Good catch! Merci beaucoup !

PR #2803, opened 2020-12-14
[Plugin] For recent versions of Gradle, use .convention to provide the default

### LINKED ISSUE #2802: Gradle Plugin: Calling apollo.customTypeMapping.put does nothing
**Summary**
Calling `customTypeMapping.put` without first calling `set` is ignored.

**Version**
2.4.6 (Gradle 6.7.1)

**Description**
To my surprise Gradle seems to ignore calls such as `MapProperty.put` if the property is not present. Here's some concrete examples:
```kotlin
apollo {
  // Prints `null`
  println(customTypeMapping.orNull)

  customTypeMapping.put("foo", "bar")
  // Prints `null`
  println(customTypeMapping.orNull)

  customTypeMapping.set(emptyMap())
  customTypeMapping.put("foo", "bar")
  // Prints `{foo=bar}`
  println(customTypeMapping.orNull)
}
```

The fix for this is simple, the Apollo Gradle Plugin just needs to set `emptyMap()` as the default value. Using `convention` does not work unfortunately.

The reason why I prefer `put` is because I often set up my Gradle scripts using custom Gradle plugins in `buildSrc` which can add some common default values.

martinbonnin:
This allows to call `customTypeMap.put()`

fixes #2802

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilerParams.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilerParams.kt
@@ -8,6 +8,7 @@ import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.SetProperty
+import org.gradle.util.GradleVersion
 import javax.inject.Inject
 
 abstract class DefaultCompilerParams @Inject constructor(objects: ObjectFactory) : CompilerParams {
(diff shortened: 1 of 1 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultCompilerParams.kt
```diff
@@ -42,11 +43,16 @@ abstract class DefaultCompilerParams @Inject constructor(objects: ObjectFactory)
   abstract override val alwaysGenerateTypesMatching: SetProperty<String>
 
   init {
-    // see https://github.com/gradle/gradle/issues/7485
-    // TODO replace with `convention(null)` when we can target Gradle 6.2
-    customTypeMapping.set(null as Map<String, String>?)
-    sealedClassesForEnumsMatching.set(null as List<String>?)
-    alwaysGenerateTypesMatching.set(null as Set<String>?)
+    if (GradleVersion.current() >= GradleVersion.version("6.2")) {
```
ansman, line 46: Seeing as Gradle is not up to 6.7 with 7 around the corner it doesn't feel unreasonable to require 6.2+

martinbonnin: Yup. At the same time, [there are still users on Gradle 5.6](https://github.com/apollographql/apollo-android/issues/2536) so if we can make this work for them, I think it's not too much of a price to pay for now.

PR #2806, opened 2020-12-16
Rework code gen (again) to better support nested fragments

sav007:


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
```diff
@@ -99,46 +97,38 @@ internal data class BackendIr(
...
-        override val selectionKeys: Set<SelectionKey>,
-        override val description: String?,
-    ) : Fragment()
+  data class Fragments(
+      val fragments: List<Fragment>,
+      val accessors: Map<String, SelectionKey>,
+  ) : List<Fragment> by fragments
+
+  data class Fragment(
+      val name: String,
```
martinbonnin, line 109: What about?

```suggestion
      val typeCondition: String,
```

As `name` could be mistaken with the fragment spread name

martinbonnin: Got it. It's not directly `typeCondition` but `typeCondition.capitalize` so it's the name of the generated class. Oh well, naming is hard... Maybe we could postpone the capitalization to the codegen phase and use only `typeCondition` in BackendIr?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/SelectionKeyUtils.kt
```diff
@@ -21,10 +21,13 @@ internal object SelectionKeyUtils {
     }
```
martinbonnin, line 21: There are two `addFieldSelectionKeys` methods here that seem unused

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrMergeUtils.kt
```diff
@@ -19,15 +19,18 @@ internal object BackendIrMergeUtils {
...
+            ),
+            accessors = this.fragments.accessors + otherField.fragments.accessors,
         ),
         selectionKeys = this.selectionKeys + otherField.selectionKeys,
     )
   }
 
-  private fun List<BackendIr.Fragment>.mergeFragments(
+  fun List<BackendIr.Fragment>.mergeFragments(
       parentSelectionSet: List<BackendIr.Field>,
```
martinbonnin, line 34: Nit: 

```suggestion
      fields: List<BackendIr.Field>,
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -452,144 +410,161 @@ internal class BackendIrBuilder constructor(
...
+              fragmentName = fragmentName,
+              fragmentTypeCondition = inlineFragment.fragmentDefinition.typeCondition.name,
+              fragmentSelectionSet = inlineFragment.fragmentDefinition.selections,
+              fragmentDescription = inlineFragment.fragmentDefinition.typeCondition.description ?: "",
+              fragmentCondition = inlineFragment.condition.toBackendIrCondition(),
+              namedFragmentSelectionKey = parentNamedFragmentSelectionKey?.plus(fragmentName),
+              parentSelectionKey = parentSelectionKey,
+              parentSelectionSet = parentSelectionSet.addFieldSelectionKey(
+                  parentNamedFragmentSelectionKey?.plus(fragmentName)
+              ),
```
martinbonnin, line 529: I hate to be the bearer of bad news but we'll need to recurse nested inline fragments as well. One exemple is there: https://github.com/apollographql/apollo-android/commit/2c2b39ad670f31a51e072701c51b035d28f64c9d

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
sav007: We could but there some logic around selection key comparison + formatting names for accessors, default implementation etc. Even if remove it there will be only 2 places at most.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: should be fixed with this commit https://github.com/apollographql/apollo-android/pull/2806/commits/a68fb53966b6626e653d061a3ae3eca7dd524d62#diff-7cee39198b0604492035a9fe87f2c7c4395fe30c6bad0c7a312849e4f73a2da4R171

PR #2807, opened 2020-12-16
Initial work for multiplatform cache

martinbonnin:
* Added `apollo-cache-interceptor` as a separate artifact
* The cache is cache_first only at the moment and provided by
  `ApolloCacheInterceptor`
* Added CacheExecutionContext and `Response.cacheContext` to access
  `fromCache`
* Added `apollo-integration-kotlin`, like `apollo-integration` but
  multiplatform. See CacheInterceptorTest for a test
* Added `apollo-testing-support` to group useful classes for testing.
  Ultimately, this could be exported to end users for mock/testing
purposes

Next steps:
* Implement different fetchers
* Make the cache work with the streaming parser


apollo-cache-interceptor/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_ARTIFACT_ID=apollo-cache-interceptor
```
sav007, line 1: why not just call it `apollo-cache`? I thought that everything that belongs to normalize cache will be packaged in one artifact?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.apollographql.apollo.internal.response.RealResponseReader
+import com.apollographql.apollo.internal.response.RealResponseWriter
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flow
+
+@ApolloExperimental
+class ApolloCacheInterceptor : ApolloRequestInterceptor {
+  val normalizedCache = SimpleNormalizedCache()
+  val readableStore = object : ReadableStore {
```
sav007, line 32: nit: private for both props?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.apollographql.apollo.interceptor.ApolloRequestInterceptor
+import com.apollographql.apollo.interceptor.ApolloResponse
+import com.apollographql.apollo.internal.response.RealResponseReader
+import com.apollographql.apollo.internal.response.RealResponseWriter
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flow
+
+@ApolloExperimental
+class ApolloCacheInterceptor : ApolloRequestInterceptor {
```
sav007, line 30: should we mark as internal? I was thinking back in a day that cache plugin will take existing `ApolloClient` and decorate it with proper execution context and setup interceptors, so user won't know existence of interceptor as it's our internal implementation abstraction.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -0,0 +1,95 @@
...
+  val normalizedCache = SimpleNormalizedCache()
```
sav007, line 31: Should we extract interface and inject implementation during construction time?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -0,0 +1,95 @@
...
+    override fun read(keys: Collection<String>, cacheHeaders: CacheHeaders): Collection<Record> {
+      return keys.mapNotNull { normalizedCache.loadRecord(it, cacheHeaders) }
+    }
+  }
+
+  override fun <D : Operation.Data> intercept(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
+    return flow {
+      val response = readFromCache(request)
+      if (response != null) {
+        emit(ApolloResponse(requestUuid = request.requestUuid, response = response, executionContext = request.executionContext + CacheExecutionContext(true)))
```
sav007, line 46: nit: could you pls put arguments on separate lines

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/CacheExecutionContext.kt
```diff
@@ -0,0 +1,17 @@
+package com.apollographql.apollo.interceptor.cache
+
+import com.apollographql.apollo.api.ApolloExperimental
+import com.apollographql.apollo.api.ExecutionContext
+import com.apollographql.apollo.api.Operation
+import com.apollographql.apollo.api.Response
+
+@ApolloExperimental
+data class CacheExecutionContext(
```
sav007, line 9: internal?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
sav007, line 32: As far as I remember the original idea is that interceptor get `*Store` implementation injected where `*Store` is constructed with cache implementation. So interceptor knows only about store, and has only one dependency.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/simple/SimpleNormalizedCache.kt
```diff
@@ -0,0 +1,45 @@
+package com.apollographql.apollo.cache.normalized.simple
+
+import com.apollographql.apollo.cache.CacheHeaders
+import com.apollographql.apollo.cache.normalized.CacheKey
+import com.apollographql.apollo.cache.normalized.NormalizedCache
+import com.apollographql.apollo.cache.normalized.Record
+
+class SimpleNormalizedCache : NormalizedCache() {
+  val map = mutableMapOf<String, Record>()
```
sav007, line 9: private?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/simple/SimpleNormalizedCache.kt
```diff
@@ -0,0 +1,45 @@
+package com.apollographql.apollo.cache.normalized.simple
+
+import com.apollographql.apollo.cache.CacheHeaders
+import com.apollographql.apollo.cache.normalized.CacheKey
+import com.apollographql.apollo.cache.normalized.NormalizedCache
+import com.apollographql.apollo.cache.normalized.Record
+
+class SimpleNormalizedCache : NormalizedCache() {
```
sav007, line 8: internal? Also should we call it `InMemoryNormalizedCache` or  `MapNormalizedCache`?

apollo-cache-interceptor/gradle.properties
martinbonnin: Excellent question! I'm not 100% sure where I'm going with this but it feels like we should bundle the sqlite cache by default so multiple artifacts would be needed. At this stage of development, my rationale was that having multiple single-responsibility modules makes for better separation of concern and that we would "polish" the user-facing API/artifacts once everything is feature complete.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Also excellent question. I'd love to find a nicer API. Did you have something in mind already? 

Should we introduce a `ApolloClientBuilder` to take care of the wiring of all the interceptors? In all cases, I think this should stay public as ultimately, we also want users to be able to plug their own interceptors and modify the list as they please?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Sure thing!

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/CacheExecutionContext.kt
martinbonnin: I used this to carry the information about whether a given response came from the cache or not so ultimately the user should have access to this. 

See `Response<D>.cacheContext()`

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Makes perfect sense, I changed that.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/simple/SimpleNormalizedCache.kt
martinbonnin: Renamed to `MapNormalizedCache`. `InMemoryNormalizedCache` might clash with the future LRU cache.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/CacheExecutionContext.kt
sav007: My bad you are right.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: @sav007 I ended up adding `ApolloClient.Builder`, one option would have been to make `ApolloClient` a data class and copying it around. Using a Builder adds some boilerplate code but also has a few advantages:

* I didn't really liked the idea of making `ApolloClient` a data class
* It makes it clear what is mutable and what's not
* It puts the arguments verification logic in a single place

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
```diff
@@ -50,4 +55,100 @@ class ApolloClient(
...
+
+  abstract class Builder {
+    internal abstract val scalarTypeAdapters: ScalarTypeAdapters
+
+    private var networkTransport: NetworkTransport? = null
+    private var subscriptionNetworkTransport: NetworkTransport? = null
+    private var interceptors: List<ApolloRequestInterceptor> = emptyList()
+    private var executionContext: ExecutionContext = ExecutionContext.Empty
+
+    fun serverUrl(serverUrl: String): Builder {
```
sav007, line 77: nit:
```
fun serverUrl(serverUrl: String): Builder = apply {
   ...
}
```
the same nit for the rest APIs

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
```diff
@@ -50,4 +55,100 @@ class ApolloClient(
...
+
+  fun newBuilder(): Builder {
+    return DefaultBuilder(scalarTypeAdapters)
+        .networkTransport(networkTransport)
+        .subscriptionNetworkTransport(subscriptionNetworkTransport)
+        .interceptors(interceptors)
+        .executionContext(executionContext)
+  }
+
+  class DefaultBuilder(override val scalarTypeAdapters: ScalarTypeAdapters = ScalarTypeAdapters.DEFAULT) : Builder()
```
sav007, line 67: is there a reason to have this `DefaultBuilder ` class, can't `Builder` be concrete class with `scalarTypeAdapters: ScalarTypeAdapters = ScalarTypeAdapters.DEFAULT`

apollo-cache-interceptor/gradle.properties
sav007: Sure, we can always change this later, let's leave it for now.

sav007 (approved):

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
martinbonnin: I was contemplating the idea of auto generating Builders with pre-built ScalarTypeAdapters like `ApolloClient.Companion.GithubBuilder()`. But that makes the generated code depend on the runtime which isn't great. I'll revert.

Issue #1864, opened 2019-12-18
Apollo subscription only onConnected() is called


developer--:
I have a subscription to my server and it is connected successfully `onConnected()` method is called,
but when I upload something to the server no subscription methods is invoked.
this is my apollo client configuration
```
private const val webSocketUrl = "ws://12.34.56.7:3000/graphql"
private val okHttp = OkHttpClient
      .Builder()
      .addInterceptor(interceptor)
      .retryOnConnectionFailure(true)
      .pingInterval(30, TimeUnit.SECONDS)
      .build()

private val transportFactory = WebSocketSubscriptionTransport.Factory(webSocketUrl, okHttp)
val apolloSocketClient: ApolloClient = ApolloClient.builder()
       .okHttpClient(okHttp)
       .serverUrl(BuildConfig.BASE_URL)
       .subscriptionTransportFactory(transportFactory)
       .build()`
```

And this is invocation

```
val mySubscriptionCall = MySubscription("MY_TOPIC")
subscription = apolloSocketClient.subscribe(mySubscriptionCall)

   subscription.execute(object : ApolloSubscriptionCall.Callback<MySubscription.Data>{
        override fun onFailure(e: ApolloException) {
          Log.e("SUBSCRIPTION_","onFailure",e)
        }

        override fun onResponse(response: Response<MySubscription.Data>) {
          Log.e("SUBSCRIPTION_","onResponse ${response.data()}")
        }

        override fun onConnected() {
          Log.e("SUBSCRIPTION_","onConnected")
        }

        override fun onTerminated() {
          Log.e("SUBSCRIPTION_","onTerminated")
        }

        override fun onCompleted() {
          Log.e("SUBSCRIPTION_","onCompleted")
        }

      })`


I'm only getting onConnected
```
D/OkHttp: <-- 101 Switching Protocols http://12.34.56.7:3000/graphql (180ms)
D/OkHttp: Upgrade: websocket
D/OkHttp: Connection: Upgrade
D/OkHttp: Sec-WebSocket-Accept: QG+xxYzpdn9O2QUYMH0ZDKaUH1U=
D/OkHttp: Sec-WebSocket-Protocol: graphql-ws
D/OkHttp: <-- END HTTP
E/SUBSCRIPTION_: onConnected
```
**Version**
apollo lib version 1.2.2

xorander00:
I don't have my project/code available to me at the moment, but I ran into the same issue a few weeks ago.  
  
[...]

PR #2809, opened 2020-12-17
Implement support for AWS AppSync in WebSocketSubscriptionTransport

### LINKED ISSUE #2808: How to include "extensions" for AppSync subscriptions
I'm hoping I can use Apollo for an integration with AWS AppSync. Mutations and queries has worked flawlessly but there's been some issues with the subscriptions.

Connections work but when Apollo sends the start command AWS responds with:
```json
{
  "type": "error",
  "payload": {
    "errors": [
      {
        "errorType": "UnsupportedOperation",
        "message": "unknown not supported through the realtime channel"
      }
    ]
  }
}
```
The start message looks like this with Apollo:
```json
{
  "id": "9f6b7d81-af63-4164-84dc-1121b12bc81e",
  "type": "start",
  "payload": {
    "variables": {
      "id": "2728459231"
    },
    "operationName": "SomeSubscription",
    "query": "subscription SomeSubscription($id: String!) { someSubscription(id: $id) { __typename stuff }"
  }
}
```

Looking through the AWS documentation it seems like the message should actually look like this:
```json
{
  "id": "9f6b7d81-af63-4164-84dc-1121b12bc81e",
  "type": "start",
  "payload": {
    "data": "<json encoded payload>",
    "extensions": {
      "authorization": {
        "host": "example.com",
        "x-api-key": "deadd00d"
      }
    }
  }
```

Looking through the source for Apollo this does not appear to be supported out of the box. Is there any way to work around this?

ansman:
This PR adds support for the AWS AppSync format of subscriptions. The only difference between AppSync and Apollo is that the payload for the `start` messages need to look like this:
```json
{
  "id": "...",
  "type": "start",
  "payload": {
    "data": "{\"variables\":{...},\"operationName\":\"...\",\"query\":\"...\"}",
    "extensions": {
      "authorization": ...
    }
  }
}
```

This PR will allow passing a flag for sending the payload as a JSON string as well as allow passing arbitrary extensions.

This fixes #2808 and #1864


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/JsonWriters.kt
```diff
@@ -0,0 +1,13 @@
+package com.apollographql.apollo.api.internal.json
+
+inline fun JsonWriter.writeObject(crossinline block: JsonWriter.() -> Unit) {
```
ansman, line 3: Ideally I would want these to be `internal` but since they live in a separate module the cannot. 

They are very handy in making the code more readable and makes manual JSON writing much less error prone.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,166 @@
...
+    webSocket.set(null)
+  }
+
+  private fun OperationClientMessage.serializeToJson(): String =
+      toJsonString(
+        writePayloadAsJsonString = writePayloadAsJsonString,
+        extensions = extensions
+    )
+
+  internal class WebSocketListener(delegate: WebSocketSubscriptionTransport) : okhttp3.WebSocketListener() {
```
ansman, line 84: I kept this as `internal` since I'm assuming it's there to avoid the synthetic accessor methods from being generated?

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,166 @@
...
+   * @param webSocketUrl The URL of the GraphQL WebSocket API
+   * @param webSocketConnectionFactory The [WebSocket.Factory] to use to create the websockets
+   * @param writePayloadAsJsonString If the payload should be serialized as a JSON String (default false). Set this to `true` for AWS
+   *                                 AppSync integrations.
+   * @param extensions Any additional extensions to when sending start commands. Must not be changed after this creating the factory.
+   */
+  class Factory @JvmOverloads constructor(
+      webSocketUrl: String,
+      private val webSocketConnectionFactory: WebSocket.Factory,
+      private val writePayloadAsJsonString: Boolean = false,
```
ansman, line 122: I'm not very happy with the format of this change and I'm happy talk alternatives. 
I like being able to pass arbitrary extensions but not how this is triggered.

One alternative would be to introduce some sort of "format" or "dialect" enum which there are 2 off. I tried to find some official specification for the websocket format but there doesn't seem to be one?

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,166 @@
...
+        .url(webSocketUrl)
+        .addHeader("Sec-WebSocket-Protocol", "graphql-ws")
+        .addHeader("Cookie", "")
+        .build()
+
+    override fun create(callback: SubscriptionTransport.Callback): SubscriptionTransport =
+        WebSocketSubscriptionTransport(webSocketRequest, webSocketConnectionFactory, callback, writePayloadAsJsonString, extensions)
+  }
+
+  class AppSyncFactory @JvmOverloads constructor(
```
ansman, line 135: This was created as a convenience to do two things for the user:
* Add the `header` and `payload` query parameters which AppSync requires
* Add the authorization as an extension to the start messages.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,166 @@
...
+      webSocketUrl: String,
+      private val webSocketConnectionFactory: WebSocket.Factory,
+      authorization: Map<String, Any?>,
+      payload: Map<String, Any?> = emptyMap()
+  ) : SubscriptionTransport.Factory by Factory(
+      webSocketUrl = webSocketUrl
+          .let {
+            HttpUrl.get(when {
```
ansman, line 143: This duplicates what `Request.url` does when passing a string. We use it so that we can leverage OkHttp's URL handling here.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,166 @@
...
+              it.startsWith("ws://", ignoreCase = true) -> "http" + it.drop(2)
+              it.startsWith("wss://", ignoreCase = true) -> "https" + it.drop(3)
+              else -> it
+            })
+          }
+          .newBuilder()
+          .setQueryParameter("header", authorization.encodeAsQueryParam())
```
ansman, line 150: I considered checking if the query param is already present but if the user is passing the header manually they shouldn't use the `AppSyncFactory`.

ansman:
I tested this in our app where we have an AppSync backend and it works fine. One concern I have is if the extensions (authorization really) should be configurable per subscription. It's unclear from the AppSync docs but it's not unthinkable that you'd want one type of authorization to open the web socket and one to start a subscription.

I think this is good for now though and it's easier to add this functionality later to the subscription object than to remove it once implemented.

ansman:
Oh, and obviously tests are still TODO

apollo-runtime/api.txt
```diff
@@ -373,38 +373,65 @@ package com.apollographql.apollo.response {
...
+    field public static final com.apollographql.apollo.subscription.OperationClientMessage.Companion Companion;
+  }
+
+  public static final class OperationClientMessage.Companion {
   }
 
   public static final class OperationClientMessage.Init extends com.apollographql.apollo.subscription.OperationClientMessage {
-    ctor public OperationClientMessage.Init(Map<String,Object>);
-    method public void writeToJson(com.apollographql.apollo.api.internal.json.JsonWriter);
+    ctor public OperationClientMessage.Init(java.util.Map<java.lang.String,?> connectionParams);
```
ansman, line 397: This is still binary compatible

apollo-runtime/api.txt
```diff
@@ -373,38 +373,65 @@ package com.apollographql.apollo.response {
...
+  }
+
   public interface OnSubscriptionManagerStateChangeListener {
     method public void onStateChange(com.apollographql.apollo.subscription.SubscriptionManagerState!, com.apollographql.apollo.subscription.SubscriptionManagerState!);
   }
 
-  public abstract class OperationClientMessage {
-    method public String toJsonString();
-    method public abstract void writeToJson(com.apollographql.apollo.api.internal.json.JsonWriter);
+  public abstract sealed class OperationClientMessage {
```
ansman, line 383: All of the changes here are fine since it was not possible to subclass this class anyway.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/JsonWriters.kt
martinbonnin: Since the package is marked as `internal`, it should be ok

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
martinbonnin: Can you add a comment about this?

ansman: Yeah, will do

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
```diff
@@ -0,0 +1,165 @@
...
+          .toString(),
+      webSocketConnectionFactory = webSocketConnectionFactory,
+      writePayloadAsJsonString = true,
+      extensions = mapOf("authorization" to authorization)
+  ) {
+    companion object {
+      private fun Map<String, Any?>.encodeAsQueryParam(): String {
+        val buffer = Buffer()
+        JsonWriter.of(buffer).jsonValue(this)
+        return Base64.getUrlEncoder().encodeToString(buffer.readByteArray())
```
martinbonnin, line 161: Maybe call this method `toBase64()` ? It's used as a query param but the method itself uses Base64 and not urlencoding.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
martinbonnin: I think this used to be package private in Java. This could certainly be `private` except we're using it from tests so `internal` seems good

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransport.kt
ansman: Done.

ansman:
I rewrote the implementation to use a "serializer" instead which turned out much nicer I think.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.apollographql.apollo.api.internal.json.JsonWriter
+import com.apollographql.apollo.api.internal.json.Utils
+import com.apollographql.apollo.api.internal.json.writeObject
+import com.apollographql.apollo.subscription.ApolloOperationMessageSerializer.writePayloadContentsTo
+import okhttp3.HttpUrl
+import okio.Buffer
+import okio.BufferedSink
+import java.util.Base64
+
+class AppSyncOperationMessageSerializer(
```
martinbonnin, line 12: Any reason this is a class and not an object like `ApolloOperationMessageSerializer` ?

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
```diff
@@ -0,0 +1,81 @@
...
+    private val authorization: Map<String, Any?>
+) : OperationMessageSerializer by ApolloOperationMessageSerializer {
```
martinbonnin, line 14: Is the implementation by delegation needed here ? I think the same result could be done by actually implementing the `readServerMessage` method

```
  override fun readServerMessage(source: BufferedSource) = ApolloOperationMessageSerializer.readServerMessage(source)
 
```

It's a few more bytes but avoid using the "implement by delegate" pattern which isn't used a lot in the codebase and it might produce a bit less bytecode too.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
```diff
@@ -0,0 +1,81 @@
...
+          .newBuilder()
+          .setQueryParameter("header", authorization.base64Encode())
+          .setQueryParameter("payload", payload.base64Encode())
+          .build()
+          .toString()
+          .let { url ->
+            when {
+              url.startsWith("http://", ignoreCase = true) -> "ws" + url.drop(4)
+              url.startsWith("https://", ignoreCase = true) -> "wss" + url.drop(5)
+              else -> url
```
martinbonnin, line 75: Ouch. I guess we want to keep the `ws://` to avoid confusing any user there, even if ultimately, okhttp is going to replace everything with `http://` ? That feels a bit awkward but I don't really have another suggestion besides using `java.net.URL` that will end up equally verbose.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationMessageSerializer.kt
```diff
@@ -0,0 +1,13 @@
+package com.apollographql.apollo.subscription
+
+import okio.BufferedSink
+import okio.BufferedSource
+import java.io.IOException
+
+interface OperationMessageSerializer {
+  @Throws(IOException::class)
+  fun OperationClientMessage.writeTo(sink: BufferedSink)
```
martinbonnin, line 9: What about `writeClientMessage(sink: BufferedSink, message: OperationClientMessage)` so that it's more symmetrical with `readServerMessage` ? I don't have a strong opinion there so either work for me though.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
ansman: Yeah, that's what I thought too. `URL` doesn't help here because it doesn't have any builder pattern.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationServerMessage.kt
```diff
@@ -0,0 +1,54 @@
...
+import okio.Buffer
+
+sealed class OperationServerMessage {
+  companion object {
+    const val JSON_KEY_ID = "id"
+    const val JSON_KEY_TYPE = "type"
+    const val JSON_KEY_PAYLOAD = "payload"
+
+    @Suppress("DeprecatedCallableAddReplaceWith")
+    @Deprecated("This method should be used any more. Instead you should use a serializer")
```
martinbonnin, line 12:
```suggestion
    @Deprecated("This method is deprecated. Use an OperationMessageSerializer instead.")
```

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationMessageSerializer.kt
ansman: Me neither. I figured the receiver pattern worked well because it would be like a function on `OperationClientMessage` so the implementation looks cleaner.

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.kt
```diff
@@ -0,0 +1,61 @@
...
+        val buffer = Buffer()
+        with(ApolloOperationMessageSerializer) { writeTo(buffer) }
+        buffer.readUtf8()
+      } catch (e: IOException) {
+        throw RuntimeException("Failed to serialize to json", e)
+      }
+
+  @Suppress("DeprecatedCallableAddReplaceWith")
+  @Throws(IOException::class)
+  @Deprecated("Use a serializer instead")
```
martinbonnin, line 22:
```suggestion
    @Deprecated("This method is deprecated. Use an OperationMessageSerializer instead.")
```

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/OperationClientMessage.kt
```diff
@@ -0,0 +1,61 @@
+package com.apollographql.apollo.subscription
+
+import com.apollographql.apollo.api.ScalarTypeAdapters
+import com.apollographql.apollo.api.Subscription
+import com.apollographql.apollo.api.internal.json.JsonWriter
+import okio.Buffer
+import java.io.IOException
+
+sealed class OperationClientMessage {
+  @Deprecated("Use a serializer instead")
```
martinbonnin, line 10:
```suggestion
    @Deprecated("This method is deprecated. Use an OperationMessageSerializer instead.")
```

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
martinbonnin: Damn, silly me 👍

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
```diff
@@ -0,0 +1,80 @@
...
+        }
+      }
+    }
+  }
+
+  companion object {
+    private const val JSON_KEY_DATA = "data"
+
+    @JvmOverloads
+    fun buildAppSyncWebSocketUrl(
```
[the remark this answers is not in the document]
martinbonnin: 👍 maybe rename to `buildWebSocketUrl` since there is already `AppSync` in `AppSyncOperationMessageSerializer`?

apollo-runtime/src/main/java/com/apollographql/apollo/subscription/AppSyncOperationMessageSerializer.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class AppSyncOperationMessageSerializer(
+    private val authorization: Map<String, Any?>
+) : OperationMessageSerializer {
+  override fun OperationClientMessage.writeTo(sink: BufferedSink) {
+    when (this) {
+      is OperationClientMessage.Start -> JsonWriter.of(sink).use { writeTo(it) }
+      is OperationClientMessage.Init,
+      is OperationClientMessage.Stop,
+      is OperationClientMessage.Terminate -> with(ApolloOperationMessageSerializer) { writeTo(sink) }
```
martinbonnin, line 22: This reads weirdly. I know it works but it can be tempting to replace with `ApolloOperationMessageSerializer.writeTo(sink)`, which doesn't. That's an argument to rename `writeTo` to `writeClientMessage(sink, message)`. This is what moshi does for an exemple and will be less surprising for users?

ansman: Yeah, the caller ergonomics isn't great. Will change.

apollo-runtime/src/test/java/com/apollographql/apollo/subscription/WebSocketSubscriptionTransportMessageTest.kt
```diff
@@ -0,0 +1,191 @@
...
+import okhttp3.Request
+import okhttp3.Response
+import okhttp3.WebSocket
+import okhttp3.WebSocketListener
+import okio.ByteString
+import org.junit.Before
+import org.junit.Test
+import java.math.BigDecimal
+
+class WebSocketSubscriptionTransportMessageTest {
```
ansman, line 15: I ended up rewriting this to Kotlin too since it had some warnings.

martinbonnin:
For the record, JapiCmp report there: https://jsfiddle.net/martinbonnin/ta2zkj51/1/show

There are a few binary breaking changes:
* `OperationClientMessage.toJsonString` became final
* `OperationClientMessage.writeToJson` became final and moved to supertype
* `OperationServerMessage.fromJsonString` is removed

I think it's mostly ok as it's for the most part internal APIs that shouldn't be used much in client code.

martinbonnin (approved):

martinbonnin:
Thanks for the awesome contribution!

ansman:
> For the record, JapiCmp report there: https://jsfiddle.net/martinbonnin/ta2zkj51/1/show
> 
> 
> 
> There are a few binary breaking changes:
> 
> * `OperationClientMessage.toJsonString` became final

Making it final is ok because the constructor of the class was package private 
> 
> * `OperationClientMessage.writeToJson` became final and moved to supertype
> 

Making it final is ok because the constructor of the class was package private. Moving it to the super class is binary compatible AFAIK.
> * `OperationServerMessage.fromJsonString` is removed

I forgot to add `@JvmStatic` to that one. If you add it it will be binary compatible. 
> 
> 
> 
> I think it's mostly ok as it's for the most part internal APIs that shouldn't be used much in client code.

I definitely think we should fix these as my existing code depended on both of these JSON methods and transitive dependencies might use them which would lead to runtime crashes.

martinbonnin:
> I definitely think we should fix these as my existing code depended on both of these JSON methods

This is code that's not needed anymore once upstreamed, right?  

>  and transitive dependencies might use them which would lead to runtime crashes.

Sound good, I'll add the `@JvmStatic` to make sure we're on the safe side.

martinbonnin:
New JapiCmp report after #2812 : https://jsfiddle.net/martinbonnin/hn0z9bao/show

ansman:
That's right. But it needs to be kept until a major bump or things like Amplify which uses Apollo could break

martinbonnin:
#2812 is merged

PR #2816, opened 2020-12-23
Remove fragment flattening

sav007:
Remove flattening and keep nested fragments.

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
+++ b/apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
@@ -116,7 +116,7 @@ class TestQuery : Query<TestQuery.Data, Operation.Variables> {
 
       fun marshaller(): ResponseFieldMarshaller
 
-      interface Character : Hero, HeroDetail, HeroDetail.Character, HumanDetail.Character {
+      interface Character : Hero, HeroDetail {
         override val __typename: String
 
         /**
@@ -124,12 +124,23 @@ class TestQuery : Query<TestQuery.Data, Operation.Variables> {
          */
         override val name: String
 
-        /**
-         * The date character was born.
-         */
-        override val birthDate: Any
-
         override fun marshaller(): ResponseFieldMarshaller
+
+        interface Character : Hero.Character, HeroDetail.Character {
+          override val __typename: String
+
+          /**
+           * The name of the character
+           */
+          override val name: String
+
+          /**
+           * The date character was born.
+           */
+          override val birthDate: Any
+
+          override fun marshaller(): ResponseFieldMarshaller
+        }
       }
 
       interface Human : Hero, HumanDetail {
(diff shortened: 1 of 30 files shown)


apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
```diff
@@ -141,43 +152,154 @@ class TestQuery : Query<TestQuery.Data, Operation.Variables> {
...
+           * What this human calls themselves
+           */
+          override val name: String
+
+          /**
+           * The date character was born.
+           */
+          override val birthDate: Any
+
+          override fun marshaller(): ResponseFieldMarshaller
```
martinbonnin, line 169: Shouldn't we put the `marshaller()` on the class rather than the interface?

apollo-compiler/src/test/graphql/com/example/inline_fragment_inside_inline_fragment/TestQuery.kt
```diff
@@ -122,64 +122,139 @@ class TestQuery : Query<TestQuery.Data, Operation.Variables> {
...
-        override fun marshaller(): ResponseFieldMarshaller
+          /**
+           * The home planet of the human, or null if unknown
+           */
+          val homePlanet: String?
+
+          override fun marshaller(): ResponseFieldMarshaller
+        }
+
+        interface Droid : Character {
```
martinbonnin, line 142: I would drop the `Character`inheritance ( and `__typename` and `name` fields here) to just keep:

```
interface Droid {
  val primaryFunction: String?
}
```

That would simplify the codegen a bunch. There's a minor annoyance that users won't be able to do `hero.asDroid()?.name` anymore but they can still achieve the same results with something like

```
when (hero) {
  is DroidHero -> // do something with hero.name and hero.primaryFunction
}
```

It might actually simplify the mental model since inline fragments would be closer to named fragments

apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
sav007: Hm, user will mostly work with interfaces when it comes to fragments, so we need to provide access to marshaller via interface rather than via fragment implementation that can be intersection of fragments.

apollo-compiler/src/test/graphql/com/example/inline_fragment_inside_inline_fragment/TestQuery.kt
sav007: We can't do that as we merge parent fields from Character, so any fields that defined in parent field / fragment will be merged down to Droid interface. Otherwise we can easily run into clash of fields when build implementation for multiple fragments, like for instance:
```
interface Droid {
  friends: List<Friend>

  interface Friend {
    ...
  }
}

interface Character {
 friends: List<Friend>

  interface Friend {
    ...
  }
}

data class DroidCharacterHero(val friends: List<Friend>): Character, Droid

```

here we going to clash on friends as it ambiguous type

apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
martinbonnin: I'm not sure how many users use `marshaller()` directly. My assumption was that very few users need to write  data/fragments and those who do use the imperative store APIs instead of calling `marshaller` directly.

More philosophically, what does marshalling an interface mean? There has to be a matching implementation to marshal something which is why it feels weird to have the method on the interface.

apollo-compiler/src/test/graphql/com/example/inline_fragment_inside_inline_fragment/TestQuery.kt
martinbonnin: > here we going to clash on friends as it ambiguous type

Since we're going to generate a data class for `Friend`, we could implement the different interfaces there?

```
interface Droid {
  friends: List<Friend>

  interface Friend {
    ...
  }
}

interface Character {
 friends: List<Friend>

  interface Friend {
    ...
  }
}

data class DroidCharacter(val friends: List<Friend>): Character, Droid {
  data class Friend([...]) : Character.Friend, Droid.Friend
}
```

This is more symmetrical.

apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
sav007: Just to make sure we are on the same page, you want to remove `marshaller ` from the generated nested interfaces and leave it only for top level generated operation data classes:  `Operation.Data#responseFieldMapper()` and for `GraphqlFragment#marshaller`, right?

apollo-compiler/src/test/graphql/com/example/inline_fragment_inside_inline_fragment/TestQuery.kt
sav007: hm, you are right.... let me try this again, I don't remember what was the issue and why this didn't work out last time, smth related to nested fragments. I think we can merge this PR and I will try to optimize this in next one.

apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
martinbonnin: The initial thought was to move them from the interfaces to the data classes but I think it makes sense to remove them from nested interfaces/classes altogether and only keep them in `Operation.Data#responseFieldMapper()` and `GraphqlFragment#marshaller`. Thinking more about this, maybe we could even remove `marshaller` from the public API and expose `Operation.Data#responseAdapter()` instead?

apollo-compiler/src/test/graphql/com/example/inline_fragment_inside_inline_fragment/TestQuery.kt
martinbonnin: Sounds good to me!

martinbonnin (approved):

apollo-compiler/src/test/graphql/com/example/fragment_used_twice/TestQuery.kt
sav007: Yeah, that makes sense, will do that in next PR

PR #2819, opened 2020-12-24
Cleanup the custom scalar APIs 1/3

martinbonnin:
First PR of a series of PR to tweak the custom scalar APIs. I wanted to do everything in 1 PR but this turned out to be pretty big already so I'll split it in multiple PRs, hopefully, not too many.

Renaming/consistency is the main change in this PR. `CustomType` is replaced by `CustomScalar` to make it clear that there is no other custom type than scalar types in GraphQL. There are a few other changes outlined in the diff.

_Note_: For the record, I considered generating the CustomScalarAdapters from the Gradle config as part of this PR so that we wouldn't need to register them manually. Something like:

```
// build.gradle.kts

apollo {
  service("github") {
    customScalar("DATE", "java.util.Date", "com.example.Iso8601DateAdapter")
    customScalar("URL", "kotlin.String", "com.example.UrlAdapter")
  }
}

// Main.kt
// GithubBuilder would be auto-generated and pre-configured with the adapters from the Gradle config
val apolloClient = ApolloClient.GithubBuilder().build()
```

While that makes the Gradle config the single source of truth for custom scalars, I ultimately gave up on this idea because:
1. it forces adapters to have a no-arg constructor, which is a weird limitation
2. more importantly, it means generated code now requires `apollo-runtime-kotlin` in the classpath, which is not something we want. We could configure that in Gradle but the additional configuration would make it more complicated than actually having to register adapters manually.


apollo-android-support/api.txt
```diff
@@ -1,31 +0,0 @@
-// Signature format: 3.0
-package com.apollographql.apollo {
```
martinbonnin, line 2: Removed the `api.txt` for now. I'll re-add them for `3.0.0-alpha01`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/BuiltinAdapters.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.apollographql.apollo.api.internal.json.Utils
+import com.apollographql.apollo.api.internal.json.use
+import okio.Buffer
+
+/**
+ * Builtin CustomScalarTypeAdapter provided for convenience. Encoding is most of the times straightforward. Decoding
+ * can involve coercion. If you need stricter decoding or different logic, define your own adapter
+ */
+object BuiltinScalarTypeAdapters {
+  val STRING_ADAPTER = adapterWithDefaultEncode { jsonElement ->
```
martinbonnin, line 13: The builtin CustomScalarTypeAdapters are now public so that users can re-use them if required.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/BuiltinAdapters.kt
```diff
@@ -0,0 +1,113 @@
...
+    when (jsonElement) {
+      is JsonElement.JsonString -> jsonElement.value
+      is JsonElement.JsonNull -> null
+      is JsonElement.JsonBoolean -> jsonElement.value.toString()
+      is JsonElement.JsonNumber -> jsonElement.value.toString()
+      is JsonElement.JsonObject,
+      is JsonElement.JsonList -> {
```
martinbonnin, line 20: `JsonElement` is the new, more type safe `CustomTypeValue`. It makes it clear that `CustomTypeAdapters` ultimately work on Json:

```kotlin
interface CustomScalarTypeAdapter<T> {
  fun decode(jsonElement: JsonElement): T
  fun encode(value: T): JsonElement
}
```

I know GraphQL is not Json-only but so far it is so we'll see when (if...) that changes.

apollo-idling-resource/src/test/java/com/apollographql/apollo/test/espresso/ApolloIdlingResourceTest.kt
```diff
@@ -0,0 +1,210 @@
+package com.apollographql.apollo.test.espresso
```
martinbonnin, line 1: Converted a bunch of files to Kotlin as I'd like to get rid of `@JvmField` annotation for `3.0.0-alpha01`

sav007:
Just for the context, back in a day having `CustomType* ` naming was from the  idea to give user more flexibility and allow to override any type defined in GraphQL schema not only custom scalar types, but it's too complex and necessary overhead

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/JsonElement.kt
```diff
@@ -0,0 +1,61 @@
+package com.apollographql.apollo.api
+
+/**
+ * A wrapper class for representation of custom GraphQL type value, used in user provided [CustomScalarAdapter]
+ * encoding / decoding functions.
+ **/
+sealed class JsonElement {
```
sav007, line 7: nit: should we move this into `/json` subfolder?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
```diff
@@ -1,17 +1,17 @@
 package com.apollographql.apollo.api
 
 /**
- * Represents a custom GraphQL scalar type
+ * Represents a mapping from a custom GraphQL scalar type to a Java/Kotlin class
  */
 interface ScalarType {
```
sav007, line 6: nit: should we rename this to `CustomScalarType`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
```diff
@@ -1,129 +1,65 @@
 package com.apollographql.apollo.api
 
-import com.apollographql.apollo.api.CustomTypeValue.*
-import com.apollographql.apollo.api.CustomTypeValue.Companion.fromRawValue
-import com.apollographql.apollo.api.internal.json.JsonWriter
-import com.apollographql.apollo.api.internal.json.Utils
-import com.apollographql.apollo.api.internal.json.use
-import okio.Buffer
-import kotlin.jvm.JvmField
+class ScalarTypeAdapters(val customScalarAdapters: Map<ScalarType, CustomScalarAdapter<*>>) {
```
sav007, line 3: same should we use naming `CustomScalar*` everywhere?

sav007 (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarType.kt
sav007: nvm I see it's done in next PR https://github.com/apollographql/apollo-android/pull/2820

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ScalarTypeAdapters.kt
sav007: nvm I see it's done in next PR #2820

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/JsonElement.kt
martinbonnin: Good question. We currently have `com.apollographql.apollo.api.internal.json` where the Json parser (`BufferedSourceJsonReader`) resides. 

Technically, we could make the json parser work with `JsonElement` but that will introduce some unnecessary wrapping and I'm afraid it'll only slow down parsing... 
So `JsonElement` is right now only used to wrap Kotlin instances into more type-safe/user-friendly types. In that sense I feel that it should be closer to `CustomScalarAdapter` than `BufferedSourceJsonReader`, even if it's about Json. Also the `com.apollographql.apollo.api` package isn't too much cluttered right now. It contains ~30 classes, some of which can maybe be moved to `internal`. All that makes me want to keep it in the root package for the time being but I'm not against reorganizing things differently if we ultimately think it makes more sense.

martinbonnin:
> Just for the context, back in a day having CustomType* naming was from the idea to give user more flexibility and allow to override any type defined in GraphQL schema not only custom scalar types, but it's too complex and necessary overhead

Got it! That explains a lot. Yea, giving the user the possibility to override arbitrary object types sounds ... complex !

Issue #2818, opened 2020-12-23
`ApolloStore#writeAndPublish` overwrites cache entries when using nested fragments


mateuszkwiecinski:
**Summary**
When calling `writeAndPublish` for an `operation` that fetches part of the schema using graphql fragments, then the information saved in cache doesn't reflect passed `operationData`.

**Version**
2.4.6

**Description**
Context: We try to treat Apollo cache as a source of a truth, leveraing `ApolloStore#writeAndPublish` as a way to help Apollo cross-cache between queries or to i.e. fill incomplete data - basically we use it in edge cases, as a last resort.
I noticed that despite manually writing data to cache we still expecrience cache misses and unexpected api calls caused by a wrong cache state. I was able to narrow down the issue down to simple repro:

Given following schema:
```graphql
type Query {
  home: Home!
}

type Home {
  sectionA: SectionA
}

type SectionA {
  id: String!
  name: String!
  imageUrl: String
}
```
and a following query:
```graphql
query Home {
  home {
    ...sectionFragment
    sectionA {
      name
    }
  }
}

fragment sectionFragment on Home {
  sectionA {
    id
    imageUrl
  }
}
```
when following code invoked:
```kotlin
apolloStore.write(
      HomeQuery(),
      HomeQuery.Data(
        HomeQuery.Home(
          sectionA = HomeQuery.SectionA(
            name = "section-name"
          ),
          fragments = HomeQuery.Home.Fragments(
            sectionFragment = SectionFragment(
              sectionA = SectionFragment.SectionA(
                id = "section-id",
                imageUrl = "https://...",
              ),
            )
          )
        )
      )
    ).execute()
```
then cache gets filled with:
```
LruNormalizedCache {
  "section-id" : {
    "__typename" : SectionA
    "id" : section-id
    "imageUrl" : https://...
  }

  "QUERY_ROOT" : {
    "home" : CacheRecordRef(home)
  }

  "home" : {
    "__typename" : Home
    "sectionA" : CacheRecordRef(section-id)
  }
}
```

Expected result:
I'd expect the `name=section-name` to be saved under "section-id" cache key. 
```
  "section-id" : {
    "__typename" : SectionA
    "id" : section-id
    "imageUrl" : https://...
    "name" : section-name
  }

```

More context: 
 - From what I understood the issue appears here:
[...]

Issue #2486, opened 2020-07-29
Tweak the `ScalarType` API


martinbonnin:
Following up on https://github.com/apollographql/apollo-android/pull/2473#issuecomment-664906541, I didn't realise that the default adapters were use to save users to call `addCustomTypeAdapter`:

 Some adapters like `java.util.Date` currently have to be manually added using `addCustomTypeAdapter` while others like `com.apollographql.apollo.api.FileUpload` are registered automatically. 

We could instead declare the implementations and let the user register them as needed:

```kotlin
// build.gradle.kts
apollo {
   customTypeMapping.set(mapOf("TimeStamp" to "java.lang.Long"))
}

// Apollo.kt
apolloClientBuilder.addCustomTypeAdapter(CustomType.TimeStamp, ApolloCustomTypeAdapter.LongAdapter)
```

This makes everything explicit.

tasomaniac:
What is the benefit of removing them? `FileUpload` type is added automatically to make it more convenient I guess. File upload feature is also weird anyways :)

martinbonnin:
> What is the benefit of removing them?

Mostly API consistency and discoverability. 

There's a super tiny argument for performance, it might be faster to lookup the correct adapter if the list is small. But I wouldn't bet too much on that, if we do this, I think this would be mostly for API design reasons.

> FileUpload type is added automatically to make it more convenient I guess.

Yep, exactly. The default adapters are a convenience for some use cases. But to me that's some magic that hides how custom adapters are working. It took me a while to figure out why it wasn't required to register an adapter.

Also somewhat related, I think there is a confusion that CustomTypeAdapters could be use to "transform" generated models (and not only scalars), like moshi [custom type adapters](https://github.com/square/moshi#custom-type-adapters). Maybe we should rename to `ScalarTypeAdapter`.

Overall, small changes would make this API hopefully easier to understand.

tasomaniac:
The last part is an important one. I am all for renaming it. Your suggestion is much more clear.

martinbonnin:
We should also throw an error if trying to register a scalar type that's not in the schema: https://github.com/apollographql/apollo-android/issues/2721

PR #2813, opened 2020-12-21
Merge main into `dev-3.x`

martinbonnin:

Issue #2721, opened 2020-11-02
Custom type generating not working


ribbon-xx:
Apollo version: 2.4.1
Build gradle: 4.1.0
Gradle wrapper: 6.5-all

build.gradle:
```
android {}
apollo {
    useSemanticNaming = true
    generateKotlinModels = true
    customTypeMapping = [
            "Date" : "java.util.Date"
    ]
}
```
CustomType.java as attachment.

Adapter
```
val sdfISO8601 = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss")
        val dateCustomTypeAdapter = object : CustomTypeAdapter<Date> {
            override fun decode(value: CustomTypeValue<*>): Date {
                return try {
                    sdfISO8601.parse(value.value.toString())
                } catch (e: ParseException) {
                    throw RuntimeException(e)
                }
            }

            override fun encode(value: Date): CustomTypeValue<*> {
                return CustomTypeValue.GraphQLString(sdfISO8601.format(value))
            }
        }
```

Usage of Adapter:
```
apolloClient = ApolloClient.builder()
            .serverUrl(BASE_URL)
            .okHttpClient(okHttpClient)
            .normalizedCache(normalizedCacheFactory, cacheKeyResolver)
            .addCustomTypeAdapter(CustomType.DATE, dateCustomTypeAdapter)
            .build()
```

Could you please tell me how to make apollo generates the right CustomType file?

![Screen Shot 2020-11-02 at 21 40 44](https://user-images.githubusercontent.com/457409/97880710-14e5e080-1d54-11eb-9053-40c8c9026e48.png)

martinbonnin:
Hi 👋 ! Looks like you're generating Kotlin models but are looking at a Java file so something's wrong there. Either set `generateKotlinModels = false` or use `CustomType.kt`

ribbon-xx:
> Hi 👋 ! Looks like you're generating Kotlin models but are looking at a Java file so something's wrong there. Either set `generateKotlinModels = false` or use `CustomType.kt`

Thanks for your reply!
I think my problem is I cannot import CustomType.kt, I've searched all the project's folders but only CustomType.java.
Could you tell me how can I have CustomType.kt, thanks you so much

ribbon-xx:
I've made a clean and rebuild task, CustomType.kt is generated but it does not contain the key in build.gradle file, please take a look at the 
![Uploading Screen Shot 2020-11-03 at 01.03.55.png…]()
[...]

PR #2820, opened 2020-12-24
Cleanup the custom scalar APIs 2/3

martinbonnin:
Follow up from #2818,

This:

* renames `ScalarTypeAdapters` to `CustomScalarAdapters` 
* renames `ScalarType` to `CustomScalar`
* generates `CustomScalar`s as data classes

I'll make one last PR to throw if trying to register an unknown CustomScalar

See https://github.com/apollographql/apollo-android/issues/2486


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CustomScalar.kt
```diff
@@ -3,15 +3,17 @@ package com.apollographql.apollo.api
...
-  val graphqlName: String
+  val graphqlName: String,
 
   /**
    * Fully qualified class name this GraphQL scalar type is mapped to (e.g. `java.lang.String`, `java.net.URL`, `java.util.DateTime`)
    */
   val className: String
+) {
+  // Do not remove, this is used by generated code
+  companion object
```
sav007, line 18: where it's used? I might missed that during reviewing PR

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/type/CustomScalars.kt
```diff
@@ -0,0 +1,8 @@
+// AUTO-GENERATED FILE. DO NOT MODIFY.
+//
+// This class was automatically generated by Apollo GraphQL plugin from the GraphQL queries it found.
+// It should not be modified by hand.
+//
+package com.example.fragment_in_fragment.type
+
+object CustomScalars
```
sav007, line 8: Should we skip empty object generation?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -1,5 +1,8 @@
 package com.apollographql.apollo.compiler.backend.ast
 
+import com.squareup.kotlinpoet.MemberName
+import com.squareup.kotlinpoet.TypeName
+
```
sav007, line 5: Idea was that this class AST supposed to be free from any specific kotlin / java poet classes and can be used with any code generation libraries. With these changes we lock AST to be used only with Kotlin Poet.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/AstBuilder.kt
```diff
@@ -665,18 +643,14 @@ internal class AstBuilder private constructor(
...
                 nullable = true,
                 schemaType = schemaTypeRef.name,
-                type = customType,
-                customEnumType = CodeGenerationAst.TypeRef(
-                    name = schemaTypeRef.name,
-                    packageName = typesPackageName,
-                    enclosingType = CodeGenerationAst.customTypeRef(typesPackageName)
-                )
+                type = className,
+                typeName = ClassName(typesPackageName, "CustomScalars", schemaTypeRef.name.capitalize())
```
sav007, line 653: Should we introduce `object CustomScalarType` in `CustomScalarType.kt` with the const for enum name `CustomScalars`. Also I was thinking maybe we should use `object`  as namespace for any top level functions?  For example move all functions in `ObjectTypeCodegen.kt` into `object ObjectTypeCodegen `? So any top level functions will have explicit namespaces

apollo-compiler/src/test/graphql/com/example/antlr_tokens/type/CustomScalars.kt
```diff
@@ -0,0 +1,16 @@
+// AUTO-GENERATED FILE. DO NOT MODIFY.
+//
+// This class was automatically generated by Apollo GraphQL plugin from the GraphQL queries it found.
+// It should not be modified by hand.
+//
+package com.example.antlr_tokens.type
+
+import com.apollographql.apollo.api.CustomScalar
+
+object CustomScalars {
```
sav007, line 10: Didn't we have PR into 2.x branch that introduces some interface for `CustomScalars` when it's generated as sealed class? Also I believe we had compiler option to generate this as sealed class, are we sure to change this behaviour in 3.x?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CustomScalar.kt
martinbonnin: My bad, it's a leftover from a previous attempt to generate CustomScalars as extension properties like `CustomScalar.Date`. But it leads to nasty import clashes (extension property vs `java.util.Date` for an exemple) so I reverted to using `CustomScalars.Date` (See https://github.com/apollographql/apollo-android/pull/2820/commits/da7da119a3d58144aa9225955e9e7a4268b65b95)

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/type/CustomScalars.kt
martinbonnin: I feel like consistency is a better dev experience and adheres to the principle of least astonishment (see https://github.com/apollographql/apollo-android/issues/2721 for an exemple)
No strong feeling there though. If generating an empty object it causes any problem, we should remove it.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
martinbonnin: Good catch! I'll fix this.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/AstBuilder.kt
martinbonnin: > Should we introduce object CustomScalarType in CustomScalarType.kt with the const for enum name CustomScalars

Not sure I get it. `CustomScalars` is now a generated `object`. I'd rather avoid having 2 instances in 2 different packages. 

> Also I was thinking maybe we should use object as namespace for any top level functions?

Yea, it's a bit heterogeneous at the moment. I don't mind it too much since this is mostly internal `compiler` API. The functions in `ObjectType.kt` are mostly extension functions though so IntelliJ will auto-add any enclosing `object` and even if we wrap them, they will look the same at the call site, only the import will change:

```
// this import is currently not needed because everything is in the same package.
import com.apollographql.apollo.compiler.backend.codegen.ObjectTypeCodegen

... 

dataType.typeSpec()
```

It's still convenient being able to completely leave out the import?

apollo-compiler/src/test/graphql/com/example/antlr_tokens/type/CustomScalars.kt
martinbonnin: This was about Enums and the `EnumValue` interface: https://github.com/apollographql/apollo-android/blob/d59f2a1b07c62b7094a8ecfd56dd57fe5f89cd63/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/EnumValue.kt#L6

This should be backported to 3.0. as part of https://github.com/apollographql/apollo-android/pull/2813

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/type/CustomScalars.kt
sav007: I think it will look odd for the user to see empty class generated and visible for import.

sav007: Also I have a suggestion to generate additional KDoc that briefly explains the purpose of this generated class, what do you think?

martinbonnin: Yup, +1000 for KDoc, I'll add that. And remove the generated class if empty.

sav007 (approved):

apollo-compiler/src/test/graphql/com/example/fragment_in_fragment/type/CustomScalars.kt
martinbonnin: Kdoc and skipping empty classes are added

PR #2824, opened 2020-12-29
Adding ApolloAndroidLogger to apollo-android-support. 

AdamMc331:
This introduces a helpful wrapper for the `apollo-android-support` module that implements the `Logger` interface from Apollo to output to the logcat. 

I've also added this logger to the KotlinSampleApp to log hits and misses to the `HttpCache`. 

Here is a sample screenshot from the logcat:
Screen Shot 2020-12-29 at 3 52 52 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ dependencies {
   implementation("com.apollographql.apollo:apollo-coroutines-support:x.y.z")
   // optional: for RxJava3 support  
   implementation("com.apollographql.apollo:apollo-rx3-support:x.y.z")
+  // optional: for additional Android utilities
+  implementation("com.apollographql.apollo:apollo-android-support:x.y.z")
   // optional: if you just want the generated models and parsers and write your own HTTP code/cache code, you can remove apollo-runtime
   // and use apollo-api instead  
   implementation("com.apollographql.apollo:apollo-api:x.y.z")
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ dependencies {
   implementation("com.apollographql.apollo:apollo-coroutines-support:x.y.z")
   // optional: for RxJava3 support  
   implementation("com.apollographql.apollo:apollo-rx3-support:x.y.z")
-  // optional: for additional Android utilities
+  // optional: Most of apollo-android does not depend on Android in practice and runs on any JVM or on Kotlin native. apollo-android-support contains a few Android-only helper classes. For an example to send logs to logcat or run callbacks on the main thread.
   implementation("com.apollographql.apollo:apollo-android-support:x.y.z")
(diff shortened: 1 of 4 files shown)


apollo-android-support/src/main/java/com/apollographql/apollo/ApolloAndroidLogger.kt
```diff
@@ -0,0 +1,23 @@
+package com.apollographql.apollo
+
+import android.util.Log
+import com.apollographql.apollo.Logger
+
+/**
+ * This is an Android wrapper around [Logger] that will take any messages Apollo wants
+ * to print and log them in the Android logcat.
+ */
+class ApolloAndroidLogger : Logger {
```
AdamMc331, line 10: It's worth thinking about what this should mean for Android devs who don't want these logs in production, and whether Apollo wants to be opinionated about this. Perhaps we can provide a solution to turn them off, or just rely on devs to make sure they don't include this logger in production.

If we want to do that, maybe we can add some additional notes inside the documentation here.

Let me know what y'all think!

README.md
```diff
@@ -69,6 +69,8 @@ dependencies {
   implementation("com.apollographql.apollo:apollo-coroutines-support:x.y.z")
   // optional: for RxJava3 support  
   implementation("com.apollographql.apollo:apollo-rx3-support:x.y.z")
+  // optional: for additional Android utilities
```
martinbonnin, line 72: Maybe make it a bit more explicit? 

```suggestion
  // optional: Most of apollo-android does not depend on Android in practice and runs on any JVM or on Kotlin native. apollo-android-support contains a few Android-only helper classes. For an example to send logs to logcat or run callbacks on the main thread.
```

apollo-android-support/src/main/java/com/apollographql/apollo/ApolloAndroidLogger.kt
martinbonnin: >  Perhaps we can provide a solution to turn them off, or just rely on devs to make sure they don't include this logger in production.

I'd prefer that decision to be left to the caller. The definition of what "production" is might vary depending the teams/products so I think it's better if Apollo Android doesn't know anything about this. Also, not sending log is as easy as not setting a logger so it should be reasonnably doable at the callsite

martinbonnin (approved):
Thanks!

apollo-android-support/src/main/java/com/apollographql/apollo/ApolloAndroidLogger.kt
AdamMc331: Agreed! So let me know if you think that warrants any additional documentation, but I don't think so - we should rely on the callers to be aware of this concept. :)

PR #2825, opened 2020-12-30
Don't copy fragment interfaces to implementation classes

sav007:
Before we were copying generated interfaces to fragment implementations. It appears to be redundant.


martinbonnin:
Looks good to me, this will certainly clash with the custom scalar stuff, I'll fix conflicts later today.

martinbonnin (approved):

PR #2830, opened 2021-01-03
Make parse / composeRequestBody as extension functions

sav007:
Remove code generation for these functions.


martinbonnin:
This looks nice 👍 . I like how it makes the generated code way smaller!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationDataExtensions.kt
```diff
@@ -1,48 +0,0 @@
...
- *
- * @param indent the indentation string to be repeated for each level of indentation in the encoded document. Must be a string
- * containing only whitespace. If [indent] is an empty String the encoded document will be compact. Otherwise the encoded
- * document will be more human-readable.
- * @param customScalarAdapters configured instance of custom GraphQL scalar type adapters. Default adapters are used if this
- * param is not provided.
- */
-@JvmName("serialize")
-@JvmOverloads
-fun Operation.Data.toJson(indent: String = "", customScalarAdapters: CustomScalarAdapters = DEFAULT): String {
```
martinbonnin, line 39: I removed it because it was clashing with the declaration from `OperationExtensions` below

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationDataExtensions.kt
```diff
@@ -1,48 +0,0 @@
...
- *    }
- * }</pre>
- *
- * @param indent the indentation string to be repeated for each level of indentation in the encoded document. Must be a string
- * containing only whitespace. If [indent] is an empty String the encoded document will be compact. Otherwise the encoded
- * document will be more human-readable.
- * @param customScalarAdapters configured instance of custom GraphQL scalar type adapters. Default adapters are used if this
- * param is not provided.
- */
-@JvmName("serialize")
```
martinbonnin, line 37: About naming, I actually like `toJson` better and I think it can be used for both Java and Kotlin callers. There's a case to be made that `ResponseAdapter.toJson` should be named `serialize` (or something else) though because it can serialize to a cache in addition to Json.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
```diff
@@ -0,0 +1,132 @@
...
+      withQueryDocument = true,
+      customScalarAdapters = customScalarAdapters
+  )
+}
+
+/**
+ * Parses GraphQL operation raw response from the [source] with provided [customScalarAdapters] and returns result [Response]
+ */
+@JvmOverloads
+fun <DATA : Operation.Data> Operation<DATA, *>.parse(
```
martinbonnin, line 117:
```suggestion
fun <D : Operation.Data> Operation<D, *>.parse(
```

To stay consistent with other parts of the code, That'll make it easier to do find & replace if needed.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
```diff
@@ -0,0 +1,132 @@
...
+    source: BufferedSource,
+    customScalarAdapters: CustomScalarAdapters = DEFAULT
+): Response<DATA> {
+  return SimpleOperationResponseParser.parse(source, this, customScalarAdapters)
+}
+
+/**
+ * Parses GraphQL operation raw response from the [byteString] with provided [customScalarAdapters] and returns result [Response]
+ */
+fun <DATA : Operation.Data> Operation<DATA, *>.parse(
```
martinbonnin, line 127: Same as above

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo/testing/MockQuery.kt
```diff
@@ -14,6 +14,9 @@ class MockQuery : Query<MockQuery.Data, Operation.Variables> {
 
   override fun responseFieldMapper(): ResponseFieldMapper<Data> {
     return ResponseFieldMapper {
+      while (it.selectField(emptyArray()) != -1) {
+        // consume the json stream
+      }
```
martinbonnin, line 19: The more I think about it, the more I think these tests should be integration tests. We have a bunch of tests that use `MockQuery/Subscription` and that are very fragile to codegen changes

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationDataExtensions.kt
sav007: Hm, strange I thought I deleted this file

martinbonnin (approved):

PR #2831, opened 2021-01-04
Add fun Operation.adapter()

martinbonnin:
This PR is based on https://github.com/apollographql/apollo-android/pull/2830 so the diff is slightly bigger than what it should be.

It removes `ResponseFieldMapper` and uses `ResponseAdapter` instead.

~Edit: somehow Github didn't start actions here 🤷  so the build status and conflicts below are mostly out of date~ Github picked up this PR now


apollo-integration/src/test/kotlin/com/apollographql/apollo/NormalizedCacheTestCase.kt
```diff
@@ -353,7 +356,7 @@ class NormalizedCacheTestCase {
         Predicate<Response<HeroAndFriendsNamesWithIDsQuery.Data>> { response -> !response.hasErrors() }
     )
     val heroWithFriendsFragment = apolloClient.apolloStore.read(
-        HeroWithFriendsFragment.Mapper(),
+        HeroWithFriendsFragmentImpl_ResponseAdapter,
```
martinbonnin, line 359: I like that using the imperative store API forces to reference the fragment implementation.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adaptable.kt
```diff
@@ -0,0 +1,7 @@
+package com.apollographql.apollo.api
+
+import com.apollographql.apollo.api.internal.ResponseAdapter
+
+interface Adaptable<T> {
+  fun adapter(): ResponseAdapter<T>
```
martinbonnin, line 6: A new interface for classes that can `Adapt` others. I'm not 100% convinced on the naming but there is some potential to uniformize a lot of the code in `RealApolloStore` for operation/fragments

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/GraphqlFragment.kt
```diff
@@ -1,14 +1,3 @@
...
- * Represents a GraphQL fragment
- */
-interface GraphqlFragment {
-
-  /**
-   * Returns marshaller [ResponseFieldMarshaller] to serialize fragment data
-   */
-  fun marshaller(): ResponseFieldMarshaller
-}
+interface GraphqlFragment
```
sav007, line 3: So fragment is no longer `Adaptable `?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/codegen/FragmentType.kt
```diff
@@ -81,6 +67,14 @@ internal fun CodeGenerationAst.FragmentType.implementationTypeSpec(generateAsInt
       .typeSpec()
       .toBuilder()
       .addSuperinterface(GraphqlFragment::class)
+      .addSuperinterface(Adaptable::class.asTypeName().parameterizedBy(defaultImplementationType.typeRef.asTypeName()))
+      .addFunction(
+          FunSpec.builder("adapter")
+              .addModifiers(KModifier.OVERRIDE)
+              .returns(ResponseAdapter::class.asTypeName().parameterizedBy(defaultImplementationType.typeRef.asTypeName()))
```
sav007, line 74: nit: 
`this.defaultImplementationType.typeRef.asAdapterTypeName()`

apollo-normalized-cache/src/jvmMain/kotlin/com/apollographql/apollo/cache/normalized/ApolloStore.kt
```diff
@@ -138,22 +137,20 @@ interface ApolloStore {
...
    *
-   * @param fieldMapper [ResponseFieldMapper] to be used for field mapping
    * @param cacheKey    [CacheKey] to be used to find cache record for the fragment
    * @param variables   [Operation.Variables] required for fragment arguments resolving
    * @param <F>         type of fragment to be read
    * @return {@ApolloStoreOperation} to be performed, that will be resolved with cached fragment data
   </F> */
-  fun <F : GraphqlFragment> read(
-      fieldMapper: ResponseFieldMapper<F>,
+  fun <F> read(
```
sav007, line 152: we should probably leave only one API (remove `fun <D : Operation.Data, V : Operation.Variables> read(...)`

apollo-normalized-cache/src/jvmMain/kotlin/com/apollographql/apollo/cache/normalized/ApolloStore.kt
```diff
@@ -199,7 +196,7 @@ interface ApolloStore {
    * have changed
    */
   fun write(
```
sav007, line 198: same here I think we can leave only one `write`. Also for symmetry should it be
```
fun <T> write(data:T, adapter: ResponseAdapter<T>...)
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/GraphqlFragment.kt
sav007, line 3: nit: keep KDoc?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/codegen/FragmentType.kt
martinbonnin: This is actually not the same:

1.
```
override fun adapter(): ResponseAdapter<Data>
```

2.
```
override fun adapter(): TestQuery_ResponseAdapter
```

I think 1. is cleaner in the generated model, the `_ResponseAdapter` file being an implementation detail?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/GraphqlFragment.kt
martinbonnin: It's not that helpful but it does make the file nicer :D ! Overall, I'd like to put a more descriptive KDoc but I'm not sure what to put there. It actually feels like this interface isn't really useful anymore...

sav007 (approved):
LGTM

apollo-normalized-cache/src/jvmMain/kotlin/com/apollographql/apollo/cache/normalized/ApolloStore.kt
martinbonnin: Follow up PR for store APIs: https://github.com/apollographql/apollo-android/pull/2839

PR #2833, opened 2021-01-04
Add normalization benchmarks

martinbonnin:
Added a bunch of benchmarks including data normalization

Also replaced `ResponseJsonStreamReader` with `Utils.readRecursively`


benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Utils.kt
```diff
@@ -0,0 +1,51 @@
...
+      .source()
+      .buffer()
+
+  private val responseNormalizer = object : ResponseNormalizer<Map<String, Any?>>() {
+    override fun cacheKeyBuilder() = RealCacheKeyBuilder()
+
+    override fun resolveCacheKey(field: ResponseField, record: Map<String, Any?>) = CacheKey.NO_KEY
+  }
+
+  fun <D : Operation.Data> computeRecordsDuringParsing(operation: Operation<D>, bufferedSource: BufferedSource): Collection<Record?>? {
```
sav007, line 28: nit: should we rename it to more explicit like `parseWithInlinedNormilizer`?
Also I don't think we need to return records, as we don't use it?

benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Utils.kt
```diff
@@ -0,0 +1,51 @@
...
+    OperationResponseParser(
+        operation,
+        CustomScalarAdapters.DEFAULT,
+        responseNormalizer
+    ).parse(bufferedSource)
+
+    return responseNormalizer.records()
+  }
+
+  fun <D : Operation.Data> computeRecordsAfterParsing(operation: Operation<D>, bufferedSource: BufferedSource): Collection<Record?>? {
```
sav007, line 38: same here should we rename it to `parseAndNormilizeAfter` and maybe add small comment what is this for?

benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Utils.kt
martinbonnin: Works for me 👍

PR #2840, opened 2021-01-06
Update codegen for fragment implementations

sav007:


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -209,13 +204,6 @@ internal data class CodeGenerationAst(
   )
 
   companion object {
-    fun customTypeRef(typesPackageName: String): TypeRef {
```
sav007, line 212: not used anywhere

apollo-compiler/src/test/graphql/com/example/arguments_simple/fragment/HeroDetail.kt
```diff
@@ -73,9 +71,5 @@ interface HeroDetail : GraphqlFragment {
         |  }
         |}
         """.trimMargin()
-
-    operator fun invoke(reader: ResponseReader): HeroDetail {
-      return HeroDetailsImpl_ResponseAdapter.fromResponse(reader)
-    }
```
martinbonnin, line 79: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
```diff
@@ -120,19 +120,19 @@ internal data class BackendIr(
...
-      ) {
+  ) {
     data class SelectionSet(
+        val name: String,
         val fields: List<Field>,
         val fragments: Fragments,
         val typeCondition: IntrospectionSchema.TypeRef,
         val possibleTypes: Set<IntrospectionSchema.TypeRef>,
+        val defaultSelectionKey: SelectionKey,
         val selectionKeys: Set<SelectionKey>,
```
martinbonnin, line 136: Does `selectionKeys` need to be a `Set`? I only see usages of it containing one value. It actually looks like it's always equal to `defaultSelectionKey` ? Or do I miss something?

martinbonnin:
Looks nice 👍 ! I took the liberty of pushing a couple of fixes for tests.

More generally, should we put the fragments implementation generation behind an `opt-in` flag? My understanding is that this will only be used to read/write fragments to the store, which is an advanced feature. We might save a lot of users some confusion and build time if we disable it by default?

PR #2841, opened 2021-01-06
Separate parsing and normalization

martinbonnin:
Remove normalization from the parsing step. Normalization is now done from the models, like it is done when using the cache write APIs.

Advantages:
- Normalizing network response now uses the same path as when writing to the cache using the imperative cache write API. This ensures the results will be identical (see apollographql#2818)
- Normalization can be better decoupled from the parsing and not applied at all if no cache is configured. This also means we can make normalization "async" and send a model to the UI without having to wait

This also enables the streaming parser in `ApolloParseInterceptor`, which is the main use case, which is something like ~25% faster on the whole parsing + normalization path, potentially event faster without normalization 🎉  (see the README.md in the benchmarks for actual numbers)

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.kt
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.kt
@@ -28,7 +28,7 @@ import com.apollographql.apollo.cache.normalized.internal.RealCacheKeyBuilder
 import com.apollographql.apollo.cache.normalized.internal.ResponseNormalizer
 import com.apollographql.apollo.cache.normalized.internal.Transaction
 import com.apollographql.apollo.cache.normalized.internal.WriteableStore
-import com.apollographql.apollo.internal.response.RealResponseWriter
+import com.apollographql.apollo.api.internal.response.RealResponseWriter
 import java.util.ArrayList
 import java.util.Collections
 import java.util.LinkedHashSet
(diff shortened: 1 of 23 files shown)


apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.kt
```diff
@@ -349,15 +355,19 @@ class RealApolloStore(normalizedCache: NormalizedCache, cacheKeyResolver: CacheK
...
+        val responseReader = RealResponseReader(
+            operation.variables(),
+            rootRecord,
+            fieldValueResolver,
+            customScalarAdapters
+        )
         return try {
-          responseNormalizer.willResolveRootQuery(operation)
           val data = operation.adapter().fromResponse(responseReader, null)
+          val records = operation.normalize(data, customScalarAdapters, networkResponseNormalizer() as ResponseNormalizer<Map<String, Any>?>)
```
martinbonnin, line 366: Since we always use `neworkResponseNormalizer` here, we can remove the parameter and do other cleanups. I didn't do them yet in order to not clash with https://github.com/apollographql/apollo-android/pull/2839

apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloStore.kt
```diff
@@ -349,15 +355,19 @@ class RealApolloStore(normalizedCache: NormalizedCache, cacheKeyResolver: CacheK
...
           builder<D>(operation)
               .data(data)
               .fromCache(true)
-              .dependentKeys(responseNormalizer.dependentKeys())
+              .dependentKeys(records.dependentKeys()) // Do we need the dependentKeys here?
```
martinbonnin, line 370: Is there a specific reason why we would need the dependent keys when **reading** from the cache?

PR #2850, opened 2021-01-08
MapResponseParser and StreamResponseParser

martinbonnin:
Some miscellaneous API clean ups to clarify which APIs are streamed and which are not:

* `MapResponseParser` parses `data` and `errors` with a `MapResponseReader`. It is used to read from the cache and to read subscriptions
* `StreamResponseParser` parses `data` and `errors` with a `StreamResponseReader`. Is is faster and used to read from the network.
* Add `Operation.parse(map)` and `Operation.parseData(map)`

Also:
* Make `M` covariant in `ValueResolver`, that removes a bunch of unsafe casts from `Record` to `Map<String, Any?>`
* Use fun interfaces for ApolloStore transactions


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/MapResponseParser.kt
```diff
@@ -0,0 +1,88 @@
...
+import com.apollographql.apollo.api.Response.Companion.builder
+import com.apollographql.apollo.api.ResponseField
+import com.apollographql.apollo.api.parseData
+
+/**
+ * [MapResponseParser] parses network responses, including data, errors and extensions from a regular Map<String, Any?>.
+ * For better performance, you can parse a response from a [okio.BufferedSource] directly with [StreamResponseParser].
+ * That will avoid the cost of having to create an entire Map in memory
+ */
+object MapResponseParser {
```
sav007, line 16: just a thought should we call it BufferedResponseParser?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
```diff
@@ -134,11 +137,43 @@ fun <D : Operation.Data> Operation<D>.parse(
...
+/**
+ * Parses GraphQL operation raw response from [map] with provided [customScalarAdapters] and returns result [Response]
+ *
+ * @param map: a [Map] representing the response. It typically include a "data" field
+ */
+fun <D : Operation.Data, M: Map<String, Any?>> Operation<D>.parseData(
+    map: M,
+    customScalarAdapters: CustomScalarAdapters = DEFAULT,
+    valueResolver: ValueResolver<M>
+): D {
```
sav007, line 170: should we keep just one API with default argument for `valueResolver`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
```diff
@@ -134,11 +137,43 @@ fun <D : Operation.Data> Operation<D>.parse(
...
   return parse(Buffer().writeUtf8(string), customScalarAdapters)
 }
+
+/**
+ * Parses GraphQL operation raw response from [map] with provided [customScalarAdapters] and returns result [Response]
+ *
+ * @param map: a [Map] representing the response. It typically include a "data" field
+ */
+fun <D : Operation.Data> Operation<D>.parse(
+    map: Map<String, Any?>,
```
sav007, line 155: just a thought if introducing type alias for this type argument in `MapResponseReader` will be more readable

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/MapResponseParser.kt
martinbonnin: I like having `Map` vs `Stream`. Buffer is more ambiguous as technically the `StreamResponseParser` works on a `BufferedSource`. At least `Map` is nowhere near `StreamResponseParser`.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
martinbonnin: The above parses a response, this one only parses `data` so we can't really merge them. `parseData` is needed because when reading from cache, we don't have a proper `Response`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
martinbonnin: We could but that would use another name and there's a finite about of them 😅 . We use `RecordSet` in some places but TBH I quite dislike it (as I always think of it as a Set<Record>, which it is not!). 

Since this map should be the result of `JsonReader.readRecursively()`, I was considering adding `JsonXYZ` aliases at some point. `Map<String, Any?>` would then be a JsonObject, or maybe a `ApolloJsonObject`, or just `AJsonObject`. But I'm not sure  how that would unfold yet.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ValueResolver.kt
```diff
@@ -2,6 +2,6 @@ package com.apollographql.apollo.api.internal
 
 import com.apollographql.apollo.api.ResponseField
 
-interface FieldValueResolver<R> {
-  fun <T> valueFor(recordSet: R, field: ResponseField): T?
+interface ValueResolver<in R> {
```
sav007, line 5: just curious why we renamed it? should it be more explicit `ResponseFieldValueResolver`?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.kt
```diff
@@ -25,12 +25,12 @@ import java.util.concurrent.CopyOnWriteArrayList
...
-class RealSubscriptionManager(customScalarAdapters: CustomScalarAdapters,
+class RealSubscriptionManager(private val customScalarAdapters: CustomScalarAdapters,
                               transportFactory: SubscriptionTransport.Factory,
-                              connectionParams: SubscriptionConnectionParamsProvider,
-                              dispatcher: Executor, connectionHeartbeatTimeoutMs: Long,
-                              responseNormalizer: Function0<ResponseNormalizer<Map<String, Any>>>,
-                              autoPersistSubscription: Boolean
+                              private val connectionParams: SubscriptionConnectionParamsProvider,
+                              private val dispatcher: Executor, private val connectionHeartbeatTimeoutMs: Long,
+                              private val responseNormalizer: Function0<ResponseNormalizer<Map<String, Any>>>,
```
sav007, line 32: should it be replaced with Kotlin functional type?

martinbonnin: Oh yes ! We have a bunch of them everywhere. I'll replace this one.

sav007 (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/ValueResolver.kt
martinbonnin: I think the rationale was something like:
- this is a small, narrow-scope class (there are 2 usages) so there's no need for a very long and descriptive type 
- a `Map` knows about `values`, not `ResponseFields`. So this class retrieves a "value" using the given "ResponseField". And not: this class retrieves a "ResponseFieldValue"

But TBH, given that it's a small API, I don't mind that much. We should rename `recordSet` to `map` though...

PR #2855, opened 2021-01-09
Fix object type singularization

sav007:
Run name singularization only for object types that derived from list fields.


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/Inflector.kt
```diff
@@ -8,7 +8,7 @@ import java.util.regex.Pattern
  * The singularization methods were heavily based off ruby on rails <link>https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflections.rb</link>
  */
 
-fun String.singularize(): String {
+internal fun String.singularize(): String {
```
martinbonnin, line 11: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/AstBuilder.kt
```diff
@@ -353,14 +347,13 @@ internal class AstBuilder private constructor(
...
     return buildObjectType(
-        name = responseName,
+        name = normalizedTypeName,
         description = schemaType.description,
         schemaTypename = this.type.rawType.name,
         fields = fields,
         fragments = fragments,
         targetPackageName = targetPackageName,
         abstract = abstract,
-
```
martinbonnin, line 363: Thanks for removing these, I think they came from a previous refactoring of mine.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -754,4 +756,15 @@ internal class BackendIrBuilder constructor(
...
   )
+
+  private val FrontendIr.Selection.Field.normalizedTypeName: String
+    get() {
+      val isListType = if (this.type is FrontendIr.Type.NonNull) {
+        this.type.ofType is FrontendIr.Type.List
+      } else {
+        this.type is FrontendIr.Type.List
+      }
+      val isObjectType = this.selections.isNotEmpty()
```
martinbonnin, line 767: Do we need to restrict to object types? If we have a list of scalar, we will never use this anyway?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -754,4 +756,15 @@ internal class BackendIrBuilder constructor(
...
+      return if (isListType && isObjectType) this.responseName.singularize().capitalize() else this.responseName
```
martinbonnin, line 768: Why `capitalize()` in one branch and not the other ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/AstBuilder.kt
```diff
@@ -730,6 +715,6 @@ internal class AstBuilder private constructor(
 
   private fun String.normalizeTypeName(): String {
```
martinbonnin, line 716: Nit: rename to `toLowerCamelCase()` ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
```diff
@@ -58,6 +58,7 @@ internal data class BackendIr(
       val name: String,
       val alias: String?,
       val type: IntrospectionSchema.TypeRef,
+      val normalizedTypeName: String,
```
martinbonnin, line 61: Nit: "normalized" always makes me think of "normalized" cache. Maybe we could find  another name there? I don't have a lot great ideas. Maybe `modelName` ? Where `Model` could refer to the generated class or interface? I think we don't have that too much in the codebase (yet :-P!).

martinbonnin:
Nice!!!! That'll make codegen way more predictable :-). I left a few comments, mainly around naming but it looks good!

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: This is a generic function used for any field as well as when we compose selection key. So we need  this logic to not touch list of scalar types otherwise the name of field and it's selection key will be off.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
sav007: will fix this.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: Do we care about selection key for scalar types? It's not like we're going to generate a model for them...

sav007: we care as we use it as indication to add override modifier. So if selection keys for this particular field has a  different key than we currently building for we add override modifier.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
sav007: how about just call it `typeName`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
martinbonnin: Got it, thanks !

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIr.kt
martinbonnin: Ahah naming is hard 🙃 . `typeName` is taken by Kotlin poet already so `modelName` is less " encumbered" but both work for me!

martinbonnin (approved):

martinbonnin:
There were some conflicts, I just rebased

PR #2856, opened 2021-01-09
Introduce generateFragmentImplementations compiler flag

sav007:
Hide generation fragment default implementation models behind `generateFragmentImplementations` compiler flag.


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -197,6 +197,15 @@ interface Service  {
   @ApolloExperimental
   val alwaysGenerateTypesMatching: SetProperty<String>
 
+  /**
+   * Whether or not generate default implementation classes for GraphQL fragments.
+   * Default value is `false`, means only interfaces are been generated.
+   *
+   * One of the use cases where generated implementation models are useful, is writing fragment response data directly to the Apollo store.
+   * To perform this operation you need fragment implementation class.
```
martinbonnin, line 205: Maybe we should highlight that it's not needed to use fragments?

```suggestion
   * Most the the times, fragment implementations are not needed because you can easily access fragments interfaces and read all data from your queries. They are needed if you want to be able to build fragments outside of an operation. For an exemple to programmatically build a fragment that is reused in another part of your code or to read and write fragments to the cache.
```

martinbonnin (approved):
👍

PR #2860, opened 2021-01-11
Optimize normalization

martinbonnin:
Introduce `NormalizationIRResponseWriter`, a ResponseWriter that writes a "normalization IR" later used to compute Records. 

This removes one intermediate `Map` and tree traversal an makes the `apolloParseAndNormalize` benchmark from ~30 to ~22 ms.

A nice side effect is that it fixes computing records for Json scalar which didn't work when normalizing models (it works in main when normalizing the network response, see https://github.com/apollographql/apollo-android/pull/2859)


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/AJ.kt
```diff
@@ -0,0 +1,45 @@
+package com.apollographql.apollo.cache.normalized.internal
+
+import com.apollographql.apollo.api.ResponseField
+
+/**
+ * A hierarchy of Json object with their fields attached used for normalization
+ */
+sealed class AJElement
```
sav007, line 8: I would recommend to change the name to smth different, as `AJ` is so abstract and not very descriptive for `augmented json`. Couple suggestion as this only related to cache normalization maybe `NormalizationNode/ Element`, `CacheNormalizationNode / Element`?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/AJResponseWriter.kt
```diff
@@ -0,0 +1,225 @@
...
+
+  private var rootObject: AJObject? = null
+  fun root(): AJObject {
+    if (rootObject == null) {
+      /**
+       * close the root object  if needed
+       */
+      check(objectStack.size == 1)
+      rootObject = AJObject(fields = objectStack.removeLast())
+    }
```
sav007, line 31: not sure but is it possible to use `lazy` intializer here?

martinbonnin: Yup, way more idiomatic

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/AJ.kt
```diff
@@ -0,0 +1,45 @@
...
+
+class AJScalar(val value: Any?) : AJElement()
+
+fun AJElement.mergeWith(other: AJElement): AJElement {
+  return when {
+    this is AJScalar && other is AJScalar -> {
+      check(this.value == other.value)
+      this
+    }
+    this is AJList && other is AJList -> throw UnsupportedOperationException()
```
sav007, line 23: hm why is that? We can't merge lists?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/AJ.kt
martinbonnin: Renamed to `NormalizationIR`

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/AJ.kt
martinbonnin: Not sure why I did that, I guess we need to merge lists nested in arrays indeed. I just added it.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/NormalizationIR.kt
```diff
@@ -0,0 +1,53 @@
+package com.apollographql.apollo.cache.normalized.internal
+
+import com.apollographql.apollo.api.ResponseField
+
+
+/**
+ * A hierarchy of Json object with their fields attached used for normalization
+ */
+internal object NormalizationIR {
+  sealed class Element {
```
sav007, line 10: up to you, just a thought if `Node` suit better?

sav007 (approved):

PR #2862, opened 2021-01-12
Drop OperationName wrapper

sav007:
Use raw string instead. We might consider inline class for this in future.


martinbonnin:
Yessss 👍

PR #2863, opened 2021-01-12
Add `operation.streamDataFromCache`

martinbonnin:
Introduce a `CacheJsonReader` that implements the `JsonReader` API so that we can use it from `StreamResponseReader`. This way we can read the cache without having to hold a `Record` in memory. 

This appears to be relatively faster but take this with a grain of salt given that `LruNormalizedCache` and `OptimisticCache` still use `Records`, I haven't run the full test suite yet.

Looking into this, I get the feeling that `LruNormalizedCache`, `OptimisticCache` and `SqlLiteNormalizedCache` serve pretty different purpose and it might be counterproductive to try to make them implement the same `NormalizedCache` interface that we have now. Will investigate this more in a follow up PR


apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -27,6 +30,19 @@ class SqlNormalizedCache internal constructor(
     return nextCache?.loadRecord(key, cacheHeaders)
   }
 
+  override fun stream(key: String, cacheHeaders: CacheHeaders): JsonReader? {
+    return try {
+      cacheQueries.recordForKey(key)
+          .executeAsList()
+          .firstOrNull()
+          ?.let {
+            BufferedSourceJsonReader(Buffer().writeUtf8(it.record))
```
martinbonnin, line 39: I'm not sure if this involves copying bytes or if the `Buffer` will automagically do the right thing. It might be worth investigating this.

PR #2867, opened 2021-01-12
[Cache] [3.x] Move normalization to the ApolloCacheInterceptor #2861

### LINKED ISSUE #2861: [Cache] [3.x] Normalization could be moved to the ApolloCacheInterceptor
The normalization is done today in the ApolloParseInterceptor (line 78), and thus executes while blocking the response callback to the UI.

Since normalization is pretty expensive, we could move the normalization to ApolloCacheInterceptor (line ~134) and benefit from the asynchronous dispatcher when passing `writeToCacheAsynchronously=true`.

olivierg13:
Fix #2861 

The normalization is done today in the ApolloParseInterceptor (line 78) and thus executes while blocking the response callback to the UI.

Since normalization is pretty expensive, moving it to ApolloCacheInterceptor (line ~134) and benefit from the asynchronous dispatcher when passing writeToCacheAsynchronously=true.

The only caveat with this approach is that we can't support the Response.dependentKeys API anymore.


apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.kt
```diff
@@ -93,6 +94,14 @@ interface ApolloCall<D : Operation.Data> : Cancelable {
      */
     abstract fun onResponse(response: Response<D>)
 
+    /**
+     * Gets called when the GraphQL response has been cached successfully.
+     * It should be called only once from the cache.
```
martinbonnin, line 99: Add a note why this is used? 

```
This is used internally by watchers to retrieve the list of dependentKeys they have to watch. 
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.kt
```diff
@@ -100,19 +103,23 @@ class ApolloCacheInterceptor(
...
-        result.add(record.toBuilder().mutationId(request.uniqueId).build())
-      }
-      result
+
+    val rawRecords = networkResponse.parsedResponse.get()?.data?.let {
+      (request.operation as Operation<Operation.Data>)
+          .normalize(it, customScalarAdapters, apolloStore.networkResponseNormalizer() as ResponseNormalizer<Map<String, Any>?>)
+    }
+    val mergeRecords = rawRecords?.map {
+      it.toBuilder().mutationId(request.uniqueId).build()
```
martinbonnin, line 112: Do everything in one step ? 

```
    val rawRecords = networkResponse.parsedResponse.get()?.data?.let {
      (request.operation as Operation<Operation.Data>)
          .normalize(it, customScalarAdapters, apolloStore.networkResponseNormalizer() as ResponseNormalizer<Map<String, Any>?>)
    }?.map { it.toBuilder().mutationId(request.uniqueId).build() }
```

Not that I mind having 2 steps but `mergeRecords` is confusing as "merge" usually refers to the process of merging several records together.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.kt
```diff
@@ -185,11 +192,4 @@ class ApolloCacheInterceptor(
       }
     }
   }
-
-  init {
-    this.apolloStore = apolloStore
-    this.dispatcher = dispatcher
-    this.logger = logger
-    this.writeToCacheAsynchronously = writeToCacheAsynchronously
```
martinbonnin, line 193: 👍

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.kt
```diff
@@ -299,10 +299,7 @@ class RealSubscriptionManager(private val customScalarAdapters: CustomScalarAdap
       val normalizer = responseNormalizer.invoke()
       val subscription = subscriptionRecord!!.subscription
       val response = try {
-        MapResponseParser.parse(message.payload, subscription, customScalarAdapters).let {
-          val records = it.data?.let { subscription.normalize(it, customScalarAdapters, normalizer as ResponseNormalizer<Map<String, Any>?>) }
-          it.copy(dependentKeys = records.dependentKeys())
-        }
+        MapResponseParser.parse(message.payload, subscription, customScalarAdapters)
```
martinbonnin, line 302: I think this breaks storing subscriptions results in the cache since `normalizer.records()` (used in notifyOnResponse below) was computed as a side effect of this `normalize()` call.

I wonder if we have tests for that. We should but for some reason, Github Actions does not trigger here...

martinbonnin:
> The only caveat with this approach is that we can't support the Response.dependentKeys API anymore.

I think this is perfectly fine to drop this as a public API, this is internal stuff.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.kt
olivierg13: Yeah I guessed, we had some of these as well when we first migrated :)

apollo-runtime/src/main/java/com/apollographql/apollo/internal/interceptor/ApolloCacheInterceptor.kt
olivierg13: Oh yeah I remember doing that cause the extension `dependentkeys()` was expecting a `Set<Record>`.
Let me check if it's not used anywhere else anymore, I can just refactor them to use a `List<Record>` instead, it would be simpler indeed.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/subscription/RealSubscriptionManager.kt
olivierg13: Ah good call let me re-introduce it there.

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloCall.kt
```diff
@@ -93,6 +94,15 @@ interface ApolloCall<D : Operation.Data> : Cancelable {
...
+     * This is used internally by watchers to retrieve the list of dependentKeys they have to watch.
+     *
+     * @param records the [List] of [Record] that was merged in cache.
+     */
+    open fun onCached(records: List<Record>) {}
```
martinbonnin, line 104: For the record (pun not intented 😁 ), we discussed using a `Set` here instead of a `List` as `records` should be unique. But since we don't need random access in this list and `Lists` are more economical than `Sets`, it make sense to keep a `List`

martinbonnin (approved):

apollo-integration/src/test/kotlin/com/apollographql/apollo/IntegrationTest.kt
```diff
@@ -307,7 +307,7 @@ class IntegrationTest {
   @Throws(Exception::class)
   fun operationParseResponseWithExtensions() {
     val source = Buffer().readFrom(javaClass.getResourceAsStream("/HeroNameResponse.json"))
-    val (_, _, _, _, _, extensions) = HeroNameQuery().parse(source)
+    val (_, _, _, _, extensions) = HeroNameQuery().parse(source)
```
martinbonnin, line 310: Nit: I don't know why we were destructuring before but it loooks like doing `val extensions = HeroNameQuery().parse(source).extensions` is a more robust way to get the extensions?

olivierg13: Yeah good point, for future refactors. Let me change that soon.

PR #2868, opened 2021-01-12
[Multiplatform] Add FetchPolicy

martinbonnin:
add multiplatform fetch policies to control the Cache Interceptor

```
enum class FetchPolicy {
  CACHE_ONLY,
  NETWORK_ONLY,
  CACHE_FIRST,
  NETWORK_FIRST,
  CACHE_AND_NETWORK
}
```

To specify a FetchPolicy, users have to build a `ApolloQueryRequest` using `ApolloQueryRequest.Builder`. We had a few incomprehensions in `2.x` about `ApolloQueryCall.responseFetcher` returning a new instance while some users expected it to mutate the existing one. Using the builder patter makes sure there's no ambiguity there. The drawback is that it's a bit more verbose but for simple cases, there are `ApolloClient.query(query: Query<D>)` overloads that will create the request under the hood. 

This also means that `ApolloRequest` is now public API so I removed `ApolloRequest.customScalarAdapters` since it's already passed to the client.


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.coroutines.flow.Flow
 
 interface ApolloCall<D: Operation.Data> {
   @ApolloExperimental
-  fun execute(executionContext: ExecutionContext = ExecutionContext.Empty): Flow<Response<D>>
+  fun execute(): Flow<Response<D>>
```
sav007, line 11: why this was removed?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/ApolloRequest.kt
```diff
@@ -4,13 +4,97 @@ import com.apollographql.apollo.api.ApolloExperimental
...
 import com.benasher44.uuid.uuid4
 
 @ApolloExperimental
-class ApolloRequest<D : Operation.Data>(
-    val operation: Operation<D>,
-    val customScalarAdapters: CustomScalarAdapters,
-    val executionContext: ExecutionContext
-) {
-  val requestUuid = uuid4()
+interface ApolloRequest<D : Operation.Data> {
```
sav007, line 14: curious if we should make it a sealed class?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -17,42 +17,86 @@ import com.apollographql.apollo.interceptor.ApolloResponse
...
+            proceed(request, chain).collect { emit(it) }
+          }
+        }
+        FetchPolicy.NETWORK_FIRST -> {
+          proceed(request, chain)
+              .catch {
+                val response = readFromCache(request, chain.customScalarAdapters)
+                if (response != null) {
+                  emit(response)
+                }
```
sav007, line 56: as far as I remember (might be wrong) but if we miss cache we deliver network error in old runtime implementation.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -17,42 +17,86 @@ import com.apollographql.apollo.interceptor.ApolloResponse
...
 
 @ApolloExperimental
 class ApolloCacheInterceptor<S>(private val store: S) : ApolloRequestInterceptor where S : WriteableStore, S : ReadableStore {
 
+  private fun <D : Operation.Data> proceed(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
+    return chain.proceed(request).map {
+      if (it.response.data != null) {
+        writeToCache(request, it.response.data!!, chain.customScalarAdapters)
+      }
+      it.setFromCache(false)
```
sav007, line 33: do we need this? I mean isn;t it set to false by default?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
```diff
@@ -62,10 +106,14 @@ class ApolloCacheInterceptor<S>(private val store: S) : ApolloRequestInterceptor
...
         RealCacheKeyBuilder()
     )
-    val data = operation.parseData(rootRecord, request.customScalarAdapters, fieldValueResolver)
+    val data = operation.parseData(rootRecord, customScalarAdapters, fieldValueResolver)
 
-    return builder<D>(operation)
-        .data(data)
-        .build()
+    return ApolloResponse(
+        response = builder<D>(operation)
```
sav007, line 112: not related to this PR, but does it make sense to have a builder for data classes in Kotlin? Data classes already kinda a builder with named arguments and default values, we probably should avoid using builders in new runtime?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/ApolloRequestInterceptor.kt
```diff
@@ -19,19 +21,22 @@ interface ApolloRequestInterceptor {
 }
 
 @ApolloExperimental
-internal class RealInterceptorChain private constructor(
+internal class RealInterceptorChain(
     private val interceptors: List<ApolloRequestInterceptor>,
-    private val index: Int
+    private val index: Int,
+    override val customScalarAdapters: CustomScalarAdapters,
```
sav007, line 27: is there a reason we move customScalarAdapters from request to interceptor chain? I'm not 100% sure if this is a right place, before we can easily access them from request if request is not right place should we store them in execution context then?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
martinbonnin: It wasn't used anywhere. I can add it back but if we do, I think we should clarify the precedence rules of ExecutionContext. Given that the API is loosely types, there's some footgun potential here. From my understanding, we have:

* Request ExecutionContext
* Response ExecutionContext
* Client ExecutionContext
* potentially Call ExecutionContext? (this one)

Should we document how each one is going to be used?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/ApolloRequest.kt
martinbonnin: Makes sense 👍

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Since ApolloResponse doesn't know anything about cache anymore, this data is in the execution context only known by the CacheInterceptor.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Good question. One advantage of `Builders` is that they "force" to use the return value. Take this exemple where we want to set the cache policy:

```
// Without builders, someone used to writing imperative/mutable code might be tempted to do that
val request = ApolloQueryRequest(HeroNameQuery())
request.fetchPolicy(NETWORK_ONLY)

apolloClient.query(request) // this doesn't use the NETWORK_ONLY policy because the return value isn't used above
```

Maybe it's just education but I feel like `Builders` are more boilerplate for us but ultimately better APIs for users

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/interceptor/ApolloRequestInterceptor.kt
martinbonnin: The idea was to make request a user-controlled type. The user would configure behaviour of the request there and create a call for it, much like OkHttp. Since `customScalarAdapters` are client-bound and not request-bound, this didn't feel like the right place. Right now it's reachable from the interceptor chain. Maybe ExecutionContext is better but I _think_ that would be less typed.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
martinbonnin: Good catch! I fixed that. Speaking about this, I think we might want to fine tune the behaviour a bit for some fetchers:

* CACHE_ONLY -> I think this should throw on error and not return an empty response
* For all fetchers that return several values, should we introduce a "generic" exception that wraps the actual ones? It feels arbitrary to return the last one or only the network one?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
sav007: I think as we are in active dev, everything can be changed as we don't have final stable API. So I guess we will document this later

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
sav007: As we discussed offline, getting value from cache only in my opinion shouldn't throw error but rather return empty response. The good analogy is Map, it doesn't throw error if there is no such key. Returning empty response with a flag that it comes from cache seems like a valid case.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo/interceptor/cache/ApolloCacheInterceptor.kt
sav007: hm, good point

martinbonnin:
I'm going to merge this one. Let's refine the ExecutionContext and error handling APIs as we go.

PR #2874, opened 2021-01-14
Multiplatform LRU cache

sav007:
Introduce multi platform (platform agnostic)  LRU cache implementation based on linked hash map for storing items and double linked list for tracking recently used items.


sav007:
Still in progress

PR #2878, opened 2021-01-14
Multiplatform LRU + Memory normalized cache

sav007:
Introduce multi platform (platform agnostic) LRU cache implementation based on linked hash map for storing items and double linked list for tracking recently used items.

Current implementation is thread safe, guaranteed by atomic fu lock that on JVM is transformed by compiler plugin into JVM ReentrantLock https://github.com/Kotlin/kotlinx.atomicfu#locks

This implementation is very close to what is shipped with Android https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/util/LruCache.java in terms of thread safety.

Unfortunately we can't introduce read / write lock here, as during reading item from the cache we must maintain recently used item property that requires moving it to the head of the LRU list, that requires delete and insert linked list operations to be atomic. 

We could look later at more performant implementation of LRU cache based on lock free implementation (for JVM), but I'm still not sure that hight throughput LRU implementation is required by mobile env, for sure that it's a different story if Apollo Client used on desktop or server side.

There are different ways how achieve this on JVM. For instance we can re-use what we have now Guava impl or we can build own based on this idea https://tech.ebayinc.com/engineering/high-throughput-thread-safe-lru-caching/

Introduce Memory cache based on LRU cache.


apollo-normalized-cache/build.gradle.kts
```diff
@@ -33,22 +33,19 @@ kotlin {
     }
 
     val jvmMain by getting {
-      dependsOn(commonMain)
```
martinbonnin, line 34: I never really understood why we did this. The Kotlin plugin does this automatically, right? Or did this have unwanted side effects?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
```diff
@@ -0,0 +1,99 @@
...
+      cachedEntry.record
+    }
+  }
+
+  override fun clearAll() {
+    lruCache.clear()
+    nextCache?.clearAll()
+  }
+
+  override fun remove(cacheKey: CacheKey, cascade: Boolean): Boolean {
```
martinbonnin, line 41: Not specifically linked to this PR but I think the `cascade` parameter should be removed from the `NormalizedCache` API and handled in a more global way. That'll make it easier to implement caches and avoid behavioral differences between caches.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/Platform.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo.cache.normalized.internal
+
+expect object Platform {
+  fun currentTimeMillis(): Long
```
martinbonnin, line 4: Should we make this a library ? :P 

Just kidding ;)

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
```diff
@@ -0,0 +1,99 @@
...
+        lruCache[apolloRecord.key] = CacheEntry(
+            record = oldRecord,
+            expireAfterMillis = expireAfterMillis
+        )
+      }
+    }
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  override fun dump() = buildMap<KClass<*>, Map<String, Record>> {
```
martinbonnin, line 73: Same here, I think this could be handled in a more global way but we can definitely do this better.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
```diff
@@ -0,0 +1,99 @@
+package com.apollographql.apollo.cache.normalized
+
+import com.apollographql.apollo.cache.ApolloCacheHeaders
+import com.apollographql.apollo.cache.CacheHeaders
+import com.apollographql.apollo.cache.normalized.internal.LruCache
+import com.apollographql.apollo.cache.normalized.internal.Platform
+import okio.internal.commonAsUtf8ToByteArray
+import kotlin.reflect.KClass
+
+class MemoryCache(
```
martinbonnin, line 10: Can you add a small KDoc header? I'm also saying this for myself, I don't want us to document each and every method call and parameter for a first drop of a new API that might change but adding a small description of the class sounds useful?

martinbonnin, line 10: Should we remove `MapNormalizedCache` and only keep one `MemoryNormalizedCache` ?

martinbonnin:
Looks good! 

Have you looked at what reentrant locks do for K/N? Since we'll most likely have to force accessing the cache from a single thread, does it make sense to wrap in a lock?

apollo-normalized-cache/build.gradle.kts
sav007: Yeah I guess it was just a legacy

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
sav007: Then we need to change the API as remove now needs to return store record instead of boolean.
Don't quite remember if there was particular reason of doing this on the low level end, we might have been thinking if we can run cascade deleting in a batch, we have to think about this carefully. Also current behaviour that we remove everything from current cache and only after that move to next chained one. With proposed changes we change that to remove one item on each cache before moving to nested referenced items

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/Platform.kt
sav007: Yeah let's do it, follow best practices from npm repo :)

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
sav007: Hm, not sure if it's possible on higher lever as we need to get this map anyway from the low level cache impl

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
sav007: Yeah I guess I forgot to do this.

sav007:
>Have you looked at what reentrant locks do for K/N? Since we'll most likely have to force accessing the cache from a single thread, does it make sense to wrap in a lock?

Yeah they handle that on low level (posix), in our case as we going to access this from one thread we should be ok.
https://github.com/Kotlin/kotlinx.atomicfu/blob/master/atomicfu/src/nativeMain/kotlin/kotlinx/atomicfu/locks/Synchronized.kt

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
martinbonnin: Yup, agreed

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
martinbonnin: Yea, I see it now, SQLite optimization require these extra APIs

martinbonnin (approved):

Issue #2882, opened 2021-01-18
Content-Type: "application/json; charset=utf-8" causes HTTP 422 Unprocessable Entity in graphql-java-spring


vikrama:
**Summary**
Interceptor sets **Content-Type: application/json; charset=utf-8** that causes SpringMVC based https://github.com/graphql-java/graphql-java-spring to throw a HTTP 422 Unprocessable Entity

**Version**
2.4.6

**Description**
Sending a mutation (POST) to a springmvc based graphql server returns a 422. Figured out that Content-Type is the issue in my case.

Between ApolloCall, ApolloServerInterceptor, ApolloHTTPNetworkTransport, OkHttp's RealCall and a bunch of interceptors there, if a mutation's RequestBody is set with a Content-Type, OkHttp uses that to set the Request Header. 

What I have tried to alleviate this problem:
* Tried setting ApolloMutationCall.RequestHeaders that overrides the Content-Type header to application/json. No luck
* Tried setting application interceptors and also network interceptors that fixes the header in a custom OkHttpClient for ApolloClient. No luck.

**Deprecation of charset=utf-8 in Content-Type header**
https://github.com/spring-projects/spring-framework/issues/22788
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html#APPLICATION_JSON_UTF8

As in the discussions & docs above, [RFC8259](https://tools.ietf.org/html/rfc8259) has made the charset subtype not necessary. All popular browsers have implemented the change. In apollo-android case, I'm not sure if OkHttp has already implemented the change. 

**Temporary fix that works**
Using a custom [GraphQLController](https://github.com/graphql-java/graphql-java-spring/blob/master/graphql-java-spring-webmvc/src/main/java/graphql/spring/web/servlet/components/GraphQLController.java#L48) in the server to handle an incoming Content-Type with charset seems to be the easier (and the only) workaround right now.

**Related bug reports**
https://github.com/apollographql/apollo-android/issues/1571

**Draft PR**
https://github.com/vikrama/apollo-android/pull/1/files

What do you think is the right step forward? If you think OkHttp is where I should be filing this bug against, let me know. Also, if there is any other way around the Apollo application interceptor, OkHttp application & network interceptor that don't seem to work for me, any help is appreciated.

martinbonnin:
Hi ! Thanks for the detailed issue. A few thoughts: 

[...]

PR #2883, opened 2021-01-18
Removing deprecated charset=utf-8 from Content-Type header

vikrama:
Addresses https://github.com/apollographql/apollo-android/issues/2882 . Keeping the string in api module as per Martin's suggestion.

### DIFF
--- a/apollo-http-cache/src/main/java/com/apollographql/apollo/cache/http/ResponseHeaderRecord.java
+++ b/apollo-http-cache/src/main/java/com/apollographql/apollo/cache/http/ResponseHeaderRecord.java
@@ -15,6 +15,7 @@
  */
 package com.apollographql.apollo.cache.http;
 
+import com.apollographql.apollo.api.network.ApolloMediaType;
 import com.apollographql.apollo.cache.http.internal.StatusLine;
 import java.io.IOException;
 import java.security.cert.Certificate;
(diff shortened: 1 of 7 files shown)


apollo-http-cache/src/main/java/com/apollographql/apollo/cache/http/ResponseHeaderRecord.java
```diff
@@ -272,7 +273,7 @@ private void writeCertList(BufferedSink sink, List<Certificate> certificates)
   Response response() {
     RequestBody body = null;
     if (HttpMethod.permitsRequestBody(requestMethod)) {
-      body = RequestBody.create(MediaType.parse("application/json; charset=utf-8"), "");
+      body = RequestBody.create(MediaType.parse(ApolloMediaType.APPLICATION_JSON_UTF8), "");
```
martinbonnin, line 276: Any reason to keep the charset here?

vikrama: Initial thought was that it won't affect the content-type header, but realized OkHttp uses the content type set in the body. Have removed the charset. Thanks.

martinbonnin:
@vikrama Thanks for the contribution 🤗 

I moved everything to the `api.internal` package, I don't think this needs to be seen from consumers. We can always add it later if needed.

vikrama:
Thx @martinbonnin for the review.

Not urgent, but trying to understand why my last build failed. Seems unrelated to my changes. Will help me be better informed while sending PRs in the future.

martinbonnin:
>  trying to understand why my last build failed. Seems unrelated to my changes. Will help me be better informed while sending PRs in the future.

Are you refering about this build failure? https://github.com/apollographql/apollo-android/runs/1724495722

This is because we track changes to the public API with metalava https://android.googlesource.com/platform/tools/metalava/

This we we're notified any time the public API breaks. For this to work, we also have to track we new additions are made like is the case here.

To fix the build, you can run `./gradlew metalavaGenerateSignature`

PR #2894, opened 2021-01-22
[3.x] Store more type information in ResponseField

martinbonnin:
Store more type information in `ResponseField`, this is the first step towards a cache reader that can batch SQL reads. 

Next step is to build the tree of `ResponseField`s


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
```diff
@@ -45,73 +44,21 @@ open class ResponseField internal constructor(
...
+     * a Named GraphQL type
+     *
+     * @param kind: whether this is an object or not. This is currently required by the CacheKeyResolver API
+     * In a typical server scenario, the resolvers would have access to the schema and would look up the complete typ
+     * but we want to stay lightweight so for now we add this information
+     */
+    class Named(val name: String, val kind: Kind): Type()
   }
 
+  enum class Kind { OBJECT, OTHER }
```
sav007, line 61: nit: Are we expecting to have more than 2 enums ? Should we then just remove it and break sealed class `Named` for 2 classes?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
```diff
@@ -140,236 +87,16 @@ open class ResponseField internal constructor(
...
     fun isArgumentValueVariableType(objectMap: Map<String, Any?>): Boolean {
       return (objectMap.containsKey(VARIABLE_IDENTIFIER_KEY)
           && objectMap[VARIABLE_IDENTIFIER_KEY] == VARIABLE_IDENTIFIER_VALUE && objectMap.containsKey(VARIABLE_NAME_KEY))
     }
+
+    fun Type.customScalarName(): String = when(this) {
+      is Type.NotNull -> ofType.customScalarName()
+      is Type.Named -> name
+      else -> error("Type '$this' is not a scalar type, check codegen")
+    }
```
sav007, line 100: I'm a little confused what this function does. It returns just a raw type `name` right? Should we call it smth like `leafName` / `rawName`

apollo-api/src/commonTest/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReaderTest.kt
```diff
@@ -1,453 +0,0 @@
...
-import com.apollographql.apollo.api.internal.json.JsonDataException
-import com.apollographql.apollo.api.internal.json.JsonUtf8Writer
-import com.apollographql.apollo.api.internal.json.Utils
-import okio.Buffer
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.assertTrue
-import kotlin.test.fail
-
-class SimpleResponseReaderTest {
```
sav007, line 20: Is this test case not relevant any more?

apollo-compiler/src/test/graphql/com/example/antlr_tokens/adapter/TestQuery_ResponseAdapter.kt
```diff
@@ -19,7 +19,13 @@ import kotlin.Suppress
...
 object TestQuery_ResponseAdapter : ResponseAdapter<TestQuery.Data> {
   private val RESPONSE_FIELDS: Array<ResponseField> = arrayOf(
-    ResponseField.forObject("typeWithGraphQLKeywords", "typeWithGraphQLKeywords", null, true, null)
+    ResponseField(
+      type = ResponseField.Type.Named("TypeWithGraphQLKeywords", ResponseField.Kind.OBJECT),
+      responseName = "typeWithGraphQLKeywords",
+      fieldName = "typeWithGraphQLKeywords",
+      arguments = emptyMap(),
+      conditions = emptyList(),
+    )
```
sav007, line 28: great to see generated named arguments!

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheValueResolver.kt
```diff
@@ -21,13 +19,24 @@ class CacheValueResolver(
...
+        if (type.kind == ResponseField.Kind.OBJECT) {
+          valueForObject(map, field) as T?
+        } else {
+          fieldValue(record = map, field)
+        }
+      }
     }
   }
 
+
```
sav007, line 39: nit: extra blank line

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
martinbonnin: Yep, I was wondering about this as well. I'm not 100% sure where this cache resolvers logic will take us. If we take the backend analogy, a server would have access to the schema and would retrieve the actual GraphQL type information from the name only. 
We still want this to be lightweight so it makes sense to embed some schema information directly inside the response field. It just feels weird to do it only for object types. I'm thinking that in the long term, we might want `fromFieldArguments()` to be applicable to something else than just object types and we might need more information stored there.
Also, `NotNull`/`List`/`Named` are taken directly from the GraphQL spec so it maps easily. 

Maybe we could do something like :

```
  sealed class Type {
    class NotNull(val ofType: Type): Type()
    class List(val ofType: Type): Type()

    sealed class Named(val name: String): Type() {
      class Object(name: String): Named(name)
      class Other(name: String): Named(name)
    }
  }
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
martinbonnin: 👍

apollo-api/src/commonTest/kotlin/com/apollographql/apollo/api/internal/SimpleResponseReaderTest.kt
martinbonnin: Once we handle cache on a separate path, I'm planning to simplify the `ResponseReader` API to be much closer to a `JsonReader` Api. It would then not work on `ResponseFields` anymore and these tests wouldn't be relevant anymore. We're not there yet though so I'll bring them back for now.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
sav007: Yeah I think proposed ^^^ sealed class works great

sav007 (approved):
Other than sealed class for `ResponseField.Type` lgtm.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
martinbonnin: Ooops sorry, I wasn't sure sealed classes could contain other sealed classes, I didn't realize this is what you proposed in the first place. This is done.

PR #2895, opened 2021-01-22
[3.x] Batch read the SQL cache

martinbonnin:
Work in progress on batching the SQL reads. The interesting part is in CacheMapBuilder.kt. 

~Doesn't work at the moment for queries with fragments on the query type and most likely other things too.~

Edit: it should be ok now.

```
benchmark:    69,688,705 ns Benchmark.apolloBatchCache
benchmark:   229,859,554 ns Benchmark.apolloReadCache
```

🎉


benchmark/src/androidTest/java/com/apollographql/apollo/benchmark/Benchmark.kt
```diff
@@ -72,6 +71,8 @@ class Benchmark {
     val data = operation.parse(bufferedSource()).data!!
 
     val records = operation.normalize(data, CustomScalarAdapters.DEFAULT, CacheKeyResolver.DEFAULT)
+
+    val cache = apolloClient.apolloStore.normalizedCache()
```
olivierg13, line 75: (not directly related to this PR) Would it make sense in the near future to have some MemoryCache benchmarks as well? Just to catch any regressions/improvements 😃

martinbonnin: Yep, definitely, see https://github.com/apollographql/apollo-android/pull/2909

I was more expecting multi-threading/contention to be the main issue on memory caches but you're right, maybe the batching introduces a cost there (for an exemple, it could allocate slightly more for all the pending references). I'll add that.

martinbonnin:
```
benchmark:     4,657,449 ns Benchmark.apolloReadCacheMemory
benchmark:     7,913,594 ns Benchmark.apolloBatchCacheMemory
benchmark:   232,572,731 ns Benchmark.apolloReadCacheSql
benchmark:    69,434,069 ns Benchmark.apolloBatchCacheSql
```

Batching slows down memory reading by almost a factor ~x2 and speeds up SQL by ~x3...

olivierg13: Yeah good point 👍 

In our Falcor batching implementation, we definitely saw a small increase in memory used, but it was alleviated by the speed bump from SQL queries batching. Not sure about Memory Batching (even if we do it, it's invisible to the UIs and I don't think we explicitly measured it 🤔)

olivierg13: I would say the raw loss in MemoryCache (4ms) is small compared to the big raw gain in SQL (160ms)?

Interestingly, in the field I've seen the `LRUMemoryCache`(relying on nytimes dependencies) response time more in the order of 100ms, do you think there were incremental improvements with the new `MemoryCache`? Or the 100ms vs 8ms could just be from multi-threading and contention like you mentioned.

martinbonnin: Oh wow, ~100ms to read from a memory cache sound a lot... If you ever get a chance to log a systrace, that'd be interesting to look at. 

I think it make sense to not batch if it's only memory and always batch if it's only SQL. Now if there is memory + SQL then I'm not sure. I'd say the 160ms are more important indeed but maybe it depends the use cases...

olivierg13: To be fair this 100ms is the time between a query is launched and when ApolloClient answers with Operation.Data (with thread switching etc...), so it's pretty good!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
```diff
@@ -17,7 +17,8 @@ class ResponseField(
     val responseName: String,
     val fieldName: String,
     val arguments: Map<String, Any?>,
-    val conditions: List<Condition>
+    val conditions: List<Condition>,
+    val possibleFieldSets: Map<String, Array<ResponseField>>,
```
sav007, line 21: just to clarify: this called `possible*` because of directives and fragments?

apollo-compiler/src/test/graphql/com/example/antlr_tokens/TestQuery.kt
```diff
@@ -50,6 +52,9 @@ data class TestQuery(
   override fun name(): String = OPERATION_NAME
 
   override fun adapter(): ResponseAdapter<Data> = TestQuery_ResponseAdapter
+  override fun responseFields(): Map<String, Array<ResponseField>> = mapOf(
+    "" to TestQuery_ResponseAdapter.RESPONSE_FIELDS
+  )
```
sav007, line 57: should we move this function into adapter? And we can get access to it via `Query.adapter().responseFeilds()`?

sav007: Also this mapping is a little confusing `"" to TestQuery_ResponseAdapter.RESPONSE_FIELDS`.

I'm thinking if we can provide some abstraction for it. If I understood this right the key here is typename and the value is root field. What do you think if we introduce this abstraction:
```
sealed class ResponseFieldRoot
  class Field(val responseField: ResponseField)
  class Fragment(val possibleSubRoots: Map<String, ResponseField>)
```

Naming could be better just throwing an idea.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
```diff
@@ -0,0 +1,116 @@
+package com.apollographql.apollo.cache.normalized.internal
```
sav007, line 1: What do you think of moving this class into sub package `com.apollographql.apollo.cache.normalized.internal.batching/batch/batchreader`?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+
+import com.apollographql.apollo.api.Operation
+import com.apollographql.apollo.api.ResponseField
+import com.apollographql.apollo.cache.CacheHeaders
+import com.apollographql.apollo.cache.normalized.CacheKey
+import com.apollographql.apollo.cache.normalized.CacheKeyResolver
+import com.apollographql.apollo.cache.normalized.CacheReference
+
+class CacheMapBuilder(
```
sav007, line 10: Also curious if different name like `CacheBatchReader` / `CacheBufferBuilder` / `CacheBatchResolver` / `CacheRecordBatchReader` will be more descriptive the intention of this class?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/OperationCacheExtensions.kt
```diff
@@ -93,6 +94,40 @@ fun <D : Fragment.Data> Fragment<D>.readDataFromCache(
   }
 }
 
+fun <D : Operation.Data> Operation<D>.batchDataFromCache(
```
sav007, line 97: should we be consistent here with name as rest API `.readBatchDataFromCahe`?
Also should we provide docs explaining what the purpose of this API?

sav007:
Is batch reading an optional feature? My concern that we can read more data than required and what should expect in worst case scenario? Is it possible that we will load the whole cache into memory?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+  private val cacheKeyBuilder = RealCacheKeyBuilder()
+
+  class PendingReference(
+      val key: String,
+      val possibleResponseFields: Map<String, Array<ResponseField>>
+  )
+
+  private val fieldSets = mutableMapOf<String, Map<String, Any?>>()
+
+  private val pendingReferences = mutableListOf<PendingReference>()
```
sav007, line 27: nit: I guess it can be local variable in `toMap()` function

sav007: I guess the same applicable for fieldSets

apollo-compiler/src/test/graphql/com/example/hero_details/adapter/HeroDetails_ResponseAdapter.kt
```diff
@@ -115,20 +123,24 @@ object HeroDetails_ResponseAdapter : ResponseAdapter<HeroDetails.Data> {
...
         ),
         ResponseField(
           type = ResponseField.Type.List(ResponseField.Type.Named.Object("FriendsEdge")),
           responseName = "edges",
           fieldName = "edges",
           arguments = emptyMap(),
           conditions = emptyList(),
+          possibleFieldSets = mapOf(
+            "" to Edge.RESPONSE_FIELDS
+          ),
```
sav007, line 143: have you considered to move this inside  `ResponseField.Type.Named.Object`? As only response field whose type is object can have selection set?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt
martinbonnin: I can drop the `possible*`, it was mainly because of different type conditions but this just calling that `fieldSets` is simpler.

apollo-compiler/src/test/graphql/com/example/antlr_tokens/TestQuery.kt
martinbonnin: >should we move this function into adapter? And we can get access to it via Query.adapter().responseFeilds()

My long term goal is actually to remove this from the adapter. Most of the contents of `ResponseField` is only used to read from the cache. For users without a cache, we can make this optional and save a bunch of bytecode and class init time.

martinbonnin: > the key here is typename 

Yes

> and the value is root field.

It's the list of fields that this typename will have. I like to think of this as a `FieldSet` because there's no fragment involved there.

We could introduce something like:

```
class FieldSet(val typename: String, val responseFields: Array<ResponseField>)
```

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
martinbonnin: Sounds good to me!

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
martinbonnin: I'd like to keep `Buffer` for okio `Buffers` and trying to limit the use of `Resolver` as it's heavily used in the GraphQL ecosystem and I'm not sure we would use it like other frameworks would so that could be surprising. 

Between `CacheBatchReader` and `CacheMapBuilder`, it depends if we want to highlight the process (batch reading) or the result (a Map). Both work for me.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/OperationCacheExtensions.kt
martinbonnin: Maybe we need a parameter ? 

```
fun <D : Operation.Data> Operation<D>.readDataFromCache(
    readableStore: ReadableStore,
    cacheKeyResolver: CacheKeyResolver,
    cacheHeaders: CacheHeaders,
    mode: ReadMode
): D?
```

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
martinbonnin: > I guess it can be local variable in toMap() function

And passing it as parameter down the call chain ? That might be a bit slower but allow to reuse the `CacheMapBuilder`. On the other hand, we're only using it once every time 🤷

apollo-compiler/src/test/graphql/com/example/hero_details/adapter/HeroDetails_ResponseAdapter.kt
martinbonnin: Yup, will do 👍

apollo-compiler/src/test/graphql/com/example/antlr_tokens/TestQuery.kt
sav007: Yeah I think that would be better.

Also maybe make it like 
```
class FieldSet(
  val responseFields: Array<ResponseField>,
  val typeCondition: String? = null
}
```
so this will indicate that this set of fields guarded by `typeCondition` if it present. 

What do you think?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
sav007: I think we should high light this so we won't  be confused what this class for. `CacheBatchReader` probably should be ok.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/OperationCacheExtensions.kt
sav007: Yeah should work 👍

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/CacheMapBuilder.kt
sav007: Yeah not a big deal TBH, just to avoid global (to class) mutable state. But.... it's really a nit, up to you.

sav007 (approved):
Other than comments LGTM.

apollo-compiler/src/test/graphql/com/example/antlr_tokens/TestQuery.kt
martinbonnin: Sounds good 👍

apollo-compiler/src/test/graphql/com/example/hero_details/adapter/HeroDetails_ResponseAdapter.kt
martinbonnin: I didn't do that part yet as I remembered I put this comment: https://github.com/apollographql/apollo-android/blob/61a3372f6d25106e9011efaa700d3ec45ed6c36e/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseField.kt#L56

The object vs non-object distinction feels a bit artificial. If we can remove `ResponseField.Type.Named.Object` ultimately, that'd be nice. Not sure we can but I'd like to explore that path a bit more before committing to the distinction

martinbonnin:
Everything should be in except https://github.com/apollographql/apollo-android/pull/2895#discussion_r573146190

I also took the opportunity to add `responseField` on Fragments.

PR #2896, opened 2021-01-25
Extract RecordsetDataSource + platform DB executors

sav007:
This is a preparation before migration NormalizedCache to coroutines.

Extract any DB related logic to RecordsetDataSource, later calling any of this API will be wrapped into appropriate platform executor.

Prepare DatabaseRequestExecutor for each platform:
- for JVM execution will be wrapped into IO coroutine
- for iOS execution will be made on background thread via Worker API.


apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/NormalizedCache.kt
```diff
@@ -72,18 +71,18 @@ abstract class NormalizedCache {
    * Calls through to [NormalizedCache.merge]. Implementations should override this method
    * if the underlying storage technology can offer an optimized manner to store multiple records.
    *
-   * @param recordCollection The collection of Records to merge.
+   * @param records The collection of Records to merge.
```
martinbonnin, line 74: +1 for `recordCollection`.

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -1,153 +1,88 @@
...
+import com.apollographql.apollo.cache.normalized.sql.internal.RecordsetDataSource.selectAllRecords
+import com.apollographql.apollo.cache.normalized.sql.internal.RecordsetDataSource.selectRecord
+import com.apollographql.apollo.cache.normalized.sql.internal.RecordsetDataSource.selectRecords
+import com.apollographql.apollo.cache.normalized.sql.internal.RecordsetDataSource.updateRecords
 import kotlin.reflect.KClass
 
 class SqlNormalizedCache internal constructor(
-    private val recordFieldAdapter: RecordFieldJsonAdapter,
     private val database: ApolloDatabase,
-    private val cacheQueries: CacheQueries
```
martinbonnin, line 18: Should we keep the `cacheQueries` parameter instead of `database` ? Looks like we don't use anything from `database` downstream.

apollo-normalized-cache-sqlite/src/iosMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/DatabaseRequestExecutor.kt
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Performs database operation on background thread via [Worker] API.
+ * This executor requires caller thread to be main.
+ * After database operation execution it resumes on main thread.
+ */
+internal actual class DatabaseRequestExecutor actual constructor(
+    private val database: ApolloDatabase
+) {
+  private val worker: Worker = Worker.start()
```
martinbonnin, line 26: Any pro/cons of using a custom queue using `dispatch_queue_create()` instead of a `Watcher` and use the same mecanism you used below to dispatch on the main thread ? My current understanding is that the Worker API is also available on other platforms like linux/windows but since we're in Apple world here, dispatch queue will always be available and might be a more platform idiomatic way to do?

apollo-normalized-cache-sqlite/src/iosMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/DatabaseRequestExecutor.kt
```diff
@@ -0,0 +1,83 @@
...
+      val database: ApolloDatabase,
+      val operation: ApolloDatabase.() -> R,
+      val continuation: Continuation<R>
+  ) {
+    fun execute() {
+      assert(NSThread.isMainThread())
+      val continuationPtr = StableRef.create(continuation).asCPointer()
+      worker.execute(
+          mode = TransferMode.SAFE,
+          producer = { (continuationPtr to { operation(database) }).freeze() },
```
martinbonnin, line 50: If I understand correctly, this will freeze the `ApolloDatabase` ? If that's the case, I'm amazed SQLDelight makes everything immutable 👍 . That subgraph must be quite a beast.

martinbonnin (approved):
A few questions but nothing blocking, LGTM!

apollo-normalized-cache-sqlite/src/iosMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/DatabaseRequestExecutor.kt
sav007: Yeah, I think it's ok to freeze DB

apollo-normalized-cache-sqlite/src/iosMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/DatabaseRequestExecutor.kt
sav007: Yeah we can do that, I think Worker is thread confined and execution of all tasks is serialized. Also if we use CGD means if it spins a new thread we need to initialize Kotlin Runtime on it. Not sure how bad it affects perf

martinbonnin: Good point. Since it's working already, let's keep it like this 👍

apollo-normalized-cache-sqlite/src/iosMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/DatabaseRequestExecutor.kt
martinbonnin: ❄️ ⛄ ❄️ ⛄ ❄️ ⛄

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.kt
```diff
@@ -56,37 +36,11 @@ class RecordFieldJsonAdapter {
...
-        }
-        else -> error("Unsupported record value type: ${value::class.qualifiedName}")
-      }
-    }
+  @Suppress("UNCHECKED_CAST")
+  fun fromJson(jsonFieldSource: String): Map<String, Any?>? {
+    val buffer = Buffer().write(jsonFieldSource.encodeUtf8())
+    return BufferedSourceJsonReader(buffer)
+        .readRecursively()
+        ?.deserializeCacheReferences() as Map<String, Any?>?
```
martinbonnin, line 44: 👍

apollo-normalized-cache-sqlite/src/androidMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/IOTaskExecutor.kt
```diff
@@ -0,0 +1,16 @@
...
+import kotlinx.coroutines.withContext
+
+/**
+ * Performs IO operation on background thread via [Dispatchers.IO].
+ */
+internal actual class IOTaskExecutor {
+  actual suspend fun <R> execute(operation: () -> R): R {
+    return withContext(Dispatchers.IO) {
+      operation()
+    }
```
martinbonnin, line 14: Not for myself: duplicating this in androidMain and jvmMain feels bad. I think this is something solved by HMPP once I get a chance to look into it.

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.kt
```diff
@@ -56,37 +36,11 @@ class RecordFieldJsonAdapter {
...
-
-    @Suppress("UNCHECKED_CAST")
-    @Throws(IOException::class)
-    private fun writeJsonValue(value: Any?, jsonWriter: JsonWriter) {
-      when (value) {
-        null -> jsonWriter.nullValue()
-        is String -> jsonWriter.value(value)
-        is Boolean -> jsonWriter.value(value)
-        is Number -> jsonWriter.value(value)
-        is CacheReference -> jsonWriter.value(value.serialize())
```
martinbonnin, line 74: Wait. Where did that logic go ? I think we still need to handle CacheReference serialization here?

apollo-normalized-cache-sqlite/src/androidMain/kotlin/com/apollographql/apollo/cache/normalized/sql/internal/IOTaskExecutor.kt
sav007: yeah, it's annoying

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/RecordFieldJsonAdapter.kt
sav007: agh......jeeez my bad

PR #2898, opened 2021-01-26
Add redirect for tutorial

designatednerd:
Added a redirect so that when you go to [https://www.apollographql.com/docs/android/tutorial](https://www.apollographql.com/docs/android/tutorial/), the first page of the tutorial loads (rather than the 404 you get at the moment). This also helps us avoid giving out hard links to the first page of the tutorial so that if that URL ever changes in the future it can get redirected easily.


designatednerd:
You can confirm via the deploy preview that the redirect works: https://deploy-preview-2898--apollo-android-docs.netlify.app/docs/android/tutorial

PR #2906, opened 2021-02-01
properly forward antlr syntax errors

martinbonnin:
The current pipeline is antlr -> GQL -> FrontendIR -> ...

The GQL transformation assumes a syntactically correct document and will throw null pointer exceptions if not. Rather than throwing an obscure null pointer exception, stop the processing and output parsing issues.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/frontend/GraphQLParser.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/frontend/GraphQLParser.kt
@@ -50,9 +50,6 @@ object GraphQLParser {
 
   /**
    * Parses and validates the given SDL executable document, containing operations and/or fragments
-   *
-   * This voluntarily does not return a GQLDocument in order to explicitly distinguish between operations and schemas
-   * throws if the schema has errors or is not a schema file
    */
   fun parseOperations(source: BufferedSource, filePath: String? = null, schema: Schema): ParseResult<GQLDocument> {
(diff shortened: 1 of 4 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/frontend/GraphQLParser.kt
```diff
@@ -76,12 +73,12 @@ object GraphQLParser {
    * A specialized version that works on multiple files and can also inject fragments from another
    * compilation unit
    */
-  fun parseExecutableFiles(files: Set<File>, schema: Schema, injectedFragmentDefinitions: List<GQLFragmentDefinition>): ParseResult<List<GQLDocument>> {
+  fun parseExecutableFiles(files: Set<File>, schema: Schema, injectedFragmentDefinitions: List<GQLFragmentDefinition>): Pair<List<GQLDocument>, List<Issue>> {
```
sav007, line 76: why `ParseResult ` can't be used here?

martinbonnin: Since `parseExecutableFiles` returns a list, it will never be null. It'll be empty instead, which is conveyed by the `Pair` type. This also signals that this API behaves inconsistently from the others in the file. i.e. it does more things like checking duplicates and non-executable definitions.

This might mean that we should move this code somewhere else and keep this file 100% about the spec.

sav007 (approved):

PR #2909, opened 2021-02-02
Remove previous implementation of LRU cache in favour of MemoryCache

sav007:
- Replace LRUCache with new MemoryCache
- Remove external dependency to `com.nytimes.android:cache`
- Clean up `NormalizedCache` from inheritance hell


apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/NormalizedCache.kt
```diff
@@ -134,9 +79,7 @@ abstract class NormalizedCache {
     leafCache.nextCache = cache
   }
 
-  open fun dump(): Map<@JvmSuppressWildcards KClass<*>, Map<String, Record>> {
-    return mapOf(this::class to emptyMap())
-  }
+  abstract fun dump(): Map<@JvmSuppressWildcards KClass<*>, Map<String, Record>>
```
martinbonnin, line 82: 👍

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/NormalizedCache.kt
```diff
@@ -107,16 +62,6 @@ abstract class NormalizedCache {
...
 
-  /**
-   * Remove cached record by the key
-   *
-   * @param cacheKey of record to be removed
-   * @return `true` if record with such key was successfully removed, `false` otherwise
-   */
-  fun remove(cacheKey: CacheKey): Boolean {
-    return remove(cacheKey, false)
-  }
```
martinbonnin, line 118: 👍

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/MemoryCache.kt
```diff
@@ -50,40 +52,53 @@ class MemoryCache(
...
     }
+
+    return changedKeys + nextCache?.merge(record, cacheHeaders).orEmpty()
+  }
+
+  override fun merge(records: Collection<Record>, cacheHeaders: CacheHeaders): Set<String> {
+    if (cacheHeaders.hasHeader(ApolloCacheHeaders.DO_NOT_STORE)) {
+      return emptySet()
+    }
+    return records.flatMap { record -> merge(record, cacheHeaders) }.toSet()
```
martinbonnin, line 95: I was going to write above that maybe we should keep the common `merge(collection)` in the parent class but given the implementation is so small, it doesn't make sense indeed. 👍

martinbonnin:
How much of a burden would it be to keep the guava cache as a separate artifact? That could be useful if we want to do benchmarks at some point.

sav007:
Sure I was thinking if we want to keep it, I guess the only reason is to keep it for benchmarking.
I will revert removing old LRUCache so we can benchmark it later, but keep MemoryCache as default one.

martinbonnin (approved):

PR #2910, opened 2021-02-02
Create gradle.yml

chihero:


martinbonnin:
Hi @chihero 👋  Can you give more context about why we need this ? We already have a workflow file there: https://github.com/apollographql/apollo-android/blob/main/.github/workflows/ci.yml

PR #2913, opened 2021-02-03
Make Record to be immutable

sav007:
Previous implementation of Record allows to mutate its state. This could be an issue with KN as we might run into issues with mutating frozen object. Also immutable implementation is more robust.


apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/Record.kt
```diff
@@ -1,141 +1,65 @@
...
-import kotlin.jvm.JvmStatic
-import kotlin.jvm.Synchronized
-import kotlin.jvm.Volatile
 
 /**
  * A normalized entry that corresponds to a response object. Object fields are stored if they are a GraphQL Scalars. If
  * a field is a GraphQL Object a [CacheReference] will be stored instead.
  */
-class Record internal constructor(
+data class Record (
```
martinbonnin, line 10: Same question I always ask, sorry in advance 😅: why do we need the `data` there? It looks like we only need a `copy` that we can implement easily. That'll remove a bunch of bytecode and make evolving the API easier, cf https://jakewharton.com/public-api-challenges-in-kotlin/

martinbonnin:
This looks so much better 🤩 . I think we can drop the `data` class requirement. Besides that, all good 👍

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/Record.kt
sav007: First I thought we needed `copy` function but we don't.

martinbonnin (approved):

PR #2915, opened 2021-02-04
Fix broken build

sav007:


martinbonnin (approved):

martinbonnin:
Oh my sorry about that. These happen since we don't rebase PRs before merging so it can be that a "green" PR is actually red after another PR is merged. Given this happens infrequently, I think it's ok and not worth adding some additional friction.

I also pushed a fix for a remaining instance of `.copy`

apollo-compiler/src/test/kotlin/com/apollographql/apollo/compiler/ValidationTest.kt
```diff
@@ -17,9 +17,9 @@ import java.io.File
 class ValidationTest(name: String, private val graphQLFile: File) {
   private val separator = "\n------------\n"
 
-  private fun List<Issue>.serialize() = map {
+  private fun List<Issue>.serialize() = joinToString(separator) {
     "${it.severity}: ${it.javaClass.simpleName} (${it.sourceLocation.line}:${it.sourceLocation.position})\n${it.message}"
```
Repoman124, line 21: f47511fbbb0be8d4fc8a6656e535149d1ea0f5e4

Repoman124 (approved):

Repoman124 (commented):
I'm new to all this, I'm eager to learn. Any help is greatly appreciated

PR #2919, opened 2021-02-05
Fix issue with named fragment in inline fragment

sav007:


apollo-compiler/src/test/graphql/com/example/named_fragment_inside_inline_fragment/GetHero.kt
```diff
@@ -0,0 +1,114 @@
...
+        override val name: String,
+        /**
+         * The movies this character appears in
+         */
+        override val appearsIn: List<Episode?>
+      ) : Hero, Character, Character.Character, CharacterName, CharacterAppearsIn
+
+      data class OtherHero(
+        override val __typename: String
+      ) : Hero
```
martinbonnin, line 77: It bugs me that we are generating 2 interfaces and 2 data classes here when we know there is only one type condition. Could we generate just one data class ? That'll make the generated code easier to read and also faster to compile. And I think this doesn't lose any feature?

martinbonnin (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ir/BackendIrBuilder.kt
```diff
@@ -371,24 +382,41 @@ internal class BackendIrBuilder constructor(
...
+    val fragmentInterfaceRootSelectionKeys = fragmentInterfaces
+        .map { selectionKey + it.name }
+
     val fragmentImplementations = fragmentInterfaces
         .flattenFragments()
         .buildFragmentImplementations(
             parentName = this.typeName,
-            parentFields = this.fields,
+            // NOTE: required by new version that removes interfaces
+            parentFields = this.fields.addFieldSelectionKey(selectionKey),
```
sav007, line 393: TBH not sure why this makes a new version of codegen that removes interfaces work but it breaks previous version. Will investigate in next PR.

PR #2923, opened 2021-02-11
Simplify normalization

martinbonnin:
Rewrite normalization once again using the new `FieldSets` and the `ResponseField` tree. This removes the `NormalizationIR` intermediate structs, saves a few CPU cycles and will allow to make the `ResponseField` codegen opt-out for faster compile times and loading times.

Also simplified `NormalizerTest` to remove the `ApolloClient` dependency


sav007 (approved):
🎉

PR #2924, opened 2021-02-14
Uniformise ResponseAdapters

martinbonnin:
Remove `__typename: String ? = null` from `fromResponse` and use the `ResponseAdapters` for enums/custom scalars in addition to objects. This makes the API more consistent and also allow things like looking up the custom scalar adapters only once and use native number types instead of `BigDecimal`.


apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/OperationExtensions.kt
```diff
@@ -36,16 +35,22 @@ import kotlin.jvm.JvmOverloads
  * @param indent the indentation string to be repeated for each level of indentation in the encoded document. Must be a string
  * containing only whitespace. If [indent] is an empty String the encoded document will be compact. Otherwise the encoded
  * document will be more human-readable.
- * @param customScalarAdapters configured instance of custom GraphQL scalar type adapters. Default adapters are used if this
+ * @param responseAdapterCache configured instance of custom GraphQL scalar type adapters. Default adapters are used if this
```
sav007, line 38: nit: `configured instance of GraphQL operation response adapters cache. Default adapters are used if this`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Fragment.kt
```diff
@@ -10,7 +10,7 @@ interface Fragment<D: Fragment.Data> {
 
   fun variables(): Operation.Variables
 
-  fun adapter(): ResponseAdapter<D>
+  fun adapter(responseAdapterCache: ResponseAdapterCache): ResponseAdapter<D>
```
sav007, line 13: should it also take optional argument `responseAdapterCache: ResponseAdapterCache = ResponseAdapterCache.DEFAULT`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseAdapterCache.kt
```diff
@@ -1,12 +1,22 @@
...
 
+import com.apollographql.apollo.api.internal.ResponseAdapter
+import com.apollographql.apollo.api.internal.json.JsonReader
+import com.apollographql.apollo.api.internal.json.JsonWriter
+import com.apollographql.apollo.api.internal.json.Utils.readRecursively
+import com.apollographql.apollo.api.internal.json.Utils.writeToJson
+import kotlin.jvm.Synchronized
+
 /**
  * A wrapper around a Map of [CustomScalarAdapter] that allows to easily retrieve an adapter for the given [CustomScalar]
```
sav007, line 11: nit: I think KDoc should be updated?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseAdapterCache.kt
```diff
@@ -1,12 +1,22 @@
...
  */
-class CustomScalarAdapters(val customScalarAdapters: Map<CustomScalar, CustomScalarAdapter<*>>) {
+class ResponseAdapterCache(val customScalarAdapters: Map<CustomScalar, CustomScalarAdapter<*>>) {
 
   private val adapterByGraphQLName = customScalarAdapters.mapKeys { it.key.graphqlName }
 
+  private val adapterByQueryName = ThreadSafeMap<String, ResponseAdapter<*>>()
+  private val adapterByFragmentName = ThreadSafeMap<String, ResponseAdapter<*>>()
```
sav007, line 18: so if accessor APIs for this class is `synchronized` then we don't need `ThreadSafeMap`, right?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/MapJsonReader.kt
```diff
@@ -64,20 +69,27 @@ class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
     val map = nextValue()
     check(map is Map<*, *>)
     push(OrderedMap(map.entries.map { Entry(it.key as String, it.value) }))
+
+    nameIndexStackSize++
+    check(nameIndexStackSize < 33) {
```
sav007, line 74: where this comes from?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/StandardResponseAdapters.kt
```diff
@@ -0,0 +1,93 @@
...
+  override fun toResponse(writer: JsonWriter, value: T?) {
+    if (value == null) {
+      writer.nullValue()
+    } else {
+      wrappedAdapter.toResponse(writer, value)
+    }
+  }
+}
+
+object stringResponseAdapter: ResponseAdapter<String> {
```
sav007, line 45: why lower case?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/Utils.kt
```diff
@@ -62,9 +62,32 @@ object Utils {
...
+            number
+          }
+        }
+      }
 
-      JsonReader.Token.LONG -> BigDecimal(nextLong())
+      JsonReader.Token.LONG -> {
+        val number = nextString()!!
+        try {
+          number.toInt()
```
sav007, line 86: `number.toLong()`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/Utils.kt
```diff
@@ -62,9 +62,32 @@ object Utils {
...
+          try {
+            number.toDouble()
+          } catch (e: Exception) {
+            number
+          }
+        } else {
+          try {
+            number.toInt()
+          } catch (e: Exception) {
+            number
```
sav007, line 78: shouldn't we just fail here if it's not proper number?

apollo-api/src/iosMain/kotlin/com/apollographql/apollo/api/ThreadSafeMap.kt
```diff
@@ -0,0 +1,36 @@
...
+import platform.darwin.DISPATCH_QUEUE_SERIAL
+import platform.darwin.dispatch_queue_create
+import platform.darwin.dispatch_queue_t
+import platform.darwin.dispatch_sync
+import kotlin.native.concurrent.AtomicReference
+import kotlin.native.concurrent.freeze
+
+actual class ThreadSafeMap<K, V> {
+  private val stableRef = StableRef.create(mutableMapOf<K, V>())
+  private val queue = dispatch_queue_create(label = "map", attr = DISPATCH_QUEUE_SERIAL as dispatch_queue_t)
```
sav007, line 13: nit: lets label it as `ThreadSafeMap`

apollo-api/src/iosMain/kotlin/com/apollographql/apollo/api/ThreadSafeMap.kt
```diff
@@ -0,0 +1,36 @@
...
+    ) {
+      initRuntimeIfNeeded()
+
+      reference.value = stableRef.get().getOrPut(key, defaultValue).freeze()
+    }
+
+    val result = reference.value
+    reference.value = null
+    return result!!
+  }
```
sav007, line 29: sorry but I'm not sure I'm following what is going in here.
Map was created outside the queue, we de-refrence it in the queue, block caller thread of `getOrPut` and mutate it outside of queue.

Is intention to make `isolate` pattern here? When accessing map confined to specific thread?

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
```diff
@@ -61,6 +61,7 @@ internal data class CodeGenerationAst(
       val typeRef: TypeRef,
       val schemaTypename: String?,
       val fragmentAccessors: List<FragmentAccessor>,
+      val isTypeCase: Boolean,
```
sav007, line 64: could you pls remind me what is `isTypeCase `?

apollo-compiler/src/test/graphql/com/example/antlr_tokens/TestQuery.kt
```diff
@@ -51,7 +52,13 @@ data class TestQuery(
 
   override fun name(): String = OPERATION_NAME
 
-  override fun adapter(): ResponseAdapter<Data> = TestQuery_ResponseAdapter
+  override fun adapter(customScalarAdapters: ResponseAdapterCache): ResponseAdapter<Data> {
```
sav007, line 55: `customScalarAdapters ` --> `responseAdapterCache `

apollo-compiler/src/test/graphql/com/example/antlr_tokens/adapter/TestQuery_ResponseAdapter.kt
```diff
@@ -5,120 +5,130 @@
...
-      conditions = emptyList(),
-      fieldSets = listOf(
-        ResponseField.FieldSet(null, TypeWithGraphQLKeywords.RESPONSE_FIELDS)
-      ),
-    )
-  )
+class TestQuery_ResponseAdapter(
+  customScalarAdapters: ResponseAdapterCache
+) : ResponseAdapter<TestQuery.Data> {
+  val nullableTypeWithGraphQLKeywordsAdapter:
```
sav007, line 27: I think this should be private

apollo-compiler/src/test/graphql/com/example/antlr_tokens/adapter/TestQuery_ResponseAdapter.kt
```diff
@@ -5,120 +5,130 @@
...
-          when (selectField(RESPONSE_FIELDS)) {
-            0 -> on = readString(RESPONSE_FIELDS[0])
-            1 -> null_ = readString(RESPONSE_FIELDS[1])
-            2 -> alias = readString(RESPONSE_FIELDS[2])
-            else -> break
-          }
+    val RESPONSE_NAMES: List<String> = RESPONSE_FIELDS.map { it.responseName }
+  }
+
+  class TypeWithGraphQLKeywords(
```
sav007, line 67: maybe also all nested classed make private?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseAdapterCache.kt
martinbonnin: Gaaa, sorry I forgot to remove the `@Synchronized` from `ResponseAdatperCache`. We need `ThreadSafeMap` for native

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Fragment.kt
martinbonnin: Since this is mostly an internal API, I don't mind the mandatory parameter. I don't think many users will need it and those who need it definitely need to be aware of how what a `ResponseAdapterCache` is. 
Also this is bytecode that's going to end up in every Operation/Fragment. I usually like to put helpers method in extensions functions.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ResponseAdapterCache.kt
martinbonnin: Definitely 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/MapJsonReader.kt
martinbonnin: That's from Moshi: https://github.com/apollographql/apollo-android/blob/caf408b6cd0f0d39547675310b1aa4f1ccebecc2/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/BufferedSourceJsonReader.kt#L50

We don't support more than 32 levels of nesting

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/StandardResponseAdapters.kt
martinbonnin: No specific reason. Reading: https://kotlinlang.org/docs/coding-conventions.html#property-names, both upper and lower case seem to work.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/Utils.kt
martinbonnin: GraphQL has no `Long`, everything other than `Int` and `Double` will fallback to a `String` and require a custom adapter and type information to be parsed.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/json/Utils.kt
martinbonnin: See above, a custom BigDecimal scalar adapter might want to read the number as a string

apollo-api/src/iosMain/kotlin/com/apollographql/apollo/api/ThreadSafeMap.kt
martinbonnin: I'm not 100% sure either. I wanted to do push this before renaming everything to apollo3 but this'll have to be properly tested and most likely reworked.

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/backend/ast/Ast.kt
martinbonnin: It's a "case" for a "type" condition. I think of it as a response shape under a certain type condition. @martijnwalraven did I get it right?

apollo-compiler/src/test/graphql/com/example/antlr_tokens/adapter/TestQuery_ResponseAdapter.kt
martinbonnin: Yep, good idea. It doesn't quite work at the moment because `ResponseFields` "jump" class nesting:

```
class Data_Adapter {
  companion object RESPONSE_FIELDS = arrayOf(ResponseField(fieldSets = listOf(
    // Cannot access 'HumanCharacterHero': it is private in 'Hero'`
    FieldSet("Human", Hero.HumanCharacterHero.RESPONSE_FIELDS)
    FieldSet(null, Hero.OtherCharacterHero.RESPONSE_FIELDS)
  ))

    private class Hero {
       private class HumanCharacterHero {...}
       private class OtherCharacterHero {...}
    }  
}
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/internal/StandardResponseAdapters.kt
martinbonnin: IntelliJ says that should be uppercase too. I made everything uppercase

apollo-api/src/iosMain/kotlin/com/apollographql/apollo/api/ThreadSafeMap.kt
martinbonnin: It's working now with https://github.com/apollographql/apollo-android/pull/2924/commits/39d3ccd860f9b987c687510b36ae89d29ce211b9. The previous code was indeed completely broken 🙃 .

sav007 (approved):
Could you please create ticket for leftovers and we can merge it

martinbonnin:
Follow up ticket https://github.com/apollographql/apollo-android/issues/2926

PR #2937, opened 2021-02-18
[API] Make Input safer

martinbonnin:
make Input more kotlin idiomatic
    
Especially, this forbid the invalid state that could happen before with
    
```
    Input<NonNullableType>.fromNullable(null)
```


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Input.kt
```diff
@@ -1,64 +1,27 @@
...
-    @JvmField val defined: Boolean
-) {
-
-  override fun equals(other: Any?): Boolean {
-    if (this === other) return true
-    if (other !is Input<*>) return false
-
-    if (value != other.value) return false
-    if (defined != other.defined) return false
+class Input<out V>(private val value: Any?) {
```
sav007, line 11: Should we make it more idiomatic and represent it as sealed class?

martinbonnin: Yup, I took inspiration from the kotlin [Result](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/util/Result.kt) class but sealed class make better use of type inference, etc.. I just pushed the change

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Input.kt
```diff
@@ -1,64 +1,13 @@
...
-     * Creates a new [Input] instance that is always undefined.
-     */
-    @JvmStatic
-    fun <V> absent(): Input<V> {
-      return Input(null, false)
-    }
-  }
+sealed class Input<out V> {
+  class Present<V>(val value: V): Input<V>()
+  class Absent<V>: Input<V>()
```
sav007, line 11: Wonder if we can do it even like this:
```
sealed class Input<out V> {
  class Present<V>(val value: V): Input<V>()
  object Absent: Input<Nothing>()
```

So `Absent` will be a singleton?

sav007 (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Input.kt
martinbonnin: Oooo, this is nice, Kotlin never ceases to impress 🤩 
I just pushed the change

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Input.kt
martinbonnin: I just bumped into that: https://github.com/Kotlin/KEEP/blob/1827a4d3da410f70797522a156405dcbd3b1069a/proposals/stdlib/result.md#representing-as-a-sealed-class

The reason `Result` is not a sealed class in the stdlib is performance since they want the values to be stored unboxed. I don't think that changes anything in our case since this is not going to be used in any "hot" path, it's fine to pay the boxing price I think.

PR #2940, opened 2021-02-19
[apollo-api] Introduce Variable

martinbonnin:
Introduce a new `VariableValue` wrapper class to represent [GraphQL Variables](https://spec.graphql.org/draft/#Variable). This gives more typesafety during codegen and also when computing the cacheKeys.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/VariableValue.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/VariableValue.kt
@@ -1,0 +1,3 @@
+package com.apollographql.apollo3.api
+
+class VariableValue(val name: String)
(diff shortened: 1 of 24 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/VariableValue.kt
```diff
@@ -0,0 +1,3 @@
+package com.apollographql.apollo3.api
+
+class VariableValue(val name: String)
```
sav007, line 3: Should we just call it as `Variable`?

martinbonnin: Maybe? I went for `VariableValue` because it represents a variable value as described in the spec: https://spec.graphql.org/draft/#Value. Also because we have `EnumValue` that is an interface right now but that I'm planning to change to a wrapper class, like for variables in a future PR. On the other hand, the spec does make an exception for variables and while there is `EnumValue`, `IntValue`, etc... `Variable` doesn't have the `Value` suffix so maybe that's better. 

Overall, GraphQL values are "almost" like json elements except that:
* json has no enum or variable
* graphql has no arbitrary precision number

I'd like to find a way to explicit that somehow so that it's easier to know what types to expect instead of the `Any?` that we have in a bunch of places? Maybe with inline classes or type aliases, not sure...

Long story short, works for me to change to `Variable` right now but something else might make more sense a bit later.

PR #2943, opened 2021-02-20
Use adapter for InputFields

martinbonnin:
Use Adapters for `InputFields` and `Variables`, this will help with a bunch of bugs around serializing custom scalars that we have had for a while and also simplifies the codegen by reusing the code from the regular adapters. 

A few notes:
* This PR broke File Upload because the Input objects are serialized earlier in the pipeline and we can't know whether some value is of `FileUpload` type or not from the interceptors. I'll make a separate PR to address this. This will also remove reflection from file upload and make it easier to add on multiplatform.
* There is something around default values and required variables/input fields that is still not 100% correct but adding that in this PR would have made  a big PR, I'll make another PR for this.


apollo-api/src/appleTest/kotlin/com/apollographql/apollo3/ResponseAdapterCacheTest.kt
```diff
@@ -45,7 +45,7 @@ class ResponseAdapterCacheTest {
 }
 
 private fun readInt(cache: ResponseAdapterCache): Int {
-  val adapter = cache.getOperationAdapter("TestQuery") { IntResponseAdapter }
+  val adapter = cache.getAdapterFor(Int::class) { IntResponseAdapter }
```
martinbonnin, line 48: This changed to use a `KClass` key because 2 operations can have the same name if they are in 2 different services

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/SortedInputFieldMapWriter.kt
```diff
@@ -1,139 +0,0 @@
...
-import com.apollographql.apollo3.api.internal.InputFieldMarshaller
-import com.apollographql.apollo3.api.internal.InputFieldWriter
-import com.apollographql.apollo3.api.internal.Throws
-import okio.IOException
-
-class SortedInputFieldMapWriter() : InputFieldWriter {
-  private val buffer = mutableMapOf<String, Any?>()
-
-  fun map(): Map<String, Any?> {
-    return buffer.toList().sortedBy { it.first }.toMap()
```
martinbonnin, line 13: Was sorting required? This PR changes this by serializing the Input Objects in the order of their input fields definition. I don't think it's an issue but maybe I'm missing something?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Operation.kt
```diff
@@ -21,94 +13,39 @@ interface Operation<D : Operation.Data> {
...
    */
-  fun name(): String
+  fun serializeVariables(writer: JsonWriter, responseAdapterCache: ResponseAdapterCache)
 
   /**
-   * Returns a unique identifier for this operation.
+   * A helper class to hold variables
    */
-  fun operationId(): String
+  class Variables(val valueMap: Map<String, Any?>)
```
sav007, line 40: should we make it as delegate to Map?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/OperationExtensions.kt
```diff
@@ -177,3 +179,32 @@ fun <D : Operation.Data, M: Map<String, Any?>> Operation<D>.parseData(
     adapter(responseAdapterCache).fromResponse(it)
   }
 }
+
+fun <D : Operation.Data> Operation<D>.variables(responseAdapterCache: ResponseAdapterCache): Operation.Variables {
```
sav007, line 183: hm looks like this API is expensive, as we always do serialization for variables. Also how user can easily get access to variables for the instance of the query? With this API it needs to provide some internal `ResponseAdapterCache `

sav007 (approved):

PR #2945, opened 2021-02-21
Use the operationId as a key for ResponseFieldMapperFactory

### LINKED ISSUE #2944: Exception while handling subscriptions
**Summary**
When using a shared delegate type for subscriptions you can get an exception when Apollo is trying to handle subscriptions.

**Version**
2.5.3

**Description**
I need to pass some additional data to a custom `OperationMessageSerializer`. To do so I've created an interface like so:
```kotlin
internal interface ExtensibleSubscription {
    val extensions: Map<String, Any?>
}

internal fun <D : Operation.Data, T, V : Operation.Variables> Subscription<D, T, V>.withExtensions(
    extensions: Map<String, Any?>,
): Subscription<D, T, V> = object : Subscription<D, T, V> by this, ExtensibleSubscription {
    override val extensions: Map<String, Any?>
        get() = ((this@withExtensions as? ExtensibleSubscription)?.extensions ?: emptyMap()) + extensions
}
```

Initially this appeared to work well but while parsing the subscription messages I got errors like this:
```
Caused by: java.lang.ClassCastException: com.example.Subscription1$Data cannot be cast to com.example.Subscription2$Data
    at com.example.Subscription2.wrapData(Subscription2.kt:38)
    at com.apollographql.apollo.response.OperationResponseParser.parse(OperationResponseParser.java:70)
    at com.apollographql.apollo.internal.subscription.RealSubscriptionManager.onOperationDataServerMessage(RealSubscriptionManager.java:412)
	... 5 more
```

After som additional research I found that the reason is that `ResponseFieldMapperFactory` uses the operation class as a key which in my case breaks things since the class isn't unique.

It feels like `ResponseFieldMapperFactory` should instead use `operationId` to prevent issues like this.

ansman:
This will prevent errors when wrapping subscriptions in a shared type.

This fixes #2944

### DIFF
--- a/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
+++ b/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt
@@ -27,7 +27,7 @@ import kotlin.coroutines.resume
  */
 @ExperimentalCoroutinesApi
 fun <T> ApolloCall<T>.toFlow(): Flow<Response<T>> = callbackFlow {
-  val clone = clone()
+  val clone = toBuilder().build()
   clone.enqueue(
       object : ApolloCall.Callback<T>() {
         override fun onResponse(response: Response<T>) {
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
@@ -27,7 +27,6 @@ import com.apollographql.apollo.internal.ApolloCallTracker;
 import com.apollographql.apollo.internal.RealApolloCall;
 import com.apollographql.apollo.internal.RealApolloPrefetch;
 import com.apollographql.apollo.internal.RealApolloSubscriptionCall;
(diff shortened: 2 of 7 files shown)


martinbonnin:
Hi 👋 ! Thanks for opening this.

One issue is that operations "could" have the same id in two different services (different schemas). Granted this is an edge case but I've seen it happen in the integration tests already. Possible solutions:
1. Wrapping each `Subscription` in a different class implementing `ExtensibleSubscription` but that doesn't scale at all (unless writing a KSP processor to generate it automatically, but I don't think we want to go that route 😆)
2. Using a Pair<Class, String> as a key, feels a bit sub-optimal but that should work
3. Removing the pool altogether. 

I think 3. makes sense as the adapters are statically allocated anyways, what is cached is a thin wrapper around them that shouldn't cost much and doesn't require caching.

ansman:
Ideally I'd want a way to pass data to the serializer built in but I'm waiting for 3.x.  I removed the usage of the factory but I left if for binary compatibility reasons. 

The class itself is in the internal package and can probably be removed but there are references to it in `ApolloCall` for example

martinbonnin (approved):

martinbonnin:
Thanks!

PR #2946, opened 2021-02-22
Reflection-free file upload

martinbonnin:
Introduce `FileUploadAwareJsonWriter` and use it to intercept calls to `JsonWriter` and remember the uploads and their corresponding path. This makes it possible to do reflection-free uploads and will make it easier to make them work on MPP.

This PR removes 15k lines mainly because it converts some schema.json into schema.sdl


apollo-api/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ kotlin {
     val commonMain by getting {
       dependencies {
         api(groovy.util.Eval.x(project, "x.dep.okio"))
+        api(groovy.util.Eval.x(project, "x.dep.uuid"))
```
martinbonnin, line 12: This is used to create UUIDs that are later used in `multipart/form-data` HTTP bodies. I think this is a small enough dependency that we can add it to `apollo-api`. If we wanted, it should be relatively easy to use a random string instead.

apollo-integration/src/test/kotlin/com/apollographql/apollo3/internal/interceptor/ApolloServerInterceptorFileUploadTest.kt
```diff
@@ -158,35 +160,36 @@ class ApolloServerInterceptorFileUploadTest {
...
         AssertHttpCallFactory(requestAssertPredicate), null, false,
-        ResponseAdapterCache(emptyMap()),
+        adapterCache,
         ApolloLogger(null))
     interceptor.httpPostCall(mutationMultiple, CacheHeaders.NONE, RequestHeaders.NONE, true, false)
   }
 
   @Test
   @Throws(Exception::class)
+  @Ignore("Creates a stack overflow at the moment")
```
martinbonnin, line 182: This is a bit unfortunate. If a `ResponseAdapter` recursively references itself, it will create a stack overflow in the constructor. I'll open a separate PR to address this.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/FileUploadAwareJsonWriter.kt
```diff
@@ -0,0 +1,65 @@
+package com.apollographql.apollo3.api.internal.json
+
+import com.apollographql.apollo3.api.Upload
+
+class FileUploadAwareJsonWriter(private val wrappedWriter: BufferedSinkJsonWriter): JsonWriter {
```
sav007, line 5: wonder if `JsonWriter ` will it be more concise if we delegate to it? And just override value(Upload)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/JsonScope.kt
```diff
@@ -46,14 +46,18 @@ object JsonScope {
    * pathIndices when doing so isn't useful.
    */
   fun getPath(stackSize: Int, stack: IntArray, pathNames: Array<String?>, pathIndices: IntArray): String {
-    val result = StringBuilder().append('$')
+    val result = StringBuilder()
+    var isRoot = true
     for (i in 0 until stackSize) {
       when (stack[i]) {
-        EMPTY_ARRAY, NONEMPTY_ARRAY -> result.append('[').append(pathIndices[i]).append(']')
+        EMPTY_ARRAY, NONEMPTY_ARRAY -> result.append('.').append(pathIndices[i])
```
sav007, line 53: is this right?

sav007 (approved):
Small questions

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/JsonScope.kt
martinbonnin: Not really but this is what FileUpload expects... I put a comment there https://github.com/apollographql/apollo-android/blob/dfc11b8dc51fe12a37e328711f789af3d705c36d/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/BufferedSinkJsonWriter.kt#L71

martinbonnin: Actually, I'm not sure if one is better than the other now that I think about it. 

Nothing prevents some object to have a `"[0]"` key....

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/FileUploadAwareJsonWriter.kt
martinbonnin: I think that won't work because it will return the `wrappedWriter` in chained calls instead of `this`. We might remove the chained calls altogether though since we rarely use them and Kotlin makes it easy to use `apply` if needed

PR #2958, opened 2021-02-25
Introduce ClientContext, RequestContext and ResponseContext

martinbonnin:
The current execution context flow is:

1. ApolloClient is built with some ClientContext
2. A new request is built with its own RequestContext
3. Before execution, the request context with be merged with the client
  context. This allows to access the store or dispatcher from any point
in the interceptor chain
4. The response context is initialized with the full context and the
  populated as it bubbles up the chain
5. Users can read response information such as HTTP headers or Cache headers

I'm a bit unsure about 4. as it potentially gives access to some internals. Maybe we 
could start the response with a new context instead, any thoughts?


apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/cache/CacheExecutionContext.kt
```diff
@@ -7,18 +7,11 @@ import com.apollographql.apollo3.api.Operation
...
-data class CacheResponseExecutionContext(
-    val fromCache: Boolean
-) : ExecutionContext.Element {
-  override val key: ExecutionContext.Key<*> = Key
-
-  companion object Key : ExecutionContext.Key<CacheResponseExecutionContext>
-}
-
-enum class FetchPolicy {
+sealed class FetchPolicy : RequestContext(Key) {
```
martinbonnin, line 14: I went with the coroutines convention of not having a wrapper `Context` object but directly making the implementing class inherit from `Context`

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ExecutionContext.kt
```diff
@@ -4,6 +4,13 @@ import com.apollographql.apollo3.api.ExecutionContext.Element
 import com.apollographql.apollo3.api.ExecutionContext.Key
 import kotlin.jvm.JvmField
 
+@ApolloExperimental
+abstract class ClientContext(override val key: Key<*>): Element
+@ApolloExperimental
+abstract class RequestContext(override val key: Key<*>): Element
+@ApolloExperimental
+abstract class ResponseContext(override val key: Key<*>): Element
```
sav007, line 12: should we provide docs?

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/cache/CacheExecutionContext.kt
```diff
@@ -27,53 +20,57 @@ enum class FetchPolicy {
    *
    * This is the default behaviour
    */
-  CACHE_FIRST,
+  object CACHE_FIRST : FetchPolicy()
```
sav007, line 23: I think we can rename it `CacheFirst` to follow Kotlin naming convention.

sav007 (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ExecutionContext.kt
martinbonnin: Yes, definitly docs 👍 . I'm not adding them first thing as the API changes **a lot** these days and tracking changes in the doc takes time. If that's ok with you, I'll make a `runtime` "API Grooming" PR like the [apollo-api one](https://github.com/apollographql/apollo-android/pull/2954) once things have settled a bit.

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/cache/CacheExecutionContext.kt
martinbonnin: Yup, 100%, will do at the same time as docs.

PR #2963, opened 2021-02-26
Merge Response and ApolloResponse

martinbonnin:
Unify the 2 classes and remove `isFromCache` from the `apollo-api` `Response` 

We might want to rename to `ApolloResponse` at some point. to avoid nameclashes with `okhttp3.Response` or any other response, not sure... At least we should make it symmetrical with `ApolloRequest` so either one of:

* `ApolloRequest` / `ApolloResponse`
or
* `Request` / `Response`


apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/cache/CacheExecutionContext.kt
```diff
@@ -61,11 +60,10 @@ fun <D : Operation.Data> ApolloQueryRequest.Builder<D>.fetchPolicy(policy: Fetch
   addExecutionContext(policy)
 }
 
-val <D : Operation.Data> Response<D>.fromCache
-  get() = executionContext[CacheInfo]?.fromCache ?: throw IllegalStateException("ApolloGraphQL: no CacheInfo")
+val <D : Operation.Data> Response<D>.isFromCache
+  get() = requireNotNull(executionContext[CacheInfo]?.isFromCache)
```
sav007, line 64: should we fallback to `false` instead throwing exception? Not sure if this is a critical issue to not having cache info context?

sav007 (approved):

apollo-cache-interceptor/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/cache/CacheExecutionContext.kt
martinbonnin: 👍  we could also add a better error message but defaulting to `false` makes sens

PR #2966, opened 2021-02-26
Introduce HttpEngine

martinbonnin:
Introduce a lower level HttpEngine interface on top of OkHttp and NSURLSession. This allows to factor the error handling code and also the code that prepares the requests. This might make multiplatform upload work.


apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpUrl.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo3.network.http
+
+expect class HttpUrlBuilder(baseUrl: String, queryParameters: Map<String, String>) {
+  fun build(): String
+}
```
sav007, line 5: is there plans to extend this class in future? It looks a little odd to have class with one fun that returns just a string? Should it be just top level function `buildUrl`? Or  even better introduce `Url / Uri` class that represents url and will have companion factory object?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -129,27 +131,21 @@ class ApolloClient private constructor(
       )
     }
 
-    /**
-     * internal because only used from tests
-     */
-    internal fun interceptors(interceptors: List<ApolloRequestInterceptor>) = apply {
```
sav007, line 135: why we making these API non internal? Are we using it in different module now?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/network/http/ApolloHttpNetworkTransport.kt
```diff
@@ -21,4 +35,98 @@ expect class ApolloHttpNetworkTransport constructor(
      * - on Android, it is used to set  [OkHttpClient.readTimeout]
      */
     readTimeoutMillis: Long = 60_000,
-) : NetworkTransport
+
+    val engine: HttpEngine = DefaultHttpEngine(connectTimeoutMillis, readTimeoutMillis)
```
sav007, line 39: all private `val`s?

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpEngine.kt
```diff
@@ -0,0 +1,56 @@
+package com.apollographql.apollo3.network.http
+
+import okio.BufferedSink
+import okio.BufferedSource
+
+/**
+ * A wrapper around platform specific engines
+ */
+interface HttpEngine {
```
sav007, line 9: Ktor effect :)

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpUrl.kt
martinbonnin: I'd love the Kotlin stlib to ship something there so we can remove this. I'm a bit hesitant on introducing a MPP `Url/Uri` as `Url` pulls a lot of potentially complex things (scheme, hosts, absolute vs relative, etc..) hence the very targetted API. Agreed there's no need for a class though, I'll replace with `buildUrl`

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: We were using it from `apollo-integration-kotlin` but it wasn't strictly need as the public API could do too, I put everything back to `internal`

apollo-runtime-kotlin/src/commonMain/kotlin/com/apollographql/apollo3/network/http/ApolloHttpNetworkTransport.kt
martinbonnin: Yup!

sav007 (approved):

PR #2972, opened 2021-03-01
Return consistently non-nullable String sets in NormalizedCache#merge()

eduardb:
Both `NormalizedCache#merge(java.util.Collection<com.apollographql.apollo.cache.normalized.Record>, com.apollographql.apollo.cache.CacheHeaders)` and `NormalizedCache#performMerge` already return non-nullable String sets, there is no reason why `NormalizedCache#merge(com.apollographql.apollo.cache.normalized.Record, com.apollographql.apollo.cache.CacheHeaders)` wouldn't do the same :)

Context: https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1614619379006200


martinbonnin (approved):
Thanks!

PR #2974, opened 2021-03-03
Add option to generate fragments as classes

sav007:


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/ResponseAdapterReadFromResponse.kt
```diff
@@ -139,6 +155,82 @@ private fun CodeGenerationAst.ObjectType.readFragmentFromResponseFunSpec(): FunS
       .build()
 }
 
+private fun CodeGenerationAst.ObjectType.readFragmentFromBufferedResponseFunSpec(): FunSpec {
```
martinbonnin, line 158: Sounds like this is duplicating a lot of the ResponseAdapter codegen. Would it be possible to somehow factor some stuff?

apollo-compiler/src/test/graphql/com/example/directive_with_fragment/fragmentsAsDataClasses/TestQuery.kt.expected
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * A character from the Star Wars universe
+     */
+    data class Hero(
+      val __typename: String = "Character",
+      /**
+       * The ID of the character
+       */
+      val id: String,
+      val asHeroDetails: HeroDetails.Data?,
```
martinbonnin, line 67: I think we used to have `asFoo` for inline fragments and have `fragments.heroDetails` for named fragments.Can we keep that?

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt
```diff
@@ -181,6 +216,18 @@ class CodegenTest(private val folder: File) {
              */
             file.isDirectory && (filterRegex == null || filterRegex.matchEntire(file.name) != null)
           }
+          .flatMap { file ->
+            if (fragmentsCodegenMode == null) {
+              listOf(
+                  arrayOf(file, FragmentsCodegenMode.FragmentsAsInterfaces),
+                  arrayOf(file, FragmentsCodegenMode.FragmentsAsDataClasses)
```
martinbonnin, line 223: Thinking more about this, it is a bit wasteful to run all the tests, event for operations without fragments like custom scalars and others test. On the other hand, it's still useful to be able to compare the generated code for operations where there are fragments, I'll try to think about something.

apollo-compiler/src/test/graphql/com/example/measurements
```diff
@@ -1,58 +0,0 @@
-fragment_in_fragment                                              100ms                1.87s
-root_query_fragment_with_nested_fragments                        74.9ms                1.27s
```
martinbonnin, line 2: Would be nice to add the number of generated lines to this file so we can compare. I'll do something

martinbonnin (approved):
Looks good overall. That's a lot of changes but the generated code looks way more managable without the interfaces. 
I'm slightly concerned about bloating the repo and CI time with all these test fixtures but we can refine that as we go.

apollo-compiler/src/test/graphql/com/example/directive_with_fragment/fragmentsAsDataClasses/adapter/TestQuery_ResponseAdapter.kt.expected
```diff
@@ -0,0 +1,141 @@
...
+      var asHumanDetails: HumanDetails.Data? = null
+      while(true) {
+        when (reader.selectName(RESPONSE_NAMES)) {
+          0 -> __typename = stringAdapter.fromResponse(reader)
+          1 -> id = stringAdapter.fromResponse(reader)
+          else -> break
+        }
+      }
+      if (__typename in arrayOf("Droid", "Human")) {
+        reader.rewind()
```
martinbonnin, line 100: 🚀

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/ResponseAdapterReadFromResponse.kt
sav007: not sure about that, even it looks like code is duplicated it's different `readFragmentFromStreamResponseFunSpec` and `readFragmentFromBufferedResponseFunSpec` emit entirely different code to for reading from response. The only place where they overlap is function declaration.

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt
sav007: I put a fix to run 2 tests for queries with fragments only.

apollo-compiler/src/test/graphql/com/example/measurements
sav007: no worries I will do that, should be trivial

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/ResponseAdapterReadFromResponse.kt
martinbonnin: Right. 

For the record, Data classes is like

```
      if (__typename in arrayOf("Droid", "Human")) {
        reader.rewind()
        heroDetails = HeroDetailsAdapter.fromResponse(reader)
      }
      if (__typename in arrayOf("Human")) {
        reader.rewind()
        humanDetails = HumanDetailsAdapter.fromResponse(reader)
      }
```

Interfaces is like

```
      return when(typename) {
        "Droid" -> CharacterHeroAdapter.fromResponse(reader, typename)
        "Human" -> CharacterHumanHeroAdapter.fromResponse(reader, typename)
        else -> OtherHeroAdapter.fromResponse(reader, typename)
      }
```

The main change is how we iterate, in Data classes, we iterate on each type condition while in Interfaces we iterate on each Type case. So agreed. We could factor in some stuff like reading the typename and function declaration but it might not be worth spreading the code in multiple functions

martinbonnin:
"Only" 15k lines diff now :)! Looks good to me. We certainly need to move the `ResponseFields` outside the adapters like discussed offline but that can be in a separate PR.

martinbonnin:
I think the CI fails because of the merge in expected files. Can you rebase?

PR #2976, opened 2021-03-03
Use RealApolloStore in apollo-runtime-kotlin

martinbonnin:


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/CacheHolder.kt
```diff
@@ -0,0 +1,27 @@
+package com.apollographql.apollo3.cache.normalized.internal
+
+import com.apollographql.apollo3.cache.normalized.ReadOnlyNormalizedCache
+
+class CacheHolder(producer: () -> OptimisticCache) {
```
sav007, line 5: I'm not sure what the purpose of the producer here as when the `CacheHolder ` is created prodicer will be called right away to initialize `optimisticCache ` property.

Also overall question, not sure what is the purpose of this class? As it looks like a wrapped around `ReentrantReadWriteLock`?

apollo-runtime/src/main/java/com/apollographql/apollo3/internal/RealApolloSubscriptionCall.kt
```diff
@@ -123,7 +128,13 @@ class RealApolloSubscriptionCall<D : Subscription.Data>(
...
       if (callback != null) {
         if (data != null && delegate!!.cachePolicy != ApolloSubscriptionCall.CachePolicy.NO_CACHE) {
-          delegate!!.apolloStore.writeOperation(response.subscription, data)
+          delegate!!.apolloStore.writeOperation(
+              response.subscription,
+              data,
+              delegate!!.responseAdapterCache,
+              CacheHeaders.NONE,
+              true,
+              )
```
sav007, line 137: nit: could you please add names for the arguments for better readability

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-milestone-1-bin.zip
```
sav007, line 3: whoa, should we migrate to m1?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/CacheHolder.kt
martinbonnin: I'm planning to turn this into an interface. We would have:

```kotlin
interface CacheHolder {
  fun <T> read(block: (ReadOnlyNormalizedCache) -> T)

  fun <T> write(block: (OptimisticCache) -> T): T

  fun writeAndForget(block: (OptimisticCache) -> Unit) 
}
```

And two different implementations:

```kotlin
# Uses Reentrant locks on the JVM
class LockCacheHolder(producer: () -> OpimisticCache): CacheHolder(producer)
# Uses a singleThreadExecutor on the JVM and a serial Queue on Apple
expect class IsolateCacheHolder(producer: () -> OpimisticCache): CacheHolder(producer)
```

You're right that the constructor doesn't belong to the interface and doesn't need to be a producer. I'll change to.

```kotlin
class LockCacheHolder(optimisticCache: OpimisticCache): CacheHolder
```

The long term plan is to have an option during `ApolloStore creation`:

```kotlin
enum JvmConcurencyMode {
  // Uses a ReadWriteLock (SQLite cannot do concurrent reads but memory cache can)
  LOCK,
  // Uses an Isolate pattern, like on native. Incurs some thread switching costs but alllows to do the writes asyncs
  ISOLATE
}
val store = ApolloStore(cacheFactory: NormalizedCacheFactory, cacheKeyResolver: CacheKeyResolver, jvmConcurrencyMode: JvmConcurencyMode)
```

apollo-runtime/src/main/java/com/apollographql/apollo3/internal/RealApolloSubscriptionCall.kt
martinbonnin: Sure thing !

gradle/wrapper/gradle-wrapper.properties
martinbonnin: I think this is fine as it's only used for building so it only has to work for us. The change is transparent for users. 

I'm looking forward to `7.0` as it comes with Kotlin 1.4 \o/. Which means we can remove `apiLevel = 1.3` in our code and start using 1.4 for real in our code. 

While we're not using any 1.4 API in our code, the same isn't always true about dependencies and I've been hit by https://github.com/Kotlin/kotlinx.atomicfu/issues/171. It's not clear why this wasn't happening before but overall, if we can adopt newer versions of Gradle faster I think that'll help moving forward.

martinbonnin: PS: talking about other kinds of m1, does K/N support Apple M1 chips ?!?

sav007: Hm, interesting question, not sure about this.

sav007 (approved):

PR #2982, opened 2021-03-04
Fix code example

BRoy98:
- Error should be logged on failure and response should be logged on success
- `onResponse` should come before `onFailure`


martinbonnin (approved):
A very welcome improvement, Thanks!

PR #2983, opened 2021-03-05
Added native mockserver

martinbonnin:
Added a simple native MockServer like OkHttp's `MockWebServer` so that we can test the native code end-to-end. 

This already allowed to find a few issues in `NSUrlSessionHttpEngine` and will make debugging the cache concurrency issues way easier.


sav007 (commented):
🎉

sav007 (approved):

PR #2986, opened 2021-03-08
Add an Isolate cache holder

martinbonnin:
Added a [Stately Isolate](https://github.com/touchlab/Stately)-inspired isolate cache for K/N. It's still blocking threads because it's using `Workers`. The `IOTaskScheduler` doesn't work here because there's no guarantee that dispatch queues will always run on the same thread, even for serial queue. That means the `suspend` modifier in the ApolloStore API aren't strictly needed at the moment but since I started adding them, I left them. That'll allow us to see if there are places where this brings complications


sav007:
Question:
> It's still blocking threads because it's using Workers
> The IOTaskScheduler doesn't work here because there's no guarantee that dispatch queues will always run on the same thread, even for serial queue.

If we enforce user to always run Apollo call on iOS from main thread then why can't we execute cache operation in Worker (and it's always single threaded) and return back to main thread via continuation? Same as we don in HTTP layer?

martinbonnin:
> If we enforce user to always run Apollo call on iOS from main thread then why can't we execute cache operation in Worker (and it's always single threaded) and return back to main thread via continuation? Same as we did in HTTP layer?

The HTTP thread doesn't use any mutable state so it can be dispatched in whatever thread GCD has handy. That doesn't work any more when we want to mutate the cache.

martinbonnin:
Sorry I misunderstood the question. That doesn't work either. For some reason I haven't been able to pass a continuation to the worker. It always throw with `Invalid Transfer State`. Maybe it'd work with coroutine-mt, not sure. I'll dig into this but it might take some time. I prefered to ship this (suboptimal/blocking) version and revisit later to make it really suspend.

PR #2991, opened 2021-03-11
Merge inline fragment selection set on same type condition

sav007:
Merge inline fragment selection with parent if it's defined on the same type condition.


martinbonnin (approved):
Looks good to me overall. I believe the `@include`/`@skip` directive handling needs to be reworked but this was the case before so we can revisit in a future PR. 

Proposal for handling the `@include` conditions:
1. For all inline fragments, remove the `@include` condition and spread it on individual fields, recursively. 
2. Flatten trivial inline fragments, i.e. inline fragments that are defined on the same type condition as their enclosing type (including potentially the base type). This has to be done after 1. because this removes intermediate inline fragments so their directives have to be spread before in order not to be lost.
3. Merge inline fragments with the same type condition inside each selection set. 
4. Merge fields with the same response name inside each selection set (This can be done in the same step as 3.)

PR #2994, opened 2021-03-11
Add :integration-tests:testAsClasses

martinbonnin:
Split the integration tests into several tasks sharing some common code so that we can test the different codegen variations.

Also add `generateFragmentsAsInterfaces` as a plugin parameter

`testAsClasses` is disabled for now


composite/integration-tests/src/testShared/kotlin/com/apollographql/apollo3/internal/interceptor/ApolloServerInterceptorFileUploadTest.kt
```diff
@@ -181,7 +181,6 @@ class ApolloServerInterceptorFileUploadTest {
 
   @Test
   @Throws(Exception::class)
-  @Ignore("Creates a stack overflow at the moment")
```
martinbonnin, line 184: I forgot to enable this test while making the ResponseAdapters stateless. It's good to see it worked :)

sav007 (approved):

PR #2996, opened 2021-03-12
Add a `fromField` function for Fragments as classes

martinbonnin:
When reading fragment as classes, integration tests were failing because `reader.beginObject()` was called multiple times for synthetic fragment fields. For these cases, introduce a `fromFields()` function that reads the field without consuming the object.

Also enable a lot more integration tests


martinbonnin:
Superseded by #2997

PR #2997, opened 2021-03-12
Introduce ObjectResponseAdapter

martinbonnin:
Supersedes https://github.com/apollographql/apollo-android/pull/2996

Instead of having `fromFields` and `fromResponse`, make all generated compound adapters read/write fields but not the `beginObject()`/`endObject()`. This simplifies the generated code and makes it possible to handle the buffering in a single place. 

As in #2996, this enables integration tests for Fragments as Classes


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/ast/Ast.kt
```diff
@@ -61,7 +64,7 @@ internal data class CodeGenerationAst(
       val typeRef: TypeRef,
       val schemaTypename: String?,
       val fragmentAccessors: List<FragmentAccessor>,
-      val isTypeCase: Boolean,
+      val isShape: Boolean,
```
sav007, line 67: not really sure what this name means, could you please provide more context?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/EnumType.kt
```diff
@@ -48,7 +51,10 @@ private fun CodeGenerationAst.EnumType.toEnumTypeSpec(generateAsInternal: Boolea
 }
 
 private fun CodeGenerationAst.EnumType.adapterTypeSpec(generateAsInternal: Boolean, asSealedClass: Boolean, packageName: String): TypeSpec {
-  val fromResponseFunSpec = fromResponseFunSpecBuilder()
+  val fromResponseFunSpec = FunSpec.builder("fromResponse")
+      .addModifiers(KModifier.OVERRIDE)
+      .addParameter(ParameterSpec.builder("reader", JsonReader::class).build())
```
sav007, line 56: nit: should we use `Identifiers.reader`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/StandardResponseAdapters.kt
```diff
@@ -125,8 +128,47 @@ object UploadResponseAdapter : ResponseAdapter<Upload> {
...
+      writer.beginObject()
+      wrappedAdapter.toResponse(writer, responseAdapterCache, value)
+      writer.endObject()
+    }
+  }
+}
+
+fun <T : Any> ResponseAdapter<T>.nullable() = NullableResponseAdapter(this)
 fun <T> ResponseAdapter<T>.list() = ListResponseAdapter(this)
+fun <T> ResponseAdapter<T>.obj(buffered: Boolean = false) = ObjectResponseAdapter(this, buffered)
```
sav007, line 171: should we use more readable name like `.toObjectResponseAdapter` or `.objectResponseAdapter`, `asObjectResponseAdapter`

sav007 (approved):
some comments around naming

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/ast/Ast.kt
martinbonnin: @martijnwalraven pointed out that `TypeCase` was originally meant for the whole `when` statement as in programming languages like modula (https://www.cs.purdue.edu/homes/hosking/m3/reference/typecase.html) and not to the individual sets of type conditions. 

I'm trying to find something that conveys this idea of "partition of possible types" and `Shape` seemed to fit the requirements. `Shape` is also mentioned in the spec: https://spec.graphql.org/draft/#SameResponseShape(). And I think that'll apply here as well. A shape is the shape of a response object (vs interfaces that are sub trees of shapes)

I'll add some Kdoc in the next PR.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/EnumType.kt
martinbonnin: Yup, make sense.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/StandardResponseAdapters.kt
martinbonnin: I'd like a "short" name as this ends up being used all over the place in generated adapters:

```kotlin
Friends.obj().nullable().list().fromResponse(reader, responseAdapterCache)

// Compared to
Friends.asObjectResponseAdapter().asNullableResponseAdapter().asListResponseAdapter().fromResponse(reader, responseAdapterCache)
```

Since there can be a bunch of chained calls, keeping them short will make sure it fits on one line in most cases and it's displayed nicely.  Ideally, I was aiming for `object()` but this is a hard keyword in kotlin so that doesn't work. Any other suggestion of short names?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/backend/codegen/EnumType.kt
martinbonnin: I made most of the response readers/writers use `Identifier.foo` for most of the codeblocks. I'll do other parts as I encounter them (if there are more, I think response adapters are the places with the most actual codeblocks)

PR #3000, opened 2021-03-15
Unified IR

martinbonnin:
A draft branch to investigate an "UnifiedIr" to unify FrontendIr and BackendIr. The goals are to:

1. Share the same IR whether we generate fragments as interfaces or fragments as classes. We need the shapes in all cases for the response fields. 
2. Generate only one interface no matter the order of fragments. If we have a onCharacter().onDroid(), the generated interface should be the same as onDroid.onCharacter() (that exemple is a bit silly but still holds with more generic interfaces)
3. Avoid nesting interfaces with the same name as their parent to avoid kapt issues.

Additionally, it could be helpful to :
* Reduce the number of interfaces implemented by data classes by omitting the interfaces included transitively
* Add a "flattening" option to workaround the 256 filename limit
* Postpone Kotlin-specific phases (capitalization/keywordEscape) to the codegen step.
* Handle type conditions on named fragments.

The general idea is to build a huuuuuuge tree of all possible response shapes and interfaces and prune it as needed. It's not yet 100% clear how that will work (if that ends up working) hence the draft status. I'll update this PR as I go.


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/unified/FragmentCollectionScope.kt
```diff
@@ -0,0 +1,84 @@
...
+        is GQLInlineFragment -> it.selectionSet.selections.collectInternal(typeSet + it.typeCondition.name)
+        is GQLFragmentSpread -> {
+          val fragmentDefinition = allGQLFragmentDefinitions[it.name]!!
+          collectedNamedFragments.add(
+              CollectedNamedFragment(
+                  it.name,
+                  fragmentDefinition.typeCondition.name,
+                  BooleanExpression.And(typeSet.map { BooleanExpression.Type(it) }.toSet())
+                      // TODO: allow @include directives on named fragments
+                      //.and(it.directives.toBooleanExpression())
```
martinbonnin, line 75: This is out of scope for now but that would be the place to support `@include` directives on named fragments.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/unified/IrBuilder.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    val inputObjects = schema.typeDefinitions.values
+        .filterIsInstance<GQLInputObjectTypeDefinition>()
+        .filter { usedInputObjects.contains(it.name) || shouldAlwaysGenerate(it.name) }
+        .map { it.toIr() }
+
+    val customScalars = schema.typeDefinitions.values
+        .filterIsInstance<GQLScalarTypeDefinition>()
+        .filter { !it.isBuiltIn() }
+        .filter { usedCustomScalars.contains(it.name) || shouldAlwaysGenerate(it.name) }
```
martinbonnin, line 70: I kind of like how the `usedTypes` stuff turned out :)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/unified/IrBuilder.kt
```diff
@@ -0,0 +1,318 @@
...
+        }
+        is GQLEnumTypeDefinition -> {
+          usedEnums.add(name)
+          IrEnumType(name)
+        }
+        is GQLObjectTypeDefinition -> IrObjectType(name)
+        is GQLInterfaceTypeDefinition -> IrInterfaceType(name)
+        is GQLUnionTypeDefinition -> IrUnionType(name)
+        is GQLInputObjectTypeDefinition -> {
+          usedInputObjects.add(name)
```
martinbonnin, line 245: This is where we record the used types. This way we have a central place to record all usages.

martinbonnin:
The main classes should be good to review now. This is not plugged anywhere yet. Next step is to generate another IR that will be easily mapable to KotlinPoet TypeSpecs.

I tried to put some doc but do not hesitate to comment on things that are unclear.

martinbonnin:
Moving this one back to draft. The previous IR had a bunch of issues I didn't find until trying to actually generate the code

martinbonnin:
Moving to "in-review" again. There's a lot of changes and the IR now produces code that compiles so it's a good milestone. The IR isn't plugged to `GraphQLCompiler` yet so all the tests are still using the existing `FrontendIR` & `BackendIR`.

Next steps are to migrate tests to the unified IR progressively
* codegen
* multi-module
* integration
* fragments as classes
* etc...

This PR also contains a high level overview of the compiler there: https://github.com/apollographql/apollo-android/blob/651874dd0814cdb7d03b96aecfc21ab73ad3c910/design-docs/Compiler.md

PR #3002, opened 2021-03-15
Make sure all publications get uploaded to the same staging repo

martinbonnin:
OSSRH has a tendency of creating multiple staging repos. This happened twice today will trying to upload the 1000+ files of `3.0.0-dev5`. Closing individual staging repos is not an option in that case because the artifacts seem randomly dispatch which means that the signature will not always be in the same repo as the jar for an example. Which makes the verifications fail.

Instead, create a new staging repo explicitely and upload all artifacts to it.

Also move all the publishing logic to `build-logic` and fix the sources artifacts that were somewhat lost in `dev-3.x`

### DIFF
--- a/apollo-compiler/build.gradle.kts
+++ b/apollo-compiler/build.gradle.kts
@@ -30,7 +30,7 @@ abstract class GeneratePluginVersion : DefaultTask() {
 
   @org.gradle.api.tasks.TaskAction
   fun taskAction() {
-    val versionFile = File(outputDir.asFile.get(), "Version.kt")
+    val versionFile = File(outputDir.asFile.get(), "com/apollographql/apollo3/compiler/Version.kt")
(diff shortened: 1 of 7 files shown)


apollo-compiler/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ abstract class GeneratePluginVersion : DefaultTask() {
 
   @org.gradle.api.tasks.TaskAction
   fun taskAction() {
-    val versionFile = File(outputDir.asFile.get(), "Version.kt")
+    val versionFile = File(outputDir.asFile.get(), "com/apollographql/apollo3/compiler/Version.kt")
```
martinbonnin, line 33: That change is not related to publishing. It makes sure the generated file is under a correct folder in the sourcesJar else we end up with `Version.kt` at the root of the Jar

PR #3006, opened 2021-03-17
Read form feed sequence as well

Romadro:
According to Google's Gson JsonReader class https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1538 you have to read escaped '\f' character as well (used for page break). And also please can you suggest some workaround for this problem for now?


martinbonnin:
Thanks! 

That'll be in the next release. I'm planning to make one next Monday. In the meantime, you'll have to replace the escape at the call-site (Something like `str.replace("\\f", "\u000c")`) or, if that happens in a lot of places, you can install a OkHttp interceptor that rewrites your json with `\u000c` that is supported by the current parser.

Romadro:
Thank you for helping me out!

martinbonnin:
2.5.5 is released, let me know how that goes!

PR #3025, opened 2021-03-30
Update apolloClient query .await()

AndreSand:
Hi, I was following the Apollo tutorial and I had to add `toDeferred()` to get it to work:
from:
val response = apolloClient.query(LaunchListQuery()).await()

to:
val response = apolloClient.query(LaunchListQuery()).toDeferred().await()

### DIFF
--- a/docs/source/tutorial/04-execute-the-query.mdx
+++ b/docs/source/tutorial/04-execute-the-query.mdx
@@ -27,7 +27,7 @@ Open `LaunchListFragment.kt`, override `onViewCreated` and launch a new coroutin
         super.onViewCreated(view, savedInstanceState)
 
         lifecycleScope.launchWhenResumed {
-            val response = apolloClient.query(LaunchListQuery()).await()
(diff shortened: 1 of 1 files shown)


martinbonnin:
Hi @AndreSand 👋 ! Thanks for reaching out! There should be a `fun ApolloCall.await()` (see https://github.com/apollographql/apollo-android/blob/main/apollo-coroutines-support/src/main/kotlin/com/apollographql/apollo/coroutines/CoroutinesExtensions.kt#L90). 

Can you double check your imports? Maybe add something like this?

```
import com.apollographql.apollo.coroutines.await
```

AndreSand:
Hi @martinbonnin, I added the import you mentioned above, but it didn't worked.

I updated to latest apollo version `2.5.3` and it worked using:
`val response = apolloClient.query(LaunchListQuery()).await()`

Maybe we can update the documentation to use apollo latest version 2.5.3.

PR #3046, opened 2021-04-19
Use KSP and Moshix to generate Json adapters

martinbonnin:
While iterating on the compiler I had wayyyyy to many stale kapt files
issue. That should fix it.
KSP itself is still experimental but I think it's ok to use since it's
run mostly on the dev machine and the generated adapter do not do
anything crazy. On the contrary, it's a way to move the ecosystem
forward by providing feedback if needed.

### DIFF
--- a/apollo-compiler/build.gradle.kts
+++ b/apollo-compiler/build.gradle.kts
@@ -57,8 +57,12 @@ configure<org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension> {
 }
 
 tasks.withType(KotlinCompile::class.java) {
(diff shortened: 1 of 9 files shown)


tasomaniac (approved):
Looks good. Glad to see KSP fixes the problems of kapt.

martinbonnin:
> Looks good. Glad to see KSP fixes the problems of kapt.

Recently I had a kapt build error each time I switched branches. Happy to see them go :) 

```
> Task :apollo-compiler:kspKotlin
e: /Users/runner/work/apollo-android/apollo-android/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/frontend/GraphQLParser.kt: (3, 43): Unresolved reference: parser
```

Looks like there's something not quite right with antlr... Maybe some task dependencies.

PR #3073, opened 2021-04-28
 Add a "Who is Apollo" section to the main README

daniman:
We're trying to add this section to the main READMEs of all our projects (e.g. https://github.com/apollographql/apollo-client#who-is-apollo).


hwillson (approved):
Thanks very much @daniman!

PR #3074, opened 2021-04-28
Normalized Cache: Make `variables` available to the `fromFieldRecordSet` method in `CacheKeyResolver`

rohandhruva:
Per slack discussion, this is a quick change to make `variables` available in `CacheKeyResolver.fromFieldRecordSet`. 

The idea is that we are now able to uniformly normalize a field in both `fromFieldArguments` and `fromFieldRecordSet`, without having the need to reflect back one of the input variables if not present in the response `recordSet`. 

Separately, one potential problem which this allows us to fix: if we are normalizing by `id`, we can now verify that the `id` we passed in `variables` is actually equal to the `id` we get back in `recordSet`. In some APIs, a new `id` returned in `recordSet` can be used as a signal by the client to treat it as a cache miss.  

Note that this is a preliminary approach to the original problem we discussed: we might decide to find other solutions, in which case this API is subject to change.


martinbonnin (approved):
LGTM! Reminder to future selves to add a test if we validate this approach.

PR #3091, opened 2021-05-04
Pretty print schema, minor typo fix

rohandhruva:
Passing an `indent` should allow the schema from introspection to be pretty-printed. I also fixed a minor typo that I found in the gradle plugin.

I've not tested this change yet because I couldn't find an existing test in `ServiceTests`, but I will add one as a follow-up.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
@@ -70,17 +70,7 @@ inline fun <reified T> T.toJson(indent: String = ""): String {
   return getJsonAdapter<T>().indent(indent).toJson(this)
 }
 
-inline fun <reified T> T.toJson(writer: JsonWriter) {
-  getJsonAdapter<T>().toJson(writer, this)
-}
-
-inline fun <reified T> T.toJson(file: File) {
-  file.outputStream().sink().buffer().use {
-    return getJsonAdapter<T>().toJson(it, this)
-  }
-}
-
-inline fun <reified T> T.toJson(file: File, indent: String) {
+inline fun <reified T> T.toJson(file: File, indent: String = "") {
   file.outputStream().sink().buffer().use {
     return getJsonAdapter<T>().indent(indent).toJson(it, this)
   }
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
@@ -66,8 +66,8 @@ inline fun <reified T> File.fromJsonList(): List<T> {
   return inputStream().fromJsonList<T>()
 }
 
-inline fun <reified T> T.toJson(): String {
-  return getJsonAdapter<T>().toJson(this)
+inline fun <reified T> T.toJson(indent: String = ""): String {
+  return getJsonAdapter<T>().indent(indent).toJson(this)
 }
 
 inline fun <reified T> T.toJson(writer: JsonWriter) {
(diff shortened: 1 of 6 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
```diff
@@ -66,8 +66,8 @@ inline fun <reified T> File.fromJsonList(): List<T> {
   return inputStream().fromJsonList<T>()
 }
 
-inline fun <reified T> T.toJson(): String {
-  return getJsonAdapter<T>().toJson(this)
+inline fun <reified T> T.toJson(indent: String = ""): String {
```
martinbonnin, line 69: Makes sense. We actually already have `toJson(file: File, indent: String)` and `toJson(file: File)`. Might be worth uniformizing the API:

```kotlin
toJson(file: File, indent: String = "")
toJson(sink: BufferedSink, indent: String = "")
toJson(indent: String = ""): String
```

I don't think we need the version that takes a `JsonWriter`. Want to look into it? Or I can push that to your branch if you prefer.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/internal/BuildDirLayout.kt
```diff
@@ -7,7 +7,7 @@ import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
 
 object BuildDirLayout {
-  internal fun operationOuput(project: Project, service: Service): Provider<RegularFile> {
+  internal fun operationOutput(project: Project, service: Service): Provider<RegularFile> {
```
martinbonnin, line 10: Thanks !

martinbonnin (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/moshi.kt
rohandhruva: I can definitely remove the `JsonWriter`.

> We actually already have toJson(file: File, indent: String) and toJson(file: File).

Let me see if I can unify those: I didn't want to change the version with a `file` param, but maybe it's ok to do so.

BTW, kinda related: what's strange is that `OperationOutput.toJson(indent: String? = null)` doesn't seem to be working...? I noticed that the `GraphQLCompiler` uses it for writing the `operationOutput.json` file like so:

> operationOutputFile.writeText(operationOutput.toJson("  "))

But as far as I can tell, the written file still uses compact JSON formatting and is not indented. Do you have any ideas on how I can debug that? I could add the fix to this PR as well!

rohandhruva: I've updated the PR with the uniform API that you suggested. I couldn't find this one:

> toJson(sink: BufferedSink, indent: String = "")

I've also updated the `OperationOutput` helpers to switch `indent` from `String?` to `String` with default value of `""` for uniformity.

martinbonnin: Looks like there's a mixup in the overloads:

```
inline fun <reified T> T.toJson(file: File, indent: String) {
```

calls 

```
inline fun <reified T> T.toJson(writer: JsonWriter) {
  getJsonAdapter<T>().toJson(writer, this)
}
```

But since this last one creates a fresh adapter with `getJsonAdapter<T>()`, the indent is lost...

rohandhruva: Looks like that's the one I ended up removing from `moshi.kt` anyway?

The one I was referring to earlier is in `OperationOutput.kt` I believe.

martinbonnin:
Thanks !

PR #3092, opened 2021-05-04
Added support for extraSchemaFiles

martinbonnin:
See `NonNullTests` for usage. It's now possible to add extra client directives at the schema level. One example where this is useful (and the only one so far) is to mark fields `@nonnull` globally without having to add the annotation to each and every query:

```graphql
extends type User @nonnull(fields: ["firstName"])
```

The `@nonnull` directive, when used on an object definition requires one `fields` param that is a list of strings. We might want to add some custom logic.

This sent me down the rabbit hole of the schema parsing API so I took this opportunity to unify a lot of the APIs there.

Everything is now an extension function so stuff like this is doable (albeit not super useful 😅 ):

```kotlin
file.toIntrospectionSchema().toGQLDocument().toIntrospectionSchema()
```


rohandhruva:
@martinbonnin , I _think_ this change might have introduced a problem where the intellij plugin (JS GraphQL) no longer recognizes the SDL schema as a valid schema :( 

This is the issue I see in the plugin: https://gist.github.com/rohandhruva/ecc16f4eea877caddea240b2d5a758fa

I'm not sure if `__Schema` always used to be written out or if it's something that changed. 

I also see an error in the generated file:

image
 
I haven't downgraded yet to confirm whether it's a dev9 ->  dev10 issue, but I will try that in a bit. I seem to remember it used to work, though.

rohandhruva:
For the time being, I switched back to using JSON schema: only to discover that the JSON pretty printing doesn't work :( I'm not sure if PR#3091 ever worked, or if it changed on this PR. I'll try to test that.

Switching to JSON schema does fix the intellij js graphql plugin issue though.

martinbonnin:
Good catch! Will dig into this!

Issue #3095, opened 2021-05-05
Subscriptions seem to be dropping items


nealsanche:
**Summary**
Subscriptions are dropping items if the items are delivered all at once, often only the first item will be delivered.

**Version**
GraphQL Android Version 2.5.6

**Description**

I wrote a small server based on https://github.com/ExpediaGroup/graphql-kotlin/tree/master/examples/server/spring-server with the following code inside it:

```kotlin
@Component
class TestSubscription : Subscription {
    @GraphQLDescription("Returns stream of values.")
    fun flowBuilder(): Publisher<String> = kotlinx.coroutines.flow.flow {
        emit("hi")
        emit("there")
        emit("world")
    }.asPublisher()
}
```
This simply emits a set of values, and then finishes.

On the apollo-android side, I have written code something like the following:

```kotlin
    val subscriptionApolloClient by lazy {
        ApolloClient(
            networkTransport = ApolloWebSocketNetworkTransport(
                webSocketFactory = ApolloWebSocketFactory(
                    serverUrl = "$serverUrl/subscriptions",
                ),
                connectionAcknowledgeTimeoutMs = 1000,
                connectionKeepAliveTimeoutMs = 1000,
                idleTimeoutMs = 5000,
            )
        )
    }

class TestSubscription {
    @Throws(QueryError::class, CancellationException::class, NullPointerException::class, ApolloNetworkException::class)
    operator fun invoke() = subscriptionApolloClient.subscribe(
        TestFlowSubscription()
    ).execute().map {
        if (it.hasErrors()) {
            throw QueryError(it.errors)
        } else {
            it.data!!.flow
        }
    }
}

```

This small class can be used in a coroutine:

```kotlin
fun testSubscription() {
  viewModelScope.launch {
     val subscription = TestSubscription()
     subscription().collect {
        Log.i("TAG", "Got $it")
     }
  }
}
```
I have been seeing mainly "hi" come back, and then the collect finishes. If I put delays between each element on the server, I will get more of the items.

In asking around, there was a suggestion that if I'm not using `graphql-ws` protocol, I should be. Is this what is used inside graphql-android?

PR #3096, opened 2021-05-05
Issue #3095: Expose channel capacity parameter.

nealsanche:
# Issue

Related to issue #3095 

# Introduction

The current implementation had Channel.CONFLATED as the capacity, which can drop messages if several come in at roughly the same time. This change simply opens that up and allows it to be configurable, with the default being how it is currently implemented.

# Testing

See the issue for a test case. I have built and tested the scenario I was dealing with and will be using Channel.BUFFERED for my case, as it works as I would expect.


martinbonnin (approved):
Thanks !

Issue #3108, opened 2021-05-14
Codegen compilation error for Float scalars with default values in Input objects


joaquim-verges:
**Summary**

We recently upgraded from Apollo Android 2.4.1 to 2.5.6, and encountered a compilation error in the autogenerated code.

In short, Input objects in our `json` schema with inputs of type `Float` and default values that omit the decimal part (eg. `"defaultValue": "1"`) causes the Java codegen to generate a default value as an `Integer` instead of `Double` which causes a compilation error:

```
private Input<Double> myFloatField = Input.fromNullable(1);
                                                        ^
equality constraints: Double
lower bounds: Integer
```

**Version**
2.5.6

**Description**

I was able to narrow down the issue to a specific codegen corner case:

- Only happens in Java codegen (issue does not happen with Kotlin models)
- Only happens with `json` schema format (issue does not happen with `sdl` schema format)

When a `INPUT_OBJECT` type has a `Float` input field with a default value of `1` or `0` (omitting the decimal part) like so:

```json
 {
  "kind": "INPUT_OBJECT",
  "name": "ReviewRefInput",
  "description": "Circle ref to review input",
  "fields": null,
  "inputFields": [
    {
      "name": "myFloatField",
      "description": "Testing float input codegen",
      "type": {
        "kind": "SCALAR",
        "name": "Float",
        "ofType": null
      },
      "defaultValue": "1"
   }
  ]
}
```

The corresponding output in the generated java file is:

```java
private Input<Double> myFloatField = Input.fromNullable(1);
```

Which is a compilation error, as `1` is considered an `Integer`, and therefore that assignment fails with:

```
private Input<Double> myFloatField = Input.fromNullable(1);
                                                        ^
equality constraints: Double
lower bounds: Integer
```

The equivalent Kotlin generated code is correct however:

```kotlin
val myFloatField: Input<Double> = Input.optional(1.0)
```

The Kotlin codegen correctly converts `1` to `1.0` since the scalar is type `Float`, but the Java codegen does not.

This also happens for mandatory fields:

java: `private double mandatoryFloat = 1;`
kotlin: `val mandatoryFloat: Double = 1.0`

Interestingly, the java codegen does the correct transformation when using schemas in `sdl` format, so the issue only happens when parsing from `json`.

[...]

PR #3109, opened 2021-05-14
Fix Float input fields Java codegen with default values in json schemas

joaquim-verges:
see https://github.com/apollographql/apollo-android/issues/3108

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
@@ -166,7 +166,7 @@ class IRBuilder(private val schema: IntrospectionSchema,
         }
       is GraphQLParser.ArrayValueContext -> {
         this.arrayValueType().valueOrVariable()?.map {
-          it.value()?.toKotlin(type)
+          it.value()?.toKotlin(type.ofType!!)
         } ?: emptyList<Any?>()
       }
       is GraphQLParser.LiteralValueContext -> {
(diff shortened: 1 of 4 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IRBuilder.kt
```diff
@@ -139,35 +139,34 @@ class IRBuilder(private val schema: IntrospectionSchema,
...
+      is GraphQLParser.NumberValueContext ->
+        when (ScalarType.forName(type.name ?: "")) {
+          ScalarType.INT -> NUMBER().text.toInt()
+          ScalarType.FLOAT -> NUMBER().text.toDouble()
+          else -> throw ParseException("Unexpected number type")
+        }
       is GraphQLParser.ArrayValueContext -> {
         this.arrayValueType().valueOrVariable()?.map {
-          it.value()?.toKotlin()
+          it.value()?.toKotlin(type)
```
martinbonnin, line 169: For lists, this should "unwrap" the item type with `type.ofType` as it recurses in order to coerce scalar items appropriately. For an example with `defaultValue = "[3]"`, `type.kind == LIST` and `type.ofType.kind == SCALAR`. Same thing for input fields where it gets even more complicated as we'll need a reference to the schema to lookup the inputFieldDefinition.

martinbonnin:
Thanks for catching this! I believe there are edge cases around lists/input objects where nested values wouldn't get the proper type. Let me know if you want to keep digging ⛏️ . Or I can also try to fix this as I introduced that bug in the first place.

joaquim-verges:
@martinbonnin yep this doesn't cover the nested values case, but does fix the plain float scalar case at least 😅 

Feel free to add another change on top of this PR (I think you can?), or do a follow up. 

Either way I'll let you decide what you want to do with this PR.

Thanks!

martinbonnin:
> Feel free to add another change on top of this PR (I think you can?)

Will do, thank you so much for the contribution !

Issue #513, opened 2017-06-07
Add support for query batching


sav007:
Ref docs:
http://dev.apollodata.com/core/network.html#query-batching
https://dev-blog.apollodata.com/query-batching-in-apollo-63acfd859862

Radivarig:
Any news on this?

sav007:
No any updates, not sure what is the priority for this feature. Not all implementations of GraphQL server supports multiple queries execution support. And with client batching (via aliases) seems to me like a workaround.

Radivarig:
I've got a dynamic list of items that needs to be inserted to a table and have thought that mapping each item to a mutation would batch them up and send as an array in a single request. Is there some existing solution I'm missing?

Radivarig:
I'm using postgraphile, last week this plugin was released [postgraphile-plugin-nested-mutations](https://github.com/mlipscombe/postgraphile-plugin-nested-mutations) mentioned in [postgraphile#793](https://github.com/graphile/postgraphile/issues/793), it allows nested mutations, I can now create or connect existing foreign key items from a variable array, all in one query :)

mandrizzle:
I have a solution to this and believe this is possible using only OkHttp interceptors. I am unsure of what the requirements are other than:

 1. There should be a custom batch timing interval and max batch size
 2. Join the queries into a single JSON array for the request
 3. Expect a JSON array as a response and dispatch a response to each item in the same order they came in
 4. If # requests do not match # responses fail the entire batch

What I have not yet done but is still possible by adding custom headers to each request:
 1. Batch queries and mutations separately  
  2. Don't batch certain requests

If the above is all that is needed, I have a working solution and would be willing to submit a PR. If not, what else needs to be handled?

DJansse:
Any update on this topic / discussion? 
We're using Apollo GQL on Web, iOS and Android and are thinking about updating the server to support transport level batching of queries...

Kind regards

mandrizzle:
So from my original comment, I have that implementation working great however, it needs to create a new thread for each request which is not ideal. I am working on a refactor that now sets a custom `CallFactory` to reduce the number of threads required to just one. Just need to test it out and I'll keep y'all updated.

imankhdrpst:
hi @mandrizzle  , Did you have anything implemented, could you post them here ?

spawrksvrbo:
[...]

PR #3117, opened 2021-05-20
Implement Query Batching for `apollo-runtime`

joaquim-verges:
### Background

Hello Apollo team!

We, the Twitch Android team, have been using Apollo Android for 3+ years now, and have been very happy with it!

However, we recently ran into scalability issues in our app. Our most busy screen has exploded in features, where each feature is encapsulated and does its own GQL query. We've reached a point where that screen does 45 GQL queries (and therefore 45 HTTP calls) when it comes on screen 🤯 This is not only costly for the client, but also very costly for our backend services.

We explored grouping queries into larger ones, but we like the modularity and encapsulation that we have right now where each feature can do their fetching independently. We also know that our web team (who also uses Apollo) solved this problem by enabling batching for queries in the web Apollo client (and that our server supports it).

We decided to fork the Android library to implement the ability to batch multiple queries in a single HTTP call, similar to what the web client offers. We're about to test our implementation in production, and should will gather data in the next few weeks.

In the meantime, we thought it would be useful to push our implementation upstream and give back to the Apollo community!

### Implementation

We followed the approach that the web Apollo client took, where batching is achieved by a "poller" that checks periodically for queries to be batched, packages them into a single request body and sends them to the server. On response, the response body is parsed to extract each sub-response and forwarded back to the corresponding callers.

The implementation below is the minimal one that we needed to get that approach working, and we're looking for early feedback before writing unit tests, samples and later port it to the `apollo-runtime-kotlin` artifact.

Let us know what you think!


apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
```diff
@@ -0,0 +1,130 @@
...
+              scalarTypeAdapters
+          )
+      )
+    }
+    // Combine all request bodies into one
+    val batchRequestBody = RequestBody.create(
+        ApolloServerInterceptor.MEDIA_TYPE,
+        createBatchRequestJsonBody(queryRequestBodyList)
+    )
+    // TODO assumes all queries are POST (does GET even make sense for batch HTTP calls?)
```
joaquim-verges, line 58: We never use GET for queries, so didn't implement it yet, but wondering what you think since batch request bodies can be quite large, feels wrong to put it all in the URL. Wondering if batching should only be enabled when using POST queries.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
```diff
@@ -0,0 +1,130 @@
...
+        .header(ApolloServerInterceptor.HEADER_CONTENT_TYPE, ApolloServerInterceptor.CONTENT_TYPE)
+        .post(batchRequestBody)
+    // For the prototype, assumes all queries have the same header
+    val firstRequest = queryList.asSequence().map { it.request }.first()
+    for (header in firstRequest.requestHeaders.headers()) {
+      val value = firstRequest.requestHeaders.headerValue(header)
+      requestBuilder.header(header, value)
+    }
+
+    // TODO add http cache headers (does it make sense for batch HTTP calls?)
```
joaquim-verges, line 71: similarly, adding the HTTP cache headers felt weird since a batch can potentially be different between 2 runs, as the queries that will be batched is timing dependent. I guess one could get lucky and get the exact same batch (maybe for single query screens), in which case HTTP caching could come into play. Let us know what you think.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
```diff
@@ -0,0 +1,82 @@
...
+      return
+    }
+    // copy and clear the current queue
+    val queryList = ArrayList(queryQueue)
+    queryQueue.clear()
+    // split into batches
+    val batches = queryList.chunked(batchConfig.maxBatchSize)
+    logger.d("Executing ${queryList.size} Queries in ${batches.size} Batch(es)")
+    for (batch in batches) {
+      dispatcher.execute { createBatchHttpCall(batch).execute() }
```
martinbonnin, line 70: Might be worth eagerly sending the `maxBatchSize` operations as soon as they're enqueued to save some latency instead of doing it from the timer?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
```diff
@@ -0,0 +1,82 @@
...
+ * as defined by the BatchConfig.
+ */
+class BatchPoller(
+    private val serverUrl: HttpUrl,
+    private val httpCallFactory: Call.Factory,
+    private val scalarTypeAdapters: ScalarTypeAdapters,
+    private val logger: ApolloLogger,
+    private val batchConfig: BatchConfig,
+    private val dispatcher: Executor,
+    private val scheduledExecutorService: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor()
```
martinbonnin, line 29: Did you consider reusing the same `dispatcher` that is used above to save a thread? Mainly curious about the decision to go with a separate thread for something that doesn't do a lot of work?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
martinbonnin: Good question. Maybe throw an error if trying to enable `GET` + `batching` at the same time? My guess is that 99% of use cases will use `POST`, we can have the discussion when/if someone needs `GET` ?

joaquim-verges: yeah that's what we were thinking 👍

martinbonnin: `apollo-client` web doesn't support Batch + GET either: https://github.com/apollographql/apollo-client/blob/main/src/link/batch-http/batchHttpLink.ts#L108

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
joaquim-verges: Since that dispatcher is passed by the application, I didn't want to make assumptions on it. Thought it was safer to control this single thread that needs to run for a long period of time internally, and make sure it doesn't reduce the number of threads available in the application owned dispatcher.

martinbonnin: Makes sense 👍

martinbonnin: I would have made this an implementation details instead of a constructor property since it's always a `newSingleThreadScheduledExecutor` ultimately but I'll leave that decision to you.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
martinbonnin: Agreed, it doesn't make a lot of sense to have HTTP cache + query batching. Ideally we could detect that HTTP + batching is used and throw an error. Maybe based on `httpCachePolicy` ?

martinbonnin: And I'd say we need same kind of thing with AutoPersistedQueries too

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
```diff
@@ -0,0 +1,82 @@
...
+    queryQueue.remove(query)
+  }
+
+  private fun maybeExecuteBatchQuery(queryQueue: Queue<QueryToBatch>) {
+    if (queryQueue.isEmpty()) {
+      return
+    }
+    // copy and clear the current queue
+    val queryList = ArrayList(queryQueue)
+    queryQueue.clear()
```
martinbonnin, line 65: I think there's a race condition here were a `QueryToBatch` can be added from another thread after (or while) the queue is copied but before the queue is cleared making the `QueryToBatch` lost forever in the multithreaded limbo .

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
```diff
@@ -383,6 +422,7 @@ Response cachedHttpResponse(String cacheKey) throws IOException {
         .useHttpGetMethodForQueries(useHttpGetMethodForQueries)
         .useHttpGetMethodForPersistedQueries(useHttpGetMethodForPersistedQueries)
         .writeToNormalizedCacheAsynchronously(writeToNormalizedCacheAsynchronously)
+        .batchPoller(batchConfig.getBatchingEnabled() && canBeBatched ? batchPoller : null)
```
martinbonnin, line 425: Nitpicking, the `batchConfig.getBatchingEnabled()` seems uneeded as `batchPoller = null` in that case
```suggestion
        .batchPoller(canBeBatched ? batchPoller : null)
```

martinbonnin:
Awesome contribution 😃 ! That'll make a lot of users happy and close the [oldest open issue in this repo](https://github.com/apollographql/apollo-android/issues/513) 🎉 . 

A few comments but looks good overall 👍

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
joaquim-verges: we actually use APQ with batching and that works really well :) 

Yeah I'll do the same for file uploads, throw an exception if used with batching with a clear message

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
joaquim-verges: I was anticipating the unit tests here :D but will see if I can hide it

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
joaquim-verges: yeah I think I need to add some synchronization here

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
martinbonnin: > we actually use APQ with batching and that works really well :)

Nice. Good to know 👍 

> Yeah I'll do the same for file uploads, throw an exception if used with batching with a clear message

File Uploads are a bit weird and work with reflexion at the moment. It might be a bit awkward to detect when it's used. If it becomes too convoluted, I think it's fine to leave that as is.

joaquim-verges: actually, since we're only enabling batching for queries and not mutations, is file upload even possible with queries? (never used it so not sure)

apollo-runtime/src/main/java/com/apollographql/apollo/ApolloClient.java
joaquim-verges: good call

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
joaquim-verges: and looking at http caching, since you can make non-batched queries with the same `ApolloClient`, would be annoying to have to turn it off entirely if you still want your single queries to take advantage of it.

I'm thinking of just not adding the http cache headers for batch calls, and just silently ignore it. Not sure if I should log something to let the caller know?

joaquim-verges:
Adding some unit tests next.

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchHttpCall.kt
martinbonnin: > is file upload even possible with queries?

In theory I think yes.. I don't see anything preventing someone to pass an `Upload` custom scalar as a field argument. For an example a service that could sign your files using a server-side private key:

```
query GetSignature($file: Upload) {
  signature(file: $file)
}
```

In practice, I've never seen anyone do that yet though...

> we're only enabling batching for queries and not mutations

Would it make sense to enable batching for mutations? I don't think a theoretical reason to limit to queries? 

> looking at http caching, since you can make non-batched queries with the same ApolloClient, would be annoying to have to turn it off entirely if you still want your single queries to take advantage of it.

Right , makes sense 👍 

> I'm thinking of just not adding the http cache headers for batch calls, and just silently ignore it. Not sure if I should log something to let the caller know?

Could you do something per-query? 

```kotlin
    if (batchPoller != null) {
      if (httpCachePolicy != null) {
        throw new IllegalArgumentException("Using httpCachePolicy and batching at the same time does not make sense");
      }
      interceptors.add(new ApolloBatchingInterceptor(batchPoller));
```

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
```diff
@@ -0,0 +1,76 @@
...
+ */
+class BatchPoller(
+    private val batchConfig: BatchConfig,
+    private val dispatcher: Executor,
+    private val batchHttpCallFactory: BatchHttpCallFactory,
+    private val logger: ApolloLogger,
+    private val periodicJobScheduler: PeriodicJobScheduler = PeriodicJobSchedulerImpl()
+) {
+
+  private val queryQueue: ConcurrentLinkedQueue<QueryToBatch> = ConcurrentLinkedQueue()
```
martinbonnin, line 24: Since the handling of the queue is now synchronized, the queue can certainly be moved to a regular `LinkedList` ? Or `Dequeue` ? I'm not super familiar with the different Java Queues but it sounds like that'd have lower overhead?

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
```diff
@@ -0,0 +1,76 @@
...
+    if (!periodicJobScheduler.isRunning()) {
+      throw ApolloException("Trying to batch queries without calling ApolloClient.startBatchPoller() first")
+    }
+    val newQueueSize = synchronized(this) {
+      queryQueue.add(query)
+      logger.d("Enqueued Query: ${query.request.operation.name().name()} for batching")
+      queryQueue.size
+    }
+    // When the amount of queries in the queue reaches the max batch size, trigger the HTTP request without waiting for the batch interval
+    if (newQueueSize >= batchConfig.maxBatchSize) {
```
martinbonnin, line 51: Nit: there's a small chance that a new query is enqueued from another thread just before or after this check happens so `maybeExecuteBatchQuery()` will be called twice and the first time, might send `maxBatchSize + 1` instead of just `maxBatchSize`. I don't think this should be an issue? If it is, I guess there's no way around holding the lock around the enqueue + send with something like `enqueueAndMaybeExecuteBatchQuery`

apollo-integration/src/test/java/com/apollographql/apollo/internal/batch/BatchPollerTest.kt
```diff
@@ -0,0 +1,154 @@
...
+import com.apollographql.apollo.api.internal.ApolloLogger
+import com.apollographql.apollo.exception.ApolloException
+import com.apollographql.apollo.integration.interceptor.AllFilmsQuery
+import com.apollographql.apollo.interceptor.ApolloInterceptor
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mockito.mock
+import java.util.concurrent.TimeUnit
+
+class BatchPollerTest {
```
martinbonnin, line 14: NIce tests 👍

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
joaquim-verges: yeah I think `LinkedList` should be sufficient now

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
joaquim-verges: yeah I see what you mean. I could make `maybeExecuteBatchQuery()` not synchronized, so I can use it inside the synchronize block above, and synchronize at the call site in the periodic callback as well. What do you think?

apollo-integration/src/test/java/com/apollographql/apollo/internal/batch/BatchPollerTest.kt
joaquim-verges: next I want to add a test with some fixtures to test the serialization / deserialization

apollo-runtime/src/main/java/com/apollographql/apollo/internal/batch/BatchPoller.kt
martinbonnin: Sounds good to me 👍

joaquim-verges:
@martinbonnin feeling pretty good about the latest implementation and tests, let me know what you think!

martinbonnin:
Thanks for adding more tests! A couple of remaining comments:

* The `testMultipleQueryBatchingError` failed, I guess on timing issues. You can use a RxJava test subscriber (like [here](https://github.com/apollographql/apollo-android/blob/dd10f00cf3dc43b63213685e0b926dffd5c34162/apollo-integration/src/test/java/com/apollographql/apollo/HttpCacheTest.java#L142)) or more recent tests use coroutines (like [here](https://github.com/apollographql/apollo-android/blob/54389d13365e2340426dca340fbbe465ec5f212f/apollo-integration/src/test/java/com/apollographql/apollo/ApolloWatcherTest.kt#L385)) for synchronization.
* API: What about `ApolloQueryCall.canBeBatched(boolean canBeBatched)` instead of `ApolloClient.batchQuery`. This would be more consistent with the other per-query parameters like cacheHeaders, responseFetcher, etc..
* We did not settle on batching mutations. Just curious if you had any thoughts on that. I think it's ok to not include it in the first version as it can always be added later.

joaquim-verges:
> The testMultipleQueryBatchingError failed, I guess on timing issues. You can use a RxJava test subscriber (like here) or more recent tests use coroutines (like here) for synchronization.
Yeah I hesitated between controlling the time or testing the automatic trigger, I'll switch it to a more controlled test

> API: What about ApolloQueryCall.canBeBatched(boolean canBeBatched) instead of ApolloClient.batchQuery. This would be more consistent with the other per-query parameters like cacheHeaders, responseFetcher, etc..
Great suggestion, should have thought of it earlier!

> We did not settle on batching mutations. Just curious if you had any thoughts on that. I think it's ok to not include it in the first version as it can always be added later.
Still torn on that, it should work fine I think but since mutations are usually slower and more isolated one-offs in general, it feels a bit weird to mix them with queries? But yeah as you mentioned, can easily be enabled later if needed.

joaquim-verges:
@martinbonnin should I add `canBeBatched` only to `ApolloQueryCall.Builder` ? or also add one `ApolloQueryCall` directly ? I see a bunch of them are deprecated in favor of the builder one?

martinbonnin:
> @martinbonnin should I add canBeBatched only to ApolloQueryCall.Builder ? or also add one ApolloQueryCall directly ? I see a bunch of them are deprecated in favor of the builder one?

Yup, you're right 👍 . `ApolloQueryCall.Builder` is the way to go now because of [this](https://github.com/apollographql/apollo-android/pull/2434). It's more verbose than I'd ideally like to and `3.0` will change that but for consistency let's put everything in `ApolloQueryCall.Builder` in main.

joaquim-verges:
running metalava locally is changing way more signatures that it should for some reasons, mind running it for me?

martinbonnin (approved):
LGTM

martinbonnin:
Thanks @joaquim-verges for the awesome contribution!

joaquim-verges:
Glad I could help!

PR #3146, opened 2021-06-02
Fix overriding `canBeBatched` value when cloning `QueryCall` with `toBuilder()`

joaquim-verges:
Missed passing the `canBeBatched` value from a `RealApolloCall` object back to its builder when cloning with `toBuilder()`. Found this while integrating the new version in our app, where we do clone queries in that way in some cases.

Added a unit test for it as well.


martinbonnin:
These builders are verbose! Thanks for fixing this and including the unit test.

martinbonnin (approved):

PR #3147, opened 2021-06-03
More robust native tests

martinbonnin:
This fixes a case where a connection from an old test was closed and caused a crash in a subsequent test


martinbonnin:
Obsoleted by https://github.com/apollographql/apollo-android/pull/3148

Issue #3144, opened 2021-06-02
[dev-3.x] Exponential blowup in codegen size


martinraison:
**Summary**
The 3.x codegen can easily lead to huge codegen sizes, which makes compilation slow and can even break it entirely.

**Version**
3.x-dev12-SNAPSHOT

**Description**
I understand the 3.x codegen is not stable at all yet, but I thought I'd put this on your radar.

Here is the repro: https://github.com/martinraison/kmm-repro-graphql-codegen-blowup

Although the GraphQL is < 100 lines (schema + query), codegen size is > 10000 lines and grows exponentially with fragment nesting. Compilation breaks with the following error:

```
Cannot fit requested classes in a single dex file (# methods: 74868 > 65536)
```

Setting `codegenModels` to `"compat"` improves things somewhat, but `CatQuery_ResponseFields.kt` is still 8000+ lines, with a lot of class nesting, hence very long `*.class` file names which can easily exceed the filesystem limit.

Another curious symptom is that in the generated `CatQuery.kt`, `OPERATION_DOCUMENT` is extremely long. It looks like the fragments get duplicated many times.

I guess there are probably some trade-offs between ease-of-use and support for large/complex GraphQL schema, but it would be nice to have an option to avoid exponential code size blow-up. For the record, I've had this kind of issue in the past with the Apollo Client for iOS (which was solved there by setting the `mergeInFieldsFromFragmentSpreads` option to `false`), however Apollo Client for Android 2.x has never been a problem.

GraphQL schema:

```graphql
type Query {
    cat: Cat
}

type Cat {
    name: String
    field1: Cat
    field2: Cat
}
```

GraphQL query file (of course this one is a bit goofy, but I've noticed similar behavior with actual real-world business logic):

```graphql
query Cat {
    cat {
        ...CatFragment1
    }
}

fragment CatFragment1 on Cat {
    field1 {
        ...CatFragment2
    }
    field2 {
        ...CatFragment2
    }
}

fragment CatFragment2 on Cat {
    field1 {
        ...CatFragment3
    }
    field2 {
        ...CatFragment3
    }
}

fragment CatFragment3 on Cat {
    field1 {
        ...CatFragment4
    }
    field2 {
        ...CatFragment4
    }
}

fragment CatFragment4 on Cat {
    field1 {
        ...CatFragment5
    }
    field2 {
        ...CatFragment5
    }
}
fragment CatFragment5 on Cat {
    field1 {
        ...CatFragment6
    }
[...]

PR #3148, opened 2021-06-03
Introduce CompiledSelection 

martinbonnin:
This moves the cache reading code from a responseBased structure to an operationBased one. The field collection and merging will have to happen at runtime but the generated code should be much smaller for deep operations. 

See #3144


martinbonnin:
I merged this PR in `more-robust-native-test` when I intended it to be merged in `dev-3.x`. I just pushed the corresponding commit: https://github.com/apollographql/apollo-android/commit/4662855cf913899e39e92cd0979ea2887a22554c

PR #3150, opened 2021-06-04
Update get-started-multiplatform.mdx to use the new 'strictly' api

fabiosanto:
Update dependency version override as per gradle new api at 

https://docs.gradle.org/current/userguide/dependency_downgrade_and_exclude.html#sec:enforcing_dependency_version


martinbonnin (approved):
Thanks!

PR #3166, opened 2021-06-10
refactor: RxJava2/3 support

aoudiamoncef:
Hi @martinbonnin 

I tried to simplify RxJava integration code in this PR. 

I'm waiting for feedback

Thanks

### DIFF
--- a/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
+++ b/apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
@@ -12,16 +12,11 @@ import com.apollographql.apollo.internal.util.Cancelable;
 import io.reactivex.BackpressureStrategy;
 import io.reactivex.Completable;
 import io.reactivex.CompletableEmitter;
-import io.reactivex.CompletableOnSubscribe;
 import io.reactivex.Flowable;
 import io.reactivex.FlowableEmitter;
-import io.reactivex.FlowableOnSubscribe;
 import io.reactivex.Observable;
 import io.reactivex.ObservableEmitter;
-import io.reactivex.ObservableOnSubscribe;
 import io.reactivex.Single;
-import io.reactivex.SingleEmitter;
-import io.reactivex.SingleOnSubscribe;
 import io.reactivex.annotations.CheckReturnValue;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.Exceptions;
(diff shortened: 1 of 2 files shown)


martinbonnin:
Hi @aoudiamoncef ! 

Thanks for sending this. Do you mind giving a high level overview of the changes? I can see some lambdas being used instead of anonymous classes. I _think_ that's ok as we have a minimum of Java8+ and Android 15+. Mind be worth double checking that the minimum AGP version we support actually has desugaring for lambdas.

Besides lamndas, is there anything else?

aoudiamoncef:
I roll backed changes from RxJava2

Lamdas for [RxJava](https://github.com/ReactiveX/RxJava) 3 are supported(Java 8+)

Replaced  `call.clone` deperectated method with `call.toBuilder.build`

Added some `@NotNull` annotations.

martinbonnin:
Sorry got side tracked!

Not sure I understand why the lambda situation would be different on RxJava2 compared to RxJava3? The two files almost have the same content today and I'd like to keep it that was if possible.

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
```diff
@@ -52,7 +52,7 @@ private Rx2Apollo() {
   public static <T> Observable<Response<T>> from(@NotNull final ApolloQueryWatcher<T> watcher) {
     checkNotNull(watcher, "watcher == null");
     return Observable.create(new ObservableOnSubscribe<Response<T>>() {
-      @Override public void subscribe(final ObservableEmitter<Response<T>> emitter) throws Exception {
+      @Override public void subscribe(@NotNull final ObservableEmitter<Response<T>> emitter) {
```
martinbonnin, line 55: +1 for `@NotNull`
+1 for removing the `throws Exception`, I guess it was carried over from the super interface but it's never used in this implementation.

apollo-rx2-support/src/main/java/com/apollographql/apollo/rx2/Rx2Apollo.java
```diff
@@ -89,8 +89,8 @@ private Rx2Apollo() {
     checkNotNull(call, "call == null");
 
     return Observable.create(new ObservableOnSubscribe<Response<T>>() {
-      @Override public void subscribe(final ObservableEmitter<Response<T>> emitter) throws Exception {
-        ApolloCall<T> clone = call.clone();
+      @Override public void subscribe(@NotNull final ObservableEmitter<Response<T>> emitter) {
+        ApolloCall<T> clone = call.toBuilder().build();
```
martinbonnin, line 93: 👍

martinbonnin:
Looks good to me! Thinking more about this, I think it's fine to use lambdas everywhere. We support JDK8+ and on Android, AGP 3.0.0 [can desugar lambdas-produced bytecode](https://developer.android.com/studio/write/java8-support). Let me know if I'm missing something.

aoudiamoncef:
Cool, if there is no issue with Android, let's add Lambdas to Rxjava2.

martinbonnin:
> Cool, if there is no issue with Android, let's add Lambdas to Rxjava2.

👍  I don't think there's an issue. Also Android can use RxJava3 as well so there's no reason not to have the same code in both.

martinbonnin:
For the record, that's the current diff between both files:

```diff
<  * The Rx2Apollo class provides methods for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to RxJava 2
---
>  * The Rx3Apollo class provides methods for converting ApolloCall, ApolloPrefetch and ApolloWatcher types to RxJava 3
31c31
< public class Rx2Apollo {
---
> public class Rx3Apollo {
33c33
<   private Rx2Apollo() {
---
>   private Rx3Apollo() {
218c218
<     emitter.setDisposable(getRx2Disposable(cancelable));
---
>     emitter.setDisposable(getRx3Disposable(cancelable));
222c222
<     emitter.setDisposable(getRx2Disposable(cancelable));
---
>     emitter.setDisposable(getRx3Disposable(cancelable));
226c226
<     emitter.setDisposable(getRx2Disposable(cancelable));
---
>     emitter.setDisposable(getRx3Disposable(cancelable));
229c229
<   private static Disposable getRx2Disposable(final Cancelable cancelable) {
---
>   private static Disposable getRx3Disposable(final Cancelable cancelable) {
```

martinbonnin (approved):

PR #3208, opened 2021-07-02
JavaScript

Pitel:
Apollo support for Kotlin/JS

:warning: Currently in heavy WIP state, consulting with @martinbonnin on Slack.

### DIFF
--- a/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
+++ b/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
@@ -1,0 +1,19 @@
+package com.apollographql.apollo3.mockserver
+
+actual class MockServer {
+  actual fun url(): String {
+    TODO()
+  }
+
+  actual fun enqueue(mockResponse: MockResponse) {
+    TODO()
+  }
+
+  actual fun takeRequest(): MockRecordedRequest {
+    TODO()
+  }
+
+  actual fun stop() {
+    TODO()
+  }
+}
--- a/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/OkHttpEngine.kt
+++ b/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/OkHttpEngine.kt
@@ -1,0 +1,10 @@
+package com.apollographql.apollo3.network.http
+
+import com.apollographql.apollo3.api.http.HttpRequest
+import com.apollographql.apollo3.api.http.HttpResponse
+
+actual class DefaultHttpEngine actual constructor(connectTimeoutMillis: Long, readTimeoutMillis: Long) : HttpEngine {
+  override suspend fun execute(request: HttpRequest): HttpResponse {
+    TODO()
+  }
+}
--- a/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
+++ b/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
@@ -1,0 +1,10 @@
+package com.apollographql.apollo3.network.ws
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+
+@OptIn(ExperimentalCoroutinesApi::class)
+actual class DefaultWebSocketEngine : WebSocketEngine {
+  override suspend fun open(url: String, headers: Map<String, String>): WebSocketConnection {
+    TODO()
+  }
+}
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,7 +18,7 @@ def versions = [
     moshix                : '0.11.2',
     okHttp4               : '4.9.0',
     okHttp                : '3.12.11', // Keep this to keep supporting older Android devices
-    okio                  : '2.10.0',
+    okio                  : '3.0.0-alpha.6',
     rxjava                : '2.2.20',
     rxjava3               : '3.0.7',
     rxandroid             : '2.0.1',
--- a/apollo-mockserver/build.gradle.kts
+++ b/apollo-mockserver/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
   kotlin("multiplatform")
 }
 
-configureMppDefaults(withJs = false)
+configureMppDefaults()
 
 kotlin {
   sourceSets {
--- a/apollo-runtime/build.gradle.kts
+++ b/apollo-runtime/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
   kotlin("multiplatform")
 }
 
-configureMppDefaults(withJs = false)
+configureMppDefaults()
 
 kotlin {
   sourceSets {
--- a/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/internal/NonMainWorker.kt
+++ b/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/internal/NonMainWorker.kt
@@ -1,0 +1,7 @@
+package com.apollographql.apollo3.internal
+
+actual class NonMainWorker {
+  actual suspend fun <R> doWork(block: () -> R): R {
+    TODO()
+  }
+}
--- a/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
+++ b/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
@@ -1,0 +1,15 @@
+package com.apollographql.apollo3.internal
+
+import kotlinx.coroutines.CoroutineDispatcher
+
+actual fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher {
+  TODO()
+}
+
+actual class WebSocketDispatcher {
+  actual val coroutineDispatcher: CoroutineDispatcher = TODO()
+
+  actual fun dispose() {
+    TODO()
+  }
+}
--- a/apollo-testing-support/build.gradle.kts
+++ b/apollo-testing-support/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
   kotlin("multiplatform")
 }
 
-configureMppDefaults(withJs = false)
+configureMppDefaults()
 
 kotlin {
   sourceSets {
--- a/apollo-testing-support/src/jsMain/kotlin/com/apollographql/apollo3/testing/readFileJs.kt
+++ b/apollo-testing-support/src/jsMain/kotlin/com/apollographql/apollo3/testing/readFileJs.kt
@@ -1,0 +1,9 @@
+package com.apollographql.apollo3.testing
+
+actual fun readFile(path: String): String {
+  TODO()
+}
+
+actual fun checkFile(actualText: String, path: String) {
+  TODO()
+}
--- a/apollo-testing-support/src/jsMain/kotlin/com/apollographql/apollo3/testing/runWithMainLoopJs.kt
+++ b/apollo-testing-support/src/jsMain/kotlin/com/apollographql/apollo3/testing/runWithMainLoopJs.kt
@@ -1,0 +1,16 @@
+package com.apollographql.apollo3.testing
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlin.coroutines.CoroutineContext
+
+actual fun <T> runBlocking(context: CoroutineContext, block: suspend CoroutineScope.() -> T): T {
+  TODO()
+}
+
+actual fun <T> runWithMainLoop(context: CoroutineContext, block: suspend CoroutineScope.() -> T): T {
+  TODO()
+}
+
+actual val MainLoopDispatcher: CoroutineDispatcher = Dispatchers.Default
--- a/build-logic/src/main/kotlin/Mpp.kt
+++ b/build-logic/src/main/kotlin/Mpp.kt
@@ -13,7 +13,7 @@ fun Project.configureMppDefaults(withJs: Boolean = true) {
     jvm()
 
     if (withJs) {
-      js {
+      js(BOTH) {
         useCommonJs()
         browser()
         nodejs()


Pitel:
@martinbonnin

```
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/BookTripMutation.kt: (22, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/BookTripMutation.kt: (24, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/CancelTripMutation.kt: (22, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/CancelTripMutation.kt: (24, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/LaunchDetailsQuery.kt: (22, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/fullstack/com/apollographql/apollo3/integration/fullstack/LaunchDetailsQuery.kt: (24, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterDetailsQuery.kt: (23, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterDetailsQuery.kt: (25, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterNameByIdQuery.kt: (21, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterNameByIdQuery.kt: (23, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterWithBirthDateQuery.kt: (22, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/CharacterWithBirthDateQuery.kt: (24, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/StarshipByIdQuery.kt: (22, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/StarshipByIdQuery.kt: (24, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/UpdateReviewMutation.kt: (23, 3): JavaScript name (id) generated for this declaration clashes with another declaration: fun id(): String
e: /home/pitel/apollo-android/composite/tests/integration-tests-kotlin/build/generated/source/apollo/normalizer/com/apollographql/apollo3/integration/normalizer/UpdateReviewMutation.kt: (26, 3): JavaScript name (id) generated for this declaration clashes with another declaration: val id: String
```

It look like you'll have to change someting in generating code, because the `id` is clashing.

martinbonnin:
Mmmm I don't think this changed recently but this is going to be a problem nonetheless... The problem can be reduced to:

```kotlin
class MyOperation(val id: String) {
  fun id(): String {
    return ""
  }
}
```

My understanding is that JS classes cannot have properties with the same name as a function? 

If that's the case, we can decrease the occurence of name clashes with `@JsName`:

```kotlin
class MyOperation(val id: String) {
  @JsName("operationId")
  fun id(): String {
    return ""
  }
}
```

To 100% remove nameclashes, it would then either be:
1. using aliases in queries
2. or escape property names in the codegen

Pitel:
@martinbonnin Do you have some ETA on the fix?

martinbonnin:
@Pitel pretty hard to tell at the moment but I think you can workaround by changing the name of [Operation.id](https://github.com/apollographql/apollo-android/blob/437bc50f4a893b6a06b8cb992fb8db1fe62b3ad2/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Operation.kt#L34) in Js:

```
  /**
   * An unique identifier for the operation. 
   */
  @JsName("operationId")
  fun id(): String
```

That should do until we find a more robust way to escape name clashes.

Pitel:
@martinbonnin Thanks for the link. I didn't know where to look.

Now it builds, and the tests are failing, as expected.

I'll probably have to modify the existing tests, because there is no `runBlocking` in JS. So I'll probably do multiplatform `runTest` method (using `runBlocking` on JVM and Apple, and Promise on JS), which will have to wrap all tests that were using the `runBlocking` and remove the `runBlocking`, because I simply can't do that in JS.

martinbonnin:
We have [runWithMainLoop](https://github.com/apollographql/apollo-android/blob/0254755bb2b0a516ca3b951d7e6a067d206f9296/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/runWithMainLoop.kt#L12) that is doing some specific stuff for Apple so it might make sense to add Js support there as well (`apollo-testing-support` module) ?

apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -0,0 +1,38 @@
...
+  private val responseQueue = mutableListOf<MockResponse>()
+
+  private val server = http.createServer { req, res ->
+    println(req)
+    res.writeHead(200)
+    res.end("Hello, World!")
+  }.listen(PORT)
+
+  actual fun url(): String {
+    //TODO Use `client.address()` but it might return null, before the server is listening. So this will have to be suspend fun
```
martinbonnin, line 14: I think `server.address()` is what we want? The url shouldn't depend on any client being connected. If getting a server address takes some time, I think making it suspend is ok. Most of the tests are run from a coroutine.

build.gradle.kts
```diff
@@ -64,6 +64,7 @@ subprojects {
   repositories {
     google()
     mavenCentral()
+    jcenter() // https://github.com/Kotlin/kotlinx-nodejs/issues/16
```
martinbonnin, line 67:
```suggestion
    jcenter {
      content {
        // https://github.com/Kotlin/kotlinx-nodejs/issues/16
        includeModule("org.jetbrains.kotlinx", "kotlinx-nodejs")
      }
    }
```

composite/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ subprojects {
   repositories {
     google()
     mavenCentral()
+    jcenter() // https://github.com/Kotlin/kotlinx-nodejs/issues/16
```
martinbonnin, line 23:
```suggestion
    jcenter {
      content {
        // https://github.com/Kotlin/kotlinx-nodejs/issues/16
        includeModule("org.jetbrains.kotlinx", "kotlinx-nodejs")
      }
    }
```

gradle/dependencies.gradle
```diff
@@ -18,12 +18,13 @@ def versions = [
     moshix                : '0.11.2',
     okHttp4               : '4.9.0',
     okHttp                : '3.12.11', // Keep this to keep supporting older Android devices
-    okio                  : '2.10.0',
+    okio                  : '3.0.0-alpha.8',
```
martinbonnin, line 21: Can you remind me why okio `3.0.0` is required for JS? It's going to be a problem if we want to ship a stable `3.0` release for the JVM.

gradle/dependencies.gradle
```diff
@@ -105,7 +107,10 @@ ext.dep = [
     kspGradlePlugin: "com.google.devtools.ksp:symbol-processing-gradle-plugin:1.5.20-1.0.0-beta04",
     kotlinxdatetime: "org.jetbrains.kotlinx:kotlinx-datetime:0.1.1",
     kotlinxserializationjson: "org.jetbrains.kotlinx:kotlinx-serialization-json:1.1.0",
-    atomicfu: "org.jetbrains.kotlinx:atomicfu:0.15.1"
+    atomicfu: "org.jetbrains.kotlinx:atomicfu:0.15.1",
+    ktor                  : [
+            clientJs: "io.ktor:ktor-client-js:$versions.ktor"
```
martinbonnin, line 112: Ultimately, I think we'll want to use `fetch` or some other API directly to save the dependency here. Can definitely be done in a later step though so 👍

apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
Pitel: Oops, my bad. It of course should be `server.address()`. And yes, it returns `null` before it binds to the port.

One nice thing with this being async is that we won't need to specify a port; Node.js will just pick one that's available.

gradle/dependencies.gradle
Pitel: Because since 3.0.0-alpha.6, they support the JS IR backend.

https://github.com/square/okio/blob/master/CHANGELOG.md#version-300-alpha6

gradle/dependencies.gradle
Pitel: Have considered using Ktor for all platforms? Because then, you won't need to write all those platform specific HTTP/WS classes, because Ktor already does that for you.

Yeah, I've noticed the "API" you have is built mostly around OkHttp, si it won't be an easy task to change it everywhere.

Pitel:
@martinbonnin Regarding the `runWithMainLoop`, will you be able to verify that my changes doesn't break Apple tests? Because I don't have a Mac, so I can't check that.

apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
martinbonnin: Yep, that'll come in handy 👍

gradle/dependencies.gradle
martinbonnin: Alright, lets cross fingers that by the time Apollo Android hits stable, Okio will be too. Else we'll have to either:
* revert to the legacy JS compiler
* or find a way to use okio 2.x for JVM and 3.x for JS

gradle/dependencies.gradle
martinbonnin: We could have a `KtorHttpEngine`, maybe in a separate artifact but I'd rather not add it as the default. 

Also, on Android/JVM, OkHttp is super widespread so users can keep their token interceptors, logging, etc...

apollo-testing-support/src/appleMain/kotlin/com/apollographql/apollo3/testing/runWithMainLoopApple.kt
```diff
@@ -20,6 +20,10 @@ import platform.darwin.dispatch_get_main_queue
 import platform.darwin.dispatch_time
 import kotlin.coroutines.CoroutineContext
 
+actual fun runTest(block: suspend () -> Unit) {
+  kotlinx.coroutines.runBlocking { block() }
```
martinbonnin, line 24: That doesn't work for Apple tests because they need a `CFRunLoop` to be alive. There's more info about it there: https://github.com/apollographql/apollo-android/blob/dev-3.x/design-docs/Threading.md

Overall, I don't think we need a new `runTest`. Could we just rename `runWithMainLoop` to `runTest` and implement it on JS ? (and leave JVM, Apple untouched?)

martinbonnin:
> @martinbonnin Regarding the runWithMainLoop, will you be able to verify that my changes doesn't break Apple tests?

I left a comment above. Apple tests really need the `runWithMainLoop` trick or else they will hang forever

Pitel:
@martinbonnin `./gradlew -p composite :tests:integration-tests-kotlin:jsNodeTest` are all passing! :tada:

* But `WriteToCacheAsynchronouslyTest[jvm]` started failing. Other JVM tests seems ok.
* The `jsNodeTest` is hanging at the end. I have to `killall node` to recover and get test reports from gradle. I must investigate why.
* I hope Apple tests are ok.
* `jsBrowserTest` is not working. Obviously, because we can't have server in browser. I'll have to find a way to selectively ignore those tests which requires MockServer when they are run in browser.

martinbonnin:
Nice 🤩 . I'll take a deeper look later today. 

Early thoughts: I think we can remove the `jsBrowser` integration tests for now. If we find bugs where the node implementation is different from the browser one (that seems unlikely but you never know), we can run a server outside the Gradle process. It'll be a bit more painful but certainly doable if we really need to.

Pitel:
@martinbonnin Hi, sorry for inactivity, I was on family vacation.

Anyways, after merging current `dev3.x` branch, I noticed there is no longer `composite` module. Digging through commits, I figured I should now run `./gradlew -p tests jsNodeTest`. Is that correct? That would also mean now have to pass even more tests, not just the `integration-tests`, right?

martinbonnin:
>  now run ./gradlew -p tests jsNodeTest. Is that correct?

Yup. Not 100% sure on the `jsNodeTest`. The CI runs `allTests` but since we can certainly ignore `browserTest`, `jsNodeTest` is certainly enough

> That would also mean now have to pass even more tests, not just the integration-tests, right?

It's the same as before really. Only removing one layer of directories. `integration-tests` are still there as the general cases and other modules are for more specific cases.

Pitel:
@martinbonnin Ok, I've fixed the tests, and there is one I can't make work.

`models-response-based` > `ParseResponseBodyTest` > `operationJsonWriter`.

Problem is with `surfaceWater` values. It expects integer (`40`), but it gets float (`40.0`). Any idea why is that? Maybe another JS quirk?

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
```diff
@@ -15,7 +15,7 @@ fun parseHeader(line: String): Pair<String, String> {
   return line.substring(0, index).trim() to line.substring(index + 1, line.length).trim()
 }
 
-class MockRecordedRequest(
+data class MockRecordedRequest(
```
martinbonnin, line 18: Why is this needed ? I'm generally opposed to data classes in public APIs unless there's a strong reason to do so (https://jakewharton.com/public-api-challenges-in-kotlin/).

martinbonnin:
```
Problem is with surfaceWater values. It expects integer (40), but it gets float (40.0). Any idea why is that? Maybe another JS quirk?
```

This is because `toString()` behaves differently:

* JVM: `40.0.toString() -> "40.0"`
* JS: `40.0.toString() -> "40"`

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/OkHttpEngine.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.ktor.client.call.receive
+import io.ktor.client.engine.js.Js
+import io.ktor.client.request.header
+import io.ktor.client.request.request
+import io.ktor.http.HttpHeaders
+import io.ktor.util.flattenEntries
+import okio.Buffer
+import okio.ByteString.Companion.toByteString
+
+actual class DefaultHttpEngine actual constructor(connectTimeoutMillis: Long, readTimeoutMillis: Long) : HttpEngine {
```
martinbonnin, line 18: Change the name of this file to `KtorHttpEngine` ?

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo3.network.ws
```
martinbonnin, line 1: Same here, change to `DefaultWebSocketEngine.kt`?

apollo-testing-support/src/appleMain/kotlin/com/apollographql/apollo3/testing/runWithMainLoopApple.kt
```diff
@@ -26,23 +26,26 @@ import kotlin.coroutines.CoroutineContext
...
+    before: suspend CoroutineScope.() -> Unit,
+    after: suspend CoroutineScope.() -> Unit,
+    block: suspend CoroutineScope.() -> Unit,
+) {
+  GlobalScope.launch(context + MainLoopDispatcher) {
+    before()
+    try {
       block()
+    } finally {
+      after()
```
martinbonnin, line 40: Can we keep a simple `runTest(block: suspend CoroutineScope.() -> T): T` ? It's important that exceptions happening inside the block are forwarded to the caller and not ignored. Getting back a return value might be useful too. And it's easy to implement before/after at the callsite:

```kotlin
  private fun runTestWithMockServer(block: (MockServer) -> Unit) {
    runTest {
      val mockServer = MockServer()
      try {
        block(mockServer)
      } finally {
        mockServer.stop()
      }
    }
  }
```

Pitel: I don't think we can.

If you want to call suspending functions from the test, then the test method in JS must return a `Promise`. There is no other way. Notice that in Js implementation, the return value is being re-casted to `dynamic`, to kinda make the compiler happy and allowing me to return that `Promise` instead of `Unit`.

Regarding the `before` and `after` parameters, there is another issue with asynchronicity. Because we've made `MockServer.url()` suspending, also the `@BeforeTest` method setting up the server and client must be suspending. So in JS world, it must return a `Promise`. But if it does, the ordering of things is different then one may expect. Because the testing framework does not wait for that promise to resolve before starting the actual test. See it [here](https://youtrack.jetbrains.com/issue/KT-19813#focus=Comments-27-3536253.0-0).

So that's the reasons why there are those 2 parameters, and why the `runTest` method looks how it looks.

martinbonnin: I understand `@BeforeTest` is not working in JS but `suspend fun` are still executed in order, right? If `runTest` is used to start the coroutine, then I think before/after can be run from this coroutine?

```kotlin
actual fun runTest(
    context: CoroutineContext,
    block: suspend CoroutineScope.() -> Unit,
): dynamic = GlobalScope.promise(context) {
  return block()
} 
```

And in the test:

```kotlin
  private fun runTestWithMockServer(block: suspend (MockServer) -> Unit) = runTest {
    val mockServer = MockServer()
    try {
      block(mockServer)
    } finally {
      mockServer.stop()
    }
  }
```

Is there a specific reason this wouldn't work? That means we can't get a return value from `runTest` which I guess is Ok. 

Side note: I wonder how the runner is waiting for the test to complete before launching other tests now :). Since there's no blocking call in JS, there has to be another way of scheduling all the tests

PR #3214, opened 2021-07-05
Appsync fixes

### LINKED ISSUE #3205: [3.x] AppSync WS protocol issue
**Summary**
The `AppSyncWsProtocol` seems to not work correctly out-of-the-box.

**Version**
v3.0.0-dev13

**Description**
When trying to connect to AppSync using the `AppSyncWsProtocol` I bumped into a couple of issues:
* The default frame type is `WsFrameType.Binary`, but for me, it only works with `WsFrameType.Text`
* The `payload.data` field needs to be stringified JSON, instead of just JSON
* There seams to not be a way yet to generate the `header` query string parameter (as what was present in: https://github.com/apollographql/apollo-android/blob/dev-3.x/deprecated-apollo-runtime-common/src/jvmMain/kotlin/com/apollographql/apollo3/subscription/AppSync.kt)

To mitigate the second issue I am using a CustomWsProtocol, that serializes the JSON using `kotlinx.serialization`:

```kotlin
class CustomAppSyncWsProtocol(private val authorization: Map<String, Any?>) : WsProtocol by AppSyncWsProtocol(authorization, WsFrameType.Text) {
    override fun <D : Operation.Data> operationStart(request: ApolloRequest<D>): Map<String, Any?> {
        return mapOf(
            "type" to "start",
            "id" to request.requestUuid.toString(),
            "payload" to mapOf(
                "data" to Json.encodeToString(AnySerializer, DefaultHttpRequestComposer.composePayload(request)),
                "extensions" to authorization
            )
        )
    }
}
```

I am using the 3.0.0 version because of my desire to use AppSync in Kotlin Multiplatform, the 2.x version implementation of AppSync seems to be JVM only. As far as I could tell this is the only KMP library with AppSync support :)

I am willing to make a PR for this change, however, I might need some help with figuring out how to best to the JSON serialization neatly without adding an extra dependency

chrisvis:
Closes #3205

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/UrlEncode.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/UrlEncode.kt
@@ -15,11 +15,11 @@ internal fun String.urlEncode(
     when {
       it in RESERVED_CHARS -> append(it.percentEncode())
       spaceToPlus && it == ' ' -> append('+')
-      else -> append(it.percentEncode())
+      else -> append(it)
     }
   }
 }
 
 private fun Char.percentEncode(): String {
-  return "%${toInt().toString(16)}"
+  return "%${toInt().toString(16)}".uppercase()
 }
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/AppSync.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/AppSync.kt
@@ -1,0 +1,46 @@
+package com.apollographql.apollo3.network
+
+import com.apollographql.apollo3.api.http.DefaultHttpRequestComposer.Companion.appendQueryParameters
+import com.apollographql.apollo3.api.internal.json.BufferedSinkJsonWriter
+import com.apollographql.apollo3.api.internal.json.Utils
+import okio.Buffer
+
+object AppSync {
+
+  /**
+   * Helper method that builds the final URL. It will append the authorization and payload arguments as query parameters.
+   * This method can be use for both the HTTP URL as well as the WebSocket URL
+   *
+   * Example:
+   * ```
+   * buildUrl(
+   *   baseUrl = "https://example1234567890000.appsync-realtime-api.us-east-1.amazonaws.com/graphql",
+   *   // This example uses an API key. See the AppSync documentation for information on what to pass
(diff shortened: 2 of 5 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/UrlEncode.kt
```diff
@@ -15,11 +15,11 @@ internal fun String.urlEncode(
     when {
       it in RESERVED_CHARS -> append(it.percentEncode())
       spaceToPlus && it == ' ' -> append('+')
-      else -> append(it.percentEncode())
+      else -> append(it)
```
martinbonnin, line 18: Yikes, thanks for catching this!

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/AppSync.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.apollographql.apollo3.api.http.DefaultHttpRequestComposer.Companion.appendQueryParameters
+import com.apollographql.apollo3.api.internal.json.BufferedSinkJsonWriter
+import com.apollographql.apollo3.api.internal.json.Utils
+import okio.Buffer
+
+object AppSync {
+
+  /**
+   * Helper method that builds the final URL. It will append the authorization and payload arguments as query parameters.
+   * This method can be use for both the HTTP URL as well as the WebSocket URL
```
martinbonnin, line 12:
```suggestion
   * This method can be used for both the HTTP URL as well as the WebSocket URL
```

martinbonnin: Also, I'm still confused by `https://` vs `wss://` is there a reason to use one over the other? If not we could add a note about it? 

```
Note: https:// and wss:// urls (and http:// and ws:// for unsecure connections) are treated in the same way be the underlying code. Using one host or the other does not matter.
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/UrlEncode.kt
chrisvis: No problem. I was in doubt to put it in this PR directly, but felt to small enough

martinbonnin (approved):
Thanks a lot!

PR #3220, opened 2021-07-06
[3.x - Runtime] Write to cache async per request

### LINKED ISSUE #3219: [3.x] The new apollo-runtime doesn't support per-request writeToCacheAsync flag
With the previous runtime (now named "deprecated-apollo-runtime"), we had the capability to override this flag per-request based. This is useful for example for prefetch requests, where we want to know where the response was merged into the cache.

We could use ApolloRequest's ExecutionContexts (CacheInput) to forward this boolean to the ApolloCacheInterceptor.

olivierg13:
Fix #3219

Allows to override the global client "writeToCacheAsynchronously" boolean by providing a boolean through ApolloRequest's ExecutionContexts (CacheInput).


martinbonnin (approved):
Thanks!

PR #3225, opened 2021-07-07
Edits to landing page and getting started

StephenBarlow:
@martinbonnin here's a first PR to get my feet wet! Not being up on JVM-land lately, I made a few assumptions about filenames and conventions in these edits that might be inaccurate, but I'll get better at this stuff with each article 😄

### DIFF
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -18,7 +18,7 @@ Hello! This article walks you through adding Apollo Android to your project.
 
 ## Add runtime dependencies
 
-```kotlin:title=build.gradle
+```kotlin:title=build.gradle[.kts]
 dependencies {
   // The core runtime dependencies
   implementation("com.apollographql.apollo3:apollo-runtime:$version")
@@ -27,7 +27,7 @@ dependencies {
 
 Or for a multiplatform project:
 
-```kotlin:title=build.gradle
+```kotlin:title=build.gradle[.kts]
 kotlin {
   sourceSets {
     val commonMain by getting {
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -59,7 +59,7 @@ val apolloClient = ApolloClient(serverUrl = "https://your.domain/graphql")
 Now you can execute your defined query with `ApolloClient` like so:
 
 ```kotlin
-// You can replace `runBlocking {}` with `viewModelScope.launch {}`
+// Replace `runBlocking {}` with `viewModelScope.launch {}`
 // or any other coroutine scope builder
 runBlocking {
   val response = try {
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -1,5 +1,6 @@
 ---
-title: Get started
+title: Get started with Apollo Android
+sidebar_title: Get started
 ---
 
 import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
@@ -9,13 +10,15 @@ import AddPlugin from "./shared/add-plugin.mdx"
 import DownloadSchema from "./shared/download-schema.mdx"
 import AddQuery from "./shared/add-query.mdx"
 
+Hello! This article walks you through adding Apollo Android to your project.
+
 ## Add the Gradle plugin
 
 <AddPlugin />
 
-## Add the runtime dependencies
+## Add runtime dependencies
 
-```kotlin
+```kotlin:title=build.gradle
 dependencies {
   // The core runtime dependencies
   implementation("com.apollographql.apollo3:apollo-runtime:$version")
@@ -24,7 +27,7 @@ dependencies {
 
 Or for a multiplatform project:
 
-```kotlin
+```kotlin:title=build.gradle
 kotlin {
   sourceSets {
     val commonMain by getting {
@@ -46,18 +49,18 @@ kotlin {
 
 ## Execute your query
 
-Create an `ApolloClient`:
+In your application code, create an instance of `ApolloClient`, providing your GraphQL server's URL where indicated:
 
 ```kotlin
-// First, create an `ApolloClient`
 // Replace the serverUrl with your GraphQL endpoint
 val apolloClient = ApolloClient(serverUrl = "https://your.domain/graphql")
 ```
 
-Apollo Android uses coroutines to handle concurrency. If you're using Java, see [RxJava2](usage/rxjava2) for an alternative way to execute your queries.
+Now you can execute your defined query with `ApolloClient` like so:
 
-```
-// Replace `runBlocking {}` by `viewModelScope.launch {}` or any other coroutine scope builder
+```kotlin
+// You can replace `runBlocking {}` with `viewModelScope.launch {}`
+// or any other coroutine scope builder
 runBlocking {
   val response = try {
     apolloClient.query(LaunchDetailsQuery(id = "83"))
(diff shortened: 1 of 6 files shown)


docs/source/get-started.mdx
```diff
@@ -9,13 +10,15 @@ import AddPlugin from "./shared/add-plugin.mdx"
...
+
 ## Add the Gradle plugin
 
 <AddPlugin />
 
-## Add the runtime dependencies
+## Add runtime dependencies
 
-```kotlin
+```kotlin:title=build.gradle
```
martinbonnin, line 21:
```suggestion
```kotlin:title=build.gradle[.kts]
```
Gradle now supports both Groovy and Kotlin scripts. Most of the times, the snippets work out of the box for both. In some cases there might be some differences but not in this case so I don't think we need a `<MultiCodeBlock>` ?

docs/source/get-started.mdx
```diff
@@ -24,7 +27,7 @@ dependencies {
 
 Or for a multiplatform project:
 
-```kotlin
+```kotlin:title=build.gradle
```
martinbonnin, line 30:
```suggestion
```kotlin:title=build.gradle[.kts]
```

docs/source/get-started.mdx
```diff
@@ -46,18 +49,18 @@ kotlin {
...
 val apolloClient = ApolloClient(serverUrl = "https://your.domain/graphql")
 ```
 
-Apollo Android uses coroutines to handle concurrency. If you're using Java, see [RxJava2](usage/rxjava2) for an alternative way to execute your queries.
+Now you can execute your defined query with `ApolloClient` like so:
 
-```
-// Replace `runBlocking {}` by `viewModelScope.launch {}` or any other coroutine scope builder
+```kotlin
+// You can replace `runBlocking {}` with `viewModelScope.launch {}`
```
martinbonnin, line 62: Out of curiosity, what are the reasons to choose `"You can replace"` vs the more imperative `"Replace"`? I never know which one to use.

docs/source/index.md
```diff
@@ -1,13 +1,13 @@
 ---
 title: Introduction
-description: A strongly-typed, caching GraphQL client for Android and Kotlin multiplatform, written in Kotlin
+description: A strongly-typed, caching GraphQL client for the JVM
```
martinbonnin, line 3: Apollo Android isn't limited to the JVM. It was already the case with 2.x and even more so with 3.x where the core runtime and cache has been rewritten in Kotlin and now supports iOS/MacOS in addition to Android.

docs/source/index.md
```diff
@@ -1,13 +1,13 @@
...
+Apollo Android executes queries and mutations against a GraphQL server and returns results as query-specific Kotlin types. This means you don't have to deal with parsing JSON, or passing around `Map`s and making clients cast values to the right type manually. You also don't have to write model types yourself, because these are generated from the GraphQL definitions your UI uses.
 
-As the generated types are query-specific, you're only able to access data you actually specify as part of a query. If you don't ask for a field in a particular query, you won't be able to access the corresponding property on the returned data structure.
+Because generated types are query-specific, you can only access data that you actually specify as part of a query. If you don't ask for a particular field in a query, you can't access the corresponding property on the returned data structure.
```
martinbonnin, line 10: FWIW, these 2 paragraphs were shamelessly copied from the iOS docs: https://www.apollographql.com/docs/ios/.

docs/source/shared/add-plugin.mdx
```diff
@@ -1,21 +1,21 @@
 import {MultiCodeBlock} from 'gatsby-theme-apollo-docs';
 
-Get the latest version of Apollo Android from [the Releases page](https://github.com/apollographql/apollo-android/releases).
+First, obtain the latest version number of Apollo Android from [the Releases page](https://github.com/apollographql/apollo-android/releases).
 
-In your app Gradle file, apply the `com.apollographql.apollo3` plugin.
+Then in your app's Gradle file, apply the `com.apollographql.apollo3` plugin using one of the following methods (replace `x.y.z` with the desired version number):
```
martinbonnin, line 5: I sometimes use `$version` instead of `x.y.z`. Might be worth uniformizing the pattern? I tend to like `$version` better as it will look more familiar to a Kotlin reader expecting a string template. Any thoughts?

docs/source/shared/add-plugin.mdx
```diff
@@ -1,21 +1,21 @@
...
 
-Using the [plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block):
+**Using the [plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block)**
 
 <MultiCodeBlock>
 
-```kotlin
+```kotlin:title=build.gradle
```
martinbonnin, line 11:
```suggestion
```kotlin:title=build.gradle.kts
```

docs/source/shared/add-plugin.mdx
```diff
@@ -24,11 +24,11 @@ plugins {
 
 </MultiCodeBlock>
 
-Or using the [legacy syntax](https://docs.gradle.org/current/userguide/plugins.html#sec:old_plugin_application):
+**Using the [legacy syntax](https://docs.gradle.org/current/userguide/plugins.html#sec:old_plugin_application)**
 
 <MultiCodeBlock>
 
-```kotlin
+```kotlin:title=build.gradle
```
martinbonnin, line 31:
```suggestion
```kotlin:title=build.gradle.kts
```

docs/source/shared/add-plugin.mdx
```diff
@@ -48,12 +48,12 @@ apply plugin: "com.apollographql.apollo3"
...
 </MultiCodeBlock>
 
-The plugin is hosted both on the Gradle plugin portal and Maven Central.
+The plugin is hosted on both the Gradle plugin portal and Maven Central.
 
-By default the plugin generates models in the root package name. You can configure this in the `apollo {}` block:
+By default, the plugin generates models in the root package name. You can configure this in the `apollo {}` block:
 
-```kotlin
+```kotlin:title=build.gradle
```
martinbonnin, line 55:
```suggestion
```kotlin:title=build.gradle[.kts]
```

docs/source/index.md
LouisCAD: Also, I don't think it's important to specify "the JVM". Java developers know that Kotlin libraries can be used in the JVM at this point, the language has been around and "advertised" for over 5 years now, I don't think we need to wait another 5.

docs/source/index.md
StephenBarlow: I think drift is fine since they're in different docsets with minimal reason to sync in general

docs/source/get-started.mdx
StephenBarlow: This actually depends on whether I am correct with my edit, so let me know if I'm incorrect!

My edit assumes that `runBlocking` is a valid option here, meaning the reader doesn't _have_ to replace it for this code to work. Because this is _optional_, they _can_ replace it. We should only use the imperative `replace` in situations where they _definitely_ have to replace something.

docs/source/index.md
LouisCAD: I don't know if it's important to specify "for Android" either, it's a given for a project named "apollo-android".

That now makes me think it'd be better named "apollo-kotlin". That would be less confusing regarding what it actually is and how compatible it is.

docs/source/shared/add-plugin.mdx
StephenBarlow: I agree we should standardize. If we go with `$version`, will that confuse any _other_ portion of our readers?

In most other docsets I'd use something large and noticeable, like "PLUGIN_VERSION_NUMBER". Would that work here or cause its own confusion?

docs/source/get-started.mdx
martinbonnin: It's usually a code smell to have `runBlocking {}` in production code. I put it here because it's the most straightforward way to create a `CoroutineScope` for tests and/or small snippet and didn't want to make any assumption about the context where the client will be called (Android vs Spring vs Ktor vs ... all have various ways to create their `CoroutineScopes`). 
In a real life app, there's 99% chances that the reader will have to replace this.

docs/source/shared/add-plugin.mdx
martinbonnin: `$version` works well for Kotlin and a bunch of other languages. I'm thinking of shell where that'd be a variable substitution as well. That doesn't work in Java though so maybe "VERSION_NUMBER" is better. Since there's only one version and it applies to the plugin + runtime + api + cache + ... I'd drop the "PLUGIN_" prefix.

docs/source/get-started.mdx
StephenBarlow: Cool thanks for the context! I'll revert to "Replace"

martinbonnin:
Thanks!

PR #3228, opened 2021-07-09
Edit pass on 3.0 migration guide

StephenBarlow:


docs/gatsby-config.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
         sidebarCategories: {
           null: [
             'index',
-            'get-started'
+            'get-started',
+            'migration/3.0',
```
StephenBarlow, line 18: Figure this'd be handier near the top

docs/gatsby-config.js
```diff
@@ -36,11 +37,6 @@ module.exports = {
             'advanced/upload',
             'advanced/using-aliases',
           ],
-          'Migration guides': [
```
StephenBarlow, line 39: Also figure we don't need to show migration guides for previous versions here anymore

docs/source/migration/1.3.mdx
```diff
@@ -1,6 +1,7 @@
 ---
-title: Migrating to 1.3
-description: Migrate to Apollo Android 1.3
+title: Migrating to Apollo Android 1.3
+description: From 1.2.x
+sidebar_title: "v1.3"
```
StephenBarlow, line 4: Ah these edits are from before I thought we should cut these articles from v3 entirely.. I'll get them reverted

docs/source/migration/3.0.mdx
```diff
@@ -1,50 +1,57 @@
...
 
-**Group id:**
+Apollo Android 3.0 uses a new identifier for its package name, maven group id, and Gradle plugin id: **`com.apollographql.apollo3`**.
 
-The maven group id used to identify the [artifacts](https://repo1.maven.org/maven2/com/apollographql/apollo3/) is now `com.apollographql.apollo3`
+This change helps avoid dependency conflicts and conflicting classes in the classpath, as encouraged in [Java Interoperability Policy for Major Version Updates](https://jakewharton.com/java-interoperability-policy-for-major-version-updates/).
+
+> In most cases, you can update identifier throughout your project by performing a find-and-replace on the 2.x identifier, `com.apollographql.apollo`.
+
+#### Package name
```
martinbonnin, line 26: I would change the order there to have:

1. Group ID
2. Plugin ID
3. Package name

While the package name is the most important change, the order in which the replacements will be done is going to be the one above. i.e. you first need to find the correct artifact then fix the plugin for the IDE to sync. Only then you can work in the code and start changing the package names

docs/source/migration/3.0.mdx
```diff
@@ -114,26 +119,26 @@ apollo {
 ```
 
 
-**Compilation units are removed**
+### Removed compilation units
```
martinbonnin, line 122:
```suggestion
### Removed `CompilationUnits`
```

docs/source/migration/3.0.mdx
```diff
@@ -148,11 +153,11 @@ apollo {
 
 ## Constructors
 
-Apollo Android 2.x uses the `Builder` pattern in different places to create new instances.
+Apollo Android 2.x uses the `Builder` pattern in various places to create new instances.
 
-Apollo Android 3.x leverages Kotlin default parameter and exposes the constructors directly as well as top-level functions.
+Apollo Android 3.x instead uses Kotlin default arguments and exposes constructors as top-level functions.
```
martinbonnin, line 158: I always get arguments & parameters mixed up, thanks 👍

docs/source/migration/3.0.mdx
```diff
@@ -275,13 +282,47 @@ val request = ApolloRequest(query).withHttpFetchPolicy(CacheFirst)
...
+
+## Optional values
+
+### The `Optional` class
+
+Apollo Android distinguishes between `null` values and absent values.
+
+Apollo Android 2.x uses `Input` to represent optional (maybe nullable) values for input types.
+
+Apollo Android 3.x uses `Optional` instead so that it can be used in places besides input types. For example, fields can be made optional with an [`@optional`](https://github.com/apollographql/apollo-android/blob/5c1ff82ccaded93ae7126cedc6d7835e63fed75a/apollo-compiler/src/main/resources/apollo.graphqls#L6) directive.
```
martinbonnin, line 323: > fields can be made optional

I used conditional here because it's still very experimental. It was an illustration of things that could be done but are not really set in stone. Maybe it's better to just remove it ?

docs/source/migration/3.0.mdx
```diff
@@ -323,52 +364,22 @@ class GetHero(val id: Input<String?> = Input.absent())
...
 ```
 
-By default the [GraphQL spec](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values) treats variables of nullable type as optional. It is valid to omit them at runtime. In practice though, this is rarely used and makes the operations declaration verbose.
+By default, the [GraphQL spec](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values) treats nullable variables as optional, so it's valid to omit them at runtime. In practice, however, this is rarely used and makes the operation's declaration verbose.
 
-Apollo Android 3.x removes the `Optional` wrapper so that it's easier to construct a query having nullable variables.
+Apollo Android 3.x removes the `Optional` wrapper so that it's more straightforward to construct a query with nullable variables.
 
-If omitting a variable is desired, it's possible to opt-in the `Optional` wrapper with the `@optional` directive:
+> **Note**: This new behavior applies _only_ to variables. Fields of input objects still use `Optional`, because it's common to omit particular input fields.
```
martinbonnin, line 371: Nitpick: maybe remove the "Note" here since we have the quotes highlighting the block already? 

```suggestion
> This new behavior applies _only_ to variables. Fields of input objects still use `Optional`, because it's common to omit particular input fields.
```

docs/source/migration/3.0.mdx
StephenBarlow: ahh ok I will try simplifying this and we can re-elaborate if it feels better

PR #3235, opened 2021-07-14
Edits to queries article for 3.0

StephenBarlow:


docs/source/essentials/00-queries.mdx
```diff
@@ -32,28 +31,32 @@ query HeroQuery($id: String!) {
...
 ```
 
-Here, `query` is the operation type and `HeroQuery` is the operation name.
-Apollo-android will generate a `HeroQuery` class that you can construct (with variables) and pass to `ApolloClient#query(query)`:
+If we define this query in its own `.graphql` file, Apollo Android will generate a corresponding `HeroQuery` class that we can then use to [execute the query](#executing).
 
-```kotlin
-val response = apolloClient.query(feedQuery)
-```
+To generate classes, Apollo Android requires your server's schema and a set of `.graphql` files (one for each operation in your application). It uses these to generate code you can use to execute queries and access typed results.
```
StephenBarlow, line 36: Does it actually have to be one operation per file? Or can you define multiple operations in a single file?

docs/source/essentials/00-queries.mdx
```diff
@@ -1,28 +1,27 @@
...
 
-Queries are represented as instances of generated classes conforming to the `Query` protocol. Constructor arguments can be used to define query variables if needed.
-You pass a query object to `ApolloClient#query(query)` to send the query to the server, execute it, and receive results.
+In Apollo Android, each query you execute is represented as an instance of a generated class that conforms to the `Query` protocol. To generate these classes, we first need to define the GraphQL operations we want to execute.
 
-For example, if you define a query called `GetHero`:
+Let's say we define a GraphQL query named `HeroQuery`, like so:
 
-```graphql
+```graphql:title=hero-query.graphql
```
StephenBarlow, line 24: Not sure what the filename convention should be here, happy to adjust

docs/source/essentials/00-queries.mdx
```diff
@@ -1,28 +1,27 @@
...
 
-> All `.graphql` files in your project (or the subset you specify as input to `apollo-codegen` if you customize the script you define as the code generation build phase) will be combined and treated as one big GraphQL document. That means fragments defined in one `.graphql` file are available to all other `.graphql` files for example, but it also means operation names and fragment names have to be unique and you will receive validation errors if they are not.
+This page also assumes that you've already set up Apollo Android for your Android/Kotlin application. For help with setup, see the [getting started guide](../#get-started).
 
-## Creating queries 
+## Defining
 
-Queries are represented as instances of generated classes conforming to the `Query` protocol. Constructor arguments can be used to define query variables if needed.
-You pass a query object to `ApolloClient#query(query)` to send the query to the server, execute it, and receive results.
+In Apollo Android, each query you execute is represented as an instance of a generated class that conforms to the `Query` protocol. To generate these classes, we first need to define the GraphQL operations we want to execute.
```
StephenBarlow, line 20: Do we have javadoc or similar for the library? If not, we can just link to wherever in the GH source the `Query` protocol is defined

docs/source/essentials/00-queries.mdx
```diff
@@ -1,28 +1,27 @@
...
 
-This page assumes some familiarity with building GraphQL queries. If you'd like a refresher, we recommend [reading this guide](http://graphql.org/learn/queries/) and practicing [running queries in GraphiQL](https://graphql.github.io/swapi-graphql/).
-Since Apollo queries are just standard GraphQL, anything you can type into the GraphiQL query explorer can also be put into `.graphql` files in your project.
+## Prerequisites
 
-The following examples assume that you've already set up Apollo Android for your Android/Kotlin application. Read our [getting started](../#get-started) guide if you need help with either of those steps.
+This page assumes some familiarity with building GraphQL queries. For a refresher, we recommend [reading this guide](http://graphql.org/learn/queries/) and practicing [running queries in Apollo Sandbox](https://studio.apollographql.com/sandbox/explorer?endpoint=https%3A%2F%2Fswapi-graphql.netlify.app%2F.netlify%2Ffunctions%2Findex&explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAOIIoBiAlgDZwDORwAOkkUQIY03V2Mtt2RAGa0GTVkKEoqKGgklCAvopVIlIADQgAbhzxUOAI3n0MIEEqA).
```
martinbonnin, line 12: 👍

docs/source/essentials/00-queries.mdx
martinbonnin: Javadoc has been on my todo list for a while. It'd be super useful for cases like this. Do you have a preferred option how to host them? I think Dokka (the documentation engine for Kotlin) can export Github markdown (example [here](https://square.github.io/okhttp/4.x/okhttp/okhttp3/)) so we could add it as a separate section in the gatsby site. This would only contain the latest API version but maybe that's ok?

docs/source/essentials/00-queries.mdx
martinbonnin: It can be multiple operations in a single file. In the end, all the `.graphql` files are concatenated together.

docs/source/essentials/00-queries.mdx
martinbonnin, line 24: We don't have really strong suggestions for naming (but I think we should, same for fragments). For the time being, I'd do:

```suggestion
```graphql:title=HeroQuery.graphql
```

docs/source/essentials/00-queries.mdx
martinbonnin: With the permalink to Github (and replacing some borrowed Swifties by Kotlin interfaces 🙈 )

```suggestion
In Apollo Android, each query you execute is represented as an instance of a generated class that conforms to the [`Query` interface](https://github.com/apollographql/apollo-android/blob/9c998e47f698bdbe5db6b6d6965dbc34183d507e/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Query.kt#L6). To generate these classes, we first need to define the GraphQL operations we want to execute.
```

docs/source/essentials/00-queries.mdx
martinbonnin: We don't have really strong suggestions for naming files/operations (but I think we should, same for fragments). For the time being, I'd do:

```suggestion
```graphql:title=HeroQuery.graphql
```

docs/source/essentials/00-queries.mdx
```diff
@@ -32,28 +31,32 @@ query HeroQuery($id: String!) {
...
 
-* On the JVM, the whole request runs in a coroutine using `Dispatchers.IO` by default. You can customize the dispatcher with the `dispatcher` constructor parameter of `ApolloClient`
-* On native, the request offloads the cache and network I/O to background threads that resume in the main dispatch queue. For this reason, the `ApolloClient` APIs assume they are called from the main thread. It's not possible to customize the dispatcher but cache and network I/O are not done on the main thread.
+## Executing
 
-Cancellation is handled through the `CoroutineScope`. Cancelling the current scope will cancel any ongoing request.
+To execute the query we [just defined](#defining), we pass a `HeroQuery` instance to the `query` method of `ApolloClient`, like so:
 
-Because it is safe to call the `ApolloClient` from the main thread, the result will also be dispatch to the main thread and you can direcly use the response to update your data.
+```kotlin{2}
```
martinbonnin, line 45: 👀  What does `{2}` do ?

StephenBarlow: Slightly highlights line 2 ✨

martinbonnin: giphy

PR #3250, opened 2021-07-16
Publish Kdoc

martinbonnin:
Publish the Kdoc on Github Pages and link it from the documentation site. 

I initially tried to generate Github flavored Markdown to inline in the documentation site but it has multiple issues:
- It increases the time to build the doc a lot because it basically requires building the project, installing an Android SDK in netlify machines, etc..
- Some links did not work because they were replaced by hashes. Not really sure what's going on
- More generally, the formatting isn't great because Markdown doesn't know anything about multiple sources sets for JVM/Native/JS, etc...
- Also Dokka [does not output code blocks](https://github.com/Kotlin/dokka/pull/1967)

All in all, publishing the Kdoc as HTML on Github pages makes it easier to navigate. Also it uses the same formatting as other libs, potentially reducing friction when reading API declarations, etc...

See it in action at https://apollographql.github.io/apollo-android/kdoc/

Also, this pull request publishes JavaDoc artifacts to maven central so we'll have the history of different versions as a nice bonus.


martinbonnin:
Merging this to include the Kdoc in the upcoming version. Feel free to leave comments here and I'll make a new PR

PR #3252, opened 2021-07-16
Edits to errors article

StephenBarlow:


docs/source/essentials/01-errors.mdx
```diff
@@ -1,16 +1,17 @@
...
+title: Error handling in Apollo Android
+sidebar_title: Errors
 ---
 
-GraphQL has fine control for error handling. When executing a GraphQL operation, errors can be of two kinds:
+Whenever you execute a GraphQL operation with Apollo Android (or any other GraphQL client), two high-level types of errors can occur:
 
-* Transport errors: a GraphQL response wasn't received because a transport error happened. This can be a SSL error, a socket error because your app is offline or a 500 or any other HTTP error. In this case you will not get any data.
-* GraphQL errors: a GraphQL response containing [errors](https://spec.graphql.org/draft/#sec-Errors) is received. This means the server wasn't able to completely process the query. You may get partial data if the server was able to process part of it.
+* **Network errors**: a GraphQL response wasn't received because an error occurred while communicating with your GraphQL server. This might be an SSL error, a socket error because your app is offline, or a 500 or any other HTTP error. When a network error occurs, no data is returned.
```
StephenBarlow, line 8: I changed "transport" to "network" to match the React client docs. Does that misrepresent anything?

martinbonnin: Good question! I like to think of "network" errors as everything that's below HTTP in the [OSI model](https://en.wikipedia.org/wiki/OSI_model). I consider HTTP to be some transport error but this might very well be a personal thing. I can see how "Network errors" can be easier to grasp.

The only thing (that's not addressed by "transport" errors either) is that cache misses will also throw exceptions. And a cache miss isn't really a "transport" neither "network" error. Maybe a solution would be to model the cache misses as part of `ApolloResponse` but that means users will have to explicitely handle them, compared to right now, handling all kinds of errors together.

But all in all, "network" sounds good to me 👍 .

docs/gatsby-config.js
```diff
@@ -19,9 +19,9 @@ module.exports = {
           ],
           'Essentials': [
             'essentials/00-queries',
-            'essentials/01-errors',
             'essentials/10-mutations',
             'essentials/20-subscriptions',
+            'essentials/01-errors',
```
martinbonnin, line 24: 👍  I was wondering whether this would make more sense. Overall, I initially tried to write the docs as users would write code. Users typically have to handle errors before doing mutations. But this is what the tutorial will be for so 👍 .

PS: because I'm a maniac and like the order of my files in my IDE to be the same as the order in the docs site, I'll rename this file to `21-errors` 🙃

martinbonnin (approved):

PR #3253, opened 2021-07-16
Edits to subscriptions article

StephenBarlow:


docs/source/essentials/20-subscriptions.mdx
```diff
@@ -1,48 +1,66 @@
...
-Subscriptions are similar to queries in syntax, the only difference being that you use the keyword `subscription` instead of `query` to indicate that the root fields on this query are going to be performing writes to the backend.
+You define a subscription in your app just like you define a query, except you use the `subscription` keyword. Here's an example subscription for getting the latest value of a number whenever that number is incremented:
+
+```graphql
+subscription NumberIncremented {
+  numberIncremented
+}
+```
+
+Unlike with queries and mutations, a subscription operation can include only _one_ field of the `Subscription` type. To subscribe to multiple fields, you create multiple subscription operations.
```
martinbonnin, line 16: TIL! Reading [the discussion in the spec](https://github.com/graphql/graphql-spec/pull/305/files), I'm still not 100% clear what is the root reason for this but good call to include it here 👍

martinbonnin (approved):

Issue #3230, opened 2021-07-12
Websocket closed error missing from graphql-ws subscription


dchappelle:
**Version**
com.apollographql.apollo3:apollo-runtime-kotlin:3.0.0-dev10

**Description**
The issue can be seen when making a subscription (that includes a security token) to a graphql-ws server. If the token is expired, the server will close the WebSocket with a 4401, backend_error – unauthorized (i.e. no cookie or other authentication data). This error is not seen by the apollo-android client though. The only error seen is "ACK not received after x amount of time". The same "ACK..." message is received on each retry. The client needs to know about the websocket being closed in order to stop retries and notify the user.

martinbonnin:
Closed with #3230 now available in `3.0.0-alpha03`

PR #3257, opened 2021-07-19
Dev 3.x: Websocket closed error missing from graphql-ws subscription (#3230)

dchappelle:
This is a minor update to communicate if the WebSocket has been closed by the server for some reason e.g., expired or invalid token.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -6,9 +6,17 @@ package com.apollographql.apollo3.exception
 open class ApolloException(message: String? = null, cause: Throwable? = null) : RuntimeException(message, cause)
 
 /**
- * A network happened: socket closed, DNS issue, TLS problem, etc...
+ * A network error happened: socket closed, DNS issue, TLS problem, etc...
  */
 class ApolloNetworkException(message: String? = null, cause: Throwable? = null) : ApolloException(message = message, cause = cause)
+
+/**
+ * A WebSocket connection could not be established: e.g., expired token
+ */
+class ApolloWebSocketException(
+    val code: Int,
+    message: String? = null,
+    cause: Throwable? = null) : ApolloException(message = message, cause = cause)
 
 /**
  * The response was received but the response code was not 200
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -13,7 +13,7 @@ class ApolloNetworkException(message: String? = null, cause: Throwable? = null) 
 /**
  * A WebSocket connection could not be established: e.g., expired token
  */
-class ApolloWebSocketException(
+class ApolloWebSocketClosedException(
     val code: Int,
     message: String? = null,
     cause: Throwable? = null) : ApolloException(message = message, cause = cause)
--- a/apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
+++ b/apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo3.network.ws
 
+import com.apollographql.apollo3.exception.ApolloWebSocketException
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.channels.Channel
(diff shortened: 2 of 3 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -6,10 +6,18 @@ package com.apollographql.apollo3.exception
 open class ApolloException(message: String? = null, cause: Throwable? = null) : RuntimeException(message, cause)
 
 /**
- * A network happened: socket closed, DNS issue, TLS problem, etc...
+ * A network error happened: socket closed, DNS issue, TLS problem, etc...
```
martinbonnin, line 9: Thanks!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -6,10 +6,18 @@ package com.apollographql.apollo3.exception
...
  */
 class ApolloNetworkException(message: String? = null, cause: Throwable? = null) : ApolloException(message = message, cause = cause)
 
+/**
+ * A WebSocket connection could not be established: e.g., expired token
+ */
+class ApolloWebSocketException(
```
martinbonnin, line 16: Do you need a separate exception for this given that there is `ApolloNetworkException`?

build-logic/src/main/kotlin/Mpp.kt
```diff
@@ -15,7 +15,7 @@ fun Project.configureMppDefaults(withJs: Boolean = true) {
     if (withJs) {
       js {
         useCommonJs()
-        browser()
+//        browser()
```
martinbonnin, line 18: Was that intended?

apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
```diff
@@ -63,6 +64,13 @@ actual class DefaultWebSocketEngine(
         messageChannel.close(t)
       }
 
+      override fun onClosing(webSocket: WebSocket?, code: Int, reason: String?) {
+        webSocketOpenResult.complete(Unit)
+
+        val t = ApolloWebSocketException(code, reason)
+        messageChannel.close(t)
```
martinbonnin, line 71: Why does this need to be in the `onClosing` method ? I would expect that `onClosed` would be called ultimately and that should trigger a `ClosedReceiveChannelException` from `readWebSocket()`. Or is `onClosed()` never called in your case?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
dchappelle: I thought about using ApolloNetworkException but then I would have to hack how the code would be included. So, I decided to introduce an explicit exception.

apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
dchappelle: In my testing, onClosing is only called when the server doesn't allow the connection e.g., expired token. I think it makes sense since that API includes a WebSocket code.

build-logic/src/main/kotlin/Mpp.kt
dchappelle: Nope. Is there a way to change this directly in Github?

apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
martinbonnin: Indeed. From the doc about `onClosing`:
```
Invoked when the remote peer has indicated that no more incoming messages will be transmitted.
```

So looks like `onClosed` = `onClosing` + client called `webSocket.close`? And the code in `onClosed` is the same as `onClosing`?

If that's the case, I think it should be fine since `messageChannel.close()` is idempotent. Being able to integration test this would be really nice. I created an issue there for this: https://github.com/apollographql/apollo-android/issues/3259

build-logic/src/main/kotlin/Mpp.kt
martinbonnin:
```suggestion
        browser()
```

martinbonnin: You can click the suggestion above 😃

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
martinbonnin: What about `ApolloWebSocketClosedException` ? This leaves the door open for potentially other `WebSocketExceptions` in the future if we need to? I think the code is only there when closing the websocket?

dchappelle: ApolloWebSocketClosedException sounds good to me.

martinbonnin (approved):

martinbonnin:
Thanks!

Issue #3265, opened 2021-07-23
v3.0.0: Gradle sync - Could not set unknown property 'customTypeMapping' for extension 'apollo'


omaksymov:
**Summary**
Gradle sync fails with
> Could not set unknown property 'customTypeMapping' for extension 'apollo' of type com.apollographql.apollo3.gradle.internal.DefaultApolloExtension

when using apollo gradle plugin v3.0.0-alpha02 (or 01)

**Version**
3.0.0-alpha02

**Description**
Trying to migrate from Apollo 2.5.9 to 3.0.0 (alpha02) using instructions at https://www.apollographql.com/docs/android/v3/migration/3.0/#gradle-configuration and can't even try that, because Gradle sync is failing with error message from summary.

build.gradle:
```
ext {
        kotlinVersion = '1.5.20'
        apolloVersion = '3.0.0-alpha02'
}
    
dependencies {
 classpath 'com.android.tools.build:gradle:4.2.2'
 classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
 classpath "com.apollographql.apollo3:apollo-gradle-plugin:$apolloVersion"
}    
```

app/build.gradle:
```
apply plugin: 'com.apollographql.apollo3'
...
apollo {
    packageName.set("...")
    customTypeMapping = [
            "IsoDateTime" : "java.util.Date",
...
            "JSON" : "full.path.to.ApolloJsonScalar"
    ]
}
```

According to https://www.apollographql.com/docs/android/essentials/plugin-configuration/ `customTypeMapping` should be still there. I didn't find any mentions of changes in this area in [release notes](https://github.com/apollographql/apollo-android/releases) as well.

What may be the problem?

martinbonnin:
Indeed, I'd expect that to work. Let me check quickly.

martinbonnin:
My bad, it's been renamed to `customScalarsMapping`:

```groovy
apollo {
    packageName.set("...")
    customScalarsMapping = [
            "IsoDateTime" : "java.util.Date",
...
            "JSON" : "full.path.to.ApolloJsonScalar"
    ]
}
```

I'll add that to the migration guide. 

Thanks for catching this!

omaksymov:
Yes, changing to `customScalarsMapping` worked. Thanks 👍

PR #3266, opened 2021-07-23
Mention `customScalarsMapping` in the migration guide

martinbonnin:
Closes https://github.com/apollographql/apollo-android/issues/3265

Thanks @omaksymov !

### DIFF
--- a/docs/source/migration/3.0.mdx
+++ b/docs/source/migration/3.0.mdx
@@ -152,7 +152,6 @@ apollo {
 }
 ```
 
-
 ### Removed `CompilationUnit`s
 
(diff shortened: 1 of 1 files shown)


omaksymov:
Not sure if it's helpful, but I see that only https://www.apollographql.com/docs/android/v3/migration/3.0/#gradle-configuration section is supposed to be updated in scope of this PR. While CustomType-related changes are quite significant in v3. For me it is not immediately obvious, how to migrate between 
```
com.apollographql.apollo.api
.CustomTypeAdapter
.CustomTypeValue
```
to 
```
com.apollographql.apollo3.api
.Adapter
.CustomScalarAdapters
.CustomScalarType
```
etc. respectively. Seems https://www.apollographql.com/docs/android/v3/essentials/30-custom-scalars/ puts more light on it, so thinking out loud that it may be worth having fully dedicated section for CustomType -> CustomScalar in migration doc (similar to "Optional values" section).

martinbonnin:
Yep, makes a lot of sense 👍 .

martinbonnin:
@omaksymov I added more details to the migration guide. I hope this helps. Let me know if you of anything else to add.

PR #3269, opened 2021-07-24
Update 04-execute-the-query.mdx

dizzersee:
Change creation of `ApolloClient` in tutorial to match with https://www.apollographql.com/docs/android/v3/migration/3.0/


martinbonnin:
Thanks for sending this! 
The tutorial in main is still using the 2.x version and documentation and also has to be kept consistent with https://github.com/apollographql/apollo-android-tutorial so I'd rather keep things as is for now. I created https://github.com/apollographql/apollo-android/issues/3270 to track the 3.x tutorial. Hopefully I can make a first drop next week.

martinbonnin:
The 3.x tutorial is now available at https://www.apollographql.com/docs/android/v3/tutorial/00-introduction/

PR #3284, opened 2021-08-02
Add error.path

martinbonnin:
Follow up from https://community.apollographql.com/t/path-in-error-type/931

Add `path` to GraphQL errors


martinbonnin:
Closed because path is already in `customAttributes` and exposing directly in `error` would be a breaking change. I'll open a follow up PR in `dev-3.x`

PR #3335, opened 2021-09-02
Java Codegen

martinbonnin:
See #2616

This PR is huge because it contains all the test fixtures. Look at the first commit for a more reasonnable diff.


martinbonnin:
The JavaCodegen can now be verified with `integration-tests` and `operation-based-tests`. This is nice because the same Kotlin code is used to test both codegens. The only thing that needed to change is the removal of named parameters that are not working when calling Java code. 

The Java codegen could be further improved with:

1. `@NonNull`/`@Nullable` annotations, ideally user-configurable
2. `Builder` APIs, especially for input objects to allow omitting input fields
3. Use of primitive types for non-nullable fields for performance reasons

Issue #3361, opened 2021-09-16
IndexOutOfBoundsException in MapJsonReader


ProVir:
Hi!

For my current schema, I get a strange error when getting data:
`com.apollographql.apollo3.exception.ApolloParseException: Failed to parse GraphQL http network response`
`java.lang.IndexOutOfBoundsException: Index: 10, Size: 1`

This bug was found in alpha04, but it is also reproduced in alpha06.

`names` has a strange index that turns out to be unrelated to it.

![image](https://user-images.githubusercontent.com/7127304/133649634-439ee098-b461-4bfa-bf44-e4a6e2e52eca.png)

martinbonnin:
Thanks for sending this. Any chance you can provide the `.graphql` query and matching `.json` response? This points into a nesting issue when trying to guess the next name.

ProVir:
My request and response data:

[GraphQL Request.txt](https://github.com/apollographql/apollo-android/files/7181077/GraphQL.Request.txt)

ProVir:
I will add to simplify the analysis of the problem: if you comment out both fields , the problem disappears. But it is only necessary to return any of them - it is reproduced again.

![image](https://user-images.githubusercontent.com/7127304/133837575-9c340cd4-a33d-42df-9498-3e9dfaf2fbea.png)

martinbonnin:
Sorry for the delay there. I can reproduce the issue. This is due to `operationBased`/`compat` codegen reading the same object twice asking less fields the second time. Will change this.

ProVir:
I tested in 3.0.0-alpha 07-SNAPSHOT - now it works. Thanks!

ProVir:
@martinbonnin  Tell me, when is release 3.0.0-alpha07 expected? There are a lot of changes in it for me to switch from 3.0.0-alpha04.

martinbonnin:
I was hoping to get the test APIs + Builders in and call that a beta01 but the test APIs are taking longer than expected. I'll release alpha07 today to unblock these issues.

martinbonnin:
`3.0.0-alpha07` is out 🎉

PR #3375, opened 2021-09-28
Fix OutOfBoundsException in MapJsonReader

martinbonnin:
Closes https://github.com/apollographql/apollo-android/issues/3361

- add a test case
- simplify the test
- rework MapJsonReader
- add a Json.md design doc


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/json/MapJsonReader.kt
```diff
@@ -277,15 +271,16 @@ class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
...
   }
 
   /**
    * Rewinds to the beginning of the current object.
    */
   override fun rewind() {
-    currentIndex = 0
-    currentName = null
+    indexStack[stackSize - 1] = 0
+    nameIndexStack[stackSize - 1] = 0
```
martinbonnin, line 283: This was the actual bug that `nameIndexStack` was not reset on `rewind()`. Everything else is refactoring and small-ish optimizations

PR #3405, opened 2021-10-12
WebSockets: better separate common code and protocol-specific code

martinbonnin:
Follow up from https://github.com/apollographql/apollo-android/pull/3401

`WebSocketNetworkTransport` now takes a `WsProtocol.Factory` instead of a `WsProtocol`. 

`WebSocketNetworkTransport` handles the opening/closing of the websocket and will create a new `WsProtocol` for each new websocket. The `WsProtocol` is responsible for the actual sending/receiving messages:

* receives commands through `WsProtocol.startOperation` and `WsProtocol.stopOperation`
* notifies `WebSocketNetworkTransport` with `listener.operationResponse`,`listener.operationError`,`listener.operationComplete` and `listener.networkError`

**Flow control:**

This change makes all buffers `UNLIMITED` so as to avoid any deadlock. This means that there is no flow control. If the client writes messages faster than what the server can handle, OkHttp will throw when trying to write more than 16MB and native will most likely OOM. I think it's OK as messages are very small and if that ever happens, it'll be the manifestation of another bug somewhere else that needs to be fixed.

**Graphql-ws "ping":**

Because "ping" is a `graphql-ws` specific message, sending "ping" messages has been moved to `GraphQLWsProtocol`. It is now controlled with `pingIntervalMillis` (disabled by default):

```kotlin
    val apolloClient = ApolloClient(
        networkTransport = WebSocketNetworkTransport(
            serverUrl = "http://localhost:9090/graphql",
            protocolFactory = GraphQLWsProtocol.Factory(
                pingIntervalMillis = 60_000
            ),
        )
    )
```

@dchappelle let me know how that works for you.  Also, I'm curious what server you're using. I've been trying to test with https://github.com/martinbonnin/graphql-ws-server and the server responds to any "ping" message with 

```
WebSocket Closed code='4400' reason='Invalid message received'
```

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/GraphQLWsProtocol.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/GraphQLWsProtocol.kt
@@ -21,6 +21,7 @@ class GraphQLWsProtocol(
     private val pongPayload: Map<String, Any?>? = null,
     private val connectionAcknowledgeTimeoutMs: Long,
     private val pingIntervalMillis: Long,
+    private val frameType: WsFrameType,
     webSocketConnection: WebSocketConnection,
     listener: Listener,
 ) : WsProtocol(webSocketConnection, listener) {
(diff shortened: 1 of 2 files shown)


dchappelle:
@martinbonnin I testing the changes but cannot establish a websocket connection now... "Connection initialization timeout"

I noticed a couple things..
- connectionAcknowledgeTimeoutMs defaults to -1 if a value is not provided, probably should default to 10_000
- The client no longer knows if the pong was not received

dchappelle:
@martinbonnin the connection issue is because the message was being sent out as binary instead of string... but I don't see WsFrameType.TEXT anymore?

PR #3431, opened 2021-10-20
Remove invalid message sent in SubscriptionWsProtocol

### LINKED ISSUE #3425: WebSocketNetworkTransport: invalid operation, and improve general error handling in WsProtocol
**Summary**
1. An invalid message is sent after 30s in `SubscriptionWsProtocol`. This was probably a test that was forgotten to be removed :)
2. When handling errors, the `id` field is actually not mandatory (it is present only when the error is about a specific operation) which could result in a crash.

**Version**
This happens in the `dev-3.x` branch only.

**Description**
- For 1., it is just a matter of removing the `sendMessageMapBinary(mapOf("type" to "ka"))` bit in `SubscriptionWsProtocol`
- For 2, let's add a new `generalError` method in `Listener`, without an `id` parameter, and call it from the `WsProtocol` implementations

BoD:
Fixes #3425:
- Remove invalid message sent in `SubscriptionWsProtocol`
- Handle error messages that don't have an `id` field, by introducing a new `generalError` fun in `WsProtocol.Listener` and a `GeneralError` event in `WebSocketNetworkTransport`
  - Like `OperationError`, this one throws an `ApolloNetworkException`, interrupting the Flow - so collectors will have to handle it. I think it makes sense but tell me if not 😅
  - Couldn't manage to add a test for this case, but at least I was able to add a test that `OperationError` still works as before

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
@@ -176,6 +176,10 @@ class WebSocketNetworkTransport(
           emit(it)
           false
         }
+        is GeneralError -> {
+          emit(it)
+          false
+        }
         else -> {
           emit(it)
           true


BoD:
> Are generic errors fatal in the fullstack tutorial case or can they be safely ignored?

It looks like the fullstack tutorial doesn't have a Subscriptions section 😅.  Couldn't find anything about this in JS client doc as well.

PR #3454, opened 2021-10-26
beta02 doc edits

martinbonnin:
A branch to be merged **only after** beta02 is merged and that contain doc edits for the beta02


martinbonnin:
Update to include Fluent APIs. Note that this doesn't touch the tutorial yet. In the future, would be nice to always prepare PRs like this before the release so that we only have to hit "merge"

BoD (approved):

Issue #3410, opened 2021-10-14
2.x Compatibility layer


martinbonnin:
In order to make migrating easier, add compatibility classes marked as `@Deprecated` that will be compatible with 2.x and will be removed down the road in a future 3.x (or 4.x?) version. 

Current identified friction points:

**Gradle plugin**:

- [x] sealedClassesForEnumsMatching
- [x] generateOptionalOperationVariables
- [x] use2xCompatDefaults()

**apollo-runtime**:
- [x] CustomTypeAdapter and CustomTypeValue
- [x] ApolloCall
- [ ] CacheKeyResolver
- [x] Input
- [x] customScalarMappings
- [x] a builder() method on ApolloClient.Companion.

PR #3482, opened 2021-11-02
Add a "compat" v2 CustomTypeAdapter to v3 Adapter API

BoD:
This is part of #3410
- duplicate `CustomTypeValue` and `CustomTypeAdapter` from v2 👉 users should be able to use them by updating imports
- add a `Version2CustomTypeAdapterToAdapter` to wrap existing `CustomTypeAdapter` and expose them as v3 `Adapter`
- add `addCustomTypeAdapter` on `ApolloClient.Builder` (and on `CustomScalarAdapters`) that uses  `Version2CustomTypeAdapterToAdapter` under the hood.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapter.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapter.kt
@@ -2,9 +2,9 @@ package com.apollographql.apollo3.api
 
 import com.apollographql.apollo3.api.internal.json.BufferedSinkJsonWriter
 import com.apollographql.apollo3.api.internal.json.BufferedSourceJsonReader
-import com.apollographql.apollo3.api.json.MapJsonReader
 import com.apollographql.apollo3.api.json.JsonReader
 import com.apollographql.apollo3.api.json.JsonWriter
+import com.apollographql.apollo3.api.json.MapJsonReader
 import okio.Buffer
 import okio.BufferedSink
 import okio.BufferedSource
@@ -53,10 +53,10 @@ interface Adapter<T> {
    * }
    * ```
    *
-   * Alternatively, you can use the the built-in [AnyAdapter] to simplify the parsing loop:
+   * Alternatively, you can use the built-in [AnyAdapter] to simplify the parsing loop:
    * ```
    * override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): Hero {
-   *   val map = AnyAdapter.fromResponse(reader) as Map<String, String>
+   *   val map = AnyAdapter.fromJson(reader) as Map<String, String>
    *
    *   return Hero(map["name"]!!, map["homeworld"]!!)
    * }
@@ -77,7 +77,7 @@ interface Adapter<T> {
    * }
    *```
    *
-   * Alternatively, you can use the the built-in [AnyAdapter]:
+   * Alternatively, you can use the built-in [AnyAdapter]:
    * ```
    * override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: Hero) {
    *   val map = mapOf("name" to value.name, "homeworld" to value.homeworld)
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
@@ -1,0 +1,73 @@
+package com.apollographql.apollo3.api
+
+import com.apollographql.apollo3.api.json.JsonReader
+import com.apollographql.apollo3.api.json.JsonWriter
+import kotlin.jvm.JvmStatic
+
+/**
+ * An [Adapter] that wraps an Apollo Android v2 style [CustomTypeAdapter], to ease migration from v2 to v3.
+ */
+@Deprecated("Used for backward compatibility with 2.x, use Adapter instead")
+class Version2CustomTypeAdapterToAdapter<T>(
+    private val v2CustomTypeAdapter: CustomTypeAdapter<T>,
+) : Adapter<T> {
+  override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): T {
+    val value: Any? = if (reader.peek() == JsonReader.Token.NULL) null else AnyAdapter.fromJson(reader)
+    return v2CustomTypeAdapter.decode(CustomTypeValue(value))
+  }
+
+  override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: T) {
+    val encoded: Any? = v2CustomTypeAdapter.encode(value).value
+    if (encoded == null) {
+      writer.nullValue()
+    } else {
+      AnyAdapter.toJson(writer, encoded)
+    }
+  }
+}
+
+/**
+ * A replica of Apollo Android v2's CustomTypeAdapter, to ease migration from v2 to v3.
+ *
+ * Make your CustomTypeAdapters implement this interface by updating the imports
+ * from `com.apollographql.apollo.api` to `com.apollographql.apollo3.api`
+ * and wrap them in [Version2CustomTypeAdapterToAdapter]s, which you can add to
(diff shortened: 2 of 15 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapter.kt
```diff
@@ -53,10 +53,10 @@ interface Adapter<T> {
    * }
    * ```
    *
-   * Alternatively, you can use the the built-in [AnyAdapter] to simplify the parsing loop:
+   * Alternatively, you can use the built-in [AnyAdapter] to simplify the parsing loop:
```
martinbonnin, line 56: 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * An [Adapter] that wraps an Apollo Android v2 style [CustomTypeAdapter], to ease migration from v2 to v3.
+ */
+@Deprecated("Used for backward compatibility with 2.x, use Adapter instead")
+class Version2CustomTypeAdapterToAdapter<T>(
+    private val v2CustomTypeAdapter: CustomTypeAdapter<T>,
+) : Adapter<T> {
+  override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): T {
+    val value: Any? = if (reader.peek() == JsonReader.Token.NULL) null else AnyAdapter.fromJson(reader)
```
martinbonnin, line 15: Nit: there is a `NullableAnyAdapter` : https://github.com/apollographql/apollo-android/blob/d852dd92ee084732d49ae993f4e1346384f2fde7/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultAdapters.kt#L224

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import com.apollographql.apollo3.api.json.JsonReader
+import com.apollographql.apollo3.api.json.JsonWriter
+import kotlin.jvm.JvmStatic
+
+/**
+ * An [Adapter] that wraps an Apollo Android v2 style [CustomTypeAdapter], to ease migration from v2 to v3.
+ */
+@Deprecated("Used for backward compatibility with 2.x, use Adapter instead")
+class Version2CustomTypeAdapterToAdapter<T>(
```
martinbonnin, line 11: Could we make that internal? And only use it from deprecated functions?

tests/integration-tests/src/commonTest/kotlin/test/ScalarAdapterTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.apollographql.apollo3.mockserver.MockServer
+import com.apollographql.apollo3.mockserver.enqueue
+import com.apollographql.apollo3.testing.runTest
+import com.example.MyDate
+import readResource
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class ScalarAdapterTest {
```
martinbonnin, line 21: Nice integration test 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
BoD: Neat! I'll amend.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
BoD: Definitely, I was not sure if it would make sense to keep it public, but I guess that would not be very useful indeed.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * A replica of Apollo Android v2's CustomTypeValue, to ease migration from v2 to v3.
+ *
+ * In your [CustomTypeAdapter], update the imports from `com.apollographql.apollo.api` to
+ * `com.apollographql.apollo3.api` to use this version.
+ */
+@Deprecated("Used for backward compatibility with 2.x, use Adapter instead")
+open class CustomTypeValue<T>(val value: T) {
+  object GraphQLNull : CustomTypeValue<Unit>(Unit)
+  class GraphQLString(value: String) : CustomTypeValue<String>(value)
```
martinbonnin, line 52: These used to implement `equals`/`hashcode`. I don't think it's required but asking just in case, is it important?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+}
+
+/**
+ * A replica of Apollo Android v2's CustomTypeAdapter, to ease migration from v2 to v3.
+ *
+ * Make your CustomTypeAdapters implement this interface by updating the imports
+ * from `com.apollographql.apollo.api` to `com.apollographql.apollo3.api`
+ * and wrap them in [Version2CustomTypeAdapterToAdapter]s, which you can add to
```
martinbonnin, line 34: Reminder to remove Version2CustomTypeAdapterToAdapter here if it becomes internal

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
BoD: Good question! I stripped them as a simplification but without too much thinking 😅. Looking at it more a bit more closely at how they are used, I don't think this would break anything.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
BoD: Oops, actually `internal` doesn't work because `ApolloClient` is in `apollo-runtime` whereas this is defined in `apollo-api`. But maybe it would make more sense to move it to `apollo-runtime` actually? I thought to put it next to `Adapter` but it would probably be needed only when using the client, so... WDYT?

martinbonnin: A third option would be to use an `Optin` `ApolloInternal` annotation to make sure anyone using this knows it's not intended for public use?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
martinbonnin: Sounds good 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapterToAdapter.kt
BoD: Ah that's a good idea! And could be useful in the future for similar situations 👍.  Let me have a 👀

BoD: Added the annotation (also moved `Version2CustomTypeAdapterToAdapter` to the `internal` package)

martinbonnin (approved):

PR #3485, opened 2021-11-02
Add parsing of non-standard Error fields

BoD:
I noticed GitHub's API returns Errors that have a non standard `type` field.

Looking at the spec, these are discouraged but still allowed - so I thought it could be useful to surface them.

In v2 these would come up in the `extensions` field - but I opted to add a specific field instead, to avoid potential clashes in the (improbable) event where fields with the same name appear at the top level *and* in `extensions`.


BoD:
Note: also opened a [ticket](https://github.com/github/feedback/discussions/7080) for GitHub about this.

martinbonnin (approved):

PR #3492, opened 2021-11-03
Add documentation for watch

nachtien:
- I moved `NetworkFirst` to the top since that is the default `FetchPolicy` and fixed the documentation to state that this is the default.
- I also added documentation to clarify what `watch` does. 

``` Overriding the [FetchPolicy] will change how the result is first queried. ```
might make more sense to be
``` Setting the [FetchPolicy] will change how the result is first queried. ```

Feel free to adjust as needed.

### DIFF
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -19,19 +19,9 @@ import kotlinx.coroutines.flow.flow
 
 enum class FetchPolicy {
   /**
-   * Try cache first, then network
+   * Try network first, then cache
    *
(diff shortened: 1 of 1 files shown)


martinbonnin:
> I moved NetworkFirst to the top since that is the default FetchPolicy and fixed the documentation to state that this is the default.

Mmm it shouldn't be. Let me check again.

martinbonnin:
Indeed, `CacheFirst` should be the default: https://github.com/apollographql/apollo-android/blob/1c703f6e5a8cd0961bd6b360e1cf918505f18774/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt#L161

Is there a case where this doesn't work ?

nachtien:
Woops, you're right. I removed the `FetchPolicy` changes.

martinbonnin (approved):

martinbonnin:
Thanks!

PR #3496, opened 2021-11-04
Make the Gradle plugin work without the Kotlin Plugin

martinbonnin:
The [JVM spec](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-5.html) is pretty loose when it comes to class loading/linking and it could happen that it failed with a runtime exception when trying to link `DefaultApolloExtension`.

Move all the Kotlin plugin code to KotlinPluginFacade so that it's never loaded in the absence of the Kotlin plugin in the runtime classpath.


BoD (approved):
👍

PR #3503, opened 2021-11-05
Fix concurrent accesses to the SQLite database

martinbonnin:
Introduce `readAccess {}` and `writeAccess {}` for separate 'read' and 'write' database accesses. Previously, everything was using the 'read' lock which could trigger `SQLiteDatabaseLockedException: database is locked (code 5)` exceptions.


BoD (approved):
👍

PR #3505, opened 2021-11-05
Update cache docs

martinbonnin:
- first drop for cache docs
- added Query watchers
- wip
- wip
- update HTTP Cache

### DIFF
--- a/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/CacheKey.kt
+++ b/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/CacheKey.kt
@@ -3,8 +3,33 @@ package com.apollographql.apollo3.cache.normalized
 import kotlin.jvm.JvmStatic
 
 /**
+ * A [CacheKey] identifies an object in the cache.
+ *
+ * @param key The key of the object in the cache. The key must be globally unique.
  */
-class CacheKey(val key: String) {
+class CacheKey constructor(val key: String) {
+
+  /**
+   * Builds a [CacheKey] from a typename and a list of Strings.
+   *
+   * This can be used for the common case where [CacheKey] use [typename] as a namespace and [values] as a path.
+   */
+  constructor(typename: String, values: List<String>): this(
+      buildString {
+        append(typename)
+        append(":")
+        values.forEach {
+          append(it)
+        }
+      }
+  )
+
+  /**
+   * Builds a [CacheKey] from a typename and a list of Strings.
+   *
+   * This can be used for the common case where [CacheKey] use [typename] as a namespace and [values] as a path.
+   */
+  constructor(typename: String, vararg values: String) : this(typename, values.toList())
 
   override fun hashCode() = key.hashCode()
   override fun equals(other: Any?): Boolean {
@@ -48,21 +73,13 @@ class CacheKey(val key: String) {
     /**
      * Helper function to build a cache key from a list of strings
      */
-    fun from(typename: String, values: List<String>): CacheKey {
-      return CacheKey(
-          buildString {
-            append(typename)
-            append(":")
-            values.forEach {
-              append(it)
-            }
-          }
-      )
-    }
+    @Deprecated("Use the constructor instead", ReplaceWith("CacheKey(typename, values)"))
+    fun from(typename: String, values: List<String>) = CacheKey(typename, values)
 
     /**
      * Helper function to build a cache key from a list of strings
      */
-    fun from(typename: String, vararg values: String) = from(typename, values.toList())
+    @Deprecated("Use the constructor instead", ReplaceWith("CacheKey(typename, values)"))
+    fun from(typename: String, vararg values: String) = CacheKey(typename, values.toList())
   }
 }
--- a/docs/source/caching/declarative-ids.mdx
+++ b/docs/source/caching/declarative-ids.mdx
@@ -11,7 +11,7 @@ It works by extending the backend schema. The GraphQL spec allows [object extens
 Cache ids are computed from two sources:
 
 1. From the object fields (`author.id`)
-2. From a field arguments (`author(id)`)
+2. From a field's arguments (`author(id)`)
 
 
 1. is happening **after** a network request and is important to merge the network response with existing data. This is the most usual case.
--- a/docs/source/caching/declarative-ids.mdx
+++ b/docs/source/caching/declarative-ids.mdx
@@ -8,10 +8,11 @@ Declaratively specifying your cache ids allows the codegen to add the id fields 
 
 It works by extending the backend schema. The GraphQL spec allows [object extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) using the `extend type` keywords. In our case, this can be used together with the `@typePolicy` and `@fieldPolicy` directives and an extra '.graphqls' file that we will name `extra.graphqls`.
 
-Cache ids are needed in two places:
+Cache ids are computed from two sources:
 
-1. As an object field (`author.id`)
-2. As a field argument (`author(id)`)
+1. From the object fields (`author.id`)
+2. From a field arguments (`author(id)`)
+
 
 1. is happening **after** a network request and is important to merge the network response with existing data. This is the most usual case.
 2. is happening **before** a network request and is important to avoid a network round trip if all the data is in the cache already. This is an optimization that can avoid some cache misses.
--- a/docs/source/caching/declarative-ids.mdx
+++ b/docs/source/caching/declarative-ids.mdx
@@ -1,0 +1,98 @@
+---
+title: Declarative cache ids
+---
+
+> Using declarative cache ids is the recommended way to define your cache ids. For a more flexible (and more complex API), you can also define your cache ids [programmatically](programmatic-ids).
+
+Declaratively specifying your cache ids allows the codegen to add the id fields automatically and generates type-safe code that can return a valid cache id for any object or field.
+
+It works by extending the backend schema. The GraphQL spec allows [object extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) using the `extend type` keywords. In our case, this can be used together with the `@typePolicy` and `@fieldPolicy` directives and an extra '.graphqls' file that we will name `extra.graphqls`.
+
+Cache ids are needed in two places:
+
+1. As an object field (`author.id`)
+2. As a field argument (`author(id)`)
+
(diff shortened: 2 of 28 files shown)


docs/source/caching/http-cache.mdx
```diff
@@ -2,81 +2,49 @@
...
-    .serverUrl("/")
-    .httpCache(directory = directory, maxSize = maxSize)
-    .build()
-                   
-// Control the fetch policy  
-val query = ApolloRequest.Builder(FeedQuery(limit = 10, type = FeedType.HOT))
-    .httpFetchPolicy(HttpFetchPolicy.CacheFirst)
-    .build()
-val dataResponse = apolloClient.query(query)
 ```
```
BoD, line 20:
```suggestion
```kotlin
```

docs/source/caching/http-cache.mdx
```diff
@@ -2,81 +2,49 @@
...
-  .build();
-
-ApolloResponse<FeedQuery.Data> response = rx2ApolloClient.query(query).blockingGet();
+val dir = File(pathToCacheDirectory)
+
+// Configure a max size of 100MB
+val maxSize = 100 * 1024 * 1024
+
+apolloClient = ApolloClient.Builder()
+    .networkTransport(
```
BoD, line 27: Shouldn't we show the `httpCache` extension?

martinbonnin: 💯 , I forgot it existed

docs/source/caching/http-cache.mdx
```diff
@@ -2,81 +2,49 @@
...
+                 .httpFetchPolicy(HttpFetchPolicy.NetworkOnly)
 
-- `NetworkOnly` - Fetch a response from the network only, ignoring any cached responses. This is the default.
-- `CacheOnly` - Fetch a response from the cache only, ignoring the network. If the cached response doesn't exist or is expired, then return
-  an error.
-- `CacheFirst` - Fetch a response from the cache first. If the response doesn't exist or is expired, then fetch a response from the network.
-- `NetworkFirst` - Fetch a response from the network first. If the network fails and the cached response isn't expired, then return cached
-  data instead.
+                 // Or only use the cache
+                 .httpFetchPolicy(HttpFetchPolicy.CacheOnly)
```
BoD, line 41: Maybe a note saying that in this case `HttpCacheMissException` will be thrown in case of a cache miss?

docs/source/caching/normalized-cache.mdx
```diff
@@ -33,68 +33,53 @@ query AuthorById($id: String!) {
...
 // Build the ApolloClient
 val apolloClient = ApolloClient.Builder()
   .serverUrl("https://...")
+  // normalizedCache() is an extension function on ApolloClient.Builder
   .normalizedCache(cacheFactory)
   .build()
 ```
 
-</MultiCodeBlock>
+Because the normalized cache is optional, `normalizedCache()` is an extension function on `ApolloClient.Builder()` that's defined in the `apollo-normalized-cache` artifact. It takes a `NormalizedCacheFactory` as a parameter so that it can create the cache in the background if needed.
```
BoD, line 61: Nit: not 100% sure, is "in the background" the correct term? Maybe "on demand" or just "as needed" is more accurate?

docs/source/caching/normalized-cache.mdx
```diff
@@ -33,68 +33,53 @@ query AuthorById($id: String!) {
...
-  implementation("com.apollographql.apollo3:apollo-normalized-cache-sqlite:x.y.z")
-}
-```
+When your app is stopped, data in the `MemoryCache` is lost forever. If you need to persist data, you can use the SQLite cache.
 
-**Note:** The `apollo-normalized-cache-sqlite` dependency has Kotlin multiplatform support and has multiple variants (`-jvm`, `-android`, `-ios-arm64`,...). If you are targeting Android and using custom `buildTypes`, you will need to help Gradle resolve the correct artifact by defining [matchingFallbacks](https://developer.android.com/studio/build/dependencies?authuser=2#resolve_matching_errors):
+## SQLite cache
 
-<MultiCodeBlock>
+The SQLite cache uses [SQLDelight](https://github.com/cashapp/sqldelight) to store data persistently if you need to to persist data accross app restarts or simply if the data becomes too big to fit in memory.
```
BoD, line 72: Nit
```suggestion
The SQLite cache uses [SQLDelight](https://github.com/cashapp/sqldelight) to store data persistently. It can be used when you need to persist data across app restarts or simply if the data becomes too big to fit in memory.
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -111,27 +96,48 @@ val apolloClient = ApolloClient.Builder()
   .build()
 ```
 
+You can then use it like the `MemoryCache
```
BoD, line 99:
```suggestion
You can then use it like the `MemoryCache`.
```

docs/source/caching/normalized-cache.mdx
martinbonnin: It's really more about not doing it on the main thread. Since it's only every creating one instance we could create it eagerly except that in most times, that would happen on the main thread. 

Maybe we just remove this sentence from the doc? And leave it for the KDoc?

BoD: Gotcha! Wasn't sure it's what you meant, in that case it's correct 👍. Yes adding this info in the KDoc could be useful. You can remove it here or also leave it, it's doesn't hurt.

docs/source/caching/declarative-ids.mdx
```diff
@@ -0,0 +1,100 @@
...
+> Using declarative cache ids is the recommended way to define your cache ids. For a more flexible (and more complex) API, you can also define your cache ids [programmatically](programmatic-ids).
+
+Declaratively specifying your cache ids allows the codegen to add the id fields automatically and generates type-safe code that can return a valid cache id for any object or field.
+
+It works by extending the backend schema. The GraphQL spec allows [object extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) using the `extend type` keywords. In our case, this can be used together with the `@typePolicy` and `@fieldPolicy` directives and an extra '.graphqls' file that we will name `extra.graphqls`.
+
+Cache ids are computed from two sources:
+
+1. From the object fields (`author.id`)
+2. From a field arguments (`author(id)`)
```
BoD, line 14:
```suggestion
2. From a field's arguments (`author(id)`)
```

docs/source/caching/programmatic-ids.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+If you need the object typename, you can access it through `context`:
+
+```kotlin
+val objectIdGenerator = object : ObjectIdGenerator {
+  override fun cacheKeyForObject(obj: Map<String, Any?>, context: ObjectIdGeneratorContext): CacheKey? {
+    val typename = field.type.leafType().name
+    val id = obj["id"] as String
+
+    return CacheKey(typename, id)
```
BoD, line 57:
```suggestion
    val typename = context.field.type.leafType().name
    val id = obj["id"] as String

    return CacheKey.from(typename, id)
```

martinbonnin: I was thinking of removing the `.from()` because it's adding yet another way to create instances:

1. Constructors
2. Builders
3. Factory Methods

My current understanding is that:

* Constructors should be preferred when there is at most one optional parameter
* Builders should be preferred otherwise
* Factory methods should be used when ???

BoD: Good question 😅.  I'd probably be in favor of a secondary constructor for this one (a Builder being a bit too heavy for this maybe?).

martinbonnin: Secondary constructor added there https://github.com/apollographql/apollo-android/pull/3505/files#diff-df868ffc210fd332a03f99ad720bf7f4dfe3adf26af956e4e6a9f5b0dc279582R17

let's remove all Factory Methods 🙌

docs/source/caching/programmatic-ids.mdx
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+```
+
+Note how this requires `id` to be present in the response. If a query does not return this field, the `CacheKey` will be inconsistent and duplication will happen.
+
+## CacheKeyResolver
+
+To customize how cache ids are built from fields arguments (before going to the network), create a new instance of `CacheKeyResolver`:
```
BoD, line 66: We may want to clarify that the resolver must then be passed as an argument to `normalizedCache`

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/CacheKey.kt
```diff
@@ -48,21 +73,13 @@ class CacheKey(val key: String) {
...
-          buildString {
-            append(typename)
-            append(":")
-            values.forEach {
-              append(it)
-            }
-          }
-      )
-    }
+    @Deprecated("Use the constructor instead", ReplaceWith("CacheKey(typename, values)"))
```
BoD, line 76: 👍

docs/source/caching/query-watchers.mdx
```diff
@@ -0,0 +1,119 @@
+---
+title: Query watchers
+---
+
+Because the normalized cache deduplicates all your data, and if you defined your cache ids, you can the cache it a the source of truth for your UI and be automatically notified when some data changes. For this you can use `ApolloQueryCall.watch`
```
BoD, line 5:
```suggestion
Because the normalized cache deduplicates all your data, and if you defined your cache ids, you can use the cache as the source of truth for your UI and be automatically notified when some data changes. For this you can use `ApolloQueryCall.watch`
```

docs/source/caching/programmatic-ids.mdx
martinbonnin: 100%, will do

docs/source/caching/programmatic-ids.mdx
```diff
@@ -0,0 +1,164 @@
...
+    )
+    .build()
+```
+
+If you need the object typename, you can access it through `context`:
+
+```kotlin
+val objectIdGenerator = object : ObjectIdGenerator {
+  override fun cacheKeyForObject(obj: Map<String, Any?>, context: ObjectIdGeneratorContext): CacheKey? {
+    val typename = field.type.leafType().name
```
BoD, line 54:
```suggestion
    val typename = context.field.type.leafType().name
```

BoD (approved):

PR #3511, opened 2021-11-08
Mark Test Builders are experimental

### LINKED ISSUE #3444: Mark generated Test Builders as `@ApolloExperimental`
Mark generated Test Builders as `@ApolloExperimental` the time the API stabilizes.

Also investigate whether it makes sense to add it to `Service.generateTestBuilders` in the Gradle plugin. Groovy callers will certainly not be aware of this but maybe it can work for `build.gralde.kts` ?

BoD:
Resolves #3444.

Add `@ApolloExperimental` to the generated code, which raises a warning when using it:

Screen Shot 2021-11-08 at 17 36 15

Also mark the Gradle plugin setting as experimental but unfortunately this doesn't raise the warning at the call site in `build.gradle.kts`:

Screen Shot 2021-11-08 at 17 48 58

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/TestBuildersBuilder.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/TestBuildersBuilder.kt
@@ -8,12 +8,10 @@ import com.apollographql.apollo3.compiler.codegen.Identifier.block
 import com.apollographql.apollo3.compiler.codegen.Identifier.customScalarAdapters
 import com.apollographql.apollo3.compiler.codegen.Identifier.fromJson
 import com.apollographql.apollo3.compiler.codegen.Identifier.testResolver
-import com.apollographql.apollo3.compiler.codegen.kotlin.KotlinContext
 import com.apollographql.apollo3.compiler.codegen.kotlin.CgFile
-import com.apollographql.apollo3.compiler.codegen.kotlin.CgFileBuilder
-import com.apollographql.apollo3.compiler.codegen.kotlin.CgOutputFileBuilder
 import com.apollographql.apollo3.compiler.codegen.kotlin.CgTestFileBuilder
 import com.apollographql.apollo3.compiler.codegen.kotlin.KotlinClassNames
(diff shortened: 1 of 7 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/TestBuildersBuilder.kt
```diff
@@ -64,6 +63,8 @@ class TestBuildersBuilder(
 
   private fun typeSpec(): TypeSpec {
     return TypeSpec.objectBuilder(simpleName)
+        // Note: the compiler forbids referencing ApolloExperimental::class directly, except in an @OptIn context
+        .addAnnotation(ClassName("com.apollographql.apollo3.api", "ApolloExperimental"))
```
martinbonnin, line 67: 👍  I also have the project to relocate `apollo-api` in the Gradle plugin jar too so that'll help.

Nit: there's a `ClassNames` file that tries to list all hardcoded class names like this, would be nice to move it there: https://github.com/apollographql/apollo-android/blob/d852dd92ee084732d49ae993f4e1346384f2fde7/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/ClassNames.kt#L8

BoD: Move it (and removed the comment which was no longer really useful)

BoD:
It didn't fail in the test because this only produces a warning, not an error - but you/re right, and I've just added the @Optin to the tests, that removes the warnings.

martinbonnin (approved):

martinbonnin:
One last question: do we get a warning when compiling `models-response-based/build.gradle.kts`? Can we remove it?

BoD:
> do we get a warning when compiling models-response-based/build.gradle.kts?

It doesn't look so. But I don't know why not 😅

martinbonnin:
> It doesn't look so. But I don't know why not 😅

It's displayed if run with `./gradlew -i` 🤷 . Also making the `@ApolloExperimental` an Error would break the build so it's somewhat working.

martinbonnin (approved):

PR #3512, opened 2021-11-08
Reduce CI build times

### LINKED ISSUE #3491: Optimize CI times
With the addition of new native targets (#3487), the CI build time went from ~45min to a whooping ~1h20. This is far from ideal, especially for PR builds. 

Several options to keep the CI times manageable:

- parallelize over multiple machines
- parallelize tests (gradle and compiler tests do take a long time)
- don't build all native targets for PRs
- other?

martinbonnin:
The goal is to have a build matrix like so:

* on PR
  * tests-gradle
    * slow gradle tests
  * tests-no-gradle
    * all root tests but not the Gradle ones and not the "exotic" apple ones (tvos, watchos)
    * all apiCheck
  * tests-integration
    * all integration tests
* on push to dev-3.x
  * ciFull
    * macOS
    * slow
    * runs all tests
  * kdoc
    * publishes KDoc
  * ciPublishSnapshot
    * publishes Snapshot
* on tag
  * ciPublishRelease

closes #3491


.github/workflows/pr.yml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Build with Gradle
+        run: |
+          ulimit -c unlimited
+          ./gradlew ciTestsGradle
+      - name: Collect Diagnostics
+        if: always()
+        run: ./scripts/collect-diagnostics.main.kts
```
BoD, line 29: Sorry I have a silly question: how come running Kotlin scripts works without needing any setup?

martinbonnin: This is thanks to @LouisCAD :) https://github.com/actions/virtual-environments/issues/3687

BoD (approved):

build.gradle.kts
```diff
@@ -96,55 +95,60 @@ fun shouldPublishSnapshots(): Boolean {
...
+
   subprojects {
-    tasks.all {
-      if (this.name == "build") {
-        this@register.dependsOn(this)
-      }
+    if (name != "apollo-gradle-plugin") {
+      dependsOn(tasks.matching { it.name == "test" })
     }
+    dependsOn(tasks.matching { it.name == "jvmTest" })
```
LouisCAD, line 142: That's using an eager API that defeats configuration avoidance FYI.

martinbonnin: Oh no, not again 😅 . I thought `TaskCollection.matching {}` was ok? What is the eager part?

LouisCAD: That's what I recall reading from @melix a few days ago. Maybe there's been come confusion?

melix: `matching` is eager. `tasks.withType(...)` is not. So it's better to use `tasks.named("jvmTest", Test)` which is lazy.

martinbonnin: Wait, I'm confused now, the page I linked above explicitly says about `TaskCollection.matching(groovy.lang.Closure)`:

```
This is OK to use because it does not require tasks to be created immediately.
```

What am I missing?

melix: `tasks.matching { ... }` is not immediately realizing the collection. However, as soon as it is queried, it requires the realization of the tasks it filters. So the `dependsOn(...)` will trigger the realization of the tasks.

If you had done `tasks.withType(Test).matching { ... }`, then the only realized tasks would be of type `Test`.

martinbonnin: That might take us a bit too far but is there a reason matching on `task.name` requires realization while matching on `task.javaClass` doesn't? Looks like there's no way to depend on a task that might or might not be already registered while only knowing its name? 

In the context of this PR, this is most likely not a huge issue as the `dependsOn` is made when configuring a task that's only meant to be called from CI. If everything goes well, chances are that only CI is affected 🤞 . In the long run, we might benefit from something like https://developer.squareup.com/blog/herding-elephants/ so that we can more precisely add the dependencies to our CI tasks (and not have to iterate all subprojects and all tasks as this PR does). Follow up issue there: https://github.com/apollographql/apollo-android/issues/3531

melix: `tasks.matching` is an old API, largely predates lazy configuration. Therefore it doesn't match on the _prototypes_ of tasks (name, type) but on the _realized_ tasks. The `withType` is more recent and creates a view which only requires the types. Similarly, `named` only requires the names, so `named(name, type)` works lazily in both cases (because it only needs the prototype which is used in `register`).

Issue #3490, opened 2021-11-03
v3 missing documentation


BoD:
- [x] Migration guide
- [x] RxJava
- [x] Section "Reading from the store programmatically"
- [x] Declarative cache
- [x] Imperative cache
- [x] File Upload
- [x] Test models
- [x] espresso
- [x] `apollo-ast`
- [x] Introspection and SDL schemas
- [x] Add `dataOrThrow` to the error handling page
- [x] Remove references to Spectrum / replace with Kotlin Slack
- [x] Update structure of the pages / left bar / getting started
- [x] Query Batching

martinbonnin:
Looks like we're done 🎉

PR #3529, opened 2021-11-09
Remove spectrum mentions and update Contributing.md

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/3490

- remove references to Spectrum
- update CONTRIBUTING.md


BoD:
👍

Note: there is also a "Discuss on Spectrum" link on the right panel showing on all pages - but couldn't immediately find where it comes from - I think it must be a Gatsby configuration?

martinbonnin:
> Note: there is also a "Discuss on Spectrum" link on the right panel showing on all pages - but couldn't immediately find where it comes from - I think it must be a Gatsby configuration?

This is most likely because `dev-3.x` has the older apollo theme when running locally. When running from netlify, it's going to use the version from `main` that doesn't reference Spectrum

BoD (approved):

PR #3530, opened 2021-11-09
Add a documentation about .json, .graphqls and .graphql

martinbonnin:


docs/source/essentials/file-types.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: File types
+---
+
+> To learn more about SDL schemas, you can also read the [blog post](https://www.apollographql.com/blog/frontend/sdl-support-in-apollo-android/) about SDL support in Apollo Android
+
+Apollo Android supports the following file types:
+
+* `.json` for introspection schemas
+* `.graphqls` for SDL schemas
```
BoD, line 10: If I'm not mistaken `.sdl` is also supported, but maybe we don't want to recommend it in the doc?

BoD (approved):
👍

docs/source/essentials/file-types.mdx
martinbonnin: Indeed, `.sdl` isn't really supported by any other tool so I'd like users to move to `.graphqls` as much as possible. I'll still add a note in the SDL section 👍

PR #3534, opened 2021-11-10
Tweak the errors doc and add ApolloResponse.dataAssertNoErrors

martinbonnin:
The doc was mostly up to date, mostly cosmetic changes.

I took this opportunity to rename `dataOrThrow` -> `dataAssertNoErrors` that I feel conveys what it's doing more explicitely.

### DIFF
--- a/docs/source/essentials/01-errors.mdx
+++ b/docs/source/essentials/01-errors.mdx
@@ -1,6 +1,5 @@
 ---
-title: Error handling in Apollo Android
-sidebar_title: Errors
+title: Error handling
 ---
 
 Whenever you execute a GraphQL operation with Apollo Android (or any other GraphQL client), two high-level types of errors can occur:
@@ -21,7 +20,7 @@ try {
 }
 ```
 
-### Possible causes
+**Non-exhaustive list of possible causes:**
 
 * The app is offline or doesn't have access to the network.
 * A DNS error occurred, making it impossible to look up the host.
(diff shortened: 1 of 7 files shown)


docs/source/essentials/01-errors.mdx
```diff
@@ -103,7 +103,9 @@ if (person != null) {
 }
 ```
 
-If you don't want to handle partial responses, you can use `dataOrThrow()`. This returns a non-nullable `data` and throws if there are any errors:
+## Ignoring partial data
+
+If you don't want to handle partial responses, you can simplify your error handling logic.`ApolloResponse.dataOrThrow` will return a non-nullable `data` and throws if there are any errors so that you can handle all eroors in a single `catch {}` block. :
```
BoD, line 108:
```suggestion
If you don't want to handle partial responses, you can simplify your error handling logic. `ApolloResponse.dataAssertNoErrors` will return a non-nullable `data` or throw if there are any errors so that you can handle all errors in a single `catch {}` block:
```

BoD (approved):

docs/source/essentials/01-errors.mdx
martinbonnin: Spelling mistakes aside, this was made on purpose so that we can merge this PR before we release beta04. The `dataAssertNoErrors` documentation changes are done in https://github.com/apollographql/apollo-android/pull/3510/commits/8a66bbdbaec3fac64c5a6c7864d1895a53147ea2

BoD: Oh I see! Makes sense. Yeah having the doc published before the new versions is tricky! 

But other instances of `dataOrThrow` were updated as part of this PR 🤔

PR #3537, opened 2021-11-10
Use only one VM for the push workflow

martinbonnin:
Contrary to the PR workflow, the Push workflow doesn't need to be fast. Run everything in the same machine so that it uses less CPU time overall (no need to checkout 3 times, run dokka 3 times, etc...)

Also, the Push workflow doesn't cancel the previous ones as it can be usefull to have all SNAPSHOTs if we ever need to bisect a regression

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,7 +97,7 @@ following events:
 
 **Workflow:** [`pr.yml`](https://github.com/apollographql/apollo-android/blob/dev-3.x/.github/workflows/pr.yml)
 
-**Jobs:**
+**Jobs (run in parallel):**
 
 - `tests-gradle`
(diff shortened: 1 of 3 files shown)


BoD (approved):

CONTRIBUTING.md
```diff
@@ -113,22 +113,20 @@ _(Will be replaced to "on pushes to `main` branch" after v3 is released.)_
 
 **Workflow:** [`push.yml`](https://github.com/apollographql/apollo-android/blob/dev-3.x/.github/workflows/push.yml)
 
-**Jobs:**
+**Job:**
 
-- `tests-all`
+- `push`
```
martinbonnin, line 118: 👍

PR #3538, opened 2021-11-10
Add "UI testing" and remove flowDecorators

martinbonnin:
flowDecorators duplicated the `ApolloInterceptor` functionality so let's only use `ApolloInterceptors`.

One subtle change is than now the dispatcher is the responsibility of individual interceptors. This is needed so that the IdlingResource interceptor can increment the busy count synchronously from the main thread. The network and cache interceptors now set `flowOn(dispatcher)`. I'm expecting the impact of that to be relatively small as the dispatcher stays the same and it should **not** create an extra buffer but that's something to keep in mind.


apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt
```diff
@@ -38,13 +45,20 @@ class ApolloIdlingResource(
...
-  return addFlowDecorator {
-    it.onStart {
-      idlingResource.operationStart()
-    }.onCompletion {
-      idlingResource.operationEnd()
+  return addInterceptor(object : ApolloInterceptor {
+    override fun <D : Operation.Data> intercept(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
+      // Do not update the idling resource on subscriptions as they will never terminate
+      return if (request.operation !is Subscription) {
+        chain.proceed(request).onStart {
```
BoD, line 52: TIL `onStart`! 👍

docs/source/advanced/ui-tests.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Apollo Android offers a built-in [IdlingResource](https://developer.android.com/reference/androidx/test/espresso/IdlingResource) to help writing UI tests with Espresso. The `ApolloIdlingResource` will make sure that your tests wait for your GraphQL queries terminate before moving on with testing.
+
+Add the `apollo-idling-resource` `dependency`:
+
+```kotlin:title=build.gradle[.kts]
+implementation("com.apollographql.apollo3:apollo-idling-resource:x.y.z")
+```
+
+> Note: If you have multiple `ApolloClients` you need to create and register multiple `ApolloIdlingResource` with different names. Registering several IdlingResources with the same name will crash.
```
BoD, line 15: 💡 Random idea: we could provide a default name like `ApolloIdlingResource-x` (x being a counter) .

apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt
```diff
@@ -38,13 +45,20 @@ class ApolloIdlingResource(
 }
 
 fun ApolloClient.Builder.idlingResource(idlingResource: ApolloIdlingResource): ApolloClient.Builder {
-  return addFlowDecorator {
-    it.onStart {
-      idlingResource.operationStart()
-    }.onCompletion {
-      idlingResource.operationEnd()
+  return addInterceptor(object : ApolloInterceptor {
```
BoD, line 48: Ideally, it could be nice to disallow calling `idlingResource` several times on the same builder - I don't see an obvious way to implement it. May be over the top anyway :)

BoD (approved):
👍

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -263,15 +254,6 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
       this.interceptors += interceptors
     }
 
-    fun addFlowDecorator(flowDecorator: FlowDecorator) = apply {
```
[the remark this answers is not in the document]
martinbonnin: I think we should be safe. That's kind of an advanced API. If anyone is using that already, they shouldn't have too many issues using interceptors instead.

apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt
martinbonnin: We could make `ApolloClient.Builder.interceptors` public and check for an instance of a previously registered interceptor. That's a tiny performance hit but certainly negligible (and only used in tests anyways)

BoD: Implemented it [here](https://github.com/apollographql/apollo-android/pull/3546) - not 100% convinced, so feel free to close if you think it's too much 😄

docs/source/advanced/ui-tests.mdx
martinbonnin: Let's see if that comes up as an issue before adding it?

apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt
martinbonnin: Merged!

PR #3542, opened 2021-11-12
Relocate the kotlin-stdlib in the Gradle Plugin

martinbonnin:
This is mainly an exercise to see what it takes to relocate the kotlin-stdlib and be able to use latest Kotlin versions regardless of the Gradle version used (see [issue #16345](https://github.com/gradle/gradle/issues/16345)  on the Gradle repo)

This PR decouples the Gradle and Apollo Kotlin versions. That's for the advantages. In terms of drawbacks, it's fragile as there could be edge cases or side effects that weren't caught by tests (but so far so good 🤞). It's also adding ~1min of build time for every change to the compiler or Gradle plugin because of R8 needing to be run again. 

The meat of the PR is the `apollo-gradle-build` build.gradle.kts and rules.pro files. There are also changes in the compiler required to remove reflection from the codegen. For more information, you can also read the [blog post](https://blog.mbonnin.net/use-kotlin-15-in-your-gradle-plugins)


apollo-gradle-plugin/build.gradle.kts
```diff
@@ -1,114 +1,73 @@
...
 
   testImplementation(groovy.util.Eval.x(project, "x.dep.junit"))
   testImplementation(groovy.util.Eval.x(project, "x.dep.truth"))
   testImplementation(groovy.util.Eval.x(project, "x.dep.assertj"))
   testImplementation(groovy.util.Eval.x(project, "x.dep.okHttp.mockWebServer4"))
 }
 
-val shadowJarTask = tasks.named("shadowJar", com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar::class.java)
-val shadowPrefix = "com.apollographql.relocated"
+if (true) {
```
BoD, line 44: Is this to quickly disable it when working on the plugin?

BoD (approved):
👏

BoD (approved):

apollo-gradle-plugin/build.gradle.kts
martinbonnin: Yep, exactly!

PR #3545, opened 2021-11-12
Allow Delete/Continue/Replace during AST transformations

martinbonnin:
- allow transform to return one of Delete/Continue/Replace
- use transform {} to add required fields


BoD (approved):
🤝

PR #3553, opened 2021-11-15
Add a doc for Query Batching

BoD:
Related to #3490.

(targets `update-readme` to avoid a few conflicts)

### DIFF
--- a/docs/source/advanced/query-batching.mdx
+++ b/docs/source/advanced/query-batching.mdx
@@ -1,0 +1,51 @@
+---
+title: Query batching
+---
+
+Apollo Android supports batching queries as a means to reduce the number of HTTP round trips. When enabled, operations
+executed consecutively in a short period of time will be sent to the server together in a single HTTP request.
+
+Note that this network optimization comes at the price of increased latency as every request in the batch is now as slow
+as the slowest one.
+
+Here is how to enable batching:
+
+```kotlin{5}
+val apolloClient = ApolloClient.Builder()
+    .networkTransport(
+        HttpNetworkTransport.Builder()
+            .serverUrl("https://...")
+            .httpEngine(BatchingHttpEngine())
+            .build()
+    )
+    .build()
+```
+
+By default, batched operations will be sent after 10 ms, or when 10 of them have been accumulated, but this can be
+configured:
+
+```kotlin
+BatchingHttpEngine(
+    batchIntervalMillis = 1000,
+    maxBatchSize = 20
+)
+```
+
+## Opting in/out
+
+If you need to disable batching for a specific operation, you can do so by using the `canBeBatched` extension:
+
+```kotlin
+apolloClient.query(MyQuery()).canBeBatched(false).execute()
+```
+
+The default behavior is to have batching enabled for all operations, unless opting-out. If instead you want it to be
+disabled by default and enable it only for specific operations, you can do so by using the `batchByDefault` parameter:
+
+```kotlin
+BatchingHttpEngine(
+    batchByDefault = false
+)
+```
+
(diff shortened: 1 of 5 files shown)


docs/source/advanced/query-batching.mdx
```diff
@@ -0,0 +1,51 @@
...
+as the slowest one.
+
+Here is how to enable batching:
+
+```kotlin{5}
+val apolloClient = ApolloClient.Builder()
+    .networkTransport(
+        HttpNetworkTransport.Builder()
+            .serverUrl("https://...")
+            .httpEngine(BatchingHttpEngine())
```
martinbonnin, line 18: There is `ApolloClient.Builder.httpEngine()` accessible

docs/source/advanced/query-batching.mdx
```diff
@@ -0,0 +1,51 @@
...
+    )
+    .build()
+```
+
+By default, batched operations will be sent after 10 ms, or when 10 of them have been accumulated, but this can be
+configured:
+
+```kotlin
+BatchingHttpEngine(
+    batchIntervalMillis = 1000,
```
martinbonnin, line 29: That's a huge interval. Maybe set `50` ?

docs/source/advanced/query-batching.mdx
```diff
@@ -0,0 +1,51 @@
...
+```kotlin
+apolloClient.query(MyQuery()).canBeBatched(false).execute()
+```
+
+The default behavior is to have batching enabled for all operations, unless opting-out. If instead you want it to be
+disabled by default and enable it only for specific operations, you can do so by using the `batchByDefault` parameter:
+
+```kotlin
+BatchingHttpEngine(
+    batchByDefault = false
```
martinbonnin, line 47: Thinking out loud. Should we remove this and instead allow `ApolloClient.canBeBatched(Boolean)` instead?

docs/source/advanced/query-batching.mdx
BoD: Oh even better! 👍

docs/source/advanced/query-batching.mdx
BoD: 💡 I think that's a good idea! That would simplify the API a bit, while also making it similar to the other APIs like `fetchPolicy` and so on. 

I'll have a look!

BoD: Done in the last commit - in the end it was just a matter of removing the constructor parameter since `canBeBatched` was already on `HasMutableExecutionContext`.

While testing this I noticed something: adding a header with `httpHeader` on `ApolloClient.Builder`  and then adding another one with `httpHeader` on an `ApolloCall` will result on the first one being ignored. This is because they are not merged but replaced. This could be surprising to users? (not sure if it would be easy to fix though)

martinbonnin: I'd vote to rename `HasMutableExecutionContext.httpHeader` to `HasMutableExecutionContext.addHttpHeader` for consistency with interceptors, etc...
And do nothing to merge `ApolloClient` and `ApolloCall` contexts as that would add yet another level of logic. Whatever is on `ApolloCall` overrides the contexts on `ApolloClient`

martinbonnin (approved):

docs/source/advanced/query-batching.mdx
BoD: > Whatever is on ApolloCall overrides the contexts on ApolloClient

👍  simple rule, I like :)

About `addHttpHeader`, I'll do it right now on a separate PR

martinbonnin: Usual stuff: keep @deprecated `httpHeader` until 3.0.0?

Issue #3152, opened 2021-06-06
Provide module information (JPMS)


tobiasdiez:
**Is your feature request related to a problem? Please describe.**
Currently it very hard to use the apollo library in a modularized application. For example, there are split package issues between apollo-runtime and apollo-api. I'll describe these problems further below in more detail (along with a workaround).

**Describe the solution you'd like**
The usage would be way easier if the apollo library would be modularized, or at least could provide `module-info.class`.

**Problems & Workarounds**
1. Adding apollo-runtime leads to split package problems like the following
```
error: module java.xml.bind reads package com.apollographql.apollo from both apollo.api.jvm and apollo.runtime
error: module java.xml.bind reads package com.apollographql.apollo.exception from both apollo.api.jvm and apollo.runtime
```
Fixed by excluding the sub-dependencyies
```
    implementation ("com.apollographql.apollo:apollo-runtime:2.5.8"){
        exclude module: "apollo-api"
        exclude module: "apollo-normalized-cache"
    }
```
and then downloading the `apollo-api-jvm.x.jar` file manually, and merging with apollo-runtime by `modularity.patchModule("apollo.runtime", "apollo-api-jvm-2.5.8.jar")`  using the gradle plugin `org.javamodularity.moduleplugin`.

2. Apollo accesses internal apis
Fixed by adding `add-exports` for
```
                'apollo.runtime/com.apollographql.apollo.ap = org.your_module',
                'apollo.runtime/com.apollographql.apollo.api.internal = org.your_module'
```
3. Access problems
```
error: cannot access Function1
        final GetUserDocumentRaw getUserDocumentRaw = reader.readObject($responseFields[0], new ResponseReader.ObjectReader<GetUserDocumentRaw>() {
                                                            ^
  class file for kotlin.jvm.functions.Function1 not found
```
Fixed by adding `requires kotlin.stdlib;` to one's module-info file. (This is probably something one still has to to even with a proper module description from apollo's side; but it should be mentioned in the documentation.)

martinbonnin:
@tobiasdiez I ended up setting up Apollo Android 3 as automatic modules in https://github.com/apollographql/apollo-android/pull/3558. Going the full `module-info.class` would be nice but that'd come with a lot of complexity. We would basically need to publish 2 versions of the artifacts. One with compiled with Java 8 and the other with Java9+

[...]

PR #3558, opened 2021-11-16
JPMS support

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/3152

This provides basic support for the [Java Platform Module System (JPMS)](https://cr.openjdk.java.net/~mr/jigsaw/spec/). 

It does so by setting "Automatic-Module-Name" in the jar manifests, meaning the plain jar can be put in the `--module-path` like other modules. The nice thing with automatic modules is that they export all their packages. I found [this stackoverflow answer](https://stackoverflow.com/a/46742802/752630) to be a good explanation.

A modular application can then use Apollo Android with a `module-info.java` like below:

```java
module com.example.app {
    requires com.apollographql.apollo3.api;
    requires com.apollographql.apollo3.rx2;
    requires com.apollographql.apollo3.runtime;
    requires com.apollographql.apollo3.cache.normalized;
    requires com.apollographql.apollo3.cache.normalized.api;
    requires com.apollographql.apollo3.cache.normalized.sql;
    requires kotlin.stdlib;
}
```

Some other fixes related to using Apollo Android in a Java only project: 

- Fix using the plugin if no Kotlin plugin is in the classpath
- make rx2/rx3 automatic modules + add some @JvmOverloads
- reorganize apollo-normalized-cache vs apollo-normalized-cache-api to remove split package

Breaking changes:

A lot of classes moved from `com.apollographql.apollo3.cache.normalized` to `com.apollographql.apollo3.cache.normalized.api`:

- NormalizedCache
- Record
- ObjectIdGenerator
- ...

CI:
Part of the CI now runs of Java11 to run the `jpms` test


.github/workflows/pr.yml
```diff
@@ -84,3 +84,28 @@ jobs:
         with:
           name: tests-integration.zip
           path: diagnostics.zip
+  tests-java9:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: set up JDK 1.8
```
BoD, line 91:
```suggestion
      - name: set up JDK 11
```

.github/workflows/push.yml
```diff
@@ -10,17 +10,20 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: set up JDK 1.8
```
BoD, line 12:
```suggestion
      - name: set up JDK 11
```

BoD (approved):

PR #3561, opened 2021-11-16
Edits to file types article

StephenBarlow:


docs/source/essentials/file-types.mdx
```diff
@@ -70,25 +58,33 @@ type schema {
...
 
 type Query {
-  field: String
+  field: String @deprecated
   ...
 }
 ```
 
-This format is more compact and easier to read and modify. Another advantage is that the SDL format carries GraphQL directives, which the Json format doesn't (see [this issue](https://github.com/graphql/graphql-spec/issues/300))
+You can't use introspection to download an SDL schema directly from a GraphQL endpoint. However, Apollo Android can convert an introspection schema to SDL automatically if you specify the `.graphqls` extension in the `--schema` option:
```
martinbonnin, line 66:
```suggestion
You can't use introspection to download a SDL schema directly from a GraphQL endpoint. However, Apollo Android can convert an introspection schema to SDL automatically if you specify the `.graphqls` extension in the `--schema` option:
```

martinbonnin (approved):
🤩 🤩 🤩

docs/source/essentials/file-types.mdx
StephenBarlow: Ah this is correct as-is, because "SDL" is pronounced "Ess Dee Ell", so it needs "an" instead of "a"

martinbonnin: TIL, Thanks!

PR #3562, opened 2021-11-17
Edits to Apollo AST article

StephenBarlow:


docs/source/advanced/apollo-ast.mdx
```diff
@@ -136,50 +127,54 @@ val schemaIssues = schemaGQLDocument.validateAsSchema()
 println(schemaIssues.map { it.severity.name + ": " + it.message })
 ```
 
-This will print `[WARNING: Unknown directive 'FooBar']`.
+When excuted, this snippet prints `[WARNING: Unknown directive 'private']`.
 
-The `validateAsExecutable` method ensures the operations are consistent with a schema, that is why it takes an instance
-of `Schema` as a parameter. This `Schema` can be obtained by calling the `GQLDocument.toSchema()` extension:
```
StephenBarlow, line 142: I'm not sure in Kotlin-land when to refer to something as a method vs. an extension. Is this library adding custom methods to a class that's defined elsewhere?

BoD: 👋  Yes you got it, extensions are when a method (... which are also called functions in Kotlin) is defined not in the class itself, but outside of it. They are used to augment an API you're not the author of (external lib), or sometimes just to organize your code in different modules / files, which is the case here.

To be accurate I used the term extension here, but I think we could have just said method and that would have worked too 😅

martinbonnin: I guess the proper term is "extension function": https://kotlinlang.org/docs/extensions.html#extension-functions

martinbonnin (approved):

docs/source/advanced/apollo-ast.mdx
```diff
@@ -136,50 +127,54 @@ val schemaIssues = schemaGQLDocument.validateAsSchema()
...
+[WARNING: Use of deprecated field 'height', ERROR: Can't query 'frends' on type 'Character']
+```
 
-This method takes as a parameter a lambda that provides a `GQLNode` and returns the transformed node (or `null` to
-remove the node). It will then visit all nodes in the AST, apply the lambda to each one, and replace the node with the
-result.
+## Outputting SDL
 
-For instance, here's how to remove any field named `foobar` from the selections:
+You can output a `GQLDocument` to standard GraphQL SDL with the `toUtf8` extensions:
```
martinbonnin, line 150: Nit, I'd remove "SDL" up here as technically operations are not SDL (as they're not schemas obvisouly). Actually, I'm not exactly sure if SDL is something inside the spec.

```suggestion
You can output a `GQLDocument` to standard GraphQL with the `toUtf8` extensions:
```

BoD, line 150:
```suggestion
You can output a `GQLDocument` to standard GraphQL syntax with the `toUtf8` extensions:
```

(as it's not just for SDL)

martinbonnin (approved):

docs/source/advanced/apollo-ast.mdx
martinbonnin: I'll resolve my comment above 😄

BoD: Haha woops we got the same reaction almost at the same minute! 😅 🤝

BoD (approved):
👍 🙏

StephenBarlow:
Thanks for the review everyone, and for making the followup edit @BoD! I was out the last couple days, ready to jump back into more of these 😄

Issue #3395, opened 2021-10-07
Remove and/or propagate experimental API usages


martinbonnin:
`dev-3.x` uses opt-in annotations such as `@OptIn(ExperimentalCoroutinesApi::class)` and `@OptIn(ExperimentalContracts::class)`. We should either add a `ApolloExperimental` or remove these before going stable to make sure they do not break compatibility.

BoD:
## TLDR

- remove 2 instances of `ExperimentalUnsignedTypes`, and `ExperimentalCoroutinesApi` on `ApolloClient.executeAsFlow` (no impact)
- mark `OkHttpWebSocketEngine` private
- remove `DelicateCoroutinesApi` on `runTest` (`runWithMainLoopApple` and `runWithMainLoopJs`): update code to not use `GlobalScope`
- add `ApolloExperimental` on `MainLoopDispatcher` in `appleMain`
- `WebSocketNetworkTransport.execute`: not sure what to do!

## More in-depth

Here's a list of all places where `@OptIn` is used and the proposed action to take:

|What                                                                           |Where                            |Why                                                                                                     |What to do                  |More info                                                                                                  |
|-------------------------------------------------------------------------------|---------------------------------|--------------------------------------------------------------------------------------------------------|----------------------|-----------------------------------------------------------------------------------------------------------|
|@OptIn(ApolloInternal::class)                                                  |CustomScalarAdapters             | Version2CustomTypeAdapterToAdapter needs to be public but is marked as internal so clients don't use it|Keep                  |There on purpose                                                                                           |
|@OptIn(ApolloInternal::class)                                                  |ApolloClient.Builder             | Version2CustomTypeAdapterToAdapter needs to be public but is marked as internal so clients don't use it|Keep                  |There on purpose                                                                                           |
|@OptIn(ExperimentalContracts::class)                                           |JsonWriters                      |Use of contract                                                                                         |Keep                  |On internal API                                                                                            |
|@OptIn(ExperimentalTime::class)                                                |CodegenTest                      |Use of Duration to monitor generation times                                                             |Keep    |In Tests. (Also, I think it's stable in Kotlin 1.6)                                                                                                   |
|@OptIn(ExperimentalStdlibApi::class)                                           |CodegenTest                      |Use of buildList                                                                                        |Keep   |In Tests (Also, I think it's stable in Kotlin 1.6)                                                                                                   |
[...]

PR #3565, opened 2021-11-17
Handle experimental opt ins

BoD:
See #3395

- [x] remove 2 instances of `ExperimentalUnsignedTypes`, and `ExperimentalCoroutinesApi` on `ApolloClient.executeAsFlow` (no impact)
- [x] refact `NSURLSessionWebSocketEngine` / `OkHttpWebSocketEngine` to avoid `invokeOnClose`, then remove annotation
- [x] refact `WebSocketNetworkTransport.execute` to avoid `transformWhile`, then remove annotation
- [ ] Apple's `DefaultDispatcher` relying on `Delay`: not sure what to do!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,7 +66,7 @@ files.
 ## Experimental / internal APIs
 
 Using Kotlin's (or other dependencies') experimental or internal APIs, such as the ones marked
-with `@ExperimentalCoroutinesApi` should be avoided as much as possible (exceptions can be made when no other option is
+with `@ExperimentalCoroutinesApi` should be avoided as much as possible (exceptions can be made for native/JS targets only when no other option is
 available). Indeed, applications using a certain version of Apollo Android could use a more up-to-date version of these
 APIs than the one used when building the library, causing crashes or other issues.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,17 @@ allowed and any public API change will fail the build.
 If that happens, you will need to run `./gradlew apiDump` and check for any incompatible changes before commiting these
 files.
 
+## Experimental / internal APIs
+
+Using Kotlin's (or other dependencies') experimental or internal APIs, such as the ones marked
+with `@ExperimentalCoroutinesApi` should be avoided as much as possible (exceptions can be made when no other option is
+available). Indeed, applications using a certain version of Apollo Android could use a more up-to-date version of these
+APIs than the one used when building the library, causing crashes or other issues.
+
+We also have the `@ApolloExperimental` annotation which can be used to mark APIs as experimental, for instance when
+feedback is wanted from the community on new APIs. This can also be used as a warning that APIs are using experimental
+features of Kotlin/Coroutines/etc. and therefore may break in certain situations.
+
 ## Releasing
 
 Releasing is done using Github Actions. The CI contains credentials to upload artifacts to Sonatype and the Gradle
--- a/apollo-runtime/api/apollo-runtime.api
+++ b/apollo-runtime/api/apollo-runtime.api
@@ -377,12 +377,6 @@ public final class com/apollographql/apollo3/network/ws/GraphQLWsProtocol$Factor
 	public fun getName ()Ljava/lang/String;
 }
 
-public final class com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine : com/apollographql/apollo3/network/ws/WebSocketEngine {
-	public fun <init> ()V
-	public fun <init> (Lokhttp3/WebSocket$Factory;)V
-	public fun open (Ljava/lang/String;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-}
-
 public final class com/apollographql/apollo3/network/ws/OkHttpWebSocketEngineKt {
 	public static final fun WebSocketEngine ()Lcom/apollographql/apollo3/network/ws/WebSocketEngine;
 }
--- a/apollo-runtime/api/apollo-runtime.api
+++ b/apollo-runtime/api/apollo-runtime.api
@@ -377,6 +377,12 @@ public final class com/apollographql/apollo3/network/ws/GraphQLWsProtocol$Factor
 	public fun getName ()Ljava/lang/String;
 }
 
+public final class com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine : com/apollographql/apollo3/network/ws/WebSocketEngine {
+	public fun <init> ()V
+	public fun <init> (Lokhttp3/WebSocket$Factory;)V
+	public fun open (Ljava/lang/String;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+}
+
 public final class com/apollographql/apollo3/network/ws/OkHttpWebSocketEngineKt {
 	public static final fun WebSocketEngine ()Lcom/apollographql/apollo3/network/ws/WebSocketEngine;
 }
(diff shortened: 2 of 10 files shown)


martinbonnin:
Re: Delay, could we use `Dispatchers.Main` instead? It wasn't in the coroutines artifact for a while but maybe it's in `1.5.2` now ? Or in the `-mt` version?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/flows.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * This is copied from [Flow], which is marked [ExperimentalCoroutinesApi].
+ * There is a risk this API will be removed or changed in the future, which could break consumers of this library
+ * - that is why we use our own copy.
+ *
+ * TODO: remove when kotlinx.coroutines.flow.Flow.transformWhile is no longer marked ExperimentalCoroutinesApi.
+ */
+@ApolloInternal
+internal fun <T, R> Flow<T>.transformWhile(
```
martinbonnin, line 19: Do we need `@ApolloInternal` on something that's already `internal`?

BoD: Nope 😅  I realized a bit later it was useless - will be gone on the next commit

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/NSURLSessionHttpEngine.kt
```diff
@@ -95,7 +95,6 @@ class NSURLSessionHttpEngine(
   }
 }
 
-@OptIn(ExperimentalUnsignedTypes::class)
```
martinbonnin, line 98: 😎  TIL this was promoted to Stable in 1.5 🎉  https://kotlinlang.org/docs/whatsnew15.html#stable-unsigned-integer-types

BoD:
> Re: Delay, could we use Dispatchers.Main instead? It wasn't in the coroutines artifact for a while but maybe it's in 1.5.2 now ? Or in the -mt version?

So I played around with this and `Dispatchers.Main` does exist _and_ works _when using the `-mt` version_. (Without `-mt`, native tests crash with `kotlin.IllegalStateException: There is no event loop. Use runBlocking { ... } to start one.`).

Now I must admit I'm not well versed enough in KMP to appreciate the full extent / subtitles of what this means - can we "just" use `-mt`? 😅 So far I didn't find many resources explaining the drawbacks of doing so (except it should no longer be needed when the [new memory model](https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/) will be stable, and "memory leaks in some concurrent execution scenarios").

martinbonnin:
We historically tried to avoid the `-mt` version of coroutines for all the reasons that your mention (memory leaks?, compatibility issues?, other?). 

If there's no way around this then I guess we're stucked with `@OptIn(InternalCoroutinesApi)` and the day it crashes release an update. I think that wouldn't be the first time a K/N lib would have an incompatibility with `kotlinx.coroutines`

BoD:
The good news is that 
- it's the only case of using this `OptIn` (albeit it's on an "important" class 😅)
- hopefully we'll be able to get rid of it "soon"

martinbonnin:
We "just" have to enable the new memory model 🙈

BoD:
Added a paragraph about this topic in Contributing.

CONTRIBUTING.md
```diff
@@ -63,6 +63,17 @@ allowed and any public API change will fail the build.
 If that happens, you will need to run `./gradlew apiDump` and check for any incompatible changes before commiting these
 files.
 
+## Experimental / internal APIs
+
+Using Kotlin's (or other dependencies') experimental or internal APIs, such as the ones marked
+with `@ExperimentalCoroutinesApi` should be avoided as much as possible (exceptions can be made when no other option is
+available). Indeed, applications using a certain version of Apollo Android could use a more up-to-date version of these
```
martinbonnin, line 70:
```suggestion
with `@ExperimentalCoroutinesApi` should be avoided as much as possible (exceptions can be made for native/JS targets only when no other option is
available). Indeed, applications using a certain version of Apollo Android could use a more up-to-date version of these
```

apollo-runtime/api/apollo-runtime.api
```diff
@@ -377,12 +377,6 @@ public final class com/apollographql/apollo3/network/ws/GraphQLWsProtocol$Factor
 	public fun getName ()Ljava/lang/String;
 }
 
-public final class com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine : com/apollographql/apollo3/network/ws/WebSocketEngine {
-	public fun <init> ()V
-	public fun <init> (Lokhttp3/WebSocket$Factory;)V
-	public fun open (Ljava/lang/String;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-}
```
martinbonnin, line 384: Can we keep this public in case someone wants to customize their `WebSocket.Factory`. Also for symmetry with `OkHttpEngine`. Then users can use them from `ApolloClient.Builder.webSocketEngine()`

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/ChannelWrapper.kt
```diff
@@ -0,0 +1,33 @@
+package com.apollographql.apollo3.internal
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.Channel
+
+/**
+ * A wrapper for [Channel] that adds a method equivalent to [invokeOnClose], which is marked [ExperimentalCoroutinesApi].
```
martinbonnin, line 7: Is this copied from `kotlinx.coroutines` ? Maybe add a link to the source if it's the case?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/flows.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flow
+
+/**
+ * This is copied from [Flow], which is marked [ExperimentalCoroutinesApi].
```
martinbonnin, line 11: Same as for `ChannelWrapper`, we should link the source for attribution + remember where we got it from in 6 months.

BoD: Good point!

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/ChannelWrapper.kt
BoD: This is not exactly copied (I just took `invokeOnClose` + simplified it) but good point I'll link to it 👍

apollo-runtime/api/apollo-runtime.api
BoD: Not sure why I thought I saw the apple one was internal, but no. Reverting 👍

martinbonnin (approved):

Issue #3570, opened 2021-11-18
Add migration helpers for useHttpGetMethodForQueries and useHttpGetMethodForPersistedQueries


martinbonnin:

BoD:
Resolved in #3578

PR #3578, opened 2021-11-19
Add 2.x compat mode for useHttpGetMethodForQueries and useHttpGetMethodForPersistedQueries

BoD:
See #3570.

Honestly I'm not sure about `useHttpGetMethodForPersistedQueries` as the semantics are not the same. Maybe the fact that it's deprecated is enough but maybe that could create more problems than it's solving?


martinbonnin:
> Honestly I'm not sure about useHttpGetMethodForPersistedQueries as the semantics are not the same. Maybe the fact that it's deprecated is enough but maybe that could create more problems than it's solving?

Excellent point 👍 . Throw inside the method with something like:

```
throw NotImplementedException("useHttpGetMethodForPersistedQueries is now configured at the same time as autopersisted queries. Use autoPersistedQueries(httpMethodForHashedQueries = HttpMethod.GET) instead.")
```

martinbonnin (approved):

PR #3584, opened 2021-11-22
Doc edits to batching, codegen, and some reorg

StephenBarlow:
Some various edits I got through at the end of last week, more to come the next couple days and the week following


docs/gatsby-config.js
```diff
@@ -51,19 +57,17 @@ module.exports = {
             'caching/http-cache',
             'caching/troubleshooting',
           ],
-          Advanced: [
+          Networking: [
+            'advanced/interceptors-http',
+            'advanced/http-engine',
+            'advanced/no-runtime',
```
martinbonnin, line 63: I love the "Networking" section!

docs/gatsby-config.js
```diff
@@ -51,19 +57,17 @@ module.exports = {
...
+          ],
+          'Advanced Config': [
```
martinbonnin, line 65: I'm not a huge fan of `Advanced Config` vs `Advanced Operations` sections as I'm not sure what the difference is. Also I feel we need to educate about is "operations" (in the GraphQL acceptation) and using it for things like "batching" or "upload" feels confusing.

docs/source/advanced/http-engine.mdx
```diff
@@ -59,25 +60,29 @@ class MyHttpEngine(val wrappedClient: MyClient) : HttpEngine {
...
 
-This is everything needed. Your GraphQL queries will then use `MyHttpEngine` to execute the HTTP requests.
+With this configuration, Apollo Android sends all of its GraphQL operation requests with `MyHttpEngine`.
 
+## Other HTTP customizations
 
+Besides implementing `HttpEngine`, Apollo Android also supports other methods for customizing HTTP behavior:
 
+* [No runtime](no-runtime): You can opt out of the Apollo Android runtime completely and only use generated models and parsers. Use this option if you don't need any of the runtime features (caching, batching, automatic persisted queries, etc.).
+* [HTTP interceptors](interceptors-http): If you want to add HTTP headers and/or logging to your requests, [HTTP interceptors](interceptors-http) enable you to do this with minimal code.
```
martinbonnin, line 88: That might belong to an `Introduction` or `Overview` page, [like for caching](https://www.apollographql.com/docs/android/v3/caching/introduction/)?

docs/source/advanced/http-engine.mdx
```diff
@@ -1,20 +1,21 @@
 ---
-title: Use your own HTTP client
+title: Using a custom HTTP library
+sidebar_title: Custom HTTP libraries
```
martinbonnin, line 3: I'd keep the "HTTP client" terminology as it's been widely popularized by "OkHttpClient"  and this feels more specific than "library"

docs/source/advanced/http-engine.mdx
```diff
@@ -59,25 +60,29 @@ class MyHttpEngine(val wrappedClient: MyClient) : HttpEngine {
...
+This example uses an asynchronous `wrappedClient` that runs the network request in a separate thread. Note that because `HttpEngine.execute` itself is called from a background thread, you can safely block in `execute()`.
 
-## Use your HttpEngine
+## Using your `HttpEngine`
 
-Once you have implemented your HttpEngine, you can register it in your [ApolloClient](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3/-apollo-client/index.html) using [HttpNetworkTransport](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3.network.http/-http-network-transport/index.html):
+After you create your `HttpEngine` implementation, you can register it with your [`ApolloClient`](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3/-apollo-client/index.html) instance using [`HttpNetworkTransport`](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3.network.http/-http-network-transport/index.html):
```
martinbonnin, line 67: Not directly  related to these edits but we added helper methods to not have to go through `NetworkTransport`:

```suggestion
After you create your `HttpEngine` implementation, you can register it with your [`ApolloClient`](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3/-apollo-client/index.html) instance using [`ApolloClient.Builder.httpEngine`](https://apollographql.github.io/apollo-android/kdoc/apollo-runtime/com.apollographql.apollo3/-apollo-client/-builder/http-engine.html):
```

docs/source/advanced/http-engine.mdx
```diff
@@ -59,25 +60,29 @@ class MyHttpEngine(val wrappedClient: MyClient) : HttpEngine {
...
-    .networkTransport(
-        HttpNetworkTransport(
-            serverUrl = "https://com.example/graphql",
-            engine = MyHttpEngine(wrappedClient)
-        )
+  .networkTransport(
+    HttpNetworkTransport(
+      serverUrl = "https://com.example/graphql",
+      engine = MyHttpEngine(wrappedClient)
     )
```
martinbonnin, line 76: Like above, use `httpEngine` directly:

```suggestion
  .serverUrl(serverUrl = "https://com.example/graphql")
  .httpEngine(httpEngine = MyHttpEngine(wrappedClient))
```

docs/source/advanced/query-batching.mdx
```diff
@@ -1,50 +1,49 @@
 ---
-title: Query batching
+title: Batching operations
```
martinbonnin, line 2: I'm always hesitating on these ones. Same for "Automatic Persisted Queries" vs "Automatic Persisted Operations". Using "Operations" is more correct but there is so much muscle memory around "Query" that I'm always afraid using "Operation" might confuse more than it helps.

docs/source/advanced/response-based-codegen.mdx
```diff
@@ -22,15 +24,20 @@ apollo {
...
-- Fragments are also generated as **interfaces**
-- Any merged fields will appear once in the models
+## The `responseBased` codegen
 
-# Inline fragments
+The `responseBased` codegen differs from the `operationBased` codegen in the following ways:
 
-Let's consider this query:
+* Generated models have a **1:1 mapping** with the JSON structure received in an operation's response.
+* Polymorphism is handled by generating **interfaces**. Possible shapes are then defined as different classes that implement a particular interface.
```
martinbonnin, line 32: A class can implement multiple interfaces:

```suggestion
* Polymorphism is handled by generating **interfaces**. Possible shapes are then defined as different classes that implement the required interfaces.
```

docs/source/advanced/response-based-codegen.mdx
```diff
@@ -61,20 +72,19 @@ when (hero) {
   }
 }
 ```
-## Accessors
 
-As a convenience, methods named `asXyz` returning the possible implementations are also generated. They show up in the IDE's completion, and allow avoiding manual casts. 
+#### Accessors
 
-In our example they can be used like so:
+As a convenience, the `responseBased` codegen generates methods with the name pattern `as<Typename>` (e.g., `asDroid` or `asHuman`) that enable you to avoid manual casting:
```
martinbonnin, line 78: It's not always "Typename". For an example: 

```graphql
query GetAnimal {
  ... on Pet {
    name
  }
  ... on WarmBlooded {
    temperature
  }
}
```

That will have `asPetWarmBlooded()` to avoid generating classes for each and every concrete type. We're [trying to define proper terminology for this](https://apollographql.quip.com/d7KaAOA9f8bz/Client-Glossary) which is not easy. I'd suggest "Shape" for the time being:

```suggestion
As a convenience, the `responseBased` codegen generates methods with the name pattern `as<ShapeName>` (e.g., `asDroid` or `asHuman`) that enable you to avoid manual casting:
```

docs/source/advanced/response-based-codegen.mdx
```diff
@@ -138,11 +148,11 @@ when (hero) {
 }
 ```
 
-## Accessors
+#### Accessors
 
-Accessors are also generated (named after the fragment) for each named fragment that a model can be cast to. 
+As a convenience, the `responseBased` codegen generates methods with the name pattern `<typename>Fields` (e.g., `droidFields` or `humanFields`) that enable you to avoid manual casting.
```
martinbonnin, line 153:
```suggestion
As a convenience, the `responseBased` codegen generates methods with the name pattern `<fragmentName>` (e.g., `droidFields` here or `humanDetails` if you named your fragment "HumanDetails") that enable you to chain calls.
```

BoD (approved):
🙏

docs/gatsby-config.js
StephenBarlow: Ok! I'll revert this and we can keep 🤔ing

docs/source/advanced/http-engine.mdx
StephenBarlow: Agreed! I'll investigate that as a next step, but will leave it here for this PR to make sure the content isn't lost.

docs/source/advanced/query-batching.mdx
StephenBarlow: I hear you! I think this is a fair place to attempt to expand out to "operations" because the reader will have already familiarized themselves with queries. And I think it behooves us in the long run to get folks more comfortable with "operations" so we can _slightly_ disambiguate the term "query"

PR #3585, opened 2021-11-22
Remove coroutines dependency where not needed

martinbonnin:
Only the following modules need coroutines:
- apollo-runtime
- apollo-normalized-cache
- apollo-normalized-cache-sqlite

This PR moves `Guard` and other `runtime` specific classes to `apollo-runtime`. Also adds a few cleanups


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Logger.kt
```diff
@@ -1,22 +0,0 @@
...
-   * @param args extra arguments to pass to the logged message.
-   */
-  fun log(priority: Int, message: String, t: Throwable?, vararg args: Any)
-
-  companion object {
-    const val DEBUG = 3
-    const val WARN = 5
-    const val ERROR = 6
-  }
-}
```
martinbonnin, line 22: We don't use `Logger` any more at this point. I think the only place where it was used was to debug cache misses which is now exposed through `CacheMissException`

build.gradle.kts
```diff
@@ -189,7 +189,10 @@ configure<kotlinx.validation.ApiValidationExtension> {
   ignoredPackages.addAll(
       listOf(
           "com.apollographql.apollo3.api.internal",
+          "com.apollographql.apollo3.internal",
           "com.apollographql.apollo3.gradle.internal",
+          "com.apollographql.apollo3.cache.http.internal",
+          "com.apollographql.apollo3.cache.normalized.api.internal",
           "com.apollographql.apollo3.cache.normalized.internal",
```
martinbonnin, line 196: We should use `nonPublicMarkers.add("com.apollographql.apollo3.ApolloInternal")` instead. Follow up PR incoming

BoD (approved):
👍

PR #3586, opened 2021-11-22
Enable JVM default methods

martinbonnin:
This PR enables `-Xjvm-default=all` in the whole project. This is a big change in the binary API so now it's now or never. 

This forces Java8 but Android supports it (it's on by default since AGP 4.2) and we generate Java8 bytecode already anyways.

Besides removing all the `DefaultImpls` classes, this allows moving some extensions to interfaces, making them way easier to call from Java (See `OperationTests`).

For more background about `@JvmDefault`: https://www.zacsweers.dev/jvmdefault-more-useful/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,8 +44,12 @@ The coding style employed here is fairly conventional Kotlin - indentations are 
 identifiers and methods are camelCased.
 
 * Use primary constructors when there is at most one optional parameter.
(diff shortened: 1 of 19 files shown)


martinbonnin:
```
Inheritance from an interface with '@JvmDefault' members is only allowed with -Xjvm-default option
```

Yikes, that's not cool. Forcing this on every consumer doesn't sound good...

BoD:
Too bad, this was cool 😿

PR #3587, opened 2021-11-22
Move the Operation extensions to Operations.kt

martinbonnin:
Substitute for https://github.com/apollographql/apollo-android/pull/3586. 

Changes Java callsites:

```java
// Before
String json = OperationKt.composeJsonData(query, data, CustomScalarAdapters.Empty, "  ");

// After
String json = Operations.composeJsonData(query, data);
```

* Default methods are blocked on https://youtrack.jetbrains.com/issue/KT-47000
* Adding static method to `Operation` (and not `Operations`) is blocked on https://youtrack.jetbrains.com/issue/KT-11968

The above seems to be an acceptable tradeoff especially as we already have this pattern of using `${ClassName}s` for extensions in Java.


BoD (approved):
👍

PR #3588, opened 2021-11-22
Add apollo-annotations

martinbonnin:
Add a new `apollo-annotations` module. It is very small but required by other modules like `apollo-mockserver` or `apollo-api` for `ApolloExpermental` and `ApolloInternal`


BoD (approved):
👍

PR #3591, opened 2021-11-22
apollo-runtime: Make a few classes internal 

BoD:


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
```diff
@@ -2,14 +2,14 @@ package com.apollographql.apollo3.internal
...
-expect fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher
+internal expect fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher
 
 /**
  * A mutex implementation for code that runs in the defaultDispatcher
  *
  * On native, this is a no-op because the defaultDispatcher is always the main loop
  */
-expect class DefaultMutex(): Mutex
+internal expect class DefaultMutex() : Mutex
```
BoD, line 12: By the way (probably silly question) - why don't we use `kotlinx.coroutines.sync.Mutex`?

martinbonnin: I think initially we used the `DefaultMutex` from non-suspending places but that doesn't seem to be the case anymore. Feel free to change to the coroutines version.

martinbonnin (approved):
Nice 🤩 ! Love it when the api size decreases !

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
BoD: 👍 I will have a look!

PR #3592, opened 2021-11-23
Edits to test builders article and category

StephenBarlow:


docs/source/advanced/test-builders.mdx
```diff
@@ -1,38 +1,32 @@
 ---
-title: 'Test Builders'
+title: Test builders
```
StephenBarlow, line 2: How do we feel about "test builders" not automatically being capitalized? We try to err on the side of not doing it except for titled features and products

docs/source/advanced/test-builders.mdx
```diff
@@ -1,38 +1,32 @@
...
 ---
 
-> Note: Test Builders are experimental and subject to change. We'd love to have your feedback on them. Things that worked well, things that can be improved, etc.. Feel free to reach out either through [Github issues](https://github.com/apollographql/apollo-android/issues/new?assignees=&labels=Type%3A+Bug&template=bug_report.md&title=[Test Builders]) or the [Kotlinlang slack channel](https://app.slack.com/client/T09229ZC6/C01A6KM1SBZ).
```
StephenBarlow, line 5: I couldn't access the kotlinlang slack channel from that URL, so I assumed others wouldn't be able to either. Replaced with a link to the forum

docs/source/advanced/test-builders.mdx
```diff
@@ -70,21 +66,18 @@ fun test() {
 }
 ```
 
-In this example we create a Human Hero with specific values for `firstName` and `age`, and the values for `lastName`
-and `height` are automatically filled with mock values.
+In this example, we create a `HumanHero` with specific values for `firstName` and `age`. The values for `lastName`and `height` are automatically filled with mock values.
```
StephenBarlow, line 69: Couple things here:

* Is `HumanHero` correct as the name of what's created here?
* Is it obvious to Apollo Android users that it's possible to create a `HumanHero` here? Or does this rely on a codegen detail that we should maybe avoid by not using an example that includes fragments?

docs/gatsby-config.js
```diff
@@ -58,6 +58,10 @@ module.exports = {
             'advanced/http-engine',
             'advanced/no-runtime',
           ],
+          Testing: [
```
StephenBarlow, line 61: This seems like a sensible-enough category but happy to revert as well if it doesn't feel right

docs/source/advanced/test-builders.mdx
BoD: 👍 
We were wondering about that - that answers it :)

docs/source/advanced/test-builders.mdx
BoD: Maybe the link works only if you're already a member :( That's too bad. We mention this Slack in other places - We could change the link to the register page (https://slack.kotl.in) instead.

docs/source/advanced/test-builders.mdx
BoD: Yes to the 1st question.  For the 2nd one, we could ditch the polymorphism / use of Fragment for a simpler example  -  however maybe it has value to show it in the doc as this will come up quickly in a real environment?

BoD (approved):
🚀

docs/source/advanced/test-builders.mdx
martinbonnin: Ahah yes we had that same question. I'd go further and ask whether this can be considered a titled feature or not :)

docs/gatsby-config.js
martinbonnin: Yesss 🤩

docs/source/advanced/test-builders.mdx
```diff
@@ -1,38 +1,32 @@
...
 apollo {
   // ...
 
   // Enable test builder generation
   generateTestBuilders.set(true)
-  
-  // Optionally, you may customize the output path
+
+  // Optionally, you can customize the output path
   testDir.set(file("test-builders"))
```
martinbonnin, line 21: I think we can remove that option as it's not needed for most use cases?

docs/source/advanced/test-builders.mdx
```diff
@@ -51,14 +45,16 @@ query HeroForEpisode($ep: Episode!) {
...
 
 ```kotlin
+// Import the query's test builder
 import com.example.test.HeroForEpisodeQuery_TestBuilder.Data
 
 @Test
 fun test() {
   val data = HeroForEpisodeQuery.Data {
+    // Set values for particular fields of the query
     hero = humanHero {
```
martinbonnin, line 58: 👍

docs/source/advanced/test-builders.mdx
martinbonnin: +1 for keeping the Kotlin slack. Also just bumped into this tweet: https://twitter.com/JimSproch/status/1463108061397061637

During developement and for feedback/discussions about experimental features, slack not being indexed is actually a feature.

docs/source/advanced/test-builders.mdx
martinbonnin: One of the values of the test builders is that it's the same test code for both operation-based and response-based codegens. It's all about the GraphQL response. What about:

```
In this example, the `hero` field is a `Human` object with specific values for `firstName` and `age`. The values for `lastName`and `height` are automatically filled with mock values.
```

So overall, the advantages are:
- mocked values
- deduplicates merged fields (mostly happens with fragments)
- prefills __typename (mostly happens with fragments)

Maybe we should reorganize a bit to have the mocked values not using fragments and the rest using them?

docs/source/advanced/test-builders.mdx
BoD: Totally, no objection to remove it.

martinbonnin, line 21:
```suggestion

```

docs/source/advanced/test-builders.mdx
StephenBarlow: I would lean toward no for simplicity's sake, given that it's pretty rare that we capitalize (for example, "schema checks" is lowercase, and those are a pretty big deal in Studio)

docs/source/advanced/test-builders.mdx
StephenBarlow: Perfect, I'll replace with the registration page. Thanks!

docs/source/advanced/test-builders.mdx
StephenBarlow: Okay cool I'm incorporating this rewrite suggestion for now, and we can expand this clarification in a followup PR. Making sure we hit all the content in a first pass for now

docs/source/advanced/test-builders.mdx
martinbonnin: Test builders are definitely a less prominent feature than schema checks 👍

Issue #3494, opened 2021-11-03
Uniformize operation APIs?


martinbonnin:
Currently we have:

```kotlin
apolloClient.query(query).execute()
apolloClient.mutate(mutation).execute()
apolloClient.subscribe(subscription).execute()
```

I find it weird, especially in the mutation/subscription case we have 2 chained verbs. Should we do this instead?

1.
```kotlin
apolloClient.query(query).execute()
apolloClient.mutation(mutation).execute()
apolloClient.subscription(subscription).execute()
```

or 

2.
```kotlin
apolloClient.queryCall(query).execute()
apolloClient.mutationCall(mutation).execute()
apolloClient.subscriptionCall(subscription).execute()
```

or 

3.
```kotlin
apolloClient.query(query).await()
apolloClient.mutate(mutation).await()
apolloClient.subscribe(subscription).toFlow()
```
Given that 3.0 is coming soon, it's now or never

BoD:
I like this change (favoring the 1st one without "call") because it mirrors the GraphQL syntax.

martinbonnin:
Silly question. Why not keeping the 2.x API?

```kotlin
apolloClient.query(query).await()
apolloClient.mutate(mutation).await()
apolloClient.subscribe(subscription).toFlow()
```

BoD:
Do you mean as a (deprecated) migration helper?

martinbonnin:
> Do you mean as a (deprecated) migration helper?

Not really. What about keeping it for real? Except it doesn't need the imports?

BoD:
So they would be a synonym of `execute`? Or do you mean rename `execute` to use these names instead?

martinbonnin:
> Or do you mean rename execute to use these names instead?

That. 

`apolloClient.mutate(mutation).execute()` feels weird but `apolloClient.mutate(mutation).await()` sounds better to me. The first verb (`mutate`) starts the action, the second (`await`) awaits completion. As a bonus, it reuses the same API as 2.x

BoD:
I see!

I personally prefer `execute` because of its consistency across the 3 operations (less surprising API) but it's pretty much subjective, and I see your point too. One thing: not sure it would be clear that `await` can be called multiple times whereas it makes sense (at least to me!) with `execute`.

martinbonnin:
>  consistency across the 3 operations (less surprising API)

The counter argument can be made that subscriptions are of different kind (Flow) so could use different name. But yea...

[...]

PR #3594, opened 2021-11-23
Renames in ApolloCall and ApolloClient

BoD:
Related to #3494

1. Rename `executeAsFlow` to `toFlow`
2. Rename `mutate` to `mutation`, `subscribe` to `subscription`

Maybe 2. is too much?  Let me know what you think.

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
@@ -1,5 +1,3 @@
-
-
 package com.apollographql.apollo3
 
 import com.apollographql.apollo3.api.ApolloRequest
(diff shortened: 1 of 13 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -24,23 +22,40 @@ abstract class ApolloCall<D: Operation.Data, E: HasMutableExecutionContext<E>>
     return this as E
   }
 
-  fun executeAsFlow(): Flow<ApolloResponse<D>> {
+  /**
+   * Returns a Flow with the results of this [ApolloCall].
```
martinbonnin, line 26:
```suggestion
   * Returns a cold Flow that produces [ApolloResponse]s for this [ApolloCall].
```

martinbonnin (approved):

martinbonnin:
I like 2. avoids 2 verbs one after the other

PR #3596, opened 2021-11-23
Mutate and subscribe deprecation message

BoD:
Actually we can keep `mutate` and `subscribe` a bit longer since it helps migration from 2.x.


martinbonnin (approved):

BoD:
Also add `await` to help migration from 2.x

PR #3602, opened 2021-11-24
Edits to languagey articles, plus reorg, plus ui tests

StephenBarlow:
Just working my way through more articles 😄

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -62,6 +62,11 @@ module.exports = {
             'advanced/ui-tests',
             'advanced/test-builders',
           ],
+          'Language Support': [
+            'advanced/kotlin-native',
+            'advanced/java',
+            'advanced/rxjava',
+          ],
           Advanced: [
             'advanced/query-batching',
             'advanced/upload',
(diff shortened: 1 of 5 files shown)


docs/gatsby-config.js
```diff
@@ -62,6 +62,11 @@ module.exports = {
             'advanced/ui-tests',
             'advanced/test-builders',
           ],
+          'Language Support': [
+            'advanced/kotlin-native',
+            'advanced/java',
+            'advanced/rxjava',
+          ],
```
martinbonnin, line 69: Feels weird-y to me. All these things aren't that related one to the other. 

`kotlin-native` is mainly about Gradle configuration and concurrency for writing Kotlin code in the end
`java` is about Java 
`rxjava` is about a framework that can be called from both Java and Kotlin

I don't have a great alternative proposal either. Maybe "concurrency" for `rxjava` + `kotlin-native`. Or if we wanted to go the different `getting-started` guides route, we could do something like:

* Getting started with Android
* Getting started with KMM
* Getting started with Java

martinbonnin (approved):

BoD (approved):
👍

docs/gatsby-config.js
StephenBarlow: Ok cool reverting for now! Might try something else in a followup

PR #3603, opened 2021-11-24
Use kotlin.Throws 

martinbonnin:
Use kotlin.Throws instead of our own and forward it to the Java codegen. Without this, it was impossible to catch `IOException` from `JsonReader`/`JsonWriter`.


BoD (approved):
👍

PR #3604, opened 2021-11-24
Fix CustomTypeAdapter compat

martinbonnin:
2.x expects `value` to be one of the sealed classes and not a direct instance of `GraphQLFoo`

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -234,12 +234,6 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
       customScalarAdaptersBuilder.add(customScalarType, customScalarAdapter)
     }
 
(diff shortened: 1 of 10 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -234,12 +234,6 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
       customScalarAdaptersBuilder.add(customScalarType, customScalarAdapter)
     }
 
-    @Deprecated("Used for backward compatibility with 2.x", ReplaceWith("addCustomScalarAdapter"))
-    fun <T> addCustomTypeAdapter(
-        customScalarType: CustomScalarType,
-        customScalarAdapter: Adapter<T>,
-    ) = addCustomScalarAdapter(customScalarType, customScalarAdapter)
-
```
martinbonnin, line 242: I removed this because it didn't exist in 2.x

BoD (approved):
👍

PR #3606, opened 2021-11-24
add a doc for Operation Variables

martinbonnin:


docs/source/advanced/operation-variables.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: Defining operation variables
+sidebar_title: Operation variables
+---
+
+GraphQL allows [passing arguments to your operations](https://graphql.org/graphql-js/passing-arguments/) with variables. This allows writing your query once and using different variables values at runtime.
+
+By default, non-nullable variables are mandatory and nullable variables are optional, i.e., it is valid to omit them. Because omitting variables is rarely used in practice, Apollo Android made the choice of making all variables mandatory.
```
BoD, line 8: Maybe explain the reason of the choice too, something like:

```suggestion
By default, non-nullable variables are mandatory and nullable variables are optional, i.e., it is valid to omit them. Because omitting variables is rarely used in practice, Apollo Android made the choice of making all variables mandatory. This allows the generated code to be simpler to call.
```

BoD (approved):
👍

PR #3608, opened 2021-11-25
🧼 🪣  Polishing of the ApolloResponse API

martinbonnin:
Add `ApolloResponse.Builder`

For this, I had to make `D` invariant in `ApolloResponse<D>`. I'm not sure why it was made covariant in the first place and can't find a use case where there would be an inheritance relation between two `Operation.Data` so I think this is ok.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
@@ -8,7 +8,7 @@ import kotlin.jvm.JvmName
 /**
  * Represents a GraphQL response. GraphQL responses can be be partial responses so it is valid to have both data != null and errors
  */
-class ApolloResponse<D : Operation.Data> @Deprecated("Please use ApolloRequest.Builder methods instead. This will be removed in v3.0.0.") constructor(
+class ApolloResponse<D : Operation.Data> @Deprecated("Please use ApolloResponse.Builder methods instead. This will be removed in v3.0.0.") constructor(
     @JvmField
     val requestUuid: Uuid,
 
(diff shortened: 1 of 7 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -8,15 +8,15 @@ import kotlin.jvm.JvmName
 /**
  * Represents a GraphQL response. GraphQL responses can be be partial responses so it is valid to have both data != null and errors
  */
-class ApolloResponse<out D : Operation.Data>(
+class ApolloResponse<D : Operation.Data> @Deprecated("Please use ApolloRequest.Builder methods instead. This will be removed in v3.0.0.") constructor(
```
BoD, line 11:
```suggestion
class ApolloResponse<D : Operation.Data> @Deprecated("Please use ApolloResponse.Builder methods instead. This will be removed in v3.0.0.") constructor(
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -78,24 +78,51 @@ class ApolloResponse<out D : Operation.Data>(
...
-    @Suppress("UNCHECKED_CAST")
-    return ApolloResponse(
-        requestUuid,
-        operation,
-        data as D?,
-        errors,
-        extensions,
-        executionContext
-    )
+  @Deprecated("Please use ApolloRequest.Builder methods instead. This will be removed in v3.0.0.")
```
BoD, line 81:
```suggestion
  @Deprecated("Please use ApolloResponse.Builder methods instead. This will be removed in v3.0.0.")
```

BoD (approved):
👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
martinbonnin: Good catch 🙏

PR #3609, opened 2021-11-25
🧼 🪣  Polishing of the CompiledGraphQL API

martinbonnin:
- use the same Builder pattern as everywhere else
- remove some companion objects


BoD (approved):
🧼

PR #3610, opened 2021-11-25
🧼🪣  Misc apollo-api polishing

martinbonnin:
- remove default parameters for Error
- add FileUpload.create compat
- added compat errors for `Response` and `ScalarTypeAdapters`

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapter.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapter.kt
@@ -2,6 +2,8 @@
 
 package com.apollographql.apollo3.api
 
+import com.apollographql.apollo3.annotations.ApolloInternal
+import com.apollographql.apollo3.api.internal.Version2CustomTypeAdapterToAdapter
 import kotlin.jvm.JvmField
 import kotlin.jvm.JvmStatic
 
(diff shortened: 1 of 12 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Version2CustomTypeAdapter.kt
```diff
@@ -51,3 +53,19 @@ sealed class CustomTypeValue<T>(@JvmField val value: T) {
     }
   }
 }
+
+/**
+ * A replica of Apollo Android v2's ScalarTypeAdapters, to ease migration from v2 to v3.
```
BoD, line 58: Nitpick: not sure 'ease migration' is the right word for this one since it will throw - OTOH I don't have a better wording either :)

apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/FileUpload.kt
```diff
@@ -5,8 +5,7 @@ import okio.buffer
 import okio.source
 import java.io.File
 
-fun Upload.Companion.fromFile(file: File, contentType: String) = object : Upload {
-  override val contentType = contentType
+class FileUpload(private val file: File, override val contentType: String) : Upload {
```
BoD, line 8: However the `fromFile` naming was consistent with the other `fromXyz` methods in the `Upload` interface.

martinbonnin: True. I was thinking of changing them to things like `ByteStringUpload` like for HttpBody https://github.com/apollographql/apollo-android/blob/f5eb27ade6185544bfa2958bfda6671172ca3b66/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/ByteStringHttpBody.kt#L7

I don't think we have a lot of `fromXyz` and TBH I don't like them too much because they:

1 create a asymmetry between the builtin implementation vs others that can be defined in other modules
2 break the typing flow

Having `ByteStringUpload` also breaks the typing flow but at least it makes it easy to see all implementations of a given interface. 

If we wanted things to be even more fluent, we could then add things like `fun String.toUpload(): Upload`. This is what we do in `apollo-ast` but I'm hesitant doing this on super common types like `String`

BoD: Yes, making the same change on the other ones is 👍 

(ps: you probably didn't forget, but just in case, reminder to impact the Upload doc 😊)

BoD (approved):

apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/FileUpload.kt
martinbonnin: Oh yes, I 100% forgot already! Thanks for the reminder!

PR #3613, opened 2021-11-25
Uniformize Upload creation

martinbonnin:
Use `Builder` for upload too:

```kotlin
val upload = DefaultUpload.Builder()
      .content("content_file0")
      // or
      .content(bufferedSource)
      .fileName("file0.txt")
      .contentType("plain/txt")
      .build()
```


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
```diff
@@ -0,0 +1,83 @@
...
+    private var contentType: String? = null
+    private var contentLength: Long = -1
+    private var fileName: String? = null
+
+    fun content(content: BufferedSource): Builder = apply {
+      this.bufferedSource = content
+    }
+
+    fun content(content: String): Builder = apply {
+      this.byteString = content.encodeUtf8()
```
BoD, line 47: To be extra careful, we could disallow setting both `bufferedSource` and `byteString` as this would probably be a mistake (like we do [here](https://github.com/apollographql/apollo-android/blob/f5eb27ade6185544bfa2958bfda6671172ca3b66/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt#L127) for instance)

BoD (approved):
Builders 👍.  I like it 😄

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
martinbonnin: Good idea 👍 . I took a slightly different approach. Let me know what you think: https://github.com/apollographql/apollo-android/pull/3613/commits/94822de45e3e1e44ec502851603332c54a4761cd

BoD: Yes, perfect 👍

BoD (approved):

PR #3614, opened 2021-11-26
Mark Ast, Compiler and MockServer  modules as experimental

martinbonnin:
`apollo-ast`, `apollo-compiler`, `apollo-mockserver` and `apollo-testing-support` are usually not run on a device and generally we have less feedback about the APIs there. Mark them `@ApolloExperimental` so we can still change the APIs if required.

This doesn't mark each and every API but the biggest entry points like `ApolloCompiler`, `toGQLDocument()`, ~`GQLNode`~, etc... so it should be pretty complicated to miss the warning.


BoD (approved):
Well, this is a bit verbose, but I agree it's for the greater good!

PR #3615, opened 2021-11-26
Enable allWarningsAsErrors

martinbonnin:
Enable allWarningsAsErrors for all root modules except `apollo-gradle-plugin`

### DIFF
--- a/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloExperimental.kt
+++ b/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloExperimental.kt
@@ -2,7 +2,7 @@ package com.apollographql.apollo3.annotations
 
 @RequiresOptIn(
     level = RequiresOptIn.Level.WARNING,
-    message = "This API is experimental and can be changed in a backwards-incompatible manner."
+    message = "ApolloGraphQL: This API is experimental and can be changed in a backwards-incompatible manner."
 )
 @Retention(AnnotationRetention.BINARY)
 @Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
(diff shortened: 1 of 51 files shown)


apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloExperimental.kt
```diff
@@ -2,7 +2,7 @@ package com.apollographql.apollo3.annotations
 
 @RequiresOptIn(
     level = RequiresOptIn.Level.WARNING,
-    message = "This API is experimental and can be changed in a backwards-incompatible manner."
+    message = "ApolloGraphQL: This API is experimental and can be changed in a backwards-incompatible manner."
```
martinbonnin, line 5: This makes it more obvious where the warning is coming from when there are a lot of other warnings

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -106,7 +106,7 @@ internal class ExecutableValidationScope(
     name.forEach {
       builder.append(if (!isDecapitalized && it.isLetter()) {
         isDecapitalized = true
-        it.toString().toLowerCase(Locale.US)
+        it.toString().lowercase()
```
martinbonnin, line 109: That's the new Kotlin 1.5 lowercase() that we should be able to use now that we shadow the whole `stdlib` 🤞

apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloExperimental.kt
BoD: Nitpick: I would use `Apollo` instead of `ApolloGraphQL` (same as in the annotation name).

BoD (approved):
So cool! 🎉  It's the best way to avoid more and more warnings slowly creeping in!
Love it ❤️

apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloExperimental.kt
martinbonnin: Yep, most likely. We do have a bunch of `ApolloGraphQL` in the codebase though so I'll do this as a follow up PR

Screenshot 2021-11-26 at 12 42 08

PR #3622, opened 2021-11-29
Simplify APQ configuration

martinbonnin:
The previous `.hashedQueries()` was not only very easy to find but it also did not reset the HTTP method. Instead, replace it with `.enableAutoPersistedQueries(Boolean)` 

Many thanks to @CoreFloDev for spotting this 💜

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -273,14 +273,13 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
      *
      * @param httpMethodForHashedQueries: the [HttpMethod] to use for the initial hashed query that does not send the actual Graphql document.
      * [HttpMethod.Get] allows to use caching when available while [HttpMethod.Post] usually allows bigger document sizes.
-     * Only used if [hashByDefault] is true
      * Default: [HttpMethod.Get]
      *
      * @param httpMethodForDocumentQueries: the [HttpMethod] to use for the follow up query that sends the full document if the initial
      * hashed query was not found.
      * Default: [HttpMethod.Post]
      *
-     * @param hashByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
+     * @param enableByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
      * sendApqExtensions=true and sendDocument=false.
      * If false it will leave them untouched. You can later use [hashedQuery] to enable them
      */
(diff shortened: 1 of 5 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -273,28 +273,25 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
...
      * hashed query was not found.
      * Default: [HttpMethod.Post]
      *
-     * @param hashByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
+     * @param enableByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
      * sendApqExtensions=true and sendDocument=false.
      * If false it will leave them untouched. You can later use [hashedQuery] to enable them
      */
     @JvmOverloads
     fun autoPersistedQueries(
```
BoD, line 287: Could we name this one `enableAutoPersistedQueries` as well? This way the API on the client build and on individual queries would be symmetric (like we did with `canBeBatched`)?  There might be an ambiguity at the call site though (`enableAutoPersistedQueries(HttpMethod, HttpMethod, Boolean)` vs `enableAutoPersistedQueries(Boolean)`)

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -273,28 +273,25 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
...
      * Default: [HttpMethod.Get]
      *
      * @param httpMethodForDocumentQueries: the [HttpMethod] to use for the follow up query that sends the full document if the initial
      * hashed query was not found.
      * Default: [HttpMethod.Post]
      *
-     * @param hashByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
+     * @param enableByDefault: whether to enable Auto Persisted Queries by default. If true, it will set httpMethodForHashedQueries,
      * sendApqExtensions=true and sendDocument=false.
      * If false it will leave them untouched. You can later use [hashedQuery] to enable them
```
BoD, line 284:
```suggestion
     * If false it will leave them untouched. You can later use [enableAutoPersistedQueries] to enable them
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: They're not really symmetrical:

* `ApolloClient.Builder.autoPersistedQueries` is the initial configuration
* `HasMutableExecutionContext.enableAutoPersistedQueries` is the on/off switch

In a way, 

* `ApolloClient.Builder.autoPersistedQueries` <-> `BatchingHttpEngine()`
* `ApolloClient.Builder.enableAutoPersistedQueries` <-> `canBeBatched`

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: 👍

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
BoD: Got it -  this actually makes sense 😊 👍

BoD (approved):
👍

PR #3624, opened 2021-11-29
Move the HTTP cache to a HttpInterceptor

martinbonnin:
This fixes a case where calling `httpCache` would overwrite the `okHttpClient` and it's overall more flexible as it potentially allows using the HttpCacheInterceptor together with other kinds of HttpInterceptors

Thanks @CoreFloDev for catching this.


BoD (approved):
Very cool 👍

PR #3627, opened 2021-11-30
Consolidate fragment articles

StephenBarlow:


BoD (approved):
👍
Good call merging them, it makes sense

martinbonnin (approved):
I like making the distinction because it hints at a clearer terminology. It's either "inline fragment" or "named fragment" but not "fragments". 

But either works really, I don't plan dying on this hill :)

Issue #3625, opened 2021-11-29
Conflicting declarations for "name" property in 3.0.0-beta04


sproctor:
**Summary**
If you have a GraphQL type with a column named "name", Apollo generates an enum class with a property "name" that conflict with the "name" from Kotlin's enum class.

**Version**
3.0.0-beta04, didn't happen in older version

martinbonnin:
Hi ! Thanks for reporting this ! `beta04` changed the default generation mode of enums to use Kotlin enums instead of sealed classes (https://github.com/apollographql/apollo-android/pull/3551).

You can fallback to the previous behaviour by adding this to your Gradle file:

```kotlin
apollo {
  sealedClassesForEnumsMatching.set(listOf(".*"))
}
```

That being said, it'd be interesting to dig more into why you had the issue. We certainly need to escape some enum values. We'll look into it.

sproctor:
`name` and `ordinal` are properties of Kotlin enums and need to be renamed. https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants

That workaround works and is good enough for me, thanks.

BoD:
A fix for this issue has just been merged - it will be available in the next version of the v3 (can also be tested with the latest Snapshot).

sproctor:
This is fixed in 3.0.0-beta05, thanks.

PR #3629, opened 2021-11-30
Escape reserved enum value names ("name" and "ordinal")

BoD:
Fix for #3625.

`name` and `ordinal` are added to the list of reserved words, specifically for enums value names (as to not disrupt the rest of the generated code - `name` is probably a very common identifier).

Potential caveat: value names are also escaped in generated Kotlin sealed classes. and Java classes, which is not really necessary - but a consistent naming between variations may be a good idea?


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -49,7 +49,7 @@ class EnumAsEnumBuilder(
         .addProperty(rawValuePropertySpec)
         .addType(companionTypeSpec())
         .apply {
-          values.forEach { value -> addEnumConstant(layout.enumName(value.name), value.enumConstTypeSpec()) }
+          values.forEach { value -> addEnumConstant(layout.enumValueName(value.name), value.enumConstTypeSpec()) }
```
martinbonnin, line 52: 👍  Good catch

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
```diff
@@ -13,6 +13,10 @@ private val KOTLIN_RESERVED_WORDS = arrayOf(
     "return", "super", "this", "throw", "true", "try", "typealias", "typeof", "val", "var", "when", "while", "yield", "it", "field"
 )
 
+private val RESERVED_ENUM_VALUE_NAMES = arrayOf("name", "ordinal")
+
 fun String.escapeJavaReservedWord() = if (this in JAVA_RESERVED_WORDS) "${this}_" else this
 
-fun String.escapeKotlinReservedWord() = if (this in (JAVA_RESERVED_WORDS + KOTLIN_RESERVED_WORDS)) "${this}_" else this
\ No newline at end of file
+fun String.escapeKotlinReservedWord() = if (this in (JAVA_RESERVED_WORDS + KOTLIN_RESERVED_WORDS)) "${this}_" else this
```
martinbonnin, line 20: Not related to this PR but we should refactor this code to allow different escaping rules for Java vs Kotlin.

apollo-compiler/src/test/graphql/com/example/enums_with_reserved_words/TestOperation.graphql
```diff
@@ -0,0 +1,7 @@
+# The generated enum class should contain all members
+query TestQuery {
+  enum
+  ... on Query {
+    enum
+  }
```
martinbonnin, line 6: Nit: the inline fragment is not needed, is it?

apollo-compiler/src/test/graphql/com/example/enums_with_reserved_words/schema.sdl
```diff
@@ -0,0 +1,8 @@
+type Query {
+  enum: Enum
+}
+
+enum Enum {
+  name
+  ordinal
+}
```
martinbonnin, line 8: What about an integration test instead? In `tests/enums`? That would allow to verify that the generated code compiles but also that the enum value is correctly generated as `name_`

apollo-compiler/src/test/graphql/com/example/enums_with_reserved_words/TestOperation.graphql
BoD: Sorry I meant to explain and then forgot 😅.  This is to trigger the operationBased codegen in `CodegenTest` (it's only triggered when there's fragments).

martinbonnin: Since this only tests enums codegen, I'd say we don't need to generate operationBased code? Saves CI time and a few test fixtures?

In general, I think we reach the point where we can't really test the matrix of all input parameters because they are too many:

* codegenModels
* targetLanguage
* sealedClassesAsEnumMatching
* languageVersion
* etc...

If we were to have 100% code coverage, compilation time would grow exponentially so we're bound to make focused tests for paramters like this

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -59,7 +60,7 @@ class CodegenLayout(
 
   // We used to write upper case enum values but the server can define different values with different cases
   // See https://github.com/apollographql/apollo-android/issues/3035
-  internal fun enumValueName(name: String) = regularIdentifier(name)
+  internal fun enumValueName(name: String) = name.escapeKotlinReservedEnumValueNames()
```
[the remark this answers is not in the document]
BoD: Yeah I mentioned it in the description 😅. I wasn't sure if making it consistent could be desirable - if not I can change it

apollo-compiler/src/test/graphql/com/example/enums_with_reserved_words/TestOperation.graphql
BoD: Oh yeah totally agree - wasn't 100% sure the enum generation was the same for both codegens but since it is, this is unneeded 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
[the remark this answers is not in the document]
BoD: I've updated it in 13fd5f010f021b12a76c0d44b53a5ef1a48d1d16 to escape only in Kotlin enum generation :)

apollo-compiler/src/test/graphql/com/example/enums_with_reserved_words/schema.sdl
BoD: Done in 13fd5f010f021b12a76c0d44b53a5ef1a48d1d16 - makes a smaller PR 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -49,7 +50,9 @@ class EnumAsEnumBuilder(
         .addProperty(rawValuePropertySpec)
         .addType(companionTypeSpec())
         .apply {
-          values.forEach { value -> addEnumConstant(layout.enumName(value.name), value.enumConstTypeSpec()) }
+          values.forEach { value ->
+            addEnumConstant(layout.enumValueName(value.name).escapeKotlinReservedEnumValueNames(), value.enumConstTypeSpec())
```
martinbonnin, line 54: Nit: I'd like the escaping to happen all in `CodegenLayout`. The layout takes GraphQL in and outputs Kotlin. So it certainly needs both `layout.enumValueName(value.name)` and `layout.sealedClassValueName(value.name)`

tests/enums/src/test/kotlin/test/EnumsTest.kt
```diff
@@ -11,6 +12,8 @@ class EnumsTest {
     assertEquals(Direction.NORTH, Direction.safeValueOf("NORTH"))
     assertEquals(Direction.north, Direction.safeValueOf("north"))
     assertEquals(Direction.UNKNOWN__, Direction.safeValueOf("newDirection"))
+    assertEquals(Reserved.name_, Reserved.safeValueOf("name"))
+    assertEquals(Reserved.ordinal_, Reserved.safeValueOf("ordinal"))
```
martinbonnin, line 16: 👍

martinbonnin: Maybe add a test for a sealed class to make sure it's using `name`?

martinbonnin (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
BoD: Will be addressed in a follow-up PR

PR #3630, opened 2021-11-30
Differentiate escaping of Java or Kotlin keywords

BoD:


apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/kotlin/responseBased/case_sensitive_enum/TestQuery.kt.expected
```diff
@@ -37,7 +37,7 @@ public class TestQuery() : Query<TestQuery.Data> {
   public override fun selections(): List<CompiledSelection> = TestQuerySelections.root
 
   public data class Data(
-    public val enum_: Enum?
+    public val `enum`: Enum?
```
BoD, line 40: I must admit I don't understand this one. Is this KotlinPoet automatically escaping it? If yes why is it not done in other places?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
```diff
@@ -8,15 +10,20 @@ private val JAVA_RESERVED_WORDS = arrayOf(
     "transient", "try", "true", "void", "volatile", "while"
 )
 
+// Reference:
+// https://kotlinlang.org/docs/keyword-reference.html#operators-and-special-symbols
+// Note: "yield" is not in this reference, but is in fact reserved (https://github.com/apollographql/apollo-android/issues/1957)
```
martinbonnin, line 15: 👍  Yay for comments !

tests/custom-scalars/src/test/kotlin/test/CustomScalarTest.kt
```diff
@@ -35,8 +35,8 @@ class CustomScalarTest {
         .execute()
         .dataAssertNoErrors
 
-    assertEquals(10_000_000_000, data.long_)
-    assertEquals(1.4f, data.float_)
+    assertEquals(10_000_000_000, data.long)
+    assertEquals(1.4f, data.float)
```
martinbonnin, line 39: Nice 👍

apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/kotlin/responseBased/case_sensitive_enum/TestQuery.kt.expected
martinbonnin: Relevant PR from kotlinpoet: https://github.com/square/kotlinpoet/pull/994

Feels weird escaping in both Apollo + KotlinPoet but I'm not sure we can rely 100% on KotlinPoet because we might use identifiers in places where KotlinPoet doesn't know enough context to escape? 

Things like:

```kotlin
CodeBlock.of("%L.value", layout.fieldName(responseName))
```

Might be worth trying to remove all escaping from Apollo and give KotlinPoet the proper hints?

BoD: Done in bc1aa75283549ead8f067e4bee39b0543beeeaf7 and e3aeb455729db79ed8df904ace4097a25dc5bf0a.

I added an integration test with keywords in different situations, and it looks like it's working. I hope some other cases of `%T` needing to be `%N` aren't missing!

apollo-compiler/src/test/graphql/com/example/reserved_keywords/TestOperation.graphql
```diff
@@ -1,6 +1,6 @@
 query TestQuery {
   yield: hero {
-    it: id
+    while: id
```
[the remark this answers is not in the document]
BoD: Oops I missed this comment - done in the last commit.

tests/escaping/src/test/kotlin/test/EscapingTest.kt
```diff
@@ -0,0 +1,23 @@
...
+        .query(WhileQuery(
+            `return` = "public",
+            `for` = Do(
+                `if` = Optional.Absent,
+                yield_ = Optional.Absent,
+            ),
+        ))
+        .execute()
+    val `while`: WhileQuery.While? = response.dataAssertNoErrors.`while`
+    val `if`: String? = `while`?.`if`
```
martinbonnin, line 21: Ahah this completely messes up with my brain! But looks good 👍

tests/escaping/src/test/kotlin/test/EscapingTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class EscapingTest {
+  // Not actually a test, we're just making sure this compiles
+  suspend fun test() {
+    val response = ApolloClient.Builder().build()
+        .query(WhileQuery(
+            `return` = "public",
+            `for` = Do(
+                `if` = Optional.Absent,
+                yield_ = Optional.Absent,
```
martinbonnin, line 16: Could this be 
```
`yield` = Optional.Absent,
```
 ? instead?

BoD: Ideally yes, but don't know how to do it, since the escaping on this one is not done by KotlinPoet but by us.

martinbonnin: Feels like that deserves another KotlinPoet PR? 'Do not escape things that are already escaped'

That won't solve it in the short term but maybe that can give us a better idea of what's going on?

BoD: Good idea, I'll have a look!

BoD: Turns out, this double escaping handling was already... handled 😅, but in the latest version, so I bumped it. Now we can have `` `yield` `` 🎉  (3323c10bda99713d8812ca39e244777f14179a18)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinCodegenLayout.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class KotlinCodegenLayout(
+    packageNameGenerator: PackageNameGenerator,
+    schemaPackageName: String,
+    useSemanticNaming: Boolean,
+) : CodegenLayout(packageNameGenerator, schemaPackageName, useSemanticNaming) {
+
+  override fun escapeReservedWord(word: String): String = word.escapeKotlinReservedWord()
+
+  internal fun sealedClassValueName(name: String) = name.escapeKotlinReservedEnumValueNames()
```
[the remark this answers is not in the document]
BoD: > The logic is inverted here
You're right 🤦.  In e3c7de8ddada712c6aaafb3b5aafb66a635bd61d I've renamed and added comments to make this absolutely less confusing

About `escapeReservedWord`: we still need to call `escapeKotlinReservedWord` because of `yield` that we escape ourselves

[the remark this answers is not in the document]
martinbonnin: > About escapeReservedWord: we still need to call escapeKotlinReservedWord because of yield that we escape ourselves

Right 👍  So it's more 3 kind of escapes

* underscore escapes
* apollo backtick escapes
* kotlinpoet backtick escapes

pheewww

martinbonnin (approved):

PR #3631, opened 2021-12-01
Edits to normalized cache article

StephenBarlow:


docs/source/caching/normalized-cache.mdx
```diff
@@ -1,53 +1,84 @@
...
+      "name": "Pierre Bordage"
     }
   }
 }
 ```
 
-In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This works because the author is stored only once in the cache and all the fields were retrieved in the initial `BookWithAuthorName` query. If you were to request more fields, like `birthdate` for example, that wouldn't work anymore.
+A normalized cache does _not_ store this response directly. Instead, it breaks it up into the following entries by default:
+
+```yaml:title=Cache
```
BoD, line 55: Nitpick
```suggestion
```json:title=Cache
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -78,7 +111,7 @@ dependencies {
 }
 ```
 
-And add the SQLite cache to your `ApolloClient`. Because different platforms use different drivers, you'll need to call the platform specific constructor:
+Then include the SQLite cache in in your `ApolloClient` initialization **according to your platform target** (different platforms use different drivers):
```
BoD, line 114: Typo
```suggestion
Then include the SQLite cache in your `ApolloClient` initialization **according to your platform target** (different platforms use different drivers):
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -95,64 +128,77 @@ val apolloClient = ApolloClient.Builder()
...
+
+  // Execute the query
+  .execute()
 ```
 
-Like for `normalizedCache(NormalizedCacheFactory)`, `fetchPolicy(FetchPolicy)` is an extension function on `ApolloClient.Builder()` so you will need `apollo-normalized-cache` in your classpath for this to work.
+As with `normalizedCache(NormalizedCacheFactory)`, `fetchPolicy(FetchPolicy)` is an extension function on `ApolloClient.Builder()`, so you need `apollo-normalized-cache` in your classpath for this to work.
 
-Because the normalized cache deduplicates all your data, it allows to react to cache changes. This is done with `watchers` that listen to cache changes. You can learn more about them [in this page](query-watchers). Using watchers requires deduplicating your objects by specifying their object ids.
+Because the normalized cache deduplicates data, it enables you to react to cache changes. You do thiswith `watchers` that listen to cache changes. [Learn more about query watchers.](./query-watchers/)
```
BoD, line 174:
```suggestion
Because the normalized cache deduplicates data, it enables you to react to cache changes. You do this with `watchers` that listen to cache changes. [Learn more about query watchers.](./query-watchers/)
```

BoD: Very nitpick: should we say "observe" instead of listen? Mixing "watch" and "listen" may be odd :) Or not!

docs/source/caching/normalized-cache.mdx
```diff
@@ -95,64 +128,77 @@ val apolloClient = ApolloClient.Builder()
...
-    author {
+    author {     # Author object
       id
       name
     }
   }
 }
+```
+
+```yaml:title=Cache
```
BoD, line 193:
```suggestion
```json:title=Cache
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -162,35 +208,40 @@ query AuthorById($id: String!) {
...
-* "author1"`: `{"id": "author1", name": "Pierre Bordage"}
-* "QUERY_ROOT"`: `{"favoriteBook": "CacheKey(book1)", "author(\"id\": \"author1\")": "CacheKey(author1)"}
-```
+* It takes up more space.
+* Modifying one of these objects does _not_ notify any watchers of the _other_ object.
+
+We want to deduplicate entries like these by making sure they're assigned the _same_ cache ID when they're written, resulting in a cache that looks more like this:
 
-The good news is that all your objects have an `id` field that we can use for that purpose. If each `id` is unique across the whole graph, you can use it directly. If it's unique per object type, you can prefix it with the type name. For an example: `"Book:$id"`.
+```yaml:title=Cache
```
BoD, line 227:
```suggestion
```json:title=Cache
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -162,35 +208,40 @@ query AuthorById($id: String!) {
...
 
-This will store the following records:
+After executing this query, our cache looks like this:
 
-```
-* "favoriteBook": {"id": "book1", "title": "Les guerriers du silence", "author": "CacheKey{favoriteBook.author}"}
-* "favoriteBook.author": {"id": "author1", name": "Pierre Bordage"}
-* "author("id": "author1")": {"id": "author1", "name": "Pierre Bordage"}
-* "QUERY_ROOT": {"favoriteBook": "CacheKey{favoriteBook}", "author(\"id\": \"author1\")": "CacheKey{author(\"id\": \"author1\")}"}
+```yaml{2-3}:title=Cache
```
BoD, line 213:
```suggestion
```json{2-3}:title=Cache
```

BoD (approved):
👍

docs/source/caching/normalized-cache.mdx
```diff
@@ -1,53 +1,84 @@
...
 }
 ```
 
-In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This works because the author is stored only once in the cache and all the fields were retrieved in the initial `BookWithAuthorName` query. If you were to request more fields, like `birthdate` for example, that wouldn't work anymore.
+A normalized cache does _not_ store this response directly. Instead, it breaks it up into the following entries by default:
+
+```yaml:title=Cache
+"favoriteBook": {"id": "bk123", "title": "Les guerriers du silence", "author": "CacheKey{favoriteBook.author}"
+"favoriteBook.author": {"id": "au456", name": "Pierre Bordage"}
+"QUERY_ROOT": {"favoriteBook": "CacheKey(favoriteBook)"}
```
martinbonnin, line 58:
```suggestion
"QUERY_ROOT": {"favoriteBook": "ApolloCacheReference{favoriteBook}"}
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -1,53 +1,84 @@
...
+```
+
+> ⚠️ These default generated cache IDs (`favoriteBook` and `favoriteBook.author`) are undesirable for data deduplication. See [Specifying cache IDs](#specifying-cache-ids).
 
-Using a normalized cache as several advantages. The main one is to avoid storing duplicate data and save some space. Another big one is that you can use the cache as a single source of truth for your UI and react to changes in the cache. To learn more about the process of normalization, check this [blog post about cache normalization](https://www.apollographql.com/blog/apollo-client/caching/demystifying-cache-normalization/).
+* Notice that the `author` field of the `Book` entry now contains the string `CacheKey{favoriteBook.author}`. This is a **reference** to the `Author` cache entry.
```
martinbonnin, line 63:
```suggestion
* Notice that the `author` field of the `Book` entry now contains the string `ApolloCacheReference{favoriteBook.author}`. This is a **reference** to the `Author` cache entry.
```

docs/source/caching/normalized-cache.mdx
```diff
@@ -1,53 +1,84 @@
...
     }
   }
 }
 ```
 
-In the above example, requesting the author of your favorite book with the `AuthorById` query will return a result from the cache if you requested your favorite book before. This works because the author is stored only once in the cache and all the fields were retrieved in the initial `BookWithAuthorName` query. If you were to request more fields, like `birthdate` for example, that wouldn't work anymore.
+A normalized cache does _not_ store this response directly. Instead, it breaks it up into the following entries by default:
+
+```yaml:title=Cache
+"favoriteBook": {"id": "bk123", "title": "Les guerriers du silence", "author": "CacheKey{favoriteBook.author}"
```
martinbonnin, line 56:
```suggestion
"favoriteBook": {"id": "bk123", "title": "Les guerriers du silence", "author": "ApolloCacheReference{favoriteBook.author}"
```

docs/source/caching/normalized-cache.mdx
martinbonnin: I'd say "listen" is more widespread at this point but that's completely subjective. Both work for me.

docs/source/caching/normalized-cache.mdx
```diff
@@ -95,64 +128,77 @@ val apolloClient = ApolloClient.Builder()
...
 ```
 
-Like for `normalizedCache(NormalizedCacheFactory)`, `fetchPolicy(FetchPolicy)` is an extension function on `ApolloClient.Builder()` so you will need `apollo-normalized-cache` in your classpath for this to work.
+As with `normalizedCache(NormalizedCacheFactory)`, `fetchPolicy(FetchPolicy)` is an extension function on `ApolloClient.Builder()`, so you need `apollo-normalized-cache` in your classpath for this to work.
 
-Because the normalized cache deduplicates all your data, it allows to react to cache changes. This is done with `watchers` that listen to cache changes. You can learn more about them [in this page](query-watchers). Using watchers requires deduplicating your objects by specifying their object ids.
+Because the normalized cache deduplicates data, it enables you to react to cache changes. You do thiswith `watchers` that listen to cache changes. [Learn more about query watchers.](./query-watchers/)
 
-## Specifying your object IDs
+## Specifying cache IDs
```
martinbonnin, line 176: Any rationale behind the use of `object id` vs `cache id` vs `cache key`? I tend to prefer `cache key` everywhere but maybe these terms have slightly different meanings?

docs/source/caching/normalized-cache.mdx
StephenBarlow: Haha fair "observation"! Yeah I agree that "listen" is such a widespread word for this that it's fine 😄

docs/source/caching/normalized-cache.mdx
StephenBarlow: Cool, I wasn't sure whether these cache blocks were going to technically be valid _anything_ so I initially went with the most relaxed syntax 😄

docs/source/caching/normalized-cache.mdx
StephenBarlow: I think we should avoid "Object ID" to reduce overlap with "an object's `id`", which might or not be the ID for the cache.

Between "cache ID" and "cache key", I went with "cache ID" purely because we already use that in the React docs and it seemed like a reasonable tie-breaker. I do think that "cache key" is a very reasonable alternative and maybe we should flip it for _every_ client in the future

Issue #3412, opened 2021-10-14
Remove temporary @Deprecated APIs


BoD:
During development of v3, some API changes have happened (for instance With-ers vs Builders), and we temporarily kept previous constructs as deprecated, to ease the transition for early adopters of v3.

Before releasing v3 stable, let's remove these temporary APIs.

To find them in the code, search for `This will be removed in v3.0.0`

PR #3632, opened 2021-12-01
Remove temporarily deprecated methods

BoD:
Fix for #3412


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -242,7 +203,7 @@ class ApolloClient @JvmOverloads @Deprecated("Please use ApolloClient.Builder in
     @Deprecated("Used for backward compatibility with 2.x", ReplaceWith("addCustomScalarAdapter"))
     fun <T> addCustomTypeAdapter(
         customScalarType: CustomScalarType,
-        @Suppress("DEPRECATION") customTypeAdapter: CustomTypeAdapter<T>,
+        @Suppress("DEPRECATION") customTypeAdapter: com.apollographql.apollo3.api.CustomTypeAdapter<T>,
```
BoD, line 206: Using FQN here to avoid a warning on the import and * import 😅

apollo-api/api/apollo-api.api
```diff
@@ -29,7 +29,7 @@ public final class com/apollographql/apollo3/api/Adapters {
 }
 
 public final class com/apollographql/apollo3/api/ApolloRequest : com/apollographql/apollo3/api/HasExecutionContext {
-	public fun <init> (Lcom/apollographql/apollo3/api/Operation;Ljava/util/UUID;Lcom/apollographql/apollo3/api/ExecutionContext;)V
+	public synthetic fun <init> (Lcom/apollographql/apollo3/api/Operation;Ljava/util/UUID;Lcom/apollographql/apollo3/api/ExecutionContext;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
```
martinbonnin, line 32: I can understand the `synthetic` (might be the way kotlin hides private constructors?) but the `DefaultConstructorMarker` is weird here. There's no default parameter at all 🤔

BoD: Looking at the rest of the file, this alternative `<init>` with `synthetic` + `DefaultConstructorMarker` parameter  is there "often but not all the time" 🤔. I wonder what the rule (and use) are.

martinbonnin: Turns out the `DefaultConstructor` marker is not (always) related to default parameters. It's just a marker (hence the name) to avoid any name clash with other possibly defined constructor.

Still weirds that it ends up in the public API. I opened an issue there: https://github.com/Kotlin/binary-compatibility-validator/issues/73

martinbonnin (approved):
Nice 👍

apollo-api/api/apollo-api.api
BoD: Indeed the constructor is `private` so it would be expected to not show up

PR #3633, opened 2021-12-01
Edits to cache ID articles

StephenBarlow:
Putting this up with just one of the two articles done in case I don't finish the programmatic side today

Edit: finished the other one 😄

### DIFF
--- a/docs/source/caching/declarative-ids.mdx
+++ b/docs/source/caching/declarative-ids.mdx
@@ -1,44 +1,38 @@
 ---
-title: Declarative cache ids
+title: Declarative cache IDs in Apollo Android
+sidebar_title: Declarative cache IDs
 ---
 
-> Using declarative cache ids is the recommended way to define your cache ids. For a more flexible (and more complex) API, you can also define your cache ids [programmatically](programmatic-ids).
+When using a [normalized cache](./normalized-cache/) in Apollo Android, it's recommended that you specify a **cache ID** for each object type in your schema. If you don't, objects are assigned a _default_ cache ID, but that ID can lead to undesirable duplication of data.
 
-Declaratively specifying your cache ids allows the codegen to add the id fields automatically and generates type-safe code that can return a valid cache id for any object or field.
+When specifying cache IDs, it's recommended that you do so **declaratively**, using the methods described in this article:
 
-It works by extending the backend schema. The GraphQL spec allows [object extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) using the `extend type` keywords. In our case, this can be used together with the `@typePolicy` and `@fieldPolicy` directives and an extra '.graphqls' file that we will name `extra.graphqls`.
-
-Cache ids are computed from two sources:
-
-1. From the object fields (`author.id`)
-2. From a field's arguments (`author(id)`)
-
-
-1. is happening **after** a network request and is important to merge the network response with existing data. This is the most usual case.
-2. is happening **before** a network request and is important to avoid a network round trip if all the data is in the cache already. This is an optimization that can avoid some cache misses.
-
-With declarative cache ids:
-
-1. uses `@typePolicy`
-2. uses `@fieldPolicy`
-
-## @typePolicy
-
-`@typePolicy` allows to specify the id from object fields returned **after** a network request.
-
-It's a client-only directive defined by:
-
-```graphql:title=apollo.graphqls
-# Marks fields as key fields. Key fields are used to compute the cache key
-# of an object `keyFields` should contain a selection set. Composite fields
-# are not supported yet.
-directive @typePolicy(keyFields: String!) on OBJECT
+```graphql
+extend type Book @typePolicy(keyFields: "id")
 ```
 
-Assuming the following schema type in your `schema.graphqls` file:
+> For advanced use cases, you can also define cache IDs [programmatically](programmatic-ids).
+
+## How they work
+
+With declarative cache IDs, the codegen process adds ID fields automatically and generates type-safe code that can return a valid cache ID for any object.
+
+To do this, Apollo Android extends your back-end schema. The GraphQL spec supports [object extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) using the `extend type` keywords. We can use this together with the `@typePolicy` and `@fieldPolicy` directives, along with an extra `.graphqls` file named `extra.graphqls`.
+
+You can obtain a given object type's cache ID from one of two sources:
+
+| Source | Directive | Description |
+|--------|-----------|-------------|
+| From a response object's fields (e.g., `Book.id`) | [`@typePolicy`](#typepolicy) | This happens _after_ a network request and is essential to merging a query result with existing cached data. This is the most common case. |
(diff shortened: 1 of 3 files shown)


docs/source/caching/normalized-cache.mdx
```diff
@@ -232,16 +232,9 @@ We want to deduplicate entries like these by making sure they're assigned the _s
 
 Fortunately, all of our objects have an `id` field that we can use for this purpose. If an `id` is unique across _all_ objects in your graph, you can use its value directly as a cache ID. Otherwise if it's unique _per object type_, you can prefix it with the type name (as shown above).
 
-Note that `id` can be obtained from two sources:
```
StephenBarlow, line 235: Removing this here because it looks like it'll have corresponding information in both of the individual ID articles

docs/source/caching/declarative-ids.mdx
```diff
@@ -1,100 +1,95 @@
...
 
 ```graphql:title=extra.graphqls
 extend type Query @fieldPolicy(forField: "author", keyArgs: "firstName lastName")
 ```
 
-The codegen will recognize the id argument and generate the correct cache id every time you use it in a query.
\ No newline at end of file
+In this case, the cache ID for an `Author` object includes the values of both its `firstName` and `lastName` fields, which are both provided as arguments to the `Query.author` field.
+
+If multiple fields of an object type have key arguments, you can apply multiple `@fieldPolicy` directives to that type.
```
StephenBarlow, line 95: Is this true? I just guessed 😄

docs/source/caching/programmatic-ids.mdx
```diff
@@ -59,25 +52,27 @@ val objectIdGenerator = object : CacheKeyGenerator {
 }
 ```
 
-Note how this requires `id` to be present in the response. If a query does not return this field, the `CacheKey` will be inconsistent and duplication will happen.
+You can also use the current object's typename to use _different_ cache ID generation logic for different object types.
```
StephenBarlow, line 55: Is this accurate? I assume this function is getting called for every object type, and you might need to `switch` on the typename or similar to use different logic per type

StephenBarlow: Pardon my kotlin ignorance. `when` 😄

docs/source/caching/declarative-ids.mdx
```diff
@@ -1,100 +1,95 @@
 ---
-title: Declarative cache ids
+title: Declarative cache IDs in Apollo Android
```
StephenBarlow, line 2: Question(s): Let's say I set up a `fieldPolicy` to get an object's ID, but that object isn't in the cache. So now I hit the network and the object is returned. When assigning the cache ID for that object, is the `fieldPolicy` value used, or is the `typePolicy` value used (assuming both are defined)? What if there's a `fieldPolicy` for a field but _not_ a `typePolicy` for the type returned by that field?

martinbonnin: `@typePolicy` is the one used after the network. `@fieldPolicy` doesn't have access to the object because it is used before the network.

Usually, you would setup a `@typePolicy` always and optionally a `fieldPolicy`. I don't think there's a valid use case where `@fieldPolicy` is defined but not `@typePolicy`

docs/source/caching/declarative-ids.mdx
martinbonnin: It is!

docs/source/caching/programmatic-ids.mdx
```diff
@@ -1,52 +1,45 @@
...
-2. uses `CacheKeyResolver`
-
-
-## CacheKeyGenerator
-
-To customize how cache ids are built from object fields, create a new instance of `CacheKeyGenerator`:
+The `CacheKeyGenerator` class enables you to generate custom cache IDs from an object's field values. This basic example generates _every_ object type's cache ID from its `id` field:
 
 ```kotlin
 val objectIdGenerator = object : CacheKeyGenerator {
```
martinbonnin, line 22:
```suggestion
val cacheKeyGenerator = object : CacheKeyGenerator {
```

docs/source/caching/programmatic-ids.mdx
```diff
@@ -1,52 +1,45 @@
...
-    .serverUrl("https://...")
-    .normalizedCache(
-        normalizedCacheFactory = cacheFactory,
-        objectIdGenerator = objectIdGenerator,
-    )
-    .build()
+  .serverUrl("https://...")
+  .normalizedCache(
+    normalizedCacheFactory = cacheFactory,
+    objectIdGenerator = objectIdGenerator,
```
martinbonnin, line 37:
```suggestion
    cacheKeyGenerator = cacheKeyGenerator,
```

docs/source/caching/programmatic-ids.mdx
```diff
@@ -1,52 +1,45 @@
...
+  )
+  .build()
 ```
 
-If you need the object typename, you can access it through `context`:
+You can get the current object's typename from the `context` object and include it in the generated ID, like so:
 
 ```kotlin
 val objectIdGenerator = object : CacheKeyGenerator {
```
martinbonnin, line 45:
```suggestion
val cacheKeyGenerator = object : CacheKeyGenerator {
```

docs/source/caching/programmatic-ids.mdx
martinbonnin: Yes, that's correct 👍 It's called for every object.

docs/source/caching/programmatic-ids.mdx
```diff
@@ -59,25 +52,27 @@ val objectIdGenerator = object : CacheKeyGenerator {
...
+## `CacheKeyResolver`
 
-To customize how cache ids are built from fields arguments (before going to the network), create a new instance of `CacheKeyResolver`:
+The `CacheKeyResolver` class enables you to generate custom cache IDs from an field's arguments. This basic example generates _every_ object type's cache ID from the `id` argument _if it's present_:
 
 ```kotlin
 val cacheKeyResolver = object: CacheKeyResolver() {
   override fun cacheKeyForField(field: CompiledField, variables: Executable.Variables): CacheKey? {
-    // [field] contains compile-time information about what type of object is being resolved
+    // `field` contains compile-time information about what type of object is being resolved.
```
martinbonnin, line 66: FWIW, the `[identifier]` syntax is the [KDoc](https://kotlinlang.org/docs/kotlin-doc.html) way to do. Backticks work too !

docs/source/caching/programmatic-ids.mdx
```diff
@@ -87,24 +82,27 @@ val cacheKeyResolver = object: CacheKeyResolver() {
...
-    .normalizedCache(
-        normalizedCacheFactory = cacheFactory,
-        objectIdGenerator = objectIdGenerator,
-        cacheResolver = cacheKeyResolver
-    )
-    .build()
+  .serverUrl("https://...")
+  .normalizedCache(
+    normalizedCacheFactory = cacheFactory,
+    objectIdGenerator = objectIdGenerator,
```
martinbonnin, line 92: We renamed that to `cacheKeyGenerator`:

```suggestion
    cacheKeyGenerator = cacheKeyGenerator,
```

martinbonnin (approved):

docs/source/caching/programmatic-ids.mdx
StephenBarlow: Ok cool! I'll go with backticks for now just for cross-docset consistency within comments

PR #3636, opened 2021-12-03
Fix collecting fields on new schema types

martinbonnin:
When querying fields on an interface:

```graphql
{
  animal {
    ... on Animal {
      species
    }
  }
}
```

The cache code relied on `fragment.possibleTypes` to determine whether a field was possible or not. This works well if the app is in sync with the schema but if the schema introduces a new type, the `possibleTypes` approach will miss all the field on the type in scope. 

That's because the client code doesn't know anything about the new type except it implements the interface. This PR takes this last condition into account.


apollo-compiler/src/test/graphql/com/example/antlr_tokens/java/operationBased/antlr_tokens/TestQuery.java.expected
```diff
@@ -111,8 +110,13 @@ public class TestQuery implements Query<TestQuery.Data> {
   }
 
   @Override
-  public List<CompiledSelection> selections() {
-    return TestQuerySelections.root;
+  public CompiledField rootField() {
+    return new CompiledField.Builder(
+      "data",
+      com.example.antlr_tokens.type.Query.type
```
BoD, line 116: Nitpick: would it be possible to have it imported to avoid the FQN?

martinbonnin: This is kotlinpoet land. 
Maybe possible, maybe not. Overall, KotlinPoet focuses at making correct code. Making the look nice is not the primary goal. We could use something like ktlint on top of KotlinPoet if we really wanted to.

BoD: Fair enough!

BoD (approved):
👏

Issue #3634, opened 2021-12-02
Subscriptions throw exception "java.lang.IllegalStateException: WeSocket queue full"


sproctor:
**Summary**
Re-subscribing after app goes to sleep throws "java.lang.IllegalStateException: WeSocket queue full".

**Version**
3.0.0-beta04

**Description**
I'm not entirely sure the cause. It happens regularly when my app wakes up causing a crash. I didn't experience it with 2.5.x

Changes:

beta04
```kotlin
val client = ApolloClient.Builder()
            .okHttpClient(okHttpClient)
            .httpServerUrl(context.getString(R.string.graphql_endpoint))
            .webSocketServerUrl(context.getString(R.string.graphql_ws_endpoint))
            .addCustomScalarAdapter(Uuid.type, uuidAdapter)
            .addCustomScalarAdapter(Timestamptz.type, KotlinxInstantAdapter)
            .addCustomScalarAdapter(Numeric.type, BigDecimalAdapter)
            .build()
```

2.5.x
```kotlin
 val client = ApolloClient.builder()
            .serverUrl(context.getString(R.string.graphql_endpoint))
            .okHttpClient(okHttpClient)
            .subscriptionTransportFactory(
                WebSocketSubscriptionTransport.Factory(
                    context.getString(
                        R.string.graphql_ws_endpoint
                    ), okHttpClient
                )
            )
            .addCustomTypeAdapter(CustomType.UUID, uuidCustomTypeAdapter)
            .addCustomTypeAdapter(CustomType.TIMESTAMPTZ, dateCustomTypeAdapter)
            .addCustomTypeAdapter(CustomType.NUMERIC, numericCustomTypeAdapter)
            .build()
```

sproctor:
Related issue from SO https://stackoverflow.com/questions/70008541/handle-network-error-when-using-graphql-subscription-with-apollo-on-android-with

martinbonnin:
Thanks for sending this! I can reproduce. Will work on a fix.

In the meantime. You can copy/paste the [DefaultWebSocketEngine code](https://github.com/apollographql/apollo-android/blob/584d5a5ae94e6deb13568552bfa8d19f389ed588/apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt) and remove the checks at line 86 and 93. And then pass it instead of the builtin `DefaultWebSocketEngine`. That should hide the issue.

I'll update this issue when a fix lands

[...]

PR #3646, opened 2021-12-06
Fix restarting a websocket after a network error

martinbonnin:
This PR serializes all messages from the `WsProtocol` in the "supervisor". 

What I call "supervisor" is a long-lived coroutine that:
- receives `Commands` from the various subscription `Flow`s, creates a `WsProtocol` when needed and forwards the commands
- receives `Events` from the `WsProtocol` and forwards them back to the various subscription `Flow`s

The `WsProtocol` owns the `WebSocketConnection`. The "supervisor" owns the `WsProtocol` and closes it after `idleTimeout`. 

The problem this PR adresses is that there was no way previously to know if a `WsProtocol` ended as an error and therefore was never re-created. With these changes, `WsProtocol.run` is required to send a terminal `NetworkError` event that will be received by the "supervisor" that can then reset its state.

Fixes https://github.com/apollographql/apollo-android/issues/3634

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
@@ -10,7 +10,18 @@ import com.apollographql.apollo3.exception.ApolloNetworkException
 import com.apollographql.apollo3.internal.BackgroundDispatcher
 import com.apollographql.apollo3.internal.transformWhile
 import com.apollographql.apollo3.network.NetworkTransport
+import com.apollographql.apollo3.network.ws.internal.Command
+import com.apollographql.apollo3.network.ws.internal.Event
+import com.apollographql.apollo3.network.ws.internal.GeneralError
+import com.apollographql.apollo3.network.ws.internal.Message
+import com.apollographql.apollo3.network.ws.internal.NetworkError
+import com.apollographql.apollo3.network.ws.internal.OperationComplete
+import com.apollographql.apollo3.network.ws.internal.OperationError
+import com.apollographql.apollo3.network.ws.internal.OperationResponse
+import com.apollographql.apollo3.network.ws.internal.StartOperation
+import com.apollographql.apollo3.network.ws.internal.StopOperation
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.BufferOverflow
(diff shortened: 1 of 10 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -77,74 +74,117 @@ private constructor(
...
+            try {
+              protocol.connectionInit()
+            } catch (e: Exception) {
+              // Error initializing the connection
+              protocol = null
+              mutableEvents.emit(NetworkError(e))
+              continue
+            }
+
+            connectionJob = scope.launch(start = CoroutineStart.UNDISPATCHED) {
```
BoD, line 163: Question: why do de need `UNDISPATCHED` here?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/internal/WsMessage.kt
```diff
@@ -0,0 +1,32 @@
+package com.apollographql.apollo3.network.ws.internal
+
+import com.apollographql.apollo3.api.ApolloRequest
+import com.apollographql.apollo3.api.Operation
+import kotlinx.coroutines.channels.Channel
+
+internal class ConnectionHolder(val events: Channel<Event>) {
```
BoD, line 7: Is this unused?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
martinbonnin: I wanted to make sure `protocol.run()` would always be called. I'm not 100% sure if `protocol` could be reset before starting to execute the coroutine so added this as an extra precaution. Maybe it's not required but it shouldn't hurt so better safe than sorry...

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/internal/WsMessage.kt
martinbonnin: Good catch ! 🙏

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
BoD: 👍 Makes sense, thanks (and TIL 😄)!

martinbonnin: Side note: Looks like the type inference is not working here. `protocol` should be nullable always. I filed https://youtrack.jetbrains.com/issue/KT-50092

BoD (approved):

PR #3647, opened 2021-12-06
Expose more configuration options on ApolloClient.Builder

martinbonnin:
Expose all the `DefaultHttpNetworkTransport` and `DefaultWebSocketNetworkTransport` options on `ApolloClient.Builder`. For the vast majority of use cases, this is easier to use than having to call a `NetworkTransport` constructor:

```kotlin
                .webSocketIdleTimeout(...)
                .wsProtocol(...)
                .webSocketServerUrl(...)
                .httpServerUrl(...)
                .addHttpInterceptor(...)
                // ...
                .build()
```

Passing a custom `NetworkTransport` is still possible but will throw if it conflicts with any of the top level properties

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -23,11 +23,9 @@ import com.apollographql.apollo3.interceptor.NetworkInterceptor
 import com.apollographql.apollo3.internal.defaultDispatcher
 import com.apollographql.apollo3.mpp.assertMainThreadOnNative
 import com.apollographql.apollo3.network.NetworkTransport
-import com.apollographql.apollo3.network.http.DefaultHttpEngine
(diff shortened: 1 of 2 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -131,44 +132,80 @@ private constructor(
...
     /**
-     * The [HttpEngine] to use for HTTP requests
+     * The timeout after which an inactive WebSocket will be closed
      *
-     * For more customization, see also [networkTransport]
+     * See also [subscriptionNetworkTransport] for more customization
      */
-    fun httpEngine(httpEngine: HttpEngine) = apply {
-      this.httpEngine = httpEngine
+    fun webSocketIdleTimeout(webSocketIdleTimeout: Long) = apply {
```
BoD, line 192: We may want to name it `webSocketIdleTimeoutMillis` to avoid any ambiguity

BoD (approved):

Issue #3550, opened 2021-11-15
Default value for `generateOptionalOperationVariables`


martinbonnin:
Decide a default value for `generateOptionalOperationVariables`. Currently, it is `false`, meaning that a query like this:

```graphql
# nullable variable
query GetHero($id: String) {
  hero(id: $id) {
    name
  }
}
```

will generate a class like this:

```kotlin
// Notice the absence of Optional<String?>
class GetHeroQuery(id: String?) {
  // ...
}
```

While this means easier and leaner callsites in the vast majority of cases, it doesn't work any more if omitting a variable is required:

```graphql
# bi-directional pagination
query GetHeroes(before: String, after: String)
```

Optional variables can be opted-in with `@optional`:

```graphql
# bi-directional pagination
query GetHeroes(before: String @optional, after: String @optional)
```

But that's not always obvious. Also the query becomes Apollo Android specific as a server will not recognize `@optional`

martinbonnin:
We decided to keep it like this because the vast majority of use cases will require a value. If omitting a value is required, it is possible through either the `@optional` directive (inside the query) or `generateOptionalOperationVariables.set(true)` in Gradle configuration. We will add a doc for that.

BoD:
# Decision log

After some discussion (including with members of the `apollo-ios` team), we have decided to go back to make the default to generate Optionals, while keeping the ability to opt-out/in globally, or on a per-variable basis.

It is reminded that Optionals on operation variables may be needed in some cases, even though we estimate that it should be rare (or that needing them may reflect some kind of a bad practice on the schema).

Pros of generating Optionals **by default** are:
- it is "technically correct" per the GraphQL spec
- it will not be surprising/confusing to users who are used to GraphQL
- when it is needed, a user using the default configuration will **not** need to search the documentation (or use Google/StackOverflow/etc.)
- it is consistent with the `apollo-ios` behavior

Cons:
- this makes the generated code more verbose to use
- in most cases, the Optionals are actually not needed

In the end, a very good argument to keep Optional by default is this "matrix" by @martinbonnin (rephrased here):

- No Optional by default
    - User didn't want Optional => User is happy 🙂 
    - User needed Optional => User is unhappy 😡  (they may have wasted time looking for how to enable it)
- Optional by default
    - User didn't want Optional =>  User is neutral 😐  (they may have to understand the spec and why we did this, which is because it's allowed by the spec)
    - User wanted Optional => User is happy 🙂

PR #3648, opened 2021-12-06
Change optional default value

BoD:
Fix for #3550

- Change the default value of `generateOptionalOperationVariables` to true (Operation variables are wrapped in `Optional` by default, but this can be opted out with the `@optional` directive)
- Remove the `variables` directive which was actually unused

- Documentation update upcoming in another PR


apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -1,7 +1,6 @@
-# Marks a field or variable as optional
-# When used on a variable definition, the corresponding variable will be optional. By default Apollo Android generates all variables as
-# non-optional because if the variable was added in the first place, it's usually to pass a value but setting this directive allows
-# to mark them optional again
+# Marks a field or variable definition as optional or required
+# By default Apollo Android generates all variables of nullable types as optional, in compliance with the GraphQL specification,
+# but this can be configured with this directive, because if the variable was added in the first place, it's usually to pass a value
 # See also @variables
```
martinbonnin, line 4:
```suggestion
```

apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -13,16 +12,9 @@ directive @optional(if: Boolean = true) on FIELD | VARIABLE_DEFINITION
 # and null in the other
 directive @nonnull(fields: String! = "") on OBJECT | FIELD
 
-# Whether to make all variables optional at the schema level
-# This is false by default. This can overriden with @optional
-directive @variables(optional: Boolean = false) on SCHEMA
```
martinbonnin, line 18: Long term, I'd like to move more of the Gradle configuration to `extra.graphqls` because:

1. it'll simplify the Gradle plugin
2. it's more granular (allows to target individual types, fields, etc...)

No need to keep stuff that's not working yet so 👍

tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
```diff
@@ -295,7 +296,7 @@ class WatcherTest {
     val channel = Channel<EpisodeHeroNameQuery.Data?>(capacity = Channel.UNLIMITED)
     val job = launch {
 
-      apolloClient.query(EpisodeHeroNameQuery(Episode.EMPIRE))
+      apolloClient.query(EpisodeHeroNameQuery(Optional.Present(Episode.EMPIRE)))
```
martinbonnin, line 299: Use `generateOptionalVariables.set(false)` in `integration-tests` to keep the tests leaner and unchanged?

apollo-ast/src/main/resources/apollo.graphqls
BoD: Makes sense!

tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
BoD: I was hesitating about that and thought it would be interesting to see the impact of the change in our tests 😅 But now that we saw it, I can revert yes.

martinbonnin (approved):

apollo-ast/src/main/resources/apollo.graphqls
martinbonnin: Another argument is:

3. makes it easier to use other build systems than Gradle (maven, buck, bazel, etc..)

PR #3650, opened 2021-12-06
Add ExecutionOptions

martinbonnin:
This PR removes `HasExecutionContext` and `HasMutableExecutionContext` in favour of `ExecutionOptions` and `MutableExecutionOptions`.

This is more verbose but also more type-safe for options that are in `apollo-runtime`. `apollo-normalized-cache` and `apollo-http-cache` still use the `ExecutionContext` because they are independant modules.

I'm also hoping that it's faster because it doesn't require traversing a linked-list every time a request parameter is required.

And it's also easier to use in Java because no extension functions.

Now for the main drawback, it requires duplicating the options in 3 different places: `ApolloRequest`, `ApolloCall` and `ApolloClient`. This process is error prone as failing to pass a parameter down the line will be unnoticed until runtime. 

Generally I tend to think that this is more work for us and less work for users so it'd be in favour of merging it. But I'm not 100% sold on it so all feedbacks welcome.


BoD (approved):
I find it generally easier to reason about (even though yes, it's more verbose). LGTM.

Issue #2119, opened 2020-03-29
Not able to get errors from body using CoroutinesExtensions toDeferred


FilipObornik:
**Summary**
When using `toDeffered()` extension method or other methods from CoroutinesExtensions, it is not possible to get errors from response body because the response body is already closed. Maybe similar approach as Retrofit's HttpException could be used for ApolloHttpExeption. Than, from Retrofit's HttpException, you can retrieve error body like this:
`(exception as HttpException).response().errorBody().string()`

I hot fixed it for myself by modifying extension function `toDeferred()` by overriding onHttpError() in ApolloCall.Callback with `deferred.completeExceptionally(e)`. But I am aware that this is not secure, because there is possibility of forgetting to close the response body. My `toDeffered()` looks like this

```
fun <T> ApolloCall<T>.toDeferred(): Deferred<Response<T>> {
    val deferred = CompletableDeferred<Response<T>>()

    deferred.invokeOnCompletion {
        if (deferred.isCancelled) {
            cancel()
        }
    }
    enqueue(object : ApolloCall.Callback<T>() {
        override fun onResponse(response: Response<T>) {
            if (deferred.isActive) {
                deferred.complete(response)
            }
        }

        override fun onHttpError(e: ApolloHttpException) {
            if (deferred.isActive) {
                deferred.completeExceptionally(e)
            }
        }

        override fun onFailure(e: ApolloException) {
            if (deferred.isActive) {
                deferred.completeExceptionally(e)
            }
        }
    })

    return deferred
}
```

**Version**
1.4.1

martinbonnin:
I'm going to close this one because it's super old. Also because `3.0.0` introduced `exposeErrorBody` so it's not possible to get the error body.

Note that `Retrofit` decided to [fully buffer the error](https://github.com/square/retrofit/blob/ee71a6858dfddcc4650f6ff4d71112911bdf6ca7/retrofit/src/main/java/retrofit2/OkHttpCall.java#L229). Given that error responses are rarely huge, that avoid the `exposeErrorBody` altogether and any risk of leaking connections. 

Something we might want to do at some point?

PR #3661, opened 2021-12-08
Add exposeErrorBody

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/2119

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpEngine.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpEngine.kt
@@ -55,11 +55,18 @@ import kotlin.jvm.JvmStatic
  *
  * @param batchIntervalMillis the interval between two batches
  * @param maxBatchSize always send the batch when this threshold is reached
+ * @param exposeErrorBody configures whether to expose the error body in [ApolloHttpException].
+ *
+ * If you're setting this to `true`, you **must** catch [ApolloHttpException] and close the body explicitly
+ * to avoid sockets and other resources leaking.
+ *
+ * Default: false
  */
 class BatchingHttpEngine @JvmOverloads constructor(
     val delegate: HttpEngine = DefaultHttpEngine(),
     val batchIntervalMillis: Long = 10,
     private val maxBatchSize: Int = 10,
+    private val exposeErrorBody: Boolean = false
 ) : HttpEngine {
   private val dispatcher = BackgroundDispatcher()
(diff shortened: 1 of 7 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -124,6 +136,18 @@ private constructor(
       this.httpRequestComposer = DefaultHttpRequestComposer(serverUrl)
     }
 
+    /**
+     * configures whether to expose the error body in [ApolloHttpException].
```
BoD, line 140:
```suggestion
     * Configures whether to expose the error body in [ApolloHttpException].
```

BoD (approved):

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpEngine.kt
```diff
@@ -55,11 +55,18 @@ import kotlin.jvm.JvmStatic
  *
  * @param batchIntervalMillis the interval between two batches
  * @param maxBatchSize always send the batch when this threshold is reached
+ * @param exposeErrorBody configures whether to expose the error body in [ApolloHttpException].
+ *
+ * If you're setting this to `true`, you **must** catch [ApolloHttpException] and close the body explicitly
+ * to avoid sockets and other resources leaking.
+ *
+ * Default: false
```
BoD, line 63: Nit: I think this needs to be on one line for KDoc to understand we're still on the `@param`
```suggestion
 * @param exposeErrorBody configures whether to expose the error body in [ApolloHttpException]. If you're setting this to `true`, you **must** catch [ApolloHttpException] and close the body explicitly to avoid sockets and other resources leaking. Default: false
```

BoD (approved):

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpEngine.kt
martinbonnin: This is what I get with the current KDoc:

Screenshot 2021-12-09 at 14 38 25

Looks okay-ish to me?

BoD: Thank you for checking, it works differently that I thought it would - TIL :) You can ignore my suggestion then!

Issue #2793, opened 2020-12-09
gradle downloadApolloSchema has different behavior if options are specified in config


sproctor:
**Summary**
`./gradlew downloadApolloSchema` does not match the behavior of specifying options on the command line

**Version**
2.4.6

**Description**
`./gradlew downloadApolloSchema` gets a schema that doesn't contain any of my actual schema and fails if the schema file doesn't exist.
`./gradlew downloadApolloSchema --endpoint=https://domain/v1/graphql --schema=path/to/schema.json --header="x-hasura-admin-secret: secret"` works exactly as expected.

Gradle config:
```groovy
apollo {
    generateKotlinModels = true
    service("hasura") {
        schemaPath = "path/to/schema.json"
        introspection {
            sourceSetName = "main"
            headers = [
                   "x-hasura-admin-secret": GRAPHQL_SECRET
            ]
            endpointUrl = "https://domain/v1/graphql"
        }
    }
}
```

sproctor:
Thanks, I figured I was doing something wrong. I did manage to figure it out after switching to the 3.x branch. Either the error messages were clearer, I was more clever, or just spent more time on it.

PR #3663, opened 2021-12-08
Add a sample how to use HttpRequestComposer

martinbonnin:
See https://github.com/apollographql/apollo-android/issues/2793


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
```diff
@@ -167,7 +167,7 @@ class OperationBuilder(
         if (generateQueryDocument) {
           addStatement("return $OPERATION_DOCUMENT")
         } else {
-          addStatement("error(\"The query document was removed from this operation. Use generateQueryDocument = true if you need it\"")
+          addStatement("error(\"The query document was removed from this operation. Use generateQueryDocument.set(true) if you need it\"")
```
BoD, line 170:
```suggestion
          addStatement("error(\"The query document was removed from this operation. Use generateQueryDocument.set(true) if you need it\")")
```

BoD (approved):
👍

Issue #3664, opened 2021-12-08
[3.0.0-beta04] crash when using the http cache


CoreFloDev:
Hi,

**Summary**

```
Fatal Exception: java.lang.IllegalStateException: HTTP cache: no snapshot
       at com.apollographql.apollo3.cache.http.DiskLruHttpCache.read(DiskLruHttpCache.java:28)
       at com.apollographql.apollo3.cache.http.CachingHttpEngine.networkMightThrow(CachingHttpEngine.java:86)
       at com.apollographql.apollo3.cache.http.CachingHttpEngine.execute(CachingHttpEngine.java:39)
       at com.apollographql.apollo3.network.http.HttpNetworkTransport$EngineInterceptor.intercept(HttpNetworkTransport.java:139)
       at com.apollographql.apollo3.network.http.DefaultHttpInterceptorChain.proceed(DefaultHttpInterceptorChain.java:21)
       at com.apollographql.apollo3.network.http.HttpNetworkTransport$execute$1.invokeSuspend(HttpNetworkTransport.java:100)
       at com.apollographql.apollo3.network.http.HttpNetworkTransport$execute$1.invoke(HttpNetworkTransport.java:2)
       at kotlinx.coroutines.flow.SafeFlow.collectSafely(SafeFlow.java:61)
       at kotlinx.coroutines.flow.AbstractFlow.collect(AbstractFlow.java:212)
       at kotlinx.coroutines.flow.internal.ChannelFlowOperatorImpl.flowCollect(ChannelFlowOperatorImpl.java:195)
       at kotlinx.coroutines.flow.internal.ChannelFlowOperator.collectTo$suspendImpl(ChannelFlowOperator.java:157)
       at kotlinx.coroutines.flow.internal.ChannelFlowOperator.collectTo(ChannelFlowOperator.java:1)
       at kotlinx.coroutines.flow.internal.ChannelFlow$collectToFun$1.invokeSuspend(ChannelFlow.java:60)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(BaseContinuationImpl.java:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.java:106)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.java:571)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.java:750)
[...]

PR #3667, opened 2021-12-08
Fix DiskLruHttpCache concurrency

BoD:
Fix for #3664.

When `networkMightThrow` is called by 2 threads in parallel we could easily arrive at a situation where `DiskLruHttpCache.write` would return `null` immediately (`cache.edit` being in the "another edit is in progress" case), which in turn would make the `DiskLruHttpCache.read` right after it throw an error.

Moving the whole write under the write lock fixes the issue.

### DIFF
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
@@ -43,35 +43,40 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
         .build()
   }
 
+  /**
+   * Store the [response] with the given [cacheKey] into the cache.
+   * Note: the response's body is not consumed nor closed.
+   */
   fun write(response: HttpResponse, cacheKey: String) {
-    val editor = cacheLock.write {
+    val editor = cacheLock.read {
       cache.edit(cacheKey)
-    } ?: return
+    }
 
-    cacheLock.write {
-      try {
-        editor.newSink(ENTRY_HEADERS).buffer().use {
-          val map = mapOf(
-              "statusCode" to response.statusCode.toString(),
-              "headers" to response.headers.map { httpHeader ->
-                // Moshi doesn't serialize Pairs by default (https://github.com/square/moshi/issues/508) so
-                // we use a Map with a single entry
-                mapOf(httpHeader.name to httpHeader.value)
-              },
-          )
-          adapter.toJson(it, map)
-        }
-        editor.newSink(ENTRY_BODY).buffer().use {
-          val responseBody = response.body
-          if (responseBody != null) {
-            it.writeAll(responseBody)
-            responseBody.close()
-          }
-        }
-        editor.commit()
-      } catch (e: Exception) {
-        editor.abort()
+    if (editor == null) {
+      return
+    }
+
+    try {
+      editor.newSink(ENTRY_HEADERS).buffer().use {
+        val map = mapOf(
+            "statusCode" to response.statusCode.toString(),
+            "headers" to response.headers.map { httpHeader ->
+              // Moshi doesn't serialize Pairs by default (https://github.com/square/moshi/issues/508) so
+              // we use a Map with a single entry
+              mapOf(httpHeader.name to httpHeader.value)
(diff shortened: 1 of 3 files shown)


apollo-http-cache/src/test/kotlin/com/apollographql/apollo3/cache/http/internal/CachingHttpInterceptorTest.kt
```diff
@@ -122,6 +122,31 @@ class CachingHttpInterceptorTest {
       }
     }
   }
+
+  @Test
+  fun cacheInParallel() {
```
martinbonnin, line 127: Nice test 👍
Out of curiosity, would it be possible to write it using `launch {}` and `Dispatchers.IO` instead of `Executors`? This way, the test would be KMP-ready?

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -44,36 +44,34 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
   }
 
   fun write(response: HttpResponse, cacheKey: String) {
-    val editor = cacheLock.read {
+    val editor = cacheLock.write {
       cache.edit(cacheKey)
```
martinbonnin, line 48: My understanding was that the cache could edit several (different) entries in parallel so putting all that under a write lock might increase contention?

More generally, this raises the question of what to do if two threads try to write the same cache entry concurrently. We can either decide that:

1. first one wins, second one is discarded
or
2. serialize everything. both get written to the disk

While 2. seems safer, I'm wondering if 1. is not better in terms of performance. Especially at this level of concurrency, there's no real notion of "first" and "second". Which one is going to come first is scheduling dependant so might as well be random. Writing the first one doesn't seem all that bad compared to writing both. 

All in all, I'm wondering if the bug isn't at https://github.com/apollographql/apollo-android/blob/ac5ed15dd4c0795daa0ffb089eb483845422d5fc/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt#L82

Maybe it's just unsafe to assume a read after a write always succeeds and we should instead clone the network response body?

apollo-http-cache/src/test/kotlin/com/apollographql/apollo3/cache/http/internal/CachingHttpInterceptorTest.kt
BoD: My first try but without `Dispatchers.IO` which didn't repro the issue - not sure why (isn't the default dispatcher also multithreaded 🤔), but with `Dispatchers.IO` it works! 👍

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
BoD: Good analysis! I did 1. in 3b4f9ae3ce3c4baa588772efaa0478be3a0fa9a1 and it works! 🎉

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -43,6 +43,10 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
         .build()
   }
 
+  /**
+   * Store the [response] with the given [cacheKey] into the cache.
+   * Note: the response's body is not consumed nor closed.
```
martinbonnin, line 48: Wait, how come the response body is not consumed and closed? Since `write` needs to write it to the cache?

BoD: Because I used `.peek()` and removed the `close()` :)

martinbonnin: Ah yes, sorry I missed that 👍

martinbonnin (approved):

Issue #3660, opened 2021-12-08
Move `BatchingHttpEngine` to a `HttpInterceptor`


martinbonnin:
Just like `CachingHttpInterceptor`, interceptors are genrally more versatile

PR #3670, opened 2021-12-09
Move BatchingHttpEngine to a HttpInterceptor

BoD:
Related to #3660

- Make `BatchingHttpEngine` an `HttpIntereptor`
  - I needed a `dispose` so I added it to the interface, with a default implementation in order to not break the API
  - `HttpNetworkTransport` has the responsibility to call `dispose` on its interceptors
- Also added a `batching` method on the client builder
- To keep compatibility with the RC I created a deprecated `BatchingHttpEngine` that delegates to the interceptor. That's a bit unfortunate, maybe we don't really need it? LMK what you think! (and I'm sorry because of this the diff is not nice to read)

Docs will need to be updated.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
@@ -73,6 +73,10 @@ private constructor(
       this.enableAutoPersistedQueries = enableAutoPersistedQueries
     }
 
+    override fun canBeBatched(canBeBatched: Boolean?): Builder<D> = apply {
+      if (canBeBatched != null) addHttpHeader(ExecutionOptions.CAN_BE_BATCHED, canBeBatched.toString())
+    }
+
     fun requestUuid(requestUuid: Uuid) = apply {
       this.requestUuid = requestUuid
     }
@@ -95,7 +99,7 @@ private constructor(
           httpHeaders = httpHeaders,
           sendApqExtensions = sendApqExtensions,
           sendDocument = sendDocument,
-          enableAutoPersistedQueries = enableAutoPersistedQueries
+          enableAutoPersistedQueries = enableAutoPersistedQueries,
       )
     }
   }
--- a/apollo-runtime/api/apollo-runtime.api
+++ b/apollo-runtime/api/apollo-runtime.api
@@ -2,6 +2,7 @@ public abstract class com/apollographql/apollo3/ApolloCall : com/apollographql/a
 	public fun <init> (Lcom/apollographql/apollo3/ApolloClient;Lcom/apollographql/apollo3/api/Operation;)V
 	public fun addExecutionContext (Lcom/apollographql/apollo3/api/ExecutionContext;)Ljava/lang/Object;
 	public fun addHttpHeader (Ljava/lang/String;Ljava/lang/String;)Ljava/lang/Object;
+	public fun canBeBatched (Ljava/lang/Boolean;)Ljava/lang/Object;
 	public fun enableAutoPersistedQueries (Ljava/lang/Boolean;)Ljava/lang/Object;
 	public final fun getApolloClient ()Lcom/apollographql/apollo3/ApolloClient;
 	public fun getEnableAutoPersistedQueries ()Ljava/lang/Boolean;
@@ -63,6 +64,8 @@ public final class com/apollographql/apollo3/ApolloClient$Builder : com/apollogr
 	public final fun autoPersistedQueries (Lcom/apollographql/apollo3/api/http/HttpMethod;Lcom/apollographql/apollo3/api/http/HttpMethod;Z)Lcom/apollographql/apollo3/ApolloClient$Builder;
 	public static synthetic fun autoPersistedQueries$default (Lcom/apollographql/apollo3/ApolloClient$Builder;Lcom/apollographql/apollo3/api/http/HttpMethod;Lcom/apollographql/apollo3/api/http/HttpMethod;ZILjava/lang/Object;)Lcom/apollographql/apollo3/ApolloClient$Builder;
 	public final fun build ()Lcom/apollographql/apollo3/ApolloClient;
+	public fun canBeBatched (Ljava/lang/Boolean;)Lcom/apollographql/apollo3/ApolloClient$Builder;
+	public synthetic fun canBeBatched (Ljava/lang/Boolean;)Ljava/lang/Object;
 	public final fun customScalarAdapters (Lcom/apollographql/apollo3/api/CustomScalarAdapters;)Lcom/apollographql/apollo3/ApolloClient$Builder;
 	public fun enableAutoPersistedQueries (Ljava/lang/Boolean;)Lcom/apollographql/apollo3/ApolloClient$Builder;
 	public synthetic fun enableAutoPersistedQueries (Ljava/lang/Boolean;)Ljava/lang/Object;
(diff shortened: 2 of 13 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -361,6 +363,28 @@ private constructor(
...
+     * Some servers might have a per-HTTP-call cache making it faster to resolve 1 big array
+     * of n queries compared to resolving the n queries separately.
+     *
+     * @param batchIntervalMillis the interval between two batches
+     * @param maxBatchSize always send the batch when this threshold is reached
+     *
+     * See also [BatchingHttpInterceptor]
+     */
+    @JvmOverloads
+    fun batching(
```
martinbonnin, line 379: Maybe `httpBatching` ?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -0,0 +1,234 @@
...
+ * Some servers might have a per-HTTP-call cache making it faster to resolve 1 big array
+ * of n queries compared to resolving the n queries separately.
+ *
+ * Because [com.apollographql.apollo3.ApolloQueryCall.execute] suspends, it only makes sense to use query batching when queries are
+ * executed from different coroutines. Use [async] to create a new coroutine if needed
+ *
+ * [BatchingHttpInterceptor] buffers the whole response, so it might additionally introduce some
+ * client-side latency as it cannot amortize parsing/building the models during network I/O.
+ *
+ * [BatchingHttpInterceptor] only works with Post requests. Trying to batch a Get requests is undefined.
```
martinbonnin, line 51:
```suggestion
 * [BatchingHttpInterceptor] only works with Post requests. Trying to batch a Get request is undefined.
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @param batchIntervalMillis the interval between two batches
```
martinbonnin, line 53:
```suggestion
 * @param batchIntervalMillis the maximum time interval between two batches before a batch is sent
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -0,0 +1,234 @@
...
+ * @param maxBatchSize always send the batch when this threshold is reached
```
martinbonnin, line 54:
```suggestion
 * @param maxBatchSize the maximum number of requests queued before a new batch is sent
```

apollo-runtime/api/apollo-runtime.api
```diff
@@ -273,9 +290,14 @@ public final class com/apollographql/apollo3/network/http/HttpInfo$Key : com/apo
 }
 
 public abstract interface class com/apollographql/apollo3/network/http/HttpInterceptor {
+	public abstract fun dispose ()V
 	public abstract fun intercept (Lcom/apollographql/apollo3/api/http/HttpRequest;Lcom/apollographql/apollo3/network/http/HttpInterceptorChain;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
+public final class com/apollographql/apollo3/network/http/HttpInterceptor$DefaultImpls {
+	public static fun dispose (Lcom/apollographql/apollo3/network/http/HttpInterceptor;)V
```
martinbonnin, line 298: I'm not a huge fan of `DefaultImpls` but yea, having to implement `dispose()` all the time isn't great either so 👍 . Apollo 4 will be able to use JVM default methods 😅

apollo-runtime/api/apollo-runtime.api
```diff
@@ -182,42 +187,53 @@ public final class com/apollographql/apollo3/network/OkHttpExtensionsKt {
...
-public final class com/apollographql/apollo3/network/http/BatchingHttpEngine$PendingRequest {
+public final class com/apollographql/apollo3/network/http/BatchingHttpInterceptor$PendingRequest {
 	public fun <init> (Lcom/apollographql/apollo3/api/http/HttpRequest;)V
 	public final fun getDeferred ()Lkotlinx/coroutines/CompletableDeferred;
 	public final fun getRequest ()Lcom/apollographql/apollo3/api/http/HttpRequest;
 }
 
-public final class com/apollographql/apollo3/network/http/BatchingHttpEngineExtensions {
+public final class com/apollographql/apollo3/network/http/BatchingHttpInterceptorKt {
 	public static final fun canBeBatched (Lcom/apollographql/apollo3/api/MutableExecutionOptions;Z)Ljava/lang/Object;
```
martinbonnin, line 231: Looks like this one can move to `MutableExecutionOptions` instead of being an extension function. Want to look into it?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -0,0 +1,234 @@
...
+        body = body,
+    ).build()
+
+    freeze(request)
+
+    var exception: ApolloException? = null
+    val result = try {
+      val response = interceptorChain!!.proceed(request)
+      if (response.statusCode !in 200..299) {
+        throw ApolloHttpException(response.statusCode, response.headers, "HTTP error ${response.statusCode} while executing batched query: '${response.body?.readUtf8()}'")
```
martinbonnin, line 165: Heads up this will conflict with the `exposeErrorBody` PR

apollo-runtime/api/apollo-runtime.api
BoD: Yes :) 👀

martinbonnin (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
```diff
@@ -73,6 +73,10 @@ private constructor(
       this.enableAutoPersistedQueries = enableAutoPersistedQueries
     }
 
+    override fun canBeBatched(canBeBatched: Boolean?): Builder<D> = apply {
+      if (canBeBatched != null) addHttpHeader(ExecutionOptions.CAN_BE_BATCHED, canBeBatched.toString())
+    }
```
martinbonnin, line 78: Sorry I'm late to the party 🙈:

Should we also add

```kotlin
override val canBeBatched: Boolean
```

I think that would:
1. make it possible to hide CAN_BE_BATCHED as an implementation detail completely
2. keep the symmetry between mutable and immutable `ExecutionOptions`. i.e. if we can mutate it, we should also be able to get it?

BoD: For 2., good point. But for 1. we still need `CAN_BE_BATCHED` from the Interceptor unfortunately (being an HttpInterceptor it has access to the headers, but not to the `ApolloRequest`)

martinbonnin: Gotcha 👍 I still think it makes sense to add it but no strong feelings there. Your call.

Issue #3666, opened 2021-12-08
Update connectionPayload for subscription on running Apollo instance (for Bearer-Token).


schoeda:
We use in our Android client wss with the Apollo-GraphQL SDK. To subscribe to GraphQL we add a `connectionPayload`-Map of the `SubscriptionWsProtocol` with a Bearer-Token. If the Bearer get invalid, we have to recreate the whole Apollo client and add all subscriptions again.

Solution:
Make it possible to update the `connectionPayload` at runtime, without a recreation of the Apollo client.

Thank you!

martinbonnin:
Hi again 👋 

How does your server signals an expired `Bearer-Token`? We might add something like this but AFAIK, there's no real standard way to do it like 401 errors with HTTP so having a sample of what the exchange would look like would help a ton!

schoeda:
Hey again, Martin :) 

I tested this only with an invalid Token at app start and we get an error in the subscription flow.

```
apolloSyncClient.subscription(ChangedFacilityPickJobsSubscription()).toFlow().collect { subscriptionResult ->
    if (subscriptionResult.hasErrors()) {
        // Errorhandling
    } else {
        // Do funny things...
    )
}
```

The error looks like this:

```
Error(message = Unauthorized, locations = null, path=null, extensions = {code=UNAUTHENTICATED, response={statusCode=401, message=Unauthorized}}, nonStandardFields = null)
```

I know, this is a little bit different than an expired token. But if that happens, we create a complete new instance of the Apollo client.

martinbonnin:
Tentative API there: https://github.com/apollographql/apollo-android/pull/3674

Because `connectionPayload` is a lambda, you should be able to provide a new value when needed. And use `reconnectWhen {}` to decide whether to reconnect or not. Let me know how that looks!

matejcikm:
Hi @martinbonnin, is possible to implement `reconnect()` method in `WebSocketNetworkTransport`? I think it will be helpful for handle bearer tokens expiration errors explained above. `reconnectWhen` lambda is not ideal for this situation, because it is called only after `NetworkError` (disconnecting from webSocket, etc...).

martinbonnin:
Hi @matejcikm 👋 

[...]

PR #3674, opened 2021-12-09
Add `WebSocketNetworkTransport.reconnectWhen {}`

martinbonnin:
It can be used like so:

```kotlin
        .httpServerUrl("http://localhost:8080/graphql")
        .webSocketServerUrl("http://localhost:8080/subscriptions")
        .wsProtocol(
            SubscriptionWsProtocol.Factory(
                connectionPayload = {
                  mapOf("token" to upToDateToken)
                }
            )
        )
        .webSocketReconnectWhen {
          // this is called when an error happens on the WebSocket
          it is ApolloWebSocketClosedException && it.code == 1001
        }
        .build()
```

If `reconnectWhen()` returns `true` the error will be consumed and the `WebSocketNetworkTransport` will try to reconnect and re-subscribe to all open subscriptions.

If `reconnectWhen()` returns `false`the error will be terminate all active subscription `Flow`s. They can individually decide to `retry()` or abort.

See https://github.com/apollographql/apollo-android/issues/3666


tests/websockets/src/commonTest/kotlin/AppSyncTest.kt
```diff
@@ -8,15 +7,42 @@ import kotlinx.coroutines.flow.collect
...
+ *
+ * The AWS project requires authentication at the moment. You can get an apiKey from the console. It would be nice to change this
+ */
 @Ignore
 class AppSyncTest {
 
   @Test
   fun simple() = runTest {
-    val apiKey = "changeMe"
+    val apiKey = "da2-mdu5pjfruvbe3kf4sukxr76bim" //TODO("changeMe")
```
BoD, line 45: Uh oh

BoD (approved):
👍

tests/websockets/src/commonTest/kotlin/AppSyncTest.kt
martinbonnin: 🙈 Thanks for catching this! apiKey disabled 🗑️

BoD (approved):

PR #3675, opened 2021-12-09
Update migration file with http cache information

StylianosGakis:
Adds a text very similar to the explanation for the normalized cache section.

To give some context. I was having trouble performing the migration because I was missing the `httpCache` and `httpFetchPolicy` methods. I hadn't found this dependency mentioned anywhere on the migration logs, nor in the readme on the tag 3.0.0-rc01 so I thought I'd contribute this. If I understood something wrong or you feel like it could be worded better, please feel free to act as you think is best. Just want to make this migration easier for other devs too.

### DIFF
--- a/docs/source/migration/3.0.mdx
+++ b/docs/source/migration/3.0.mdx
@@ -432,6 +432,15 @@ val response = apolloClient.query(request)
 
 #### HTTP cache
 
+To add http cache support, add the dependency to your gradle file:
+
+```kotlin
+dependencies {
+  // Add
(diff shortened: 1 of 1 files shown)


StylianosGakis:
> @StylianosGakis: Thank you for submitting a pull request! Before we can merge it, you'll need to sign the Apollo Contributor License Agreement here: https://contribute.apollographql.com/

Done ✅

martinbonnin:
Thank you so much 💙 ! This makes a ton of sense 👍

PR #3676, opened 2021-12-10
Merge ApolloQueryCall, ApolloSubscriptionCall, ApolloMutationCall

martinbonnin:
Last minute breaking change to remove `ApolloQueryCall`, `ApolloSubscriptionCall` and `ApolloMutationCall` in favor of just `ApolloCall<D>`.

**Rationale**:

From the caller point of view, having a second `E` type parameter was confusing as it wasn't super clear what it was for. Using `ApolloCall<D>` reduces the cognitive load around this.

The only thing that `ApolloQueryCall`, `ApolloSubscriptionCall` and `ApolloMutationCall` allowed previously was preventing using `.execute()` for subscriptions but we find that in practice, it shouldn't be a big issue as it's clear from the function signature that calling `.execute()` on a subscription will return a single value. There might even be valid use cases to do so?

Having a single `ApolloCall<D>` is symmetric with `ApolloRequest<D>` and `ApolloResponse<D>` so the API is overall more consistent.

It's still possible to restrict some calls like `fetchPolicy()` for an example using extension functions:

```kotlin
fun <D: Query.Data>ApolloCall<D>.fetchPolicy(fetchPolicy: FetchPolicy) {
  // 
}
```

Out of curiosity, I opened [this question](https://discuss.kotlinlang.org/t/restrict-calling-some-functions-of-a-class-with-generic-type-parameters/23643) to investigate whether we can do so without extension functions but I don't have many hopes there. 

All in all, code is simpler to maintain, cognitive load is smaller for users and we don't lose much type safety, might even gain some flexibility so I think it's a win. Feedbacks welcome

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
@@ -3,6 +3,7 @@ package com.apollographql.apollo3
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.ExecutionContext
+import com.apollographql.apollo3.api.ExecutionOptions
 import com.apollographql.apollo3.api.MutableExecutionOptions
 import com.apollographql.apollo3.api.Operation
 import com.apollographql.apollo3.api.http.HttpHeader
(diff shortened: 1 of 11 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -101,79 +90,13 @@ abstract class ApolloCall<D : Operation.Data, E>(val apolloClient: ApolloClient,
...
-   * [ApolloMutationCall] can be executed several times
-   *
-   * Example:
-   * ```
-   * val response = apolloClient.mutation(SetHeroName("Luke"))
-   *                  .addHttpHeader("Authorization", myToken)
-   *                  .optimisticData(data)
-   *                  .execute()
-   * ```
+   * A shorthand for `toFlow().execute()`.
```
BoD, line 95:
```suggestion
   * A shorthand for `toFlow().single()`.
```

BoD (approved):
👍

PR #3677, opened 2021-12-10
Add canBeBatched var to ExecutionOptions

BoD:
It is not used currently but makes sense in terms of consistency

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
@@ -20,6 +20,7 @@ private constructor(
     override val sendApqExtensions: Boolean?,
     override val sendDocument: Boolean?,
     override val enableAutoPersistedQueries: Boolean?,
+    override val canBeBatched: Boolean?,
 ) : ExecutionOptions {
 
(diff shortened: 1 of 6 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -70,7 +70,10 @@ abstract class ApolloCall<D : Operation.Data, E>(val apolloClient: ApolloClient,
     return this as E
   }
 
+  override var canBeBatched: Boolean? = null
```
BoD, line 73: Sorry it will conflict with #3676

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
```diff
@@ -31,6 +32,7 @@ private constructor(
         .sendApqExtensions(sendApqExtensions)
         .sendDocument(sendDocument)
         .enableAutoPersistedQueries(enableAutoPersistedQueries)
+        .canBeBatched(canBeBatched)
```
martinbonnin, line 35: 👍  Good catch

martinbonnin (approved):

PR #3678, opened 2021-12-10
Multi-module doc: disambiguate root vs graphql-root module

martinbonnin:
Add some explanations about the GraphQL root module in the context of multi-module as Gradle also has its own "root" project.

Let me know what you think and I'll forward port to dev-3.x if/once it looks ok.

### DIFF
--- a/docs/source/advanced/multi-modules.mdx
+++ b/docs/source/advanced/multi-modules.mdx
@@ -9,18 +9,18 @@ Support for multi-modules build is currently experimental and we'd love to have 
 
 ## Setup
 
-Multi-modules requires that one module and only one contains a schema. This is the schema that all other modules can reuse. In this document, we'll name it "graphql-root". This is **not to be confused with Gradle root project** that contains all your modules.
+Multi-modules requires that one module and only one contains a schema. This is the schema that all other modules can reuse. In this document, we'll refer to it as the "schema module". 
 
-Configure your "graphql-root" module to generate Apollo metadata:
+Configure your schema module to generate Apollo metadata:
 
 ```kotlin
-// graphql-root/build.gradle.kts
+// schema/build.gradle.kts
 apollo {
     generateApolloMetadata.set(true)
 }
 ```
 
-And declare your graphql-root module as a dependency of your feature module:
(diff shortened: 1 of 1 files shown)


docs/source/advanced/multi-modules.mdx
```diff
@@ -9,16 +9,18 @@ Support for multi-modules build is currently experimental and we'd love to have
 
 ## Setup
 
-Configure your parent module to generate Apollo metadata:
+Multi-modules requires that one module and only one contains a schema. This is the schema that all other modules can reuse. In this document, we'll name it "graphql-root". This is **not to be confused with Gradle root project** that contains all your modules.
```
BoD, line 12:
```suggestion
Multi-modules requires that one and only one module contains a schema. This is the schema that all other modules can reuse. In this document, we'll name it "graphql-root". This is **not to be confused with Gradle root project** that contains all your modules.
```

BoD (commented):
I'm not 100% sure "root" is the best term if other modules must depend on it (sometimes in a dependency tree the "root" is the app module, the one who depends on the other ones).

Could we name it the "schema module"?

martinbonnin:
> I'm not 100% sure "root" is the best term if other modules must depend on it (sometimes in a dependency tree the "root" is the app module, the one who depends on the other ones).
> 
> Could we name it the "schema module"?

Agreed. All the things are renamed

docs/source/advanced/multi-modules.mdx
```diff
@@ -27,37 +29,37 @@ dependencies {
...
+// This module can use fragments and types from 'schema'
 dependencies {
-    apolloMetadata(project(":root"))
+    apolloMetadata(project(":schema"))
     generateApolloMetadata.set(true)
 }
 
-// root/build.gradle.kts
-// This module is the root module
+// schemabuild.gradle.kts
```
BoD, line 61:
```suggestion
// schema/build.gradle.kts
```

BoD (approved):
👍

joeldenke (approved):
Looks good, less confusing of root gradle :)

PR #3679, opened 2021-12-10
Add back some 2.x symbols

martinbonnin:
Add a few symbols to help with the migration. Most of them are quite advanced symbols so I went straight to `DeprecationLevel.ERROR` but at least that will indicate what the replacement is.

This also reshuffles a bit the RxJava artifacts:

* I removed a lot of `ApolloStore`/`Interceptor` APIs as I feel they're not really battle tested and I rather not ship them compared to ship something half backed. We can always add them later if needed/when we have proper use cases
* I added `Rx2Apollo` back as a Kotlin-accessible class for Kotlin users not using function extensions


tests/rxjava/src/test/kotlin/RxJavaTest.kt
```diff
@@ -63,18 +57,5 @@ class RxJavaTest {
           it is ApolloException
         }
   }
-
-  @Test
-  fun writingToTheStoreWorks() {
-    rx2ApolloStore.rxWriteOperation(GetRandomQuery(), GetRandomQuery.Data(random = 43)).blockingGet()
```
martinbonnin, line 69: This is the part that I feel is better not shipping until we get a better idea of the problem we're trying to solve.

Note how the rest of the tests didn't change

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Error.kt
```diff
@@ -34,6 +34,17 @@ class Error(
...
+  /**
+   * Custom attributes associated with this error
+   */
+  @Deprecated(
+      message = "Used for backward compatibility with 2.x",
+      replaceWith = ReplaceWith("nonStandardFields"),
+      level = DeprecationLevel.ERROR
+  )
+  val customAttributes: Map<String, Any?>
+    get() = error("Use nonStandardFields instead")
```
martinbonnin, line 46: Thanks @michgauz for catching this 💙 !

BoD (approved):
👍

PR #3680, opened 2021-12-10
Add ApolloIdlingResource migration guide

martinbonnin:
Also add `ApolloIdlingResource.create` with `DeprecationLevel.Error` to point to the migration guide.

Thanks again @michgauz for catching this 🙏


BoD (approved):
2️⃣ 🚀

PR #3685, opened 2021-12-10
Fix `replapcing` typo

mateuszkwiecinski:
I started going through migration guide and noticed a typo that doesn't seem to be fixed in any of opened PRs. Please let my know if there is another way of reporting typos or if you simply don't want to see such nitpicky changes at this moment 😬


martinbonnin:
PR are perfect 👍 . Thanks for sending this 🙏

PR #3692, opened 2021-12-13
Add operation.composeJsonRequest and matching test case

martinbonnin:
While reviewing https://www.apollographql.com/docs/android/v3/advanced/no-runtime/, I found that we had a missing method there. 

Documentation update to follow


BoD (approved):
👍

PR #3700, opened 2021-12-13
Update 3.x gatsby config for release

StephenBarlow:


martinbonnin (approved):
👍 
Question: Will that take effect when we merge dev-3.x in main? Or is it going to change something when we merge that PR already?

StephenBarlow:
To the best of my understanding, this will have no effect until it ends up on `main`. I asked to double check and have yet to know for _sure_, but if it's time to do any merging tomorrow before I'm around, I bet you can go ahead and go for it and everything will be fine.

PR #3715, opened 2021-12-14
move json helpers out of internal

martinbonnin:
We're referencing them in the doc and hopefully they are well-defined enough so we can live with them in the long run


BoD (approved):
👍

PR #3717, opened 2021-12-15
Mention the apollo-idling-resource artifact in the migration docs

StylianosGakis:
The new dependency required for idlingResource isn't mentioned in the migration docs.
As always, feel free to edit/improve the wording or remove the link that I added to the function if you don't want to maintain the link (currently points at the rc03 version).

### DIFF
--- a/docs/source/migration/3.0.mdx
+++ b/docs/source/migration/3.0.mdx
@@ -623,6 +623,15 @@ val apolloClient = ApolloClient.Builder()
     .build()
 ```
 
+The idlingResource function is hosted under the [new artifact](https://github.com/apollographql/apollo-kotlin/blob/d881829a652d5745a593197fc5384db5f08f2760/apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt#L60-L63) `apollo-idling-resource`.
+Add the new dependency like so:
+
+```kotlin
+dependencies {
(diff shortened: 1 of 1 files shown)


docs/source/migration/3.0.mdx
```diff
@@ -623,6 +623,15 @@ val apolloClient = ApolloClient.Builder()
     .build()
 ```
 
+The [idlingResource function](https://github.com/apollographql/apollo-kotlin/blob/v3.0.0-rc03/apollo-idling-resource/src/main/java/com/apollographql/apollo3/android/ApolloIdlingResource.kt#L60-L63) is hosted under the new artifact `apollo-idling-resource`.
+Add the new dependency like so:
+
+```kotlin
+dependencies {
+  testImplementation("com.apollographql.apollo3:apollo-idling-resource:$version")
```
martinbonnin, line 631: I think this used to be at `com.apollographql.apollo:apollo-idling-resource` so it didn't change much? Or were you using something else?

StylianosGakis: You are 100% correct. While doing the migration I had accidentally deleted the old implementation so I thought that this one was a similar case like how `apollo-http-cache` was added. Sorry!

StylianosGakis:
Closed as it was unnecessary. This existed in 2.0 too, so there is no need for a special mention in the 2->3 migration doc

PR #3719, opened 2021-12-15
Add a Data.toJson that uses reflection to lookup the adapter

martinbonnin:
A helper function to get a Json from a `Data` instance. This is a band-aid until we get better and multiplatform mockserver APIs

### DIFF
--- a/apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/-Data.kt
+++ b/apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/-Data.kt
@@ -1,0 +1,27 @@
+package com.apollographql.apollo3.api
+
+import com.apollographql.apollo3.api.json.JsonWriter
+
+private fun Operation.Data.adapter(): Adapter<Operation.Data> {
+  val name = this::class.java.name
+
+  val operationQualifiedName = name.removeSuffix("${'$'}Data")
+  val operationName = operationQualifiedName.substringAfterLast(".")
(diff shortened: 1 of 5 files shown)


apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/-Data.kt
```diff
@@ -0,0 +1,27 @@
...
+
+  val field = clazz.getDeclaredField("INSTANCE")
+
+  @Suppress("UNCHECKED_CAST")
+  val adapter = field.get(null) as Adapter<Operation.Data>
+
+  return adapter.obj()
+}
+
+fun Operation.Data.toJson(jsonWriter: JsonWriter, customScalarAdapters: CustomScalarAdapters = CustomScalarAdapters.Empty) {
```
BoD, line 24: Should we mark it `@ApolloExperimental`?

martinbonnin: I was wondering the same. I decided not to because it's 2.x functionality for which we don't have a multiplatform equivalent yet. So it's not really new and it's not really deprecated either. 

Once we have a multiplatform mockserver API, we'll mark this one as deprecated?

BoD: Sounds good 👍

BoD (approved):

PR #3720, opened 2021-12-15
Disable key fields check if unnecessary and optimize its perf

BoD:
This branch investigates a performance issue in the compiler around check key fields, that has been reported.
- Do not check key fields if there are no `typePolicy` directives
- Cache `implementedTypes` and `keyFields`
- Cache `collectFields` entirely

👆 when profiling, these changes seem to show the check runs faster, however we still don't have a convincing repro steps where it was taking a significant time overall, so caution must be taken

### DIFF
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
@@ -94,7 +94,7 @@ class Schema(
   /**
    * Returns whether the `typePolicy` directive is present on at least one object in the schema
    */
-  fun hasAnyTypePolicyDirectives(): Boolean {
+  fun hasTypeWithTypePolicy(): Boolean {
     return typeDefinitions.values.filterIsInstance<GQLObjectTypeDefinition>().any { objectType ->
       objectType.directives.any { it.name == TYPE_POLICY }
     }
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
@@ -2,7 +2,6 @@ package com.apollographql.apollo3.ast
 
 import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.ast.internal.buffer
-import okio.Buffer
 
 /**
  * A wrapper around a schema GQLDocument that:
@@ -89,6 +88,15 @@ class Schema(
       is GQLEnumTypeDefinition,
       -> setOf(name)
       else -> error("Cannot determine implementedTypes of $name")
+    }
+  }
+
+  /**
+   * Returns whether the `typePolicy` directive is present on at least one object in the schema
+   */
+  fun hasAnyTypePolicyDirectives(): Boolean {
+    return typeDefinitions.values.filterIsInstance<GQLObjectTypeDefinition>().any { objectType ->
+      objectType.directives.any { it.name == TYPE_POLICY }
     }
   }
 
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
@@ -41,7 +41,7 @@ private fun CheckKeyFieldsScope.checkField(
 private fun CheckKeyFieldsScope.checkFieldSet(path: String, selections: List<GQLSelection>, parentType: String, possibleType: String) {
   val implementedTypes = implementedTypes(possibleType)
 
-  val mergedFields = collectFields(selections, parentType, implementedTypes).groupBy {
+  val mergedFields = collectFieldsCached(selections, parentType, implementedTypes).groupBy {
     it.field.name
   }.values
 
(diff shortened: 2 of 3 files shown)


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
```diff
@@ -92,6 +91,15 @@ class Schema(
     }
   }
 
+  /**
+   * Returns whether the `typePolicy` directive is present on at least one object in the schema
+   */
+  fun hasAnyTypePolicyDirectives(): Boolean {
```
martinbonnin, line 97: Nit: rename to `hasTypeWithTypePolicy` ?

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
```diff
@@ -3,14 +3,28 @@ package com.apollographql.apollo3.ast
...
-)
+) {
+  private val implementedTypesCache = mutableMapOf<String, Set<String>>()
+  fun implementedTypes(name: String) = implementedTypesCache.getOrPut(name) {
+    schema.implementedTypes(name)
+  }
+
+  private val keyFieldsCache = mutableMapOf<String, Set<String>>()
+  fun keyFields(name: String) = keyFieldsCache.getOrPut(name) {
+    schema.keyFields(name)
```
martinbonnin, line 14: 👍

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
```diff
@@ -53,6 +67,15 @@ private fun CheckKeyFieldsScope.checkFieldSet(path: String, selections: List<GQL
 
 private class FieldWithParent(val field: GQLField, val parentType: String)
 
+private val collectFieldsCache = mutableMapOf<String, List<FieldWithParent>>()
+private fun CheckKeyFieldsScope.collectFieldsCached(
+    selections: List<GQLSelection>,
+    parentType: String,
+    implementedTypes: Set<String>,
+) = collectFieldsCache.getOrPut("$selections $parentType $implementedTypes") {
+  collectFields(selections, parentType, implementedTypes)
```
martinbonnin, line 76: Do you have any idea the cache hit ratio for this? I'd expect it to be relatively low?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -116,11 +116,15 @@ object ApolloCompiler {
...
     val allFragmentDefinitions = (fragments + incomingFragments).associateBy { it.name }
 
-    operations.forEach {
-      checkKeyFields(it, options.schema, allFragmentDefinitions)
-    }
-    fragments.forEach {
-      checkKeyFields(it, options.schema, allFragmentDefinitions)
+    // Check if all the key fields are present in operations and fragments
+    // (do this only if there are key fields as it may be costly)
+    if (options.schema.hasAnyTypePolicyDirectives()) {
```
martinbonnin, line 121: 👍

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/PerfTest.kt
```diff
@@ -0,0 +1,102 @@
...
+import com.apollographql.apollo3.annotations.ApolloExperimental
+import com.apollographql.apollo3.compiler.TargetLanguage.KOTLIN_1_5
+import com.apollographql.apollo3.compiler.TestUtils.checkTestFixture
+import org.junit.Test
+import java.io.File
+import kotlin.time.ExperimentalTime
+import kotlin.time.measureTime
+
+@OptIn(ExperimentalTime::class, ApolloExperimental::class)
+class PerfTest {
```
martinbonnin, line 12: Could we do everything from `CodegenTest` ? `CodegenTest` writes a `measurements` file containing the codegen time IIRC so it might be enough?

BoD: Yeah I really don't think we should push that test and indeed it should be one of the regular tests ran by `CodegenTest` (in fact it already does because where the files are located, and fails because I didn't generate the .expected 😅). But since it's fairly slow, and we may want to continue investigations, it's easier to have a dedicated test temporarily. If we want to merge this branch, let's remove or move it first.

martinbonnin: Would be nice to merge the other fixes quickly. My moving the test to another PR?

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
BoD: Withe the test in the PR I can see that the method was run `2693` times overall, and only `405` times after making it cached. So `2288` duplicated calls, meaning ~80% cache hit?

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/PerfTest.kt
BoD: 👍  let's do it!

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
```diff
@@ -92,6 +91,15 @@ class Schema(
     }
   }
 
+  /**
+   * Returns whether the `typePolicy` directive is present on at least one object in the schema
+   */
+  fun hasTypeWithTypePolicy(): Boolean {
+    return typeDefinitions.values.filterIsInstance<GQLObjectTypeDefinition>().any { objectType ->
```
martinbonnin, line 98: Note for our future selves: we'll want to support `@typePolicy` on interfaces and unions (see https://github.com/apollographql/apollo-kotlin/issues/3356). This will need to be updated when this happens

martinbonnin (approved):
Nice 👍

PR #3724, opened 2021-12-15
Add an easy way to log cache misses 

martinbonnin:
For users coming from 2.x and relying on `ApolloLogger`, they can now use

```kotlin
ApolloClient.Builder()
   .logCacheMisses()
```


BoD (approved):
💪

PR #3738, opened 2021-12-16
Add a design doc about operation execution

BoD:
I'm not 100% happy with the diagram, but I hope this has some value! At least the process of making it was useful to me 😅


design-docs/Operation execution.md
```diff
@@ -0,0 +1,78 @@
...
+### `Operation` (interface)
+
+Represents a GraphQL operation (mutation, query or subscription).
+
+The generated classes for operations implement this interface.
+
+### `ApolloCall`
+
+Has a reference to an `ApolloClient` and an `Operation` and provides a method to make the `ApolloClient` execute
+the `Operation`.
```
martinbonnin, line 23: One key thing about `ApolloCall` is that it is mutable. It's one of the only classes (maybe the only?) mutable class that is not a `Builder`. It allows to chain configuration calls.

martinbonnin (approved):
Yay for more docs 🙌

PR #3740, opened 2021-12-17
Avoid key field checks on children of fragment spreads

apramana:
In the collectFields method when collecting fields from a GQLFragmentSpread, it seems like only the top-level fields are capable of satisfying the key field requirements of its parent, and so all of the fields in its selection set should have their children removed.

Also, adding tests for fragment spread key field checking.

### DIFF
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
@@ -100,7 +100,13 @@ private fun CheckKeyFieldsScope.collectFields(
         }
 
         val fragmentDefinition = allFragmentDefinitions[it.name]!!
-        collectFields(fragmentDefinition.selectionSet.selections, fragmentDefinition.typeCondition.name, implementedTypes)
+        collectFields(
+            fragmentDefinition.selectionSet.selections.map {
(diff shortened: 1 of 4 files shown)


martinbonnin:
Hi 👋  Thanks for sending this! Unfortunately, I don't think there's an easy way out of this because the fields can be merged, etc...

Removing the nested fields doesn't work in a case like this:

```graphql
query GetBook {
  ...queryFragment
}

fragment queryFragment on Query {
  book {
    # this field will not be visited by the algorithm with this patch and throw while in fact the keyField was correctly queried
    isbn
  }
}
```

I made a small reproducer there: https://github.com/apollographql/apollo-kotlin/commit/5247d61b54f5b1cb18de300d8a42c08b2cfa656e. Let me know if I'm missing anything.

apramana:
Good catch - our project doesn't typically use fragments on operations, so I forgot to consider that case. Would it work if the child removal was limited to fragments on types?

martinbonnin:
> our project doesn't typically use fragments on operations, so I forgot to consider that case. Would it work if the child removal was limited to fragments on types?

That's the thing, the root `Query` type is a normal type like `Book` and the other ones. There's no "fragment on operation" vs "fragment on type", it's all fragment on type. I made a top-level fragment to avoid another level of nesting but wherever you do it, it'll be the same.

apramana:
Ah ok thanks, it makes sense now. I'm going to close this PR and continue discussion in the issue.

PR #3742, opened 2021-12-17
☁️  update to new sonatype host

martinbonnin:
I pushed some snapshots, seems to work 🎉 

https://s01.oss.sonatype.org/content/repositories/snapshots/com/apollographql/apollo3/


BoD (approved):
👍

PR #3744, opened 2021-12-17
Commit yarn.lock

martinbonnin:
Kotlin 1.6.10 introduced the possibility to persist yarn installations and writes a `yarn.lock` file with all the currently installed packages. I understand this needs to be checked in source control so that builds are reproductible

See https://youtrack.jetbrains.com/issue/KT-34014


BoD (approved):
👍

PR #3750, opened 2021-12-20
Specify node version

BoD:
This fixes building on Apple silicon


martinbonnin (approved):

martinbonnin:
Nice 👍 . 

Side note: This logic could also be factored in `build-logic` and later on in a [convention plugin](https://github.com/apollographql/apollo-kotlin/issues/3531)

Issue #3470, opened 2021-10-28
Android Tutorial:  --schema and --endpoint options aren't working for downloadApolloSchema task


aormsby:
**Summary**
For some reason, the provide gradle task to download the Android tutorial schema fails. Well... technically, it succeeds. But it doesn't download anything.
```shell
./gradlew :app:downloadApolloSchema --endpoint='https://apollo-fullstack-tutorial.herokuapp.com/introspectionSchema.json' --schema='app/src/main/graphql/com/example/rocketreserver/schema.json'
```

**Version**
I ran the task with both 2.5.9 and 2.5.10.

**Description**
I wondered if maybe the input values were incorrect, but then I used the deprecated `-Pcom.apollographql.apollo.endpoint` and `-Pcom.apollographql.apollo.endpoint` options. This was successful.

I'd love to think I missed a step or something, but I'm not sure.

martinbonnin:
@aormsby anything else we can help with here?

aormsby:
Nah, worked out well enough

jclement92:
Ran into this issue as well. The path `app/src/main/graphql/com/example/rocketreserver/` did not exist in the initial branch. Added it manually and the download was successful

BoD:
Thanks for reporting it!  We indeed could reproduce the issue this time and fixed the root cause as well as the tutorial instructions so they work with the current version

PR #3752, opened 2021-12-20
Fix downloadApolloSchema task

BoD:
Related to #3470

### DIFF
--- a/docs/source/tutorial/02-add-the-graphql-schema.mdx
+++ b/docs/source/tutorial/02-add-the-graphql-schema.mdx
@@ -27,6 +27,7 @@ In this tutorial, we will use the `apolloDownloadSchema` Gradle task that is cre
 From the root of the project, run the following in Android Studio's Terminal tab:
 
 ```shell:title=(shell)
+mkdir -p app/src/main/graphql/com/example/rocketreserver
 ./gradlew :app:downloadApolloSchema --endpoint='https://apollo-fullstack-tutorial.herokuapp.com/graphql' --schema='app/src/main/graphql/com/example/rocketreserver/schema.graphqls'
 ```
 
--- a/docs/source/tutorial/02-add-the-graphql-schema.mdx
+++ b/docs/source/tutorial/02-add-the-graphql-schema.mdx
@@ -27,8 +27,7 @@ In this tutorial, we will use the `apolloDownloadSchema` Gradle task that is cre
 From the root of the project, run the following in Android Studio's Terminal tab:
 
 ```shell:title=(shell)
-mkdir -p app/src/main/graphql/com/example/rocketreserver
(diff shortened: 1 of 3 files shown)


docs/source/tutorial/02-add-the-graphql-schema.mdx
```diff
@@ -27,6 +27,7 @@ In this tutorial, we will use the `apolloDownloadSchema` Gradle task that is cre
 From the root of the project, run the following in Android Studio's Terminal tab:
 
 ```shell:title=(shell)
+mkdir -p app/src/main/graphql/com/example/rocketreserver
 ./gradlew :app:downloadApolloSchema --endpoint='https://apollo-fullstack-tutorial.herokuapp.com/graphql' --schema='app/src/main/graphql/com/example/rocketreserver/schema.graphqls'
```
martinbonnin, line 31: How come that's not made by the task? 

See https://github.com/apollographql/apollo-kotlin/blob/6299f2acc8d2afd15a5a4669a817daa7f140e7fa/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloDownloadSchemaTask.kt#L120

BoD: It appears the file path was resolved relatively to the "current working directory" which is not the path from which the gradle command is ran, but the gradle folder in the user's home directory. I was convinced the instructions in the tutorial have been working, so I don't understand. But it is possible that it never worked actually?

I changed the task to use a path relative to the project's root (could not find any way to get the "real" current working directory).

I also changed the tutorial's instructions to use `pwd` so it works also with the current published plugin.

martinbonnin: > "current working directory" which is not the path from which the gradle command is ran, but the gradle folder in the user's home directory

This sounds like a bug ?

>  But it is possible that it never worked actually?

Works for me (™️)  using 

* https://github.com/apollographql/apollo-kotlin-tutorial/commit/a415255ff0e1a7d5b5181636b70e0c7def356585
* OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.7+10)

I don't like the idea of using the root project for this as every command line utility out there uses paths relative to the current working directory. Anything else will be super confusing.

BoD: Just to investigate I've added a simple task showing the path for `File(".")` - can you tell me what `./gradlew printcwd` prints in your case?

BoD: Other related issue: https://github.com/gradle/gradle/issues/5187 and https://github.com/gradle/gradle/pull/6139

martinbonnin: So turns out that on my laptop using a debugger, I have:

```kotlin
File("foo").writeText((File("foo").absolutePath))
```

output:
```
$ cat /Users/mbonnin/git/apollo-android-3-tutorial/foo 
/Users/mbonnin/.gradle/daemon/6.7/foo
```

martinbonnin (approved):

PR #3755, opened 2021-12-21
v3.0.1 documentation edits

BoD:
To be merged after 3.0.1 is released


docs/gatsby-config.js
```diff
@@ -64,8 +64,9 @@ module.exports = {
             'advanced/no-runtime',
           ],
           Testing: [
-            'advanced/ui-tests',
-            'advanced/test-builders',
+            'testing/ui-tests',
```
BoD, line 67: Not strictly necessary but I thought it would make sense to move these pages to their own folder (to be consistent with the structure of the doc site). Added redirections in `static/_redirects`  (hoping it will work!)

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
+---
+title: Mocking responses
+---
+
+Apollo Kotlin provides two tools to mock responses in tests:
```
BoD, line 5: If this is too long we can also split in 2 pages (with cross references). LMK what you think :)

ROADMAP.md
```diff
@@ -22,7 +22,7 @@ The declarative cache makes working with the cache and defining unique object id
 
 ### Test server
 
-Apollo Kotlin 3 introduces [test builders](https://www.apollographql.com/docs/kotlin/advanced/test-builders/) as a new type-safe way to build test data. We can go a step further and build an all-in-one solution that includes a GraphQL aware server.
+Apollo Kotlin 3 introduces [test builders](https://www.apollographql.com/docs/kotlin/testing/test-builders/) as a new type-safe way to build test data. We can go a step further and build an all-in-one solution that includes a GraphQL aware server.
```
martinbonnin, line 25: Nice catch 👍

docs/gatsby-config.js
martinbonnin: Agree this is cleaner 👍 

If we're rehauling this, I'd suggest changing the names a little bit and adding an `overview` page like for normalized cache:

* Overview
* Mock HTTP requests
* Mock GraphQL requests
* Test builders
* Espresso IdlingResource

The overview page would explain that it's possible to test end-to-end with a mock HTTP server. In that case, the same ApolloClient can be used for tests and production. Or for lighter tests that don't need to create a server, it's possible to use a MockNetworkTransport.

* For GraphQL tests, the Test Builders can produce an `Operation.Data`.
* For HTTP tests, the Test Builders can produce an Json String (this part still needs to be done but it shouldn't be too hard).

docs/source/advanced/plugin-configuration.mdx
```diff
@@ -73,7 +73,7 @@ apollo {
 By default, the Gradle plugin registers a `downloadApolloSchema` task that you can use from the command line:
 
 ```bash
-# --schema is interpreted relative to the current working directory. This example
+# --schema is interpreted relative to the project's root directory (can also be an absolute path). This example
```
martinbonnin, line 76: Something I realized during the holidays (don't ask me why I was thinking about this!) and I'm dumping here before I forget: We have the same issue with `./gradlew convertApolloSchema`. Just opened https://github.com/apollographql/apollo-kotlin/issues/3764

docs/source/testing/mocking-responses.mdx
martinbonnin: Yep, I'd be in favour of splitting.

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+- `TestNetworkTransport`: allows to specify the GraphQL responses returned by `ApolloClient`
+- `MockServer`: a low-level HTTP server that can be used to mock any responses
+
+In most cases, `TestNetworkTransport` will be more straightforward to use in your tests, while `MockServer` can be
+useful to test specific server behavior, such as error cases, custom headers, timeouts, etc.
```
martinbonnin, line 11: Do we want to be opinionated there? I'm personally the kind of person that'd prefer `MockServer` because this touched as little of the production code as possible (basically just changes the `serverUrl`). Maybe we can gather some feedback what's deemed easiest/more flexible, etc... before making a recommendation?

martinbonnin, line 11: Maybe something like:

```suggestion
`MockServer` creates a full HTTP server. It is useful to test specific server behaviours, such as error cases, HTTP headers, timeouts, etc. It will also require very little changes to your production code as you'll only need to change `serverUrl`.
`TestNetworkTransport` will replace the HTTP call with a preregistered GraphQL response. It is more lightweight and requires more modifications to your production code.
```

Basically changing `"straighforward"` to `"lightweight"`

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+Enable the `TestNetworkTransport` by passing it to the `ApolloClient` builder:
+
+```kotlin{2}
+val apolloClient = ApolloClient.Builder()
+    .networkTransport(TestNetworkTransport())
+    .build()
+```
+
+Either keep a reference to the `TestNetworkTransport` instance, or use the `testNetworkTransport` extension function to
+retrieve it:
```
martinbonnin, line 34: I'd recommend using the extension function like for `ApolloClient.apolloStore`. From my experience this is what most users do.

```suggestion
Use the `testNetworkTransport` extension function to retrieve the `TestNetworkTransport`:
```

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+```kotlin
+val testNetworkTransport = apolloClient.testNetworkTransport
+```
+
+You can then use the `register` function to specify the responses that will be returned in function of the received
+operations:
+
+```kotlin{7}
+val query = GetHeroQuery("001")
```
martinbonnin, line 44:
```suggestion
val testQuery = GetHeroQuery("001")
```

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+val testData = GetHeroQuery.Data {
+  hero = droidHero {
+    name = "R2D2"
+  }
+}
+apolloClient.testNetworkTransport.register(query, testData)
+
+val actual = apolloClient.query(query).execute().data!!
```
martinbonnin, line 52:
```suggestion
apolloClient.testNetworkTransport.register(testQuery, testData)

val actual = apolloClient.query(testQuery).execute().data!!
```

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+assertEquals(testData.hero.name, actual.hero.name)
+```
+
+The `register` function accepts an `Operation` (here, the query) and the `ApolloResponse` to be returned when it is
+executed. As a shortcut you can pass a `Data` directly instead of an `ApolloResponse`, like in the example above.
+
+Note how you can use the [Test builders](./test-builders/) to easily create the data.
```
martinbonnin, line 59: Life is not easy :P. 

Joking asides, "easy" is a very relative concept so different readers might have different expectations. I'd replace with "type-safe", which is objectively true and the same for everyone

```suggestion
Note how you can use the [Test builders](./test-builders/) to create the data in a type-safe way.
```

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+    .execute()
+
+// Don't forget to stop the server when you're done
+mockServer.stop()
+```
+
+Note how the `enqueue` function normally takes a `MockResponse` as a parameter, while a shorthand version taking
+a `String` also exists.
+
+Also note how a delay can be passed to the `MockResponse` constructor to simulate network latency.
```
martinbonnin, line 152: Put this as comments in the sample code? By the time I reached this, I'm too lazy to scroll back 😅

docs/source/testing/mocking-responses.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+Note how you can use the [Test builders](./test-builders/) to easily create the data.
+
+### Advanced usage
+
+In some cases it can be useful to have different responses returned when executing the same operations consecutively.
+
+To that effect, instead of mapping operations to responses with `register`, it is possible to **enqueue** responses:
+they will be returned in the order they were added. To do so, pass a `QueueTestNetworkTransportHandler` to
+the `TestNetworkTransport`, and use its `enqueue` function:
```
martinbonnin, line 67: I'd keep the symmetry with MockServer advanced usage and make this section primarily about defining a custom `TestNetworkTransportHandler`

Reading that doc, I wonder if we should even push the symmetry further and have both `TestNetworkTransport` and `MockServer` use a queue by default. That will make it easier to learn one coming from the other. Or (thinking out loud) have both `Map` and `Queue` options by detault:

```kotlin
class TestNetworkTransport(handler: TestNetworkTransportHandler = QueueTestNetworkTransportHandler) {

  fun enqueue(response: ApolloResponse) // throws if handler is not QueueTestNetworkTransportHandler
  fun register(operation: Operation, response: ApolloResponse) // throws if handler is not RegistryTestNetworkTransportHandler
}

class MockServer(handler: MockServerHandler) {

  fun enqueue(response: HttpResponse)
  fun register(request: HttpRequest, response: HttpResponse)
}

```

docs/gatsby-config.js
StephenBarlow: The redirects look like they'll look great! 😄

BoD: Agreed! I'll do these changes now.

docs/source/testing/mocking-responses.mdx
BoD: Totally agree. Will use your suggestion below.

docs/source/testing/mocking-responses.mdx
BoD: Totally agree 🙏 Trying to avoid this kind of terms, but they slip 😅

docs/source/testing/mocking-responses.mdx
BoD: Good point!

On the MockServer side, I didn't make an equivalent `MapMockServerHandler` at all (not 100% sure it would be useful, and raises a few questions about how `MockRequest` equality should be implemented - I'll have one more look at it).

For `TestNetworkTransport`, I felt `MapTestNetworkTransportHandler` was a sensitive default but tbh I didn't think too much about it :). In the end it'd be nice if the default reflects on "how most users use it", but we don't really know, and I agree that having the same default as the MockServer is a plus (yay consistency!). I'll change the default to use Queue handler, and also add the `enqueue` method (and keep the register one), and update the doc.

Issue #3737, opened 2021-12-16
TestNetworkTransport


martinbonnin:
Now that we have [test builders](https://www.apollographql.com/docs/kotlin/advanced/test-builders/), we can use them to make writing tests easier. The testing can happen at 2 levels:

1. HTTP
2. GraphQL (this issue)

The `NetworkTransport` interface is relatively simple:

```kotlin
interface NetworkTransport {

  fun <D : Operation.Data> execute(
      request: ApolloRequest<D>,
  ): Flow<ApolloResponse<D>>

  fun dispose()
}
```

`TestNetworkTransport` can maintain a `MutableMap<Operation, ApolloResponse>` so that it responds with the pre-recorded `ApolloResponse` for the given operation (that would be build from the test builders).

Building a test client:

```kotlin
val testApolloClient = ApolloClient.Builder()
       .networkTransport(TestNetworkTransport())
       .build()
```

Ideally, the API would look like:

```kotlin
val testNetworkTransport = testApolloClient.testNetworkTransport // this is an extension function that throws if the networkTransport is not a TestNetworkTransport

val query = HeroQuery()
testNetworkTransport.registerResponse(query) {
  hero = humanHero {
    name = "Luke"
  }
}

// Use it
testApolloClient.query(query).execute().data?.hero?.name 
// "Luke" 
```

The `registerOperation {}` part is the hard part because there's no simple generic way to get the `HeroQuery_TestBuilder` from a `HeroQuery` reified type parameter. We might be able to use some compiler plugin magic like [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization/issues/1348) but that adds a new component into the equation.

Another solution might be to have a global registry of `_TestBuilder` keyed by `KClass` but even there I'm not sure how to retrieve them without reflection, which sounds like it's doable on the JVM but not always desired and not possible at all on MPP

BoD:
Just a thought: for the logic returning the `ApolloResponse`, in addition to a "map oriented" API it may be useful to also have a "queue oriented" one and/or one where you can pass a lambda for custom logic.  (I'm thinking about a test scenario where the same query executed twice returns different responses). Something like:

```kotlin
var count = 0
testNetworkTransport.register(query) { query ->
  count++
  when (count) {
    1 -> {
      hero = humanHero {
        name = "Luke"
      }
    }
    2 -> {
      hero = humanHero {
        name = "Han"
      }
   }
}
```

martinbonnin:
[...]

PR #3757, opened 2021-12-23
MockServer improvements and TestNetworkTransport

BoD:
# Step 1
- 0f830092546614e59994e1b9fa6b33d6ea6ccb67
- Add `MockDispatcher` interface which has a `fun dispatch(request: MockRecordedRequest): MockResponse` and modify `MockServer` to use it
- Add an implementation `QueueMockDispatcher`, make it the default one in `MockServer`
- Now `MockServer.enqueue` delegates to this queue dispatcher

**Caveat**: the Apple implementation of `MockServer` needs to `freeze()` the `MockDispatcher`, which prompted me to declare a `copy` method on the `MockDispatcher` interface - this allows to mutate the copy and re-freeze. Could there be a better way? 

# Step 2
- 28a676fd98188306d0f023be8e1db211af5d27fb
- In `apollo-testing-support`, add `ApolloMockDispatcher` interface which is analogous to `MockDispatcher` but returning `ApolloResponse`
- Thanks to a `ApolloMockDispatcherBridge` class and a `fun MockServer(apolloMockDispatcher: ApolloMockDispatcher)` function, it can be used as a regular `MockDispatcher`
- Add 2 implementations:
    - `QueueApolloMockDispatcher` : analogous to `QueueMockDispatcher`
    - `MapApolloMockDispatcher` : has methods to associate `Operation` or `Operation.Data` to `ApolloResponse`

# Step 3
- 387454335bd08e55bcbbf57e284f58af9482b86a
- In `apollo-testing-support`, add `TestNetworkTransport` and a `TestNetworkTransportDispatcher` interface
- `TestNetworkTransportDispatcher` has 2 implementations:
    - `QueueTestNetworkTransportDispatcher`
    - `MapTestNetworkTransportDispatcher` (the default one)
- This is basically what was described in #3737, minus the “magic” to have the ability to pass a test builder lambda

# Thoughts

After looking at this, I am not convinced we should keep what was done in Step 2 (`ApolloMockDispatcher`), because:
- I find the `TestNetworkTransport` to be a better fitting (less low-level) API for users wishing to mock ApolloClient responses
- For our own internal tests, there was already a `MockServer.enqueue(Operation, Data)` extension in `apollo-testing-support` which addresses pretty much the same usecase (I saw it later 🙈)
- With the dispatcher approach you can’t both enqueue a string and an Operation on the same `MockServer`


apollo-api/api/apollo-api.api
```diff
@@ -101,6 +101,12 @@ public final class com/apollographql/apollo3/api/ApolloResponse$Builder {
 	public final fun requestUuid (Ljava/util/UUID;)Lcom/apollographql/apollo3/api/ApolloResponse$Builder;
 }
 
+public final class com/apollographql/apollo3/api/ApolloResponsesKt {
```
martinbonnin, line 104: Use `@JvmName("ApolloResponses")` to make Java interop nicer?

martinbonnin: Also what about merging with `ResponseParser` and have something like `ApolloResponseAdapter` that can adapt to//from json like we do for other models?

apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -31,10 +31,11 @@ import kotlin.native.concurrent.freeze
...
  */
 actual class MockServer(
-    private val acceptDelayMillis: Long
-): MockServerInterface {
+    private val acceptDelayMillis: Long,
+    mockDispatcher: MockDispatcher = QueueMockDispatcher(),
+) : BaseMockServer(mockDispatcher) {
 
-  actual constructor(): this(0)
+  actual constructor(mockDispatcher: MockDispatcher) : this(0, mockDispatcher)
```
martinbonnin, line 38: I'm slightly concerned about the `"Dispatcher"` naming here since it's also used extensively in coroutines code. May I suggest `"Handler"` instead, like in [Servlets](https://archive.eclipse.org/jetty/8.0.0.M1/apidocs/org/eclipse/jetty/servlet/ServletHandler.html) and [Ktor](https://ktor.io/docs/requests.html) ?

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -26,7 +33,12 @@ interface MockServerInterface {
   fun takeRequest(): MockRecordedRequest
 }
 
+abstract class BaseMockServer(override val mockDispatcher: MockDispatcher) : MockServerInterface {
+  override fun enqueue(mockResponse: MockResponse) {
+    (mockDispatcher as QueueMockDispatcher).enqueue(mockResponse)
+  }
+}
```
martinbonnin, line 40: Nitpicking: I understand this is used to factor in the `enqueue` between all platforms? I wouldn't mind inlining it to avoid the abstract class.

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -26,7 +33,12 @@ interface MockServerInterface {
   fun takeRequest(): MockRecordedRequest
 }
 
+abstract class BaseMockServer(override val mockDispatcher: MockDispatcher) : MockServerInterface {
+  override fun enqueue(mockResponse: MockResponse) {
+    (mockDispatcher as QueueMockDispatcher).enqueue(mockResponse)
```
martinbonnin, line 38:
```suggestion
    (mockDispatcher as? QueueMockDispatcher)?.enqueue(mockResponse) ?: error("Apollo: cannot call MockeServer.enqueue() with a custom dispatcher")
```

(or s/dispatcher/handler if we decide to rename)

martinbonnin: In retrospect, I'm not a huge fan of the `MockServer` naming either because of the possible "mockito" confusion but certainly too late to change this 🤷

apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/MockServerCommonTest.kt
```diff
@@ -1,12 +1,36 @@
...
 class MockServerCommonTest {
+  private lateinit var mockServer: MockServer
+
+  private fun setUp() {
+    mockServer = MockServer()
+  }
+
+  private suspend fun tearDown() {
+    mockServer.stop()
+  }
```
martinbonnin, line 32: Can we split this file in 2? 

* `ReadRequestTest` (doesn't require a mockServer)
* `MockServerTest` (requires a mockServer)

martinbonnin: 3 actually 😇 ?

* `ReadRequestTest` (doesn't require a mockServer)
* `EnqueueTest` (requires a default mockServer) 
* `CustomDispatcherTest` (requires a custom mockServer)

apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/MockServerCommonTest.kt
```diff
@@ -48,4 +72,88 @@ class MockServerCommonTest {
...
+      val httpResponse = engine.execute(HttpRequest.Builder(HttpMethod.Get, mockServer.url()).build())
+      assertMockResponse(mockResponse, httpResponse)
+    }
+  }
+
+  @Test
+  fun exceptionWhenNothingWasEnqueued() = runTest(before = { setUp() }, after = { tearDown() }) {
+    val engine = DefaultHttpEngine(timeoutMillis = 10)
+    assertFailsWith<ApolloNetworkException> {
+      engine.execute(HttpRequest.Builder(HttpMethod.Get, mockServer.url()).build())
```
martinbonnin, line 104: Very cool to explicitely require that behaviour 👍 . I was never sure.

apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
BoD: Agreed!

BoD: Are you ok with `MockServerHandler`?  A bit verbose, but I'm afraid the shorter `MockHandler` would be too broad.

martinbonnin: That works for me. I can also suggest `HttpHandler`. Your call.

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/ApolloMockDispatcher.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.apollographql.apollo3.api.composeJsonResponse
+import com.apollographql.apollo3.api.json.buildJsonString
+import com.apollographql.apollo3.mockserver.MockDispatcher
+import com.apollographql.apollo3.mockserver.MockRecordedRequest
+import com.apollographql.apollo3.mockserver.MockResponse
+import com.apollographql.apollo3.mockserver.MockServer
+
+interface ApolloMockDispatcher {
+  val customScalarAdapters: CustomScalarAdapters
+  fun dispatch(request: MockRecordedRequest): ApolloResponse<out Operation.Data>
```
martinbonnin, line 16: I'm still digging into the diff but I agree with your general sentiment that this is not really needed as there is:

* `MockServer` for testing at the HTTP level
* `TestNetworkTransport` for testing at the GraphQL level

settings.gradle.kts
```diff
@@ -1,4 +1,4 @@
-rootProject.name = "apollo-android"
+rootProject.name = "apollo-kotlin"
```
martinbonnin, line 1: 👍

tests/test-network-transport/src/test/kotlin/testnetworktransport/MapTestNetworkTransportDispatcherTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        )
+    )
+    apolloClient.testNetworkTransport.register(query, testData)
+
+    val actual = apolloClient.query(query).execute().data
+    assertEquals(testData, actual)
+  }
+
+  @Test
+  fun registerDataTestBuilder() = runTest(before = { setUp() }, after = { tearDown() }) {
```
martinbonnin, line 113: I don't think we need both `Manual` and `TestBuilder` versions of this test. How the `data` is constructed is tested in other tests already.

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
```diff
@@ -51,10 +50,15 @@ class MockResponse(
   constructor(
       body: String,
       statusCode: Int = 200,
-      headers: Map<String, String> = emptyMap()
+      headers: Map<String, String> = emptyMap(),
   ) : this(statusCode, body.encodeUtf8(), headers)
 }
 
+interface MockDispatcher {
+  fun dispatch(request: MockRecordedRequest): MockResponse
```
martinbonnin, line 58: Naming: it feels weird having `"Recorded"` in this API as it's not really recorded yet. What about introducing `MockRequest` ?

BoD: Agreed, `MockRequest` makes more sense!

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
BoD: 👍 done in 313ef8bffbd82dd309f20e7713f04f1e5a395684

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/ApolloMockDispatcher.kt
BoD: Yeah that seems sensible. If you're ok I'll axe this part (no regrets, at least I learned a few things doing it!).

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
```diff
@@ -51,11 +50,16 @@ class MockResponse(
...
       statusCode: Int = 200,
-      headers: Map<String, String> = emptyMap()
+      headers: Map<String, String> = emptyMap(),
   ) : this(statusCode, body.encodeUtf8(), headers)
 }
 
-internal fun readRequest(source: BufferedSource): MockRecordedRequest? {
+interface MockServerHandler {
+  fun handle(request: MockRequest): MockResponse
+  fun copy(): MockServerHandler
```
martinbonnin, line 59: Not a huge fan of the `copy()` because:

1. it's a K/N detail that JVM implementers shouldn't need to worry about.
2. in some scenarios, it makes it possible to "miss" some enqueued response. For an example if enqueuing a response while another one is being handled, the queue weil be overwritten.

I think we can make a `QueueMockDispatcher` that survives freezing by using a `NSMutableArray`  under the hood, like in `Socket.kt`. Let me know if you want to try, if not, I'll give it a shot.

BoD: Yeah I agree it's less than ideal!

Ok I'll try making `QueueMockDispatcher` `expect` with a specific implementation for Apple. 

There's still an expectation that custom implementations must be "freeze safe" - but a comment on the interface may be enough?  Something like _"Note that on Apple platform, instances will be frozen, and thus need to be immutable (this is not required for other platforms)"_?

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/QueueMockServerHandler.kt
```diff
@@ -0,0 +1,19 @@
+package com.apollographql.apollo3.mockserver
+
+class QueueMockServerHandler : MockServerHandler {
```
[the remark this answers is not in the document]
BoD: No, you're right 👍

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
martinbonnin: I think this is a reasonable expectation. No need to mention `freeze` explicitely, especially if we're planning to remove them in the mid-term. I'd focus at the "multi-threaded" portion. Then every platform can do what they want (freeze, lock, remove all side effects, etc...)

```
  /**
   * handles the given [MockRequest]
   * 
   * This method is called from one or several background thread and should be thread-safe.  
   */
  fun handle(request: MockRequest): MockResponse
```

martinbonnin: mmmm and re-reading my comment, I just understand what you meant 🤦 ... Maybe we can remove the `freeze()` from the apollo code? Leave it to the implementation instead? 

`QueueMockServerHandler` would just "freeze" itself from its constructor. Someone using the new MM (or just having a pure handler) could skip this

BoD: The thing is `MockServerHandler` is not frozen explicitly but "transitively" as it's a property of `Socket` which needs to be frozen. 

I actually agree with your comment on not mentioning `freeze` explicitly and instead talking about threading, I think it makes sense.

Making an Apple specific `QueueMockDispatcher` using `NSMutableArray` did work! 💪  (used `expect`/`actual` in the end) - coming in the next commit.

martinbonnin: > The thing is MockServerHandler is not frozen explicitly but "transitively" as it's a property of Socket which needs to be frozen.

Right. I see... We _could_ expose a `Factory` to move the creation to a later point in time. But TBH I don't think it's worth the extra complexity. Let's start with freezing everything and if we need to we can always introduce factories later on?

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/TestNetworkTransport.kt
```diff
@@ -0,0 +1,86 @@
...
+
+interface TestNetworkTransportHandler {
+  fun handle(request: ApolloRequest<*>): ApolloResponse<*>
+}
+
+class QueueTestNetworkTransportHandler : TestNetworkTransportHandler {
+  private val queue = ArrayDeque<ApolloResponse<out Operation.Data>>()
+
+  fun <D : Operation.Data> enqueue(response: ApolloResponse<D>) {
+    queue.add(response)
```
martinbonnin, line 43: I think this needs to be locked. On native it doesn't throw because we run the flow on the main thread always but it might not always be the case. And on the JVM we cannot make any assumption about what dispatcher is going to be used.

martinbonnin (approved):
One last concurrency comment but looks great 🤩 . That'll be a nice addition!

martinbonnin (approved):

BoD:
Added some documentation [here](https://github.com/apollographql/apollo-kotlin/pull/3755/commits/f4cd5e20cdec1c5d86430d283b7d6d38829a30c2)

martinbonnin (approved):

Issue #3759, opened 2021-12-29
Allow building an ApolloClient on top of another one changing the underlying OkHttpClient


francescocervone:
**Use case**
Let's say I have 
```kotlin
val okHttpClient1: OkHttpClient = someFactory()
val client1: ApolloClient = ApolloClient.Builder()
  .okHttpClient(okHttpClient1)
  .build()
```

Then I need to create a new Apollo client on top of `client1` passing a new `OkHttpClient`

```kotlin
val okHttpClient2: OkHttpClient = someNewFactory()
val client2: ApolloClient = client1.newBuilder()
  .okHttpClient(okHttpClient2)
  .build()
```

With Apollo 3.0.0 this is not possible, since `build` throws an exception
```
Apollo: 'httpEngine' has no effect if 'networkTransport' is set
```

I cannot even put `networkTransport` to `null` because `ApolloClient.Builder.networkTransport(...)` expects a non-nullable variable.

**Describe the solution you'd like**
I expect that Apollo allows me to build an `ApolloClient` on top of another one (built with an `OkHttpClient`) with a new `OkHttpClient`.

martinbonnin:
Good catch, that's a side effect of `ApolloClient.Builder.okHttpClient()` being a "shortcut" to simplify the setup. 

You can use `ApolloClient.Builder.networkTransport()` to override it:

```kotlin
    apolloClient
        .newBuilder()
        .networkTransport(
            HttpNetworkTransport.Builder()
                .serverUrl(serverUrl2)
                .okHttpClient(okHttpClient2)
                .build()
        )
```

I think making `networkTransport` nullable in `ApolloClient.Builder.networkTransport()` is a reasonable thing too to reset the network transport.

The other solution to respect 100% the `Builder` semantics is to store all "shortcut" properties in `ApolloClient` and forward them in `newBuilder()`. No strong opinion on which one is better yet. That last one is more consistent I believe but has a small price in terms of making `ApolloClient` bigger and more complex. Any thoughts?

francescocervone:
I think making `networkTransport` nullable is a better fit.
Since `ApolloClient.Builder.okHttpClient` is an extension function hiding the detail about of how Apollo actually works with `OkHttpClient`, it should also hide the detail that the `networkTransport` (and maybe `subscriptionNetworkTransport`) must be set to null again.

I expect something like this:
```kotlin
fun ApolloClient.Builder.okHttpClient(okHttpClient: OkHttpClient) = apply {
[...]

PR #3771, opened 2022-01-06
Allow new builder with shortcuts

BoD:
Related to #3759.

In the [first commit](https://github.com/apollographql/apollo-kotlin/pull/3771/commits/ef8638cdef75e7616d7c1a9306cb24d8fafe1934) I kept builder "shortcuts" references in `ApolloClient` in order to use them in `newBuilder`. That's a bit "boilerplate-y" but then I realized if we're going to do this, why not simply keep the builder itself, which encapsulates all these references? Which I did in [second commit](https://github.com/apollographql/apollo-kotlin/pull/3771/commits/23223b17385df21d01a138a3860bf0d86507b26f). This almost seems too easy so not sure if I overlooked something 😅? Let me know what you think.

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -560,12 +560,14 @@ private constructor(
           .sendDocument(sendDocument)
           .enableAutoPersistedQueries(enableAutoPersistedQueries)
           .canBeBatched(canBeBatched)
+      _networkTransport?.let { builder.networkTransport(it) }
       httpServerUrl?.let { builder.httpServerUrl(it) }
       httpEngine?.let { builder.httpEngine(it) }
       httpExposeErrorBody?.let { builder.httpExposeErrorBody(it) }
       for (httpInterceptor in httpInterceptors) {
         builder.addHttpInterceptor(httpInterceptor)
       }
+      subscriptionNetworkTransport?.let { builder.subscriptionNetworkTransport(it) }
       webSocketServerUrl?.let { builder.webSocketServerUrl(it) }
       webSocketEngine?.let { builder.webSocketEngine(it) }
       webSocketReconnectWhen?.let { builder.webSocketReconnectWhen(it) }
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -57,6 +57,19 @@ private constructor(
     override val sendDocument: Boolean?,
     override val enableAutoPersistedQueries: Boolean?,
     override val canBeBatched: Boolean?,
+
+    // Builder shortcuts for networkTransport
+    private val httpServerUrl: String?,
+    private val httpEngine: HttpEngine?,
+    private val httpInterceptors: List<HttpInterceptor>,
+    private val httpExposeErrorBody: Boolean?,
+
+    // Builder shortcuts for subscriptionNetworkTransport
+    private val webSocketServerUrl: String?,
+    private val webSocketEngine: WebSocketEngine?,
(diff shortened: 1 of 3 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -541,24 +541,15 @@ private constructor(
...
-        .interceptors(interceptors)
-        .requestedDispatcher(requestedDispatcher)
-        .executionContext(executionContext)
-        .httpMethod(httpMethod)
-        .httpHeaders(httpHeaders)
-        .sendApqExtensions(sendApqExtensions)
-        .sendDocument(sendDocument)
-        .enableAutoPersistedQueries(enableAutoPersistedQueries)
-        .canBeBatched(canBeBatched)
+    return builder
```
martinbonnin, line 552: This needs to copy the builder else doing

```
val builder1 = apolloClient.newBuilder()
val builder2 = apolloClient.newBuilder()

builder1.serverUrl("https://example.com/graphql1")
// builder2.serverUrl has changed too

```

And since copying will require some boilerplate, might as well keep everything under `ApolloClient` to avoid storing properties twice?

BoD: Ah you're right, this could be misleading.  For the sake of comparison I've tried the `copy` method in the [last commit](https://github.com/apollographql/apollo-kotlin/pull/3771/commits/a4bfcf7f461c4e6b9da1dd773352f42a3048e70e), I think the code looks a bit nicer than having this logic in `ApolloClient`, but let me know what you think.

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -541,24 +541,41 @@ private constructor(
           sendDocument = sendDocument,
           enableAutoPersistedQueries = enableAutoPersistedQueries,
           canBeBatched = canBeBatched,
+
+          // Keep a reference to this Builder shortcuts so newBuilder() behaves as expected
```
martinbonnin, line 545:
```suggestion
          // Keep a reference to the Builder so we can keep track of `httpEngine` and other properties that 
          // are important to rebuild `networkTransport` (and potentially others)
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -541,24 +541,41 @@ private constructor(
...
+      httpEngine?.let { builder.httpEngine(it) }
+      httpExposeErrorBody?.let { builder.httpExposeErrorBody(it) }
+      for (httpInterceptor in httpInterceptors) {
+        builder.addHttpInterceptor(httpInterceptor)
+      }
+      webSocketServerUrl?.let { builder.webSocketServerUrl(it) }
+      webSocketEngine?.let { builder.webSocketEngine(it) }
+      webSocketReconnectWhen?.let { builder.webSocketReconnectWhen(it) }
+      webSocketIdleTimeoutMillis?.let { builder.webSocketIdleTimeoutMillis(it) }
+      wsProtocolFactory?.let { builder.wsProtocol(it) }
```
martinbonnin, line 572: I think `networkTransport` is missing here?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: Works for me!

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
BoD: Wow good catch! Fixed.

martinbonnin (approved):

PR #3772, opened 2022-01-07
Make `reconnectWhen` suspend and pass attempt number

### LINKED ISSUE #3766: Migrating to Apollo 3: Usage of `SubscriptionManager`
While migrating from Apollo 2 to Apollo 3, I can't find a trivial replacement for usage of `ApolloClient::addOnSubscriptionManagerStateChangeListener` and `ApolloClient::subscriptionManager.reconnect()`.

Any clues/suggestions ? 🙏

AchrafAmil:
### Reason

- Fixes #3766

### Changelist

- Accept `suspend` function as arg of `ApolloClient::webSocketReconnectWhen` and `WebSocketNetworkTransport::reconnectWhen`.
- Along with the `Throwable`, also pass the attempt number to the `reconnectWhen` predicate.

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -323,7 +323,8 @@ private constructor(
       this.webSocketReconnectWhen = webSocketReconnectWhen
     }
 
-    fun webSocketReconnectWhen(reconnectWhen: (suspend (Throwable) -> Boolean)?) = apply {
+    @Deprecated("Use the overload where the lambda is suspend and has more args")
+    fun webSocketReconnectWhen(reconnectWhen: ((Throwable) -> Boolean)?) = apply {
       this.webSocketReconnectWhen = reconnectWhen?.let {
         val adaptedLambda: suspend (Throwable, Long) -> Boolean = { throwable, _ -> reconnectWhen(throwable) }
         adaptedLambda
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -581,7 +581,7 @@ private constructor(
       subscriptionNetworkTransport?.let { builder.subscriptionNetworkTransport(it) }
       webSocketServerUrl?.let { builder.webSocketServerUrl(it) }
       webSocketEngine?.let { builder.webSocketEngine(it) }
-      webSocketReconnectWhen?.let { builder.webSocketReconnectWhen(it) }
+      webSocketReopenWhen?.let { builder.webSocketReopenWhen(it) }
       webSocketIdleTimeoutMillis?.let { builder.webSocketIdleTimeoutMillis(it) }
       wsProtocolFactory?.let { builder.wsProtocol(it) }
       return builder
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -180,7 +180,7 @@ private constructor(
     private var wsProtocolFactory: WsProtocol.Factory? = null
     private var httpExposeErrorBody: Boolean? = null
     private var webSocketEngine: WebSocketEngine? = null
-    private var webSocketReconnectWhen: ((Throwable) -> Boolean)? = null
+    private var webSocketReconnectWhen: (suspend (Throwable) -> Boolean)? = null
 
     override var httpMethod: HttpMethod? = null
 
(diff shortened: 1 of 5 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -313,15 +313,23 @@ private constructor(
...
+     * It is a suspending function, so it can be used to introduce delay before retry (e.g. backoff strategy).
      *
      * See also [subscriptionNetworkTransport] for more customization
      */
-    fun webSocketReconnectWhen(webSocketReconnectWhen: ((Throwable) -> Boolean)) = apply {
+    fun webSocketReconnectWhen(webSocketReconnectWhen: (suspend (Throwable, attempt: Long) -> Boolean)) = apply {
       this.webSocketReconnectWhen = webSocketReconnectWhen
     }
 
+    fun webSocketReconnectWhen(reconnectWhen: (suspend (Throwable) -> Boolean)?) = apply {
```
BoD, line 326: Thanks for keeping this one to avoid a breaking change 🙏!  Would it be possible to keep the lambda non suspending for this one? That would ensure binary compatibility as well.

Also I think we can mark it as `@Deprecated`.
```suggestion
    @Deprecated("Use webSocketReconnectWhen(webSocketReconnectWhen: (suspend (Throwable, attempt: Long) -> Boolean))")
    @ApolloDeprecatedSince(v3_0_1)
    fun webSocketReconnectWhen(reconnectWhen: ((Throwable) -> Boolean)?) = apply {
```

(Same goes for WebSocketNetworkTransport.kt below)

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -180,7 +180,7 @@ private constructor(
     private var wsProtocolFactory: WsProtocol.Factory? = null
     private var httpExposeErrorBody: Boolean? = null
     private var webSocketEngine: WebSocketEngine? = null
-    private var webSocketReconnectWhen: ((Throwable) -> Boolean)? = null
+    private var webSocketReconnectWhen: (suspend (Throwable, attempt: Long) -> Boolean)? = null
```
BoD, line 183: Nitpick: an Int is probably enough?

AchrafAmil: Having an int will be fine for almost all cases indeed.
Though, using a `Long` doesn't cost anything and helps avoid silent bugs for anyone trying an intensive usage (e.g. tests).
Moreover, I just mimicked kotlin flow's `retryWhen`
```kotlin
fun <T> Flow<T>.retryWhen(predicate: suspend FlowCollector<T>.(cause: Throwable, attempt: Long) -> Boolean): Flow<T>
```

BoD: Fair enough! 👍

BoD (approved):
Thanks a lot! 🎉

BoD:
Oh noes [this test](https://github.com/apollographql/apollo-kotlin/blob/main/tests/websockets/src/jvmTest/kotlin/SampleServerTest.kt#L235) doesn't compile because there is now an ambiguity 😞 . The lambda we pass clearly has only one parameter, so it should be clear which method we mean, but it looks like the compiler is not "smart" enough to see that.

One possible solution would be to rename the new one but that's unfortunate - I can't think of a better name.

martinbonnin:
For the record, the Youtrack issue about overload resolution ambiguity: https://youtrack.jetbrains.com/issue/KT-11265

> Another option is to consider this is not a big deal and allow this "source only" breaking change (and the fix is trivial).

I'm pretty sure adding `suspend` is a binary incompatible change. Because the compiler generates a JVM method that takes a `Continuation`, etc...

What about we use `DeprecationLevel.HIDDEN`?

```kotlin
@Deprecated(message = "use the suspend version with 2 parameters instead", level = DeprecationLevel.HIDDEN)
fun webSocketReconnectWhen(webSocketReconnectWhen: ((Throwable) -> Boolean)) {
  ...
}

fun webSocketReconnectWhen(webSocketReconnectWhen: suspend ((Throwable, Int) -> Boolean)) {
  ...
}
```

That's a source-incompatible change (because callers will have to add the second lambda parameter). But I _think_ it's binary compatible?

AchrafAmil:
What about renaming `reconnectWhen` to `reopenWhen` (and `webSocketReopenWhen`) ?

That being said, the "source only" breaking change with a trivial fix seems ok to me, too.

I'll let you decide whatever you think is better.

AchrafAmil:
What about renaming `reconnectWhen` to `reopenWhen` (and `webSocketReopenWhen`) ?

That being said, the "source only" breaking change with a trivial fix seems ok to me, too.

I'll let you decide whatever you think is better.

martinbonnin:
> What about renaming reconnectWhen to reopenWhen (and webSocketReopenWhen) ?

I like it 👍

BoD:
`reopenWhen` is ok for me too 👍

AchrafAmil:
Renamed!

Please notice I also replaced/renamed usage in tests. Let me know if we prefer to keep testing the deprecated one.

BoD:
(the build failed because of changes on `main`, I've rebased / pushed)

Issue #3769, opened 2022-01-06
Http Headers Not Included When Batching is Enabled on Version 3


simon-the-canadian:
Version 3.0.0

Hi, I am adding an Http Header to my graphQL call using `ApolloClient.Builder().addHttpHeader()`.
The header along with some other apollo headers show up in the network request made when batching is turned off. However when I turn on batching then none of those headers are in the request.

This is a screenshot of the request headers when batching is off.
![Screen Shot 2022-01-06 at 11 26 19 AM](https://user-images.githubusercontent.com/17653830/148432668-c42a8c03-e1f4-48e8-99b8-5bc000a2ecf6.png)

This is a screenshot of the same request but with batching turned on. All I change is add `.httpBatching(maxBatchSize = 5)` to the client.
![Screen Shot 2022-01-06 at 11 19 19 AM](https://user-images.githubusercontent.com/17653830/148432810-63aa7aa5-b0a9-417b-b1c2-2c593c19c193.png)

I haven't tried adding an interceptor to the OkHttpClient, maybe that would work. It's just a lot simpler if I can set the http header directly on the apollo client.

BoD:
Hi! Thanks for reaching out.

I believe this is because the headers are not taken into account when creating a "composite" `HttpRequest` [here](https://github.com/apollographql/apollo-kotlin/blob/4b02a22ff2b39b9c2aa05a68aaff467276fc8669/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt#L159), which I think is an oversight (but should not be too hard to fix).

In the meantime, yes I believe using an `OkHttp` interceptor should work!

martinbonnin:
There's a question about how to "merge" HTTP headers. If we batch several operations, they will have different `X-APOLLO-OPERATION-ID` for an example. Which one should we ultimately send over the wire?

BoD:
@martinbonnin Good point!

I think we should continue omitting the X-APOLLO-* headers (unless I’m missing something!). For the other ones, I can see several options:

1. Keep current behavior (discard headers). Since several merging strategies can exist, it may be better for the library to avoid choosing one, as it may not match some users’ needs. We can clarify this and recommend using an interceptor in the doc.
2. Merge headers, keep all duplicates
request1: `Header1: A`, `Header2: A`
request2: `Header1: A`, `Header2: B`, `Header3: A`, 
[...]

PR #3776, opened 2022-01-10
Merge HTTP headers when batching

BoD:
Fix for #3769.

Now headers present with the same value on all requests will be kept.


martinbonnin (approved):
Tiny nit: I'd use `commonHeaders` instead of `mergedHeaders` because merging always make me thing of a "union" more than an "intersection" but really that's not super important. 

Looks good to me 👍

Issue #3748, opened 2021-12-18
Apollo 3 customScalarsMapping for ID


Saecki:
After upgrading to apollo 3, and replacing `customTypeMapping` with `customScalarsMapping`, mapping the `ID` scalar to `kotlin.Long` will throw this error: 
```
Apollo: unknown custom scalar(s) in customScalarsMapping: ID
```
Is there another way to map standard scalars now, or is this usecase simply not supported?

martinbonnin:
Good catch. Indeed, this hasn't made it to 3.x. The overall reason is that `ID` isn't technically a custom scalar since it's builtin and should always be serializable to a `String`. From [the spec](https://spec.graphql.org/draft/#sec-ID):

```
The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. 
The ID type is serialized in the same way as a String; however, it is not intended to be human-readable. 
While it is often numeric, it should always serialize as a String.
```

I'm curious what the use case is. Do you need to do some maths on the IDs? Or is it because storing a Long is more compact than a String? Or something else?

Saecki:
Yeah, peace of mind. It probably doesn't matter but I like the thought of `ID`s not being heap allocated and fixed size.
Also comparing elements by an `ID` of a numeric type should be much faster than comparing strings.

martinbonnin:
👍 That's fair. We'll look into adding this back. 

If we're doing this, I think `ID` shouldn't be a special case but it should be possible to map any scalar (builtin or custom):

```kotlin
/**
 *
 * @param graphqlName: the name of the scalar in the schema
 * @param kotlinName: fully qualified name of the kotlin type to use in models
 * @param adapterName: fully qualified name of the adapter used to adapt the Graph type to the Kotlin type
 * The adapter must have a no-arg constructor or be an `object` singleton
 */
fun Service.mapScalar(graphqlName: String, kotlinName: String, adapterName: String?)
```

That'd make it possible to avoid the lookups in the `CustomScalarAdapters` at runtime:

```kotlin
apollo {
  mapScalar("ID", "kotlin.Long", "com.apollographql.apollo3.api.LongAdapter")
}
```

We could even think of nice overrides for the (Apollo) builtin adapters:

```kotlin
apollo {
  mapScalarToLong("ID")
  mapScalarToMap("Json")
  ...
}
```

Saecki:
That looks great!

simon-the-canadian:
[...]

PR #3779, opened 2022-01-10
Scalar mapping and adapter configuration improvements

BoD:
Related to #3748.

For now here is a design document with possible changes.


design-docs/Scalar adapter configuration.md
```diff
@@ -0,0 +1,195 @@
...
+sealed interface AdapterInitializer
+
+// The adapter will be instantiated in the generated code
+class SingletonAdapterInitializer(val qualifiedName: String): AdapterInitializer
+
+// The adapter will be used as-is (it's an object or a public val)
+class NoArgConstructorAdapterInitializer(val qualifiedName: String): AdapterInitializer
+
+// The adapter will be looked up in the `customScalarAdapters` parameter (same as current behavior)
+object RuntimeAdapterInitializer: AdapterInitializer
```
martinbonnin, line 107: I know I'm the one who suggested this initially but I think we should avoid Kotlin APIs in the Gradle plugin API. 

Mainly for Groovy users + because we use R8 to relocate all of the Kotlin-stdlib, this feels dangerous. I'd suggest something like this instead:

```kotlin
class Service {
  fun mapScalar(graphQLName: String, kotlinName: String)
  fun mapScalarUsingKotlinSingleton(graphQLName: String, kotlinName: String, singletonName: String)
  fun mapScalarUsingKotlinConstructor(graphQLName: String, kotlinName: String, constructorName: String)

  fun mapScalarToKotlinLong(graphQLName: String)
  fun mapScalarToKotlinInt(graphQLName: String)
  fun mapScalarToKotlinString(graphQLName: String)
  fun mapScalarToKotlinBoolean(graphQLName: String)
  fun mapScalarToKotlinMap(graphQLName: String)

  fun mapScalarToJavaLong(graphQLName: String)
  fun mapScalarToJavaInteger(graphQLName: String)
  fun mapScalarToJavaString(graphQLName: String)
  fun mapScalarToJavaBoolean(graphQLName: String)
  fun mapScalarToJavaMap(graphQLName: String)
}
```

design-docs/Scalar adapter configuration.md
```diff
@@ -0,0 +1,195 @@
...
+
+Let's also have convenience shortcuts for the types for which we have built-in adapters:
+
+```kotlin
+apollo {
+  // equivalent to mapScalar("ID", "kotlin.Long", SingletonAdapterInitializer("com.apollographql.apollo3.api.LongAdapter")
+  mapScalarToLong("ID")
+
+  // equivalent of mapScalar("Json", "kotlin.Any", SingletonAdapterInitializer("com.apollographql.apollo3.api.AnyAdapter")
+  mapScalarToMap("Json")
```
martinbonnin, line 124:
```suggestion
  mapScalarToKotlinAny("Json")
```

design-docs/Scalar adapter configuration.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+With this, it is no longer necessary (but still possible) to register the adapters at runtime with `addCustomScalarAdapter`.
+
+#### Compatibility
+
+We need to keep the current mechanism (`customScalarsMapping` + `addCustomScalarAdapter`) working of course.
+
+Let's make `customScalarsMapping` call `mapScalarToMap` internally and mark it as deprecated.
```
martinbonnin, line 136:
```suggestion
Let's make `customScalarsMapping` call `mapScalar` internally and mark it as deprecated.
```

design-docs/Scalar adapter configuration.md
```diff
@@ -0,0 +1,195 @@
...
+Let's change it to a `Map<String, ScalarInfo>`, with `ScalarInfo` being `data class ScalarInfo(val targetName: String, val adapterInitializer: AdapterInitializer)`.
+
+We need to pass it to `KotlinResolver` / `JavaResolver` because this is where the logic to get the adapter lies (`adapterInitializer()`, `nonNullableAdapterInitializer()`). 
+
+This can be done by passing it to `KotlinCodeGen` / `JavaCodeGen` which instantiate the resolvers.
+
+#### Remaining points
+
+- Avoid instantiating the no arg constructors adapters at each use
+  -> Generate fields in a specific object (e.g. `ScalarAdapters`) and use them in the generated code?
```
martinbonnin, line 194: We've historically tried to avoid that kind of lazy logic because it requires synchronization for proper multi-threading support. I think making a new instance everytime is ok. The caller can always use a singleton if they want to optimize

design-docs/Scalar adapter configuration.md
```diff
@@ -0,0 +1,195 @@
...
+- Need a Type class for built-in scalars (e.g. `ID`) when using `RuntimeAdapterInitializer`
```
martinbonnin, line 195: Not sure I'm following what's the issue here?

design-docs/Scalar adapter configuration.md
BoD: Agreed! 
Plus I actually find the syntax easier to use 👍

Shouldn't `mapScalarUsingKotlinSingleton` / `mapScalarUsingKotlinConstructor` be `mapScalarUsingSingleton` / `mapScalarUsingConstructor`?

design-docs/Scalar adapter configuration.md
BoD: Oh ok - I actually have a prototype that works, but you're right about potential risk / it being not that useful. Removing (will make the implementation simpler) 👍

design-docs/Scalar adapter configuration.md
BoD: Just me not seeing yet how this works 😅  As it happens it's not an issue.

design-docs/Scalar adapter configuration.md
martinbonnin: > Shouldn't mapScalarUsingKotlinSingleton / mapScalarUsingKotlinConstructor be mapScalarUsingSingleton / mapScalarUsingConstructor?

I was wondering the same. `Singleton` is a Kotlin concept so it will not work with Java but maybe we can throw a nice error if that happens?

BoD: I was assuming this would work also in Java with static fields:
```java
public class MyAdapters {
  public static final Adapter<MyDate> DATE_ADAPTER = new Adapter<MyDate>() {...}
}
```

```kotlin
mapScalarUsingSingleton("Date", "com.example.MyDate", "com.example.MyAdapters.DATE_ADAPTER")
```

But now I'm realizing maybe the name "singleton" is not the most accurate?

BoD: I think I agree, the distinction is not that useful. The documentation can explain how it works, and the fact that an instance or an instantiation expression can both be used.

In that case, maybe we can keep the name `mapScalar`?

```kotlin
  /**
   * Map a GraphQL scalar type to the Java/Kotlin type and adapter configured at runtime.
   * 
   * For example:
   * `mapScalar("Date", "com.example.Date")`
   */
  fun mapScalar(graphQLName: String, targetName: String)

  /**
   * Map a GraphQL scalar type to the Java/Kotlin type and provided adapter expression.
   *
   * For example:
   * `mapScalar("Date", "com.example.Date", "com.example.DateAdapter")` (an instance property or object)
   * `mapScalar("Date", "com.example.Date", "com.example.DateAdapter()")` (instantiate the class on the fly)
   */
  fun mapScalar(graphQLName: String, targetName: String, adatpterExpression: String)

```

BoD: `mapScalarToKotlinMap` / `mapScalarToJavaMap` should be `mapScalarToKotlinAny` / `mapScalarToJavaObject` or am I missing something?

martinbonnin: Yes, indeed. Until we have https://github.com/apollographql/apollo-kotlin/issues/3243, we can't use Map or other parametrized classes.

BoD:
I've pushed the Kotlin implementation for review

Remaining:
- [ ] Java
- [ ] Documentation updates

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -170,6 +170,11 @@ object ApolloCompiler {
       options.operationOutputFile.writeText(operationOutput.toJson("  "))
     }
 
+    // Merge scalar mappings from upstream modules and this module's options
+    val allScalarMapping = options.incomingCompilerMetadata.fold(options.scalarMapping.toMutableMap()) { acc, it ->
+      acc.apply { putAll(it.scalarMapping) }
```
martinbonnin, line 175: This feels dangerous. What if there's a conflicting mapping in 2 modules? Should we mandate that all scalar mappings are defined in the schema module? 

I can't think of any use case where different modules could use different mappings but if something comes up we can always relax the restriction.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -267,3 +270,26 @@ class Options(
   }
 }
 
+/**
+ * Controls how scalar adapters are used in the generated code.
+ */
+@JsonClass(generateAdapter = true, generator = "sealed:type")
+sealed interface AdapterInitializer : Serializable
```
martinbonnin, line 277: Why does it need to be `Serializable` ?

martinbonnin: Got it. This is need by Gradle to snapshot the input properties. 

I always feel weird implementing `Serializable` since it's already serialized to Json. Also using Kotlin features like sealed interfaces and data classes in Gradle APIs feels dangerous too (think Groovy callers).

I would certainly have used 2 maps:

```kotlin
scalarsMapping: Map<String, String>
scalarAdapters: Map<String, String>
```

But I can't find a strong reason for the change so your choice!

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
```diff
@@ -132,8 +133,8 @@ internal class IrBuilder(
       visitedTypes.add(name)
       val typeDefinition = schema.typeDefinition(name)
       if (typeDefinition.isBuiltIn()) {
-        // We don't generate builtin types
-        continue
+        // We don't generate builtin types, unless they are explicitly mapped
```
martinbonnin, line 136:
```suggestion
        // We don't generate builtin types, unless they are explicitly mapped
        // Because users need String.type (or others) to register their adapters
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -95,17 +99,62 @@ interface Service {
...
   @ApolloDeprecatedSince(v3_0_0)
   val customTypeMapping: MapProperty<String, String>
 
+  /**
+   * Map from GraphQL scalar types to the Java/Kotlin type and adapter.
+   * Do not use this property directly, instead use [mapScalar].
+   *
+   * Default value: the empty map
+   */
+  val scalarMapping: MapProperty<String, ScalarInfo>
```
martinbonnin, line 121: Can we hide this from the API?

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -455,7 +462,14 @@ abstract class DefaultApolloExtension(
...
       task.failOnWarnings.set(service.failOnWarnings)
       @Suppress("DEPRECATION")
-      task.customScalarsMapping.set(service.customScalarsMapping.orElse(service.customTypeMapping))
+      task.scalarsMapping.set(
+          service.scalarMapping.orElse(
+              service.customScalarsMapping.orElse(
+                  service.customTypeMapping
+              )
+                  .getOrNull()?.asScalarInfoMapping() ?: emptyMap()
+          )
```
martinbonnin, line 471: throw if both `customScalarsMapping` and `sclarMapping` are set?

tests/custom-scalars/src/test/kotlin/test/CustomScalarTest.kt
```diff
@@ -34,14 +45,23 @@ class CustomScalarTest {
         .query(GetAllQuery())
         .execute()
         .dataAssertNoErrors
-
+    assertEquals(1, data.id)
```
martinbonnin, line 48: Wait, how come that works? Isn't this comparing an `Int` with a `Long`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinResolver.kt
```diff
@@ -60,26 +63,35 @@ class KotlinResolver(entries: List<ResolverEntry>, val next: KotlinResolver?) {
...
+      type is IrScalarType -> resolveIrScalarType(type)
       type is IrModelType -> resolveAndAssert(ResolverKeyKind.Model, type.path)
       type is IrNamedType -> resolveAndAssert(ResolverKeyKind.SchemaType, type.name)
       else -> error("$type is not a schema type")
     }.copy(nullable = true)
   }
 
+  private fun resolveIrScalarType(type: IrScalarType): ClassName {
+    // Try mapping first, then built-ins, then fallback to Any
+    return resolve(ResolverKeyKind.CustomScalarTarget, type.name) ?: when (type.name) {
```
martinbonnin, line 75: Rename `CustomScalarTarget` to `ScalarTarget`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
martinbonnin: Thinking a bit more about this. I think it makes sense to keep the builtin types in `apollo-api` always and not generate them. This means for builtin types that require a runtime adapter, users need to do something like:

```kotlin
apolloClientBuilder.addScalarAdapter(CompiledIDType, MyIDAdapter)
// instead of
// apolloClientBuilder.addScalarAdapter(ID.type, MyIDAdapter)
```

But I think this is such an advanced case that it's ok? It keeps the logic more symmetric. Builting types are in `apollo-api`, others get generated

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinResolver.kt
```diff
@@ -156,6 +159,43 @@ class KotlinResolver(entries: List<ResolverEntry>, val next: KotlinResolver?) {
...
+            resolveCompiledType(type.name)
+        )
+      }
+      else -> {
+        when (type.name) {
+          "Boolean" -> CodeBlock.of("%M", KotlinSymbols.BooleanAdapter)
+          "ID" -> CodeBlock.of("%M", KotlinSymbols.StringAdapter)
+          "String" -> CodeBlock.of("%M", KotlinSymbols.StringAdapter)
+          "Int" -> CodeBlock.of("%M", KotlinSymbols.IntAdapter)
+          "Float" -> CodeBlock.of("%M", KotlinSymbols.DoubleAdapter)
```
martinbonnin, line 181: I'd love this to be moved up a few layers so that the resolver can apply the same logic everywhere. Not really sure how to do this yet but I might investigate this in a follow up PR unless you have ideas how to do this already.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
BoD: I thought it was already the case actually. The multi module [doc](https://www.apollographql.com/docs/kotlin/advanced/multi-modules/) says "The schema module and only the schema module must define customScalarsMapping".

Should I ensure there are no duplicates and throw if there are any? Or should we not merge them at all (not sure how that would work)?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
BoD: Yes sorry I should have commented :). I agree, I'm not very fond of making them as `Serializable` either. Using `String` feels safer, I'll make the change.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
BoD: 👍  I'll have a look!

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
BoD: Good point!

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
martinbonnin: Better throw fast if we see something like this.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
martinbonnin: In retrospect, `CompiledIDType` should certainly have been `BuiltinIDType` but that'll be for v4 😅

tests/custom-scalars/src/test/kotlin/test/CustomScalarTest.kt
BoD: `assertEquals` uses `.equals` under the hood

martinbonnin: Yes but...

Screenshot 2022-01-17 at 09 54 16

BoD: 🤔  Indeed!

Looks like a generics thing, but to be honest I'm surprised!

```kotlin
fun <T> isEqual(a: T, b: T): Boolean {
  println(a!!::class)
  println(b!!::class)
  return a == b
}

isEqual(1, 1L)
```
```
class kotlin.Long
class kotlin.Long
true
```

martinbonnin: TIL, Kotlin has implicit conversions for number literals: https://discuss.kotlinlang.org/t/at-which-situations-does-implicit-conversion-happen/16394/2

(And for the verbose version: https://kotlinlang.org/spec/built-in-types-and-their-semantics.html#integer-type-widening)

All in all, maybe use `assertEquals(1L, data.id)` to make it explicit that we're also checking the `Long` type here ?

BoD: Very cool TIL too! And yes, totally agree :)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
BoD: Hmm after more digging, the `CustomScalarAdapters` methods take instances of `CustomScalarType`s, whereas the compiled types are `ScalarType` - so `apolloClientBuilder.addScalarAdapter(ID.type, MyIDAdapter)` for instance doesn't work.

We could make `CustomScalarType` be a subclass of `ScalarType`, and change `CustomScalarAdapters` methods to take `ScalarType` instead. But that looks like a breaking change 😅

Or we can add `ScalarType` based methods (`responseAdapterFor`, `add`, `addCustomScalarAdapter`) in addition to the existing `CustomScalarType` based ones.

WDYT?

martinbonnin: Gosh this is hard. I'm not too much a fan of duplicating everything either. I think your initial approach of generating a `CustomScalarType` for builtin types if they have a mapping was best then. `CustomScalarType` will not be a super accurate name and the whole thing feels a bit asymmetrical but I don't see a better way.

Or should we always generate all builtin types? And mark `CompiledStringType`, etc.. as deprecated? I was initially not too fond of this because it puts `String`, `Int`, etc.. as top level classes but maybe since they're using a specific package name, that'll be ok...

BoD: > Or should we always generate all builtin types?

I'll five it a go so we can see what it looks like :)

martinbonnin (approved):

martinbonnin (approved):

Issue #3761, opened 2022-01-04
Add compat helper for STORE_PARTIAL_RESPONSES


martinbonnin:
Put back the 2.x `STORE_PARTIAL_RESPONSES` as `DeprecationLevel.Error` that points to the `storePartialResponses()` extension method

BoD:
Fixed via #3784

PR #3784, opened 2022-01-11
Add STORE_PARTIAL_RESPONSES from 2.x, with a deprecated message

BoD:
Fix for #3761


apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/ApolloCacheHeaders.kt
```diff
@@ -13,4 +16,14 @@ object ApolloCacheHeaders {
...
+
+  /**
+   * To configure whether to store partial responses, please use the storePartialResponses() extension instead.
+   */
+  @Deprecated(
+      message = "Used for backward compatibility with 2.x",
+      level = DeprecationLevel.ERROR
+  )
+  @ApolloDeprecatedSince(Version.v3_0_1)
+  const val STORE_PARTIAL_RESPONSES: String = ""
```
BoD, line 28: Wanted to add a getter that throws at first, but that would make it less obvious for Java users, as it compiles as a method.

martinbonnin (approved):

PR #3793, opened 2022-01-18
Mark compiled types as deprecated and no longer use them in generated code

BoD:
Related to [this comment](https://github.com/apollographql/apollo-kotlin/pull/3779#discussion_r786311491) on the feature branch, this removes usage of the compiled types altogether.

Caveat: when `generateSchema` is set to `true` we generate a `__Schema` object which now clashes with the generated `__Schema` type. So I renamed to `___Schema` but this is a breaking change and the name is a bit cumbersome. Maybe we could keep the name but generate it in another package (still breaking change but maybe not as bad)? Not sure!

Let me know what you think!


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -239,46 +241,68 @@ fun resolveVariables(value: Any?, variables: Executable.Variables): Any? {
   }
 }
 
+@Deprecated("No longer used by the codegen")
```
martinbonnin, line 244:
```suggestion
@Deprecated("Use the generated CustomScalarType instead")
```

martinbonnin (changes requested):
In order to avoid breaking changes, we should keep `__Schema` and alllow the user to override the name if required.

```kotlin
apollo {
  generatedSchemaName.set("ApolloSchema")
}
```

martinbonnin (approved):

PR #3803, opened 2022-01-21
Fix disposing the WebSocketNetworkTransport if no subscription has been executed

martinbonnin:
The `Dispose` message wasn't handled in that case.

Fixes https://github.com/apollographql/apollo-kotlin/issues/3802


BoD (approved):
👍

BoD (approved):

PR #3811, opened 2022-01-24
Tweaks to the scalar adapters PR

martinbonnin:
- rename custom-scalar test to scalar-adapters
- fix KDoc for registerOperations
- add a section in CONTRIBUTING.md about naming Gradle APIs
- Kdoc tweaks
- add a test to make sure we don't allow mapScalar + customScalarMapping at the same time
- remove ScalarTarget as a resolver entry and instead use scalarMapping

I also tried moving all the builtin logic & adapters to a `ExpressionAdapterInitializer` but:
1. Since these are used **a lot**, we'll need to keep the imports from javapoet/kotlinpoet or the code will look ugly so we can't just output an expression as a literal.
2. There is an optimization right now to have the nullable versions of the adapters pre-computed so that'd need to be part of the API too somehow.


BoD (approved):
Very nice simplifications! 🙏

Didn't grasp the distinction between `CommonMetadata` and `CompilerMetadata`, now I do :)

PR #3813, opened 2022-01-25
Tunnel variables in CustomScalarAdapters

### LINKED ISSUE #3810: Issues with `@include` / `@skip` on fragment spreads / inline fragments
**Summary**
Fragment spreads / inline fragments with `@include(if: $variable)` are not parsed even when `$variable` is `true`.

**Version**
3.0.0

**Description**

When using `@include(if: true)` or `@include(if: false)` on a fragment spread or inline fragment, for instance:

```graphql
query LaunchList($cursor: String) {
  launches(after: $cursor) {
    cursor
    hasMore
    launches {
      ...launchFields @include(if: true)
    }
  }
}

fragment launchFields on Launch {
  id
  site
  mission {
    name
    missionPatch(size: SMALL)
  }
}
```

The generated code looks like:

```kotlin
// ...

reader.rewind()
var launchFields: LaunchFields? = null
if (BooleanExpression.True.evaluate(emptySet(), __typename)) {
  launchFields = com.example.rocketreserver.fragment.LaunchFieldsImpl_ResponseAdapter.LaunchFields.fromJson(reader,
      customScalarAdapters)
}

```

`BooleanExpression.True.evaluate(emptySet(), __typename)` evaluates to `true` so the fields of the `launchFields` fragment are correctly parsed (and inversely for `False`) ✅

However when using a variable:

```graphql
query LaunchList($cursor: String, $includeLaunchFields: Boolean!) {
  launches(after: $cursor) {
    cursor
    hasMore
    launches {
      ...launchFields @include(if: $includeLaunchFields)
    }
  }
}
```

The generated code looks like this:

```graphql
if (variable("includeLaunchFields").evaluate(emptySet(), __typename)) {
  launchFields = com.example.rocketreserver.fragment.LaunchFieldsImpl_ResponseAdapter.LaunchFields.fromJson(reader,
      customScalarAdapters)
}
```

`variable("includeLaunchFields").evaluate(emptySet(), __typename)` always return `false` (no matter the value of `$includeLaunchFields` ) so the fields are never parsed which is incorrect when `$includeLaunchFields` is `true`   ❌

### Possible fix

The generated adapter has no knowledge of the variables.

Therefore, it shouldn’t assume anything and always “try” to parse.

To do this, the code generated to evaluate a `BooleanExpression` must handle the `BVariable` case in such a way that calling `evaluate` on it returns true. One simple way to do this is to return  `BooleanExpression.True`.

However, making this change will now try to parse the fields even when they are not returned (the case when `$includeLaunchFields` is `false`). This will crash for any non nullable field in the fragment.

A possible (but not very satisfactory) fix is to add a `try catch` block and return `null`  if parsing fails.

In other words, the generated code could look like this:

```graphql
if (BooleanExpression.True.evaluate(emptySet(), __typename)) {
  launchFields = try { com.example.rocketreserver.fragment.LaunchFieldsImpl_ResponseAdapter.LaunchFields.fromJson(reader,
      customScalarAdapters) } catch (_: Exception) { null }
}
```

Of course having a `try catch` in the generated code is not very compelling... Could there be another solution I'm not seeing?

martinbonnin:
Make operation variables available in `CustomScalarAdapters`. This is not the cleanest thing but it's an easy and non-breaking way to make the variables accessible from `Adapters`. Moving forward, we should separate two interfaces:

* `ScalarAdapter` can be implemented by users for scalars. Doesn't need access to variables or other scalar adapters.
* `Adapter` would be mostly internal to be used by codegen for `Operation.adapter()`. 

Close #3810


apollo-api/api/apollo-api.api
```diff
@@ -328,14 +328,15 @@ public final class com/apollographql/apollo3/api/CompiledVariable {
 public final class com/apollographql/apollo3/api/CustomScalarAdapters : com/apollographql/apollo3/api/ExecutionContext$Element {
 	public static final field Empty Lcom/apollographql/apollo3/api/CustomScalarAdapters;
 	public static final field Key Lcom/apollographql/apollo3/api/CustomScalarAdapters$Key;
-	public synthetic fun <init> (Ljava/util/Map;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Ljava/util/Map;Lcom/apollographql/apollo3/api/Executable$Variables;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
```
martinbonnin, line 331: This is the same issue as https://github.com/Kotlin/binary-compatibility-validator/issues/73.

This constructor is private so the breaking change should be ok here

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CustomScalarAdapters.kt
```diff
@@ -53,6 +53,18 @@ private constructor(
     } as Adapter<T>
   }
 
+  fun variables(): Set<String> {
+    if (variables == null) {
+      return emptySet()
+    }
+
+    return variables.valueMap.entries.filter {
+      it.value == true
```
BoD, line 62: Could be a bit tricky to understand, should the name for `variables()` be more descriptive? Maybe `trueVariableNames()` (not great...) `variablesThatAreTrue()` ?  Or maybe return the Map as-is and filter in `Operations.kt` (or even in `AdapterCommon`) instead?

tests/include-skip-operation-based/src/test/kotlin/IncludeTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import kotlin.test.assertEquals
+
+class IncludeTest {
+  @Test
+  fun includeTrue() = runBlocking {
+    val operation = GetCatQuery(withCat = true)
+
+    val dataMap = GetCatQuery_TestBuilder.DataBuilder().apply {
+      animal = catAnimal {
+        meow = "meeoooowwwww"
```
BoD, line 18: 😂

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/AdapterCommon.kt
```diff
@@ -107,7 +107,7 @@ internal fun readFromResponseCodeBlock(
                 context.layout.variableName(property.info.responseName),
                 context.resolver.resolveIrType(property.info.type).copy(nullable = !property.info.type.isOptional()),
             )
-            beginControlFlow("if·(%L.%M(emptySet(),·$__typename))", property.condition.codeBlock(), evaluate)
+            beginControlFlow("if·(%L.%M($customScalarAdapters.variables(),·$__typename))", property.condition.codeBlock(), evaluate)
```
BoD, line 110: How come this doesn't impact a bunch of `.exepected` files?

BoD: Ah I guess it does actually 😅

martinbonnin: Yup, my bad, forgot to regenerate those, doing it now.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CustomScalarAdapters.kt
```diff
@@ -9,9 +9,9 @@ import kotlin.jvm.JvmField
 /**
  * A wrapper around a Map<String, [Adapter]> used to retrieve custom scalar adapters at runtime
  */
-class CustomScalarAdapters
-private constructor(
+class CustomScalarAdapters private constructor(
     customScalarAdapters: Map<String, Adapter<*>>,
+    private val variables: Executable.Variables?
```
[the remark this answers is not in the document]
martinbonnin: Agreed 👍 . PRs in code is fine by me 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CustomScalarAdapters.kt
martinbonnin: In the world were `CustomScalarAdapters` becomes `AdapterContext`, I don't want to expose all variables because only boolean ones are actually needed during parsing. 

I went with plain `variables` because it's also the terminology used in `evaluate`: https://github.com/apollographql/apollo-kotlin/blob/72630f31ad8082c285e3a50e207345776ff86d77/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt#L112

Agree it's not great but don't really have good alternatives. `variablesSet`? `definedVariables`?

BoD: I'm ok with `variables` 👍

Nitpick: I think it can be simplified as 
```kotlin
    return variables.valueMap.filter {
      it.value == true
    }.keys
```

BoD (approved):
👍

PR #3814, opened 2022-01-25
Fix @skip with if=true/false not being inverted

BoD:
A follow-up of #3813

Skip is the opposite of include so `toBooleanExpression` should return the opposite of what its `if` is set to (if that makes sense!), which was correctly done for a variable, but not for a hardcoded `true`/`false`.


martinbonnin (approved):
👍

Issue #3786, opened 2022-01-11
Allow to specify error handling for watch()


martinbonnin:
Currently, `watch()` ignores all errors (not only cache) which is different from 2.x. 

I'm not sure what the different use cases are but getting an error for the initial request seems like an important thing. To allow for maximum configurability, what about adding a flag to watch:

```kotlin
enum class WatchErrorHandling {
  EMIT_CACHE_ERRORS,
  EMIT_NETWORK_ERRORS,
  EMIT_CACHE_AND_NETWORK_ERRORS,
  IGNORE_ERRORS
}

fun ApolloQueryCall<D>.watch(errorHandling: WatchErrorHandling)
```

This can be complexified that we have both `fetchPolicy` and `refetchPolicy` at the moment so we might want to have separate error handling there but a possible solution could be to leverage coroutines instead:

```kotlin
// The initial query will ignore cache errors and rethrow other errors
apolloClient.query(query).toFlow()
 .catch { if (it !is CacheMissException) throw it}
 .onCompletion {
  emitAll(apolloClient.query.watch())
}
```

peterplamenovpenchev:
Hey all, we are currently migrating to Apollo 3.0 & it is a lot of fun.
We stumbled upon that discrepancy as well.

Is there any visibility when will this be available? I guess in 3.0.1 ?

BoD:
In the end we went for a simpler API: `watch(fetchThrows: Boolean = false, refetchThrows: Boolean = false)` and also added a lower level `watch(data: D?, retryWhen: suspend (cause: Throwable, attempt: Long) -> Boolean = { _, _ -> true })` API that can be used in more advanced scenarios.

This will be in the next version - in the meantime can be tried out with the [snapshots](https://github.com/apollographql/apollo-kotlin#snapshots).

martinbonnin:
Fixed in 3.1.0

PR #3817, opened 2022-01-26
Allow to specify error handling for watch()

BoD:
Related to #3786.

Introduce `WatchErrorHandling` enum and use it in `WatcherInterceptor`.

### DIFF
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -95,12 +95,21 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
       .writeToCacheAsynchronously(writeToCacheAsynchronously)
 }
 
+enum class WatchErrorHandling {
+  EMIT_CACHE_ERRORS,
+  EMIT_NETWORK_ERRORS,
+  EMIT_CACHE_AND_NETWORK_ERRORS,
+  IGNORE_ERRORS
+}
+
 /***
  * Gets the result from the network, then observes the cache for any changes.
  * Overriding the [FetchPolicy] will change how the result is first queried.
+ * Exception are ignored by default, this can be changed by setting [watchErrorHandling].
  */
-fun <D : Query.Data> ApolloCall<D>.watch(): Flow<ApolloResponse<D>> {
-  return copy().addExecutionContext(WatchContext(true)).toFlow()
+@JvmOverloads
+fun <D : Query.Data> ApolloCall<D>.watch(errorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS): Flow<ApolloResponse<D>> {
+  return copy().addExecutionContext(WatchContext(errorHandling)).toFlow()
 }
 
 /**
@@ -261,8 +270,8 @@ internal val <D : Mutation.Data> ApolloRequest<D>.optimisticData
 internal val <D : Operation.Data> ApolloRequest<D>.cacheHeaders
   get() = executionContext[CacheHeadersContext]?.value ?: CacheHeaders.NONE
 
-internal val <D : Operation.Data> ApolloRequest<D>.watch
-  get() = executionContext[WatchContext]?.value ?: false
+internal val <D : Operation.Data> ApolloRequest<D>.watchContext: WatchContext?
+  get() = executionContext[WatchContext]
 
 /**
  * @param isCacheHit true if this was a cache hit
@@ -441,7 +450,7 @@ internal class OptimisticUpdatesContext<D : Mutation.Data>(val value: D) : Execu
   companion object Key : ExecutionContext.Key<OptimisticUpdatesContext<*>>
 }
 
-internal class WatchContext(val value: Boolean) : ExecutionContext.Element {
+internal class WatchContext(val errorHandling: WatchErrorHandling) : ExecutionContext.Element {
   override val key: ExecutionContext.Key<*>
     get() = Key
 
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -125,7 +125,8 @@ fun <D : Query.Data> ApolloCall<D>.watch(
         data = it.data
       }.onCompletion {
         emitAll(
-            watch(data)
+            copy().fetchPolicyInterceptor(refetchPolicyInterceptor)
+                .watch(data)
                 .catch {
                   maybeThrow(it, refetchErrorHandling)
                 }
@@ -135,7 +136,7 @@ fun <D : Query.Data> ApolloCall<D>.watch(
 
 /**
  * Observes the cache for the given data. Unlike [watch], no initial request is executed on the network.
- * The refetch policy set by [refetchPolicy] will be used.
+ * The fetch policy set by [fetchPolicy] will be used.
  * Cache and network exceptions are propagated.
  */
 fun <D : Query.Data> ApolloCall<D>.watch(data: D?): Flow<ApolloResponse<D>> {
@@ -316,7 +317,7 @@ fun <D : Mutation.Data> ApolloCall<D>.optimisticUpdates(data: D) = addExecutionC
 internal val <D : Operation.Data> ApolloRequest<D>.fetchPolicyInterceptor
   get() = executionContext[FetchPolicyContext]?.interceptor ?: CacheFirstInterceptor
 
-internal val <D : Operation.Data> ApolloRequest<D>.refetchPolicyInterceptor
+private val <T> MutableExecutionOptions<T>.refetchPolicyInterceptor
   get() = executionContext[RefetchPolicyContext]?.interceptor ?: CacheOnlyInterceptor
 
 internal val <D : Operation.Data> ApolloRequest<D>.doNotStore
@@ -528,25 +529,10 @@ internal class FetchFromCacheContext(val value: Boolean) : ExecutionContext.Elem
   companion object Key : ExecutionContext.Key<FetchFromCacheContext>
 }
 
-internal class IsRefetching(val value: Boolean) : ExecutionContext.Element {
-  override val key: ExecutionContext.Key<*>
-    get() = Key
-
-  companion object Key : ExecutionContext.Key<IsRefetching>
-}
-
 internal fun <D : Operation.Data> ApolloRequest.Builder<D>.fetchFromCache(fetchFromCache: Boolean) = apply {
   addExecutionContext(FetchFromCacheContext(fetchFromCache))
 }
 
-internal fun <D : Operation.Data> ApolloRequest.Builder<D>.isRefetching(isRefetching: Boolean) = apply {
-  addExecutionContext(IsRefetching(isRefetching))
-}
-
 internal val <D : Operation.Data> ApolloRequest<D>.fetchFromCache
   get() = executionContext[FetchFromCacheContext]?.value ?: false
-
-
-internal val <D : Operation.Data> ApolloRequest<D>.isRefetching
-  get() = executionContext[IsRefetching]?.value ?: false
 
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -105,11 +105,18 @@ enum class WatchErrorHandling {
 /***
  * Gets the result from the network, then observes the cache for any changes.
  * Overriding the [FetchPolicy] will change how the result is first queried.
- * Exception are ignored by default, this can be changed by setting [watchErrorHandling].
+ * Exception are ignored by default, this can be changed by setting [fetchErrorHandling] for the first fetch and [refetchErrorHandling]
+ * for subsequent fetches.
  */
 @JvmOverloads
-fun <D : Query.Data> ApolloCall<D>.watch(errorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS): Flow<ApolloResponse<D>> {
-  return copy().addExecutionContext(WatchContext(errorHandling)).toFlow()
+fun <D : Query.Data> ApolloCall<D>.watch(
+    fetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
+    refetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
+): Flow<ApolloResponse<D>> {
+  return copy().addExecutionContext(WatchContext(
+      fetchErrorHandling = fetchErrorHandling,
+      refetchErrorHandling = refetchErrorHandling,
+  )).toFlow()
 }
 
 /**
@@ -450,7 +457,10 @@ internal class OptimisticUpdatesContext<D : Mutation.Data>(val value: D) : Execu
   companion object Key : ExecutionContext.Key<OptimisticUpdatesContext<*>>
 }
 
-internal class WatchContext(val errorHandling: WatchErrorHandling) : ExecutionContext.Element {
+internal class WatchContext(
+    val fetchErrorHandling: WatchErrorHandling,
+    val refetchErrorHandling: WatchErrorHandling,
+) : ExecutionContext.Element {
   override val key: ExecutionContext.Key<*>
     get() = Key
 
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -31,6 +31,7 @@ import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.retryWhen
 import kotlin.jvm.JvmName
 import kotlin.jvm.JvmOverloads
 
@@ -124,15 +125,19 @@ fun <D : Query.Data> ApolloCall<D>.watch(
   var data: D? = null
   return toFlow()
       .catch {
-        maybeThrow(it, fetchErrorHandling)
+        if (it.shouldThrow(fetchErrorHandling)) throw it
       }.onEach {
         data = it.data
       }.onCompletion {
         emitAll(
             copy().fetchPolicyInterceptor(refetchPolicyInterceptor)
                 .watch(data)
+                .retryWhen { cause, _ ->
+                  // If this exception is ignored (shouldThrow is false), we should continue watching - so retry
+                  !cause.shouldThrow(refetchErrorHandling)
+                }
                 .catch {
-                  maybeThrow(it, refetchErrorHandling)
+                  if (it.shouldThrow(refetchErrorHandling)) throw it
                 }
         )
       }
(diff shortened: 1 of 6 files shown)


martinbonnin:
I just restarted the CI that got stuck on `integration-tests`

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -95,14 +99,83 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
  */
-fun <D : Query.Data> ApolloCall<D>.watch(): Flow<ApolloResponse<D>> {
-  return copy().addExecutionContext(WatchContext(true)).toFlow()
+fun <D : Query.Data> ApolloCall<D>.watch(data: D?): Flow<ApolloResponse<D>> {
+  return copy().addExecutionContext(WatchContext(data)).toFlow()
 }
 
+private fun maybeThrow(throwable: Throwable, errorHandling: WatchErrorHandling) {
+  // Only potentially swallow Apollo exceptions, the rest must be surfaced
+  if (throwable !is ApolloException) {
```
martinbonnin, line 147: Is that required? `Flow.catch` [rethrows the cancellation exceptions](https://github.com/Kotlin/kotlinx.coroutines/blob/ae0c842d9f5d8e4718d54960325c260dc551426e/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt#L169) so we should only get our own exceptions there?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -95,14 +99,83 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
+    throw throwable
+  }
+  if (errorHandling == WatchErrorHandling.IGNORE_ERRORS) return
+  val throwCacheErrors = errorHandling == WatchErrorHandling.THROW_CACHE_AND_NETWORK_ERRORS || errorHandling == WatchErrorHandling.THROW_CACHE_ERRORS
+  val throwNetworkErrors = errorHandling == WatchErrorHandling.THROW_CACHE_AND_NETWORK_ERRORS || errorHandling == WatchErrorHandling.THROW_NETWORK_ERRORS
```
martinbonnin, line 152: Makes me want a bitfield API but this is Kotlin and not C so I guess this is fine too :)

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -95,14 +99,83 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
+
+/**
+ * Observes the cache for the given data. Unlike [watch], no initial request is executed on the network.
+ * The refetch policy set by [refetchPolicy] will be used.
+ * Cache and network exceptions are propagated.
  */
-fun <D : Query.Data> ApolloCall<D>.watch(): Flow<ApolloResponse<D>> {
-  return copy().addExecutionContext(WatchContext(true)).toFlow()
+fun <D : Query.Data> ApolloCall<D>.watch(data: D?): Flow<ApolloResponse<D>> {
+  return copy().addExecutionContext(WatchContext(data)).toFlow()
```
martinbonnin, line 142: Food for thoughts (no action required): instead of adding an execution context here, we could conditionally add the `WatcherInterceptor` so that it's only present in the chain if `watch()` is called.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: I was thinking other "unexpected" exceptions (OOM?) should not be swallowed here? But maybe I'm overthinking it :)

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: Haha yeah :) Java has `EnumSet` but they don't seem to be very popular in Kotlin

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/WatcherInterceptor.kt
```diff
@@ -8,59 +8,40 @@ import com.apollographql.apollo3.api.Query
...
         }.map {
-          chain.proceed(request.newBuilder().isRefetching(isRefetching).build())
-              .catch {
-                if (it !is ApolloException) {
-                  // Re-throw cancellation exceptions
-                  throw it
-                }
-                // Else just ignore errors
-              }
+          chain.proceed(request.newBuilder().isRefetching(true).build())
```
martinbonnin, line 39: I was thinking we could even remove `isRefetching` completely so that `refetchPolicy` is only used by the `watch(fetchError, refetchError)` function? "refetching" would be a concept only known by this one and have no effect else? Keep things more local?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: I have scars from trying to shrink a jar that was heavily using `EnumSet` because it uses reflection somewhere (can't remember where...)

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: Good point about OOM 👍 . Can you add a doc to the enum?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/WatcherInterceptor.kt
BoD: The thing is currently, `isRefetching` is used to route to either `fetchPolicyInterceptor` or `refetchPolicyInterceptor`. And since users can set their own  `refetchPolicyInterceptor` we want to keep this mechanism? Or was this never released so we can remove it?

martinbonnin: Also, passing `fetchError` and `refetchError` kind of breaks the usual pattern of how `ApolloCalls` are configured. Maybe this instead?

```kotlin
apolloClient.query(query)
             .fetchPolicy(CacheFirst)
             .refetchPolicy(CacheOnly)
             .errorHandling(THROW_CACHE_AND_NETWORK_ERRORS)
             .refetchErrorHandling(IGNORE_ERROR)
             .watch()
``
`

2 advantages: 

1. It's more consistent
2. It allows to clone the call and keep all parameters (not sure if that is ever going to be used but maybe? )

Comment (martinbonnin) [reply apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/WatcherInterceptor.kt:39]: (had another comment but deleted it to keep this conversation focused)

We definitely want to keep the behaviour but the choice can be made by `watch(errorHandling, refetchErrorHandling)`:

```kotlin
fun <D : Query.Data> ApolloCall<D>.watch(
    fetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
    refetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
): Flow<ApolloResponse<D>> {
  var data: D? = null
  return toFlow()
      .catch {
        maybeThrow(it, fetchErrorHandling)
      }.onEach {
        data = it.data
      }.onCompletion {
        emitAll(
            copy().fetchPolicyInterceptor(refetchPolicyInterceptor).watch(data)
                .catch {
                  maybeThrow(it, refetchErrorHandling)
                }
        )
      }
}
```

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -95,14 +99,83 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
 /***
  * Gets the result from the network, then observes the cache for any changes.
  * Overriding the [FetchPolicy] will change how the result is first queried.
+ * Exception are ignored by default, this can be changed by setting [fetchErrorHandling] for the first fetch and [refetchErrorHandling]
+ * for subsequent fetches.
+ */
+@JvmOverloads
+fun <D : Query.Data> ApolloCall<D>.watch(
+    fetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
+    refetchErrorHandling: WatchErrorHandling = WatchErrorHandling.IGNORE_ERRORS,
```
martinbonnin, line 118: Also, passing `fetchError` and `refetchError` kind of breaks the usual pattern of how `ApolloCalls` are configured. Maybe this instead?

```kotlin
apolloClient.query(query)
             .fetchPolicy(CacheFirst)
             .refetchPolicy(CacheOnly)
             .errorHandling(THROW_CACHE_AND_NETWORK_ERRORS)
             .refetchErrorHandling(IGNORE_ERROR)
             .watch()
```

2 advantages:

1. It's more consistent
2. It allows to clone the call and keep all parameters (not sure if that is ever going to be used but maybe? )

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/WatcherInterceptor.kt
BoD: Oooh! Indeed, didn't think of that :). I'll make the change 👍

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: Agreed! 👍

I'll also make them CamelCase (we use both conventions in the codebase, but there's `FetchPolicy` right above it which uses CamelCase so it will be at least consistent in the file :)

martinbonnin: Yay for CamelCase 🙌 

Also maybe `ThrowAll` instead of the verbose `ThrowCacheAndNetworkErrors` ?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -95,14 +99,83 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
+      }
+      val cacheMissException = throwable.first as? CacheMissException ?: throwable.second as? CacheMissException
+      // If it's *not* a CacheMissException we consider it a network error (could be ApolloNetworkException, ApolloHttpException, ApolloParseException...)
+      val networkException = throwable.first.takeIf { it !is CacheMissException } ?: throwable.second.takeIf { it !is CacheMissException }
+      if (cacheMissException != null && throwCacheErrors) {
+        throw cacheMissException
+      }
+      if (networkException != null && throwNetworkErrors) {
+        throw networkException
+      }
```
martinbonnin, line 166: I'm wondering if the `ApolloCompositeException` case is really required here. Could we instead consider `ApolloCompositeException` a network error? At the end of the day,  when `ApolloCompositeException` is thrown, it always means there was a network error (and maybe a cache error too). It keeps things simple. If anyone wants more fined grained error handling, they can always use the low-level API?

tests/integration-tests/src/commonTest/kotlin/test/WatcherErrorHandlingTest.kt
```diff
@@ -0,0 +1,274 @@
...
+      apolloClient.query(EpisodeHeroNameQuery(Episode.EMPIRE))
+          .fetchPolicy(FetchPolicy.NetworkOnly)
+          .watch()
+          .collect {
+            channel.send(it.data)
+          }
+    }
+
+    channel.assertEmpty()
+    jobs.forEach { it.cancel() }
```
martinbonnin, line 93: Doesn't this cancels before anything could happen? I would expect this test to throw an uncaught exception in the MockServer thread because to response is enqueued.

tests/integration-tests/src/commonTest/kotlin/test/WatcherErrorHandlingTest.kt
```diff
@@ -0,0 +1,274 @@
...
+          .watch()
+          .collect {
+            channel.send(it.data)
+          }
+    }
+    assertEquals(channel.receiveOrTimeout()?.hero?.name, "R2-D2")
+
+    // Another newer call gets updated information with "Artoo"
+    // Due to .refetchPolicy(FetchPolicy.NetworkOnly), a subsequent call will be executed in watch()
+    // we didn't enqueue anything in the mockServer so a network exception is thrown, and ignored by default
```
martinbonnin, line 117: Same here I think? If you want a network exception, enqueue an Error response else I would expect the MockServer to throw?

tests/integration-tests/src/commonTest/kotlin/test/WatcherErrorHandlingTest.kt
martinbonnin: I missed that `assertEmpty()` had a timeout. Then I guess it's ok. I still can't explain why the mockserver doesn't throw though?

BoD: I changed the behavior when working on MockServer, now it handles the case by [returning 500](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/apollo-mockserver/src/jvmMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt#L45) instead of crashing.

martinbonnin: Ah, I see. But [native is still crashing](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/QueueMockServerHandler.kt#L15).

Overall, I like the behaviour of crashing better. IIRC, this is what OkHttp MockServer does so it might be less surprising. Also it's more explicit?

BoD: For native it's handled [here](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/Socket.kt#L137) actually! ([here](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt#L37) for JS) There's also a test [here](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/EnqueueTest.kt#L52).

Don't remember precisely, but I think the crash was kind of hard to handle consistently between platforms (I think it was JS hanging forever?). To me a `500` is more realistic as in, a real server would not crash your tests. But I also agree that crashing is more explicit/obvious while crafting the tests. Should I revert?

martinbonnin: > For native it's handled here actually!

👍 Sorry I missed that.

> JS hanging forever

Urg, that's no good...

> a real server would not crash your tests

That's the thing, MockServer is not a real server. By returning 500 if nothing is enqueued, we're letting tests rely on that behaviour and the GraphQL calls are not balanced with the enqueue() anymore. That might or might not make the tests harder to read.

> Should I revert? 

Your call :-). If you decide to keep the 500 default behaviour, this should be in the release notes somewhere.

BoD: All right I'll let the fate decide 😅 : if I can make it crash consistently I'll do that.

But in any case I agree with you that the test should be explicitly enquing an error if it expects it later, instead of implicitly relying on this behavior! I'll update that.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: Question: did you mean that `errorHandling` should be taken into account also outside of `watch`? If yes it may be slightly confusing as the default is to ignore in `watch` but throw all in `toFlow`/`execute` - but this can be clarified in the KDocs.

martinbonnin: > did you mean that errorHandling should be taken into account also outside of watch

I'm not sure... On one hand, it being a general thing makes it more consistent. On the other hand, it doesn't bring much over using `.catch {}` directly so it's adding noise in the API. I think I'd limit this to `watch()` for the time being?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: Let's keep it as parameters then? To highlight the fact that it's more of a special case?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -100,88 +99,46 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
-  IgnoreErrors,
-}
-
 /***
  * Gets the result from the network, then observes the cache for any changes.
  * Overriding the [FetchPolicy] will change how the result is first queried.
- * Exception are ignored by default, this can be changed by setting [fetchErrorHandling] for the first fetch and [refetchErrorHandling]
- * for subsequent fetches.
+ * Network and cache exceptions are ignored by default, this can be changed by setting [fetchThrows] for the first fetch and [refetchThrows]
+ * for subsequent fetches (non Apollo exceptions like `OutOfMemoryError` are always propagated).
```
martinbonnin, line 106: Nit pick. Use `@param`? It duplicates some content but from a consumer point of view, it's easier to find the information?

```kotlin
/**
 * Gets an initial result then observes the cache for any changes.
 * Use [fetchPolicy] to control the initial fetch and [refetchPolicy] for the subsequent ones.
 * 
 * Network and cache exceptions are ignored by default, this can be changed by setting [fetchThrows] for the first fetch and [refetchThrows]
 * for subsequent fetches (non Apollo exceptions like `OutOfMemoryError` are always propagated).
 *
 * @param fetchThrows whether to throw if an [ApolloException] happens during the initial fetch. Default: false
 * @param refetchThrows whether to throw if an [ApolloException] happens during a refetch. Default: false
 */
```

martinbonnin (approved):

PR #3822, opened 2022-01-28
Make MockServer crash when querying with no enqueued response

BoD:
I looked at making MockServer crash when querying with no enqueued response, instead of it returning a `500` response, following [this discussion](https://github.com/apollographql/apollo-kotlin/pull/3817#discussion_r794396672).

When that happens inside a test:
- on JVM:
  - MockServer crashes its own thread and a log is printed
  - the test itself continues
  - (depending on the test) the test will crash also after 1 minute (read timeout)
  - the test is thus marked as failed
- on Apple:
  - same, except
  - the test crash was not propagated to the test runner, and we hung forever
  - ⬆️  This PR addresses this (see `runWithMainLoopApple.kt`), so now the test is marked as failed as well
- on JS:
  - MockServer crashes and a log is printed
  - the test is immediately marked as failed at this point (I guess any crash stops the runner)

The behavior is not 100% consistent (in fact I had to remove the `status500WhenNothingWasEnqueued` test), but I think it's still clear enough to make MockServer crash like it was previously.

### DIFF
--- a/apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/Socket.kt
+++ b/apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/Socket.kt
@@ -134,7 +134,7 @@ class Socket(
           try {
             mockServerHandler.handle(request)
           } catch (e: Exception) {
-            MockResponse("MockServerHandler.handle() threw an exception: ${e.message}", 500)
+            throw Exception("MockServerHandler.handle() threw an exception: ${e.message}", e)
           }
         }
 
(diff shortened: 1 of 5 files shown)


apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/Socket.kt
```diff
@@ -134,7 +134,7 @@ class Socket(
           try {
             mockServerHandler.handle(request)
           } catch (e: Exception) {
-            MockResponse("MockServerHandler.handle() threw an exception: ${e.message}", 500)
+            throw Exception("MockServerHandler.handle() threw an exception: ${e.message}", e)
```
martinbonnin, line 137: Why not let the original exception bubble up?

martinbonnin (approved):

apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/Socket.kt
BoD: I was thinking it would make the message a bit more predictable/less dependent on a handler

martinbonnin: Fair enough 👍

PR #3825, opened 2022-01-30
Correct reference to addCustomScalarAdapter in 3.0 migration doc

dhritzkiv:
Correct reference to `addCustomScalarAdapter` for replacing `addCustomTypeAdapter` in "The quick route"


martinbonnin:
Thanks for sending this! I think the reasoning was that since this is the "fast" route, we'd keep the "old" name. `addCustomTypeAdapter` is still present (even if deprecated) so it's not strictly required to change that (even though it's more future proof). Or is there another reason to do the change?

dhritzkiv:
Nope, no reason – I misunderstood / got jumbled up with the naming and types. My mistake!

Issue #3763, opened 2022-01-04
Add support for CACHE_AND_NETWORK with watch()


demoritas:
**Use case**
following scenario, we have potentially data in cache but we need a network refresh and want to watch the cache after that:

cache first -> to have a fast response to the user
network second -> to have the current data in case it changed
watch -> we need cache updates

`executeCacheAndNetwork` is not useable in this scenario since it directly executes.

There is no `FetchPolicy` for this scenario so we can not call `watch`

**Describe the solution you'd like**

a.  `FetchPolicy.CacheAndNetwork` so we can decide to call `toFlow`, `execute` or `watch` after setting it
This has the downside that `execute()` has a definition problem sind we could have 2 results

b. add an extension function `watchCacheAndNetwork`

BoD:
With #3829 a new `CacheAndNetwork` policy has been added 🎉. This will be in the next release, in the meantime it can be tried out with the [snapshots](https://github.com/apollographql/apollo-kotlin#snapshots).

martinbonnin:
Fixed in 3.1.0

PR #3829, opened 2022-01-31
CacheAndNetwork FetchPolicy

BoD:
This is another approach to #3763, to compare and choose the best one :).

Instead of having a specific `watchCacheAndNetwork()` method, a new `FetchPolicy.CacheAndNetwork` can be used with `watch()`.

(This one targets the branch of the first approach)

Thoughts: I thought I could replace the implementation of `executeCacheAndNetwork` with a call with the new policy, but the exception handling is different (current `executeCacheAndNetwork` throws only if both cache and network failed).

### DIFF
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -31,32 +31,36 @@ import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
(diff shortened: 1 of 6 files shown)


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -31,32 +31,36 @@ import kotlinx.coroutines.flow.emitAll
...
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.flow.onStart
 import kotlin.jvm.JvmName
 import kotlin.jvm.JvmOverloads
 
 enum class FetchPolicy {
   /**
-   * Try cache first, then network
+   * Try the cache, if that failed, try the network
```
BoD, line 39: Add exception behavior, and possible number of elements

martinbonnin (approved):

PR #3833, opened 2022-02-02
Fix null pointer exception in LruCache

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/3818

It turns out there were two issues:

* LruCache cannot use a read lock only because accesses actually write data structures
* The `moveNodeToHead` function wasn't updating the tail pointer


BoD (approved):
👍

PR #3835, opened 2022-02-02
Add a test using the cache with scalar adapter declared in build.gradle

BoD:


tests/integration-tests/src/main/graphql/com/apollographql/apollo3/integration/normalizer/schema.graphqls
```diff
@@ -88,10 +90,18 @@ interface Character {
...
+scalar Instant
+
 """
 UnsupportedType for testing
 """
 scalar UnsupportedType
+
+type InstantContainer {
+  instant: Instant
+}
```
martinbonnin, line 103: FWIW, I don't think the container is needed here, adding `Query.instant` would work as well for the purpose of the test

tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
```diff
@@ -162,4 +165,21 @@ class OtherCacheTest {
       assertTrue(e.message!!.contains("has no field named 'friends'"))
     }
   }
-}
\ No newline at end of file
+
+  @Test
+  fun withCustomScalarDeclaredInGradle() = runTest(before = { setUp() }, after = { tearDown() }) {
```
martinbonnin, line 170: Nit: I'd call this `withCompileTimeScalarAdapter` because:

1. We could support other build systems like Maven or even declare scalar adapters in some `config.graphqls` file so the test isn't directly related to Gradle
2. As any Scalar can now be mapped, there's no need to specify "Custom"

tests/integration-tests/build.gradle.kts
```diff
@@ -65,9 +65,8 @@ fun configureApollo(generateKotlinModels: Boolean) {
               }
               "normalizer" -> {
                 generateFragmentImplementations.set(true)
-                customScalarsMapping.set(mapOf(
-                    "Date" to "kotlinx.datetime.LocalDate"
-                ))
+                mapScalar("Date", "kotlinx.datetime.LocalDate")
+                mapScalar("Instant", "kotlinx.datetime.Instant", "com.example.ScalarAdapters.InstantAdapter")
```
martinbonnin, line 69: Could you save a wrapper here? 

```suggestion
                mapScalar("Instant", "kotlinx.datetime.Instant", "com.apollographql.apollo3.adapter.KotlinxInstantAdapter")
```

tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
BoD: Totally right!

tests/integration-tests/build.gradle.kts
BoD: That's what I tried first, but this doesn't work in Java (needs `.INSTANCE`).  I thought of trying `@JvmStatic` on the funs but the compiler doesn't like it because they are `override` :)

martinbonnin: I see. This is a failure at observing our [guidelines for Java interop](https://github.com/apollographql/apollo-kotlin/blob/a1536a269ffe5896f1b5290e55bd361e2d398861/CONTRIBUTING.md#L61). 

You could do something like this instead:

```kotlin
if (generateKotlinModels) {
    mapScalar("Instant", "kotlinx.datetime.Instant", "com.apollographql.apollo3.adapter.KotlinxInstantAdapter")
} else {
    mapScalar("Instant", "kotlinx.datetime.Instant", "com.apollographql.apollo3.adapter.KotlinxInstantAdapter.INSTANCE")
}
```

BoD: > This is a failure at observing our guidelines for Java interop.

Yeah 😿! To be fair, here a top level variable wouldn't helpe either (Kotlin: `com.apollographql.apollo3.adapter.KotlinxInstantAdapter` / Java: `com.apollographql.apollo3.adapter.KotlinTimeAdapters.KotlinxInstantAdapter`).

martinbonnin: Right ! oh well life is hard 😅

martinbonnin (approved):

Issue #3836, opened 2022-02-02
ApolloParseException caused by check failed when invoking guessNumber


rev0lgenX:
**Summary**
When mapping fromJson to generated class, in `JsonReaders#guessNumber` if `nextInt(MapJsonReader)` throws exception then rest of the method inside `guessNumber` fails and `nextNumber` will throw exception causing ApolloParseException .

**Version**
3.0.0

**Description**
When `nextInt` from `MapJsonReader` is invoked from `JsonReaders#guessNumber,` then  `MapJsonReader#nextValue` is invoked which will make `pendingName = null` in case of OrderedMap. Then, if nextInt throws error then rest of method call inside `JsonReaders#guessNumber` will fail because pendingName is null so when nextNumber is invoked, it invokes `MapJsonReader#nextValue` which will do `check(pendingName != null)` and throws exception.

![1](https://user-images.githubusercontent.com/20517801/152172262-618b4f90-6bee-4a8b-a565-dee0145c24d7.PNG)

![2](https://user-images.githubusercontent.com/20517801/152172288-4101a97b-8ddb-4ece-8324-87a775013257.PNG)

HELP!!!

rev0lgenX:
Hi, 
Here is a sample that demonstrate the issue. [Sample](https://github.com/rev0lgenX/apollo-kotlin-sample)

Parse exception only occurs when fragment is used. Here is the query and it's response

## Query
```
query MediaListQuery($userId:Int,$mediaId:Int $type:MediaType){

    mediaList: MediaList(mediaId: $mediaId, userId: $userId, type: $type){
        ... mediaListEntry

    }
}


fragment mediaListEntry on MediaList {
    advancedScores
    media{
        title{
            english
        }
    }
}
```

## Response

```
{
  "data": {
    "MediaList": {
      "advancedScores": {
        "Story": 9.5,
        "Characters": 0,
        "Audio": 0,
        "Enjoyment": 0
      },
      "media": {
        "title": {
          "english": "The Rising of the Shield Hero"
        }
      }
    }
  }
}
```

#### Issue is with the advancedScores in fragment which have double as value. Exception is not thrown when I don't use fragment.

BoD:
Thanks a lot for the repro project, it helped a ton! 🙏

There is an issue when parsing float numbers in custom scalars which are mapped to an `Any` type - which is the case of `Json` in your example - due to trying to read a value several times.

[...]

PR #3839, opened 2022-02-03
Make guessNumber read the next value only once

BoD:
Related to #3836

When using `AnyAdapter` (e.g. a custom scalar with no specified mapping), `JsonReader.readAny()` is used, which calls `guessNumber()`. This was calling `MapJsonReader.nextValue()` multiple times if the value could not be parsed as an `Int`, which isn't supported.

With this fix, we instead read the value once as a `JsonNumber` and try the different cases here.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
@@ -70,23 +70,32 @@ class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
     pendingName = null
   }
 
+  /**
+   * Calls to this method must be followed by calling [consumeValue] in the normal case. This permits retrying reading the same value in the
+   * exceptional case.
+   */
   private fun nextValue(): Any? {
     return when (val data = dataStack[stackSize - 1]) {
       is List<*> -> {
-        data[indexStack[stackSize - 1]++]
+        data[indexStack[stackSize - 1]]
       }
       is OrderedMap -> {
         check(pendingName != null)
-        pendingName = null
-        data.entries[indexStack[stackSize - 1]++].value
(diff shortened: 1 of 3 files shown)


martinbonnin:
Nice catch 👍 ! 

One drawback is that calling `nextNumber()` allocates a `String` which is sub-optimal, especially because `BufferedSourceJsonReader` can stream integers (it does not store a string representation of the integer). 

Is that a `MapJsonReader` thing only? If yes, could we modify `MapJsonReader` to not consume the value until it is confirmed that it can be parsed correctly?

BoD:
> Is that a MapJsonReader thing only? If yes, could we modify MapJsonReader to not consume the value until it is confirmed that it can be parsed correctly?

Yes, the issue doesn't exist with `BufferedSourceJsonReader`.

I'll amend with a the other approach, stay tuned :)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -220,11 +241,14 @@ class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
...
       is JsonNumber -> value
       else -> error("Expected JsonNumber but got $value instead")
+    }.also {
+      consumeValue()
     }
   }
 
   override fun skipValue() {
     nextValue()
+    consumeValue()
```
martinbonnin, line 251: I think you can remove `nextValue` here?

martinbonnin (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
BoD: Of course! 💪

martinbonnin (approved):

PR #3844, opened 2022-02-04
Fix error reporting when there is a "schema.graphqls" but it doesn't contain any type definition

martinbonnin:
If you have an empty `src/main/graphql/schema.graphqls` (or one containing executable definitions), the plugin fails with 

```
> Multiple schemas found:
  
  Use different services for different schemas
```

Which is completely misleading. Instead, output a better error:

```
Schema(s) found:
src/main/graphql/schema.graphqls

But none of them contain type definitions.
```

Many thanks to @pcarrier for catching this 💜


pcarrier (approved):

BoD (approved):
👍

PR #3852, opened 2022-02-08
fix typo in 3.0 migration guide

Saecki:


BoD:
The original thinking was to keep `addCustomTypeAdapter` in the "quick route" section because even though it's been marked deprecated, it still works.

But if we're going to touch these lines, we might as well change them with the new method name 😁. 

So I would merge this 👍

BoD (approved):

martinbonnin:
This has come up before so twice is enough to see a pattern there 👍

martinbonnin (approved):

PR #3853, opened 2022-02-08
Make watchers subscribe to the store earlier

olivierg13:
Just to reproduce an issue we've experienced while using two watchers on the same query.

### DIFF
--- a/tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
+++ b/tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
@@ -85,36 +85,32 @@ class WatcherTest {
 
     // The first query should get a "R2-D2" name
     mapResponses.register(query, ApolloResponse.Builder(query, uuid4(), episodeHeroNameData).build())
-    val job = launch {
-      apolloClient.query(query).fetchPolicy(FetchPolicy.NetworkOnly).watch().collect {
-        channel.send(it.data)
-      }
-    }
-    val job2 = launch {
-      apolloClient.query(query).fetchPolicy(FetchPolicy.NetworkOnly).watch().collect {
-        channel2.send(it.data)
-      }
-    }
-
-    assertEquals(channel.receiveOrTimeout()?.hero?.name, "R2-D2")
-    assertEquals(channel2.receiveOrTimeout()?.hero?.name, "R2-D2")
-
-    // Another newer call gets updated information with "Artoo"
     mapResponses.register(query, ApolloResponse.Builder(query, uuid4(), episodeHeroNameChangedData).build())
-    apolloClient.query(query).fetchPolicy(FetchPolicy.NetworkOnly).execute()
 
-    apolloClient.query(query).toFlow()
-        .catch {
-          // Handle errors here
+    repeat(100) {
+      val job = launch {
+        apolloClient.query(query).fetchPolicy(FetchPolicy.NetworkOnly).watch().collect {
+          channel.send(it.data)
         }
-        .onCompletion {
-          emitAll(apolloClient.query(query).watch())
+      }
+      val job2 = launch {
+        apolloClient.query(query).fetchPolicy(FetchPolicy.NetworkOnly).watch().collect {
+          channel2.send(it.data)
         }
-    assertEquals(channel.receiveOrTimeout()?.hero?.name, "Artoo")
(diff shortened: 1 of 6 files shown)


tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
```diff
@@ -74,19 +81,26 @@ class WatcherTest {
...
+        channel2.send(it.data)
+      }
+    }
 
     assertEquals(channel.receiveOrTimeout()?.hero?.name, "R2-D2")
+    assertEquals(channel2.receiveOrTimeout()?.hero?.name, "R2-D2")
 
     // Another newer call gets updated information with "Artoo"
-    apolloClient.enqueueTestResponse(query, episodeHeroNameChangedData)
+    mapResponses.register(query, ApolloResponse.Builder(query, uuid4(), episodeHeroNameChangedData).build())
```
BoD, line 103: In a similar vein as what @martinbonnin tried yesterday, if I add a single `yield()` right before this line, the test passes for me. Of course this doesn't really help explaining what's wrong... 😅

olivierg13: haha yeah I still don't know if it's a unit test environment issue or a real issue that could happen on a real device 🤔

tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
```diff
@@ -74,19 +81,26 @@ class WatcherTest {
   fun sameQueryTriggersWatcher() = runTest(before = { setUp() }) {
     val query = EpisodeHeroNameQuery(Episode.EMPIRE)
     val channel = Channel<EpisodeHeroNameQuery.Data?>()
+    val channel2 = Channel<EpisodeHeroNameQuery.Data?>()
```
BoD, line 84: Replacing these 2 lines with 
```kotlin
    val channel = Channel<EpisodeHeroNameQuery.Data?>(Channel.UNLIMITED)
    val channel2 = Channel<EpisodeHeroNameQuery.Data?>(Channel.UNLIMITED)
```

also fixes the test for me. 

I must admit I don't understand why. Without it, the `send` could block?

olivierg13: Ohhh interesting, maybe the Channels all share the same worker underneath. Not sure why it would block though yeah 🤔

olivierg13: The test passes in our CI as well with this change 🤔 
I'm not sure what could have happened. Maybe a recent change affected the watchers scopes underneath?

olivierg13: Channel.BUFFERED works as well. Wondering if the first `receiveOrTimeout()` is not somehow affecting the other Channel as well 🤔  although it wasn't before.

martinbonnin:
After some slack discussion, we realized this is a race between 2 coroutines:

coroutine1:
1. get R2-D2
1. emit R2-D2
1. subscribe to cache changes

coroutine2:
1. get Artoo
1. emit Artoo
1. publish Artoo change

What happens when the bug happens is that the coroutine2 publishes the changed keys before coroutine1 has a chance to subscribe to the store. This has very little chances of happening with a real network but is happening in tests because the network is mocked with a synchronous implementation.

I slightly modified the runtime to subscribe to the store before emiting the first item in a watch. This way, coroutine 2 can use the first item as a synchronisation point before starting the "Artoo" request.

Many thanks @olivierg13 for catching and troubleshooting this 💙

BoD (approved):
Yes, makes sense to me! 👍

olivierg13:
Thanks for the throrough investigations! Looks good to me! 😃

martinbonnin:
Heads up we had to change the implementation of this a bit to support the `CacheAndNetwork` fetchPolicy: https://github.com/apollographql/apollo-kotlin/issues/3946

I think the `buffer(2)` should still allow these tests to pass but let us if you notice anything weird

olivierg13:
Sounds good thanks!
Which version are these new changes gonna be included onto?

martinbonnin:
`3.1.1+` (most likely going to be a `3.2.0` but not 100% sure yet)

PR #3854, opened 2022-02-08
Minor edits to 3.1 docs

StephenBarlow:
The 3.1 doc changes look great! These are really minor changes I happened to make while reading through.

### DIFF
--- a/docs/source/essentials/custom-scalars.mdx
+++ b/docs/source/essentials/custom-scalars.mdx
@@ -14,13 +14,13 @@ To interact with custom scalars in your Apollo Kotlin app, you need to define a 
 ```kotlin:title=build.gradle[.kts]
 apollo {
   mapScalar("GeoPoint", "com.example.GeoPoint")
-  
+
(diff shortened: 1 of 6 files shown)


docs/source/essentials/custom-scalars.mdx
```diff
@@ -14,13 +14,13 @@ To interact with custom scalars in your Apollo Kotlin app, you need to define a
...
   mapScalar("GeoPoint", "com.example.GeoPoint")
-  
+
   // Shortcuts exist for standard types - equivalent to mapScalar("Long", "kotlin.Long")
   mapScalarToKotlinLong("Long")
 }
 ```
 
-Note: if needed, you can also do this with built-in scalars (like `ID`) to override the type used by default.
+> If needed, you can also do this with _default_ scalars (like `ID`) to override the type used by default.
```
martinbonnin, line 23: Nit: I like "built-in" better because this is the terminology from the spec: https://spec.graphql.org/draft/#sec-Scalars

martinbonnin (approved):
One small comment but 🤩 otherwise 🙏

PR #3858, opened 2022-02-09
@defer design docs

BoD:
In this doc, let's describe how the API of supporting `@defer` in Apollo-Kotlin would look like.

### DIFF
--- a/design-docs/Defer draft API.md
+++ b/design-docs/Defer draft API.md
@@ -1,0 +1,852 @@
+# Apollo Kotlin: @defer draft API
+
+This document outlines proposed changes to the user-facing API of Apollo Kotlin related to supporting the new `@defer`
+directive.
+
+Implementation is not discussed for now, the idea is to first agree on the desired API, other documents will dive into
+implementation.
+
+>
+> 💡 Note: even though both `@defer` and `@stream` are part of the same spec, it has been decided to focus on `@defer`
+> first, as it appears `@stream` has a lower demand.
+>
+
+## Context
+
+- Proposed GraphQL spec changes: https://github.com/graphql/graphql-spec/pull/742
+- Original RFC: https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md
+- Working Group: https://github.com/robrichard/defer-stream-wg
+
+This directive applies to fragment spreads and inline fragments. When present, the fragment’s fields can be received
+incrementally.
+
+For example:
+
+```graphql
+type Query {
+  computers: [Computer!]!
+}
+
+type Computer {
+  id: ID!
+  cpu: String!
+  year: Int!
+  screen: Screen!
+}
+
+type Screen {
+  resolution: String!
+  isColor: Boolean!
+}
+```
+
+```graphql
+query Query {
+  computers {
+    id
+    ... on Computer @defer {
+      cpu
+      year
+      screen {
+        resolution
+        ... on Screen @defer {
+          isColor
+        }
+      }
+    }
+  }
+}
+```
+
+Would be received as:
+
+```json
+// Payload 1
+{
+  "data": {
+    "computers": [
+      {
+        "id": "Computer1"
+      },
+      {
+        "id": "Computer2"
+      }
+    ]
+  },
+  "hasNext": true
+}
+
+// Payload 2
+{
+  "data": {
+    "__typename": "Computer",
+    "cpu": "386",
+    "year": 1993,
+    "screen": {
+      "resolution": "640x480"
+    }
+  },
+  "path": [
+    "computers",
+    0
+  ],
+  "hasNext": true
+}
+
+// Payload 3
+{
+  "data": {
+    "__typename": "Computer",
+    "cpu": "486",
+    "year": 1996,
+    "screen": {
+      "resolution": "800x600"
+    }
+  },
+  "path": [
+    "computers",
+    1
+  ],
+  "hasNext": true
+}
+
+// Payload 4
+{
+  "data": {
+    "__typename": "Screen",
+    "isColor": false
+  },
+  "path": [
+    "computers",
+    0,
+    "screen"
+  ],
+  "hasNext": true
+}
+
+// Payload 5 (final)
+{
+  "data": {
+    "__typename": "Screen",
+    "isColor": false
+  },
+  "path": [
+    "computers",
+    1,
+    "screen"
+  ],
+  "hasNext": false
+}
+```
+
+## User facing API changes
+
+### Generated models
+
+**Response based**
+
+Not supported at first. Like for `@skip` / `@include`, compilation will fail if `@defer` is used on fragments when using
+response based codegen.
+
+**Operation based**
+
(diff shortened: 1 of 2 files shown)


design-docs/Defer draft API.md
```diff
@@ -0,0 +1,852 @@
...
+      }
+    ]
+  },
+  "hasNext": true
+}
+
+// Payload 2
+{
+  "data": {
+    "__typename": "Computer",
```
martinbonnin, line 82: Is this __typename automatically sent with all defered payload? This looks weird this it wasn't requested in the query.

design-docs/Defer draft API.md
```diff
@@ -0,0 +1,852 @@
...
+```
+
+## User facing API changes
+
+### Generated models
+
+**Response based**
+
+Not supported at first. Like for `@skip` / `@include`, compilation will fail if `@defer` is used on fragments when using
+response based codegen.
```
martinbonnin, line 149: Agreed 👍 Let's keep things simple there

design-docs/Defer draft API.md
```diff
@@ -0,0 +1,852 @@
...
+    1. A Computer *(from request **a**)*
+    2. A Computer, with computerFields *(from request **a**)*
+    3. A Computer *(from request **b**)*
+    4. A Computer, with computerFields, with screenFields *(from request **a**)*
+    5. A Computer, with computerFields *(from request **b**)*
+    6. A Computer, with computerFields, with screenFields *(from request **b**)*
+
+Open questions
+
+- What happens if the key fields are deferred? Maybe this could be forbidden via a compile time check?
```
martinbonnin, line 801: Yikes 😱 . The compile time check should even be more complicated than this...

This is allowed:

```graphql
query Query {
  ... on Computer @defer {
    computer {
      id
      cpu
      year
    }
  }
}
```

This is not:


```graphql
query Query {
  ... on Computer @defer {
    computer {
      id
    }
  }
  computer {
     # we get cpu and year without the id yet so we cannot write it
      cpu
      year
  }
}
```

Or maybe this is allowed but the write only happens once the id is received 🤔

design-docs/Defer draft API.md
BoD: Good catch! No, they're not automatically sent, my mistake!

design-docs/Defer draft API.md
```diff
@@ -0,0 +1,852 @@
...
+
+**A. Incremental data is cached as it is being received**
+
+Pros:
+
+- When using `watch` the UI can be as reactive as possible and reflect the received data in real time
+
+Cons:
+
+- When using `query` with `FetchPolicy.CacheOnly` partial data can be returned, with no way to know this
```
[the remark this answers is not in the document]
BoD: > The normalized cache knows of null vs absent so if data is not there yet, it's going to do a cache miss

Oh I didn't realize that, that's a good thing! So that's one less "con" for this option.

> Overall, I think the cache should act as a server that simply ignores all @defer directives.

I'm not exactly sure if you mean this in favor or option A or B?

design-docs/Defer draft API.md
BoD: > Or maybe this is allowed but the write only happens once the id is received 🤔

Oh yeah If doable, this sounds like the best behavior for users (it "just works")!

design-docs/Defer draft API.md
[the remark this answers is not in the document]
martinbonnin: > I'm not exactly sure if you mean this in favor or option A or B?

I'm team A. Writers write data as soon as it's available. Readers read what's available from the cache. If someone tries to read a query containing `@defer` directives from the cache, the cache works as if `@defer` was not present and tries to resolve the deferred fragment as soon as possible. If it's not there, it's a cache miss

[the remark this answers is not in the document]
BoD: 👍  sounds fair! 

What about `watch` though? I feel like `watch` should emit the partial results as soon as they're received, because if not, it basically defeats the purpose of `@defer` when using the cache as the source of truth in a 'unidirectional data flow' architecture.

But that could mean some inconsistency: `watch` emits a partial result, while querying the cache directly doesn't return it.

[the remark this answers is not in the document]
martinbonnin: I think the question is not too much `A. store incremental data` vs `B. store only the final data`  but more how to handle reads. 

I think it makes sense to store the data as soon as it is received. Ideally, even store the individual payloads to not have to rewrite the initial ones all the time (requires the id but we will need it in all cases as discussed in the [comment below](https://github.com/apollographql/apollo-kotlin/pull/3858#pullrequestreview-877682891)).

> I feel like watch should emit the partial results as soon as they're received, because if not, it basically defeats the purpose of @defer when using the cache as the source of truth in a 'unidirectional data flow' architecture.

I'm not sure about this. Using `@defer` in a query that only hits the cache sounds like a red flag to me. The cache is fast, there should be no need to `@defer` fragments.

I _think_ `@defer` is another reason to move away from the current `watch()` (the one that supports fetchPolicy and refetchPolicy) and use `watch(data: Data?)` instead (the one that only reacts to store changes, without that initial fetch)

My favorite pattern for unidirectional data flows would be something like:

* each UI component has its own fragment 
* each UI component watches the store (CacheOnly, means we certainly need to introduce a watch() for fragments or else we'll have to wrap fragments in queries which is ok but cumbersome) 
* when the screen starts, a NetworkOnly query is made aggregating the various fragments, potentially using `@defer`

In that world, it's ok for the cache to ignore `@defer` completely (or maybe even throw). One issue with that model is that it makes it harder to communicate errors from the NetworkOnly query to the CacheOnly ones. But maybe once we can store more metadata in the cache we can store some error information there too?

The alternative is to start watching the cache while the initial fetch is still ongoing but I'm afraid this is trying to do "too many things at once"...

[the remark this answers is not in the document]
BoD: That makes a lot of sense 🙏.

I'd say having UI components mapped to fragments matches well with the "spirit" of the spec - so having a way to observe fragments indeed would be useful.

So to summarize:
- data is stored in the cache as it's received incrementally
- querying the cache with fragments that are not in the cache (yet) leads to cache miss
- same is true with `watch(Data?)`: only 'full' data is emitted from the cache

I'll update the doc with this.

design-docs/Defer merging implementation.md
```diff
@@ -0,0 +1,445 @@
...
+- Fairly complex / difficult (if even possible?) code generation to merge data classes
+
+
+Approach B:
+
+**Pros:**
+- Changes to the generated Adapter are minimal
+
+**Cons:**
+- JSON payloads must be parsed into Maps, which are kept in memory for a while
```
martinbonnin, line 443: Technically, that's better than keep the data classes, which may contain duplicated merged fields, in memory.

BoD:
I think we can merge this now?

martinbonnin (approved):
Ah good catch, yes 👍

PR #3861, opened 2022-02-11
Fix issue in Java codegen where selectors returned ImmutableMapBuilder instances instead of Map

anderssandven:
Added missing build statement to generated code. This issue caused JsonWriter.writeAny to fail as it doesnt handle the case where we pass in an ImmutableMapBuilder.

### DIFF
--- a/tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
+++ b/tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
@@ -185,7 +185,7 @@ class OtherCacheTest {
   }
 
   @Test
-  fun mutationWithoutVariable() = runTest(before = { setUp() }, after = { tearDown() }) {
+  fun mutationWithObjectValueArgument() = runTest(before = { setUp() }, after = { tearDown() }) {
     val mutation = UpdateReviewWithoutVariableMutation()
     val data = UpdateReviewWithoutVariableMutation.Data(
         UpdateReviewWithoutVariableMutation.UpdateReview(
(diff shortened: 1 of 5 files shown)


BoD:
Thanks a lot for this! 🙏 I've pushed the .`expected` + one little repro test

tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
```diff
@@ -182,4 +183,21 @@ class OtherCacheTest {
 
     assertEquals(instant, response.data!!.instant)
   }
+
+  @Test
+  fun mutationWithoutVariable() = runTest(before = { setUp() }, after = { tearDown() }) {
```
martinbonnin, line 188: Nit

```suggestion
  fun mutationWithObjectValueArgument() = runTest(before = { setUp() }, after = { tearDown() }) {
```

martinbonnin (approved):
Nice 🤩 . Thanks a lot 💙

tests/integration-tests/src/commonTest/kotlin/test/OtherCacheTest.kt
```diff
@@ -182,4 +183,21 @@ class OtherCacheTest {
 
     assertEquals(instant, response.data!!.instant)
   }
+
+  @Test
+  fun mutationWithObjectValueArgument() = runTest(before = { setUp() }, after = { tearDown() }) {
```
martinbonnin, line 188: Even more nitpicking (sorry 😅 ). It doesn't have to be a mutation but it has to be using the cache so after some more thoughts, I'd go with:

```suggestion
  fun cacheFieldWithObjectValueArgument() = runTest(before = { setUp() }, after = { tearDown() }) {
```

BoD (approved):

PR #3866, opened 2022-02-13
Mention we require Kotlin 1.6 for native

martinbonnin:
Moving forward, we should be more conservative about updating the Kotlin version as consumers will require at least the same version as the one we're using for building the lib. See also https://youtrack.jetbrains.com/issue/KT-42293


BoD (approved):
👍

PR #3873, opened 2022-02-16
🗄️  HttpCache: do not cache mutations by default

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/3872


apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -153,6 +160,11 @@ class CachingHttpInterceptor(
      */
     const val CACHE_FETCH_POLICY_HEADER = "X-APOLLO-CACHE-FETCH-POLICY"
 
+    /**
+     * Cache fetch strategy http header
```
BoD, line 164: Nitpick:
```suggestion
     * Cache operation type http header
```

BoD (approved):
👍

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
martinbonnin: Thanks!

PR #3874, opened 2022-02-16
Fix -1 body length in BatchingHttpInterceptor

BoD:
The code was trying to use the `Content-Length` header but it's actually never there, so the result was always `-1`. This resulted in OkHttp to switch to chunk transfer-encoding on the request (not an issue in itself, but it's nicer to send a Content-Length if we know it, I guess!).

Instead, combine the bodies as a String, and use that.

- Pros: correct size
- Cons: temporarily keep the combined array in memory, while before the data could be streamed

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
@@ -133,13 +133,6 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
 
     val firstRequest = pending.first().request
 
-    val allLengths = pending.map { it.request.headers.valueOf("Content-Length")?.toLongOrNull() ?: -1L }
-    val contentLength = if (allLengths.contains(-1)) {
-      -1
-    } else {
-      allLengths.sum()
-    }
-
     val allBodies = pending.map { it.request.body ?: error("empty body while batching queries") }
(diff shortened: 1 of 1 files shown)


martinbonnin:
The previous code was very wrong but also kind of more optimized too.

What about making this configurable? By default we'll stream. If a backend doesn't support that, the user can set an option and we'll buffer all the data in memory and send it alongside its Content-Length?

BoD:
In the end I think it's ok to keep only the streaming version :) It avoids complexifying the API/code and we don't have reasons to think chunked encoding could cause problems (it looks like it has been working well so far!). LMK if you're ok with that!

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -150,7 +143,11 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
 
     val body = object : HttpBody {
       override val contentType = "application/json"
-      override val contentLength = contentLength
+
+      // We don't know the combined size at that point.
+      // Note: returning -1 here will switch to chunked encoding when using OkHttp.
```
martinbonnin, line 148:
```suggestion
      // Note: this assumes the underlying HttpEngine will use chunked encoding when contentLength is -1.
```

martinbonnin (approved):
Works for me. I've definitely seen servers that did not support chunked encoding before but maybe it's much more widespread now 👍

PR #3879, opened 2022-02-17
✅  Validation: allow nullable variables in non-null locations if there is a default value

martinbonnin:
Implement http://spec.graphql.org/draft/#sec-All-Variable-Usages-are-Allowed.Allowing-optional-variables-when-default-values-exist

Closes https://github.com/apollographql/apollo-kotlin/issues/3227 
See https://github.com/apollographql/apollo-kotlin/issues/3878


BoD (approved):
💪

PR #3889, opened 2022-02-21
Add support for receiving multiple bodies with multipart

BoD:
- Import `MultipartReader` from OkHttp
- Use it in `HttpNetworkTransport`: when a response is in multipart content type, then each part is parsed as an individual response, and emitted in the Flow
  - At this point this doesn't make sense in terms of GraphQL, as parts are parsed as "full" (non incremental) payloads, which would never happen in real life: multipart should only happen in a `@defer`/`@stream` scenario.
  - But it's a necessary intermediary step (I think it's ok to merge this in `main` but we could start a "defer" branch instead if that's preferable)

#### ⚠️ I enabled `-Xjvm-default=compatibility` in this PR
- In `MultipartReader`, Okio's  `Source` is implemented
- This breaks compilation with `Inheritance from an interface with '@JvmDefault' members is only allowed with -Xjvm-default option`
- `Source` doesn't have default methods, and it looks like this is a compiler bug (https://youtrack.jetbrains.com/issue/KT-45919 - which is marked as resolved in upcoming `1.6.20`)
- A way to get rid of this error is to enable `-Xjvm-default=compatibility`, which from reading [this](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/) I think has no impact since we don't use `@JvmDefault` anywhere.
- (Tried [`all-compatibility`](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/) first, but it breaks in generated code due to the same bug)

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
@@ -181,6 +181,7 @@ class MultipartReader constructor(
   }
 
   /** A single part in a multipart body. */
+  @ApolloInternal
   class Part(
       val headers: List<HttpHeader>,
       val body: BufferedSource,
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
@@ -1,0 +1,209 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ * Copyright (c) 2022 Apollo Graph, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.apollographql.apollo3.internal
+
+import com.apollographql.apollo3.api.http.HttpHeader
+import com.apollographql.apollo3.exception.ApolloException
+import okio.Buffer
+import okio.BufferedSource
+import okio.ByteString.Companion.encodeUtf8
+import okio.Closeable
+import okio.Options
+import okio.Source
+import okio.buffer
+
+/**
+ * This was copied from OkHttp and slightly modified to work in a Multiplatform project.
+ *
+ * Original source: https://github.com/square/okhttp/blob/parent-4.9.2/okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
+ *
+ * Reads a stream of [RFC 2046][rfc_2046] multipart body parts. Callers read parts one-at-a-time
+ * until [nextPart] returns null. After calling [nextPart] any preceding parts should not be read.
+ *
+ * Typical use loops over the parts in sequence:
+ *
+ * ```
+ * val response: Response = call.execute()
+ * val multipartReader = MultipartReader(response.body!!)
+ *
+ * multipartReader.use {
+ *   while (true) {
+ *     val part = multipartReader.nextPart() ?: break
+ *     process(part.headers, part.body)
+ *   }
+ * }
+ * ```
+ *
+ * Note that [nextPart] will skip any unprocessed data from the preceding part. If the preceding
+ * part is particularly large or if the underlying source is particularly slow, the [nextPart] call
+ * may be slow!
+ *
+ * Closing a part **does not** close this multipart reader; callers must explicitly close this with
+ * [close].
+ *
+ * [rfc_2046]: http://www.ietf.org/rfc/rfc2046.txt
+ */
+internal class MultipartReader constructor(
+    private val source: BufferedSource,
+    val boundary: String,
+) : Closeable {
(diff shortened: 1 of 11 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
```diff
@@ -0,0 +1,209 @@
...
+import okio.ByteString.Companion.encodeUtf8
+import okio.Closeable
+import okio.Options
+import okio.Source
+import okio.buffer
+
+/**
+ * This was copied from OkHttp and slightly modified to work in a Multiplatform project.
+ *
+ * Original source: https://github.com/square/okhttp/blob/parent-4.9.2/okhttp/src/main/kotlin/okhttp3/MultipartReader.kt
```
martinbonnin, line 32: Heads up we have another multipart reader [there](https://github.com/apollographql/apollo-kotlin/blob/f476e024c0c74e222270cc357ee48202df43ee0a/tests/integration-tests/src/commonTest/kotlin/test/FileUploadTest.kt#L155). Might be worth removing it?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -67,32 +73,41 @@ private constructor(
...
-                      startMillis = millisStart,
-                      endMillis = currentTimeMillis(),
-                      statusCode = httpResponse.statusCode,
-                      headers = httpResponse.headers
-                  )
-              )
-              .build()
+      emitAll(
+          bodies.map { body ->
+            val operation = request.operation
```
martinbonnin, line 86:
```suggestion
            // do not capture request
            val operation = request.operation
```

I'm not sure why this was important but I'd keep the comment for context

build-logic/src/main/kotlin/TargetCompatibility.kt
```diff
@@ -10,7 +10,13 @@ fun Project.configureJavaAndKotlinCompilers() {
...
     tasks.withType(KotlinCompile::class.java) {
       it.kotlinOptions {
-        freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
+        freeCompilerArgs = freeCompilerArgs +
+            "-Xopt-in=kotlin.RequiresOptIn" +
+            // This is a workaround for https://youtrack.jetbrains.com/issue/KT-47000 (fixed in Koltin 1.6.20)
+            // Since we don't use @JvmDefault anywhere, the option has no effect, but suppresses the bogus compiler error
+            // See also https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/
+            // See also https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/
+            "-Xjvm-default=compatibility"
```
martinbonnin, line 19: Out of curiosity, why `compatibility` and not `enable`?

BoD: My thinking was we're not using `@JvmDefault` anywhere at the moment, plus we should probably remove this "soonish" when upgrading to 1.6.20.
But if we don't, and we do use `@JvmDefault` at some point, I'm thinking the `compatibility` mode is safer (the doc says "binary compatible")?  Not 100% sure it's really useful since if this happens, the Api Dump should warn us anyway, right?

martinbonnin: Not sure how much of apiDump can catch that. I did a run of japicmp with and without `-Xjvm-default=compatibility` and it didn't show any difference so I guess it is safe. I'll feel better once we upgrade to 1.6.20 though

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -67,32 +76,45 @@ private constructor(
...
+            // On native, body will be freezed in worker.doWork, but reading a BufferedSource is a mutating operation.
+            // So we read the bytes into a ByteString first, and create a new BufferedSource from it after freezing.
+            val bodyByteString = if (platform() == Platform.Native) body.readByteString() else null
```
BoD, line 92: Here's a trick to not crash inside `worker.doWork` when reading the body on native. 

I'm not very satisfied with it, especially the explicit call to `platform()`... 

Performance-wise we're already dealing with bodies fully read in memory on Apple, so I think it's ok.

In any case LMK if you can think of a better way to deal with it 🙏

hakib1 (approved):

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
martinbonnin: Yea no silver bullet... That works for me 👍

tests/integration-tests/src/commonTest/kotlin/test/DeferTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.apollographql.apollo3.integration.normalizer.HeroNameQuery
+import com.apollographql.apollo3.mockserver.MockServer
+import com.apollographql.apollo3.mockserver.enqueueMultipart
+import com.apollographql.apollo3.testing.runTest
+import kotlinx.coroutines.flow.toList
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@OptIn(ApolloExperimental::class)
+class DeferTest {
```
martinbonnin, line 16: Might be a bit premature to call this "DeferTest" ;). But works for me!

martinbonnin (approved):

PR #3893, opened 2022-02-24
Detect missing arguments when there are no arguments at all

martinbonnin:
Because we were doing a safe call, the validation logic was skipped if no arguments were present at all

Closes https://github.com/apollographql/apollo-kotlin/issues/3891


BoD (approved):
👍

PR #3894, opened 2022-02-24
👷 Test Builders: Fix enums in test resolver

martinbonnin:
When no value is specified for an enum, the code was using the composite path (objects/interfaces) and failed with a division by zero. 

Add a specific path for enums to avoid the crash.

Many thanks to @StylianosGakis for the heads up and investigations!


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -61,6 +66,27 @@ class EnumAsEnumBuilder(
     return TypeSpec.companionObjectBuilder()
         .addProperty(typePropertySpec())
         .addFunction(safeValueOfFunSpec())
+        .addFunction(knownValuesFunSpec())
```
martinbonnin, line 69: This is so that we can do `SomeEnum.Companion.knownValues()` for both enums as enums and enums as sealed classes.

BoD (approved):
💪

PR #3895, opened 2022-02-25
⚙️  Add support for triple quotes escapes

martinbonnin:
Add support for `\"""`:

```graphql
"""
It is valid to have \""" here
"""
```

Also improve the indendation trimming to match better with https://spec.graphql.org/draft/#BlockStringValue(). Previously, it would strip indents on the first line too.


BoD (approved):
"""
👍
"""

PR #3896, opened 2022-02-25
🐘  Use resolved version and not declared version for checkApolloVersions

martinbonnin:
See https://github.com/jmfayard/refreshVersions/issues/507

Instead of getting the version from the declared dependencies, get it from the resolved one. Fixes usage with [refreshVersions](https://github.com/jmfayard/refreshVersions).

This is going to resolve the "runtimeClasspath" configuration earlier than was the case before so I hope that won't trigger any issues. The resolution is still done lazily so it should be done after all the plugin have contributed their own dependencies 🤞.

If that causes any problem, the check can be disabled with 

```kotlin
tasks.configureEach {
  if (name == "checkApolloVersions") {
    enabled = false
  }
}
```


BoD (approved):
🎉

Issue #3882, opened 2022-02-21
@defer: codegen changes


BoD:
- [x] Add directive to built-ins
- For now, only the `operationBased` codegen will be impacted. 
  - [x] Output an error when using `@defer` with the responseBased codegen.
- [x] Handle fragment synthetic fields nullability and merging
  - [x] Fragment spreads
  - [x] Inline fragment
- [x] In generated adapters, read fields or not depending on presence of fragment's path+label

PR #3901, opened 2022-02-28
Defer: `DeferredJsonMerger`

BoD:
Starting to implement "Approach B" from [this doc](https://github.com/apollographql/apollo-kotlin/pull/3858/files#diff-3fe0960286149c9e1b604842ea5bd3e18a326f6f59fac3cce97e3ad53e0714e8R293):

- Introduce `DeferredJsonMerger` (not sure of the name!) which is a utility to merge JSON payloads
- Add `@defer` to the builtins ~and automatically add unique `label` argument in `ApolloCompiler`~

Related to #3882


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.apollographql.apollo3.ast.GQLDirective
+import com.apollographql.apollo3.ast.GQLFragmentDefinition
+import com.apollographql.apollo3.ast.GQLNode
+import com.apollographql.apollo3.ast.GQLOperationDefinition
+import com.apollographql.apollo3.ast.GQLStringValue
+import com.apollographql.apollo3.ast.NodeTransformer
+import com.apollographql.apollo3.ast.TransformResult
+import com.apollographql.apollo3.ast.transform
+
+fun addDeferLabel(operation: GQLOperationDefinition): GQLOperationDefinition {
```
martinbonnin, line 14: Why this is required? From the [RFC](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md), I understand the label is optional. If that's the case, I'd avoid modifying the user document too much. We do it for `__typename` and some ids but the less we modify the better.

From my very superficial understanding of the RFC, the label is only required when 2 deferred fragments have the same path? If that's the case, I'd rather much have a validation rule to fail compilation if 2 fragments with the same paths don't have a "label" (and I guess one having a label but not the other is enough to disambiguate)

apollo-ast/src/main/resources/builtins.graphqls
```diff
@@ -131,4 +131,9 @@ directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
...
 directive @deprecated(
   reason: String = "No longer supported"
-) on FIELD_DEFINITION | ENUM_VALUE
\ No newline at end of file
+) on FIELD_DEFINITION | ENUM_VALUE
+
+directive @defer(
+  label: String
+  if: Boolean
+) on FRAGMENT_SPREAD | INLINE_FRAGMENT
```
martinbonnin, line 139: 👍

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt
```diff
@@ -0,0 +1,112 @@
...
+ * Each call to [merge] will merge the given chunk into the [merged] Map, and will also update the [mergedFragmentLabels] Set with the
+ * value of its `label` field (if present).
+ *
+ * The fields in `data` are merged into the node found in [merged] at `path` (for the first call to [merge], the payload is
+ * copied to [merged] as-is).
+ *
+ * Errors are merged by adding them to [merged]'s `errors` list field.
+ *
+ * Extensions are merged by adding them into [merged]'s `extensions` field - if several payloads have the same key in `extensions`,
+ * their values are stored as a list.
```
martinbonnin, line 20: For errors/extensions, I _think_ I'd prefer only exposing the last value to users. If users need the list, they can always build it as payloads are received.

tests/defer/src/test/kotlin/test/DeferTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        )
+    )
+    val jsonList = expectedDataList.map { data ->
+      buildJsonString {
+        query.composeJsonResponse(jsonWriter = this, data = data)
+      }
+    }
+    mockServer.enqueueMultipart(jsonList)
+
+    val actualDataList = apolloClient.query(query).toFlow().toList().map { it.dataAssertNoErrors }
```
martinbonnin, line 82: This is not valid GraphQL, right? (Getting multipart HTTP without `hasNext` and `path` outside the context of `@defer`/`@stream`).

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt
```diff
@@ -0,0 +1,112 @@
...
+  @Suppress("UNCHECKED_CAST")
+  private fun mergeData(payloadMap: Map<String, Any?>) {
+    val payloadData = payloadMap["data"] as Map<String, Any?>?
+    val payloadPath = payloadMap["path"] as List<Any>
+    val mergedData = merged["data"] as Map<String, Any?>
+
+    // payloadData can be null if there are errors
+    if (payloadData != null) {
+      val nodeToMergeInto = nodeAtPath(mergedData, payloadPath) as MutableMap<String, Any?>
+      nodeToMergeInto += payloadData
```
martinbonnin, line 53: This will overwrite merged fields if any, you need to deep merge. 

Example:

```graphql
query Query1 {
  computers {
    id
    screen {
       isColor
    }
    ...ComputerFields @defer
  }
}

fragment ComputerFields on Computer {
  # writing screen will lose `isColor`
  screen {
    resolution
  }
}```

File: apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt (hunk 1 above)
Comment (BoD) [reply apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt:14]: Correct, it is optional. The [idea](https://github.com/apollographql/apollo-kotlin/pull/3858/files#diff-3fe0960286149c9e1b604842ea5bd3e18a326f6f59fac3cce97e3ad53e0714e8R373) is to use the label as a way to know when to / not to parse a fragment's fields. Since labels must be unique across a document they are a good match for this - from the RFC they're used "to identify the data from responses and associate it with the corresponding defer directive".

Indeed, to avoid automatically adding labels, we may be able to use "path+label" instead, and ensure that all "path+label" are unique. Inside the generated adapter code, I'm not yet sure if it's easy to know the current path?

However, going this route, whenever several defer fragments are present at the same level, the path will be the same, and it will be up to the user to resolve the ambiguity by adding a unique label. From a user's perspective, wouldn't it be nicer for the SDK to take care of this automatically?

File: apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt (hunk 3 above)
Comment (BoD) [reply apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt:20]: Wouldn't it make the API a bit inconsistent if some parts are merged but not all parts? Do you see disadvantages of having the errors/extensions merged? In the case of extensions, I agree it's a bit awkward to have to create a list when the same key is used in several chunks...

File: apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt (hunk 5 above)
Comment (BoD) [reply apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt:53]: Good catch! 👍

File: apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt (hunk 1 above)
Comment (martinbonnin) [reply apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt:14]: > From a user's perspective, wouldn't it be nicer for the SDK to take care of this automatically?

I'd prefer to give the user an error and let them come up with the label. This way, they can copy/paste the GraphQL document to/from GraphiQL/Studio and it'll work out of the box. We can add "auto-label" down the road if it comes up a lot but if we start with "auto-label" from day 1, we will never be able to go back if we ever regret the decision.

File: tests/defer/src/test/kotlin/test/DeferTest.kt (hunk 4 above)
Comment (BoD) [reply tests/defer/src/test/kotlin/test/DeferTest.kt:82]: Yes totally - this tests the current "preliminary" implementation in `HttpNetworkTransport` that can receive several full payloads inside chunks, which should never happen. This test will need to be updated or removed - I'll add a comment / TODO to clarify this.

File: apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt (hunk 3 above)
Comment (martinbonnin) [reply apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt:20]: > Wouldn't it make the API a bit inconsistent if some parts are merged but not all parts?

The spec is inconsistent. It defines how to merge data but doesn't tell how to merge errors/extensions.

> Do you see disadvantages of having the errors/extensions merged?

Let's say you have 

```
payload1 - data1 - (error1, error2)

payload2 - data2
```

When receiving payload2, one has to make a diff with payload1 to see if there are any new errors. 

So if you want to handle errors per payload, it's more complicated. Maybe that's never going to happen but maybe it is, not sure.

We could even think of a scenario where the error gets "cleared":

```graphql 

query GetHero {
  hero (
    # get the details asap for optimistic rendering, might fail
    ...heroDetails
    # also get the details defered, will never fail
    ...heroDetails @defer
  }
}
```

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt
BoD: I see what you mean. Just to be 100% thorough, OTOH, there could be cases where a valid query (that works in GraphiQL/Studio) would be rejected by the compiler, as the rule to have unique path+label is not in the spec but specific to us.

martinbonnin: >  the rule to have unique path+label is not in the spec but specific to us.

Would it make sense to put it in the spec? 

Something like below sounds ambiguous:

```graphql
query GetHero {
  hero {
    ...heroDetails1 @defer
    ...heroDetails2 @defer
  }
}
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt
BoD: Merging errors and extensions:
  - Pros:
    - Consistency: everything is merged in the emitted ApolloResponse
    - Looking at the last response is as if there was no `@defer` at all (I was thinking of maybe making `ApolloRequest.execute` return the last element - perhaps it is not that useful and throwing is actually better)
    - Having centralized error handling may be easier?
  - Cons:
    - To know if an error is new (specific to a partial response), users would need to diff manually
    - No way to clear errors if some are present on the same fragment on payload n and not on payload n+1 (example above)
    - Merging extensions is awkward because they're maps

I have the impression that the cons outweigh the pros 😅 For me the big plus was the consistency one but maybe it's not a big deal after all, as long as the doc is clear.

So I'll remove it (we can always revisit this if we have second thoughts!).

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/transformation/add_defer_labels.kt
BoD: A follow-up: this was not 100% clear to me before but now I realize in any case, we *must* use path+label because whenever arrays are involved, the label alone is not enough. In [this example](https://github.com/apollographql/apollo-kotlin/pull/3858/files#diff-3fe0960286149c9e1b604842ea5bd3e18a326f6f59fac3cce97e3ad53e0714e8R351), the fields for the `ComputerFields` fragment must be parsed for `["computers",0]` and not for `["computers",1]`.

BoD: Update: removed the automatically computed label (and created #3938).

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/DeferredJsonMerger.kt
```diff
@@ -50,12 +50,27 @@ class DeferredJsonMerger {
     // payloadData can be null if there are errors
     if (payloadData != null) {
       val nodeToMergeInto = nodeAtPath(mergedData, payloadPath) as MutableMap<String, Any?>
```
martinbonnin, line 58: What is guaranteeing that the cast to MutableMap is safe?

martinbonnin: ALright, because it's all coming from `_merged = mutableMapOf<String, Any?>()` above 👍

PR #3902, opened 2022-03-01
⚙️  add `Service.typePackageName`

martinbonnin:
Add `typePackageName` as a way to override the `**.type.*` package name and keep 3.0.0 behaviour:

```
dependencies {
  apolloMetadata(project(":graphqlSchema"))
}

apollo {
  packageName.set("com.feature1")
  // Without this, codegen will use the packageName of the schema module
  typePackageName.set("com.feature1")
}
```

By default:

* for modules containing a schema, it is `packageName` or the result of applying `packageNameGenerator` to the schema file
* for multi-modules child modules, it is the typePackageName of the schema module

For for multi-modules child modules, this default value allows to change the module a given type is generated easily with `alwaysGenerateTypesMatching`. Also this default value doesn't require a file path for users of `packageNamesFromFilePaths()`

Many thanks to @childofthehorn for catching this.


martinbonnin:
Closing this for now. Let's try to avoid complexifying the (already complex) package name handling. If there are more compelling reasons to re-introduce this, I'll reopen this PR.

PR #3905, opened 2022-03-02
🗄️  fix the cache keys of fragments written programmatically 

martinbonnin:
Do not strip the first cache key component if it's not QUERY_ROOT

Closes https://github.com/apollographql/apollo-kotlin/issues/3875


BoD (approved):
👍

PR #3907, opened 2022-03-04
Adds JvmOverloads annotation to Kotlin Type generation if enabled

theBradfo:
Per https://github.com/apollographql/apollo-kotlin/issues/3848 this is the start of the introduction to add `@JvmOverloads` to kotlin generated sources.  

I'm not sure though what the expectations are for adding tests in this project, given it seems these are generated "expected" files - so some guidance there would be appreciated!

I'm also not completely sure how these `Options` are written, do i need to update `ApolloGenerateSourcesTask` to mutate this new field?


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/OperationBuilder.kt
```diff
@@ -79,6 +80,9 @@ class OperationBuilder(
         .addSuperinterface(superInterfaceType())
         .maybeAddDescription(operation.description)
         .makeDataClass(operation.variables.map { it.toNamedType().toParameterSpec(context) })
+        .applyIf(useJvmAnnotations) {
```
theBradfo, line 83: Also curious to hear if there's a better way to know if this annotation would even be useful (there are arguments that have default values) 🤔

martinbonnin: I think that's a good idea. Keeps the generated code more readable. It'd be confusing to see a `@JvmOverloads`
 annotation on a function without any default parameter.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/OperationBuilder.kt
```diff
@@ -79,6 +80,9 @@ class OperationBuilder(
...
+          addAnnotation(JvmOverloads::class)
```
martinbonnin, line 84: I think I misguided you 🤦 . This is adding the annotation on the class itself, not the constructor. So I don't think it'd work?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/OperationBuilder.kt
```diff
@@ -35,6 +35,7 @@ class OperationBuilder(
     private val generateQueryDocument: Boolean,
     private val operation: IrOperation,
     flatten: Boolean,
+    private val useJvmAnnotations: Boolean = false,
```
martinbonnin, line 38: We'll need this to be available in `FragmentBuilder`. It's a lot more niche but it's used when reading fragments from the cache.

martinbonnin:
Thanks a lot 💜 ! 

A few comments but looks good otherwise!

Also, do you mind renaming `useJvmAnnotations` to `addJvmOverloads`? I'm not sure there are any other JVM annotations we might want to opt-in in the future and if we do, we might want them to be independant, i.e. add overloads but maybe not other annotations.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/OperationBuilder.kt
theBradfo: Oops - you're right.  More on that though - what is the best way to test this?  Should be feasible to add a unit test, but i'm not sure what the best approach there is given the existing test infra?

martinbonnin: You can turn the option on in [`CodegenTests`](https://github.com/apollographql/apollo-kotlin/blob/a1536a269ffe5896f1b5290e55bd361e2d398861/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt#L271). Pick a directory that has variables with default values (`variable_default_value` sounds like a good one :)) and it will make sure it compiles. You'll need to regenerate the test fixtures with `./gradlew :apollo-compiler:test -DupdateTestFixtures=true` and commit the changed files

theBradfo:
Added a test and updated per your feedback, thanks @martinbonnin for the guidance here 🎉.  I did see updating the expected test source regenerated the `measurements` file, but seems like this is expected given the commit history?  Let me know if I should roll that back.

martinbonnin:
> updating the expected test source regenerated the measurements file, but seems like this is expected given the commit history?

Yup, this is fine. This file is mostly there so we can track performance regressions in the codegen but we do not enforce anything yet.

One last thing missing from this PR is to expose `addJvmOverloads` to the Gradle build scripts from the `Service` class. Let me know if you want to add it or I can take it from there if you prefer.

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -321,6 +321,13 @@ interface Service {
    */
   val useSchemaPackageNameForFragments: Property<Boolean>
 
+  /**
+   * Whether to generate kotlin constructors with @JvmOverloads for more graceful Java interop experience when default values are present.
```
martinbonnin, line 325: 👍

martinbonnin (approved):
Thanks!

martinbonnin:
The CI failed on `apiCheck` because `Service.addJvmOverloads` wasn't in the api dump. I just pushed this.

theBradfo:
Thanks @martinbonnin for the push!  It looks like there was a [network error connecting to maven central](https://github.com/apollographql/apollo-kotlin/runs/5447199822?check_suite_focus=true#step:4:3140), is this something you've seen in the past in the CI setup for this project?

martinbonnin:
> It looks like there was a [network error connecting to maven central](https://github.com/apollographql/apollo-kotlin/runs/5447199822?check_suite_focus=true#step:4:3140), is this something you've seen in the past in the CI setup for this project?

This stuff happens. I just relaunched the build

PR #3911, opened 2022-03-08
⚙️ Detect case insensitive filesystems (like MacOS default one) and rename classes when that happens

### LINKED ISSUE #3880: Handle scalar types with the same name but different case (e.g. "URL" and "Url")
**Summary**
Scalar types with the same name but different case (e.g. "URL" and "Url") result in compilation errors in the generated code.

**Version**
3.1.0

**Description**
I have a schema using "scalar URL" in some places, "scalar Url" in others. I was able to register a custom scalar type for "URL" in v2 and it worked fine, but now in v3 I'm only getting a type generated for `URL` with compilation failures looking for type `Url`.

I realize I should just use the same case in all places, and I'm working with other teams in our org to fix this. Nevertheless it is a confusing error and a change in behavior.

martinbonnin:
If a schema has 2 types with names differing only by their case (which is allowed in GraphQL), the generated files ended up overwriting each other:

```kotlin
scalar Url
# URL is a distinct GraphQL type but URL.kt is the same file as Url.kt on most MacBooks
scalar URL
```

When that happens, we'll rename the Kotlin class to avoid file clashes. In that example, `URL.kt` and `Url1.kt` will be generated. The names are sorted using the natural sort order before suffixes are added.

Closes #3880


tests/filesystem-sensitivity/src/test/kotlin/MainTest.kt
```diff
@@ -0,0 +1,45 @@
+import kotlin.test.Test
+
+class MainTest {
+  @Test
+  fun test() {
+
+    val os = System.getProperty("os.name").lowercase()
+    if (os.contains("mac")) {
```
BoD, line 8: Won't work on Macs with case-sensitive FS but we could use/duplicate `isFileSystemCaseSensitive` if that ever happens to be a problem :)

BoD (approved):
👍

tests/filesystem-sensitivity/src/test/kotlin/MainTest.kt
martinbonnin: I was hoping that both Github MacOS runners and most dev laptops would use case insensitive filesystems. The good thing is that if it's the case, is also tests `isFileSystemCaseSensitive`. 🤞 it'll be the case. If not I'll find another test I guess.

PR #3913, opened 2022-03-09
 WebSocket support for JS targets

### LINKED ISSUE #2463: WebSocket support for JS targets

- Many projects include a JS target. Supporting JS targets will increase Apollo’s viability when being considered for multi-platform projects.
- It is not straightforward to use Apollo in projects that include a JS target. This is due to the `apollo-runtme-kotlin` artifact, which can’t be used in `commonMain`. One workaround is to include `apollo-runtime-kotlin-jvm` in `androidMain` and `apollo-runtime-kotlin-ios-x64` in `iosMain`, then write dozen of expect/actual to fill the gap in `jsMain`

- Add `js` target in the build definition.
- Provide `actual` implementations for the required classes (i.e `ApolloHttpNetworkTransport`).
- Probably many more things.

undermark5:
Resolves #2463


apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/KtorWebSocketEngine.kt
```diff
@@ -1,8 +1,53 @@
...
+    val messageChannel = ChannelWrapper(Channel<String>(Channel.UNLIMITED))
+    val socket = WebSocket(url)
+    val webSocketOpenResult = CompletableDeferred<Unit>()
+    socket.onopen = { webSocketOpenResult.complete(Unit) }
+    socket.onmessage = { messageEvent: MessageEvent ->
+      val data = messageEvent.data
+      if (data is String) {
+        messageChannel.trySend(data)
+      }
+      if (data is ByteString) {
```
undermark5, line 27: This is the big thing that I'm not 100% certain about. The type of data is `Any?` so I'm not sure if we will be getting. I know that at least in some cases (the server I was using to verify that this worked) resulted in strings over the socket, but I'm not certain what format a binary message would come back as. I guessed something that could maybe be interpreted as a `ByteString`, but I really don't know, it could be `ByteArray` or `Uint8Array` or something completely different. If anyone has a way to determine it or know of a better way of solving the problem of receiving the messages that would be wonderful.

martinbonnin:
Thanks for sending this 💜 .

Ideally, we'll need some tests for this. We have an issue about [adding MPP websocket support to MockServer](https://github.com/apollographql/apollo-kotlin/issues/3259) but it's nowhere near ready yet. In the meantime, you can use the `FullStackTest`. You can use this [commit](https://github.com/martinbonnin/apollo-android/commit/821293d9afbaa81c54ae885d4bb9e685e5923d4d) to enable it and trigger an event from [sandbox](https://studio.apollographql.com/sandbox/explorer?endpoint=https%3A%2F%2Fapollo-fullstack-tutorial.herokuapp.com%2Fgraphql&explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4RxighigSwiQAIAhCCAawBUAnAgBwAoASAG1xiSgAsBJMAGd0JANr8AIgF0AhAEoSwADqkSAI0q0GjIc07c%2BgkSQ5ceA4YpVqSJITChQEQoarsBfVR5AAaEABuuAy46uwuGCA2dsogBhbGsaJisQCMAAwAnLHS3iAeQA). It currently fails with

```
FullStackTest.simple STANDARD_OUT
    ReferenceError: WebSocket is not defined
        at $openCOROUTINE$15.doResume_0_k$ (/Users/mbonnin/git/apollo-android/tests/KtorWebSocketEngine.kt:19:18)
        at DefaultWebSocketEngine.open_5uin04_k$ (/Users/mbonnin/git/apollo-android/tests/KtorWebSocketEngine.kt:14:20)
        at $superviseCOROUTINE$8.doResume_0_k$ (/Users/mbonnin/git/apollo-android/tests/WebSocketNetworkTransport.kt:174:31)
        at supervise (/Users/mbonnin/git/apollo-android/tests/WebSocketNetworkTransport.kt:110:19)
        at _no_name_provided__301.doResume_0_k$ (/Users/mbonnin/git/apollo-android/tests/WebSocketNetworkTransport.kt:80:7)
        at _no_name_provided__301.invoke_2i3g7c_k$ (/Users/mbonnin/git/apollo-android/tests/WebSocketNetworkTransport.kt:79:27)
        at l (/Users/mbonnin/git/apollo-android/tests/build/js/packages/tests-websockets-test/kotlin/tests-websockets-test.js:129583:16)
        at _no_name_provided__1_3.doResume_2_0_k$ (/Users/mbonnin/git/apollo-android/tests/IntrinsicsJs.kt:163:40)
        at _no_name_provided__1_3.doResume_0_k$ (/Users/mbonnin/git/apollo-android/tests/build/js/packages/tests-websockets-test/kotlin/tests-websockets-test.js:28878:17)
        at _no_name_provided__1_3.CoroutineImpl_0.resumeWith_jccoe6_k$ (/Users/mbonnin/git/apollo-android/tests/Standard.kt:55:40)
        at _no_name_provided__1_3.CoroutineImpl_0.resumeWith_bnunh2_k$ (/Users/mbonnin/git/apollo-android/tests/build/js/packages/tests-websockets-test/kotlin/tests-websockets-test.js:28763:17)
        at DispatchedContinuation.DispatchedTask.run_sv8swh_k$ (/Users/mbonnin/git/apollo-android/tests/Continuation.kt:45:5)
        at ScheduledMessageQueue.MessageQueue.process_sv8swh_k$ (/Users/mbonnin/git/apollo-android/tests/JSDispatcher.kt:148:25)
        at _no_name_provided__120.invoke_sv8swh_k$ (/Users/mbonnin/git/apollo-android/tests/JSDispatcher.kt:19:48)
        at /Users/mbonnin/git/apollo-android/tests/build/js/packages/tests-websockets-test/kotlin/tests-websockets-test.js:77223:9
        at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/KtorWebSocketEngine.kt
martinbonnin: I _think_ `ByteArray` is more likely than `ByteString`. `ByteString` being an okio thing, I don't see how it could be part of an `org.w3c` API. 

According to the [W3C spec](https://websockets.spec.whatwg.org//#feedback-from-the-protocol), `data` can be a `Blob` or an `ArrayBuffer`, not 100% sure how that translates to Kotlin

undermark5:
> there any chance ktor would handle both at the same time?

I can look into this, but looking at stuff yesterday, I was under the impression that ktor 1.6.7 did not support websockets on JS targets either. I built my client around ktor 2.0.0-beta-1, and I wasn't sure that bringing that in was ideal so I switched to vanilla JS (though I suppose we could keep these changes out until ktor 2.0.0 goes to public release or something). So unless that is what we want to do, I'll look into the support in 1.6.7 because I could be totally misunderstanding something, I never actually validated that was the case, just kinda under the impression that it was.

martinbonnin:
I'm not sure about using beta versions. I'd never do it for JVM code but JS is still overall experimental so it might be OK actually. If there's an easy way with 1.6.7, let's use it. If not, we can revisit 2.0.0-beta-1 and it's most likely going to be OK.

Another option would be to use something like https://github.com/heineiuo/isomorphic-ws, not sure how easy it is to pull it as a dependency.

undermark5:
Thanks, that would explain why it wasn't being treated as a module and showing that there were tests that could be run.

One other thing, when trying to book a trip in the sandbox, an internal server error occurs 
image

martinbonnin:
You'll need to set a HTTP header. This wasn't part of the link I sent for security reasons but you can get a new one by using the `login` mutation. Or just reuse mine: `Zm9vQGJhci5jb20=`

Screenshot 2022-03-09 at 19 03 43

undermark5:
So, I wasn't actually able to verify with the sandbox, not exactly sure why.

martinbonnin:
From the [Ktor docs](https://ktor.io/docs/websocket-client.html#example), all the `send()`/`receive()` happens from the `webSocketSession {}` block. My suspicion is that `webSocketSession {}` doesn't return until the socket is closed so it just hangs in there. Not sure how to fix this since the Apollo code expects individual calls. 

Maybe `open` could start a coroutine and return immediatly while the coroutine would run the `webSocketSession {}` block?

undermark5:
The sample on the docs shows the function `webSocket` not `webSocketSession`. The first gives the session in the block, but the second (the one the code uses) returns the `WebSocketSession` and only suspends to make the upgrade request, and the block is `HttpRequestBuilder.() -> Unit` and is used to construct the upgrade request.

I've verified that the code works on my box by publishing to maven local and depending on 3.1.1-SNAPSHOT in the project I'm working on (the one I made my basic client for). So given that plus the information above, I suspect something else is the issue, but I'm not certain what exactly. I suppose it is possible that since `webSocketSession {}` suspends until the socket is open that under Node ktor doesn't actually support websockets and fails to ever open and suspend indefinitely.

martinbonnin:
Ohh right, sorry I missed the `webSocket` vs `webSocketConnection` thing. Might be worth upgrading the ktor version to latest stable 1.6.7, just in case?

Ktor does [have Node WebSocket code](https://github.com/ktorio/ktor/blame/05d8386300e0d50955a8c9db98c4929781046eb7/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt#L61)

undermark5:
So, even with updating to 1.6.7, I'm still unable to verify the test locally, it is suspending on `collect` and not printing anything. The `StartOperation` seems to be sent successfully, but no incoming messages are received.

martinbonnin:
I took a quick look, it has to do with Ktor defaulting to [`ws://`](https://github.com/ktorio/ktor/blob/6763a7344059c1f9dfdb2f0816e98559862ef604/ktor-client/ktor-client-core/common/src/io/ktor/client/features/websocket/builders.kt#L48) for the protocol instead of `wss`. I could make it go one step further with https://github.com/martinbonnin/apollo-android/commit/746e8d296b8c1d3d6f536f4150baf522b4af3177.

But now it fails with 

```
WebSocketException: {"target":{"_events":{},"_eventsCount":2,"readyState":2,"protocol":"","_binaryType":"nodebuffer","_closeFrameReceived":false,"_closeFrameSent":false,"_closeMessage":"","_closeTimer":null,"_closeCode":1006,"_extensions":{},"_receiver":null,"_sender":null,"_socket":null,"_isServer":false,"_redirects":0,"url":"wss://apollo-fullstack-tutorial.herokuapp.com/graphql","_req":null},"type":"error","message":"Server sent a subprotocol but none was requested","error":{}}
```

undermark5:
Alright, so after further digging, it turns out that either ktor or websocket/ws is potentially incorrectly configuring the websocket in this instance. The issue appears to lie in the fact that we set a subprotocol header, and the subprotocol header is ignored by websocket/ws when determining if the client requested a subprotocol. So, I think we've got a few options, scrap using ktor and instead look into what it would take to bring in [isomorphic-ws](https://github.com/heineiuo/isomorphic-ws), or do what ktor is doing, but in such a way that websocket/ws actually recognizes the subprotocol set in the header as being requested (possibly filing issues with either ktor or websocket/ws depending on who thinks the other is to blame).

Ultimately, I think the level of effort between those two is going to be pretty much the same as both will require adapters for websocket/js.

undermark5:
commit [ba7319a](https://github.com/apollographql/apollo-kotlin/pull/3913/commits/ba7319a8f79bf5d9b6aad8670830926615db055b) should now pass the test. If anything of the copyright notice associated with the code pulled from ktor needs to change let me know (still new to this).

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt
```diff
@@ -0,0 +1,117 @@
...
+   * The following applies for lines 71-114
+   * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+   * Some lines have been added/modified from the original source (https://github.com/ktorio/ktor/blob/f723638afd4d36024c390c5b79108b53ab513943/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt#L62)
+   */
+  // Adding "_capturingHack" to reduce chances of JS IR backend to rename variable,
+  // so it can be accessed inside js("") function
+  @Suppress("UNUSED_PARAMETER", "UnsafeCastFromDynamic", "UNUSED_VARIABLE", "LocalVariableName")
+  private fun createWebSocket(urlString_capturingHack: String, headers: Headers): WebSocket =
+      if (PlatformUtils.IS_NODE) {
+        val ws_capturingHack = js("eval('require')('ws')")
```
martinbonnin, line 76: Note that this only works because we use Ktor for the HTTP engine and Ktor imports `ws`. I'm fine with that but maybe leave a comment?

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+    }
+  }
+
+  /*
+   * The following applies for lines 71-114
+   * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+   * Some lines have been added/modified from the original source (https://github.com/ktorio/ktor/blob/f723638afd4d36024c390c5b79108b53ab513943/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt#L62)
+   */
```
martinbonnin, line 70: LGTM. Maybe mention why the lines where modified?

```suggestion
  /*
   * The following applies for lines below
   * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
   * Some lines have been added/modified from the original source (https://github.com/ktorio/ktor/blob/f723638afd4d36024c390c5b79108b53ab513943/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt#L62)
   * in order to fix an issue with subprotocols on Node (https://youtrack.jetbrains.com/issue/KTOR-4001)
   */
```

apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt
undermark5: I can either add a comment or add `ws` as a dependency, whichever way make more sense. Adding the dependency will make this independent of ktor being used for the HTTP engine.

martinbonnin: That works too !

martinbonnin (approved):

BoD (approved):

martinbonnin:
Thanks for the contribution 🟣 !

PR #3919, opened 2022-03-11
🗄️ add apolloClient.httpCache

martinbonnin:
This allows things like 

```kotlin
apolloClient.httpCache.clearAll()
```

Thanks @StylianosGakis for the heads up!

### DIFF
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
@@ -24,7 +24,7 @@ class CachingHttpInterceptor(
 ) : HttpInterceptor {
   private val lruHttpCache = DiskLruHttpCache(fileSystem, directory, maxSize)
 
-  val cache: HttpCache = lruHttpCache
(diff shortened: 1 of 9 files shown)


apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -128,12 +131,14 @@ class CachingHttpInterceptor(
     return response
   }
 
+  @Deprecated("Use store.clearAll() instead", ReplaceWith("store.clearAll()"))
```
BoD, line 134: Nitpick: let's add the `@ApolloDeprecatedSince` annotation to keep track of these (`v3_2_0` needs to be added to the enum).

BoD (approved):
👍

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
martinbonnin: Good call 👍 . I went with `v3_1_1` so we're not stuck on the versioning scheme and may decided to ship either `3.1.1` or `3.2.0`

PR #3921, opened 2022-03-14
Defer: AdapterContext

BoD:
In this PR:
- move `variables` to a dedicated `AdapterContext` class (carried over by `CustomScalarAdapters`)
- add `AdapterContext.mergedDeferredFragmentIds` to keep track of which fragments are to be parsed
- ~~add `AdapterContext.currentPath` to keep track of the current path being parsed~~  <- let's use JsonReader.getPath()
  - ~~and make `ListAdapter` and generated Adapter code (Kotlin only for now) update current path~~

(This targets the draft PR with the initial work: `DeferredJsonMerger`, etc.)


apollo-api/api/apollo-api.api
```diff
@@ -329,25 +343,25 @@ public final class com/apollographql/apollo3/api/CompiledVariable {
...
+	public synthetic fun <init> (Ljava/util/Map;Lcom/apollographql/apollo3/api/AdapterContext;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun fold (Ljava/lang/Object;Lkotlin/jvm/functions/Function2;)Ljava/lang/Object;
 	public fun get (Lcom/apollographql/apollo3/api/ExecutionContext$Key;)Lcom/apollographql/apollo3/api/ExecutionContext$Element;
+	public final fun getAdapterContext ()Lcom/apollographql/apollo3/api/AdapterContext;
 	public fun getKey ()Lcom/apollographql/apollo3/api/ExecutionContext$Key;
 	public fun minusKey (Lcom/apollographql/apollo3/api/ExecutionContext$Key;)Lcom/apollographql/apollo3/api/ExecutionContext;
 	public final fun newBuilder ()Lcom/apollographql/apollo3/api/CustomScalarAdapters$Builder;
 	public fun plus (Lcom/apollographql/apollo3/api/ExecutionContext;)Lcom/apollographql/apollo3/api/ExecutionContext;
 	public final fun responseAdapterFor (Lcom/apollographql/apollo3/api/CustomScalarType;)Lcom/apollographql/apollo3/api/Adapter;
-	public final fun variables ()Ljava/util/Set;
```
BoD, line 340: Technically a breaking change - but do we expect it to impact anyone?

martinbonnin: Yes, it's binary breaking... Meaning no one consumer source code is going to be affected but if a transitive dependency is running < 3.1.1 it will crash at runtime

BoD: You're right! Will keep / mark deprecated.

PR #3926, opened 2022-03-15
🔄  JS Websocket follow up

martinbonnin:
Small follow up from https://github.com/apollographql/apollo-kotlin/pull/3913:

Give more time to JS tests and dispose the apolloClient

### DIFF
--- a/tests/websockets/build.gradle.kts
+++ b/tests/websockets/build.gradle.kts
@@ -6,6 +6,17 @@ plugins {
 configureMppTestsDefaults()
 
 kotlin {
+  js {
+    nodejs {
+      testTask {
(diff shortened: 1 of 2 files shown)


BoD (approved):
👍

martinbonnin:
The yellow card was for the Javascript language but felt a bit soccer-y 😅 . Replaced it with a 🔄  for bidirectional websockets :)

BoD:
Also opened #3927 with the mention about WS not being available on JS removed 🎉

martinbonnin:
Looks like the Netlify docs aren't super happy about this PR. @trevorblades is this something we should worry about?

Issue #3924, opened 2022-03-15
MapTestNetworkTransport doesn't work with zero-arg queries


BenTilbrook:
**Summary**
`MapTestNetworkTransport` doesn't work with zero-arg queries. This seems to be because the generated code is not a `data class`, unlike other queries/mutations, and therefore has no `equals/hashCode`, which is required by `MapTestNetworkTransport` because it uses the `Operation` as a `Map` key.

**Version**
Apollo Kotlin version: `3.1.0`

**Description**
See attached modified tutorial project. Run `./gradlew testDebugUnitTest`. It will fail.

[apollo-android-tutorial-map-transport-bug.zip](https://github.com/apollographql/apollo-kotlin/files/8249612/apollo-android-tutorial-map-transport-bug.zip)

martinbonnin:
In all cases, I think it makes sense to make the operations data classes always, it keeps everything consistent and predictable.

BoD:
> I'm wondering if we could use the Operation's id instead

Actually that wouldn't be enough as we need to distinguish same operation with different parameters. 

> it makes sense to make the operations data classes always

But `Data class must have at least one primary constructor parameter` 😅

BenTilbrook:
Could make a zero arg operation an object. It's singleton status would make it a suitable map key.

martinbonnin:
> But Data class must have at least one primary constructor parameter

Right!

> Could make a zero arg operation an object. It's singleton status would make it a suitable map key.

True but that'd make callsites less consistent... It'd be very weird for newcomers to have to instantiate some operations but not others.

martinbonnin:
> Actually that wouldn't be enough as we need to distinguish same operation with different parameters.

What about we implement `hashCode/equals` with `operation.id` for operations without variables? (and leave the other as regular data classes)

BoD:
> What about we implement hashCode/equals with operation.id for operations without variables? (and leave the other as regular data classes)

Good idea! 👍 I'll have a look!

> Or just based on the instance so that it works for Fragments too that don't have an id?

That's equivalent to the default equals/hashcode, right?

martinbonnin:
> That's equivalent to the default equals/hashcode, right?

Sorry, I meant based on the **class**, not instance 🤦

BoD:
> based on the class, not instance
> 2 operations with the same class and no variables should be equal?

Ooh yeah, makes sense!

martinbonnin:
Closed in `3.2.0`

PR #3928, opened 2022-03-15
Add equals and hashCode implementations when a data class cannot be used

BoD:
Related to #3924


tests/models-response-based/src/commonTest/kotlin/test/BasicTest.kt
```diff
@@ -105,4 +106,12 @@ class BasicTest {
     }
     assertEquals(name, "R2-D2")
   }
+
+  @Test
+  fun equalsAndHashCodeAreCorrect() = runTest {
```
martinbonnin, line 111: FWIW, `model-xyz-based` tests should ideally be only be about differences between codegens. 

Since this doesn't depend on the codegen, I was going to say to move it to `integration-tests` but that won't work either because `integration-tests` are also run in Java. So long story short LGTM 👍. (And life is hard 😅)

martinbonnin (approved):

tests/models-response-based/src/commonTest/kotlin/test/BasicTest.kt
martinbonnin: `integration-tests` is:

* JS (with Java codegen)
* Native (with Kotlin codegen)
* JVM (with Kotlin codegen and Java codegen)

See [here](https://github.com/apollographql/apollo-kotlin/blob/09f4a04a9609b609d9c011b267a7b767309fe244/tests/integration-tests/build.gradle.kts#L10) for the setup.

The nice thing is that it makes sure the same callsites work exactly the same with both Kotlin and Java models 😎 

The only thing that doesn't work is default arguments and named arguments because they do not exist in Java. For this we have a specific module named [`kotlin-codegen`](https://github.com/apollographql/apollo-kotlin/tree/09e654447c87dd8c4177b796c590878ce8ee380d/tests/kotlin-codegen)

martinbonnin: If it works in Java, it might be worth moving the test to `integration-test` then, increases the test coverage and makes more sense from a separation of concernes point of view

martinbonnin (approved):

PR #3930, opened 2022-03-15
⚙️ Make adapters code work without relying on having a` __typename` IrProperty

martinbonnin:
First step towards https://github.com/apollographql/apollo-kotlin/issues/3774

This changes the adapter code so that it doesn't rely on models having a `__typename` IrProperty. Instead, the operationBased models can get the typename from `JsonReader.readTypename`. This makes the adapters robust to queries like this:

```graphql
{
  animal {
   # __typename is present at the root of the field
    __typename
    ... on Animal {
      # Here we need typename but it's not a property of the OnAnimal model 
      # It's ok because the adapter will buffer and get the typename from the json
      ... on Cat {
        meoooww
      }
    }
  }
}
```

This is mostly a refactoring. No API is changed besides some `RESPONSE_NAMES` properties deep down the adapters that shouldn't be used in user code.


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -109,7 +109,7 @@ fun <T : Any> BooleanExpression<T>.evaluate(block: (T) -> Boolean): Boolean {
   }
 }
 
-fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String): Boolean {
+fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?): Boolean {
```
martinbonnin, line 112: The `typename` argument becomes nullable because we might want to call it in cases where we don't have a typename. For an example:

```graphql
query GetAnimal($withDetails: Boolean!) {
  animal {
    ... on Animal @include(if: $withDetails) {
      species
    }
  }
}
```

The code to evaluate the condition will be like:

```kotlin
variable("withDetails").evaluate(customScalarAdapters.variables(), null)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/Ir.kt
```diff
@@ -169,21 +170,26 @@ data class IrModel(
...
- * @param hidden allows to hide a property from the model but still have it part of the selections.
- * This is used for typename in compat models because the adapters need to read __typename
+ * @param requiresBuffering true if this property contains synthetic properties and needs to be buffered
  */
 data class IrProperty(
     val info: IrFieldInfo,
     val override: Boolean,
     val condition: BooleanExpression<BTerm>,
-    val requiresBuffering: Boolean,
-    val hidden: Boolean,
```
martinbonnin, line 181: Nice side effect of not relying on `IrProperty` to get the __typename in adapters, we don't need this `hidden` hack to hide `__typename` from the compat `Fragments` model

martinbonnin:
Sorry, bad rebase 😬 . Should be alright now

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -135,3 +135,14 @@ data class BPossibleTypes(val possibleTypes: Set<String>) : BTerm() {
   constructor(vararg types: String): this(types.toSet())
 }
 
+
+fun <T : Any> BooleanExpression<T>.containsPossibleTypes(): Boolean {
+  return when (this) {
+    BooleanExpression.True -> false
+    BooleanExpression.False -> false
+    is BooleanExpression.Not -> !operand.containsPossibleTypes()
```
BoD, line 143: Should it be `operand.containsPossibleTypes()` 🤔 To check that a type is _not_ in the list you need the type?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/Ir.kt
```diff
@@ -169,21 +170,26 @@ data class IrModel(
...
-  // synthetic properties are special as we need to rewind the reader before reading them
+  /**
+   * synthetic properties are special as we need to rewind the reader before reading them
+   * They are read in a second pass and are not real json names
+   */
   val isSynthetic: Boolean
     get() = info.gqlType == null
 
+  /**
+   *
```
BoD, line 189: Nitpick: empty doc :)

BoD (approved):
💪

PR #3931, opened 2022-03-15
Remove files from old Gatsby docs theme (v2)

jgarrow:


jgarrow:
@trevorblades Removing the `netlify.toml` will also mean that the `docs/generate-docs.sh` file will never get run. Do we need to add an extra config somewhere in the `docs` repo to account for this?

trevorblades:
Good question @jgarrow. That script takes the repo readme and writes it as the `index.md` file in the docs. My hunch is that we'll be fine because the `index.md` file that it generates is already checked into git, and this is an older version of the software, so perhaps its readme isn't going to get updated very much going forward. If it does, we can always manually run `docs/generate-docs.sh` and check the result in to git.

@martinbonnin what are your thoughts?

martinbonnin:
Sounds fine 👍 . We had this script to avoid some duplication between the repo README.md and the docs index.md but `release-2.x` isn't going to move much going forward so it's fine to edit in 2 places if needed.

martinbonnin (approved):

PR #3934, opened 2022-03-16
Add getPath() and refact MapJsonReader

BoD:
Note: my first approach was to only use `dataStack`/`indexStack` but couldn't manage to get the same exact behavior as `BufferedSourceJsonReader`. In the end, managing a specific `path` array did the trick.

### DIFF
--- a/apollo-api/src/commonTest/kotlin/test/PathTest.kt
+++ b/apollo-api/src/commonTest/kotlin/test/PathTest.kt
@@ -1,0 +1,217 @@
+package test
+
+import com.apollographql.apollo3.api.json.BufferedSourceJsonReader
+import com.apollographql.apollo3.api.json.JsonReader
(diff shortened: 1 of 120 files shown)


apollo-api/src/commonTest/kotlin/test/PathTest.kt
```diff
@@ -0,0 +1,217 @@
...
+        END_ARRAY,
+        "menu.popup.arrayList",
+        END_OBJECT,
+        "menu.popup",
+        END_OBJECT,
+        "menu",
+        END_OBJECT,
+        "",
+    )
+    assertEquals(expected, tokensAndPaths)
```
martinbonnin, line 203: Awesome test 😃

BoD:
Also contains the refact from #3935

BoD:
Closing because this was incorporated in #3941

PR #3935, opened 2022-03-16
Try to reduce allocations in MapJsonReader

martinbonnin:


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -23,285 +24,298 @@ import com.apollographql.apollo3.api.json.internal.toLongExact
...
-    is Map<*, *> -> JsonReader.Token.BEGIN_OBJECT
-    is Int -> JsonReader.Token.NUMBER
-    is Long -> JsonReader.Token.NUMBER
-    is Double -> JsonReader.Token.NUMBER
-    is String -> JsonReader.Token.STRING
-    is Boolean -> JsonReader.Token.BOOLEAN
-    else -> error("Unsupported value $any")
+  override fun endObject() = apply {
+    if (peek() != JsonReader.Token.END_OBJECT) {
+      throw JsonDataException("Expected END_OBJECT but was ${peek()} at path ${getPath()}")
```
martinbonnin, line 152: This is more strict than the previous implementation and will fail on consumers that do not consume all tokens until the end of the object (like BufferedSourceJsonReader)

apollo-compiler/src/test/graphql/com/example/deprecated_merged_field/java/operationBased/deprecated_merged_field/adapter/CatQuery_ResponseAdapter.java.expected
```diff
@@ -77,9 +77,9 @@ public class CatQuery_ResponseAdapter {
       }
 
       Assertions.checkFieldNotMissing(__typename, "__typename");
-      reader.rewind();
       AnimalFragment animalFragment = null;
       if (BooleanExpressions.evaluate(new BooleanExpression.Element<BTerm>(new BPossibleTypes("Cat","Dog")), Collections.emptySet(), __typename)) {
+        reader.rewind();
```
martinbonnin, line 82: Because `endObject()` is now more strict, we need to make sure to rewind() only if really required

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -23,285 +24,330 @@ import com.apollographql.apollo3.api.json.internal.toLongExact
...
 class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
-  // Like a Map but easier to access sequentially
-  // Could be optimized
-  private class OrderedMap(val entries: List<Entry>)
-  private class Entry(val key: String, val value: Any?)
 
-  private val dataStack = arrayOfNulls<Any>(MAX_STACK_SIZE)
-  private val indexStack = IntArray(MAX_STACK_SIZE)
-  private val nameIndexStack = IntArray(MAX_STACK_SIZE)
+  private var peekedToken: JsonReader.Token = JsonReader.Token.END_OBJECT
```
BoD, line 31: It could be set to `JsonReader.Token.BEGIN_OBJECT` here and the assignment in `init` be removed?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -23,285 +24,330 @@ import com.apollographql.apollo3.api.json.internal.toLongExact
...
-    // Allow garbage collection
-    dataStack[stackSize] = null
-    stackSize--
-    pendingName = null
+  private fun anyToToken(any: Any?) = when (any) {
+    null -> JsonReader.Token.NULL
+    is List<*> -> JsonReader.Token.BEGIN_ARRAY
+    is Map<*, *> -> JsonReader.Token.BEGIN_OBJECT
+    is Int -> JsonReader.Token.NUMBER
+    is Long -> JsonReader.Token.NUMBER
```
BoD, line 67: Would it make sense to return `JsonReader.Token.LONG` for this one?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -23,285 +24,330 @@ import com.apollographql.apollo3.api.json.internal.toLongExact
...
-    pop()
-    rewindPath()
-    incrementPathIndex()
+    if (peek() != JsonReader.Token.END_ARRAY) {
+      throw JsonDataException("Expected END_ARRAY but was ${peek()} at path ${getPath()}")
+    }
+    stackSize--
+    iteratorStack[stackSize] = null // allow garbage collection
+    path[stackSize] = null // allow garbage collection
+    advanceIterator()
```
BoD, line 133: Should we also nullify `containerStack[staclSize]`?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -23,285 +24,330 @@ import com.apollographql.apollo3.api.json.internal.toLongExact
...
-      is List<*> -> {
-        indexStack[stackSize - 1] < data.size
-      }
-      is OrderedMap -> {
-        indexStack[stackSize - 1] < data.entries.size
-      }
-      else -> error("")
+    return when (peek()) {
+      JsonReader.Token.END_OBJECT -> false
+      JsonReader.Token.END_ARRAY -> false
```
BoD, line 164: I guess `JsonReader.Token.END_DOCUMENT -> false` too?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -311,25 +357,30 @@ class MapJsonReader(val root: Map<String, Any?>) : JsonReader {
...
+    return buildString {
+      for (index in 0.until(stackSize)) {
+        val element = path[index]
+        if (element is String) {
+          if (index > 0) {
+            append('.')
+          }
+          append(element)
+        } else if (element is Int) {
+          append("[$element]")
```
BoD, line 377:
```suggestion
          append(".$element")
```

To be the same format as `BufferedSourceJsonReader`

BoD: Also it looks like there's a +1 in there somewhere, with `PathTest` we can see the indices are 1 too high 😓

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
martinbonnin: Looks like it 👍 . I must admit I'm not 100% clear the implications of this as `nextInt()`, `nextDouble()`, etc.. do some coercion. But it's certainly better to hint at long there.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
martinbonnin: Aaarg, I've run the wrong test... Ok let me double check that.

martinbonnin: I've also just realized that the current `getPath()` as a String is pretty bad for `@defer` because these 2 things cannot be distinguished:

```json
{
  "0": true
}
```

and 

```
[
  true
]

```

BoD: But "0" is not a valid GraphQL field name so it's ok, right?

BoD (approved):
💪

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
martinbonnin: I guess so...

PR #3939, opened 2022-03-17
⚙️  Introduce addTypename Gradle parameter

martinbonnin:
Introduce `addTypename`:

```kotlin
apollo {
  /**
   * Add '__typename' for every selection set that contains fragments (inline or named) (Default)
   * 
   * This is adding a lot more '__typename' than the above 'ifPolymorphic" and will be removed in 
   * a future version. If you require '__typename' explicitely, you should add it to your queries.
   **/
  addTypename.set("ifFragments")

  /**
   * Add '__typename' for polymorphic fields, i.e. fields that contains a subfragment 
   * (inline or named) whose type condition isn't a super type of the field type.
   * If a field is monomorphic, no '__typename' will be added.
   * 
   * Note: It also adds '__typename' on fragment definitions that satisfy the same property because fragments
   * could be read from the cache and we don't have a containing field in that case.
   **/
  addTypename.set("ifPolymorphic")

  /**
   * Add '__typename' for abstract fields, i.e. fields that are of union or interface type
   * 
   * Note: It also adds '__typename' on fragment definitions that satisfy the same property because fragments
   * could be read from the cache and we don't have a containing field in that case.
   **/
  addTypename.set("ifAbstract")
}
```

`"ifAbstract"` is certainly a good default to go with because it is simple enough and will work well with the cache but it's aalso a significant change so it'll be changed down the road depending the feedback.


martinbonnin:
(sorry for the extra notification, I pressed [enter] too fast 🤦 . Title and original description are now updated)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/adapter/AdapterCommon.kt
```diff
@@ -103,9 +109,19 @@ internal fun readFromResponseCodeBlock(
...
+    CodeBlock.builder()
+        .apply {
+          if (regularProperties.none { it.info.responseName == "__typename" }) {
+            // We are in a nested fragment that needs access to __typename, get it from the buffered reader
+            add("$reader.rewind();\n")
+            add(javaTypenameFromReaderCodeBlock())
+          } else {
+            add("$T.checkFieldNotMissing($__typename, $S);", JavaClassNames.Assertions, __typename)
+          }
+        }.build()
```
martinbonnin, line 124: These changes are leftovers from https://github.com/apollographql/apollo-kotlin/pull/3930 that were not added to the Java codegen

BoD (approved):
💪💪💪

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -22,6 +23,10 @@ const val MODELS_OPERATION_BASED = "operationBased"
 @ApolloDeprecatedSince(v3_0_0)
 const val MODELS_COMPAT = "compat"
 
+const val ADD_TYPENAME_IF_FRAGMENTS = "ifFragments"
+const val ADD_TYPENAME_IF_POLYMORPHIC = "ifPolymorphic"
+const val ADD_TYPENAME_IF_ABSTRACT = "ifAbstract"
```
martinbonnin, line 28: Note: we do not check the API on apollo-compiler so this didn't appear in the api dump. Something we could do at some point

martinbonnin:
Just pushed a few more changes. I backtracked on the idea of tunneling some information through the `_apolloHasTypename` internal directive. It turns out it was only used by operationBased codegen and we can actually do better:

```
{
  # animal is polymorphic
  animal {
    # __typename needs to be added here
    ... on Cat {
      meow
    }
    ... on Node {
      # But this is always true and we don't need to check __typename here
      ... on Animal {
        # this is also always true and we don't need to check __typename here
        # even if not all Nodes are Animals
      }
    }
  }
}
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -435,6 +435,13 @@ interface Service {
    */
   val codegenModels: Property<String>
 
+  /**
+   * When to add __typename. One of "ifFragments", "ifAbstract" or "ifPolymorphic"
+   *
+   * Default value: "ifAbstract"
```
BoD, line 441:
```suggestion
   * Default value: "ifFragments"
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultService.kt
```diff
@@ -117,6 +118,10 @@ abstract class DefaultService @Inject constructor(val project: Project, override
   override fun useVersion2Compat(rootPackageName: String?) {
     packageNamesFromFilePaths(rootPackageName)
     codegenModels.set(MODELS_COMPAT)
+    // This isn't really exact. We could add a 3rd mode that matches 1:1 what 2.x is doing but
+    // ADD_TYPENAME_IF_FRAGMENTS is closer than ADD_TYPENAME_IF_POLYMORPHIC
+    addTypename.set(ADD_TYPENAME_IF_FRAGMENTS)
+
```
BoD, line 124: No longer necessary if it's the default now.
```suggestion
```

apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
BoD: Maybe we can copy the nice explanation of each value you put in the description of this PR here?

BoD (approved):

PR #3942, opened 2022-03-18
Optimize CI a bit

Goooler:
Mainly for using gradle-build-action to speed up build on CI, and bump action versions by the way.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ jobs:
   tests-gradle:
     runs-on: macos-11
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: set up JDK 1.8
         uses: actions/setup-java@v1
         with:
@@ -39,7 +39,7 @@ jobs:
   tests-no-gradle:
     runs-on: macos-11
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: set up JDK 1.8
         uses: actions/setup-java@v1
         with:
@@ -63,7 +63,7 @@ jobs:
   tests-integration:
     runs-on: macos-11
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: set up JDK 1.8
         uses: actions/setup-java@v1
         with:
@@ -87,7 +87,7 @@ jobs:
   tests-java9:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: set up JDK 11
         uses: actions/setup-java@v2
         with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: macos-11
     steps:
       - uses: actions/checkout@v3
-      - name: set up JDK 1.8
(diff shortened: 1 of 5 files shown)


.github/workflows/pr.yml
```diff
@@ -15,11 +15,12 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - name: set up JDK 1.8
-        uses: actions/setup-java@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: 'temurin'
+          java-version: 8
+      - uses: gradle/gradle-build-action@v2
```
Goooler, line 23: https://github.com/gradle/gradle-build-action

Ref https://github.com/detekt/detekt/pull/3723, it's really so fast to run jobs, and many other advantages.

.github/workflows/push.yml
```diff
@@ -5,15 +5,16 @@ on:
     branches: [ main ]
 
 jobs:
-  push:
+  publish:
     runs-on: macos-11
+    if: github.repository == 'apollographql/apollo-kotlin'
```
Goooler, line 10: See https://github.com/square/kotlinpoet/pull/1119.

martinbonnin (approved):
LGTM, thanks!

.github/workflows/push.yml
BoD: This could also be added to the `publish` in `tag.yml` too? (This one is for the SNAPSHOTS, whereas the one in `tag` publishes releases).

Goooler: 8fa5fe58f8a99f444b79ae14433e1220e3c17f12

.github/workflows/pr.yml
```diff
@@ -31,19 +32,21 @@ jobs:
...
-      - uses: actions/upload-artifact@v2.2.4
+      - uses: actions/upload-artifact@v3
         if: always()
         with:
           name: tests-gradle.zip
           path: diagnostics.zip
+
   tests-no-gradle:
     runs-on: macos-11
     steps:
```
Goooler, line 43: And if we should run CI on ubuntu? It should be faster.

.github/workflows/push.yml
BoD: (just added a small commit to update the job name referenced in a doc)

CONTRIBUTING.md
```diff
@@ -161,13 +161,11 @@ following events:
...
 ### On pushes to `main` branch
 
-_(Will be replaced to "on pushes to `main` branch" after v3 is released.)_
-
 **Workflow:** [`push.yml`](https://github.com/apollographql/apollo-android/blob/main/.github/workflows/push.yml)
 
 **Job:**
 
-- `push`
+- `publish`
```
Goooler, line 168: I've just renamed this job to `deploy`, should we keep `deploy`?

.github/workflows/pr.yml
martinbonnin: I wish we could do that. Unfortunatly, some of the Gradle tests run multiplatform projects that include Apple targets. We could split them but I don't think it's worth the trouble for now.

CONTRIBUTING.md
BoD: Oh haha 😅. Well it doesn't matter too much, as long as the name is the same here.

BoD (approved):
🙏

PR #3943, opened 2022-03-21
Fix JamesIves/github-pages-deploy-action version

BoD:
Following the merge of #3942

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -40,7 +40,7 @@ jobs:
           name: push.zip
           path: push.zip
       - name: Deploy Kdoc to github pages
-        uses: JamesIves/github-pages-deploy-action@4.2.5
+        uses: JamesIves/github-pages-deploy-action@v4
         with:
           branch: gh-pages # The branch the action should deploy to.
           folder: build/dokkaHtml # The folder the action should deploy.
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '11'
-      - uses: gradle/gradle-build-action@v2
+      - uses: gradle/gradle-build-action@0d13054264b0bb894ded474f08ebb30921341cee #v2.1.4
       - name: Build with Gradle
         run: |
           ulimit -c unlimited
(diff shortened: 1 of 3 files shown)


martinbonnin (approved):

.github/workflows/push.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           name: push.zip
           path: push.zip
       - name: Deploy Kdoc to github pages
-        uses: JamesIves/github-pages-deploy-action@4
+        uses: JamesIves/github-pages-deploy-action@4.2.5
         with:
```
Goooler, line 44: Sorry for the mistake, it might be better to use `@v4`?

BoD: Woops indeed  a `v` is needed here, right? Sorry I'm not super familiar with Github actions, does`v4` mean use the latest from 4.x.y?

Goooler: yep

martinbonnin: That seems a recipe for the kind of supply chain attacks that the npm ecosystem is struggling with at the moment? I'd rather keep the version fixed even if it means updating it manually. At least  we keep some control...

It won't solve anything because Github will certainly allow tags to be re-created (which is a bit scary now that I think about it...). But at least it won't make it any worse?

BoD: Sounds good to me 👍

martinbonnin: Actually, we should pin versions to their sha1 (see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). 

@BoD want to take a look as part of this PR?

BoD: Sure! I'll have a 👀.

martinbonnin (approved):
🔒  👍

.github/workflows/push.yml
BoD: Used sha1 for `gradle` and `JamesIves` and kept the tag names for `actions` as I figured we can trust that? LMK what you think.

martinbonnin: Works for me. If Github goes rogue, I guess we'll have bigger problems...

PR #3947, opened 2022-03-21
Defer: add a BDefer boolean expression element

BoD:
In this PR:
- add a `BDefer` boolean expression element
- add a `toIncludeAndDeferBooleanExpression` fun that returns `BDefer` in addition to `BVariable` when `@defer` is present
- use that to disambiguate fragment synthetic fields
- update `evaluate` to take an `AdapterContext` and a JSON path and use that for `BDefer` expressions

Note: added a test to see how the generated code looks like (need `JsonReader.getPath()` from #3934 for the code to compile)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -125,14 +132,21 @@ sealed class BTerm
 
 /**
  * A term that comes from @include/@skip directives and that needs to be matched against operation variables
+ * TODO rename to BInclude
```
BoD, line 135: Should we? Not 100% sure of the naming to be honest but "variable" sounds off to me.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
```diff
@@ -584,10 +588,12 @@ internal fun GQLValue.toIrValue(): IrValue {
  * - False
  * - (!)Variable
  * - (!)Variable & (!)Variable
+ *
+ * TODO: rename to toIncludeBooleanExpression
```
BoD, line 592: Same here, naming is hard :) But need to distinguish this one that is about include/skip and the other one which is this one + defer.

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt
```diff
@@ -191,7 +191,7 @@ class CodegenTest {
                     Parameters(file, MODELS_COMPAT, true)
                 )
 
-                if (file.name in listOf("inline_fragment_with_include_directive", "fragment_spread_with_include_directive")) {
+                if (file.name in listOf("inline_fragment_with_include_directive", "fragment_spread_with_include_directive", "fragments_with_defer_and_include_directives")) {
                   // These do not support responseBased models because of include directives on fragments
```
martinbonnin, line 195:
```suggestion
                  // These do not support responseBased models because of include or defer directives on fragments
```

apollo-compiler/src/test/graphql/com/example/fragments_with_defer_and_include_directives/TestOperation.graphql
```diff
@@ -0,0 +1,71 @@
...
+      name
+    }
+    ... on Droid @defer(label: "second") {
+      id
+    }
+    ... on Droid @defer(label: "third") @include(if: $cond3) {
+      name
+    }
+    ... on Droid @defer(label: "fourth") @skip(if: $cond4) {
+      id
```
martinbonnin, line 68: Can we have a test case with a conditional defer ? `@defer(if: $someVar)`

BoD: Yes good idea! 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
martinbonnin: Ideally we can "uniformize" that. Something like this?

1. `@include(if: $var1)` = BVariable("var1")
2. `@skip(if: $var1)` = !BVariable("var1")
3. `@defer(label: $lbl1)` = !BLabel("lbl1")
4. `@defer(label: $lbl1, if: $var1)` = BLabel("lbl1") || !BVariable("var1")

BoD: 👍 Yes it makes sense! Done in e2e26ce772f89b80f055efc235ef86165acb663d.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -109,10 +110,11 @@ fun <T : Any> BooleanExpression<T>.evaluate(block: (T) -> Boolean): Boolean {
   }
 }
 
-fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?): Boolean {
+fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?, adapterContext: AdapterContext, path: String): Boolean {
```
martinbonnin, line 113: I missed that the `path` was part of the fragment identifier. What about replacing `DeferredFragmentIdentifier` with plain strings? And `BLabel` with `BFragmentId`. This way we can have 

```kotlin
fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?, fragmentIds: Set<String>): Boolean {
  return evaluate {
    when (it) {
      is BVariable -> variables.contains(it.name)
      is BFragmentId -> fragmentIds.contains(it.id)
      is BPossibleTypes -> it.possibleTypes.contains(typename)
    }
  }
}
```

And `BooleanExpression` remains relatively standalone.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/AdapterCommon.kt
```diff
@@ -122,7 +122,7 @@ internal fun readFromResponseCodeBlock(
             } else {
               "null"
             }
-            beginControlFlow("if·(%L.%M($customScalarAdapters.adapterContext.variables(),·$typenameLiteral))", property.condition.codeBlock(), evaluate)
+            beginControlFlow("if·(%L.%M($customScalarAdapters.adapterContext.variables(),·$typenameLiteral,·$customScalarAdapters.adapterContext,·$reader.getPath()))", property.condition.codeBlock(), evaluate)
```
martinbonnin, line 125: With the above comment, we'd ideally have 

```
$customScalarAdapters.adapterContext.variables(),·$typenameLiteral,·$customScalarAdapters.adapterContext.fragmentIds()
```

And it's all nicely symmetrical? I don't think we need `getPath()` here? Better use it when merging the json?

BoD: We need the path here because that's part of the id of a received fragment. For instance if we're about to read the fields for `ComputerFields` and we are at `computers.2`, we need to know if we received it. We store the ids of the merged fragments at merging time.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
BoD: > the path was part of the fragment identifier

yes but it's the current path that you can only get while parsing the JSON - that is why we still need to get it from `JsonReader.getPath` and pass it to this method. Unless there's a way around it that I'm missing?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/AdapterCommon.kt
martinbonnin: I thinnk the part I'm missing is when [`AdapterContext.deferredFragmentLabels()`](https://github.com/apollographql/apollo-kotlin/blob/a73c9c06037c8075f41555917e95a3489b4d6445/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt#L31) is going to be called. 

My understanding would be that it would be setup at the beginning of the parsing. Something like (pseudo code):

```kotlin
adapterContext = adapterContext.newBuilder().deferredFragmentLabels(deferredJsonMerger.mergedFragmentIds)).build()

customScalarAdapters = customScalarAdapters.newBuilder().adapterContext(adapterContext).build()

adapter.fromJson(MapJsonReader(deferredJsonMerger.merge), customScalarAdapters)

```

Then you don't the path in the individual "fromJson" since you have a "Map" with additional data (all the fragment ids present in the map)?

martinbonnin: > yes but it's the current path that you can only get while parsing the JSON - that is why we still need to get it from JsonReader.getPath and pass it to this method. Unless there's a way around it that I'm missing?

To answer this question, I think this fragmentId could be a compile time constant encoded in the BooleanExpression?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
martinbonnin: Let's continue the discussion in a single thread [below](https://github.com/apollographql/apollo-kotlin/pull/3947#discussion_r832013147)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/AdapterCommon.kt
BoD: I think the reason we can't do it without a "dynamic" path is because of arrays.

Remember, we're going to call `fromJson` in a loop (once per payload), and pass a json that contains more fragments each time. 
Let's re-use this example:

**Received payloads:**

```json
// Payload 1
{"data":{"computers":[{"id":"Computer1"},{"id":"Computer2"}]},"hasNext":true}

// Payload 2
{"data":{"cpu":"386","year":1993,"screen":{"resolution":"640x480"}},"path":["computers",0],"hasNext":true}

// Payload 3
{"data":{"cpu":"486","year":1996,"screen":{"resolution":"640x480"}},"path":["computers",1],"hasNext":true}
```

On the 1st item we'll call `fromJson` with a json that looks like this:

```json
{
  "computers": [
    {
      "id": "Computer1"
    },
    {
      "id": "Computer2"
    }
  ]
}
```

and `mergedFragmentIds` = `[]`

On the 2nd item we'll call `fromJson` with a json that looks like this:

```json
{
  "computers": [
    {
      "id": "Computer1",
      "cpu": "386",
      "year": 1993,
      "screen": {
        "resolution": "640x480"
      }
    },
    {
      "id": "Computer2"
    }
  ]
}
```

and `mergedFragmentIds` = `[ {path="computers.0", label=null} ]`

Inside the `Computer` adapter, the code needs to know if it should read the ComputerFields: Yes for path `computers.0`, No for path `computers.1`.

Is there a way I'm not seeing to do it without using the current parsing path?

BoD: Yes it is! 😅 

In that case do you still think it would be better to rename BLabel? And pass the set of "ids" to evaluate instead of passing the FragmentContext and delegate to it?

martinbonnin: Yea, no need to rename `BLabel`. I'm still a bit annoyed to add `AdapterContext` as a dependency to `BooleanExpression` but can't really see a way around this so we can refine down the road

BoD: Yeah I see what you mean. But all right for now, next steps for this PR are:
- replicate for Java codegen
- rebase on top of a branch containing `getPath`
- regenerate the `.expected`

Then I think we can merge it. 

Then, the next PR will actually wire things (use `DeferredJsonMerger` from `HttpNetworkTransport`)

martinbonnin: Sounds good 👍

martinbonnin (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/AdapterCommon.kt
```diff
@@ -56,6 +56,12 @@ internal fun readFromResponseCodeBlock(
     )
   }.joinToCode(separator = "\n", suffix = "\n")
 
+  val path = if (syntheticProperties.any { it.condition != BooleanExpression.True }) {
+    CodeBlock.of("val $__path = $reader.$getPath()")
```
martinbonnin, line 60: Can we refine the condition here and only call `getPath()` if we know we're going to need it (because there's a `BLabel` condition? This is a hot path and will save some parsing time for the vast majority of users that do not use `@defer`

BoD: I think yes although, at the moment the path is always passed to `evaluate`. But we could make the `evaluate`'s `path` parameter a `String?`, and initialize it to `null` if not needed instead of not defining it at all - WDYT?

martinbonnin: > I think yes although, at the moment the path is always passed to evaluate. But we could make the evaluate's path parameter a String?, and initialize it to null if not needed instead of not defining it at all - WDYT?

Agreed, passing null is the way to go. Either go with `val __path = null` or pass `null` to `evaluate()` directly like when `__typename` is not needed

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
```diff
@@ -584,10 +589,12 @@ internal fun GQLValue.toIrValue(): IrValue {
...
```
martinbonnin, line 593: Let's do it? 

* `toBooleanExpression` looks at all directives (for fragments)
* `toIncludeBooleanExpression` only looks at `include` and `skip` directives (for fields where we know there is no defer)
* `toDeferBooleanExpression` only looks at `defer` directives (for balance ⚖️ )

BoD: Woops forgot about it! Yes makes sense 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
```diff
@@ -112,23 +114,44 @@ internal class OperationBasedModelGroupBuilder(
...
 
   private fun InlineFragmentKey.toName(): String = buildString {
     append(typeCondition.capitalizeFirstLetter())
     if (condition != BooleanExpression.True) {
-      append("If")
-      append(condition.toName())
+      // If at least one condition is a BLabel, it is enough to make a unique name, because the label is unique
+      val deferCondition: BLabel? = condition.firstElementOfType(BLabel::class)
+      if (deferCondition != null) {
+        append("Defer")
```
martinbonnin, line 138: I would just skip "Defer" there. The [samples from the wg](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) already include "defer" so that'd be a lot of "defer"

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
```diff
@@ -112,23 +114,44 @@ internal class OperationBasedModelGroupBuilder(
...
+    else -> error("Unexpected boolean expression type in toName")
   }
 
   private fun InlineFragmentKey.toName(): String = buildString {
     append(typeCondition.capitalizeFirstLetter())
     if (condition != BooleanExpression.True) {
-      append("If")
-      append(condition.toName())
+      // If at least one condition is a BLabel, it is enough to make a unique name, because the label is unique
+      val deferCondition: BLabel? = condition.firstElementOfType(BLabel::class)
```
martinbonnin, line 136: Maybe check that there is only one "label"? Else it's not balanced ⚖️  if there are multiple:

```graphql
{
  computer {
    ...computerDetails @defer("label1")
    ...computerDetails @defer("label2")
  }
}
```

I know this sample is contrieved and will most likely never happen but if it happens, the longer name shouldn't be too much of an issue (`ComputerDetailsLabel1OrLabel2`)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -96,6 +98,7 @@ fun <T : Any> or(vararg other: BooleanExpression<T>): BooleanExpression<T> = Boo
 fun <T : Any> and(vararg other: BooleanExpression<T>): BooleanExpression<T> = BooleanExpression.And((other.toList()).toSet())
 fun <T : Any> not(other: BooleanExpression<T>): BooleanExpression<T> = BooleanExpression.Not(other)
 fun variable(name: String): BooleanExpression<BVariable> = BooleanExpression.Element(BVariable(name))
+fun label(label: String?): BooleanExpression<BLabel> = BooleanExpression.Element(BLabel(label))
```
martinbonnin, line 101: You could make it `label(label: String? = null)` so that the generated code reads nicer when there is no label:

```
label().evaluate(...)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
BoD: This would be ambiguous if there's no Label though. 

```graphql
computers {
   ... on Computer { id }  # 'onComputer'
   ... on Computer @defer { cpu }  # also 'onComputer'
}
```

But we could append `Defer` only in that case?

martinbonnin: > This would be ambiguous if there's no Label though.

Right!

> But we could append Defer only in that case?

Except if someone does `@defer(label: "defer")` 🤪 

Oh man... Ok let's stick with `"Defer"`

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
BoD: Actually I don't understand 😅  This is only used for inline fragments, right?

martinbonnin: Right 👍 And inline fragments will always ever have one label. Works for me 👍

PR #3950, opened 2022-03-21
🗄️  Do not delay emission of items in watch()

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/3946

### DIFF
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -534,3 +534,17 @@ internal fun <D : Operation.Data> ApolloRequest.Builder<D>.fetchFromCache(fetchF
 internal val <D : Operation.Data> ApolloRequest<D>.fetchFromCache
   get() = executionContext[FetchFromCacheContext]?.value ?: false
 
+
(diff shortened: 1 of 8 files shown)


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -150,13 +148,20 @@ fun <D : Query.Data> ApolloCall<D>.watch(
...
-              }
             }
     )
   }
+
+  /**
+   * We buffer the returned flow so that the subscription to the store happens in the same stacktrace as the last element is
+   * received and *before* the last element is emitted downstream.
+   *
+   * This allows callers to use the last element as a synchronisation point to modify the store and still have the watcher
```
BoD, line 159: 👏

BoD (approved):
👍

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: Well, this broke the stress test... I'm assuming because when the coroutine run on different threads, just relying on the callstack is not enough as coroutine #1 can be interrupted any time...

BoD (approved):
👍

PR #3952, opened 2022-03-21
JS Websocket properly handle remote close

undermark5:
In the addition of support for JS Websockets (PR 3913), I forgot to add a listener for the close events, without this, if the remote closes the connection, the client doesn't get notified. This should resolve that.


BoD (approved):
👍

martinbonnin (approved):
Thanks!

PR #3954, opened 2022-03-23
Add macosArm64 or macosX64 target for tests depending on architecture

BoD:
This makes running `appleTest` on an ARM based Mac work. Note: GitHub workflows don't have ARM runners yet ([issue](https://github.com/actions/virtual-environments/issues/2187), [issue](https://github.com/actions/runner/issues/805)).

Not sure if there's a better/simpler way to do this!

### DIFF
--- a/build-logic/src/main/kotlin/Mpp.kt
+++ b/build-logic/src/main/kotlin/Mpp.kt
@@ -80,18 +80,7 @@ fun Project.configureMppDefaults(withJs: Boolean = true, withLinux: Boolean = tr
 
     addTestDependencies(withJs)
 
-    if (System.getProperty("idea.sync.active") == null) {
-      /**
-       * Evil tasks to fool IntelliJ into running the appropriate tests when clicking the green triangle in the gutter
-       * IntelliJ "sees" apple during sync but the actual tasks are macosX64
-       */
-      tasks.register("cleanAppleTest") {
-        it.dependsOn("cleanMacosX64Test")
-      }
-      tasks.register("appleTest") {
-        it.dependsOn("macosX64Test")
-      }
-    }
+    registerAppleTasks()
   }
 }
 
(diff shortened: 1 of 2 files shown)


gradle/dependencies.gradle
```diff
@@ -99,7 +99,7 @@ ext.dep = [
     gr8                      : "com.gradleup:gr8-plugin:0.2",
     kspGradlePlugin          : "com.google.devtools.ksp:symbol-processing-gradle-plugin:1.6.10-1.0.2",
     kotlinxdatetime          : "org.jetbrains.kotlinx:kotlinx-datetime:0.3.1",
-    kotlinxserializationjson : "org.jetbrains.kotlinx:kotlinx-serialization-json:1.1.0",
+    kotlinxserializationjson : "org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.2",
```
BoD, line 102: Updated because the AppleArm64 version of the artifact didn't exist on `1.1.0`.

build-logic/src/main/kotlin/Mpp.kt
```diff
@@ -113,7 +114,11 @@ fun Project.configureMppTestsDefaults(withJs: Boolean = true) {
         nodejs()
       }
     }
-    macosX64("apple")
+    if (hostArchitecture() == "arm64") {
+      macosArm64("apple")
+    } else {
+      macosX64("apple")
+    }
```
martinbonnin, line 121: Could we just register both? I'm always a bit concerned this might break stuff like build cache or configuration cache (if we ever wanted to enable it in the project)

BoD: Good point! I knew there was probably a simpler way 😅

martinbonnin: Not sure how much simpler it'll be because now we're going to have to lie to IntelliJ to have both ARM and Intel use the same source set... 😅

martinbonnin: Ah yes, just like that 👍 LGTM!

martinbonnin: Oh well maybe not 🙃 . Do we still have IDE support with this? (Life is hard...)

martinbonnin: This works because it's the `commonTest` SourceSet. The problem with `macos("macosArm64")` is that it won't recognize `appleTest` as a source set like `macos("apple")` did. 

See HttpEngineTest.kt for an example

Screenshot 2022-03-24 at 00 37 45

Until we have HMPP, I don't see a way around the `if (sync)` workaround

martinbonnin (approved):

PR #3957, opened 2022-03-24
HttpCacheMissException masks ApolloException in case of NETWORK_FIRST

akshay253101:
see https://github.com/apollographql/apollo-kotlin/issues/3955

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -102,16 +102,12 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
  * [cause] might change to null and second exception will also be added as suppressed exception
  */
 class ApolloCompositeException(
-    private val firstException: Throwable?, private val secondException: Throwable?
+    firstException: Throwable?, secondException: Throwable?,
 ) : ApolloException(message = "multiple exceptions happened", secondException) {
 
-  val first: ApolloException
-    @Throws(RuntimeException::class)
-    get() = (firstException as? ApolloException) ?: throw RuntimeException("unexpected first exception", firstException)
+  val first = (firstException as? ApolloException) ?: throw RuntimeException("unexpected first exception", firstException)
 
-  val second: ApolloException
-    @Throws(RuntimeException::class)
-    get() = (secondException as? ApolloException) ?: throw RuntimeException("unexpected second exception", secondException)
+  val second = (secondException as? ApolloException) ?: throw RuntimeException("unexpected second exception", secondException)
 
   init {
     if (firstException != null && secondException != null) addSuppressed(firstException)
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -98,11 +98,24 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
 class HttpCacheMissException(message: String, cause: Exception? = null) : ApolloException(message = message, cause = cause)
 
 /**
- * Multiple exceptions happened, for an exemple with a [CacheFirst] fetch policy
+ * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
  */
-class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
-  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
-  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
+class ApolloCompositeException(
+    private val original: Throwable?, private val suppressed: Throwable?
+) : ApolloException(message = "multiple exceptions happened", original ?: suppressed) {
+
+  val first
+    @Throws(RuntimeException::class)
+    get() = (original as? ApolloException) ?: throw RuntimeException("unexpected first exception", original)
+
+  val second
+    @Throws(RuntimeException::class)
+    get() = (suppressed as? ApolloException) ?: throw RuntimeException("unexpected second exception", suppressed)
+
+  init {
+    if (original != null && suppressed != null) addSuppressed(suppressed)
+  }
+
 }
 
 class AutoPersistedQueriesNotSupported : ApolloException(message = "The server does not support auto persisted queries")
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -118,10 +118,8 @@ class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloEx
     }
 
   init {
-    if (first != null && second != null) {
-      addSuppressed(first)
-      addSuppressed(second)
-    }
+    if (first != null) addSuppressed(first)
+    if (second != null) addSuppressed(second)
   }
 
 }
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -99,7 +99,7 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
 
 /**
  * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
- * [cause] might change to null and second exception will also be added as suppressed exception
+ * [cause] might change in future where both exceptions will be added as suppressed exception
  */
 class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
 
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -101,16 +101,14 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
  * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
  * [cause] might change to null and second exception will also be added as suppressed exception
  */
-class ApolloCompositeException(
-    firstException: Throwable?, secondException: Throwable?,
-) : ApolloException(message = "multiple exceptions happened", secondException) {
+class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
 
-  val first = (firstException as? ApolloException) ?: throw RuntimeException("unexpected first exception", firstException)
+  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
 
-  val second = (secondException as? ApolloException) ?: throw RuntimeException("unexpected second exception", secondException)
+  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
 
   init {
-    if (firstException != null && secondException != null) addSuppressed(firstException)
+    if (first != null && second != null) addSuppressed(first)
   }
 
 }
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -99,21 +99,22 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
 
 /**
  * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
+ * [cause] might change to null and second exception will also be added as suppressed exception
  */
 class ApolloCompositeException(
-    private val original: Throwable?, private val suppressed: Throwable?
-) : ApolloException(message = "multiple exceptions happened", original ?: suppressed) {
+    private val firstException: Throwable?, private val secondException: Throwable?
+) : ApolloException(message = "multiple exceptions happened", secondException) {
 
-  val first
+  val first: ApolloException
     @Throws(RuntimeException::class)
-    get() = (original as? ApolloException) ?: throw RuntimeException("unexpected first exception", original)
+    get() = (firstException as? ApolloException) ?: throw RuntimeException("unexpected first exception", firstException)
 
-  val second
+  val second: ApolloException
     @Throws(RuntimeException::class)
-    get() = (suppressed as? ApolloException) ?: throw RuntimeException("unexpected second exception", suppressed)
+    get() = (secondException as? ApolloException) ?: throw RuntimeException("unexpected second exception", secondException)
 
   init {
-    if (original != null && suppressed != null) addSuppressed(suppressed)
+    if (firstException != null && secondException != null) addSuppressed(firstException)
   }
 
 }
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
@@ -103,12 +103,25 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
  */
 class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
 
-  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
+  @get:Deprecated("Use suppressedExceptions instead", ReplaceWith("suppressedExceptions.first()"))
+  val first: ApolloException
+    get() {
+      val firstException = suppressedExceptions.firstOrNull()
+      return (firstException as? ApolloException) ?: throw RuntimeException("unexpected first exception", firstException)
+    }
 
-  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
+  @get:Deprecated("Use suppressedExceptions instead", ReplaceWith("suppressedExceptions.getOrNull(1)"))
+  val second: ApolloException
+    get() {
+      val secondException = suppressedExceptions.getOrNull(1)
+      return (secondException as? ApolloException) ?: throw RuntimeException("unexpected second exception", secondException)
+    }
 
   init {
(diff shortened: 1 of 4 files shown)


apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -48,17 +48,29 @@ class CachingHttpInterceptor(
...
 
+        try {
+          return cacheMightThrow(request, cacheKey)
+        } catch (cacheMissException: HttpCacheMissException) {
+          // In case of exception thrown by network request,
+          // ApolloException is the root cause and cache miss exception will be suppressed
+          networkException?.addSuppressed(cacheMissException)
+
+          // Throw network exception, but in case of response status code not in 200..299 throw cache miss exception
+          throw networkException ?: cacheMissException
```
BoD, line 71: Could we throw a `ApolloCompositeException` here instead (to be more consistent with what we do [here](https://github.com/apollographql/apollo-kotlin/blob/42e3d90abc1e8f75b794fc1440c61c1c14457d26/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt#L180) and a few other places)?

akshay253101: Here ApolloCompositeException only throws second
```` 
ApolloException(message = "multiple exceptions happened", second)
````
If we do this then clients must catch exception and then use first.addSuppressed(second) before re-throwing the exception otherwise there network failure will be masked by HttpCacheMissException

Comment (martinbonnin) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: What if we added the suppressed exception always? 

```kotlin
class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
  init {
    if (first != null) {
      addSuppressed(first)
    }
  }
}
``
`

TIL about suppressed exceptions but it looks like they're designed for exactly that case?

Comment (akshay253101) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: Works

Comment (akshay253101) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: Should i proceed with the above changes?

Comment (martinbonnin) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: We can't change the 'first' and 'second' property names nor types, that'd be a breaking change.

We could have just thrown the second one and add first as suppressed one but it's too late now because of API stability (and also because I'm not sure native and js have suppressed exceptions)

Id there a specific reason you're adding the suppressed exception to the second one (and not to the ApolloComposite one) ?

Comment (akshay253101) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: ApolloCompositeException's cause is always 'second' that means if there is no 'second' exception both exceptions stacktrace will be lost

Comment (akshay253101) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: For native and js this might help - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/suppressed-exceptions.html

Comment (martinbonnin) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: > For native and js this might help

👍 Nice. Looks like it's working for JS and native too! Still not a change we can do today due to backward compatibility reasons

> ApolloCompositeException's cause is always 'second' that means if there is no 'second' exception both exceptions stacktrace will be lost

Isn't it possible to add suppressed exception on the exception itself (and not on its cause?)

Comment (BoD) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: I have the impression that throwing `HttpCacheMissException` if there was no network exception, and `ApolloCompositeException` if there was one will allow the calling code to have all the information it needs? The code can check the type of the exception, and if it is composite, access `first` and `second`.  Am I missing something? Adding the suppressed in `ApolloCompositeException` is nice though!

Comment (akshay253101) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: Yes it is possible
````
class ApolloCompositeException(
    original: Throwable?, suppressed: Throwable?,
) : ApolloException(message = "multiple exceptions happened", original ?: suppressed) {

  val first = (original as? ApolloException) ?: throw RuntimeException("unexpected first exception", original)
  val second = (suppressed as? ApolloException) ?: throw RuntimeException("unexpected second exception", suppressed)

  init {
    if (original != null && suppressed != null) addSuppressed(suppressed)
  }
}
````

Comment (BoD) [reply apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt:71]: LGTM!

File: apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt [hunk 2]
```diff
@@ -98,11 +98,24 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
 class HttpCacheMissException(message: String, cause: Exception? = null) : ApolloException(message = message, cause = cause)
 
 /**
- * Multiple exceptions happened, for an exemple with a [CacheFirst] fetch policy
+ * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
```
martinbonnin, line 101: Thanks !

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -98,11 +98,24 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
...
  */
-class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
-  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
-  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
+class ApolloCompositeException(
+    private val original: Throwable?, private val suppressed: Throwable?
```
martinbonnin, line 104: This is a Kotlin source incompatible change because of named parameters. Can you keep `first` and `second`?

akshay253101: okay

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -98,11 +98,24 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
...
+) : ApolloException(message = "multiple exceptions happened", original ?: suppressed) {
```
martinbonnin, line 105: What's the reason to put the first exception as the cause when it was the second before? I'm not sure there are strong reasons to prefer one over the other but if we change it we should document it.

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -48,17 +49,31 @@ class CachingHttpInterceptor(
         return networkMightThrow(request, chain, cacheKey)
       }
       NETWORK_FIRST -> {
+        var networkException: ApolloException? = null
         try {
           val response = networkMightThrow(request, chain, cacheKey)
           if (response.statusCode in 200..299) {
             //  let HTTP errors through
             return response
           }
```
martinbonnin, line 58: You can throw an ApolloHttpResponse here:

```suggestion
          if (response.statusCode in 200..299) {
            //  let HTTP errors through
            return response
          } else {
            throw ApolloHttpException(response.statusCode, response.headers)
          }
```

That'll simplify the block below

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -48,17 +49,31 @@ class CachingHttpInterceptor(
...
+          // Original cause of network request failure
+          networkException = e
         }
 
-        return cacheMightThrow(request, cacheKey)
+        try {
+          return cacheMightThrow(request, cacheKey)
+        } catch (cacheMissException: HttpCacheMissException) {
+          // In case of response status code not in 200..299 throw cache miss exception
+          if (networkException == null) throw cacheMissException
```
martinbonnin, line 68: With the suggestion above, this is not needed

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: CacheFirstInterceptor work is to fetch from cache first then network, if cacheMightThrow throws an exception it should be logged first and networkMightThrow should suppressed by it and same for NetworkFirstInterceptor and CacheAndNetworkInterceptor
This is the reason why first operations exception is thrown first and second is suppressed

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
akshay253101: Agree

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: accessing first and second might throw RuntimeException so it would easy if client is aware when using such properties.
But throws annotation is not applicable to target 'member property without backing field or delegate' that is why 
1. property getter is added
2. getter function cannot access non val constructor params
3. params marked val must name different then properties name

That's the reason had to rename constructor params and set them to private val

martinbonnin: I see 👍 . We're not declaring the `@Throws` exceptions consistently in the codebase. Mainly because Kotlin doesn't have checked exceptions but that'll help with Java interop if someone ever watch to catch these. So works for me 👍

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
BoD: For the cause, I don't have a strong opinion either - in fact I'm not even sure it makes sense to set the cause at all 😅: `ApolloCompositeException` means "2 exception happened", is it really true to say that one is the _cause_ of the other? But I kind of agree that changing this now could be a bit dangerous if existing code depends on it.

For the suppressed... Well now that I think even more about it, I think both should be marked as suppressed? There can be more than 1, right?

akshay253101: Yes, it can be multiple as it supports list of suppressed exception

akshay253101: So one doubt here if we keep both as suppressed, default cause will be null
````
ApolloException(message = "multiple exceptions happened", null)
````
Which will change the current behavior

Comment (akshay253101) [reply apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt:105]: Agree, i also think that would be safe and we can update it in future

Comment (akshay253101) [reply apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt:105]: Which one you think will do
````
second ?: first 
````
or 
````
second
````

Comment (martinbonnin) [reply apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt:105]: Since we're going for backward compatibility, let's keep

```
second
```

Comment (akshay253101) [reply apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt]: Updated

File: apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt [hunk 7]
```diff
@@ -98,11 +98,25 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
...
 /**
- * Multiple exceptions happened, for an exemple with a [CacheFirst] fetch policy
+ * Multiple exceptions happened, for an example with a [CacheFirst] fetch policy
+ * [cause] might change to null and second exception will also be added as suppressed exception
  */
-class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
-  val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
-  val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
+class ApolloCompositeException(
+    private val firstException: Throwable?, private val secondException: Throwable?
```
martinbonnin, line 105: Thanks for all the work! One last comment: this really needs to be `first` and `second` in case anyone uses named parameters.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -102,16 +102,12 @@ class HttpCacheMissException(message: String, cause: Exception? = null) : Apollo
  * [cause] might change to null and second exception will also be added as suppressed exception
  */
 class ApolloCompositeException(
-    private val firstException: Throwable?, private val secondException: Throwable?
+    firstException: Throwable?, secondException: Throwable?,
 ) : ApolloException(message = "multiple exceptions happened", secondException) {
 
-  val first: ApolloException
-    @Throws(RuntimeException::class)
```
martinbonnin: I think it's available in common code. We're using it there for an example: https://github.com/apollographql/apollo-kotlin/blob/31c873c5abbc4873c98e31ba17da7c16e90109ef/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/JsonReader.kt#L45

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: Removed Throws annotation as it only available in jvm - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: yes, my bad 
Have checked the failing test cases it seem it is only available for jvm
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: okay

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
akshay253101: For 'native' it can only be set on AnnotationTarget.Function, while in 'common' it can be set on AnnotationTarget.PROPERTY_GETTER

akshay253101: Changing to function instead of property to annotate with error for java-interop, will be a breaking change. so i think we can remove annotation for now?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
```diff
@@ -98,11 +98,19 @@ class CacheMissException(val key: String, val fieldName: String? = null) : Apoll
...
+ * [cause] might change to null and second exception will also be added as suppressed exception
  */
 class ApolloCompositeException(first: Throwable?, second: Throwable?) : ApolloException(message = "multiple exceptions happened", second) {
+
   val first = (first as? ApolloException) ?: throw RuntimeException("unexpected first exception", first)
+
   val second = (second as? ApolloException) ?: throw RuntimeException("unexpected second exception", second)
+
+  init {
+    if (first != null && second != null) addSuppressed(first)
```
BoD, line 111: Ok to keep `second` for the cause, but to add both `first` and `second` to the suppressed? So something like this:

```suggestion
    if (first != null) addSuppressed(first)
    if (second != null) addSuppressed(second)
```

akshay253101: then there will be a duplicate stack trace like
1. SecondException
2. FirstException
3. SecondException

BoD: Oh ok didn't realize they would appear in the stack trace by default!

I'm not sure it's really an issue though? Maybe it's ok to have the `second` one appearing twice (once for cause, and once for suppressed), at least you can't miss it 😅

akshay253101: Okay, that way should i also add previously suggested deprecation on first and second usages?

BoD: Yes that would be nice!

martinbonnin (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/exception/Exceptions.kt
martinbonnin: +1 to both suppressed for consistency and symmmetry. Also to make sure we have `second` when we will remove it from the `cause` later on.

BoD (approved):

martinbonnin:
Thanks!

PR #3960, opened 2022-03-24
Defer: Use DeferredJsonMerger in HttpNetworkTransport

BoD:
In this PR:
- use DeferredJsonMerger in HttpNetworkTransport
- fix `AdapterContext.newBuilder` and rename for consistency
- test defer cases 🎉 (need more tests! For instance error cases)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -78,58 +80,90 @@ private constructor(
...
+  private suspend fun <D : Operation.Data> multipleResponses(
+      operation: Operation<D>,
+      customScalarAdapters: CustomScalarAdapters,
+      httpResponse: HttpResponse,
+  ): Flow<ApolloResponse<D>> {
+    val jsonMerger = DeferredJsonMerger()
+    return multipartBodyFlow(httpResponse).map { part ->
+      try {
+        // On native, components of jsonMerger are frozen inside worker.doWork, making mutability impossible.
+        // So we clone them before they are captured.
```
martinbonnin, line 126:
```suggestion
        // On native, we cannot pass `jsonMerger._merge` to `worker.doWork` or it will become frozen making
        // any subsequent payload merge throw.
        // So we clone them before they are captured.
        // XXX: revisit with the new memory model
```

BoD: Much better! 👍

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -78,58 +80,92 @@ private constructor(
...
-      )
+  private suspend fun <D : Operation.Data> singleResponse(
+      operation: Operation<D>,
+      customScalarAdapters: CustomScalarAdapters,
+      httpResponse: HttpResponse,
+  ): ApolloResponse<D> {
+    val response = try {
+        worker.doWork {
+          operation.parseJsonResponse(
+              jsonReader = httpResponse.body!!.jsonReader(),
```
martinbonnin, line 105: How come we don't need the `val bodyByteString = if (platform() == Platform.Native) body.readByteString() else null` thing?

tests/defer/src/commonTest/kotlin/test/DeferTest.kt
```diff
@@ -0,0 +1,271 @@
...
+
+private fun assertResponseListEquals(expectedDataList: List<ApolloResponse<*>>, actualResponseList: List<ApolloResponse<*>>) {
+  assertContentEquals(expectedDataList, actualResponseList) { expectedResponse, actualResponse ->
+    assertEquals(expectedResponse.data, actualResponse.data)
+    assertContentEquals(expectedResponse.errors, actualResponse.errors) { expectedError, actualError ->
+      assertEquals(expectedError.message, actualError.message)
+      assertContentEquals(expectedError.path, actualError.path)
+      assertContentEquals(expectedError.locations, actualError.locations) { expectedLocation, actualLocation ->
+        assertEquals(expectedLocation.line, actualLocation.line)
+        assertEquals(expectedLocation.column, actualLocation.column)
```
martinbonnin, line 254: No data class for Errors 😅. Not sure if there's a specific reason for that. Let's keep this in mind, we could add it if we find more evidence this could help.

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
BoD: This was needed when capturing/freezing `body`, but now we capture/freeze `httpResponse` so it can be simple again.

tests/defer/src/commonTest/kotlin/test/DeferTest.kt
```diff
@@ -0,0 +1,271 @@
...
+                        ComputerFields.Screen("Screen", "800x600",
+                            ScreenFields(true)))),
+                )
+            )
+        ).build(),
+    )
+
+    mockServer.enqueueMultipart(jsonList)
+    val actualResponseList = apolloClient.query(query).toFlow().toList()
+    assertResponseListEquals(expectedDataList, actualResponseList)
```
martinbonnin, line 242: Nice tests! 👍 

It'd be interesting to test against a real server too. I think helix has defer support. We could make a JS-only test that runs helix? Like we have JVM-only tests running graphql-kotlin ?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
martinbonnin: But freezing `httpResponse` will deep freeze all its properties so it's the same thing no?

BoD: Yes but `HttpResponse.body` is `get() = bodySource ?: bodyString?.let { Buffer().write(it) }` and in native we're in the `bodyString` case.

martinbonnin: 👍  got it!

martinbonnin (approved):

tests/defer/src/commonTest/kotlin/test/DeferTest.kt
BoD: Yeah wasn't sure if it was OK to change it at this point :) We could also only add equals/hashCode without making a data class if that's a bit less "API changing".

tests/defer/src/commonTest/kotlin/test/DeferTest.kt
BoD: Totally! I've been using Helix locally to test stuff. Not sure how hard it would be to integrate it inside Kotlin tests but that definitely would be useful! I'll create a specific ticket.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt
```diff
@@ -17,23 +19,24 @@ class AdapterContext private constructor(
   }
 
   fun hasDeferredFragment(path: String, label: String?): Boolean {
-    return mergedDeferredFragmentIds?.contains(DeferredFragmentIdentifier(path, label)) == true
+    val sanitizedPath = path.removePrefix("data.").trim('.')
```
martinbonnin, line 22: Could you avoid adding "data." in the first place to save bytes and CPU?

apollo-api/api/apollo-api.api
```diff
@@ -897,7 +896,7 @@ public final class com/apollographql/apollo3/api/json/BufferedSourceJsonReader :
 	public fun close ()V
 	public fun endArray ()Lcom/apollographql/apollo3/api/json/JsonReader;
 	public fun endObject ()Lcom/apollographql/apollo3/api/json/JsonReader;
-	public fun getPath ()Ljava/lang/String;
+	public fun getPath ()Ljava/util/List;
```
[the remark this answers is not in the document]
martinbonnin: My mistake, you're free to go sir, have a nice day!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -126,11 +126,15 @@ fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?)
...
 
-fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?, adapterContext: AdapterContext, path: String?): Boolean {
+fun BooleanExpression<BTerm>.evaluate(variables: Set<String>, typename: String?, adapterContext: AdapterContext, path: List<Any>?): Boolean {
   return evaluate {
     when (it) {
       is BVariable -> variables.contains(it.name)
-      is BLabel -> adapterContext.hasDeferredFragment(path!!, it.label)
+      is BLabel -> {
+        // Remove "data" from the path
+        val croppedPath = path!!.drop(1)
```
martinbonnin, line 135: Move it out of the lambda so that it's called only once

PR #3961, opened 2022-03-24
Add note to tutorial about `graphql-ws` library to tutorial

rkoron007:


docs/source/tutorial/11-subscriptions.mdx
```diff
@@ -58,6 +58,7 @@ subscription TripsBooked {
 ```
 
 ## Configure your ApolloClient to use subscriptions
+> By default, Apollo Kotlin uses the [`subscriptions-transport-ws`](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) library to set up subscriptions. **This library is no longer actively maintained.** We recommend using the [`graphql-ws`](https://www.npmjs.com/package/graphql-ws) library instead, which implements its own WebSocket subprotocol, `GraphQLWsProtocol`. Note that the two libraries do not use the same WebSocket subprotocol, so you need to ensure that your servers, clients, and tools use the **same library and subprotocol**. For more information and examples, see [GraphQL over WebSocket protocols](../essentials/subscriptions/#customizing-your-websocket-protocol).
```
BoD, line 61: Should we update [this section](https://www.apollographql.com/docs/kotlin/essentials/subscriptions#customizing-your-websocket-protocol) with the same info? 

And I guess changing the protocol used by default would be nice... but that could break existing users when they upgrade 🫤 But at least we can mark it as Deprecated in the code.

BoD (approved):

docs/source/tutorial/11-subscriptions.mdx
martinbonnin: > Should we update [this section](https://www.apollographql.com/docs/kotlin/essentials/subscriptions#customizing-your-websocket-protocol) with the same info?

+1 to that. I took the liberty of pushing a paragraph there and a few wording changes. Let me know what you think

martinbonnin:
Thanks for sending this! I meant to do it some time ago and completely forgot 🤦 . Thanks a lot 💜

PR #3967, opened 2022-03-25
⬇️  Also use ApolloCompositeException for CacheFirst 

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/3957 to keep everything balanced ⚖️


BoD (approved):
Makes sense 👍

PR #3969, opened 2022-03-25
Typename: add  "always"

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/3965

```kotlin
 /**
   * When to add __typename. One of "always", "ifFragments", "ifImplementsAbstractType", "ifAbstract" or "ifPolymorphic"
   *
   * - "always": Add '__typename' for every compound field
   *
   * - "ifFragments": Add '__typename' for every selection set that contains fragments (inline or named)
   * This causes cache misses when introducing fragments where no fragment was present before and will be certainly removed in
   * a future version.
   *
   * - "ifAbstract": Add '__typename' for abstract fields, i.e. fields that are of union or interface type
   * Note: It also adds '__typename' on fragment definitions that satisfy the same property because fragments
   * could be read from the cache and we don't have a containing field in that case.
   *
   * - "ifPolymorphic": Add '__typename' for polymorphic fields, i.e. fields that contains a subfragment
   * (inline or named) whose type condition isn't a super type of the field type.
   * If a field is monomorphic, no '__typename' will be added.
   * This adds the bare minimum amount of __typename but the logic is substantially more complex and
   * it could cause cache misses when using fragments on monomorphic fields because __typename can be
   * required in some cases.
   *
   * Note: It also adds '__typename' on fragment definitions that satisfy the same property because fragments
   * could be read from the cache and we don't have a containing field in that case.
   *
   * Default value: "ifFragments"
   */
```


BoD (approved):
👍

martinbonnin:
I removed "ifImplementsAbstractType" as it's becoming really complex and doesn't handle schema changes adding new interfaces on existing types. Typenames.md is updated to reflect that. Unless there are objections, I'd like to go with that for `3.2.0`

BoD (approved):

PR #3972, opened 2022-03-29
Defer: add defer specific validation

BoD:


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -310,6 +320,111 @@ internal class ExecutableValidationScope(
...
+   */
+  private fun GQLSelectionSet.validateDeferDirectives(path: String) {
+    for (selection in selections) {
+      when (selection) {
+        is GQLField -> {
+          val fieldPath = if (path.isEmpty()) selection.name else "$path.${selection.name}"
+          selection.selectionSet?.validateDeferDirectives(fieldPath)
+        }
+        is GQLInlineFragment -> {
+          val deferDirective = selection.children.firstOrNull { it is GQLDirective && it.name == "defer" } as GQLDirective?
```
martinbonnin, line 375: There should be a `GQLInlineFragment.directives` property to avoid the use of the more generic `children`

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -310,6 +320,111 @@ internal class ExecutableValidationScope(
...
+          if (deferDirective != null && deferDirective.validateDeferDirective(path, deferDirective.sourceLocation)) return
+          selection.selectionSet.validateDeferDirectives(path)
+        }
+        is GQLFragmentSpread -> {
+          val deferDirective = selection.children.firstOrNull { it.name == "defer" }
```
martinbonnin, line 380: Same here for `GQLFragmentSpread.directives`

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -70,6 +75,11 @@ internal class ExecutableValidationScope(
       it.validate()
     }
 
+    deferDirectiveLabels.clear()
+    operations.forEach {
+      it.validateDeferDirectives()
```
martinbonnin, line 80: Could you plug into the more generic `validate()` above (and pass a path along the way). No strong preference but the other validations use that and it saves traversing the operation mutiple times. Not sure if that's a real concern or not. Just thought I'd bring it up

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -310,6 +320,111 @@ internal class ExecutableValidationScope(
...
+      registerIssue(
+          message = "@defer label argument must not be a variable",
+          sourceLocation = sourceLocation
+      )
+      return true
+    }
+
+    var labelStringValue = ""
+    if (label != null) {
+      labelStringValue = (label as GQLStringValue).value
```
martinbonnin, line 403: register an issue instead of throwing if label is not a `GQLStringValue`?

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -310,6 +320,111 @@ internal class ExecutableValidationScope(
...
+      if (labelStringValue in deferDirectiveLabels) {
+        registerIssue(
+            message = "@defer label '$labelStringValue' must be unique within all other @defer directives in the document",
+            sourceLocation = sourceLocation
```
martinbonnin, line 407: It could be nice to "remember" where the other `@defer` sourceLocation is to give more actionable errors

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
BoD: This should be caught beforehand by the argument type validation! But I did a test and it crashed indeed :( I'll investigate why.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -320,11 +315,9 @@ internal class ExecutableValidationScope(
         ))
       }
     }
-  }
 
-  private fun GQLOperationDefinition.validateDeferDirectives() {
-    deferDirectivePathAndLabels.clear()
-    selectionSet.validateDeferDirectives("")
+    if (issues.isNotEmpty()) return
+    selectionSet.validateDeferDirectives()
```
martinbonnin, line 320: Could you go even one step further and reuse `GQLSelectionSet.validate()` (and `GQLField.validate()`, etc...)? This is where the traversal is happening. Not 100% sure if that's possible but feels like it should be?

BoD: Hehe I contemplated it without really trying and my first impression was that it would make the code more complex/harder to read - but let me try a bit more so we can know for sure 👍

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -320,11 +315,9 @@ internal class ExecutableValidationScope(
         ))
       }
     }
-  }
 
-  private fun GQLOperationDefinition.validateDeferDirectives() {
-    deferDirectivePathAndLabels.clear()
-    selectionSet.validateDeferDirectives("")
+    if (issues.isNotEmpty()) return
```
martinbonnin, line 319: I'd like to avoid stopping the validation early. This code could be used to display warnings in an editor or something else and we want to have as many issues as possible in a single pass. For an example:

```graphql
{
  computer {
     ...fragment1 @defer
     ...fragment2 @defer
     ...fragment3 @defer(label: 42)
  }
}
```
Ideally, we want an issue to be output about fragment1 and fragment2 (and fragment3 too)

BoD: Oh ok, that was my way of avoiding to handle `labelStringValue = (label as GQLStringValue).value` later :) I can skip instead.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
BoD: All right in the end it's not bad I think 😊 I had to pass `path` around. 
Also there's a small trick to **not** do the `@defer` specific validation in the case where we come from `GQLFragmentDefinition.validate()` because in that case the path+label unicity doesn't make sense.  LMK what you think.

martinbonnin: We could pass a `TraversalContext` down the rabbit hole if we want more data but this is all internal API so we can always change later if needed 👍

PR #3973, opened 2022-03-30
⚙️  Fix fragments on the root query type in operationBased codegen

martinbonnin:
Because the adapter was never buffered, we could never rewind.

Closes https://github.com/apollographql/apollo-kotlin/issues/3914


BoD (approved):
👍

PR #3974, opened 2022-03-30
Fixed link to http interceptors

neugartf:


martinbonnin:
Thanks for sending this! It's "working as designed" for me as the link redirects to the "HTTP interceptors" documentation page (https://www.apollographql.com/docs/kotlin/advanced/interceptors-http). We could link to the KDoc but I'd prefer reserving that to actual class names like [`HttpInterceptor`](https://apollographql.github.io/apollo-kotlin/kdoc/apollo-runtime/com.apollographql.apollo3.network.http/-http-interceptor/index.html). Would that work for you?

neugartf:
Sure! My bad, it was just giving me a 404 browsing here on GitHub 👍.

martinbonnin:
Ah yes, the docs are processed so they all end up on the documentation site. I wish we could make that more prominent while browsing on Github. Not sure how though. Worst case we can maybe use absolute links everywhere 🤔 . I'll ask around.

PR #3986, opened 2022-04-01
Defer: Fix false positive on duplicate labels validation

BoD:
Fragments can be reused of course, so when finding a matching label, make sure it's not actually the same node we're looking at.

### DIFF
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
@@ -58,7 +58,7 @@ internal class ExecutableValidationScope(
    */
   override val variableUsages = mutableListOf<VariableUsage>()
 
-  private val deferDirectiveLabels = mutableMapOf<String, GQLDirective>()
+  private val deferDirectiveLabels = mutableMapOf<String, SourceLocation>()
   private val deferDirectivePathAndLabels = mutableMapOf<String, SourceLocation>()
 
   fun validate(document: GQLDocument): List<Issue> {
@@ -70,7 +70,6 @@ internal class ExecutableValidationScope(
       it.validate()
     }
 
-    deferDirectiveLabels.clear()
     val operations = document.definitions.filterIsInstance<GQLOperationDefinition>()
     operations.checkDuplicateOperations()
     operations.forEach {
@@ -293,6 +292,7 @@ internal class ExecutableValidationScope(
 
   private fun GQLOperationDefinition.validate() {
     variableUsages.clear()
+    deferDirectiveLabels.clear()
     deferDirectivePathAndLabels.clear()
 
     val rootTypeDefinition = rootTypeDefinition(schema)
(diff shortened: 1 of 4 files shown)


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -392,15 +392,15 @@ internal class ExecutableValidationScope(
         )
       }
 
-      if (labelStringValue in deferDirectiveLabels) {
+      if (labelStringValue in deferDirectiveLabels && deferDirectiveLabels[labelStringValue] != this) {
         registerIssue(
             message = "@defer label '$labelStringValue' must be unique within all other @defer directives in the document. " +
-                "Same label found in ${deferDirectiveLabels[labelStringValue]!!.pretty()}",
+                "Same label found in ${deferDirectiveLabels[labelStringValue]!!.sourceLocation.pretty()}",
             sourceLocation = sourceLocation
```
martinbonnin, line 399: Why not using the same logic as for `deferDirectivePathAndLabels` and call `deferDirectiveLabels.clear()` from `GQLOperationDefinition.validate()`?

BoD: The labels if present, must be unique across the whole document, whereas the path+labels must be unique within a specific operation.

martinbonnin: `document` is very broad. When it reaches validation, it's a document containing all executable definitions. Meaning something like this would trigger a validation error:

```graphql
query Query1 {
  field {
    ...fieldDetails1 @defer(label: "details")
  }
}

query Query2 {
  field {
    ...fieldDetails2 @defer(label: "details")
  }
}
```

I'm usually the kind that would rather add more restrictions (because is always easier to lift them afterwards than introduce new ones) but I can't see why we would ever need that restriction. Or is there something that could go wrong?

BoD: I think you're right.

The [current wording](https://github.com/graphql/graphql-spec/pull/742/files#diff-607ee7e6b71821eecadde7d92451b978e8a75e23d596150950799dc5f8afa43eR1536) uses the term "document", and I took it to mean the whole document but now I'm thinking maybe it's meant to mean something like "the document that is sent when executing an operation" (if that makes sense?). But yes, I think it makes more sense for the unicity to be in the context of a particular operation 👍  I can also ask for confirmation on the WG's repo 😊

BoD: I got [confirmation](https://github.com/robrichard/defer-stream-wg/discussions/35#discussioncomment-2501662) that document indeed meant the whole document but not because it's strictly necessary, but to ease the  implementation of the validation. In our case it's actually more consistent to keep this per operation so I'll make the change.

martinbonnin (approved):

PR #3987, opened 2022-04-01
Defer: take @defer cases into account with normalized cache

BoD:
With defer, multiple elements can be emitted in Flows returned by the network, which prompted a few adaptations in the normalized cache interceptors.

Also, always read deferred fragments in the adapter when used to parse cached values.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt
@@ -19,7 +19,11 @@ class AdapterContext private constructor(
   }
 
   fun hasDeferredFragment(path: List<Any>, label: String?): Boolean {
-    return mergedDeferredFragmentIds?.contains(DeferredFragmentIdentifier(path, label)) == true
+    if (mergedDeferredFragmentIds == null) {
+      // By default, parse all deferred fragments - this is the case when parsing from the normalized cache.
+      return true
+    }
+    return mergedDeferredFragmentIds.contains(DeferredFragmentIdentifier(path, label))
   }
 
   class Builder {
--- a/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
+++ b/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
@@ -13,7 +13,9 @@ import com.apollographql.apollo3.interceptor.ApolloInterceptor
 import com.apollographql.apollo3.interceptor.ApolloInterceptorChain
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.singleOrNull
 import kotlin.jvm.JvmName
 
@@ -72,26 +74,25 @@ val CacheFirstInterceptor = object : ApolloInterceptor {
         } else {
           throw it
         }
-      }.singleOrNull()
-
-      if (networkResponse != null) {
-        emit(
-            networkResponse.newBuilder()
-                .cacheInfo(
-                    networkResponse.cacheInfo!!
-                        .newBuilder()
-                        .cacheMissException(cacheException as? CacheMissException)
-                        .build()
-                )
-                .build()
-        )
-        return@flow
+      }.map { response ->
+        response.newBuilder()
+            .cacheInfo(
+                response.cacheInfo!!
+                    .newBuilder()
+                    .cacheMissException(cacheException as? CacheMissException)
+                    .build()
+            )
+            .build()
       }
 
-      throw ApolloCompositeException(
-          first = cacheException,
-          second = networkException
-      )
+      emitAll(networkResponse)
+
+      if (networkException != null) {
+        throw ApolloCompositeException(
+            first = cacheException,
+            second = networkException
+        )
+      }
     }
   }
 }
@@ -110,10 +111,10 @@ val NetworkFirstInterceptor = object : ApolloInterceptor {
         } else {
           throw it
         }
-      }.singleOrNull()
+      }
 
-      if (networkResponse != null) {
-        emit(networkResponse)
+      emitAll(networkResponse)
+      if (networkException == null) {
         return@flow
       }
 
@@ -186,28 +187,29 @@ val CacheAndNetworkInterceptor = object : ApolloInterceptor {
             } else {
               throw it
             }
-          }.singleOrNull()
-
-      if (networkResponse != null) {
-        emit(
-            networkResponse.newBuilder()
+          }
+          .map { response ->
+            response.newBuilder()
                 .cacheInfo(
-                    networkResponse.cacheInfo!!
+                    response.cacheInfo!!
                         .newBuilder()
                         .cacheMissException(cacheException as? CacheMissException)
                         .build()
                 )
                 .build()
-        )
-        return@flow
+          }
+
+      emitAll(networkResponse)
+
+      if (networkException != null) {
+        if (cacheException != null) {
+          throw ApolloCompositeException(
+              first = cacheException,
+              second = networkException
+          )
+        }
+        throw networkException!!
       }
-      if (cacheException != null) {
-        throw ApolloCompositeException(
-            first = cacheException,
-            second = networkException
-        )
-      }
-      throw networkException!!
     }
   }
 }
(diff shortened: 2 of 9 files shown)


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
```diff
@@ -134,26 +137,38 @@ internal class ApolloCacheInterceptor(
...
-        maybeWriteToCache(request, response, customScalarAdapters, optimisticKeys)
+        maybeWriteToCache(request, response, customScalarAdapters, optimisticKeys!!)
         emit(response)
-      } else {
-        store.publish(optimisticKeys)
-        throw exception!!
+      }
+
+      if (networkException != null) {
+        if (optimisticKeys == null) optimisticKeys = if (optimisticData != null) {
```
BoD, line 164: I'm sorry I have to repeat this block (also not 100% certain this is correct). Could there be a nicer way to write this?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
```diff
@@ -229,18 +242,18 @@ internal class ApolloCacheInterceptor(
       request: ApolloRequest<D>,
       chain: ApolloInterceptorChain,
       customScalarAdapters: CustomScalarAdapters,
-  ): ApolloResponse<D> {
+  ): Flow<ApolloResponse<D>> {
     val startMillis = currentTimeMillis()
-    val networkResponse = chain.proceed(request).onEach {
+    return chain.proceed(request).onEach {
       maybeWriteToCache(request, it, customScalarAdapters)
```
BoD, line 248: Would it make more sense to cache only the last (full) response? I'm not sure caching intermediate responses have merit?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt
```diff
@@ -19,7 +19,11 @@ class AdapterContext private constructor(
   }
 
   fun hasDeferredFragment(path: List<Any>, label: String?): Boolean {
-    return mergedDeferredFragmentIds?.contains(DeferredFragmentIdentifier(path, label)) == true
+    if (mergedDeferredFragmentIds == null) {
+      // By default, parse all deferred fragments - this is the case when parsing from the normalized cache.
+      return true
+    }
+    return mergedDeferredFragmentIds.contains(DeferredFragmentIdentifier(path, label))
```
martinbonnin, line 26: This means that reading from the cache will basically ignore the `@defer`, right? It'd be interesting to call this out somewhere, future documentation I guess

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -72,26 +74,25 @@ val CacheFirstInterceptor = object : ApolloInterceptor {
         } else {
           throw it
         }
-      }.singleOrNull()
+      }.map { response ->
```
martinbonnin, line 77: This breaks the contract that `CacheFirst` [emits only once](https://github.com/apollographql/apollo-kotlin/blob/2e21bd2952d75e86c58920c498568ab73d632774/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt#L40). I guess there's no real way around this but the doc needs to be updated

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
```diff
@@ -134,26 +137,38 @@ internal class ApolloCacheInterceptor(
...
        */
-      var response: ApolloResponse<D>? = null
-      var exception: ApolloException? = null
-      try {
-        response = chain.proceed(request).single()
-      } catch (e: ApolloException) {
-        exception = e
-      }
+      var networkException: ApolloException? = null
+      val networkResponse: Flow<ApolloResponse<D>> = chain.proceed(request)
```
martinbonnin, line 141: Nit: if it's a Flow, it should be plural: `networkResponses`

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
martinbonnin: The more I think about it, the more I think optimistic updates + defer doesn't make sense:

```graphql
mutation {
  updateUser(name: $name, email: $email) {
    name
    email @defer
  }
}
```

What if we set `"John Doe"`/`"john@doe.com"` as optimistic data and we get the first payload but the second fails? We're now in a weird state where the local cache doesn't match the backend because we can't rollback individual fields. 

For a first version, I would throw at runtime if there is optimistic data but downstream emits several responses. We can always revisit and introduce "incremental" optimistic data (but I hope we won't have to 😄 )

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -186,28 +187,29 @@ val CacheAndNetworkInterceptor = object : ApolloInterceptor {
...
+      if (networkException != null) {
+        if (cacheException != null) {
+          throw ApolloCompositeException(
+              first = cacheException,
+              second = networkException
+          )
+        }
+        throw networkException!!
       }
-      throw networkException!!
```
martinbonnin, line 210: Not a fan of what all the custom logic in this file has become. In retrospect, I think that throwing on cache misses was a mistake (see https://github.com/apollographql/apollo-kotlin/issues/3976) but throwing on network errors might also be (because throwing always break the flow).

If we emit `ApolloResponse` all the time (with a potential `.exception` field) we can move the `ApolloCompositeException` logic to `.execute()`:

```kotlin
  suspend fun execute(): ApolloResponse<D> {
    val responses = toFlow().toList()
    
    val (errors, successes) = responses.partition { it.exception != null }
    
    if (successes.size > 1) {
      throw ApolloException("The operation returned multiple items, use .toFlow() instead of .execute()")
    } else if (successes.isEmpty()) {
      if (errors.size == 1) {
        throw errors.single().exception!!
      } else if (errors.size > 1) {
        throw ApolloCompositeException(errors.map { it.exception })
      } else {
        throw ApolloException("The operation did not emit any item, check your interceptor chain")
      }
    } else {
      return successes.first()
    }
  }
```

This would put this logic in a central and final place and keep the flow running for as long as possible.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/AdapterContext.kt
BoD: Yes exactly. If some fields are not yet received, this will be a cache miss, otherwise, we read all the fragments as if `@defer` didn't exist.

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
BoD: Hmm I see! Indeed that is not ideal and we can forbid this case for now 👍

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
BoD: I played with the idea [here](https://github.com/apollographql/apollo-kotlin/commit/587167a16e81e04d6a2261613ba6beb8a229a984), and this indeed simplifies `FetchPolicyInterceptors` + it's nice to have the exception handling centralized here.

This breaks some tests though and I'm not sure we can do this change without changing the contract of some of the calls due to how they currently throw exceptions in `toFlow` vs the idea of emitting them with a non-null `exception` field instead.

Ok for me to open an issue to keep track of this idea?

martinbonnin: Sure thing! It's a big change that might require some tweaks (maybe even breaking changes). Let's follow up in a separate issue

BoD: I've opened https://github.com/apollographql/apollo-kotlin/issues/4003.

tests/defer/src/commonTest/kotlin/test/DeferNormalizedCacheTest.kt
```diff
@@ -0,0 +1,462 @@
...
+    assertFailsWith<CacheMissException> {
+      apolloClient.query(WithFragmentSpreadsQuery()).execute()
+    }
+
+    // Fill the cache by doing a network only request
+    val jsonList = listOf(
+        """{"data":{"computers":[{"__typename":"Computer","id":"Computer1"}]},"hasNext":true}""",
+        """{"data":{"cpu":"386","year":1993,"screen":{"__typename":"Screen","resolution":"640x480"}},"path":["computers",0],"hasNext":true}""",
+        """{"data":{"isColor":false},"path":["computers",0,"screen"],"hasNext":false,"label":"a"}""",
+    )
```
martinbonnin, line 73: (Dog) Food for thought: looks like we could use test builders to simplify those tests

PR #3988, opened 2022-04-03
Creating a copy() funtion for the ApolloRequest 

AdamMTGreenberg:
**Purpose:** to allow setting a new `Operation` which consumes the exact same values as an existing request and returns the result. Useful for when we want to intercept and relay the `ApolloRequest` through various stages of the network request.

There are various points in the API where one may want to wrap or proxy calls in the chain and without access to the `Operation`, we cannot. This `copy` function would allow the consumer to be sure that all values (beside the UUID) would be copied over, both now and in the future.

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
@@ -23,8 +23,12 @@ private constructor(
     override val canBeBatched: Boolean?,
 ) : ExecutionOptions {
 
-  fun newBuilder(): Builder<D> {
-    return Builder(operation)
+  fun newBuilder(): Builder<D> = duplicate(Builder(operation))
+
+  fun copy(operation: Operation<D>): ApolloRequest<D> = duplicate(Builder(operation)).build()
+
+  private fun duplicate(builder: Builder<D>): Builder<D> {
+    return builder
         .requestUuid(requestUuid)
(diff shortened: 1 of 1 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
```diff
@@ -23,8 +23,12 @@ private constructor(
     override val canBeBatched: Boolean?,
 ) : ExecutionOptions {
 
-  fun newBuilder(): Builder<D> {
-    return Builder(operation)
+  fun newBuilder(): Builder<D> = duplicate(Builder(operation))
+
+  fun copy(operation: Operation<D>): ApolloRequest<D> = duplicate(Builder(operation)).build()
```
martinbonnin, line 28: Thanks for sending this ! I understand the use case but I'd like to avoid the "copy" vocabulary here because it hints heavily at data classes which are not really used here.

Instead, maybe use `newBuilder(operation: Operation<E>)`? It also makes it more discoverable?

```kotlin
  fun newBuilder(): Builder<D> {
    return newBuilder(operation)
  }

  fun <E: Operation.Data> newBuilder(operation: Operation<E>): Builder<E> {
```

martinbonnin:
Hi 👋 . I took the liberty of pushing `newBuilder(operation)` as we have a new release coming up. It's `@ApolloExperimental` so we can always change it if needed.

BoD (approved):
👍

Issue #3964, opened 2022-03-25
Do not crash if the previous record cannot be parsed


martinbonnin:
The sqlite cache `merge` first reads the existing record before overwriting it. This is done so that we can compare by value and only publish cache keys that have really changed (and avoid triggering all watchers all the time).

When the initial read fails because of an IO error or because the data was corrupted, the app currently crashes while it's technically ok to continue and just publish all the cache keys.

This issue is to catch this error and try to recover.

BoD:
Done in  #3989.

Issue #3963, opened 2022-03-25
[Cache] Expose global hook for cache write errors


martinbonnin:
Although very unlikely, there can be rare occurrences where a cache write fails. Especially when using SQLite, there can be errors writing to persistent storage or anything else. Because writing to the cache can be asynchronous, we can't expose these errors to the caller. 

When this happens, we currently crash with an uncaught exception which is not ideal. 

Instead, we could add a static `ApolloCacheInterceptor.errorHandler: (Throwable) -> Unit` so that these errors can be logged for diagnostics purposes. We could also make it log to "stdout" by default.

BoD:
In  #3989 we went with a globally usable `ApolloExceptionHandler` (marked experimental for the time being).

PR #3989, opened 2022-04-04
Add exception handlers to ApolloCacheInterceptor and SqlNormalizedCache

BoD:
This handles #3963 and #3964 

Adds an optional `asyncExceptionHandler` parameter to `ApolloCacheInterceptor` (default value is to log).

In `SqlNormalizedCache`, in addition to treating read fails as cache misses, I ended up also adding another exception handler so this case can be logged - but I’m not convinced it’s worth it. Let me know what you think!


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -108,10 +117,15 @@ fun ApolloClient.Builder.logCacheMisses(
   return addInterceptor(CacheMissLoggingInterceptor(log))
 }
 
-fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: Boolean = false): ApolloClient.Builder {
+@JvmOverloads
+fun ApolloClient.Builder.store(
+    store: ApolloStore,
+    writeToCacheAsynchronously: Boolean = false,
+    asyncExceptionHandler: (Throwable) -> Unit = DEFAULT_ASYNC_EXCEPTION_HANDLER,
```
martinbonnin, line 124: Same comment as above. Do we need this to be surfaced in the main API?

PR #3995, opened 2022-04-05
Add configurable headers to ws network transport

sproctor:
I changed the WebSocketEngine open method to take a list of HttpHeader. It might be a breaking change, I can switch it back to a map if you think it's a bad idea. I thought having a consistent interface between the HttpEngine and WebSocketEngine would make sense.

I couldn't test changes for apple because I don't have access to a Mac currently.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -20,9 +20,6 @@
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="50" />
       <option name="ALLOW_TRAILING_COMMA" value="true" />
     </JetCodeStyleSettings>
-    <XML>
-      <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
-    </XML>
     <codeStyleSettings language="GraphQL">
       <indentOptions>
         <option name="INDENT_SIZE" value="2" />
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -20,6 +20,9 @@
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="50" />
       <option name="ALLOW_TRAILING_COMMA" value="true" />
     </JetCodeStyleSettings>
+    <XML>
+      <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
+    </XML>
     <codeStyleSettings language="GraphQL">
       <indentOptions>
         <option name="INDENT_SIZE" value="2" />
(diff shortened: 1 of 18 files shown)


BoD:
Hi! Thank you for this!

Indeed `HttpHeader` would be nicer but that would be a breaking change to the public API. We _could_ have 2 versions of `open`, one being deprecated which calls the new one - but I don't think it's worth the trouble.

Also, for our curiosity, can you tell us what is your use-case for needing this, is it for auth or something else? Thanks a lot!

sproctor:
Yes, it's for auth with Hasura. I talked to @martinbonnin about it a month ago, then got busy.

https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1645656182878689

Want me to revert to map or add the deprecation?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -173,9 +175,7 @@ private constructor(
             val webSocketConnection = try {
               webSocketEngine.open(
                   url = serverUrl,
-                  headers = mapOf(
-                      "Sec-WebSocket-Protocol" to protocolFactory.name,
-                  )
+                  headers = headers + HttpHeader("Sec-WebSocket-Protocol", protocolFactory.name),
```
martinbonnin, line 178: Should we only add the "Sec-WebSocket-Protocol" header if it's not present already? That'd allow callers to customize the header if really required? (not sure this will ever happen but it might?)

BoD:
> Yes, it's for auth with Hasura. I talked to @martinbonnin about it a month ago, then got busy.

Oh right! 😊 👍 

> Adding the deprecation should work? Means you have both...

That works for me 👍

BoD (approved):
👍

.idea/codeStyles/Project.xml
```diff
@@ -20,9 +20,6 @@
       <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="50" />
       <option name="ALLOW_TRAILING_COMMA" value="true" />
     </JetCodeStyleSettings>
-    <XML>
```
BoD, line 23: 👀

sproctor: Oof. I missed that in the list of changes, sorry. I hate it when the IDE does stuff randomly.

BoD: No worries 😅

martinbonnin (approved):

apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/ws/OkHttpWebSocketEngine.kt
```diff
@@ -105,4 +108,9 @@ actual class DefaultWebSocketEngine(
       }
     }
   }
+
+  @Deprecated("Use open(String, List<HttpHeader>) instead.", ReplaceWith("open(String, List<HttpHeader>)"))
```
martinbonnin, line 112:
```suggestion
  @Deprecated("Use open(String, List<HttpHeader>) instead.", ReplaceWith("open(url, headers)"))
```

I think the Kotlin intelliJ plugin is smart enough to replace the arguments values

martinbonnin: @sproctor want me to look into this?

sproctor:
Sorry, I hadn't used replaceWith before, I updated the other usages. Should the interface have a `replaceWith`? I removed it for now.

martinbonnin (approved):

PR #3999, opened 2022-04-06
⏰  Add tests for date adapters

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/3997

### DIFF
--- a/apollo-adapters/src/jvmTest/kotlin/JavaTimeAdaptersTest.kt
+++ b/apollo-adapters/src/jvmTest/kotlin/JavaTimeAdaptersTest.kt
@@ -66,7 +66,7 @@ class JavaTimeAdaptersTest {
   @Test
   fun localDate() {
     val localDate = JavaLocalDateAdapter.fromJson("2010-06-01")
-    assertEquals(1275430784, localDate.toEpochSecond(LocalTime.parse("22:19:44.475"), ZoneOffset.UTC))
+    assertEquals(1275430784, localDate.atTime(LocalTime.parse("22:19:44.475")).toEpochSecond(ZoneOffset.UTC))
     assertEquals("2010-06-01", JavaLocalDateAdapter.toJson(localDate))
   }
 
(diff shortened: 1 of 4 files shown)


apollo-adapters/src/jvmTest/kotlin/JavaTimeAdaptersTest.kt
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Test
+  fun instant() {
+    var instant = JavaInstantAdapter.fromJson("2010-06-01T22:19:44.475Z")
+    assertEquals(1275430784475, instant.toEpochMilli())
+    assertEquals("2010-06-01T22:19:44.475Z", JavaInstantAdapter.toJson(instant))
+
+    instant = JavaInstantAdapter.fromJson("2010-06-01T23:19:44.475+01:00")
+    assertEquals(1275430784475, instant.toEpochMilli())
```
BoD, line 42: Is it really true that `2010-06-01T22:19:44.475Z` and `2010-06-01T23:19:44.475+01:00` are the same instant `1275430784475`? Or does that mean that the offset is also ignored when parsing?

martinbonnin: I think it's true? Second one is 22:19 +1? So both are the same milliseconds offset from epoch (dates are hard...)

BoD: So sorry! Didn't see the `22` vs `23`. Yes it makes sense 👍 .

Too bad about the lost timezone on the writing side, but I don't think it should cause too much trouble.

BoD (approved):
👍

apollo-adapters/src/jvmTest/kotlin/JavaTimeAdaptersTest.kt
martinbonnin: I think it is fine. This is all used to convey "instant" so shouldn't be used to carry a future local date (I have no idea how one would do that)

Issue #4001, opened 2022-04-07
Handle IllegalStateException: attempt to re-open an already-closed object


olivierg13:
**Summary**
We have been seeing this crash (` IllegalStateException: attempt to re-open an already-closed object`) a few times per day in our application. At least we could handle it in `SqlNormalizedCache.merge()` so it doesn't crash the user's application.

**Version**
3.2.1

**Description**
Here is the full stacktrace:
```
java.lang.IllegalStateException: attempt to re-open an already-closed object: SQLiteDatabase: /data/user/0/com.netflix.mediaclient/databases/apollo_cache_6Q7OBFCHLRDLXJGF7BS3GP2FKA
        at android.database.sqlite.SQLiteClosable.acquireReference(SQLiteClosable.java:58)
        at android.database.sqlite.SQLiteDatabase.beginTransaction(SQLiteDatabase.java:569)
        at android.database.sqlite.SQLiteDatabase.beginTransactionNonExclusive(SQLiteDatabase.java:505)
        at androidx.sqlite.db.framework.FrameworkSQLiteDatabase.beginTransactionNonExclusive(FrameworkSQLiteDatabase:74)
        at com.squareup.sqldelight.android.AndroidSqliteDriver.newTransaction(AndroidSqliteDriver:88)
        at com.squareup.sqldelight.TransacterImpl.transactionWithWrapper(TransacterImpl:225)
        at com.squareup.sqldelight.TransacterImpl.notifyQueries(TransacterImpl:214)
        at com.squareup.sqldelight.TransacterImpl.transaction(TransacterImpl:214)
        at com.squareup.sqldelight.Transacter$DefaultImpls.transaction$default(Transacter:72)
        at com.apollographql.apollo3.cache.normalized.sql.internal.CacheQueriesHelpers.updateRecord(CacheQueriesHelpers:66)
        at com.apollographql.apollo3.cache.normalized.sql.SqlNormalizedCache.merge(SqlNormalizedCache:82)
        at com.apollographql.apollo3.cache.normalized.api.MemoryCache.merge(MemoryCache:113)
        at com.apollographql.apollo3.cache.normalized.api.MemoryCache.merge(MemoryCache:120)
        at com.apollographql.apollo3.cache.normalized.api.internal.OptimisticCache.merge(OptimisticCache:38)
        at com.apollographql.apollo3.cache.normalized.internal.DefaultApolloStore$writeOperationWithRecords$2.invoke(DefaultApolloStore:205)
[...]

PR #4002, opened 2022-04-07
Handle SqlNormalizedCache merge APIs Exceptions with ApolloExceptionHandler

olivierg13:
Handle merge exceptions that could happen as described in #4001, so it doesn't crash the application andwe can use the exception handler to still investigate these issues.


martinbonnin (approved):
Thanks!

BoD (approved):

Issue #4004, opened 2022-04-07
Apollo Unable to generate classes for operations with Default Arguments on v3.2.0


AnkitPatanaik:
**Summary**
Hi again! I'm running into an issue where Apollo is unable to generate classes when there are default arguments in my Query/Mutation. It specifically throws this error: 

```
Caused by: java.lang.IllegalArgumentException: couldn't make a guess for [com.apollographql.relocated.cl](http://com.apollographql.relocated.cl/)
```

**Version**
3.2.0, things were working as expected on 3.1.0

**Description**

**These queries don't work**

_Mandatory Arguments, no defaults_
```
mutation MyMutation($firstName: CustomScalar!, $lastName: CustomScalar!) {
  my_put_operation(firstName: $firstName, lastName: $lastName)
}
```
_Hardcoded default at field level w/ mandatory arguments_
```
mutation MyMutation($firstName: CustomScalar!, $lastName: CustomScalar!) {
  my_put_operation(firstName: "MyDefault123" lastName: $lastName)
}
```

**The following don't work**

_Default for non-mandatory argument for mutation_
```
mutation MyMutation($firstName: CustomScalar = "MyDefault123", $lastName: CustomScalar!) {
  my_put_operation(firstName: $firstName, lastName: $lastName)
}
```
_Non-Mandatory argument, default value hardcoded in field_
```
mutation MyMutation($firstName: CustomScalar , $lastName: CustomScalar!) {
  my_put_operation(firstName: "MyDefault123" lastName: $lastName)
}
```

**Here's the full stack trace for more context:**
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':subsystem:myFeature:generateServiceApolloSources'.
> couldn't make a guess for com.apollographql.relocated.cl

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 34s
53 actionable tasks: 3 executed, 50 up-to-date

Publishing build scan...

ERRORS (1)

FAILED TASK: :subsystem:myFeature:generateServiceApolloSources
  Description: org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':subsystem:myFeature'.

    Caused by: java.lang.IllegalArgumentException: couldn't make a guess for com.apollographql.relocated.cl
      // ...
      at com.apollographql.relocated.d4.a(SourceFile:7)
[...]

PR #4008, opened 2022-04-08
Use a constant for JvmOverloads to avoid a crash due to relocation

BoD:
A fix for #4004.

There were also a few other instances or classes used directly, which didn't seem to cause trouble, but also fixed here for consistency.


martinbonnin (approved):
👍  I think the JDK classes (like `java.lang.Override` and `java.lang.String`) are not relocated because they're not part of the fat jar so that'd explain why this has worked so far.

Doesn't hurt to not use them reflectively though 👍

martinbonnin:
CI fails with

```
fullstack/LaunchDetailsQuery.kt: (22, 19):  Declaration annotated with '@OptionalExpectation' can only be used in common module sources
```

BoD:
Nothing's ever easy is it? 😅
I'm guessing this is due to the specific wiring in `integration-tests` to have both Java and Kotlin codegens?

I moved the test to a specific submodule instead - JVM only but also tried locally on a MP one which also didn't have the issue.

martinbonnin:
> I'm guessing this is due to the specific wiring in integration-tests to have both Java and Kotlin codegens?

How would that be an issue? Java codegen is not impacted and Kotlin should compile in all cases?

BoD:
> How would that be an issue? Java codegen is not impacted and Kotlin should compile in all cases?

That was kind of a wild guess, but in fact I just noticed it's `compileTestKotlinJs` that's failing 🤔. Will investigate more.

BoD:
I can repro `Declaration annotated with '@OptionalExpectation' can only be used in common module sources` by setting this on my test project:

```
  outputDirConnection {
    connectToKotlinSourceSet("jsTest")
    connectToKotlinSourceSet("jvmTest")
    connectToKotlinSourceSet("appleTest")
  }
```
 
but to be honest I'm still bewildered

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -177,12 +176,13 @@ class Options(
     val generateOptionalOperationVariables: Boolean = defaultGenerateOptionalOperationVariables,
 
     /**
-     * Whether to generate kotlin constructors with @JvmOverloads for more graceful Java interop experience when default values are present.
+     * Whether to generate kotlin constructors with `@JvmOverloads` for more graceful Java interop experience when default values are present.
+     * Note: when enabled in a multi-platform setup, the generated code can only be used in the common or JVM sourcesets.
```
martinbonnin, line 180: 👍

martinbonnin (approved):

PR #4009, opened 2022-04-08
🗃️ initial pass at a CacheControl document

martinbonnin:


design-docs/CacheControl.md
```diff
@@ -0,0 +1,138 @@
...
+```graphql
+extend type Session @cacheControl(maxAge: 600) {
+  startDate: Date! @cacheControl(maxAge: 30)
+  endDate: Date! @cacheControl(maxAge: 30)
+}
+
+extend type Speaker @cacheControl(maxAge: 3600) 
+```
+
+Or if introspection ever supports directive usages, it could be retrieved (and cached) at runtime.
```
BoD, line 64: TIL that it's not supported 😢

design-docs/CacheControl.md
```diff
@@ -0,0 +1,138 @@
...
+
+When writing a record, the Date timestamp can be passed in `CacheHeaders` without changing the current API:
+
+```kotlin
+abstract fun merge(record: Record, cacheHeaders: CacheHeaders): Set<String>
+```
+
+To keep the database size manageable, the Records can be serialized in binary format, skipping the Json overhead and compressing timestamps when the same timestamp is used for multiple fields.
+
+To avoid storing another Long in the DB, it is proposed to define cache control rules in the client. Ideally using directives:
```
BoD, line 98: Sorry I'm not sure I understand this sentence: we still need to store the dates in the db right?

martinbonnin: Yes, `date` but not `date + maxAge` (so it saves 64bits). I'll update that part.

design-docs/CacheControl.md
```diff
@@ -0,0 +1,138 @@
...
+## Storing the cache control information
+
+We can think of several fields to store (loosely related to their equivalent HTTP definitions):
+
+- Date: the date the data originated (details TBC: created vs sent vs receive vs other?)
+- MaxAge: the maximum age before the data becomes stale
+- StaleWhileRevalidate
+- ...
+
+Date sounds required in all cases. MaxAge is required if we want to store that information when coming from the server but could be spared if the rules are defined on the client side.
```
ynkm169, line 75: > Date sounds required in all cases.
Ideally date is optional. If server does not send it, client can generate the "received" date.

design-docs/CacheControl.md
```diff
@@ -0,0 +1,138 @@
...
+
+Because any operation can query any field, this information needs to be stored per-field. This is the hard part. Adding one (or several) 64 bit timestamp for each field might double (or worse) the size of the cache. On the other hand a lot of fields will share the same timestamp, opening opportunities for compression (see #possible-optimisations).
```
ynkm169, line 77: > this information needs to be stored per-field
Question, does this mean the following code need to store expiry for all 3 fields? (name, bio, twitter)

```
type Speaker @cacheControl(maxAge: 3600) {
  name: String!
  bio: String!
  twitter: String
}
```

design-docs/CacheControl.md
```diff
@@ -0,0 +1,138 @@
...
+
+Similarly, updating one field in a Record requires rewriting all the Record. Storing fields separately might help.
+
+### Tweaking memory caches
+
+We currently recommend chaining a memory cache and a SQLite cache. While this works well, this adds a 3rd memory cache:
+
+- OS I/O cache
+- SQLite page cache
+- Apollo Memory cache
```
ynkm169, line 136: Clarification question:
If we don't use SQL cache at all, will we be able to use expiry in just the memory cache?

PR #4022, opened 2022-04-12
🐘  Fix the Moshi adapter used for OperationOutput. 

martinbonnin:
Moshi cannot get the type parameters from the typealias automagically


BoD (approved):
👍

PR #4033, opened 2022-04-14
Kotlin 1.6.21 & HMPP

martinbonnin:
- use Kotlin 1.6.20
- make ios-test pass
- better workaround

See also https://github.com/apollographql/apollo-kotlin/issues/4019


build-logic/src/main/kotlin/TargetCompatibility.kt
```diff
@@ -10,13 +10,7 @@ fun Project.configureJavaAndKotlinCompilers() {
     // For Kotlin JVM projects
     tasks.withType(KotlinCompile::class.java) {
       it.kotlinOptions {
-        freeCompilerArgs = freeCompilerArgs +
-            "-Xopt-in=kotlin.RequiresOptIn" +
-            // This is a workaround for https://youtrack.jetbrains.com/issue/KT-47000 (fixed in Koltin 1.6.20)
-            // Since we don't use @JvmDefault anywhere, the option has no effect, but suppresses the bogus compiler error
-            // See also https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/
-            // See also https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/
-            "-Xjvm-default=compatibility"
```
martinbonnin, line 19: @BoD is it ok to remove this?

martinbonnin:
Ouch. This seems to break the IDE a bunch :/. We might want to wait a bit before we enable this...

Screenshot 2022-04-14 at 11 52 57

build-logic/src/main/kotlin/TargetCompatibility.kt
BoD: Yes, should be good 👍

tests/build.gradle.kts
```diff
@@ -28,6 +28,21 @@ subprojects {
     }
   }
 
+  // Workaround for https://youtrack.jetbrains.com/issue/KT-51970
```
BoD, line 31: _Oh my..._ 😅

BoD (approved):

martinbonnin:
Just added okio 3.1.0 and HMPP support but somehow IDE is still not working. I opened this as a follow up https://youtrack.jetbrains.com/issue/KT-52059/

gradle.properties
```diff
@@ -21,5 +21,4 @@ org.gradle.jvmargs=-Xmx8g
 # silence the "multiplatform alpha" warning
 kotlin.mpp.stability.nowarn=true
 
-# https://github.com/gradle/gradle/issues/11412
-systemProp.org.gradle.internal.publish.checksums.insecure=true
```
martinbonnin, line 25: Another workaround bites the dust. This shouldn't be needed anymore now that we're publishing to MavenCentral only

martinbonnin:
On hold for https://youtrack.jetbrains.com/issue/KTIJ-21583

martinbonnin:
Marked as ready for review again as we don't need to wait for https://youtrack.jetbrains.com/issue/KTIJ-21583. We can actually build with HMPP-ready `1.6.20` and develop in IDEA with `1.6.10` based on the `idea.sync.active` property. It's not pretty but it works!

BoD (approved):
Hope we can remove the workarounds soon 🤞 but it's great that it works! 🎉

martinbonnin:
> Hope we can remove the workarounds soon 🤞 but it's great that it works! 🎉
🤞

PR #4038, opened 2022-04-15
☁️  add HttpRequest.newBuilder(url, method)

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/4037

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
@@ -41,7 +41,7 @@ private constructor(
 ) {
 
   @JvmOverloads
-  fun newBuilder(url: String = this.url, method: HttpMethod = this.method) = Builder(
+  fun newBuilder(method: HttpMethod = this.method, url: String = this.url) = Builder(
       method = method,
       url = url,
   ).apply {
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
@@ -4,6 +4,7 @@ import okio.Buffer
 import okio.BufferedSink
 import okio.BufferedSource
 import okio.ByteString
+import kotlin.jvm.JvmOverloads
 
 enum class HttpMethod {
   Get, Post
@@ -39,7 +40,8 @@ private constructor(
     val body: HttpBody?,
 ) {
 
-  fun newBuilder() = Builder(
+  @JvmOverloads
+  fun newBuilder(url: String = this.url, method: HttpMethod = this.method) = Builder(
       method = method,
       url = url,
   ).apply {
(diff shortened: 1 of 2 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
```diff
@@ -39,7 +40,8 @@ private constructor(
     val body: HttpBody?,
 ) {
 
-  fun newBuilder() = Builder(
+  @JvmOverloads
+  fun newBuilder(url: String = this.url, method: HttpMethod = this.method) = Builder(
```
kevincianfarini, line 44: Do you think it would make more sense to expose an API like so? 

```kt
val newRequest = request.newBuilder()
  .url(new = "someUrl")
  .method(new = someHttpMethod)
  .addHeader(...)
  .body(...)
  .build()
```

It feels weird that we'd have two different ways of configure the builder.

kevincianfarini: The only reason I exposed them in a `newBuilder` like function in my workaround was because it was the only place I had the ability to configure them.

martinbonnin: > Do you think it would make more sense to expose an API like so?

Good question. I did it like this because I'm spending some time in KotlinPoet these days and [this is how they do it](https://github.com/martinbonnin/kotlinpoet/blob/61fcce3d73c555dda42defc17a3281f7fb60037d/kotlinpoet/src/main/java/com/squareup/kotlinpoet/FileSpec.kt#L231). But on the other hand, OkHttp is doing [the opposite thing](https://github.com/square/okhttp/blob/master/okhttp/src/commonMain/kotlin/okhttp3/Request.kt#L47)

No strong opinion here but whatever we choose we must stick to it. I'd lean towards the KotlinPoet way to do it because `okhttp3.Request` also doesn't have a mandatory `url` parameter to the Builder constructor. So it's all a runtime thing.

kevincianfarini: I don't have much of a strong opinion, either. I think this is good!

BoD: It makes sense to me to have the parameters in `newBuilder` as it's consistent with the builder's constructor 👍

BoD (approved):

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
BoD, line 44: (micro nitpick: they're not in the same order, in the Builder, you have `method` first)

```suggestion
  fun newBuilder(method: HttpMethod = this.method, url: String = this.url) = Builder(
```

martinbonnin: Good catch, thanks!

Issue #4054, opened 2022-04-20
(V3 Migration) okhttp.OkHttpClient: <-- HTTP FAILED: java.lang.IllegalStateException: closed 


kodekid20:
Hi, I am migrating to V3, but keep getting the following error when uploading file(s).
>okhttp.OkHttpClient: <-- HTTP FAILED: java.lang.IllegalStateException: closed 

In version 2, everything works fine based on my implementation below: 
```kotlin
apolloClient.mutate(
     CreateProjectMutation(
          ...
          video = FileUpload(mimetype = type, filePath = parameters.video),
          thumbnail = FileUpload(mimetype = thumbType, filePath = parameters.thumbnail)
       )
 )
```

This is my implementation in V3:
```kotlin
val upload = DefaultUpload.Builder()
      .content(File(parameters.video)) // using DefaultUpload.Builder content ext
      .contentType(type)
      .build()

val thumbnail = DefaultUpload.Builder()
      .content(File(parameters.thumbnail))
      .contentType(thumbType)
      .build()

apolloClient.mutation(
     CreateProjectMutation(
          ...
          video = upload,
          thumbnail = thumbnail
       )
 )
```

For reference, here is my mutation
```graphql
mutation CreateProject(
    ...
    $thumbnail: Upload!,
    $video: Upload!
) ...
```

Here's the network log
```okhttp
I/okhttp.OkHttpClient: Content-Type: multipart/form-data; boundary=44a3c556-0e96-44e1-824a-f3bfa67322a3
I/okhttp.OkHttpClient: X-APOLLO-OPERATION-ID: 3ad17cbe8d5f05812ba6d19dabd9f896304b6cb1808d105b0697ee188a074dc9
I/okhttp.OkHttpClient: X-APOLLO-OPERATION-NAME: CreateProject
I/okhttp.OkHttpClient: Authorization: JWT eyJ0eXAiOiJKV1QiLCJhb...
I/okhttp.OkHttpClient: apollographql-client-name: Android
I/okhttp.OkHttpClient: apollographql-client-version: 1.0.0
...
I/okhttp.OkHttpClient: �e��T��o�~����VJ/6�y{�_a�Op����/���5���.����˂�Η���˷)�i�Vj�"�Ô��3q�2a�J
I/okhttp.OkHttpClient: ZZZZZZZ^
I/okhttp.OkHttpClient: --> END POST (-1-byte body)
[...]

PR #4056, opened 2022-04-21
Provide a Content-Length when using Upload

BoD:
Fix for #4054

### DIFF
--- a/tests/integration-tests/src/jvmTest/resources/file1.jpg
+++ b/tests/integration-tests/src/jvmTest/resources/file1.jpg
@@ -1,0 +1,1 @@
+Supposedly an image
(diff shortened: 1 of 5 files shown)


tests/integration-tests/src/jvmTest/resources/file1.jpg
```diff
@@ -0,0 +1 @@
+Supposedly an image
```
martinbonnin, line 1: 😄

martinbonnin (approved):
👍 

Now that okio has multiplatform Filesystem, we could also add `DefaultUpload.content(okio.Path)`. But that can be done in a later PR

Issue #3884, opened 2022-02-21
@defer: handle incremental delivery at GraphQL level (Subscriptions)


BoD:
In `WebSocketNetworkTransport`, combine and emit responses (including data, errors, and extensions) as incremental payloads are received.

PR #4058, opened 2022-04-22
defer: Merge payloads in subscriptions

BoD:
This adds support for `@defer` in subscriptions.

Alas! I couldn't make a working unit test using the embedded Helix at this point, so the tests rely on running an external Helix instance manually.

Related to #3884

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
@@ -2,6 +2,7 @@ package com.apollographql.apollo3.network.ws
 
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_0_1
+import com.apollographql.apollo3.annotations.ApolloInternal
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.CustomScalarAdapters
@@ -9,8 +10,11 @@ import com.apollographql.apollo3.api.Operation
 import com.apollographql.apollo3.api.http.HttpHeader
 import com.apollographql.apollo3.api.json.jsonReader
 import com.apollographql.apollo3.api.parseJsonResponse
+import com.apollographql.apollo3.api.withDeferredFragmentIds
 import com.apollographql.apollo3.exception.ApolloNetworkException
 import com.apollographql.apollo3.internal.BackgroundDispatcher
+import com.apollographql.apollo3.internal.DeferredJsonMerger
+import com.apollographql.apollo3.internal.isDeferred
 import com.apollographql.apollo3.internal.transformWhile
 import com.apollographql.apollo3.network.NetworkTransport
 import com.apollographql.apollo3.network.ws.internal.Command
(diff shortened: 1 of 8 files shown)


apollo-api/api/apollo-api.api
```diff
@@ -17,6 +17,9 @@ public final class com/apollographql/apollo3/api/AdapterContext$Builder {
 	public final fun variables (Lcom/apollographql/apollo3/api/Executable$Variables;)Lcom/apollographql/apollo3/api/AdapterContext$Builder;
 }
 
+public final class com/apollographql/apollo3/api/AdapterContextKt {
+}
```
martinbonnin, line 21: Interesting that this is adding a public class given that all its symbols should be internal IIUC. Maybe rename to `-AdapterContext.kt` to make sure it's never reachable in Java ?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -238,9 +242,12 @@ private constructor(
     }
   }
 
+  @OptIn(ApolloInternal::class)
   override fun <D : Operation.Data> execute(
       request: ApolloRequest<D>,
   ): Flow<ApolloResponse<D>> {
+    val deferredJsonMergers = mutableMapOf<String, DeferredJsonMerger>()
```
martinbonnin, line 249: Could you use a `val deferredJsonMerger: DeferredJsonMerger` here? This coroutine will only ever see this request so it shouldn't need to care about others?

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -266,18 +273,34 @@ private constructor(
...
+        is OperationResponse -> {
+          val responsePayload = response.payload
+          val requestCustomScalarAdapters = request.executionContext[CustomScalarAdapters]!!
+          val (payload, customScalarAdapters) = if (responsePayload.isDeferred()) {
+            val responseId = response.id ?: ""
+            val deferredJsonMerger = deferredJsonMergers.getOrPut(responseId) { DeferredJsonMerger() }
+            deferredJsonMerger.merge(responsePayload)
+            if (responsePayload["hasNext"] == false) {
+              // Last deferred payload: discard the deferredJsonMerger
+              deferredJsonMergers.remove(responseId)
```
martinbonnin, line 287: Bonus point: you don't need this housekeeping if you have only one deferredJsonMerger. The GC should do all the work.

tests/defer/build.gradle.kts
```diff
@@ -13,6 +13,17 @@ kotlin {
     withJava()
   }
 
+  js {
+    nodejs {
+      testTask {
+        useMocha {
+          // Override default timeout (needed for subscriptions tests)
+          timeout = "120s"
```
martinbonnin, line 21: At this point, we could consider making this the default everywhere 😅

BoD: Done! :)

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
BoD: You're totally right! It's a lot simpler like this. (Still need to reset it after the last payload, so I added a reset method)

PR #4067, opened 2022-04-27
chore fixes link to change log

jpvajda:
this link reference was wrong, and instead of linking it to the file itself,  I switched it to always point to main's change link via a direct link.


martinbonnin (approved):
👍

PR #4068, opened 2022-04-28
🔒  deprecate BearerTokenInterceptor and provide tests and docs instead

martinbonnin:


docs/source/advanced/authentication.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Authenticating your HTTP requests with OkHttp
+
+[OkHttp Interceptors](https://square.github.io/okhttp/features/interceptors/) are an easy way to add an `"Authorization"` header to your HTTP requests.
+
+OkHttp Interceptors have been around for a long time and work well but only work on Android and the JVM.
+
+## Authenticating your HTTP requests with Apollo HttpInterceptor
+
+In order to authenticate your HTTP requests in a multi-platform way, you can use Apollo [HttpInterceptor](./interceptors-http).
```
BoD, line 15: Nitpick
```suggestion
In order to authenticate your HTTP requests in a multi-platform way, you can use an Apollo [HttpInterceptor](./interceptors-http).
```

docs/source/advanced/authentication.mdx
```diff
@@ -0,0 +1,75 @@
...
+      token = mutex.withLock {
+        // get new token
+      }
+      chain.proceed(request.newBuilder().addHeader("Authorization", "Bearer $token").build())
+    } else {
+      response
+    }
+  }
+}
+```
```
BoD, line 40: What about something like "For a more advanced example, you can have a look at (link to AuthorizationInterceptor)"?

apollo-runtime/src/commonTest/kotlin/com/apollographql/apollo3/network/AuthorizationInterceptor.kt
```diff
@@ -0,0 +1,164 @@
...
+     * @param oldToken the previous token or null if there was none
+     * @return a new token from the oldToken
+     */
+    suspend fun refreshToken(oldToken: String?): Token
+  }
+
+  /**
+   * A token link in the chain
+   */
+  class TokenLink(
```
BoD, line 62: I'm not quite sure I follow why we need to keep a chain of tokens, since only the most recent one is supposed to be valid?

martinbonnin: That's because of this test: https://github.com/apollographql/apollo-kotlin/pull/4068/files#diff-54b37c733642763f10a821c0cec37d9562c5c4ec48f24c1e499100b13e9515d5R162

In theory, you can have a very long request blocked on a very old token. If other concurent requests trigger mutliple refreshes, you need to "follow the link" to get the latest one or else you're taking the risk of generating an extra token.

It's all very theoretical but the "maxSize == 1" is "just" a special case

BoD: I see! Thanks :)

docs/source/advanced/authentication.mdx
BoD: Perfect 👍

BoD (approved):

PR #4070, opened 2022-04-29
Enable Multiplatform new memory manager

BoD:


martinbonnin:
Should we close this one?

BoD:
> Should we close this one?

Yeah, we can resurrect it when we fell ready 😊

PR #4073, opened 2022-05-02
Chunked support on Apple in specific HttpEngine

BoD:
Adds a new `StreamingNSURLSessionHttpEngine` that supports chunked. This only works with the new MM however (which this PR also enables).


apollo-mpp-utils/src/appleMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -1,24 +1,24 @@
...
 import platform.posix.pthread_self
 import kotlin.native.concurrent.ensureNeverFrozen
 import kotlin.native.concurrent.freeze
 import kotlin.native.concurrent.isFrozen
+import kotlin.system.getTimeMillis
 
 actual fun currentTimeMillis(): Long {
-  val nanoseconds: Long = dispatch_time(DISPATCH_TIME_NOW, 0).convert()
-  return nanoseconds / 1_000_000L
+  return getTimeMillis()
```
martinbonnin, line 11: TIL. 👍

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
```diff
@@ -0,0 +1,313 @@
...
+                data = httpDataSource,
+                httpResponse = response,
+                error = null,
+            )
+        )
+      }
+
+      override fun onData(data: NSData) {
+        httpDataSink.write(data.toByteString())
+        httpDataSink.flush()
```
martinbonnin, line 125: Out of curiosity, is this called on every chunk? Every x kB? Something else? How much do we need to flush?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
```diff
@@ -0,0 +1,313 @@
...
+ * - didCompleteWithError (with a non-null error)
+ */
+private class StreamingDataDelegate : NSObject(), NSURLSessionDataDelegateProtocol {
+  interface Handler {
+    fun onResponse(response: NSHTTPURLResponse)
+    fun onData(data: NSData)
+    fun onComplete(error: NSError?)
+  }
+
+  private val handlers = mutableMapOf<NSURLSessionTask, Handler>()
```
martinbonnin, line 207: I'm guessing you could create a new nsUrlSession and a new NSURLSessionDataDelegateProtocol for every execute() call and that'd save this indirection and make things a bit more immutable ?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
```diff
@@ -0,0 +1,313 @@
...
+    override fun flush() {}
+    override fun timeout() = Timeout.NONE
+  }
+
+  val source: Source = object : Source {
+    override fun read(sink: Buffer, byteCount: Long): Long {
+      pthread_mutex_lock(mutex.ptr)
+      while (buffer.size == 0L && !isSinkClosed) {
+        pthread_cond_wait(cond.ptr, mutex.ptr)
+      }
```
martinbonnin, line 300: return -1 if the buffer is empty and the sink is closed?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
```diff
@@ -0,0 +1,313 @@
...
+      val readCount = buffer.read(sink, byteCount)
+      pthread_mutex_unlock(mutex.ptr)
+      return readCount
+    }
+
+    override fun close() {
+      pthread_mutex_destroy(mutex.ptr)
+      pthread_cond_destroy(cond.ptr)
```
martinbonnin, line 308: This is potentially dangerous if the source is closed while the request is still ongoing. I don't really see a way around waiting for both the sink & the source to be closed before releasing the mutex & cond. That's more hoops but I don't think we can make the assumption that the source is going to be closed last?

build-logic/src/main/kotlin/Mpp.kt
```diff
@@ -42,6 +38,8 @@ fun Project.configureMppDefaults(withJs: Boolean = true, withLinux: Boolean = tr
     )
 
     addTestDependencies(withJs)
+
+    enableNewMemoryManager()
```
martinbonnin, line 42: I think this one is useless. The new memory model is only used on binaries but all the modules calling into this should produce libraries, not binaries.

build-logic/src/main/kotlin/Mpp.kt
```diff
@@ -122,6 +120,8 @@ fun Project.configureMppTestsDefaults(withJs: Boolean = true) {
     configureAppleTargets("macosX64", "macosArm64")
 
     addTestDependencies(withJs)
+
+    enableNewMemoryManager()
```
martinbonnin, line 124: Can we only enable it for defer tests? It's important to keep testing the old memory model if we continue maintaining it. (And I think we should given that [some users reported pretty bad performance at the moment](https://twitter.com/agLuque/status/1519408400907522051))

tests/defer/src/appleTest/kotlin/test/DeferAppleStreamingTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.apollographql.apollo3.mockserver.MockServer
+import com.apollographql.apollo3.mockserver.enqueueMultipart
+import com.apollographql.apollo3.mpp.currentTimeMillis
+import com.apollographql.apollo3.testing.runTest
+import defer.WithFragmentSpreadsQuery
+import kotlinx.coroutines.flow.collect
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+// TODO Move this test to the common test suite when chunked support works on JS
```
martinbonnin, line 14: You could move this code to commonTest now and return early if the test is running on JS. We have this in a few other places like https://github.com/apollographql/apollo-kotlin/blob/c4c605e84cb1006d8b81936886cbaed49ae0d262/tests/models-response-based/src/commonTest/kotlin/test/ParseResponseBodyTest.kt#L60

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
BoD: From what I see in my tests, it's called once per chunk, so flushing here makes the payload available to the reader

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
BoD: That would simplify the code but I think that we're not supposed to create a new session per request, for performance - a bit like we wouldn't create a new OkHttpClient every time. For instance I guess it can keep/re-use certain connections alive.

[Here](https://developer.apple.com/forums/thread/84663) I found an answer recommending against it.

build-logic/src/main/kotlin/Mpp.kt
BoD: Oh! Will try it.

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
BoD: I think this already works because `buffer.read(sink, byteCount)` will return -1 in this case unless I'm missing something?

build-logic/src/main/kotlin/Mpp.kt
BoD: Yes you're totally right 😁

build-logic/src/main/kotlin/Mpp.kt
BoD: Good point! I've enabled it only on the project (+ added a small readme to make it more visible).

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
martinbonnin: Makes sense 👍 . Maybe add a comment up there?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/StreamingNSURLSessionHttpEngine.kt
martinbonnin: Gotcha 👍 .

martinbonnin (approved):

Issue #3981, opened 2022-03-31
@defer: don't block in CachingHttpInterceptor


BoD:
When using `@defer` the expectation is that the deferred parts are surfaced as soon as they arrive.

However, the current implementation of `CachingHttpInterceptor` [is reading](https://github.com/apollographql/apollo-kotlin/blob/fb19d466996db800147ae58a1a99567eaaed417d/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt#L75) the whole `HttpResponse` in order to cache it, before returning it. In effect this waits for all the chunks to be received, and so they're not emitted as they go.

A possible solution would be to have a body 'delegate' which writes to the cache as it is read. An example of this existed in v2 [here](https://github.com/apollographql/apollo-kotlin/blob/release-2.x/apollo-http-cache/src/main/java/com/apollographql/apollo/cache/http/ResponseBodyProxy.java).

Note: this is also true of `LoggingInterceptor` albeit since this is used as a debugging tool, this can be considered OK (adding a warning on the doc would be nice).

BoD:
Done in #4096

PR #4076, opened 2022-05-03
Do not buffer entire body in Http Cache

BoD:
See #3981

### DIFF
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
@@ -1,16 +1,27 @@
 package com.apollographql.apollo3.cache.http
 
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_2_3
 import com.apollographql.apollo3.api.http.HttpResponse
 import java.io.IOException
 
 interface ApolloHttpCache {
   fun read(cacheKey: String): HttpResponse
 
+  @Deprecated("Implementations should implement writeIncremental instead", ReplaceWith("writeIncremental(cacheKey, response)"))
+  @ApolloDeprecatedSince(v3_2_3)
+  fun write(response: HttpResponse, cacheKey: String)
+
   /**
    * Store the [response] with the given [cacheKey] into the cache.
-   * Note: the response's body is not consumed nor closed.
+   * A new [HttpResponse] is returned whose body, when read, will write the contents to the cache.
+   * The response's body is not consumed nor closed.
    */
-  fun write(response: HttpResponse, cacheKey: String)
+  fun writeIncremental(response: HttpResponse, cacheKey: String): HttpResponse {
+    @Suppress("DEPRECATION")
+    write(response, cacheKey)
+    return response
+  }
 
   @Throws(IOException::class)
   fun clearAll()
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
@@ -1,27 +1,17 @@
 package com.apollographql.apollo3.cache.http
 
-import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
-import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_2_3
 import com.apollographql.apollo3.api.http.HttpResponse
 import java.io.IOException
 
 interface ApolloHttpCache {
   fun read(cacheKey: String): HttpResponse
 
-  @Deprecated("Implementations should implement writeIncremental instead", ReplaceWith("writeIncremental(cacheKey, response)"))
-  @ApolloDeprecatedSince(v3_2_3)
-  fun write(response: HttpResponse, cacheKey: String)
-
   /**
    * Store the [response] with the given [cacheKey] into the cache.
-   * A new [HttpResponse] is returned whose body, when read, will write the contents to the cache.
    * The response's body is not consumed nor closed.
+   * @return a new [HttpResponse] whose body, when read, will write the contents to the cache.
    */
-  fun writeIncremental(response: HttpResponse, cacheKey: String): HttpResponse {
-    @Suppress("DEPRECATION")
-    write(response, cacheKey)
-    return response
-  }
+  fun write(response: HttpResponse, cacheKey: String): HttpResponse
 
   @Throws(IOException::class)
   fun clearAll()
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
@@ -106,6 +106,10 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
 
   /**
    * A [Source] that writes to the given cache sink as it is read.
+   *
+   * It commits all successful reads, even if they do not read until EOF. This is so that we can cache Json with extra trailing whitespace.
+   * If an error happens when reading the original source or writing to the cache sink, the edit is aborted.
+   * The commit or abort is done on [close].
    */
   private class ProxySource(
       private val originalSource: Source,
@@ -115,57 +119,49 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
 
     private val buffer = Buffer()
     private var hasClosedAndCommitted: Boolean = false
+    private var hasReadError: Boolean = false
 
     override fun read(sink: Buffer, byteCount: Long): Long {
       val read = try {
         originalSource.read(buffer, byteCount)
       } catch (e: Exception) {
-        abortEdit()
+        hasReadError = true
         throw e
       }
 
       if (read == -1L) {
-        // We've read fully, commit the cache edit
-        closeAndCommitCache()
+        // We're at EOF
         return -1L
       }
       try {
         buffer.peek().readAll(this.sink)
       } catch (e: Exception) {
-        abortEdit()
+        hasReadError = true
       }
       try {
         sink.writeAll(buffer)
       } catch (e: Exception) {
-        abortEdit()
+        hasReadError = true
         throw e
       }
       return read
     }
 
     override fun close() {
-      closeAndCommitCache()
-      originalSource.close()
-    }
-
-    private fun closeAndCommitCache() {
       if (!hasClosedAndCommitted) {
         try {
           sink.close()
-          cacheEditor.commit()
+          if (hasReadError) {
+            cacheEditor.abort()
+          } else {
+            cacheEditor.commit()
+          }
         } catch (e: Exception) {
           // Silently ignore cache write errors
         } finally {
           hasClosedAndCommitted = true
         }
-      }
-    }
-
-    private fun abortEdit() {
-      try {
-        cacheEditor.abort()
-      } catch (e: Exception) {
-        // Silently ignore cache write errors
+        originalSource.close()
       }
     }
 
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
@@ -49,44 +49,11 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
   }
 
   /**
-   * This is not actually called by the current version of the Interceptor, but is kept for backward binary compatibility.
-   */
-  @Deprecated("Kept for backward binary compatibility")
-  override fun write(response: HttpResponse, cacheKey: String) {
-    val editor = cacheLock.read {
-      cache.edit(cacheKey)
-    } ?: return
-
-    try {
-      editor.newSink(ENTRY_HEADERS).buffer().use {
-        val map = mapOf(
-            "statusCode" to response.statusCode.toString(),
-            "headers" to response.headers.map { httpHeader ->
-              // Moshi doesn't serialize Pairs by default (https://github.com/square/moshi/issues/508) so
-              // we use a Map with a single entry
-              mapOf(httpHeader.name to httpHeader.value)
-            },
-        )
-        adapter.toJson(it, map)
-      }
-      editor.newSink(ENTRY_BODY).buffer().use {
-        val responseBody = response.body
-        if (responseBody != null) {
-          it.writeAll(responseBody.peek())
-        }
-      }
-      editor.commit()
-    } catch (e: Exception) {
-      editor.abort()
-    }
-  }
-
-  /**
    * Store the [response] with the given [cacheKey] into the cache.
    * A new [HttpResponse] is returned whose body, when read, will write the contents to the cache.
    * The response's body is not consumed nor closed.
    */
-  override fun writeIncremental(response: HttpResponse, cacheKey: String): HttpResponse {
+  override fun write(response: HttpResponse, cacheKey: String): HttpResponse {
     val editor = cacheLock.read {
       cache.edit(cacheKey)
     } ?: return response
--- a/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
+++ b/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
@@ -120,7 +120,7 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
       val read = try {
         originalSource.read(buffer, byteCount)
       } catch (e: Exception) {
-        cacheEditor.abort()
+        abortEdit()
         throw e
       }
 
@@ -129,8 +129,17 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
         closeAndCommitCache()
         return -1L
       }
-      buffer.peek().readAll(this.sink)
-      sink.writeAll(buffer)
+      try {
+        buffer.peek().readAll(this.sink)
+      } catch (e: Exception) {
+        abortEdit()
+      }
+      try {
+        sink.writeAll(buffer)
+      } catch (e: Exception) {
+        abortEdit()
+        throw e
+      }
       return read
     }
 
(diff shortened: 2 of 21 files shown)


apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
```diff
@@ -1,20 +1,31 @@
...
+  fun write(response: HttpResponse, cacheKey: String)
+
   /**
    * Store the [response] with the given [cacheKey] into the cache.
-   * Note: the response's body is not consumed nor closed.
+   * A new [HttpResponse] is returned whose body, when read, will write the contents to the cache.
+   * The response's body is not consumed nor closed.
    */
-  fun write(response: HttpResponse, cacheKey: String)
+  fun writeIncremental(response: HttpResponse, cacheKey: String): HttpResponse {
```
BoD, line 20: A bit unfortunate. Couldn't think of another way to do this.

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -104,6 +137,51 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
...
+      val buffer = Buffer()
+      val read = try {
+        originalBody.read(buffer, byteCount)
+      } catch (e: Exception) {
+        cacheEditor.abort()
+        throw e
+      }
+
+      if (read == -1L) {
+        // We've read fully, commit the cache edit
```
BoD, line 161: Not 100% sure this is needed.

tests/defer/src/jvmTest/kotlin/test/DeferJvmTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        """{"data":{"isColor":true},"path":["computers",1,"screen"],"hasNext":false,"label":"a"}""",
+    )
+
+    val delay = 200L
+    mockServer.enqueueMultipart(jsonList, chunksDelayMillis = delay)
+
+    var lastEmitTime = currentTimeMillis()
+    apolloClient.query(WithFragmentSpreadsQuery()).toFlow().collect {
+      // Allow a 10% margin for inaccuracies
+      assertTrue(currentTimeMillis() - lastEmitTime >= delay / 1.1)
```
martinbonnin, line 50: It'd be nice to have a lower level `MockServer` API where we enqueue the chunks one after the other and read them as we go. That'd remove the potentiality of flaky timing issues.

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
martinbonnin: Is that even backward compatible? If someone implements `ApolloHttpCache` using `3.2.2` and ends up using `3.3.0` at runtime, what's going to happen? 

The Kotlin bytecode fir [CachingHttpInterceptor](https://github.com/apollographql/apollo-kotlin/blob/4b75f69669c26c0b3367eac0f50fab9dbd816a63/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt#L113) shows 
```
    INVOKEVIRTUAL com/apollographql/apollo3/cache/http/DiskLruHttpCache.writeIncremental (Lcom/apollographql/apollo3/api/http/HttpResponse;Ljava/lang/String;)Lcom/apollographql/apollo3/api/http/HttpResponse;
```

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -104,6 +137,51 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
     }
   }
 
+  /**
+   * A [BufferedSource] that writes to the given cache sink as it is read.
+   */
+  private class WriteToCacheSource(
+      private val originalBody: Source,
+      private val cacheSink: Sink,
```
martinbonnin, line 145: Nit: what about proxy? 

```suggestion
  private class ProxySource(
      private val originalSource: Source,
      private val sink: Sink,
```

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -104,6 +137,51 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
...
+      private val cacheEditor: DiskLruCache.Editor,
+  ) : Source by originalBody {
```
martinbonnin, line 147: [Interface delegation is one of Andrey Breslav regrets in Kotlin design](https://discuss.kotlinlang.org/t/what-is-wrong-about-implementation-by-delegation/19873) so I'd prefer avoiding it. 

Especially in that case, looks like the only thing we need to proxy is `timeout`?

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
martinbonnin, line 145: Nit: what about proxy? 

```suggestion
   * A [Source] that writes to the given cache sink as it is read.
   */
   private class ProxySource(
      private val originalSource: Source,
      private val sink: Sink,
```

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
martinbonnin: The `DeferJvmTest` tests still pass for me without the `closeAndCommitCache()` can you add one that fails?

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -104,6 +137,51 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
...
+
+    private var hasClosedAndCommitted: Boolean = false
+
+    override fun read(sink: Buffer, byteCount: Long): Long {
+      val buffer = Buffer()
```
martinbonnin, line 152: Sounds like this could be optimized at some point but we can always see that later ™️ :)

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
BoD: After looking a bit more at what we do with this interface:
- `CachingHttpInterceptor` instantiates `DiskLruHttpCache` directly, we don't provide a way to provide another implementation
- the `ApolloClient.httpCache: ApolloHttpCache` extension exposes the interface, but I think the only useful use-cases is to call `clearAll()` on it?

Could that mean that the interface is not really needed, and we could instead expose an extension `fun ApolloClient.clearHttpCache()`?

If yes we could:
- mark the whole interface as `@Deprecated`
- not implement it in `DiskLruHttpCache`
- mark `ApolloClient.httpCache` as deprecated, and make it return an implementation that only has `clearAll` working and the other ones being no-op

tests/defer/src/jvmTest/kotlin/test/DeferJvmTest.kt
BoD: Oh yeah 😊  I'll have a look!

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -104,6 +137,51 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
...
+        throw e
+      }
+
+      if (read == -1L) {
+        // We've read fully, commit the cache edit
+        closeAndCommitCache()
+        return -1L
+      }
+      buffer.peek().readAll(cacheSink)
+      sink.writeAll(buffer)
```
martinbonnin, line 166: Sounds like something that can fail too? Should we abort the editor in that case?

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
BoD: I hesitated naming it more "generically" but because of the `cacheEditor` it's pretty specific to the cache - OTOH  `ProxySource` is pretty clear too 👍

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
martinbonnin: > Could that mean that the interface is not really needed

I guess users could still call `write()` manually to warm the cache programmatically

> return an implementation that only has clearAll working and the other ones being no-op

That will make the above (hopefully very rare) use case fail so at this point, we might as well add the `HttpResponse` return value. The only thing that could break is if someone calls `ApolloHttpCache.write()` manually (very unlikely) and relies on the return value being `Unit` (even more unlikely)

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
martinbonnin: Your call. The only thing I'm pretty positive about is that it is a `Source` and not a `BufferedSource`

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/ApolloHttpCache.kt
BoD: A lot simpler 😅  That makes sense 👍

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
```diff
@@ -68,18 +71,64 @@ suspend fun writeResponse(sink: BufferedSink, mockResponse: MockResponse, versio
 class MockResponse(
     val statusCode: Int = 200,
     val body: ByteString = ByteString.EMPTY,
-    val chunks: List<MockChunk> = emptyList(),
+    chunks: List<MockChunk> = emptyList(),
     val headers: Map<String, String> = emptyMap(),
     val delayMillis: Long = 0,
+    private val waitForMoreChunks: Boolean = false,
 ) {
+  var chunks: List<MockChunk> = chunks
```
BoD, line 79: In addition to the channel, still keep the List, as it is useful to do asserts in tests

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
martinbonnin: Did we ever find the reason behind this?

BoD: Sorry I thought your previous comment was about the test passing without committing because it wasn't checking the presence of data in the cache - but now I'm wondering if you meant something else 😅

martinbonnin: Let me try again

martinbonnin: Still the case. What I'm doing is I'm commenting out the `closeAndCommitCache()` on line 129. And then run tests:

```
./gradlew -p tests build
```

And everything is still ✅  leaving me to figure out what this is for? Or do I need to run other tests?

BoD: Ah got it (thought you were commenting the inside of `closeAndCommitCache`).

This path can be reached when an incomplete response is fully read: the JSON reader pulls more bytes, but there aren't any, so `originalSource.read` returns -1. In that case this bit of code commits early, but:

1. it's actually not useful, because `close` will still be called after a parse exception occurs (which commits at that point)
2. not 100% sure if we actually want to commit in that case? On the one hand, this would be consistent with the current behavior (currently, the data is cached before it's parsed). On the other hand, if we know for sure that -1 here means an incomplete response, it may be nice to avoid caching that?

WDYT?

martinbonnin: Fell into this rabbit hole and turns out it's not too hard to implement. I gave it a try there: https://github.com/apollographql/apollo-kotlin/pull/4087. This is wildly untested. Feel free to include in this PR or I'll rebase and open a follow up PR later

BoD: Yes this looks good! 🙏  Merging into this one and I'll add 2 tests for it.

BoD: hmm 2 tests failed - I'll have a look before merging :P

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerExtensions.kt
```diff
@@ -47,25 +47,46 @@ fun createMultipartMixedChunkedResponse(
     responseDelayMillis: Long = 0,
     chunksDelayMillis: Long = 0,
     boundary: String = "-",
+    waitForMoreChunks: Boolean = false,
```
martinbonnin, line 50: That's a lot of constructor arguments. I think MockResponse is due for a little refactoring. 

We could make it take a `Source` as body:

```kotlin
class MockResponse(
    val statusCode: Int = 200,
    val body: Source = Buffer(),
    val headers: Map<String, String> = emptyMap(),
)
```

Then `delay` and `chunks` "just" become specific `Source` implementations:

```kotlin
class DelayedSource(
    val delayMillis: Long,
    val wrappedSource:Source
    ): Source{
  private var firstRead = true

  override fun close() {
    wrappedSource.close()
  }

  override fun read(sink: Buffer, byteCount: Long): Long {
    if (firstRead) {
      // TODO: see if we can implement this more efficiently with timeouts
      usleep(delayMillis*1000)
    }
    return wrappedSource.read(sink, byteCount)
  }

  override fun timeout(): Timeout {
    TODO("Not yet implemented")
  }
}

class ChunkedSource(): Source {
  // ...
}
```

apollo-mpp-utils/src/commonMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -1,7 +1,13 @@
+@file:JvmName("-utils")
+
 package com.apollographql.apollo3.mpp
 
+import kotlin.jvm.JvmName
+
 expect fun currentTimeMillis(): Long
+expect fun currentTimeFormatted(): String
```
martinbonnin, line 8:
```suggestion
/**
 * The current time as a human readable String. Used for debugging
 */
@ApolloInternal
expect fun currentTimeFormatted(): String
```

apollo-mpp-utils/src/appleMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -11,10 +13,24 @@ actual fun currentTimeMillis(): Long {
...
+
 actual fun currentThreadId(): String {
   return pthread_self()?.rawValue.toString()
 }
 
+actual fun currentThreadName(): String {
+  return if (NSThread.isMainThread) {
+    "main"
+  } else {
+    currentThreadId()
```
martinbonnin, line 30: Could you use [pthread_get_name_np](https://linux.die.net/man/3/pthread_getname_np) instead? Does iOS put anything relevant there?

BoD: Sadly, this returns an empty string (same for `NSThread.currentThread.name`)!

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerExtensions.kt
martinbonnin: Could we do some kind of non-blocking/polling I/O with okio? have `read()` return zero in the source implementation while waiting and something >0 when data is available again?

BoD: So I think 0 would make callers call again immediately, in a loop, but we could sleep for a few ms to not hog the cpu? Not ideal but maybe enough for tests?

martinbonnin: Right. `okio.Source` is not the good abstraction then. But we could write our own `Source` abstraction with coroutines support to give a chance to JS land to "sleep"/"delay"

BoD: What about `Flow<ByteString>`?

```
class MockResponse(
    val statusCode: Int = 200,
    val body: Flow<ByteString> = emptyFlow(),
    val headers: Map<String, String> = emptyMap(),
)
```

Example usage:

```
fun MockServer.enqueue(string: String, delayMs: Long = 0) {
  enqueue(MockResponse(
      statusCode = statusCode,
      body = flow {
        delay(delayMillis)
        emit(body.encodeUtf8())
      },
      headers = mapOf("Content-Length" to string.length.toString()),
  ))
}
```

```
val chunks = Channel<String>(Channel.UNLIMITED)
delay(200)
chunks.send("chunk 1")
delay(200)
chunks.send("chunk 2")
chunks.close()
val body: Flow<ByteString> = chunks.receiveAsFlow().map { it.encodeUtf8() }
```

BoD: If you're OK with it I can remove the MockServer changes from this PR and open a dedicated one 😅

martinbonnin: Sure thing!

PR #4078, opened 2022-05-04
🐘  publish `apollo-gradle-plugin-external`

martinbonnin:
_Note: We might want to hold off until 3.3 before we merge this._

This creates a new `apollo-gradle-plugin-external` module that contains the code for the Gradle plugin. The tests are still in `apollo-gradle-plugin` because they are relying on the `com.apollographql.apollo3` plugin ID.

This means users that are aware of Gradle classpath issues (or just want to consume the jar) can use the external publication:

```kotlin
plugins {
  id("com.apollographql.apollo3.external")
}
```

or to use the jar:

```kotlin
dependencies {
  implementation("com.apollographql.apollo3:apollo-gradle-plugin-external:$version")
}
```

An alternative would have been to publish different Gradle variants but:

1. It makes it more complicated to consume the plugin compared to just changing the plugin ID
2. It isn't compatible with maven projects or Kotlin scripts

Closes https://github.com/apollographql/apollo-kotlin/issues/4025


apollo-gradle-plugin/src/test/files/gradle/build.gradle.kts.template
```diff
@@ -18,8 +18,8 @@ buildscript {
     maven {
       url = uri("../../../build/localMaven")
     }
-    google()
     mavenCentral()
+    google()
```
martinbonnin, line 22: I moved `google()` down as part of this PR because the Google maven repo was having issues and it shouldn't hurt but this is not related to `apollo-gradle-plugin-external`

BoD (approved):
👍

PR #4083, opened 2022-05-06
Fix test directories not cleared

asimonigh:
Fix TestBuilders classes not cleared after a gql fragment change


martinbonnin (approved):
Thanks !!

BoD (approved):
Good catch 👍

asimonigh:
> Good catch 👍

good catch from @martinbonnin ;)

PR #4086, opened 2022-05-07
Do not use 'header' as a enum value name as it breaks the Kotlin compiler

martinbonnin:
See https://youtrack.jetbrains.com/issue/KT-52315

Thanks @ephemient for catching this!

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
@@ -18,7 +18,7 @@ fun String.escapeKotlinReservedWord() = this
 fun String.escapeKotlinReservedEnumValueNames() : String {
   return when {
     // https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants:~:text=properties%20for%20obtaining%20its%20name%20and%20position
-    this in arrayOf("name", "ordinal") -> "${this}_"
+    "(?:name|ordinal)_*".toRegex().matches(this) -> "${this}_"
     // "header" and "impl" are added to this list because of https://youtrack.jetbrains.com/issue/KT-52315
     this in arrayOf("header", "impl") -> "`${this}`"
     else -> this
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
@@ -13,7 +13,8 @@ private val JAVA_RESERVED_WORDS = arrayOf(
 
 // Reference:
 // https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants:~:text=properties%20for%20obtaining%20its%20name%20and%20position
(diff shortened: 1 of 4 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
```diff
@@ -13,7 +13,8 @@ private val JAVA_RESERVED_WORDS = arrayOf(
 
 // Reference:
 // https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants:~:text=properties%20for%20obtaining%20its%20name%20and%20position
-private val KOTLIN_RESERVED_ENUM_VALUE_NAMES = arrayOf("name", "ordinal")
+// "header" and "impl" are added to this list because of https://youtrack.jetbrains.com/issue/KT-52315
```
BoD, line 16: 😱

BoD (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
martinbonnin: PR to handle this at the KotlinPoet level: https://github.com/square/kotlinpoet/pull/1248

martinbonnin:
Updated this PR with a better fix that escapes with backticks. If this gets fixed in the Kotlin compiler, no change will be needed on the caller side

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ReservedKeywords.kt
```diff
@@ -10,14 +10,17 @@ private val JAVA_RESERVED_WORDS = arrayOf(
...
 fun String.escapeJavaReservedWord() = if (this in JAVA_RESERVED_WORDS) "${this}_" else this
 
 // Does nothing. KotlinPoet will add the backticks
 fun String.escapeKotlinReservedWord() = this
 
-fun String.escapeKotlinReservedEnumValueNames() = if (this in KOTLIN_RESERVED_ENUM_VALUE_NAMES) "${this}_" else this
+fun String.escapeKotlinReservedEnumValueNames() : String {
+  return when {
+    // https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants:~:text=properties%20for%20obtaining%20its%20name%20and%20position
+    this in arrayOf("name", "ordinal") -> "${this}_"
```
BoD, line 21: Do you remember why we don't use backticks for these two as well? (we wouldn't want to change this now as it would be breaking, but I'm curious if we simply didn't think of it or a good reason).

ephemient: I wonder if you want to
```kotlin
"(?:name|ordinal)_*".toRegex().matches(this)
```
to prevent collisions if a schema defines both `name` and `name_`.

martinbonnin: Excellent point, much better 👍

PR #4087, opened 2022-05-09
Do not cache HTTP responses with GraphQL errors

martinbonnin:


BoD (approved):

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
```diff
@@ -144,20 +143,16 @@ class DiskLruHttpCache(private val fileSystem: FileSystem, private val directory
...
         try {
           sink.close()
-          cacheEditor.commit()
+          cacheEditor.commitUnlessAborted()
         } catch (e: Exception) {
           // Silently ignore cache write errors
         } finally {
           hasClosedAndCommitted = true
         }
+        originalSource.close()
```
martinbonnin, line 155: Should that be in a try/catch as well? I think it can throw

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/internal/DiskLruCache.kt
```diff
@@ -863,6 +863,16 @@ internal class DiskLruCache(
       }
     }
 
+    fun commitUnlessAborted() {
```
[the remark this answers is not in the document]
BoD: Ahah yeah it felt a bit weird, but there was already a `abortUnlessCommitted` (unused) 😅. Agreed, I'll put it in `ProxySource`.

apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/DiskLruHttpCache.kt
BoD: If we successfully read the source, but then when trying to close it it threw, I think we should still commit? If yes, then no need to catch it, it will throw after the commit (or abort if there was an error).

martinbonnin: Right thanks for the deep dive 👍 . Sometimes I miss checked exceptions...

martinbonnin (commented):
Can't approve my own PR 😄 . But looks good to me!

PR #4089, opened 2022-05-10
add DslMarker for test builders

hrach:
Introduce a custom DslMarker for MapBuilder. This way nesting builders
won't see their parent implicit receiver and users cannot define
parent receiver's properties without an explicit receiver reference.

### DIFF
--- a/apollo-api/api/apollo-api.api
+++ b/apollo-api/api/apollo-api.api
@@ -1051,6 +1051,9 @@ public final class com/apollographql/apollo3/api/json/MapJsonWriter : com/apollo
 	public fun value (Z)Lcom/apollographql/apollo3/api/json/MapJsonWriter;
 }
 
+public abstract interface annotation class com/apollographql/apollo3/api/test/ApolloTestBuilderMarker : java/lang/annotation/Annotation {
+}
+
 public final class com/apollographql/apollo3/api/test/TestResolverKt {
 }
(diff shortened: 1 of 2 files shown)


martinbonnin:
I love this! Looks really cool 👍! 

To be on the exhaustive side might want to add a test but TBH this is going to be hard to test because we need to test that something does not compile. It's doable but I don't think it's worth the trouble. @BoD any thoughts.

BoD:
Agreed 😊 I was not familiar with the annotation so an example of what was possible before / not possible after would be cool but we can just try on the branch, I agree a test is not necessary.

martinbonnin:
@hrach can you run `./gradlew apiDump` and commit the changed API files? This is needed so that we can track the public API

BoD:
### Before
Screen Shot 2022-05-10 at 10 34 47

### After
Screen Shot 2022-05-10 at 10 35 00

👍

hrach:
Hi guys, thanks for guidance for the API dump. I'll do this.

Testing: I was thinking about doing something, but "not compiling" requirement seemed rather very difficult to do for some trivial code. Thank you for trying out. :)

hrach:
rebased

martinbonnin (approved):
Thanks!

BoD (approved):
👍

PR #4090, opened 2022-05-10
Sse sus/side channel payload

fmatosqg:
- Moves payload data classes to final module/package in apollo-runtime
- Creates hello world test in test module `sse` that starts server and checks it's running


fmatosqg:
Created on the wrong repo, should be on the fork https://github.com/fmatosqg/apollo-kotlin/pull/3

PR #4091, opened 2022-05-10
🧪 @experimental support

martinbonnin:
Add support for the [`@experimental` directive](https://github.com/graphql/graphql-spec/issues/943). By default, it annotates experimental fields, input fields and enum values with  `com.apollographql.apollo3.annotations.@ApolloExperimental`. This can be changed with a Gradle setting:

```kotlin
apollo {
  experimentalAnnotation.set("com.example.MyOptInAnnotation")
  // "none" is a special value that will not emit any annotation
  experimentalAnnotation.set("none")
}
```


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
```diff
@@ -183,6 +184,7 @@ internal class IrBuilder(
         implements = implementsInterfaces,
         keyFields = schema.keyFields(name),
         description = description,
+        // XXX: this is not spec-compliant. Directive cannot be on object definitions
```
martinbonnin, line 187: I kept those because it shouldn't harm too much but we should not hesitate to remove them if anything comes up.

kotlin-js-store/yarn.lock
```diff
@@ -19,6 +19,14 @@ abort-controller@3.0.0:
...
     event-target-shim "^5.0.0"
 
+accepts@~1.3.8:
+  version "1.3.8"
+  resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.3.8.tgz#0bf0be125b67014adcb0b0921e62db7bffe16b2e"
+  integrity sha512-PYAthTa2m2VKxuvSD3DPC/Gy+U+sOA1LAuT8mkmRuvw+NACSaeXEQ+NHcVF7rONl6qcaxV3Uuemwawk+7+SJLw==
+  dependencies:
+    mime-types "~2.1.34"
+    negotiator "0.6.3"
+
```
martinbonnin, line 29: No idea why this was added.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/NamedType.kt
```diff
@@ -26,7 +27,9 @@ internal fun NamedType.toParameterSpec(context: KotlinContext): ParameterSpec {
           name = context.layout.propertyName(graphQlName),
           type = context.resolver.resolveIrType(type)
       )
-      .applyIf(description?.isNotBlank() == true) { addKdoc("%L\n", description!!) }
+      .maybeAddDescription(description)
+      .maybeAddDeprecation(deprecationReason)
```
BoD, line 31: Deprecation was missing?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrBuilder.kt
BoD: Oh I think that answers my previous comment 😅

apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -511,6 +511,18 @@ interface Service {
...
 
+  /**
+   * The annotation to use for experimental fields/inputFields/enumValues
+   *
+   * You can pass the special value "none" to disable adding an annotation.
+   * If you're using a custom annotation, it must be able to target:
+   * - AnnotationTarget.PROPERTY
+   * - AnnotationTarget.CLASS
+   *
+   * Default: "com.apollographql.apollo3.annotations.ApolloExperimental"
```
BoD, line 522: I'm not 100% convinced we should re-use `ApolloExperimental` for the default as it is currently kind of specific to the project's codebase, so maybe it would be worth it to introduce a specific `Experimental` annotation.
Not a big deal for the time being!

BoD (approved):
Very cool!

apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -18,3 +18,9 @@ directive @typePolicy(keyFields: String!) on OBJECT
 # Indicates how to compute a key from a field arguments.
 # `keyArgs` should contain a selection set. Composite args are not supported yet.
 directive @fieldPolicy(forField: String!, keyArgs: String!) repeatable on OBJECT
+
+# Indicates that the given field or enum value is still experimental and might be changed
+# in a backward incompatible manner
+directive @experimental(
+  reason: String! = "Experimental"
+) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION | ENUM_VALUE
```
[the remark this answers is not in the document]
martinbonnin: There's a [graphql-spec work item for that](https://github.com/graphql/graphql-spec/pull/805/) 😅 

I can do this as part of a separate PR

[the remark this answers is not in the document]
BoD: Oh! 😅. Well I guess it's not urgent until it's accepted but also wouldn't hurt to be ready in advance :)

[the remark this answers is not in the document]
martinbonnin: Just added `@deprecated` on input fields and field arguments. Arguments are an interesting case because I haven't found a way to "touch" them from Kotlin code. They are either GraphQL variables or values. I added validation rules so using them will emit a warning during codegen. But no annotation is emitted there.

apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
martinbonnin: Agreed the semantics are different but didn't find a good alternative name. `GraphQLExperimental` maybe ?

BoD: I was thinking just `Experimental` but I like `GraphQLExperimental` too. Another idea: `SchemaExperimental`.

apollo-ast/src/main/resources/apollo.graphqls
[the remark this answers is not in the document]
BoD: Very cool! 👍

apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
martinbonnin: I was a bit concerned `Experimental` would clash with other annotation but maybe it's not the case. Maybe we should make `@Experimental` `@ApolloExperimental` 🙃 . 

Just kidding. I'll go with `@Experimental` and we'll see how that unfolds.

kotlin-js-store/yarn.lock
martinbonnin: > My current guess is that it depends if it's the root project or the tests project that is run last

confirmed. I'll make a follow up PR

martinbonnin: https://github.com/apollographql/apollo-kotlin/pull/4094

BoD (approved):

PR #4094, opened 2022-05-10
🌍 use separate kotlin-js-store

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/pull/4091#discussion_r869231816


BoD (approved):
👍

PR #4096, opened 2022-05-11
MockServer: make MockResponse.body a Flow<ByteString>

BoD:
In this PR, `MockResponse.body` becomes a `Flow<ByteString>` (previously was `ByteString`). This allows to not treat chunks as a special case.

The JS MockServer was also changed to use a `net.Server` (instead of `http.Server`), which allows sharing more common code (previously, `MockServerCommon.writeResponse` was used on JVM and Apple, but not JS).


apollo-mockserver/api/apollo-mockserver.api
```diff
@@ -27,16 +17,16 @@ public final class com/apollographql/apollo3/mockserver/MockRequest {
 
 public final class com/apollographql/apollo3/mockserver/MockResponse {
 	public fun <init> ()V
-	public fun <init> (ILokio/ByteString;Ljava/util/List;Ljava/util/Map;J)V
-	public synthetic fun <init> (ILokio/ByteString;Ljava/util/List;Ljava/util/Map;JILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (ILkotlinx/coroutines/flow/Flow;Ljava/util/Map;J)V
```
BoD, line 20: I think considering people were most probably not using chunks, this is probably ok. The main ways to construct a MockResponse with a `String` or a `ByteString` are still there.

martinbonnin: 👍 Speaking of which.. Should we make this use the `Builder` pattern, like other parts of the API?

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
```diff
@@ -27,70 +29,50 @@ class MockRequest(
...
 }
 
 class MockResponse(
     val statusCode: Int = 200,
-    val body: ByteString = ByteString.EMPTY,
-    val chunks: List<MockChunk> = emptyList(),
-    val headers: Map<String, String> = emptyMap(),
+    val body: Flow<ByteString> = emptyFlow(),
+    val headers: Map<String, String> = mapOf("Content-Length" to "0"),
     val delayMillis: Long = 0,
```
martinbonnin, line 50: Can we remove this as a property and "just" delay in the `Flow`?

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerExtensions.kt
```diff
@@ -48,24 +54,97 @@ fun createMultipartMixedChunkedResponse(
...
+  val contentLengthHeader = "Content-Length: ${content.length}"
+  val contentTypeHeader = "Content-Type: $contentType"
+  val endBoundary = if (isLast) "--$boundary--" else "--$boundary"
+  return startBoundary +
+      "$contentLengthHeader\r\n" +
+      "$contentTypeHeader\r\n" +
+      "\r\n" +
+      "$content\r\n" +
+      "$endBoundary\r\n"
+}
```
martinbonnin, line 110: This is really cool 👍

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerExtensions.kt
```diff
@@ -48,24 +54,97 @@ fun createMultipartMixedChunkedResponse(
...
-                "\r\n" +
-                "$part\r\n" +
-                "$endBoundary\r\n",
-            delayMillis = chunksDelayMillis
-        )
-      }
+      body = chunkChannel.receiveAsFlow().map {
+        delay(chunksDelayMillis)
+        it
+      }.asChunked()
```
martinbonnin, line 133: 👌

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
BoD: We could but the semantics of this delay are actually a bit different, as it applies before *anything* is sent. Whereas if we did this, the server would send the status code and headers immediately, then delay before sending the body. Maybe that's ok though?

apollo-mockserver/api/apollo-mockserver.api
BoD: I think it would make sense yeah 😁. I'll take a look.

apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/assertMockResponse.kt
```diff
@@ -2,23 +2,21 @@ package com.apollographql.apollo3.mockserver.test
 
 import com.apollographql.apollo3.api.http.HttpResponse
 import com.apollographql.apollo3.mockserver.MockResponse
+import okio.ByteString
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 
 fun assertMockResponse(
     mockResponse: MockResponse,
+    body: ByteString,
```
martinbonnin, line 11: Instead of adding a new parameter, could you do this instead?

```kotlin
  assertEquals(
      mockResponse.body.toList().fold(Buffer()) { buffer, byteString ->
        buffer.write(byteString)
      }.readByteString(),
      httpResponse.body!!.readByteString()
  )
```

This can suspend so you don't want to use it when delays are involved but I think so far it's only used in non-suspending code.

apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerCommon.kt
martinbonnin: Aaahh yes, good catch 👍

apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/assertMockResponse.kt
BoD: Arh it seemed like a good idea, but it's a bit tricky. When using chunked you pass something like `c\r\nFirst chunk\n\r\nc\r\nSecond chunk\r\n0\r\n\r\n` to the `MockResponse` but what you get in the `HttpResponse` is `First chunk\nSecond chunk`. So you can't really compare them as-is.

martinbonnin: Looks like we need a chunk decoder then ? Should be relatively easy to implement with all the okio/kotlin goodness?

martinbonnin (approved):
👍

PR #4104, opened 2022-05-13
💧 first drop for a SQLite backend that stores when each field was last updated

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/3566

First drop for a SQLite backend that stores when each field was last updated. It introduces a new SQLite table that stores Records in a binary format that also holds the last updated timestamp.

This allows to throw a cache miss when a given field is older that a configured max-age

```kotlin
    val client = ApolloClient.Builder()
        .normalizedCache(
            normalizedCacheFactory(),
            TypePolicyCacheKeyGenerator,
            MaxAgeCacheResolver(10) // Allow 10s for the max Age
        )
        .serverUrl("https://...")
        .build()

    val records = query.normalize(data, CustomScalarAdapters.Empty, TypePolicyCacheKeyGenerator).values

    client.apolloStore.accessCache {
      // store records 11s in the past
      it.merge(records, cacheHeaders(currentTimeMillis()/1000 - 11))
    }

    try {
      client.query(GetUserQuery()).fetchPolicy(FetchPolicy.CacheOnly).execute()
      fail("An exception was expected")
    } catch (e: CacheMissException) {
      assertTrue(e.age != null) // Age will be 11s
    }
```

Next steps: 

* add more tests
* make the maxAge configurable per-type/per-field programmatically
* make the maxAge configurable per-type/per-field declaratively
* add benchmarks


apollo-api/api/apollo-api.api
```diff
@@ -1120,8 +1120,8 @@ public final class com/apollographql/apollo3/exception/AutoPersistedQueriesNotSu
 
 public final class com/apollographql/apollo3/exception/CacheMissException : com/apollographql/apollo3/exception/ApolloException {
 	public static final field Companion Lcom/apollographql/apollo3/exception/CacheMissException$Companion;
-	public fun <init> (Ljava/lang/String;Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/Long;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun getAge ()Ljava/lang/Long;
```
martinbonnin, line 1124: This is a false positive of https://github.com/Kotlin/binary-compatibility-validator/issues/36. This API shouldn't be public just yet

apollo-normalized-cache-api/api/apollo-normalized-cache-api.api
```diff
@@ -164,6 +166,7 @@ public final class com/apollographql/apollo3/cache/normalized/api/Record : java/
 	public final fun getFields ()Ljava/util/Map;
 	public final fun getKey ()Ljava/lang/String;
 	public fun getKeys ()Ljava/util/Set;
+	public final fun getLastUpdated ()Ljava/util/Map;
```
martinbonnin, line 169: Same here. This is marked as ApolloExperiemental unfortunately, that's not taken into account here

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/CacheHeaders.kt
```diff
@@ -28,7 +28,10 @@ class CacheHeaders internal constructor(private val headerMap: Map<String, Strin
   /**
    * @return A [CacheHeaders.Builder] with a copy of this [CacheHeaders] values.
    */
-  fun toBuilder(): Builder = builder().addHeaders(headerMap)
+  @Deprecated("Use newBuilder() instead", ReplaceWith("newBuilder()"))
+  fun toBuilder(): Builder = newBuilder()
+
+  fun newBuilder(): Builder = builder().addHeaders(headerMap)
```
martinbonnin, line 34: For consistency with other Builders

apollo-normalized-cache-sqlite/api/android/apollo-normalized-cache-sqlite.api
```diff
@@ -1,77 +1,10 @@
-public abstract interface class com/apollographql/apollo3/cache/normalized/sql/ApolloDatabase : com/squareup/sqldelight/Transacter {
-	public static final field Companion Lcom/apollographql/apollo3/cache/normalized/sql/ApolloDatabase$Companion;
-	public abstract fun getCacheQueries ()Lcom/apollographql/apollo3/cache/normalized/sql/CacheQueries;
-}
-
-public final class com/apollographql/apollo3/cache/normalized/sql/ApolloDatabase$Companion {
-	public final fun getSchema ()Lcom/squareup/sqldelight/db/SqlDriver$Schema;
-	public final fun invoke (Lcom/squareup/sqldelight/db/SqlDriver;)Lcom/apollographql/apollo3/cache/normalized/sql/ApolloDatabase;
-}
-
```
martinbonnin, line 10: I moved all the SQLDelight generated classes into an `internal` package. Ideally we would generate them internal but it's not possible just yet, see https://github.com/cashapp/sqldelight/issues/1333

This is a breaking change for any code calling this from outside the module but as this is an implementation detail, I think we can afford this.

apollo-normalized-cache-sqlite/build.gradle.kts
```diff
@@ -26,6 +32,7 @@ configure<org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension> {
       dependencies {
         api(projects.apolloApi)
         api(projects.apolloNormalizedCacheApi)
+        api(projects.apolloNormalizedCache)
```
martinbonnin, line 35: Our setup is a bit weird and allows using `apolloNormalizedCacheSqlite` without `apolloNormalizedCache` which is a bit confusing.

`apolloNormalizedCache` is the one with the `apollo-runtime` dependency so it pulls some code but from the consumer side of things I don't really see a reason to use `apolloNormalizedCacheSqlite` without `apolloNormalizedCache`. I'm happy to revert and keep the current behaviour though

apollo-normalized-cache-sqlite/src/commonMain/sqldelight/blob/com/apollographql/apollo3/cache/internal/blob/blob.sq
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE blobs (
+  key TEXT NOT NULL PRIMARY KEY,
+  blob BLOB NOT NULL
+) WITHOUT ROWID;
```
martinbonnin, line 4: The [WITHOUT ROWID](https://www.sqlite.org/withoutrowid.html) looks very promising. I'm hoping it can get us smaller/faster queries.

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/sqlNormalizedCacheFactory.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.apollographql.apollo3.cache.internal.json.JsonDatabase
+import com.apollographql.apollo3.cache.normalized.sql.internal.BlobRecordDatabase
+import com.apollographql.apollo3.cache.normalized.sql.internal.JsonRecordDatabase
+import com.apollographql.apollo3.cache.normalized.sql.internal.RecordDatabase
+import com.squareup.sqldelight.db.SqlDriver
+
+expect class SqlNormalizedCacheFactory internal constructor(
+    driver: SqlDriver,
+    withAge: Boolean = false,
+) : NormalizedCacheFactory
```
martinbonnin, line 14: I'm pretty unhappy about the current state of things of how to build a `SqlNormalizedCacheFactory`. The more I think about it, the more I think the expect/actual here doesn't make much sense. Initialization has to be platform specific so the expect constructor has little use in practice and puts unecessary constraints on the different platforms. Thoughts about deprecating it?

BoD (commented):
On thing that I'm not entirely sure I understand: the date stored along with fields - with the `MaxAgeCacheResolver` they're meant to represent the time at which they were received, right?  But with a different Resolver, they could also represent an expiration date in the future (whose value could come from the server), right? So it is a generic date - its usage can adapt to needs?

apollo-mpp-utils/src/appleMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -3,14 +3,15 @@ package com.apollographql.apollo3.mpp
...
+import platform.Foundation.timeIntervalSince1970
 import platform.posix.pthread_self
 import kotlin.native.concurrent.ensureNeverFrozen
 import kotlin.native.concurrent.freeze
 import kotlin.native.concurrent.isFrozen
 import kotlin.system.getTimeMillis
 
 actual fun currentTimeMillis(): Long {
-  return getTimeMillis()
+  return (NSDate().timeIntervalSince1970 * 1000).toLong()
```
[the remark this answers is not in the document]
martinbonnin: [getTimeMillis](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/get-time-millis.html) gets current system time in milliseconds since certain moment in the past. I want the EPOCH here. I'll update the KDoc too while I'm at it.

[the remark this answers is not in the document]
BoD: woops my bad!

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/CacheResolver.kt
```diff
@@ -92,17 +95,53 @@ object DefaultCacheResolver: CacheResolver {
...
     return parent[name]
   }
 }
 
+
+/**
+ * A cache resolver that uses the parent to resolve fields and use a constant max A
+ */
+@ApolloExperimental
+class MaxAgeCacheResolver(private val maxAge: Int) : CacheResolver {
```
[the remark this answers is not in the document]
martinbonnin: I'm usually all for that but in that case I was thinking the similarity with the HTTP headers and existing [`@cacheControl` spec](https://www.apollographql.com/docs/apollo-server/performance/caching/) could maybe save the verbosity and make things symmetrical. Not sure...

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
```diff
@@ -25,32 +30,63 @@ class Record (
     val mutationId: Uuid? = null,
 ) : Map<String, Any?> by fields {
 
-  val sizeInBytes = calculateBytes(this)
+  @ApolloExperimental
+  var lastUpdated: Map<String, Long?>? = null
```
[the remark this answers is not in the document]
martinbonnin: `private set` it is 👍

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
```diff
@@ -25,32 +30,63 @@ class Record (
...
+      return calculateBytes(this) + datesSize
+    }
 
   /**
    * Returns a merge result record and a set of field keys which have changed, or were added.
    * A field key incorporates any GraphQL arguments in addition to the field name.
    */
-  fun mergeWith(otherRecord: Record): Pair<Record, Set<String>> {
+  @ApolloExperimental
+  fun mergeWith(newRecord: Record, date: Long?): Pair<Record, Set<String>> {
```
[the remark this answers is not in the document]
martinbonnin: I hear you... But also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control doesn't add the unit...

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/BlobRecordSerializer.kt
```diff
@@ -0,0 +1,165 @@
...
+      else -> error("Trying to write unsupported Record value: $value")
+    }
+  }
+  private fun Buffer.readAny(): Any? {
+    return when (val what = readByte().toInt()) {
+      STRING -> readString()
+      INT -> readInt()
+      LONG ->readLong()
+      DOUBLE -> readString().toDouble()
+      BOOLEAN -> readInt() > 0
```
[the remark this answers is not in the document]
martinbonnin: Good catch 👍

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchOnMain.kt
```diff
@@ -25,6 +26,7 @@ suspend fun <R> suspendAndResumeOnMain(block: (MainContinuation<R>, InvokeOnCanc
   assertMainThreadOnNative()
 
   return suspendCancellableCoroutine { continuation ->
+    continuation.ensureNeverFrozen()
```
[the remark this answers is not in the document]
martinbonnin: I removed it in a subsequent commit. It indeed breaks coroutines-mt.

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/sqlNormalizedCacheFactory.kt
martinbonnin: I was thinking 

* AndroidSqliteNormalizedCacheFactory
* NativeSqliteNormalizedCacheFactory
* JdbcSqliteNormalizedCacheFactory

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchOnMain.kt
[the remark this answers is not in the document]
BoD: Yeah sorry I noticed later :)

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/sqlNormalizedCacheFactory.kt
BoD: Oh yeah that works!

martinbonnin:
> One thing that I'm not entirely sure I understand: the date stored along with fields - with the MaxAgeCacheResolver they're meant to represent the time at which they were received, right? But with a different Resolver, they could also represent an expiration date in the future (whose value could come from the server), right? So it is a generic date - its usage can adapt to needs?

Excellent question. At the moment it's mostly the receive time. We could store the expiration time too I guess. It wouldn't be too hard I think. Something I want to improve is to make sure a given database is always open with the same assumption (blob vs json, date vs expire). I'll suggest something

martinbonnin:
Also closes https://github.com/apollographql/apollo-kotlin/issues/3758

apollo-normalized-cache-sqlite/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/sqlNormalizedCacheFactory.kt
martinbonnin: Nevermind, with app startup, we can expose a common constructor:

```
SqliteNormalizedCacheFactory(name: String?, withDates: Boolean)
```

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/CacheResolver.kt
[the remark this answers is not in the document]
martinbonnin: I went with `Long` for dates and `Int` for ages. Because ages are typically much smaller than dates and this is a hint that huge values aren't supported there.

apollo-normalized-cache-sqlite/src/main/AndroidManifest.xml
```diff
@@ -1 +1,14 @@
-<manifest package="com.apollographql.apollo3.cache.normalized.sql" />
+<manifest package="com.apollographql.apollo3.cache.normalized.sql" xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools">
+  <application>
+    <provider
+        android:name="androidx.startup.InitializationProvider"
+        android:authorities="${applicationId}.androidx-startup"
+        android:exported="false"
+        tools:node="merge">
+      <!-- This entry makes ExampleLoggerInitializer discoverable. -->
```
BoD, line 9:
```suggestion
      <!-- This entry makes ApolloInitializer discoverable. -->
```

build.gradle.kts
```diff
@@ -194,12 +194,13 @@ configure<kotlinx.validation.ApiValidationExtension> {
   ignoredPackages.addAll(
       listOf(
           /**
-           * We rely on annotations or "internal" visibility to hide the non-public APIs in general
+           * In general, we rely on annotations or "internal" visibility to hide the non-public APIs. Bur there are a few exceptions
```
BoD, line 197:
```suggestion
           * In general, we rely on annotations or "internal" visibility to hide the non-public APIs. But there are a few exceptions:
```

BoD (approved):

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/CacheResolver.kt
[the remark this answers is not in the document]
BoD: Got it 👍

apollo-normalized-cache-sqlite/build.gradle.kts
```diff
@@ -78,6 +88,31 @@ if (System.getProperty("idea.sync.active") == null) {
...
+     *
+     * In general, there is so little Android code here, it's not really worth running lint
+     */
+    enabled = false
+  }
+
+  tasks.configureEach {
+    if (name.endsWith("UnitTest")) {
+      /**
+       * Because there is no App Startup in Android unit tests, the Android tests
```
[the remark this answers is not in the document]
martinbonnin: Ooops, thanks for catching it!

PR #4115, opened 2022-05-17
update POM_URL and other maven information

yshrsmz:
I was looking at search.maven.org and found out that POM_URL and others are pointing to older repository url.


yshrsmz:
I sined it

BoD (approved):
Good catch! Thank you 🙏

martinbonnin (approved):

PR #4116, opened 2022-05-17
Update list of custom scalar adapters in the migration doc.

StylianosGakis:
Taken from apollo-kotlin/apollo-adapters/README.md which I presume is more up to date. I believe this is a simple case of that doc being a tiny bit out of sync right?


BoD (approved):
Good catch! Thank you 🙏

martinbonnin (approved):

PR #4117, opened 2022-05-17
Gradle tweaks + fix README about minimum supported Kotlin version

BoD:
Based on the `kotlin-1-7` branch.

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
@@ -2,6 +2,7 @@ package com.apollographql.apollo3.compiler
 
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_0_0
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_3_1
 import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.ast.Schema
 import com.apollographql.apollo3.ast.toSchema
@@ -30,7 +31,11 @@ const val ADD_TYPENAME_ALWAYS = "always"
 enum class TargetLanguage {
   // The order is important. See [isTargetLanguageVersionAtLeast]
   JAVA,
+
+  @Deprecated("Use KOTLIN_1_5", replaceWith = ReplaceWith("KOTLIN_1_5"))
+  @ApolloDeprecatedSince(v3_3_1)
   KOTLIN_1_4,
+
   KOTLIN_1_5,
 }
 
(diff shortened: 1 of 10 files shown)


apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/KotlinPluginVersionTests.kt
```diff
@@ -29,12 +29,26 @@ class KotlinPluginVersionTests {
...
-  fun kotlin14Succeeds() {
+  fun kotlin14Fails() {
     TestUtils.withTestProject("kotlin-plugin-version") { dir ->
       dir.setKotlinPluginVersion("1.4.32")
-      val result = TestUtils.executeTask("help", dir)
+      try {
+        TestUtils.executeTask("build", dir)
+        fail("An exception was expected")
+      } catch (e: UnexpectedBuildFailure) {
+        Truth.assertThat(e.message).contains("Incompatible classes were found in dependencies. Remove them from the classpath or use '-Xskip-metadata-version-check' to suppress errors")
```
martinbonnin, line 39: Do we really need this test? It's interesting to know but it downloads yet another version of Kotlin, spawns yet another GradleRunner instance, etc... So I'd be happy with a comment instead :)

apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/KotlinPluginVersionTests.kt
```diff
@@ -29,12 +29,26 @@ class KotlinPluginVersionTests {
...
+  }
 
-      Truth.assertThat(result.task(":help")!!.outcome).isEqualTo(TaskOutcome.SUCCESS)
+  @Test
+  fun kotlin15Succeeds() {
+    TestUtils.withTestProject("kotlin-plugin-version") { dir ->
+      dir.setKotlinPluginVersion("1.5.31")
+      val result = TestUtils.executeTask("build", dir)
+
+      Truth.assertThat(result.task(":build")!!.outcome).isEqualTo(TaskOutcome.SUCCESS)
```
martinbonnin, line 50: 👍

martinbonnin (approved):

apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/KotlinPluginVersionTests.kt
martinbonnin: Oh yea, definitely 👍

martinbonnin (changes requested):
Can you change the check in the Gradle plugin? See `checkKotlinPluginVersion`. It also means we can compile the Apollo Gradle Plugin with Kotlin Gradle Plugin 1.5 which I think is an OK change to do?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -30,7 +31,11 @@ const val ADD_TYPENAME_ALWAYS = "always"
 enum class TargetLanguage {
   // The order is important. See [isTargetLanguageVersionAtLeast]
   JAVA,
+
+  @Deprecated("Use KOTLIN_1_5", replaceWith = ReplaceWith("KOTLIN_1_5"))
```
martinbonnin, line 35:
```suggestion
  @Deprecated("Compilation with 1.4 is not supported anymore", replaceWith = ReplaceWith("KOTLIN_1_5"))
```

martinbonnin: Mmm now I see your comment below. Gotcha. Let's keep it like this 👍

martinbonnin (approved):

Issue #4093, opened 2022-05-10
Fail on unused return values in test builders 


martinbonnin:
When using the test builders, it is tempting to write GraphQL-like syntax:

```kotlin
val data = UserDataQuery.Data {
  userViewer {
    userId = "userId"
  }
}
```

This has no effect because the result of `userViewer` is never used. Investigate if we could detect that at runtime at fail early as a hint towards the working code:

```kotlin
val data = UserDataQuery.Data {
  viewer = userViewer {
    userId = "userId"
  }
}
```

(or use a Kotlin `@CheckResult` annotation if one becomes available at some point, see https://youtrack.jetbrains.com/issue/KT-12719/Add-CheckResult-MustUseReturnValue-to-warn-about-unused-return-v)

BoD:
Change available in v3.3.1.

PR #4122, opened 2022-05-18
Check for unassigned fields in Test Builders

BoD:
See also #4093

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/test/MapBuilder.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/test/MapBuilder.kt
@@ -18,12 +18,16 @@ abstract class MapBuilder {
    * with fields
    */
   protected val __map = mutableMapOf<String, Any?>()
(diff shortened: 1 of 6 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/test/MapBuilder.kt
```diff
@@ -63,7 +67,7 @@ class MandatoryTypenameProperty {
       property: KProperty<*>,
   ): String {
     check(typename != null) {
-      "__typename is not known at compile-time for this type. Please specify it explicitely"
+      "__typename is not known at compile-time for this type. Please specify it explicitly"
```
BoD, line 70: Unrelated but noticed this typo 😅  Was also elsewhere.

martinbonnin: Yep, that's me! Thanks!

martinbonnin (approved):

PR #4124, opened 2022-05-19
Add Operation.Data.toJsonString() convenience function for the jvm

StylianosGakis:
Allows a more convenient way to simply get the Json representation from `Operation.Data` without having to manually deal with Buffers etc in client code which the current `toJson` function requires.

This could be imitated on the clients by instead calling the existing `toJson` function like this:
```kotlin
fun Operation.Data.toJsonString(
    scalarTypeAdapters: CustomScalarAdapters = CustomScalarAdapters.Empty,
    indent: String? = null,
): String {
    val buffer = Buffer()
    val jsonWriter = BufferedSinkJsonWriter(buffer, indent)
    toJson(jsonWriter, scalarTypeAdapters)
    return buffer.readUtf8()
}
```
Which should be functionally equivalent to what the introduced function does (if we inlined `buildJsonString` which `toJsonString` uses internally)

If you believe that the API doesn't need this addition that would be totally understandable, and you can feel free to simply close this PR 🙌


martinbonnin (approved):
Thanks!

BoD (approved):
👍

PR #4125, opened 2022-05-19
Uploads should be read only once even when logging

BoD:
In this PR, 2 commits:
1. In LoggingInterceptor, fully read the request's body and redispatch downstream + add an option to not log the request's body
2. add an `isOneShot` field to `HttpBody`, and use it with with the OkHttp engine. This makes using OkHttp's log interceptor not break Uploads

Pretty sure 2. is a breaking change, but I still wanted to see where it takes us. (I can think of a hack to still pass the option to OkHttp without adding the field to the interface)

See also #4054

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
@@ -22,13 +22,6 @@ interface HttpBody {
    * This can be called several times
    */
   fun writeTo(bufferedSink: BufferedSink)
-
-  /**
-   * Returns true if this body expects at most one call to writeTo and can be transmitted at most once.
-   * This would be the case when reading the source of data consumes it, for instance when uploading files.
-   */
-  val isOneShot: Boolean
-    get() = false
 }
 
 /**
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
@@ -22,6 +22,13 @@ interface HttpBody {
    * This can be called several times
    */
   fun writeTo(bufferedSink: BufferedSink)
+
+  /**
+   * Returns true if this body expects at most one call to writeTo and can be transmitted at most once.
+   * This would be the case when reading the source of data consumes it, for instance when uploading files.
+   */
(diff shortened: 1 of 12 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
```diff
@@ -22,6 +22,13 @@ interface HttpBody {
    * This can be called several times
    */
   fun writeTo(bufferedSink: BufferedSink)
+
+  /**
+   * Returns true if this body expects at most one call to writeTo and can be transmitted at most once.
+   * This would be the case when reading the source of data consumes it, for instance when uploading files.
+   */
+  val isOneShot: Boolean
+    get() = false
```
martinbonnin, line 31: I don't like this too much. That one more API that's useful to a small portion of the usecases but accounts for 25% of the interface. It also introduce default methods, which is something I'm never sure how backward compatible it is.

Instead, I'd much rather fail early in `DefaultUpload` if the body has been transmitted already.

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/LoggingInterceptor.kt
```diff
@@ -1,10 +1,24 @@
...
+import kotlin.jvm.JvmOverloads
+
+/**
+ * An interceptor that logs requests and responses.
+ *
+ * @param logRequestBody whether to log the request body. Caution: when uploading files, setting this to `true` will cause the files to
+ * be fully loaded into memory, which may cause OutOfMemoryErrors.
+ */
+class LoggingInterceptor(
+    private val logRequestBody: Boolean,
```
martinbonnin, line 16: Make this a `level`, like in OkHttp?

apollo-runtime/src/jvmMain/kotlin/com/apollographql/apollo3/network/http/OkHttpEngine.kt
```diff
@@ -51,6 +52,10 @@ actual class DefaultHttpEngine constructor(
 
               override fun contentLength() = body.contentLength
 
+              // This prevents OkHttp from reading the body several times (e.g. when using its logging interceptor)
+              // which could consume files when using Uploads
+              override fun isOneShot() = body is UploadsHttpBody
```
BoD, line 57: A bit hacky...

PR #4127, opened 2022-05-20
Add non-breaking spaces after 'return'

BoD:
With fields with long names, the generated code could look like this:

```kotlin
 return
      FieldWithAVeryLongName().apply(block).build()
}
```

instead of:

```kotlin
 return FieldWithAVeryLongName().apply(block).build()
}
```

which made the compiler choke.

Also added the character in all `return` places just in case! 

Thank you @StylianosGakis for reporting this!


martinbonnin (approved):

StylianosGakis:
Amazing 🥳🥳🥳

PR #4128, opened 2022-05-20
⚡  add a note about WebSocket token refresh

martinbonnin:
This is a question that comes back every once in a while and it's not an easy topic so I figured out we might as well add a doc for it.

### DIFF
--- a/docs/source/advanced/authentication.mdx
+++ b/docs/source/advanced/authentication.mdx
@@ -76,7 +76,7 @@ val apolloClient = ApolloClient.Builder()
 
 ### Updating your WebSocket token
 
-When using subscriptions over WebSockets, you might want to refresh your token. Because it was invalidated or simply because it expired.
+When using authenticated subscriptions over WebSockets, you might want to refresh your token. Because it was invalidated or simply because it expired.
 
 > Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead to avoid sharing cache and/or other state between users.
 
--- a/docs/source/advanced/authentication.mdx
+++ b/docs/source/advanced/authentication.mdx
@@ -78,7 +78,7 @@ val apolloClient = ApolloClient.Builder()
 
 When using subscriptions over WebSockets, you might want to refresh your token. Because it was invalidated or simply because it expired.
 
-> Note: if your user logged out, you might want to consider creating a new `ApolloClient` to avoid sharing cache and/or other state between users.
+> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead to avoid sharing cache and/or other state between users.
 
 Sometimes the websocket will emit an error that you can use in `webSocketReopenWhen` to create a new WebSocket. Other times that information will come from elsewhere in your app and you will have to force disconnect the WebSocket.
 
--- a/docs/source/advanced/authentication.mdx
+++ b/docs/source/advanced/authentication.mdx
@@ -78,7 +78,7 @@ val apolloClient = ApolloClient.Builder()
 
 When using authenticated subscriptions over WebSockets, you might want to refresh your authentication token because it was invalidated or simply because it expired.
 
-> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead to avoid sharing cache and/or other state between users.
+> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead. This will make sure that all other state besides the WebSocket, like cache for an example, is new and not shared amongst different users.
 
 Sometimes the websocket will emit an error that you can use in `webSocketReopenWhen` to create a new WebSocket. Other times that information will come from elsewhere in your app and you will have to force disconnect the WebSocket.
 
--- a/docs/source/advanced/authentication.mdx
+++ b/docs/source/advanced/authentication.mdx
@@ -76,7 +76,7 @@ val apolloClient = ApolloClient.Builder()
 
 ### Updating your WebSocket token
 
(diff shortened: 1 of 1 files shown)


docs/source/advanced/authentication.mdx
```diff
@@ -74,4 +74,50 @@ val apolloClient = ApolloClient.Builder()
     .build()
 ```
 
+### Updating your WebSocket token
+
+When using authenticated subscriptions over WebSockets, you might want to refresh your token. Because it was invalidated or simply because it expired.
+
+> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead to avoid sharing cache and/or other state between users.
```
BoD, line 81:
```suggestion
> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead of avoiding sharing cache and/or other state between users.
```

docs/source/advanced/authentication.mdx
```diff
@@ -74,4 +74,50 @@ val apolloClient = ApolloClient.Builder()
     .build()
 ```
 
+### Updating your WebSocket token
+
+When using authenticated subscriptions over WebSockets, you might want to refresh your token. Because it was invalidated or simply because it expired.
```
BoD, line 79: (nit)
```suggestion
When using authenticated subscriptions over WebSockets, you might want to refresh your authentication token because it was invalidated or simply because it expired.
```

docs/source/advanced/authentication.mdx
```diff
@@ -74,4 +74,50 @@ val apolloClient = ApolloClient.Builder()
...
+            SubscriptionWsProtocol.Factory(
+                connectionPayload = {
+                  // the connection_init payload
+                  // This is an example. You will most likely have to tweak it for your backend
+                  mapOf("Authorization" to token)
+                }
+            )
+        )
+        .webSocketReopenWhen { e, attempt ->
+          if (WebSocketReconnectException) {
```
BoD, line 105:
```suggestion
          if (e is WebSocketReconnectException) {
```

docs/source/advanced/authentication.mdx
```diff
@@ -74,4 +74,50 @@ val apolloClient = ApolloClient.Builder()
...
+        .build()
+```
+
+When you need to reopen the WebSocket, call `closeConnection`:
+
+```kotlin
+apolloClient.subscriptionNetworkTransport.closeConnection(WebSocketReconnectException())`
+```
+
+The WebSocket will close and reopen with a fresh new `connectionPayload`
```
BoD, line 122:
```suggestion
The WebSocket will close and reopen with a fresh new `connectionPayload`.
```

BoD (approved):
👍

docs/source/advanced/authentication.mdx
martinbonnin: I'll rephrase a bit. I wanted to put the emphasis on the fact that caches do need to be separated.

martinbonnin: Updated with 

```
> Note: if your user logged out, you might want to consider creating a new `ApolloClient` instead. This will make sure that all other state besides the WebSocket, like cache for an example, is new and not shared amongst different users.
```

Let me know what you think

BoD: Yes, it is clearer 👍

AlexHartford:
Can this be expanded upon? This does not suffice as a standalone example. It's missing key details, like how is the custom exception thrown and caught?

martinbonnin:
Hi @AlexHartford 

`apolloClient.subscriptionNetworkTransport.closeConnection` is an [extension function](https://github.com/apollographql/apollo-kotlin/blob/e6c02d6cb9d0a24ef76af237f827d68a4ff57c27/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt#L420), you'll need to import it in your IDE. Usually Android Studio/IntelliJ provide assistance for that.

AlexHartford:
Thanks for the quick reply, let me give that a try.

AlexHartford:
After updating to the latest version of the SDK, my network errors are now being caught globally in `subscriptionNetworkTransport.reopenWhen`. Since it's not being caught in the retries of individual subscriptions, how can I close the connection so it tries to reauthenticate with a new token?

martinbonnin:
> I believe the issue is actually the serverUrl, since the serverUrl requires the auth as well, but it isn't updated when the subscription is reopened

It's been a while I haven't looked at the AppSync doc but the url authentication used an API key last time I checked:

```
  private val authorization = mapOf(
    "host" to "example1234567890000.appsync-api.us-east-1.amazonaws.com",
    "x-api-key" to "da2-12345678901234567890123456"
  )
```

So it's not supposed to expire? Or is it? Do you have any reference about this? We can add a `suspend` serverUrl if updating the url during the lifecycle of the app is needed.

AlexHartford:
Well, you can use many authentication methods for AppSync, but API keys aren't very secure so we are using Bearer auth through OpenID Connect (we use Firebase for auth). [Here's the docs](https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html) on the different authentication methods with AppSync.

Basically, the desired behavior would be any time the protocol is refreshed due to reopening the connection, the serverUrl would be updated as well. I could see this being implemented the same way using a Factory.

For what it's worth, as far as I know this isn't an issue with the Swift SDK.

AlexHartford:
https://github.com/apollographql/apollo-ios

The appsync SDK's are... not the best :)

Issue #4126, opened 2022-05-20
downloadApolloSchema on v3 don't fetch the interface


Canato:
## **Summary**
When using the command `./gradlew downloadApolloSchema --endpoint="https://...graphql/playground" --schema=."..."` the downloaded schema is not complete.

## **Version**
- `2.5.9` working
- `3.3.0` not working
Same endpoint

## **Description**
What I retrieve on apollo version `2.5.9`
```json
{
   "kind":"OBJECT",
   "name":"CustomerAddressInvalidCountry",
   "description":null,
   "fields":[
      {
         "name":"message",
         "description":null,
         "args":[
            
         ],
         "type":{
            "kind":"NON_NULL",
            "name":null,
            "ofType":{
               "kind":"SCALAR",
               "name":"String",
               "ofType":null
            }
         },
         "isDeprecated":false,
         "deprecationReason":null
      }
   ],
   "inputFields":null,
   "interfaces":[
      {
         "kind":"INTERFACE",
         "name":"Error",
         "ofType":null
      }
   ],
   "enumValues":null,
   "possibleTypes":null
}
```

What I retrieve on apollo version `3.3.0`
```json
{
   "kind":"OBJECT",
   "name":"CustomerAddressInvalidCountry",
   "fields":[
      {
         "name":"message",
         "isDeprecated":false,
         "type":{
            "kind":"NON_NULL",
            "ofType":{
               "kind":"SCALAR",
               "name":"String"
            }
         },
         "args":[
            
         ]
      }
   ]
}
```

What is breaking for me is the missing `interface` value. 
What broke my `CommonError` fragment. 

Because now we have
```kt
type CustomerAddressInvalidCountry {
    message: String!
}
```
Instead of 
```kt
type CustomerAddressInvalidCountry implements Error {
    message: String!
}
```

BoD:
HI! 

Is the endpoint publicly accessible, so we can have a look?

Canato:
Sadly is not @BoD =/ 

But I call it twice using the commit before we update the lib and the commit after =/

Canato:
Perfect! thanks =D

BoD:
Fix available in v3.3.1.

PR #4129, opened 2022-05-20
Keep the 'interfaces' field on the JSON introspection

BoD:
Related to #4126.

I'm wondering if we should save the introspection JSON as-is instead of parsing it and then serializing it, in `SchemaDownloader`? 🤔


martinbonnin (approved):
> I'm wondering if we should save the introspection JSON as-is instead of parsing it and then serializing it, in SchemaDownloader? 

That'd avoid this kind of things moving forwad. I'm all for it 👍

martinbonnin (approved):

PR #4130, opened 2022-05-22
🧶 Use a getter instead of a const val OPERATION_QUERY

ephemient:
Because enormous (64kB) strings cannot fit into a single constant in JVM
bytecode, but the Kotlin compiler will compile huge string literals as
runtime concatenation of smaller constants.

Using a getter instead of a val with a backing field so that for most
users, this does not have any practical impact (as r8 should be able to
inline the simple getter function into a string constant load).


martinbonnin (approved):
Thanks!

BoD (approved):

PR #4131, opened 2022-05-23
Allow @typePolicy directive on interfaces

### LINKED ISSUE #3356: Support @typePolicy for interfaces

[typePolicy](https://github.com/apollographql/apollo-android/blob/4a402d5ccc4e1a313dbc3918c75eb68239e537e6/apollo-ast/src/main/resources/apollo.graphqls) is currently only supported on objects.

```
directive @typePolicy(keyFields: String!) on OBJECT
```

This means we have to define the policy for every object even if it inherits from an interface which contains a common ID field.

in the example from the release notes:

```kotlin
    assertEquals(
        setOf(Cat.type, Dog.type, Crocodile.type),
        __Schema.possibleTypes(Animal.type).toSet()
    )
```
we have to define `typePolicy` for all Types of `Animal`

```
extend type Cat @typePolicy(keyFields: "id")
extend type Dog @typePolicy(keyFields: "id")
extend type Crocodile @typePolicy(keyFields: "id")
```

in addition if a new type of Animal is being added we have to remember to add the `typePolicy` or risk running into cache errors.

Allowing `typePolicy` on the interface would allow us to just declare

```
extend interface Animal @typePolicy(keyFields: "id")
```

In case we need to override this behaviour we could still define on the type level as override

bubba:
Closes #3356 (if this is the desirable behaviour!)
This fixes a couple of things:
It changes the @typePolicy directive definition in apollo.graphqls to allow it to be used on interfaces.
And in order for it to be picked up, this extends GQLInterfaceTypeExtension with a directives field.
Finally, it also checks for interfaces as well as objects when checking the key fields

@martinbonnin


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
```diff
@@ -33,7 +33,9 @@ private fun CheckKeyFieldsScope.checkField(
     selections: List<GQLSelection>,
     parentType: String,
 ) {
-  schema.typeDefinitions.values.filterIsInstance<GQLObjectTypeDefinition>().forEach {
+  schema.typeDefinitions.values.filter {
+    it is GQLObjectTypeDefinition || it is GQLInterfaceTypeDefinition
+  }.forEach {
```
martinbonnin, line 38: I don't think we need to iterate the interfaces here, this check is used to detect cases like this:

```graphql
# schema.graphqls
extend type Cat @typePolicy(keyFields: "id")
```

```graphql
# operation.graphql
query GetCat {
  animal {
    # id is missing here
    species
  }
}
```

So it only cares about actual concrete objects because at the end of the day, all objects must have a concrete type.

martinbonnin (commented):
Thank you so much for this, this is really cool! 2 small comments but look good to me overall!

bubba:
@martinbonnin Where would be the best place to throw an exception, and what exception type should I use for a user facing error, i.e. ApolloException? I’ve currently added it inside schema.kt within the `keyFields` function, but it uses `error` which looks like it’s meant for internal errors.  I could alternatively check and throw exceptions in `check_key_fields.kt`

martinbonnin:
> what exception type should I use for a user facing error

Good question. We're historically been quite liberal in what exceptions are thrown in the Gradle codegen side of things. While the runtime side of things is using `ApolloException` everywhere, the codegen side of things uses a bunch of different exceptions. Ideally we should log [Issues](https://github.com/apollographql/apollo-kotlin/blob/f890e3408d61acaa60aa5f6f2a3da93256bdf1ce/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Issue.kt#L7) so that the codegen can output all issues at once without the user having to fix one issue, recompile, fix the next one, etc...

That being said, `Issue` is in the `apollo-ast` module and `checkKeyFields` is in the Apollo-specific `apollo-compiler` module. I guess this is why it just used "error" because it was easier.

For the purpose of this PR, feel free to throw `IllegalStateException` or just `error` and this should be fine. `ApolloException` is for runtime only. I don't even think it's in the compiler classpath.

> Where would be the best place to throw an exception

I think the best place is in `GQLDocument.validateAsSchema`. This validation doesn't need to know the operation, it can operate on the schema, which is way faster. Make a mapping of `TypeDefinition` -> `keyFields` and check that inside a given type hierarchy, all the keyFields are the same? Ultimately, we might want to factor in this code, i.e. precompile the mapping just after getting the schema and reuse that one from `checkKeyFields` (instead of computing it lazily as we do now)

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
```diff
@@ -185,3 +189,25 @@ private fun SchemaValidationScope.validateNoIntrospectionNames() {
...
+  val keyFields = mergedTypeDefinitions.filterValues { it.canHaveKeyFields() }
+      .values.associate { it.name to it.keyFields(mergedTypeDefinitions) }
+
+  mergedTypeDefinitions.values.filterIsInstance<GQLObjectTypeDefinition>().forEach { o ->
+    if (keyFields[o.name]!!.isNotEmpty()) {
+      val interfacesWithKfs = o.implementsInterfaces.map { it to keyFields[it]!! }.filter { it.second.isNotEmpty() }.map { it.first }
+      if (interfacesWithKfs.isNotEmpty()) {
+        registerIssue(
+            "Type '${o.name}' cannot have key fields since it implements the following interfaces which also have key fields: $interfacesWithKfs",
+            sourceLocation = o.sourceLocation
```
martinbonnin, line 208: 👍 my initial thoughts were "ensure the key fields are the same" but this is better.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gqltypedefinition.kt
```diff
@@ -44,3 +44,73 @@ fun GQLTypeDefinition.implementsAbstractType(schema: Schema): Boolean {
...
+    is GQLInterfaceTypeDefinition -> {
+      val kf = directives.toKeyFields()
+      if (kf != null) {
+        kf
+      } else {
+        val kfs = implementsInterfaces.map { it to typeDefinitions[it]!!.keyFields(typeDefinitions) }.filter { it.second.isNotEmpty() }
+        if (kfs.isNotEmpty()) {
+          check(kfs.size == 1) {
+            val candidates = kfs.joinToString("\n") { "${it.first}: ${it.second}" }
+            "Interface '$name' inherits different keys from different interfaces:\n$candidates\nSpecify @${Schema.TYPE_POLICY} explicitly"
```
martinbonnin, line 94: Same here 
```suggestion
            "Interface '$name' cannot inherit different keys from different interfaces:\n$candidates"
```

martinbonnin:
And [integration-tests](https://github.com/apollographql/apollo-kotlin/blob/8efa7629b233c90d1b51a27d732aa08f54ecc00f/tests/integration-tests/src/commonTest/kotlin/test/declarativecache/graphql/extra.graphqls#L7) had some tests using the previous logic of "overriding" the key fields. Sorry about this, things were a bit "in flight" (which is part of the reason why we disabled `@typePolicy` on interfaces). Let me know if it's too cumbersome, I can take it from there.

bubba:
Whoops, didn’t realise the integration tests were failing. Fixing now

bubba:
Potentially stupid question: How can I run the integration tests locally? Is there a specific gradle task for them or do I need to run the `allTests` task. IntelliJ isn’t suggesting to run them from within the IDE either

martinbonnin:
You'll need to open the `tests` project in IDE. It's a composite build that will import the integration tests.

Or you can run the tests from the command line with:

```
./gradlew -p tests allTest
```

BoD (approved):
🙏 👍

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
```diff
@@ -185,3 +186,79 @@ private fun SchemaValidationScope.validateNoIntrospectionNames() {
...
+  return directives.flatMap {
+    (it.arguments!!.arguments.first().value as GQLStringValue).value.buffer().parseAsGQLSelections().valueAssertNoErrors().map { gqlSelection ->
+      // No need to check here, this should be done during validation
+      (gqlSelection as GQLField).name
+    }
+  }.toSet()
+}
+
+/**
+ * To prevent surprising behaviour, objects that declare key fields that also implement interfaces that declare key fields are an error
```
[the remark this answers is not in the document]
martinbonnin: I added another comment on the method above and kept this one as it's the entry point of validation.

martinbonnin:
Thanks for the contribution!

bubba:
Hi, thanks for patching up the PR! I was able to pull it from the snapshots repository which was great, but it had some surprising behaviour with my own schema, my `extras.graphqls` is just this one line:

```graphql
extend interface Node @typePolicy(keyField: "id”)
```

And my server schema has types like this:
```graphql
interface Node {}
interface UniformResourceLocatable {}
# and some more interfaces etc.
type Foo implements Node, UniformResourceLocatable {}
```

With the latest version of this PR the compiler plugin errors because `Foo` inherits different keys from different interfaces:

```
Cause: e: /Users/luke/Source/minm-android/app/src/main/graphql/schema.graphqls: (10, 1): Apollo: Type 'Album' cannot inherit different keys from different interfaces:
LibraryItem: []
Node: [id]
UniformResourceLocatable: []
```

Even though `Node` is the only interface with a `@typePolicy` directive, the other interfaces technically have keyFields of `[]`. I’m not really sure why this is happening: the distinct key fields should just be the one (id):

https://github.com/bubba/apollo-kotlin/blob/7de9b02bda657abc9e09e4664bf0ea63988d1cd6/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt#L206-L216

martinbonnin:
Oops, my bad, I forgot to exclude the 'empty' key fields special case. I'll make a patch in a few

martinbonnin:
Fix is [there](https://github.com/apollographql/apollo-kotlin/pull/4131#issuecomment-1138429890), sorry for this. 

Interestingly, this does not distinguish between `interface Foo @typePolicy(keyFields: "")` and just `interface Foo`. It could be interesting to throw a validation error on `keyFields: ""` as I don't think there's a use case for this.

Issue #4137, opened 2022-05-25
SchemaBuilder not generating deterministic Schema.kt


eboudrant:
**Summary**
The `Schema.kt` is generated by Apollo in `build/generated/source/apollo/...` folder. The content of the file is different from CI and Local builds causing lot of miss in our Gradle build cache.

**Version**
3.2.2

**Description**
The difference in the file come from the `all` list :

Local build would have :
```
  public val all: List<CompiledType> = listOf(
        graphql.models.type.TypeA.type,
        graphql.models.type.TypeB.type,
        graphql.models.type.TypeC.type,
            graphql.models.type.TypeD.type)
```
CI would have : 
```
  public val all: List<CompiledType> = listOf(
        graphql.models.type.TypeA.type,
        graphql.models.type.TypeC.type,
        graphql.models.type.TypeB.type,
            graphql.models.type.TypeD.type)
```

PR #4138, opened 2022-05-25
Sort the type names in the list so the code gen is deterministic.

eboudrant:
Fix for #4137

### DIFF
--- a/apollo-compiler/src/test/graphql/com/example/measurements
+++ b/apollo-compiler/src/test/graphql/com/example/measurements
@@ -2,8 +2,8 @@
 // If you updated the codegen and test fixtures, you should commit this file too.
 
 Test:                                                                                      Total LOC:
-aggregate-all                                                                                  189197
-aggregate-kotlin-responseBased                                                                  48615
+aggregate-all                                                                                  189514
+aggregate-kotlin-responseBased                                                                  48932
 aggregate-kotlin-operationBased                                                                 34136
 aggregate-kotlin-compat                                                                         35553
 aggregate-java-operationBased                                                                   70893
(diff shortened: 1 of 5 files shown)


martinbonnin:
Thanks for sending this! Out of curiosity, are you seeing discrepancies only in `Schema.kt`? 
I was expecting `mutableMapOf<>()` to have stable iteration order so that the codegen would be stable too. If it's not the case we might have issues in other places too....

martinbonnin:
There's some info in the [CONTRIBUTING.md](https://github.com/apollographql/apollo-kotlin/blob/main/CONTRIBUTING.md#using-a-local-version-of-apollo-kotlin). TLDR; `./gradlew publishToMavenLocal`

eboudrant:
We're looking if there was a change in CI. Not at my knowledge but we've seen similar issue with Autovalue Gson code gen.

eboudrant:
@martinbonnin should I also update the class `com.apollographql.apollo3.compiler.codegen.java.file.SchemaBuilder`?

martinbonnin:
> should I also update the class com.apollographql.apollo3.compiler.codegen.java.file.SchemaBuilder

That'd be nice, if the bug exists with Kotlin codegen there's a high chance it is in the Java codegen too. Also keeps the both codegen symmetrical so 👍

apollo-compiler/src/test/graphql/com/example/measurements
```diff
@@ -240,6 +240,7 @@ kotlin-responseBased-subscriptions
 kotlin-responseBased-enums_as_sealed                                                              340
 kotlin-responseBased-operation_id_generator                                                       325
 kotlin-responseBased-merged_include                                                               323
+kotlin-responseBased-big_query                                                                    317
 kotlin-responseBased-enum_field                                                                   316
```
eboudrant, line 244: I'm not sure why this line was added?

martinbonnin: 2 PRs were merged in parallel:

* https://github.com/apollographql/apollo-kotlin/pull/4133/files added a test but did not update measurements (not sure why)
* https://github.com/apollographql/apollo-kotlin/pull/4132 changed the format of measurements

So in the end 4132 was merged without the new test that you just added. So thank you :) 

FWIW, this file is informative only. It's used to keep track of the size of codegen. It's handy to have it commited because it allows to easily bisect if there is a regression but we might as well put it in .gitignore

martinbonnin (approved):

BoD (approved):
👍

PR #4143, opened 2022-05-26
Parse the interface's interface field in introspection

bubba:
This fixes the schema downloading tool not generating the "implements" part of any interfaces it introspects.
Ran into it wondering why the generated types for my interfaces, weren't reporting the correct key fields, it was because the schema downloaded wasn't complete!

Wasn't sure how to write a test case for this: Happy to add one though if someone can nudge me in the right direction!


BoD:
Thanks a lot for this! 🙏

Just to be sure I understand, your case is that the schema has interfaces implementing other interfaces, but they don't appear in the resulting `.graphqls` file?

In terms of tests, there's [this one](https://github.com/apollographql/apollo-kotlin/blob/main/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/ConvertSchemaTests.kt#L65). You could tweak [the json schema](https://github.com/apollographql/apollo-kotlin/tree/main/apollo-gradle-plugin/testProjects/convertSchema/schemas) it's using to include your case.

bubba:
@BoD Yeah, the interfaces field on interface types weren't being parsed from the introspection query, and so not getting getting generated in the resulting schema file. 
Will add tests today

BoD (approved):
Thanks a lot! 👍 🙏

martinbonnin (approved):
Thanks!

PR #4147, opened 2022-05-30
Do not bypass websocketReopenWhen {}

martinbonnin:
When an error happened while initializing the websocket, it was bypassing `websocketReopenWhen{}` completely. Depending on the subscriptions retry policies, this could loop forever. 

Also add a WebSocket errors doc.


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -184,7 +193,7 @@ private constructor(
               )
             } catch (e: Exception) {
               // Error opening the websocket
-              mutableEvents.emit(NetworkError(e))
+              messages.send(NetworkError(e))
```
martinbonnin, line 196: This is the actual fix

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -292,7 +301,7 @@ private constructor(
           }
           apolloResponse
         }
-        is OperationError -> throw ApolloNetworkException("Operation error ${request.operation.name()}: ${response.payload}")
+        is OperationError -> throw SubscriptionOperationException(request.operation.name(), response.payload)
```
martinbonnin, line 304: That's a small behaviour change, I added it to the Changelog

BoD (approved):
👏

PR #4153, opened 2022-06-01
🧰  introduce apollo-tooling and apollo-cli

martinbonnin:
Shuffle a bit the dependencies:

* the introspection API goes into `apollo-ast` because they work with the `apollo-ast` Schema. It's a bit of a stretch (this module should be `apollo-language` or something like this) but I think it's okay.
* It is kept as a convenience in `apollo-compiler` even though it was technically experimental before so we can certainly remove this before 4.0
* `SchemaDownloader`, `SchemaUploader`, `RegisterOperations` go into `apollo-tooling`
* Both `apollo-cli` and `apollo-gradle-plugin` use `apollo-tooling` to expose functionality either through Gradle or through the cli
* There's a `apollo-kotlin.main.kts` helper script that's an easy way to start the cli. It'd be interesting to ditribute it on sdkman, npm, or another package manager one of these days.


apollo-cli/src/main/kotlin/com/apollographql/apollo3/cli/Main.kt
```diff
@@ -0,0 +1,17 @@
+package com.apollographql.apollo3.cli
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
+
+private class MainCommand: CliktCommand() {
+  override fun run() {
+    println("apollo-cli is experimental and might change in backward incomapatible ways")
```
BoD, line 8:
```suggestion
    println("apollo-cli is experimental and might change in backward incompatible ways")
```

apollo-cli/apollo-kotlin.main.kts
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env kotlin
+
+@file:Repository("https://repo.maven.apache.org/maven2/")
+@file:Repository("file://Users/mbonnin/.m2/repository")
```
BoD, line 4:
```suggestion
```
:)

apollo-cli/apollo-kotlin.main.kts
```diff
@@ -0,0 +1,10 @@
...
+
+@file:DependsOn("com.apollographql.apollo3:apollo-cli:3.3.1-SNAPSHOT")
```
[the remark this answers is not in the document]
martinbonnin: TBH I'm not really sure where I'm going with this. It's just an easy way to call into the cli without having to manually download the jar and invoke java. I'm happy to remove it and we can look into packaging/distribution later.

Issue #4152, opened 2022-06-01
Capitalized letters in schema fields - differences in v2 and v3


anderssandven:
Im using a Prisma based API which uses capitalized field names in types. V3 does not seem to like [this](https://github.com/apollographql/apollo-kotlin/blame/6216553fba26b560b2b11de73dcf963895e88e41/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt#L141). I cant find any documentation on this and it seems like big change. It makes it very annoying to create queries when we have to alias all types. Are there any other ways to go about this?

BoD:
Hi! Thanks for reporting this.

Indeed this was done because in the generated code, a class is generated for each composite fields, with the name of the field capitalized (`fooBar` -> `FooBar`). But if the field itself is named `FooBar` then there will be a conflict. 

This can be resolved with an alias but if your fields all are named with a capital letter this will be impractical indeed.

There is an option `flattenModels` to make the classes generated individually instead of nested, which avoid some clashes, but not all of them. I'm going to have a look at a way to fix the other clashes in the codeGen so this check can be disabled in conjunction with `flattenModels`.

anderssandven:
That would be great. It probably save people some time if its mentioned in the migration guide as well ;)

BoD:
A fix has just been merged where when using `flattenModels.set(true)`, capitalized fields names are accepted. The restriction still exists with nested models however.

This will be in the next release, in the meantime you can try it out with the [snapshots](https://github.com/apollographql/apollo-kotlin#snapshots).

anderssandven:
Great job! Looking forward to testing it :)

BoD:
Fix available in v3.3.1.

PR #4154, opened 2022-06-01
Allow capitalized field names if flattenModels is true

BoD:
- If flattenModels is true, then it's OK to have field names with the same name as the corresponding model - disable the validation in that case.
- Fix a few other cases where there were still name conflicts (Adapter and Selections)

Related to #4152


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/selections/CompiledSelectionsBuilder.kt
```diff
@@ -68,7 +68,8 @@ class CompiledSelectionsBuilder(
   }
 
   private fun List<GQLSelection>.walk(name: String, private: Boolean, parentType: String): List<FieldSpec> {
-    val propertyName = resolveNameClashes(usedNames, context.layout.propertyName(name))
+    val maybeEscapedName = if (private) context.layout.escapedPropertyName(name) else context.layout.propertyName(name)
+    val propertyName = resolveNameClashes(usedNames, maybeEscapedName)
```
martinbonnin, line 72: Looks like the only case where this is public is the `root` field? In that case,  you could rename the root field to `__root` (technically source breaking but I think that's ok?) and escape indiscriminately, just like for variables?

All in all, I think I'd do this:

* __typename -> never escape
* __root -> never escape
* variables -> always escape
* add a new `compiledSelectionsName()` that would always escape.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -87,7 +87,9 @@ abstract class CodegenLayout(
 
   // variables keep the same case as their declared name
   internal fun variableName(name: String) = regularIdentifier(name)
+  internal fun escapedVariableName(name: String) = regularIdentifier("_$name")
```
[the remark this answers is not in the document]
BoD: You're right! I thought `variableName` was also used elsewhere but it's actually not :)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/adapter/AdapterCommon.kt
```diff
@@ -116,6 +116,8 @@ internal fun readFromResponseCodeBlock(
 
   val checkedProperties = mutableSetOf<String>()
 
+  val typeName = context.layout.escapedVariableName(__typename)
```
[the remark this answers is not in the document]
BoD: Yeah but the code above prefixes all fields indiscriminately so I thought let's escape everything to avoid special cases but it looks like it may be actually more complex like this 😅. I'll see if I can simplify.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Issue.kt
```diff
@@ -42,6 +42,7 @@ sealed class Issue(
   /**
    * Upper case fields are not supported as Kotlin doesn't allow a property name with the same name as a nested class.
    * If this happens, the easiest solution is to add an alias with a lower case first letter.
+   * The Apollo compiler option `flattenModels` can also be used in which case the check is skipped.
```
martinbonnin, line 45:
```suggestion
   * When models are nested, upper case fields are not supported as Kotlin doesn't allow a property name
   * with the same name as a nested class.
   * If this happens, the easiest solution is to add an alias with a lower case first letter.
   * If there are a lot of such fields, the Apollo compiler option `flattenModels` can also be used to circumvent this
   * error at the price of possible suffixes in model names.
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -93,6 +93,10 @@ object ApolloCompiler {
 
     findApolloReservedEnumValueNames(schema).checkNoErrors()
 
+    if (!options.flattenModels) {
+      findCapitalizedFields(definitions + incomingFragments).checkNoErrors()
```
martinbonnin, line 97: No need to check the incomingFragments, they'll be checked by upstream modules:

```suggestion
      findCapitalizedFields(definitions).checkNoErrors()
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkNoCapitalizedFields.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.apollographql.apollo3.ast.GQLFragmentDefinition
+import com.apollographql.apollo3.ast.GQLFragmentSpread
+import com.apollographql.apollo3.ast.GQLInlineFragment
+import com.apollographql.apollo3.ast.GQLOperationDefinition
+import com.apollographql.apollo3.ast.GQLSelection
+import com.apollographql.apollo3.ast.Issue
+import com.apollographql.apollo3.ast.internal.IssuesScope
+
+@ApolloInternal
+fun findCapitalizedFields(definitions: List<GQLDefinition>): List<Issue> {
```
martinbonnin, line 15: Super nit. I'd expect a method called "find" to return capitalized fields. Use "check" instead?

```suggestion
fun checkCapitalizedFields(definitions: List<GQLDefinition>): List<Issue> {
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkNoCapitalizedFields.kt
```diff
@@ -0,0 +1,85 @@
...
+      }
+      is GQLInlineFragment -> findCapitalizedFields(it.selectionSet.selections)
+      is GQLFragmentSpread -> findCapitalizedFields(fragmentsByName[it.name]!!.selectionSet.selections)
+    }
+  }
+}
+
+private interface ValidationScope : IssuesScope {
+  override val issues: MutableList<Issue>
+  val fragmentsByName: Map<String, GQLFragmentDefinition>
```
martinbonnin, line 66: Random thoughts: It'd be interesting to see if the Kotlin 1.7 context receivers can make the whole "scopes" thing a bit easier to diggest

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/adapter/AdapterCommon.kt
```diff
@@ -77,7 +77,7 @@ internal fun readFromResponseCodeBlock(
     CodeBlock.of(
         "$T $L = $L;",
         context.resolver.resolveIrType(property.info.type),
-        context.layout.variableName(property.info.responseName),
+        property.responseName(context),
```
martinbonnin, line 80: `responseName` has a big "GraphQL" connotation, I'd avoid using it for kotlin names. 

For some general context about what I was trying to achieve with `CodegenLayout`, `CodegenLayout` takes GraphQL names as input and outputs "Target" names where "Target" is either Java or Kotlin.

In that specific case, I'd put the check for `__typename` in `CodegenLayout.variableName` itself. I don't know if that makes sense?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/selections/CompiledSelectionsBuilder.kt
```diff
@@ -68,7 +68,8 @@ class CompiledSelectionsBuilder(
   }
 
   private fun List<GQLSelection>.walk(name: String, private: Boolean, parentType: String): List<FieldSpec> {
-    val propertyName = resolveNameClashes(usedNames, context.layout.propertyName(name))
+    val modelName = if (name == root) name else context.layout.compiledSelectionsName(name)
```
martinbonnin, line 71: Nit: you could rename `private` to `isRoot` and use that instead of doing a string comparison here

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkNoCapitalizedFields.kt
BoD: Good point! I've renamed the 2 other `find` as well.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/adapter/AdapterCommon.kt
BoD: I was hesitating about that 😅 Yeah thanks for the context it makes sense! Ok to do it in `variableName` then.

apollo-compiler/api/apollo-compiler.api
```diff
@@ -1,3 +1,9 @@
+public final class com/apollographql/apollo3/compiler/-checkApolloReservedEnumValueNames {
+}
+
+public final class com/apollographql/apollo3/compiler/-checkCapitalizedFields {
```
martinbonnin, line 4: I think that's a false positive. Just filed https://github.com/Kotlin/binary-compatibility-validator/issues/87

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/selections/CompiledSelectionsBuilder.kt
```diff
@@ -67,15 +67,16 @@ class CompiledSelectionsBuilder(
...
+    val modelName = if (isRoot) root else context.layout.compiledSelectionsName(name)
+    val propertyName = resolveNameClashes(usedNames, modelName)
 
     val results = mapNotNull { it.walk(true, parentType) }
 
     val property = FieldSpec.builder(ParameterizedTypeName.get(JavaClassNames.List, JavaClassNames.CompiledSelection), propertyName)
         .initializer(results.map { it.initializer }.toListInitializerCodeblock(withNewLines = true))
         .addModifiers(Modifier.STATIC)
-        .addModifiers(if (private) Modifier.PRIVATE else Modifier.PUBLIC)
+        .addModifiers(if (isRoot) Modifier.PRIVATE else Modifier.PUBLIC)
```
BoD, line 79: This is incorrect of course 🤦‍♀️. Fixing it right now.

martinbonnin, line 79: opposite?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/selections/CompiledSelectionsBuilder.kt
```diff
@@ -77,7 +78,7 @@ class CompiledSelectionsBuilder(
 
     val property = PropertySpec.builder(propertyName, KotlinSymbols.List.parameterizedBy(KotlinSymbols.CompiledSelection))
         .initializer(builder.build())
-        .applyIf(private) {
+        .applyIf(isRoot) {
```
martinbonnin, line 81: opposite too

apollo-compiler/src/test/graphql/com/example/antlr_tokens/java/operationBased/antlr_tokens/adapter/TestQuery_ResponseAdapter.java.expected
```diff
@@ -59,24 +59,24 @@ public class TestQuery_ResponseAdapter {
     @Override
     public TestQuery.TypeWithGraphQLKeywords fromJson(JsonReader reader,
         CustomScalarAdapters customScalarAdapters) throws IOException {
-      String on = null;
-      String null_ = null;
-      String alias = null;
+      String _on = null;
+      String _null = null;
+      String _alias = null;
```
martinbonnin, line 64: That's kind of nice as it handles the `null` case too :)

martinbonnin (approved):

martinbonnin:
🚀

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/selections/CompiledSelectionsBuilder.kt
martinbonnin: Digging into this again. Do you remember why we needed to escape the compiled selections? They look like this:

```kotlin
public object TestQuerySelections {
  private val __onDroid: List<CompiledSelection> = listOf(..)
  private val __hero: List<CompiledSelection> = listOf(..)
  // ...
  public val __root: List<CompiledSelection> = listOf(..)
}
```

This is all private (except __root) and this file has some code to resolve name clashes: https://github.com/apollographql/apollo-kotlin/blob/89a439cbaa1a9b6bc02427df7f2d13abf8e953f3/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/selections/CompiledSelectionsBuilder.kt#L51. So escaping doesn't sound needed? Do you remember the details?

BoD: You're right, it looks like escaping is unnecessary - not sure why we did this 😅 Maybe we overlooked the fact that only `__root` is public?

PR #4156, opened 2022-06-02
⏳ Add a test for server side expiration

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/3566


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -310,14 +317,61 @@ fun <T> MutableExecutionOptions<T>.storePartialResponses(storePartialResponses:
...
+/**
+ * @param storeExpirationDate Whether to store the expiration date in the cache.
+ *
+ * The expiration date is computed from the response HTTP headers
+ *
+ * Default: false
+ */
+@ApolloExperimental
+fun ApolloClient.Builder.storeExpirationDate(storeExpirationDate: Boolean): ApolloClient.Builder = apply {
+  addExecutionContext(StoreExpirationDateContext(storeExpirationDate))
```
BoD, line 338: I think this is unused but maybe I'm missing something?

martinbonnin: You are 100% right. The initial idea was to make this configurable but I lost track at some point. Looking into this.

BoD (approved):
👍

PR #4161, opened 2022-06-02
🧽  reorganize dependencies.gradle

martinbonnin:
Something I've been wanting to do for a while: flatten the dependencies and order them alphabetically so that it's easier to find them. No functional change


BoD (approved):
👍

PR #4162, opened 2022-06-02
Fix incorrect content-length in MockServer

BoD:
When a string contains non 8 bit characters, its Java length is not equal to its utf-8 encoded length. This caused receiving truncated responses from MockServer.

Thanks @StylianosGakis for finding this! 🙏


martinbonnin (approved):
👍

PR #4164, opened 2022-06-03
Add a command to dump the completion script

martinbonnin:
You can now dump the completion with

```
apollo_kotlin.main.kts generate-completion --shell zsh --name apollo_kotlin.main.kts > ~/.apollo_kotlin_completion
```


BoD (approved):
Neat!

PR #4166, opened 2022-06-03
☕ make it possible to build the project with Java17

martinbonnin:
There were 2 issues:

* AGP using [Gson + Reflection](https://github.com/google/gson/issues/1979)
* Gradle 5.6 not supporting Java-17 for some Groovy reason

Leverage [Gradle toolchains](https://docs.gradle.org/current/userguide/toolchains.html) to compile with Java8 and run tests with Java11.


gradle/dependencies.gradle
```diff
@@ -1,5 +1,6 @@
 def versions = [
-    androidPlugin       : '4.2.2',
+    androidPlugin       : '7.0.0',
```
martinbonnin, line 2: This version bump is actually not required if using toolchains but since IntelliJ now supports 7.0.0 it shouldn't hurt to bump.

BoD (approved):
👍

martinbonnin:
@BoD I just pushed a couple of commits to use only one toolchain in CI (11) and instead set the target version to 8 using compiler options. This might shave a bit of time in CI. Do you mind checking again (sorry for the last minute change)

BoD:
Yes no worries, and this looks good! 👍

BoD (approved):

martinbonnin:
FWIW, that's the latest CI run:

Screenshot 2022-06-06 at 11 44 38

Compared to before:

Screenshot 2022-06-06 at 11 45 53

Not a huge difference (espectially, since the times are not 100% reproducible) but I like to see we gained a couple of minutes.

PR #4174, opened 2022-06-06
🏷️  Allow namespacing and renaming of directives

martinbonnin:
## namespacing/renaming

This PR adds support for namespacing and renaming directive (and other definitions) using the same `@link` directive as in Federation2:

```graphql
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", 
                    import: ["@optional", "@nonnull", "@typePolicy", "@fieldPolicy"])
```

[In a future version](https://github.com/apollographql/apollo-kotlin/issues/4171) (but not yet), in order to use these directives in `extra.graphqls` or in client queries, users will have to opt-in with the above snippet. 

This PR keeps the existing behaviour to minimize friction while upgrading:
1. operation directives ignore the absence of a matching directive definition
2. apollo definitions are added after schema validation and overwrite any pre-existing definition

[This will be changed in a future version](https://github.com/apollographql/apollo-kotlin/issues/4171) in order to support proper validation of operation directives/

All in all, the current behaviour is:
* schema has existing `@nonnull` definition -> the directive is overridden by the Apollo directive. If it's not the desired behaviour, the user has to import `@nonnull` with a different name (`@apollo__nonnull` for an example)
* schema does not have existing `@nonnull` definition -> the operation `@nonnull` usages are validated against the auto-added Apollo `@nonnull`
* schema does not have existing (non-apollo) `@foo` definition -> the operation `@nonnull` usages are note validated

The target behaviour is:
* schema has existing `@nonnull` definition -> the operation `@nonnull` usages are validated against the existing definition. If it doesn't match, the user has to import `@nonnull` with a different name
* schema does not have existing `@nonnull` definition -> the user has to import `@nonnull` using `@link`
* schema does not have existing (non-apollo) `@foo` definition -> the operation fails, the user has to add the directive to their schema

PS: because this PR is big already, it does include support for [renaming `@link` itself](https://github.com/apollographql/apollo-kotlin/issues/4173)


apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
```diff
@@ -149,6 +203,167 @@ internal fun ValidationScope.syntheticSchemaDefinition(): GQLSchemaDefinition {
...
+
+        if (components.size < 2) {
+          issues.add(Issue.ValidationError("Invalid @link url: 'url'", gqlDirective.sourceLocation))
+          return@eachDirective
+        }
+
+        val foreignName = components[components.size - 2]
+
+        if (prefix == null) {
+          prefix = foreignName
```
BoD, line 268: If I understand correctly it means that `extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", import: ["@nonnull"])` won't import `@nonnull` but `@kotlin_labs__nonnull`?

martinbonnin: Not exactly:

```graphql
# import removes the namespace
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", import: ["@nonnull"])

type Foo {
  field: String! @nonnull
}
```

```graphql
# without import uses the default namespace
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1")

type Foo {
  field: String! @kotlin_labs__nonnull
}
```

```graphql
# as changes the namespace
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", as: "apollo")

type Foo {
  field: String! @apollo__nonnull
}
```

BoD: Ah ok got it! 👍  Makes sense actually 😊

martinbonnin: I think the confusing part is that "import" is more about "rename" than "import" but once you get past that, it all makes sense

BoD (approved):
👏

PR #4175, opened 2022-06-06
@experimental -> @requiresOptIn

martinbonnin:
Rename `@experimental` to `@requiresOptIn`

Also: 
- support renaming `@targetName`
- add tests that read the opt-in annotations using reflection


apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -519,16 +517,17 @@ interface Service {
   val sealedClassesForEnumsMatching: ListProperty<String>
 
   /**
-   * The annotation to use for experimental fields/inputFields/enumValues
+   * The annotation to use for `@requiresOptIn` fields/inputFields/enumValues
```
BoD, line 520: I know this is all still "cooking" :) but it would make sense to have a mapping (feature -> annotation) instead?

So for instance
```graphql
type Foo {
  field: String
  betaField: String @requiresOptIn(feature: "beta")
  alphaField: String @requiresOptIn(feature: "alpha")
}
```

could generate 
```kotlin
data class Foo(
  field: String,
  @Beta
  betaField: String,
  @Alpha
  alphaField: String,
)
```

BoD (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/KDoc.kt
```diff
@@ -55,29 +55,29 @@ internal fun ParameterSpec.Builder.maybeAddDeprecation(deprecationReason: String
   return addAnnotation(deprecatedAnnotation(deprecationReason))
 }
 
-internal fun TypeSpec.Builder.maybeAddExperimental(resolver: KotlinResolver, experimentalReason: String?): TypeSpec.Builder {
-  if (experimentalReason.isNullOrBlank()) {
+internal fun TypeSpec.Builder.maybeAddExperimental(resolver: KotlinResolver, optInFeature: String?): TypeSpec.Builder {
```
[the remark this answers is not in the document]
martinbonnin: Yup, this one fell through the cracks, fixing it now. Thanks for catching this!

apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
martinbonnin: Yep, good point. Ultimately, I think we'll want to do this with GraphQL configuration, not Gradle configuration:

```graphql
extend directive @requiresOptIn @map(feature: "alpha", annotation: "com.example.Alpha")
                     @map(feature: "beta", annotation: "com.example.Beta")
```

or if we decide to go the "markers" road:

```graphql
directive @alpha on FIELD @requiresOptIn
directive @beta on FIELD @requiresOptIn

extend directive @alpha @targetName(name "com.example.Alpha")
extend directive @beta @targetName(name "com.example.Beta")
```


What do you think?

BoD: Oh yeah doing this with GraphQL is nice! I guess it'll depend on which direction the RFC goes but both approaches look good to me :)

Issue #4183, opened 2022-06-13
Apollo v2 Kotlin 1.7.0 incompatibility


remcomokveld:
**Summary**
The v2 com.apollographql.apollo Gradle fails to configure the project after updating Kotlin to 1.7.0.

**Version**
2.5.11

**Description**
I completely understand if you don't want to invest any additional time in the 2.x releases so I'm also happy to dig deeper and put up a PR to fix this, but it would be nice to not force users of Apollo 2.x to stick to Kotlin 1.6 until they've migrated to 1.7.

TBH I have not yet looked at the upgrade path for 3.x so if there is strong recommedation to do that instead, also happy to do that. But if the fix for 2.x is simple I think that would be nice for anyone else that's not fully up-to-date on their apollo updates.

I created a bare mimimum sample repository here. https://github.com/remcomokveld/apollo-kotlin-repro

BoD:
Hi! Thank you for reporting this, and the repro project!

Indeed we recommend moving to to 3.x, however if a fix is possible, a new release of 2.x is not out of the question. If you want to investigate you are welcome!

martinbonnin:
+1 to doing a maintenance release there. 

@remcomokveld also the upgrade guide to 3.x is there: https://www.apollographql.com/docs/kotlin/migration/3.0.

BoD:
The issue was a signature change in the KotlinCompile task. A fix is in #4187, and we'll release a new 2.x version soon.

remcomokveld:
Thank you! 🙏

BoD:
The fix is in v2.5.12 which was just released.

PR #4187, opened 2022-06-13
Fix Kotlin 1.7 compatibility

BoD:
Fix for #4183.

This is the same issue/fix as https://github.com/square/wire/pull/2273

### DIFF
--- a/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/KotlinCompileCompat.kt
+++ b/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/KotlinCompileCompat.kt
@@ -2,6 +2,7 @@ package com.apollographql.apollo.gradle.internal
 
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
+// From https://github.com/square/wire/pull/2273:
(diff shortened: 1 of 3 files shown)


apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/KotlinCompileCompat.kt
```diff
@@ -0,0 +1,14 @@
+package com.apollographql.apollo.gradle.internal
+
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+// The signature of this function changed in Kotlin 1.7, so we invoke it reflectively
+// to support both.
+// 1.6.x: `fun source(vararg sources: Any): SourceTask
+// 1.7.x: `fun source(vararg sources: Any)
```
martinbonnin, line 8: Attribute the fix in the source too?

```suggestion
// From https://github.com/square/wire/pull/2273
// The signature of this function changed in Kotlin 1.7, so we invoke it reflectively
// to support both.
// 1.6.x: `fun source(vararg sources: Any): SourceTask
// 1.7.x: `fun source(vararg sources: Any)
```

martinbonnin (approved):

PR #4189, opened 2022-06-13
Introduce ApolloResolver and use it to allow max-stale queries

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/4157

Can be used like so:

```
    val response1 = client.query(GetUserQuery()).fetchPolicy(FetchPolicy.CacheOnly)
        .cacheHeaders(CacheHeaders.Builder().addHeader(ApolloCacheHeaders.MAX_STALE, "10").build())
        .execute()
```

See integration tests for an example


apollo-normalized-cache-api/api/apollo-normalized-cache-api.api
```diff
@@ -136,7 +142,8 @@ public final class com/apollographql/apollo3/cache/normalized/api/OperationCache
...
-	public static final fun readDataFromCache (Lcom/apollographql/apollo3/api/Fragment;Lcom/apollographql/apollo3/cache/normalized/api/CacheKey;Lcom/apollographql/apollo3/api/CustomScalarAdapters;Lcom/apollographql/apollo3/cache/normalized/api/ReadOnlyNormalizedCache;Lcom/apollographql/apollo3/cache/normalized/api/CacheResolver;Lcom/apollographql/apollo3/cache/normalized/api/CacheHeaders;)Lcom/apollographql/apollo3/api/Fragment$Data;
+	public static final fun readDataFromCache (Lcom/apollographql/apollo3/api/Executable;Lcom/apollographql/apollo3/cache/normalized/api/CacheKey;Lcom/apollographql/apollo3/api/CustomScalarAdapters;Lcom/apollographql/apollo3/cache/normalized/api/ReadOnlyNormalizedCache;Lcom/apollographql/apollo3/cache/normalized/api/ApolloResolver;Lcom/apollographql/apollo3/cache/normalized/api/CacheHeaders;)Lcom/apollographql/apollo3/api/Executable$Data;
```
martinbonnin, line 145: I'm having a hard time figuring out if this is a compatible change. Since `Fragment` implements `Executable` and same for `Fragment.Data` and `Executable.Data` I would say so but maybe I'm missing something?

BoD: Could we keep the existing Fragment one in addition to the new one (and mark it as deprecated maybe)?

BoD (approved):
Do you think of making `CacheResolver` deprecated (not now since `ApolloResolver` is experimental)?

Looks good 👍

martinbonnin:
> Do you think of making CacheResolver deprecated (not now since ApolloResolver is experimental)?

I think yes. We have the same issue in Adapters to with `fromJson()`and `CustomScalarAdapters` being the de-facto `Context` there. Lesson learned: in interface function, use a `Context` if there are more than a couple parameters to make it easier to evolve the API

apollo-normalized-cache-api/api/apollo-normalized-cache-api.api
BoD: In any case, the risk is that an app calls `Fragment.readDataFromCache` and that it will no longer resolve, right? Possible because it's public, but not likely?

martinbonnin: It should still resolve because `Fragment: Executable`

PR #4192, opened 2022-06-14
✅ update binary compat validation

martinbonnin:
This version removes `@ApolloExperimental` and `@ApolloInternal` symbols from the public API dumps

See https://github.com/Kotlin/binary-compatibility-validator/issues/58


BoD (approved):
👍

PR #4196, opened 2022-06-15
Add @typePolicy(embeddedFields: String! = "")

martinbonnin:
`@embed` allows "embedding" objects in their parent records. Will be useful for pagination. This schema:

```grahql
type Query @typePolicy(embeddedFields: "hero"){
  hero: Hero! 
}

type Hero @typePolicy(embeddedFields: "friends"){
  friends: [Friend!]! 
}

type Friend {
  name: String!
}
```

Will only produce a single Record. Note that it's opt-in because in the worst case where no object has an id, we could rewrite the whole cache on each query.


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -158,17 +158,111 @@ class CustomScalarType(
...
+  constructor(
+      name: String,
+      keyFields: List<String> = emptyList(),
+      implements: List<InterfaceType> = emptyList(),
+  ) : this(name, keyFields, implements, emptyList())
+
+  fun newBuilder(): Builder = Builder(this)
+
+  class Builder(internal val name: String) {
+    internal var keyFields: List<String> = emptyList()
```
BoD, line 182: Nit: could be `private`?

BoD (approved):

martinbonnin:
I updated this PR to leverage the existing `@typePolicy`, no only is this more in line with `keyFields` but it also makes possible to extend the existing types while keeping a GraphQL compatible syntax. Let me know what you think

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/Schema.kt
```diff
@@ -188,6 +193,7 @@ class Schema internal constructor(
     const val NONNULL = "nonnull"
     const val OPTIONAL = "optional"
     const val REQUIRES_OPT_IN = "requiresOptIn"
+    const val EMBED = "embed"
```
BoD, line 196: Can be deleted

kotlin-js-store/yarn.lock
```diff
@@ -49,11 +49,6 @@ argparse@^2.0.1:
   resolved "https://registry.yarnpkg.com/argparse/-/argparse-2.0.1.tgz#246f50f3ca78a3240f6c997e8a9bd1eac49e4b38"
   integrity sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==
 
-async-limiter@~1.0.0:
```
BoD, line 52: It happened again!

BoD (approved):

BoD:
> I updated this PR to leverage the existing @typePolicy, no only is this more in line with keyFields but it also makes possible to extend the existing types while keeping a GraphQL compatible syntax. Let me know what you think

Yes, even better, smaller API surface 👍

kotlin-js-store/yarn.lock
martinbonnin: Fixed with https://github.com/apollographql/apollo-kotlin/pull/4199

PR #4201, opened 2022-06-17
Fix NPE in checkCapitalizedFields

BoD:
Don't crash if a Fragment definition from upstream is not found when checking capitalized fields

### DIFF
--- a/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkCapitalizedFields.kt
+++ b/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkCapitalizedFields.kt
@@ -58,7 +58,7 @@ private fun ValidationScope.checkCapitalizedFields(selections: List<GQLSelection
         }
       }
       is GQLInlineFragment -> checkCapitalizedFields(it.selectionSet.selections)
-      is GQLFragmentSpread -> checkCapitalizedFields(fragmentsByName[it.name]!!.selectionSet.selections)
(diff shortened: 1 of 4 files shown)


martinbonnin (approved):

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkCapitalizedFields.kt
```diff
@@ -58,7 +58,7 @@ private fun ValidationScope.checkCapitalizedFields(selections: List<GQLSelection
         }
       }
       is GQLInlineFragment -> checkCapitalizedFields(it.selectionSet.selections)
-      is GQLFragmentSpread -> checkCapitalizedFields(fragmentsByName[it.name]!!.selectionSet.selections)
+      is GQLFragmentSpread -> fragmentsByName[it.name]?.let { fragment -> checkCapitalizedFields(fragment.selectionSet.selections) }
```
martinbonnin, line 61:
```suggestion
      // it might be that the fragment is defined in an upstream module. In that case, it is validated
      // already, no need to check it again
      is GQLFragmentSpread -> fragmentsByName[it.name]?.let { fragment -> checkCapitalizedFields(fragment.selectionSet.selections) }
```

BoD: 💯

PR #4207, opened 2022-06-17
Pagination: arguments and metadata on Record and MetadataGenerator

BoD:
Add `arguments` and `metadata` on `Record`. Arguments are set in Normalizer. Metadata are also set in Normaliizer but by implementing a new interface: `MetadataGenerator`.

(Also includes the fix for the embedded fields from the `embed-test` branch)

### DIFF
--- a/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
+++ b/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
@@ -31,14 +31,32 @@ class Record(
   var date: Map<String, Long?>? = null
     private set
 
+  /**
+   * The arguments of the field this Record represents in its parent.
+   */
+  @ApolloExperimental
+  var arguments: Map<String, Any?> = emptyMap()
+    private set
+
+  /**
+   * Arbitrary metadata that can be attached to a Record.
+   */
+  @ApolloExperimental
+  var metadata: Map<String, Any?> = emptyMap()
+    private set
+
   @ApolloInternal
   constructor(
       key: String,
       fields: Map<String, Any?>,
       mutationId: Uuid?,
       date: Map<String, Long?>,
+      arguments: Map<String, Any?>,
+      metadata: Map<String, Any?>,
   ) : this(key, fields, mutationId) {
     this.date = date
+    this.arguments = arguments
+    this.metadata = metadata
   }
 
   val sizeInBytes: Int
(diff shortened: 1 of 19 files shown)


martinbonnin (approved):

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
```diff
@@ -31,14 +31,32 @@ class Record(
   var date: Map<String, Long?>? = null
     private set
 
+  /**
+   * The arguments of the field this Record represents in its parent.
+   */
+  @ApolloExperimental
+  var arguments: Map<String, Any?> = emptyMap()
```
martinbonnin, line 38: As discussed, I think this should be a `Map<Map<String, Any?>>` to have the arguments per-field. Also there's always the case of list of scalars when you don't have a Record

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/Record.kt
```diff
@@ -31,14 +31,32 @@ class Record(
...
+    private set
+
+  /**
+   * Arbitrary metadata that can be attached to a Record.
+   */
+  @ApolloExperimental
+  var metadata: Map<String, Any?> = emptyMap()
```
martinbonnin, line 45: Maybe we should use this metadata to store the date and arguments 🤔 ?

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
```diff
@@ -89,10 +91,14 @@ internal class DefaultApolloStore(
       data: D,
       customScalarAdapters: CustomScalarAdapters,
   ): Map<String, Record> {
+    // Capture a local reference so as not to freeze "this"
+    val metadataGenerator = metadataGenerator
```
martinbonnin, line 95: Something's off here. There's no lambda capturing `metadataGenerator`. Also if there were, the same logic would apply to `cacheKeyGenerator`

tests/pagination/src/commonTest/kotlin/CursorPaginationTest.kt
```diff
@@ -0,0 +1,190 @@
...
+import com.apollographql.apollo3.cache.normalized.api.Record
+import com.apollographql.apollo3.cache.normalized.api.RecordMerger
+import com.apollographql.apollo3.cache.normalized.api.TypePolicyCacheKeyGenerator
+import com.apollographql.apollo3.cache.normalized.apolloStore
+import com.apollographql.apollo3.cache.normalized.normalizedCache
+import com.apollographql.apollo3.testing.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class RecordMergerTest {
```
martinbonnin, line 23: Rename to `class CursorPaginationTest` ? We usually give the test classes the same name as their containing file

BoD: ✅  in #4211

PR #4211, opened 2022-06-20
Persist Record arguments/metadata with the SQL Json backend

BoD:
Also, add a test/sample for offset based pagination.


tests/pagination/src/commonMain/graphql/schema.graphqls
```diff
@@ -1,5 +1,7 @@
 type Query {
-  users(first: Int = 10, after: String = null, last: Int = null, before: String = null): UserConnection!
+  usersCursorBased(first: Int = 10, after: String = null, last: Int = null, before: String = null): UserConnection!
+  usersOffsetBased1(offset: Int = 0, limit: Int = 10): [User!]!
```
BoD, line 3: Alas, I can't find a way to make this example work with the recent API changes. The fact that it returns an a list directly and not an object containing the list makes things difficult.

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/JsonRecordSerializer.kt
```diff
@@ -16,18 +17,30 @@ import okio.use
  */
 @ApolloInternal
 object JsonRecordSerializer {
+  private const val KEY_ARGUMENTS = "__arguments"
+  private const val KEY_METADATA = "__metadata"
```
martinbonnin, line 21: Let's use `__apm` in case https://github.com/graphql/graphql-spec/issues/300 lands one day and uses `__metadata` ?

(`ap` is for Apollo and `m` for metadata, since this is going to be written multiple times on disk, we might as well save a few bytes)

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/JsonRecordSerializer.kt
```diff
@@ -16,18 +17,30 @@ import okio.use
...
+      jsonWriter.name(KEY_ARGUMENTS).writeObject {
+        for ((key, value) in record.arguments) {
+          jsonWriter.name(key).writeJsonValue(value)
+        }
+      }
+      jsonWriter.name(KEY_METADATA).writeObject {
+        for ((key, value) in record.metadata) {
+          jsonWriter.name(key).writeJsonValue(value)
+        }
+      }
```
martinbonnin, line 43:
```suggestion
      jsonWriter.name(KEY_ARGUMENTS)
      jsonWriter.writeJsonValue(record.arguments)

      jsonWriter.name(KEY_METADATA)
      jsonWriter.writeJsonValue(record.metadata)
```

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/JsonRecordSerializer.kt
BoD: Good call!

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -307,7 +314,19 @@ class CompiledVariable(val name: String)
...
  * Note: for now, enums are mapped to Strings
  */
-class CompiledArgument(val name: String, val value: Any?, val isKey: Boolean = false)
+class CompiledArgument(
+    val name: String,
+    val value: Any?,
+    val isKey: Boolean = false,
+    @ApolloExperimental
+    val isPagination: Boolean = false,
+) {
```
martinbonnin, line 323:
```suggestion
class CompiledArgument @ApolloExperimental constructor(
    val name: String,
    val value: Any?,
    val isKey: Boolean = false,
    @ApolloExperimental
    val isPagination: Boolean = false,
) {
```

apollo-ast/api/apollo-ast.api
```diff
@@ -1115,6 +1115,7 @@ public final class com/apollographql/apollo3/ast/Schema {
 	public static final field FIELD_POLICY Ljava/lang/String;
 	public static final field FIELD_POLICY_FOR_FIELD Ljava/lang/String;
 	public static final field FIELD_POLICY_KEY_ARGS Ljava/lang/String;
+	public static final field FIELD_POLICY_PAGINATION_ARGS Ljava/lang/String;
```
martinbonnin, line 1118: This should not be part of the public API as it's marked `@ApolloExperimental` 🤔

BoD: I'm not sure why, but if I make it not `const` it disappears. But if I put `@JvmStatic` it comes back. Could this be a binary-compatibility-validator bug?

martinbonnin: Making it const removes the getter on the Companion object... I guess this is why. I'll investigate see if there's a quick fix for this.

martinbonnin: It's not a quick fix unfortunately :(. I opened https://github.com/Kotlin/binary-compatibility-validator/issues/90

apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -24,9 +24,10 @@ directive @nonnull(fields: String! = "") on OBJECT | FIELD
 # embedded in their parent instead. Order is unimportant.
 directive @typePolicy(keyFields: String! = "", embeddedFields: String! = "") on OBJECT | INTERFACE | UNION
 
-# Indicates how to compute a key from a field arguments.
-# `keyArgs` should contain a selection set. Composite args are not supported yet.
-directive @fieldPolicy(forField: String!, keyArgs: String!) repeatable on OBJECT
+# Attach extra information to a given field
+# `keyArgs`: a selection set used to compute a key from a field arguments. Composite args are not supported yet.
+# `paginationArgs` (experimental): a selection set used to compute a pagination key from a field arguments.
```
martinbonnin, line 29:
```suggestion
# `paginationArgs` (experimental): a selection set representing the arguments 
# that vary when requesting different pages. These arguments are omitted when 
# computing the cache key of this field.
```

apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -24,9 +24,10 @@ directive @nonnull(fields: String! = "") on OBJECT | FIELD
 # embedded in their parent instead. Order is unimportant.
 directive @typePolicy(keyFields: String! = "", embeddedFields: String! = "") on OBJECT | INTERFACE | UNION
 
-# Indicates how to compute a key from a field arguments.
-# `keyArgs` should contain a selection set. Composite args are not supported yet.
-directive @fieldPolicy(forField: String!, keyArgs: String!) repeatable on OBJECT
+# Attach extra information to a given field
+# `keyArgs`: a selection set used to compute a key from a field arguments. Composite args are not supported yet.
```
martinbonnin, line 28:
```suggestion
# `keyArgs`: a selection set used to compute the cache key of the object this field is pointing to. 
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/fieldPolicyArgs.kt
```diff
@@ -20,23 +30,23 @@ internal fun GQLTypeDefinition.keyArgs(fieldName: String, schema: Schema): Set<S
   return directives.filter { schema.originalDirectiveName(it.name) == Schema.FIELD_POLICY }.filter {
     (it.arguments?.arguments?.single { it.name == Schema.FIELD_POLICY_FOR_FIELD }?.value as GQLStringValue).value == fieldName
   }.flatMap {
-    val keyArgsValue = it.arguments?.arguments?.single { it.name == Schema.FIELD_POLICY_KEY_ARGS }?.value
+    val argValue = it.arguments?.arguments?.singleOrNull { it.name == argumentName }?.value ?: return emptySet()
 
-    if (keyArgsValue !is GQLStringValue) {
-      throw SourceAwareException("Apollo: no keyArgs found or wrong keyArgs type", it.sourceLocation)
+    if (argValue !is GQLStringValue) {
+      throw SourceAwareException("Apollo: no $argumentName found or wrong keyArgs type", it.sourceLocation)
```
martinbonnin, line 36:
```suggestion
      throw SourceAwareException("Apollo: no $argumentName found or wrong $argumentName type", it.sourceLocation)
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -307,7 +314,19 @@ class CompiledVariable(val name: String)
  *
  * Note: for now, enums are mapped to Strings
  */
-class CompiledArgument(val name: String, val value: Any?, val isKey: Boolean = false)
+class CompiledArgument @ApolloExperimental constructor(
+    val name: String,
+    val value: Any?,
+    val isKey: Boolean = false,
+    @ApolloExperimental
+    val isPagination: Boolean = false,
```
martinbonnin, line 322: Two parameters with default values => Sounds like it's time for a `Builder`?

tests/pagination/src/commonMain/graphql/schema.graphqls
```diff
@@ -1,5 +1,14 @@
-type Query {
-  users(first: Int = 10, after: String = null, last: Int = null, before: String = null): UserConnection!
+type Query
+@typePolicy(embeddedFields: "usersCursorBased, usersOffsetBasedWithPage")
```
martinbonnin, line 2: Do you need to embed `usersOffsetBasedWithPage ` here?

tests/pagination/src/commonTest/kotlin/CursorBasedPaginationTest.kt
```diff
@@ -0,0 +1,197 @@
...
+    val client = ApolloClient.Builder()
+        .normalizedCache(
+            normalizedCacheFactory = cacheFactory,
+            cacheKeyGenerator = TypePolicyCacheKeyGenerator,
+            metadataGenerator = CursorPaginationMetadataGenerator(),
+            apolloResolver = FieldPolicyApolloResolver,
+            recordMerger = CursorPaginationRecordMerger()
+        )
+        .serverUrl("unused")
+        .build()
```
martinbonnin, line 46: Looks like you're not using any of the `ApolloClient` functionality here? If that's the case, you can instanciate an `ApolloStore` directly:

```kolin
ApolloStore(
          normalizedCacheFactory = normalizedCacheFactory,
          cacheKeyGenerator = cacheKeyGenerator,
          metadataGenerator = metadataGenerator,
          apolloResolver = apolloResolver,
          recordMerger = recordMerger
      )
```

tests/pagination/src/commonTest/kotlin/OffsetBasedWithArrayPaginationTest.kt
```diff
@@ -0,0 +1,194 @@
...
+          val incomingOffset = incoming.arguments[fieldKey]!!["offset"] as? Int
+          if (existingOffset == null || incomingOffset == null) {
+            mergedFields[fieldKey] = incomingFieldValue
+            mergedMetadata[fieldKey] = incoming.metadata[fieldKey] as Map<String, Any?>
+          } else {
+            val existingList = existing[fieldKey] as List<*>
+            val incomingList = incomingFieldValue as List<*>
+            val mergedList = mergeLists(existingList, incomingList, existingOffset, incomingOffset)
+            mergedFields[fieldKey] = mergedList
+            mergedMetadata[fieldKey] = mapOf("offset" to min(existingOffset, incomingOffset))
```
martinbonnin, line 149: Using `min` doesn't take into account the case where existing and incoming do not join (the early returns in `mergeLists`)

apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/Normalizer.kt
```diff
@@ -83,14 +89,16 @@ internal class Normalizer(
...
-      fieldKey to replaceObjects(
+      val value = replaceObjects(
           entry.value,
           mergedField,
           mergedField.type,
           base.append(fieldKey),
           embeddedFields
       )
+      val arguments = mergedField.argumentsWithValue(variables)
+      val metadata = metadataGenerator.metadataForObject(entry.value, MetadataGeneratorContext(field = mergedField, variables))
```
martinbonnin, line 100: Would it be possible to store `arguments` in metadata? Looks like they're both computed in the same place and `metadataGenerator` has access to the arguments if it wants to.

This way it stores less data in the `Record` and the API is simpler?

BoD: Oh yeah, the `MetadataGenerator` can store arguments values if it needs them, but we can avoid doing it in the general case. I think that works 👍.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
BoD: Let's go :)

tests/pagination/src/commonMain/graphql/schema.graphqls
BoD: It's not obvious, but if we don't, we'll end up having to merge these 2 records:
`usersOffsetBasedWithPage` -> `{"users": ["cacheReference(User:42)", "cacheReference(User:43)"]}`
`usersOffsetBasedWithPage` -> `{"users": ["cacheReference(User:44)", "cacheReference(User:45)"]}`

but without any metadata.

What we need instead is to merge:
`ROOT_KEY` -> `{"usersOffsetBasedWithPage": {"users": ["cacheReference(User:42)", "cacheReference(User:43)"]}, (...)}` (+ metadata for `usersOffsetBasedWithPage`)
`ROOT_KEY` -> `{"usersOffsetBasedWithPage": {"users": ["cacheReference(User:44)", "cacheReference(User:45)"]}}` (+ metadata for `usersOffsetBasedWithPage`)

Hope it makes sense?

Also, maybe there's a problem there because I think it means that users will need to do this for all the query fields that paginate (at least with Relay style pagination). For some APIs that's probably a lot!

martinbonnin (approved):

PR #4214, opened 2022-06-21
🏖️  Relax MapJsonReader endObject, fixes reading inline + named fragments with compat models

### LINKED ISSUE #4212: Parsing error when having both inline and normal fragments at the same time
**Summary**
There seems to be a problem parsing a response of a union type where we're asking for the alternatives in both inline fragments and normal fragments at the same time. According to a discussion in the kotlin slack it seems to be a problem with rewinding the json reader and it not being at the end of the object at the end of parsing.

**Version**
3.3.1

**Description**
Given a schema with a section that looks like this:
```graphqls
union EmbarkResponse = EmbarkGroupedResponse|EmbarkResponseExpression|EmbarkMessage

fragment MessageFragment on EmbarkMessage {
  // stuff
}

fragment ResponseExpressionFragment on EmbarkResponseExpression {
  // stuff
}
```
And a query which at some point asks for a `EmbarkResponse` as a result of `response`
```graphqls
response {
  ...MessageFragment
  ...ResponseExpressionFragment
  ... on EmbarkGroupedResponse {
    // inline fragment stuff
  }
}
```

Parsing the response fails with an exception:
"Expected END_OBJECT but was NAME at path data.embarkStory.passages.5.response.__typename"

Adding this fragment:
```graphqls
fragment GroupedResponseFragment on EmbarkGroupedResponse {
  // stuff
}
```
And changing the query from the above to:
```graphql
response {
  ...MessageFragment
  ...ResponseExpressionFragment
  ...GroupedResponseFragment
}
```
works perfectly fine.

A lot more context exists in the discussion in the kotlin slack of the channel #apollo-kotlin [here](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1655733067620679)

martinbonnin:
Because the synthetic `Fragments` field was rewinding the stream, it caused leftover fields at the end of objects. 

Closes #4212

This is a simple, minimal fix. 

In order to support https://github.com/apollographql/apollo-kotlin/issues/3344, we'll need something more involved. Especially, because we know which models are buffered and which are not, I think we could generate code that doesn't use the `JsonReader` API for buffered adapters. That should be faster because there's no need to iterate all the keys for each fragments, the adapters could do a hashmap lookup instead. Because that's a bigger change, I've added it to the 4.0 list at https://github.com/apollographql/apollo-kotlin/issues/4171

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
@@ -156,9 +156,11 @@ constructor(
   }
 
   override fun endObject() = apply {
-    if (peek() != JsonReader.Token.END_OBJECT) {
-      throw JsonDataException("Expected END_OBJECT but was ${peek()} at path ${getPathAsString()}")
-    }
+    // Do not fail if there are trailing names in buffered readers.
+    // See https://github.com/apollographql/apollo-kotlin/issues/4212
+//    if (peek() != JsonReader.Token.END_OBJECT) {
+//      throw JsonDataException("Expected END_OBJECT but was ${peek()} at path ${getPathAsString()}")
+//    }
     stackSize--
     iteratorStack[stackSize] = null // allow garbage collection
     path[stackSize] = null // allow garbage collection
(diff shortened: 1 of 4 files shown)


BoD (approved):
👍

PR #4216, opened 2022-06-22
Pagination: FieldRecordMerger

BoD:
This is a tentative simplified API for the merging part. When merging Records it looks like the logic will be the same in most cases: iterate over the fields and merge them, handle `changedKeys` and `date`. With this, implementers can focus on the field merging part only. 

This still keeps the lower level `RecordMerger` API, even though at this point I'm not 100% sure if it's useful.


martinbonnin (approved):
This allocates twice in something that's potentially a hot path. Maybe it's an issue, maybe it's not. In all cases, it's not the only place where we're liberal with allocations so I assumes it's okay. We should really benchmark all of this though.

martinbonnin (approved):

PR #4226, opened 2022-06-27
Don't check git status on Kotlin 1.7 branch

BoD:
On the Kotlin 1.7 branch, this check fails because of differences in yarn.lock. Adds a way to ignore it with a flag.


BoD:
> could we put the 1.7 dependencies in yarn.lock and expect 1.6 to use these (because they are more recent and should be backward compatible)?

I think it should work, but `kotlinNpmInstall` overrides `kotlin-js-store/yarn.lock` so it's confusing 😅. If I understand correctly it builds with the versions in the checked-in file, and then updates it? If yes we could push the Kotlin 1.7 version but then the git status check will fail on Kotlin 1.6 builds.

martinbonnin:
Follow up PR there: https://github.com/apollographql/apollo-kotlin/pull/4230

PR #4227, opened 2022-06-27
One more example how to use Kotlin object in Java codegen for mapScalar

stengvac:
I was looking for way how to write Adapter for custom scalar in Kotlin with Java codegen target for some time. Without creating new instances.


martinbonnin (approved):
Thanks a lot for the contribution!

PR #4228, opened 2022-06-27
make it possible to reuse a File Upload

martinbonnin:
See #4054

In addition to making the File Upload reusable, this adds:

```
fun File.toUpload(contentType: String): Upload
fun Path.toUpload(contentType: String, fileSystem: FileSystem = systemFileSystem): Upload
```

### DIFF
--- a/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
+++ b/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
@@ -1,70 +1,87 @@
 package com.apollographql.apollo3.api
 
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import okio.BufferedSink
 import okio.BufferedSource
 import okio.ByteString
-import okio.ByteString.Companion.encodeUtf8
-import okio.ByteString.Companion.toByteString
+import okio.FileSystem
+import okio.Path
+import okio.buffer
 import okio.use
+import okio.utf8Size
 
 /**
- * An [Upload] that writes data from the provided content
- *
- * If the content is a [bufferedSource], the [DefaultUpload] will close it once uploaded
- * If the content is a [byteString], the [DefaultUpload] can be reused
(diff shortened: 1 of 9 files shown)


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
```diff
@@ -1,70 +1,87 @@
...
     private var contentType: String? = null
     private var contentLength: Long = -1
     private var fileName: String? = null
-    private val hasContent: Boolean
-      get() = bufferedSource != null || byteString != null
 
+    @Deprecated("This API is dangerous because the resulting upload can only be used once and can also lead to resource leaks.",
+        ReplaceWith("content {sink ->\nval source = openSource()\nsource.use {sink.writeAll(it)}\n}"))
+    @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v3_3_3)
     fun content(content: BufferedSource): Builder = apply {
```
martinbonnin, line 47: This method has a lot of potential foot guns so I marked it as deprecated but it might be going a bit too far. I'm fine keeping it too.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/DefaultUpload.kt
```diff
@@ -1,70 +1,87 @@
...
-      check(!hasContent) { "content() can only be called once" }
-      this.bufferedSource = content
+      check(writeTo == null) { "content() can only be called once" }
+      var consumed = false
+      this.writeTo = { sink ->
+        check(!consumed) { "Apollo: DefaultUpload body can only be read once. If you want to read it several times for logging or other purposes, either buffer it in memory or use your own `Upload` implementation." }
```
BoD, line 51:
```suggestion
        check(!consumed) { "Apollo: DefaultUpload BufferedSource body can only be read once. If you want to read it several times for logging or other purposes, either use a different kind of body or use your own `Upload` implementation." }
```

PR #4229, opened 2022-06-28
Add more cache benchmarks and run them in CI

martinbonnin:
This PR adds Cache benchmarks based on @sebj data. 

Thanks a lot to him for providing real life data to monitor the cache performance 💙 

It also adds a new CI target that runs the benchmarks in Firebase Test Lab.


scripts/run-benchmarks.main.kts
```diff
@@ -0,0 +1,286 @@
...
+import okio.buffer
+import okio.source
+import java.io.File
+import java.util.Date
+
+
+/**
+ * This script expects:
+ *
+ * - `gcloud` in the path
```
BoD, line 28: Is gcloud present by default on the GH actions image, or did I miss something?

BoD (approved):

scripts/run-benchmarks.main.kts
martinbonnin: Good question. Looks like [it is preinstalled indeed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#cli-tools)

BoD: Handy! 👍

martinbonnin:
On hold for https://github.com/gradle/gradle/issues/21125

martinbonnin:
> On hold for https://github.com/gradle/gradle/issues/21125

Never mind, we can use AGP alpha04 for now so we can stick with Gradle 7.4.2. Marking as ready for review again

BoD (approved):

PR #4230, opened 2022-06-28
use different kotlin-js-store for different Kotlin versions

martinbonnin:
Kudos to @turansky for suggesting this solution on [the Kotlin slack](https://kotlinlang.slack.com/archives/C0B8L3U69/p1656414273860479?thread_ts=1656413245.311039&cid=C0B8L3U69)


BoD (approved):
Very cool!

PR #4197, opened 2022-06-16
Introduce RecordMerger

BoD:

PR #4240, opened 2022-06-30
Revert cache changes

martinbonnin:
As there are more and more exploratory work on the cache, let's make a separate branch for this in order to keep main stable for releases.

- Revert "Pagination: fixes in FieldRecordMerger and MemoryCache (#4237)"
- Revert "Persist Record arguments/metadata with the SQL Json backend (#4211)"
- Revert "Introduce RecordMerger (#4197)"
- Revert "Add @typePolicy(embeddedFields: String! = "") (#4196)"
- Revert "Introduce ApolloResolver and use it to allow max-stale queries (#4189)"

They could all be reverted automatically except #4189 that touched `@ApolloDeprecatedSince` and required a small manual conflict resolution


BoD (approved):
👍

PR #4241, opened 2022-07-01
Incubating modules for the next gen cache

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/4240. Introduce

* `apollo-normalized-cache-incubating`
* `apollo-normalized-cache-sqlite-incubating`
* `apollo-normalized-cache-api-incubating`

This way we're making it explicit that the components are unstable


BoD (approved):
🚀

Issue #4238, opened 2022-06-30
Linking globals named 'OBJC_CLASS_$_GatewayApiLong': symbol multiply defined!


sentinelweb:
**Summary**
I am testing using the Kotlin Multiplatform version of Apollo. Our schema works fine when compiling for android but we are getting the above error when building iOS library : `./gradlew clean gateway_api:iosX64MainBinaries`

**Version**
3.3.2

**Description**
Add context about the problem here. How it happened and how to reproduce it:  

Our schema has no `scalar Long` declaration but there are quite a few places that use an adhoc scalar Long definition in fields (e.g. below)

I am guessing this is the cause of the above error so I wanted to check if simply declaring a  `scalar Long` in the schema and using it for fields would be the correct solution? As it might quite a big change for our organization. 

Or is there some compilation flag i could add that would ignore this error?

i have also tried adding `codegenModels.set("responseBased")` and modifying our code to use this codegen structure - but i ended up getting the same error after the change.
```
{
  "name": "netUnitsSold",
  "description": "Net unit sold across aggregation",
  "args": [],
  "type": {
    "kind": "SCALAR",
    "name": "Long",
    "ofType": null
  },
  "isDeprecated": false,
  "deprecationReason": null
},
```
This is the error stacktrace when executing `./gradlew clean gateway_api:iosX64MainBinaries` - same for `pod install` in XCode.
```
> Task :gateway_api:linkPodDebugFrameworkIosX64 FAILED
e: Compilation failed: Linking globals named 'OBJC_CLASS_$_GatewayApiLong': symbol multiply defined!

 * Source files: 
 * Compiler version info: Konan: 1.6.21 / Kotlin: 1.6.21
 * Output kind: FRAMEWORK

e: java.lang.Error: Linking globals named 'OBJC_CLASS_$_GatewayApiLong': symbol multiply defined!
        at org.jetbrains.kotlin.backend.konan.llvm.DefaultLlvmDiagnosticHandler.handle(diagnosticReport.kt:24)
        at org.jetbrains.kotlin.backend.konan.llvm.LlvmDiagnosticCollector.flush(diagnostics.kt:36)
        at org.jetbrains.kotlin.backend.konan.llvm.LinkModulesKt.llvmLinkModules2(linkModules.kt:36)
        at org.jetbrains.kotlin.backend.konan.llvm.objc.LinkObjCKt.linkObjC(linkObjC.kt:32)
[...]

PR #4243, opened 2022-07-05
Support @targetName on Interfaces, Enums, Unions, Scalars and Input objects

BoD:
This allows setting `@targetName` to change the generated type name on more things. This should help work around collisions such as #4238.


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/CustomScalarBuilder.kt
```diff
@@ -16,7 +16,7 @@ class CustomScalarBuilder(
 ): JavaClassBuilder {
   private val layout = context.layout
   private val packageName = layout.typePackageName()
-  private val simpleName = prefixBuiltinScalarNames(layout.compiledTypeName(name = customScalar.name))
+  private val simpleName = prefixBuiltinScalarNames(layout.compiledTypeName(name = customScalar.targetName))
```
martinbonnin, line 19: This is not needed because `compiledTypeName` uses `nameToClassName` under the hood

martinbonnin: More generally, I'd like `CodegenLayout` to take GraphQL names as input and outputs Kotlin names

apollo-compiler/api/apollo-compiler.api
```diff
@@ -1969,18 +1980,20 @@ public abstract class com/apollographql/apollo3/compiler/ir/IrType {
...
+	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;)Lcom/apollographql/apollo3/compiler/ir/IrUnion;
+	public static synthetic fun copy$default (Lcom/apollographql/apollo3/compiler/ir/IrUnion;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Lcom/apollographql/apollo3/compiler/ir/IrUnion;
 	public fun equals (Ljava/lang/Object;)Z
 	public final fun getDeprecationReason ()Ljava/lang/String;
 	public final fun getDescription ()Ljava/lang/String;
 	public final fun getMembers ()Ljava/util/List;
 	public fun getName ()Ljava/lang/String;
+	public fun getTargetName ()Ljava/lang/String;
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
```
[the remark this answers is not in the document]
BoD: The compiler doesn't like it because they are part of the public API of builders (`CustomScalarBuilder`, `EnumAsEnumBuilder`, etc.). I could make these internal too?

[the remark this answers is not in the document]
martinbonnin: Yep, the public API of `apollo-compiler` is just `ApolloCompiler` and `Options` at the moment

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/CustomScalarBuilder.kt
BoD: 👍 I've reverted to using `name` in builders, and map them to `targetName` via `nameToClassName` in c0d9274f1b10a7244b88e31d50a8d98e6d6ac2d6

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -269,33 +269,33 @@ object ApolloCompiler {
   }
 }
 
-internal fun maybeMakeNamesUnique(names: List<String>, outputDir: File): Map<String, String> {
+internal fun maybeMakeNamesUnique(names: Map<String, String>, outputDir: File): Map<String, String> {
```
martinbonnin, line 272: Right 👍 Keeping the original name is required.

apollo-compiler/api/apollo-compiler.api
```diff
@@ -1426,600 +964,6 @@ public final class com/apollographql/apollo3/compiler/introspection/Schema_to_in
...
-	public fun hashCode ()I
-	public fun toString ()Ljava/lang/String;
-}
-
-public abstract interface class com/apollographql/apollo3/compiler/ir/ModelGroupBuilder {
-	public abstract fun buildFragmentData (Ljava/lang/String;)Lkotlin/Pair;
-	public abstract fun buildFragmentInterface (Ljava/lang/String;)Lcom/apollographql/apollo3/compiler/ir/IrModelGroup;
-	public abstract fun buildOperationData (Ljava/util/List;Ljava/lang/String;Ljava/lang/String;)Lkotlin/Pair;
-}
-
```
martinbonnin, line 2022: 🙌

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkApolloDuplicateTargetNames.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.apollographql.apollo3.ast.GQLObjectTypeDefinition
+import com.apollographql.apollo3.ast.GQLScalarTypeDefinition
+import com.apollographql.apollo3.ast.GQLTypeDefinition
+import com.apollographql.apollo3.ast.GQLUnionTypeDefinition
+import com.apollographql.apollo3.ast.Issue
+import com.apollographql.apollo3.ast.Schema
+import com.apollographql.apollo3.ast.findTargetName
+
+@ApolloInternal
+fun checkApolloDuplicateTargetNames(schema: Schema): List<Issue> {
```
martinbonnin, line 19: Make it internal?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkApolloDuplicateTargetNames.kt
```diff
@@ -0,0 +1,58 @@
...
+  val issues = mutableListOf<Issue>()
+
+  val visitedTypes = mutableListOf<GQLTypeDefinition>()
+  for (type in schema
+      .typeDefinitions
+      .values
+  ) {
+    val targetName = type.directives.findTargetName(schema)
+    if (targetName == null) {
+      visitedTypes += type
```
martinbonnin, line 29: Somewhat a personal preference but I'm not a fan of the augmentedAssigment operator on mutable lists as it actually doesn't assign anything 🤷

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -17,13 +20,62 @@ import com.apollographql.apollo3.compiler.singularize
...
+
+  private fun computeClassNames(): Map<IrSchemaType, String> {
+    val schemaTypeToClassName = mutableMapOf<IrSchemaType, String>()
+    val allTypes: List<IrSchemaType> = ir.customScalars + ir.objects + ir.enums + ir.interfaces + ir.inputObjects + ir.unions
+    val usedNames = mutableSetOf<String>()
+    for (type in allTypes) {
+      val targetName = type.targetName
+      val className = if (targetName != null) {
+        // If a targetName was specified, honor it verbatim
+        targetName
```
martinbonnin, line 40: targetName certainly needs to be added to `usedNames` here

martinbonnin: Unless this is caught by validation earlier 🤔 ? In which case add a comment?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -17,13 +20,62 @@ import com.apollographql.apollo3.compiler.singularize
...
+      } else {
+        val uniqueName = uniqueName(type.name, usedNames)
+        // Capitalize all types except enums
```
martinbonnin, line 43: This smells. I can't really remember why such a decision was made 🤔 . But it feels asymmetrical...

My guess is that this predates the "case insensitive filesystem" issues and the assumption was to allow different enums like `Direction` and `direction` but the same is true about other schema types too

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -17,13 +20,62 @@ import com.apollographql.apollo3.compiler.singularize
...
-abstract class CodegenLayout(
+internal abstract class CodegenLayout(
+    private val ir: Ir,
     private val packageNameGenerator: PackageNameGenerator,
     private val schemaPackageName: String,
     private val useSemanticNaming: Boolean,
     private val useSchemaPackageNameForFragments: Boolean,
-    private val nameToClassName: (String) -> String,
 ) {
+  private val schemaTypeToClassName: Map<IrSchemaType, String> = computeClassNames()
```
martinbonnin, line 30: Do you need to key by `IrSchemaType` instead of just `name`? I'd prefer keying by `name` because it makes it explicit that only this is used. Also allows to mutate the `Ir` if we need to down the road while maintaining stable keys.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
martinbonnin: I _think_ the algorithm should be like so:

1. for all types without a `targetName`, assign a unique (possibly suffixed) name
2. for all types with a `targetName`, assert they are using amongst all types with a `targetName` + the names assigned in 1.

I believe the current algorithm fails in this case:

```graphql
type URL @targetName(newName: "Url2")
type Url 
type url
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
martinbonnin: Certainly something to change in 4.x

apollo-compiler/src/test/graphql/schema.sdl
```diff
@@ -311,8 +311,35 @@ type Query {
...
+
+input ReservedInput @targetName(name: "renamedInput") {
+  field: String
+}
+
+scalar ReservedScalar @targetName(name: "renamedScalar")
+
+enum ReservedEnum @targetName(name: "renamedEnum") {
+  VALUE
+}
```
martinbonnin, line 340: Mind adding this in a separate schema (in the target_name folder) so that it keeps this "fallback" schema reasonnably small (also avoids diffs in the `__schema` folder)?

apollo-compiler/src/test/validation/schema/duplicate-target-name.expected
```diff
@@ -0,0 +1,14 @@
...
+A type named 'RenamedEnum' is already defined, please use a different target name. First definition is: duplicate-target-name.graphql: (13, 1)
+------------
+ERROR: ReservedEnumValueName (26:0)
+A type named 'Type' is already defined, please use a different target name. First definition is: duplicate-target-name.graphql: (22, 1)
+------------
+ERROR: ReservedEnumValueName (34:0)
+A type named 'RenamedType' is already defined, please use a different target name. First definition is: duplicate-target-name.graphql: (30, 1)
+------------
+ERROR: ReservedEnumValueName (39:0)
+A type named 'Query' is already defined, please use a different target name. First definition is: duplicate-target-name.graphql: (1, 1)
```
martinbonnin, line 14: 👍

apollo-compiler/src/test/validation/schema/reserved-enum-value-names.graphql
```diff
@@ -9,3 +9,17 @@ enum Enum {
...
+
+enum Enum4 {
+  foo
+  bar @targetName(name: "foo")
+}
+
+enum Enum5 {
+  foo @targetName(name: "bar")
+  bar2 @targetName(name: "bar")
+}
```
martinbonnin, line 25: Can you add 

```graphql
enum Enum6 {
  type @targetName(name: "newType")
}
```

I think this fails right now (but shouldn't)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
BoD: Oh yeah, we can keep the name keys 👍

apollo-compiler/src/test/validation/schema/reserved-enum-value-names.graphql
BoD: I did, doesn't seem to fail but you must be onto something? 😅

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/ValidationTest.kt
```diff
@@ -45,6 +44,8 @@ class ValidationTest(name: String, private val graphQLFile: File) {
         val schemaResult = parseResult.valueAssertNoErrors().validateAsSchemaAndAddApolloDefinition()
         schemaResult.issues + if (graphQLFile.name == "reserved-enum-value-names.graphql") {
           checkApolloReservedEnumValueNames(schemaResult.value!!)
+        } else if (graphQLFile.name == "duplicate-target-name.graphql") {
+          checkApolloDuplicateTargetNames(schemaResult.value!!)
         } else {
```
martinbonnin, line 49: Sounds like the apollo-specific validation could apply for all schemas, irrespective of their name?

apollo-compiler/src/test/validation/schema/reserved-enum-value-names.graphql
martinbonnin: It's all good, sorry, got a bit confused by the different branches. May I suggest a (slightly) more compact way to present things:

```kotlin
    val enumNames = mutableSetOf<String>()
    for (value in enumDefinition.enumValues) {
      val targetName = value.directives.findTargetName(schema)

      val name = targetName ?: value.name
      when {
        name.isApolloReservedEnumValueName() -> {
          val message = if (targetName == null) {
            "'$name' is a reserved enum value name, please use the @targetName directive to specify a target name"
          } else {
            "'$name' is a reserved enum value name, please use a different name"
          }
          issues.add(
              Issue.ReservedEnumValueName(
                  message = message,
                  sourceLocation = value.sourceLocation
              )
          )
        }
        name in enumNames -> {
          issues.add(
              Issue.ReservedEnumValueName(
                  message = "'${targetName}' is already defined in this enum, please use a different name",
                  sourceLocation = value.sourceLocation
              )
          )
        }
        else -> {
          // all good
          enumNames.add(name)
        }
      }
    }
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
BoD: Good point. All right, assert it is!

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -53,6 +46,18 @@ internal abstract class CodegenLayout(
       usedNames.add(className.lowercase())
       this[type.name] = className
     }
+
+    // 2. Use targetName verbatim for types that define it
+    // Note: prior validation in checkApolloDuplicateTargetNames checks for clashes but there may still
+    // be a clash due to step 1.
```
martinbonnin, line 52: Consolidate both validations in a single method? Not only is it less code but it'll allow displaying a nice `SourceAwareException` with some context if we ever end up in this case?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -16,13 +19,48 @@ import com.apollographql.apollo3.compiler.singularize
...
+      usedNames.add(className.lowercase())
+      this[type.name] = className
+    }
+
+    // 2. Use targetName verbatim for types that define it
+    for (type in allTypes.filter { it.targetName != null }) {
+      val className = type.targetName!!
+      if (usedNames.contains(className.lowercase())) {
+        error("Apollo: '$className' cannot be used as a target name for '${type.name}' because it clashes with another class name")
+      }
```
martinbonnin, line 55: Do you need that check here now that it is done during validation?

martinbonnin (approved):

PR #4248, opened 2022-07-07
Minor copyedits to docs

StephenBarlow:
I had some unstructured time and figured I'd take a quick pass on Kotlin docs, I might do a bit more in the next week or two!

### DIFF
--- a/docs/source/essentials/mutations.mdx
+++ b/docs/source/essentials/mutations.mdx
@@ -84,7 +84,7 @@ mutation UpvotePost($postId: Int!) {
 
 ## Passing input objects
 
-The GraphQL type system includes [input types](http://graphql.org/learn/schema/#input-types) as a way to pass complex values to fields. Input types are often used for mutation variables, because they provide a compact way to pass in objects to be created:
+The GraphQL type system includes [input objects](http://graphql.org/learn/schema/#input-types) as a way to pass complex values to fields. Input objects are often used for mutation variables, because they provide a compact way to pass in objects to be created:
 
 ```graphql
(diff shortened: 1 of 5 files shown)


docs/source/essentials/mutations.mdx
```diff
@@ -84,7 +84,7 @@ mutation UpvotePost($postId: Int!) {
 
 ## Passing input objects
 
-The GraphQL type system includes [input objects](http://graphql.org/learn/schema/#input-types) as a way to pass complex values to fields. Input objects are often used for mutation variables, because they give you a compact way to pass in objects to be created:
+The GraphQL type system includes [input types](http://graphql.org/learn/schema/#input-types) as a way to pass complex values to fields. Input types are often used for mutation variables, because they provide a compact way to pass in objects to be created:
```
martinbonnin, line 87: This section focuses on "input objects" (complex values) so I'd keep the "input objects" wording there. 

```suggestion
The GraphQL type system includes [input objects](http://graphql.org/learn/schema/#input-types) as a way to pass complex values to fields. Input objects are often used for mutation variables, because they provide a compact way to pass in objects to be created:
```

I think objects were highlighted here because they are generated as separate classes compared to other types that are mapped directly to their built-in Kotlin equivalent (Boolean, Int, String, etc...)

martinbonnin (approved):

martinbonnin:
Thanks for looking at this 💜 . A couple comments but looks good to me!

StephenBarlow:
Thanks @martinbonnin! 🙏

PR #4254, opened 2022-07-08
Includes terminating brace in the JSON field entry

ArjanSM:
Very Trivial: The JSON entry for the type `favoriteBook` in the cache contents example doesn't have a closing brace (`}`).


martinbonnin (approved):
Thanks for catching this 🙏

PR #4257, opened 2022-07-11
🚧  remove withDates

martinbonnin:
Remove `withDates` parameter in SqlNormalizedCache. As there's more to come, we moved everything to `apollo-normalized-cache-sqlite-incubating`

This is a breaking change, sorry about this. Hopefully, it hasn't been there long enough that it's a big issue


BoD (approved):
👍

PR #4265, opened 2022-07-13
Pagination: add connectionFields argument to @typePolicy

BoD:
To simplify the Relay/Connection style pagination, a new `connectionsFields` argument is added to `@typePolicy`:

```graphql
extend type Query
@typePolicy(connectionFields: "sessions, rooms")

# Equivalent to:

extend type Query
@typePolicy(embeddedFields: "sessions, rooms")
@fieldPolicy(forField: "sessions", paginationArgs: "first, after, last, before")
@fieldPolicy(forField: "rooms", paginationArgs: "first, after, last, before")

extend type SessionConnection @typePolicy(embeddedFields: "edges")
extend type RoomConnection @typePolicy(embeddedFields: "edges")
```

This also generates a new `Pagination` object with a `connectionTypes: Set<String>` field, which can be passed to `ConnectionMetadataGenerator` (now available in `apollo-normalized-cache-api-incubating`, among with `ConnectionMetadataGenerator`).

The `connectionFields` argument must be opted-in by linking (importing) the `kotlin_labs` v0.2 directives like so:

```graphql
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.2", import: ["@typePolicy", "@fieldPolicy"])
```

See also [this PR](https://github.com/apollographql/specs/pull/24) that adds kotlin_labs v0.2 to the public specs site.

### DIFF
--- a/apollo-ast/api/apollo-ast.api
+++ b/apollo-ast/api/apollo-ast.api
@@ -982,6 +982,7 @@ public final class com/apollographql/apollo3/ast/GqldirectiveKt {
 
 public final class com/apollographql/apollo3/ast/GqldocumentKt {
 	public static final fun apolloDefinitions ()Ljava/util/List;
+	public static final fun apolloDefinitions (Ljava/lang/String;)Ljava/util/List;
 	public static final fun builtinDefinitions ()Ljava/util/List;
 	public static final fun linkDefinitions ()Ljava/util/List;
 	public static final fun withApolloDefinitions (Lcom/apollographql/apollo3/ast/GQLDocument;)Lcom/apollographql/apollo3/ast/GQLDocument;
--- a/apollo-ast/api/apollo-ast.api
+++ b/apollo-ast/api/apollo-ast.api
@@ -1121,6 +1121,7 @@ public final class com/apollographql/apollo3/ast/Schema {
 	public static final field REQUIRES_OPT_IN Ljava/lang/String;
 	public static final field TYPE_POLICY Ljava/lang/String;
 	public fun <init> (Ljava/util/List;)V
+	public final fun getConnectionTypes ()Ljava/util/Set;
 	public final fun getDirectiveDefinitions ()Ljava/util/Map;
 	public final fun getForeignNames ()Ljava/util/Map;
 	public final fun getMutationTypeDefinition ()Lcom/apollographql/apollo3/ast/GQLTypeDefinition;
--- a/apollo-ast/api/apollo-ast.api
+++ b/apollo-ast/api/apollo-ast.api
@@ -1121,7 +1121,6 @@ public final class com/apollographql/apollo3/ast/Schema {
 	public static final field REQUIRES_OPT_IN Ljava/lang/String;
 	public static final field TYPE_POLICY Ljava/lang/String;
 	public fun <init> (Ljava/util/List;)V
-	public final fun getConnectionTypes ()Ljava/util/Set;
 	public final fun getDirectiveDefinitions ()Ljava/util/Map;
 	public final fun getForeignNames ()Ljava/util/Map;
 	public final fun getMutationTypeDefinition ()Lcom/apollographql/apollo3/ast/GQLTypeDefinition;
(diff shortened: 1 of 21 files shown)


apollo-ast/api/apollo-ast.api
```diff
@@ -1121,6 +1121,7 @@ public final class com/apollographql/apollo3/ast/Schema {
 	public static final field REQUIRES_OPT_IN Ljava/lang/String;
 	public static final field TYPE_POLICY Ljava/lang/String;
 	public fun <init> (Ljava/util/List;)V
+	public final fun getConnectionTypes ()Ljava/util/Set;
```
martinbonnin, line 1124: can you make this `@ApolloInternal`? I'd like to make `Schema` (and apollo-ast) generic enough that it doesn't expose the Apollo specific parts. `foreignNames` is already an exception to this rule but it's not reason not to do it for this one while we can.

apollo-ast/src/main/resources/apollo.graphqls
```diff
@@ -22,7 +22,8 @@ directive @nonnull(fields: String! = "") on OBJECT | FIELD
 # `keyFields`: a selection set containing fields used to compute the cache key of an object. Order is important.
 # `embeddedFields`: a selection set containing fields that shouldn't create a new cache Record and should be
 # embedded in their parent instead. Order is unimportant.
-directive @typePolicy(keyFields: String! = "", embeddedFields: String! = "") on OBJECT | INTERFACE | UNION
+# `connectionFields`: a selection set containing fields that should be treated as Relay Connection fields. Order is unimportant.
+directive @typePolicy(keyFields: String! = "", embeddedFields: String! = "", connectionFields: String! = "") on OBJECT | INTERFACE | UNION
```
martinbonnin, line 26: Wondering if you could use `@requiresOptIn` on `connectionFields` 🥫?

martinbonnin (approved):
LGTM!

apollo-ast/src/main/resources/apollo.graphqls
BoD: We could do it as a purely 'indicative' thing? But it won't impact the generated code as it's on a directive argument definition

martinbonnin: indicative is better than nothing. + we could output a warning during codegen?

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gqldocument.kt
```diff
@@ -44,9 +44,10 @@ fun builtinDefinitions() = definitionsFromResources("builtins.graphqls")
 fun linkDefinitions() = definitionsFromResources("link.graphqls")
 
 /**
- * Extra apollo specific definitions from https://specs.apollo.dev/kotlin_labs/v0.1
+ * Extra apollo specific definitions from https://specs.apollo.dev/kotlin_labs/<[version]>
  */
-fun apolloDefinitions() = definitionsFromResources("apollo.graphqls")
+fun apolloDefinitions() = apolloDefinitions("v0.1")
```
martinbonnin, line 49: Make Deprecated?

martinbonnin: Probably make `apolloDefinitions(version: String)` public if we want to keep offering that possibility. I'd be surprised if anyone is relying on this at the moment but it might be useful for some use cases?

apollo-ast/src/main/resources/apollo-v0.1.graphqls
```diff
@@ -1,7 +1,7 @@
-# The kotlin_labs directives
-# You can import them with
+# The kotlin_labs v0.1 directives
+# You can import them with:
 #
-# extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", import: ["@optional", "@nonnull", "@typePolicy", "@fieldPolicy"])
+# extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.1", import: ["@optional", "@nonnull", "@typePolicy", "@fieldPolicy", "requiresOptIn", "targetName"])
```
martinbonnin, line 4: 👍

PR #4273, opened 2022-07-18
Pin GitHub actions

martinbonnin:
Turns out we had most of them pinned already except the first-party ones (starting with "actions/") and  `gradle/gradle-build-action`. I'm still pinning all of them because even if we obvisouly trust GitHub, they might still get hacked and reducing the attack surface could help if that were (unfortunately) to happen.


BoD (approved):
🔒

Issue #4100, opened 2022-05-12
Kotlin Native: remove the old memory manager restrictions


BoD:
The [new memory manager](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md) simplifies sharing code between targets by lifting the specific concurrency rules that have existed on Kotlin Native (more details in [this Kotlin blog post](https://blog.jetbrains.com/kotlin/2021/05/kotlin-native-memory-management-update/)).

Switching to the new MM will mean:
- removing `freeze()` calls and some object duplications
- removing the need to call `runTest`
- removing the requirement to execute calls on the main thread on Apple

Users of the library will need to switch their project to the new MM as well.

Let's use this issue to track progress on this change, but also to gather user feedback on potential blockers preventing such a switch.

martinbonnin:
Can we rename this as "Kotlin Native: remove the old memory manager restrictions". As technically we don't have to switch to the new memory manager (except for tests)

BoD:
[Looks like](https://github.com/cashapp/sqldelight/blob/master/gradle.properties#L26) SQLDelight is planned to make the switch in v2.0.0

Tieru:
I still hope you will allow us using ApolloGraphql with the new memory model a bit sooner. Currently I'm struggling with the old MM, and new one would make my life 100 times easier, but I have choose between "not working part of my app" and "entire app is not working due to asserts in apollo"

BoD:
Hi @Tieru! We are under the impression that the library should already work with projects using the new MM. A sample project is [here](https://github.com/BoD/apollo-kotlin-template-mpp) but maybe it is too simplistic and we are missing something? Can you tell us what is not working with your project?

Tieru:
Sorry for delay, but well, it does work!
My problem was that I was starting requests on background thread. Somehow it worked this way earlier on old MM, before I switched to the new MM.

Thanks a lot for your work :)

BoD:
Done with #4287 and is part of version 3.5.0

PR #4287, opened 2022-07-22
Native: assume New Memory Manager

BoD:
Assume running under the [New Memory Manager (aka New Memory Model)](https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/).

Related to #4100

- Replace custom dispatcher implementations using the main thread on Apple, with `newFixedThreadPoolContext` / `newSingleThreadContext`
- Remove custom code to dispatch callbacks on main thread and checks to be called from the main thread
- Use a Mutex instead of a single thread queue in Guard
- Remove all calls to `freeze` and related
- Also mark `runTest` as deprecated


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
```diff
@@ -145,8 +145,8 @@ private constructor(
     }
 
     /**
-     * An immutable body that can be freezed when used from Kotlin native.
-     * Prefer [bodySource] on non-native so that the response can be streamed.
+     * An immutable body.
+     * Prefer [bodySource] so that the response can be streamed.
      */
     fun body(bodyString: ByteString) = apply {
```
martinbonnin, line 151: Make `@Deprecated` ?

apollo-mpp-utils/api/apollo-mpp-utils.api
```diff
@@ -7,14 +13,10 @@ public final class com/apollographql/apollo3/mpp/Platform : java/lang/Enum {
...
 
 public final class com/apollographql/apollo3/mpp/UtilsKt {
-	public static final fun assertMainThreadOnNative ()V
 	public static final fun currentThreadId ()Ljava/lang/String;
 	public static final fun currentThreadName ()Ljava/lang/String;
 	public static final fun currentTimeFormatted ()Ljava/lang/String;
 	public static final fun currentTimeMillis ()J
-	public static final fun ensureNeverFrozen (Ljava/lang/Object;)V
-	public static final fun freeze (Ljava/lang/Object;)V
-	public static final fun isFrozen (Ljava/lang/Object;)Z
```
martinbonnin, line 17: Sounds like a breaking change. Mark them deprecated with `Error` level instead?

apollo-normalized-cache-incubating/src/appleMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/Guard.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  actual fun writeAndForget(block: (R) -> Unit) = runBlocking {
+    mutex.withLock {
+      block(resource)
+    }
+  }
+
+  actual fun dispose() {
+  }
+}
```
martinbonnin, line 31: Looks like all the `Guard` implementations are the same now? Can you simplify everything and put the mutex in `DefaultApolloStore` ?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
```diff
@@ -1,79 +1,29 @@
...
-  }
+private val defaultDispatcher = newFixedThreadPoolContext(nThreads = 32, name = "Apollo Default Dispatcher")
 
-  return DefaultDispatcher
+internal actual fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher {
+  return requested ?: defaultDispatcher
 }
 
-
 internal actual class BackgroundDispatcher actual constructor() {
```
martinbonnin, line 14: Can you rename this to CloseableDispatcher and make it implement `Closeable`?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
```diff
@@ -1,79 +1,29 @@
...
-    "Changing the dispatcher is not supported on Apple targets"
-  }
-  check(NSThread.isMainThread) {
-    "defaultDispatcher must be called from the main thread"
-  }
+private val defaultDispatcher = newFixedThreadPoolContext(nThreads = 32, name = "Apollo Default Dispatcher")
 
-  return DefaultDispatcher
+internal actual fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher {
+  return requested ?: defaultDispatcher
```
martinbonnin, line 11: It's a bit weird to have the `requested` parameter now. Can you move the logic at the callsite?

```
val actualDispatcher = dispatcher ?: defaultDispatcher()
```

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/NSURLSessionHttpEngine.kt
```diff
@@ -41,15 +39,11 @@ actual class DefaultHttpEngine constructor(
   actual constructor(timeoutMillis: Long) : this(timeoutMillis, DefaultDataTaskFactory())
 
   @Suppress("UNCHECKED_CAST")
-  override suspend fun execute(request: HttpRequest) = suspendAndResumeOnMain<HttpResponse> { mainContinuation, invokeOnCancellation ->
-    assertMainThreadOnNative()
-
-    request.freeze()
-
+  override suspend fun execute(request: HttpRequest): HttpResponse = suspendCancellableCoroutine { continuation ->
```
martinbonnin, line 42: Woohooo 🚀 🚀 🚀 Long live the `suspendAndResumeOnMain`

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
martinbonnin: Also the KDoc for the `expect` declaration needs a bit of tuning now

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/multipart.kt
```diff
@@ -26,7 +25,7 @@ internal fun multipartBodyFlow(response: HttpResponse): Flow<BufferedSource> {
     )
     while (true) {
       // Read the body in a background thread because it is blocking
```
martinbonnin, line 27:
```suggestion
```

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -37,8 +33,6 @@ private constructor(
     val interceptors: List<HttpInterceptor>,
     val exposeErrorBody: Boolean,
 ) : NetworkTransport {
```
martinbonnin, line 35: There's now a bunch of unrequired `suspend` modifiers in this file

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
```diff
@@ -1,79 +1,29 @@
...
+import kotlinx.coroutines.newSingleThreadContext
 
-internal actual fun defaultDispatcher(requested: CoroutineDispatcher?): CoroutineDispatcher {
-  check(requested == null || requested is DefaultDispatcher) {
-    "Changing the dispatcher is not supported on Apple targets"
-  }
-  check(NSThread.isMainThread) {
-    "defaultDispatcher must be called from the main thread"
-  }
+private val defaultDispatcher = newFixedThreadPoolContext(nThreads = 32, name = "Apollo Default Dispatcher")
```
martinbonnin, line 8: Could we use `Dispatchers.Default` here? This way it pools the [`global queue`](https://developer.apple.com/documentation/dispatch/1452927-dispatch_get_global_queue) and saves a bunch of memory/threads?

martinbonnin (changes requested):
Very cool 🚀 . That's a very welcome change! I think we can even simplify a bit more. Let me know what you think!

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
BoD: Looks like it! 

(Can't remember where I (mis?)read that Dispatchers.Default and Main were the same on Apple  - nope that's fake news ❌)

apollo-mpp-utils/api/apollo-mpp-utils.api
BoD: The change is actually due to removing the `actual` implementations - I've added them back but kept `HIDDEN` deprecation level, LMK if that's 👍 with you.

apollo-normalized-cache-incubating/src/appleMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/Guard.kt
BoD: Did this in 1de5b7592c6b805ce4742dfd49f713f1e9cdb4e3.

On the JVM, we used to use a `ReentrantReadWriteLock`, which has a different behavior for read and write - I'm wondering if we could lose some performance by using a simple Mutex instead.

apollo-mpp-utils/src/appleMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -33,19 +29,13 @@ actual fun currentThreadName(): String {
...
-
+actual fun isFrozen(obj: Any) = false
 actual fun freeze(obj: Any) {
-  obj.freeze()
 }
 
 actual fun assertMainThreadOnNative() {
-  check(NSThread.isMainThread()) {
-    "Non-main native call"
-  }
```
martinbonnin, line 48: Something else I forgot to mention: can we have an assert and graceful error message for users that don't use the new MM?

apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/dispatchers.kt
BoD: Ah found it, it was on the [doc for the Main dispatcher](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html) saying _"On JS and Native it is equivalent to the Default dispatcher"_

But the doc must be outdated for the new MM.


```kotlin
        withContext(Dispatchers.Main) {
            println(NSThread.isMainThread)
        }
        withContext(Dispatchers.Default) {
            println(NSThread.isMainThread)
        }
```
prints
- `true`, `true` with the old MM
- `true`, `false` with the new MM

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/Http.kt
```diff
@@ -145,9 +148,11 @@ private constructor(
     }
 
     /**
-     * An immutable body that can be freezed when used from Kotlin native.
-     * Prefer [bodySource] on non-native so that the response can be streamed.
+     * An immutable body.
+     * Prefer [bodySource] so that the response can be streamed.
      */
+    @Deprecated("Use body(BufferedSource) instead", ReplaceWith("body(bodySource)"))
```
martinbonnin, line 154:
```suggestion
    @Deprecated("Use body(BufferedSource) instead", ReplaceWith("Buffer().write(bodyString)", "okio.Buffer"))
```

apollo-mpp-utils/api/apollo-mpp-utils.api
martinbonnin: I like `ERROR` better. It gives a better upgrade path for users. Instead of getting:

```
Cannot find symbol 'ensureNeverFrozen'
```

They will get:

```
With the new Memory Manager this method is no longer needed
```

This leads me to realize that keeping binary compatibility here is a gimmick because any lib that would call that would most likely break under the new memory model. But it shouldn't harm either to keep these symbols around

apollo-normalized-cache-incubating/src/appleMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/Guard.kt
martinbonnin: LGTM. Still needs to be done for `-incubating` artifacts

apollo-mockserver/gradle.properties
```diff
@@ -1,3 +1,5 @@
 POM_ARTIFACT_ID=apollo-mockserver
 POM_NAME=Apollo Mockserver
 POM_DESCRIPTION=Apollo Mockserver
+# See https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md
+kotlin.native.binary.memoryModel=experimental
```
martinbonnin, line 5: I'd like to use KGP APIs instead of Gradle properties because:
1. It's easier to find usage and remove them once they become obsolete
2. It allows more fined grained control (see also https://youtrack.jetbrains.com/issue/KT-50552/Gradle-plugin-provide-extension-APIs-instead-of-Gradle-properties)

Here in particular (and in multiple other places), feels like we only need it for the tests. Could we enable it only for the test targets?

tests/coroutines-mt/src/appleTest/kotlin/macos/app/MainTest.kt
```diff
@@ -40,25 +38,13 @@ class MainTest {
...
-    val server = MockServer()
-    server.enqueue(json)
-    val client = ApolloClient.Builder().serverUrl(server.url()).build().freeze()
-    withContext(Dispatchers.Default) {
-      assertFailsWith(IllegalStateException::class) {
-        client.query(GetRandomQuery()).execute()
-      }
-    }
-  }
-
```
martinbonnin, line 54: 👍

tests/coroutines-mt/src/appleTest/kotlin/macos/app/MainTest.kt
```diff
@@ -26,7 +24,7 @@ class MainTest {
   fun coroutinesMtCanWork() = runTest {
     withContext(Dispatchers.Default) {
       println("Dispatchers.Default: ${currentThreadId()}")
-      withContext(Dispatchers.Main) {
+      withContext(Dispatchers.Main.immediate) {
```
martinbonnin, line 27: I understand `immediate` is required to avoid a deadlock but that feels quite unrepresentative of a real use case.

Especially since we're planning to use `kotlin.runTest {}`, the default dispatcher [should be something like Dispatcher.Main](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-standard-test-dispatcher.html). In which case something like

```kotlin
  fun coroutinesMtCanWork() = runTest {
    withContext(Dispatchers.Default) {
      // execute GraphQL query here
```

should be enough?

martinbonnin (changes requested):
See comments above

apollo-mpp-utils/api/apollo-mpp-utils.api
BoD: Agreed!

martinbonnin (approved):
👏

BoD:
@martinbonnin I've just applied minor tweaks to Threading.md to remove / update obviously no longer correct or relevant bits. but it may benefit from bigger revamp :) Don't hesitate to make more changes if you feel inspired!

martinbonnin (approved):

PR #4292, opened 2022-07-26
Use internal runTest in all tests

BoD:
- Adds a new internal version of `runTest` that uses `kotlinx.coroutines.test.runTest` under the hood
- Use it in all tests to avoid using the public one which doesn't use `kotlinx.coroutines.test.runTest` and is deprecated
- Except for a few tests that need particular delays / timing -  for those, we use `runTestBlocking` introduced in this PR, also internal, which does the same as the public `runTest` (calls `runBlocking`)


apollo-testing-support/api/apollo-testing-support.api
```diff
@@ -18,11 +18,17 @@ public final class com/apollographql/apollo3/testing/ReadFileJvmKt {
...
+public final class com/apollographql/apollo3/testing/RunTestKt {
 }
 
 public final class com/apollographql/apollo3/testing/TestNetworkTransportKt {
 }
 
+public final class com/apollographql/apollo3/testing/internal/RunTestBlockingKt {
+	public static final fun runTestBlocking (Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;)V
+	public static synthetic fun runTestBlocking$default (Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
+}
```
martinbonnin, line 30: I believe this is a false positive from the binary compatibility tracker as they are marked `@ApolloInternal`

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/internal/runTest.kt
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Similar to [runTest] but uses `runBlocking` instead of [kotlinx.coroutines.test.runTest].
+ * Use it when delays/timing are important to the test or [kotlinx.coroutines.test.runTest] can't otherwise be used.
+ */
+@ApolloInternal
+expect fun runTestBlocking(
+    context: CoroutineContext = EmptyCoroutineContext,
+    before: suspend CoroutineScope.() -> Unit = {},
+    after: suspend CoroutineScope.() -> Unit = {},
+    block: suspend CoroutineScope.() -> Unit,
```
martinbonnin, line 39: Can we merge both? 

```kotlin
@ApolloInternal
@OptIn(ExperimentalCoroutinesApi::class)
fun runTest(
    skipDelays: Boolean = false, 
    before: suspend CoroutineScope.() -> Unit = {},
    after: suspend CoroutineScope.() -> Unit = {},
    block: suspend CoroutineScope.() -> Unit,
) {
  // I think using Dispatchers.Main will override the default `TestDispatcher` (but not sure)
  val context = if (skipDelays) EmptyCoroutinesContext else Dispatchers.Main
  kotlinx.coroutines.test.runTest(context) {
    before()
    try {
      block()
    } finally {
      after()
    }
  }
}
```

This way we can get rid of our own `Promise` and expect/actual stuff?

tests/test-network-transport/src/test/kotlin/testnetworktransport/QueueTestNetworkTransportHandlerTest.kt
```diff
@@ -7,29 +7,34 @@ import com.apollographql.apollo3.api.Error
...
 
+@OptIn(ExperimentalCoroutinesApi::class)
 class QueueTestNetworkTransportHandlerTest {
   private lateinit var apolloClient: ApolloClient
 
-  private fun setUp() {
+  private fun TestScope.setUp() {
     apolloClient = ApolloClient.Builder()
         .networkTransport(QueueTestNetworkTransport())
+        .dispatcher(StandardTestDispatcher(testScheduler))
```
martinbonnin, line 28: Why do we need to change the dispatcher here?

martinbonnin (commented):
Mostly questions at this point. I'm not sold on the idea that skipping delays by default is a good thing for our tests as a lot of them still rely on delays Some of them we could maybe rewrite but some others maybe not (watchers emitting several values are typically never idle). And in some cases, delays is part of the tested behaviour (websocket keep alive)

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/internal/runTest.kt
martinbonnin: Damn, too bad. `expect`/`actual` it is then. Nit, I'd rename `runTestBlocking` to something more explicit like `runBlockingOrPromise`. It's not pretty but it's internal only. And JS is not blocking at all so it's more accurate.

BoD: Agreed!

tests/test-network-transport/src/test/kotlin/testnetworktransport/QueueTestNetworkTransportHandlerTest.kt
martinbonnin: So it was not needed after all?

BoD: Sorry for not replying earlier! In the previous version which was using `kotlinx.coroutines.test.runTest` this was to make the client use the `testScheduler` - without it, turbine's awaitXxx methods time out but I must admit I'm not sure why!

Now we've reverted to using `runBlocking` under the hood, so need for this.

martinbonnin (approved):

PR #4295, opened 2022-07-27
Automatically escape `type` in enum values

martinbonnin:
Since we can detect, we might as well to it on behalf of the user, this saves some configuration. See https://github.com/apollographql/apollo-kotlin/issues/4293

Note that if someone has already done something like this:

```
extend enum SomeEnum {
  type @targetName(name: "type_")
}
```

they will have to remove this to keep `type_` or else it'll be renamed to `type__`. 

Also remove a workaround we had for 'header' and 'impl' that was adressed in KotlinPoet with https://github.com/square/kotlinpoet/pull/1248


BoD (approved):
👍

PR #4296, opened 2022-07-27
Update v2 docs now that v3 is available

remcomokveld:
Fairly trivial but since the release/2.x branch is kept for reference it's nice to have it be accurate about the status of v3.


martinbonnin (approved):
Good catch! Thanks a lot! 🙏

PR #4303, opened 2022-07-29
Remove schema and AST from the IR

martinbonnin:
Prep-work for schemaBased builders, cleanup the IR by removing the schema and AST types from there. This allows to factor some of the logic between Java/Kotlin codegen and more generally, it's better separation of concerns


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
```diff
@@ -201,6 +201,7 @@ internal class DefaultApolloStore(
           cacheKeyGenerator = objectIdGenerator
       )
 
+      records.flatMap { it.value.fieldKeys() }
```
BoD, line 204: Not intentional?

BoD (approved):
⚙️

apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
martinbonnin: uh oh good catch 👍

martinbonnin:
As it's a relatively big PR, I'm keeping this one open until we release 3.5.0

PR #4308, opened 2022-08-01
Add a native-benchmarks project

BoD:


martinbonnin (approved):
LGTM 🚀 As discussed, commiting machine dependent timings will most break the build cache (once we have it 😅 ) but it's good to have in until we merge the new MM and can see the difference

PR #4310, opened 2022-08-02
Add native benchmarks and upload results to DataDog

BoD:


martinbonnin (approved):
LGTM
You could also save the serialization by make a native executable that uploads to datadog directly from Kotlin native using HttpEngine 🤓

Issue #4080, opened 2022-05-05
Update to Kotlin 1.7


BoD:
Kotlin 1.7 Beta [has been released](https://blog.jetbrains.com/kotlin/2022/05/kotlin-1-7-0-beta/), let's try it on the project and see if we encounter any issue with it.

BoD:
Initial findings:
- the workaround to pick an older version of Kotlin when inside the IDE for [this issue](https://youtrack.jetbrains.com/issue/KTIJ-21583/HMPP:-1.6.20-breaks-autocomplete-in-multiplatform-composite-buil#focus=Comments-27-6022244.0-0)) doesn't work. Issue open [here](https://youtrack.jetbrains.com/issue/KT-52375/With-1-7-0-Beta-can-t-ise-Gradle-s-useModule)
- needed to update `org.jetbrains.kotlinx:kotlinx-datetime` to latest version to avoid a native compiler crash
- other than that, all tests pass 👍

martinbonnin:
Update the name of this issue to "Update to Kotlin 1.7" ? 

That being said, I don't really see us adopting it in the near future. [Using 1.7 features will break compiling with 1.6](https://kotlinlang.org/docs/kotlin-evolution.html#compatibility-tools), which means unfortunately, we'll have to wait until a significant portion of the user base has switched to Kotlin 1.7 before enabling it.

martinbonnin:
Note: we could still have a [dedicated shadow branch](https://slack.engineering/shadow-jobs/) to test 1.7 on CI and make sure it doesn't break.

BoD:
WIP branch [here](https://github.com/apollographql/apollo-kotlin/compare/kotlin-1-7?expand=1).

BoD:
Done.

PR #4314, opened 2022-08-03
Update Kotlin dependency to 1.7.10

BoD:
Also remove the Kotlin 1.7 shadow build.

Related to #4080.


README.md
```diff
@@ -215,7 +215,7 @@ Some platforms have specific requirements:
 For building, it requires:
 
 * Gradle 5.6
-* Kotlin 1.5+ (1.6+ for native)
+* Kotlin 1.6+ (1.7+ for native)
```
BoD, line 218: Actually, 1.5+ works for building, but needs `-Xskip-metadata-version-check` compiler flag, otherwise it fails with errors like `org.jetbrains.kotlin/kotlin-stdlib-common/1.7.10/bac80c520d0a9e3f3673bc2658c6ed02ef45a76a/kotlin-stdlib-common-1.7.10.jar!/META-INF/kotlin-stdlib-common.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.`

BoD: My understanding is that this ensures Apollo Kotlin doesn't use any >1.5 language feature, but the error above is about trying to use the Kotlin stdlib 1.7 (a dependency of Apollo Kotlin), on a project using 1.5. Now I'm not sure if removing the warning with the flag is really safe. It worked on my little sample project 😅

Instead of this flag, another way is to pin the stdlib to 1.5 - and Apollo Kotlin will work thanks to being built with `-api-version 1.5`. But that's probably dangerous because of other dependencies that may not use this flag...

martinbonnin: > the error above is about trying to use the Kotlin stdlib 1.7 (a dependency of Apollo Kotlin), on a project using 1.5.

Aah yes of course! Thanks for the reminder. Damn this stuff is h ard!

> Instead of this flag, another way is to pin the stdlib to 1.5 - and Apollo Kotlin will work thanks to being built with -api-version 1.5. But that's probably dangerous because of other dependencies that may not use this flag...

No need to pin, just set

```
kotlin.stdlib.default.dependency=false
```

And we will use the "biggest" kotlin-stdlib version. `apollo-compiler` should use 1.7 because of KotlinPoet. `apollo-gradle-plugin` should use nothing because it's relocated. Everything else will keep using 1.6 until we start using a dependency that requires 1.7.

martinbonnin: PS: I would expect KGP to add the stdlib that's matching the `apiVersion` by default actually 🤔 . I wonder if this should be considered a bug.

BoD: > kotlin.stdlib.default.dependency=false

Looks like it's working! I had trouble with JS but explicitly adding the `stdlib-js` dependency fixed it (I think it shouldn't be necessary, so it looks like a bug). Using `1.6.21` (most recent version compatible with 1.5).

> I would expect KGP to add the stdlib that's matching the apiVersion by default actually 🤔 . I wonder if this should be considered a bug.

Agreed ✅ Filed as [KT-53462](https://youtrack.jetbrains.com/issue/KT-53462).

apollo-annotations/build.gradle.kts
```diff
@@ -7,7 +7,17 @@ configureMppDefaults()
...
   sourceSets {
     val commonMain by getting {
+      dependencies {
+        api(kotlin("stdlib", groovy.util.Eval.x(project, "x.versions.kotlinStdlib").toString()))
+      }
     }
+
+    val jsMain by getting {
+      dependencies {
+        api(kotlin("stdlib-js", groovy.util.Eval.x(project, "x.versions.kotlinStdlib").toString()))
```
martinbonnin, line 17: How come Gradle metadata doesn't pick this up?

apollo-gradle-plugin-external/gradle.properties
```diff
@@ -2,3 +2,4 @@ POM_ARTIFACT_ID=apollo-gradle-plugin-external
 POM_NAME=Apollo Gradle Plugin (not shadowed)
 POM_DESCRIPTION=Gradle plugin for generating java/kotlin classes for graphql files
 
+kotlin.stdlib.default.dependency=false
```
martinbonnin, line 5: Could that be set once in the rootProject? I think Gradle properties have a notion of "hierarchy"

apollo-gradle-plugin/build.gradle.kts
```diff
@@ -41,7 +41,7 @@ if (relocateJar) {
 
       exclude(".*MANIFEST.MF")
       exclude("META-INF/versions/9/module-info\\.class")
-      exclude("META-INF/kotlin-stdlib.*\\.kotlin_module")
+//      exclude("META-INF/kotlin-stdlib.*\\.kotlin_module")
```
martinbonnin, line 44: Can you ellaborate on that change?

build-logic/src/main/kotlin/CompilerOptions.kt
```diff
@@ -24,8 +24,8 @@ fun Project.configureJavaAndKotlinCompilers() {
           // See also https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/
           freeCompilerArgs = freeCompilerArgs + "-Xjvm-default=compatibility"
         }
-        apiVersion = "1.5"
-        languageVersion = "1.5"
+        apiVersion = groovy.util.Eval.x(project, "x.versions.kotlinCompatVersion").toString()
+        languageVersion = groovy.util.Eval.x(project, "x.versions.kotlinCompatVersion").toString()
```
martinbonnin, line 28: 👍

apollo-gradle-plugin/build.gradle.kts
BoD: Woops committed by mistake 😅

apollo-gradle-plugin-external/gradle.properties
BoD: You're right! TIL

apollo-annotations/build.gradle.kts
BoD: I don't know! But without it JS compilation complains - looks like a bug.

apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/KotlinPluginVersionTests.kt
```diff
@@ -17,12 +17,7 @@ class KotlinPluginVersionTests {
    */
   @Test
   fun `kotlin n-1 succeeds`() {
-    val currentKotlinVersion = System.getenv("COM_APOLLOGRAPHQL_VERSION_KOTLIN_PLUGIN") ?: "1.6"
-    val kotlinPluginVersionToTest = if (currentKotlinVersion.startsWith("1.7")) {
-      "1.6.0"
-    } else {
-      "1.5.31"
-    }
+    val kotlinPluginVersionToTest = "1.6.0"
```
martinbonnin, line 20: Should this test be `kotlin JVM min version succeeds` and use 1.5? Since this is what we claim 1.5+ support?

BoD: Good point! ✅

martinbonnin (approved):

PR #4316, opened 2022-08-03
[release-2.x] fix converting introspection to SDL

martinbonnin:
The 2.x code was trying to re-encode something that was already GraphQL resulting in extra quotes


BoD (approved):
👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/parser/introspection/introspection2sdl.kt
```diff
@@ -5,13 +5,24 @@ import okio.buffer
...
+ *
+ * This is wrong. The spec mandates defaultValue to be GraphQL encoded String:
+ * "defaultValue": "[0, 1]"
+ *
+ * For backward compatibility, use [legacyDefaultValues] = true
+ */
+fun IntrospectionSchema.toSDL(file: File, legacyDefaultValues: Boolean) {
   file.sink().buffer().use {
-    toSDL(it)
+    toSDL(it, legacyDefaultValues)
```
martinbonnin, line 19: This is not great but at this point, I'm not sure how many wrongly encoded json are in the wild so I went for the tactical approach of making the least invasive change possible

BoD (approved):

PR #4319, opened 2022-08-04
Update Threading.md and change the dispatcher earlier in the chain

martinbonnin:


apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -216,10 +218,11 @@ val CacheAndNetworkInterceptor = object : ApolloInterceptor {
...
   override fun <D : Operation.Data> intercept(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
-    if (request.operation !is Query) {
+    return if (request.operation !is Query) {
       // Subscriptions and Mutations do not support fetchPolicies
-      return chain.proceed(request)
+      chain.proceed(request)
+    } else {
+      request.fetchPolicyInterceptor.intercept(request, chain)
     }
-    return request.fetchPolicyInterceptor.intercept(request, chain)
```
martinbonnin, line 223: That's cosmetics only

tests/simple-mpp/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+        implementation("com.apollographql.apollo3:apollo-mockserver")
+        implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4")
+      }
+    }
+  }
+}
+
+apollo {
+  packageName.set("com.example")
+}
```
martinbonnin, line 28: That's a new simple mpp integration test. I used it to do some quick tests. More generally, it's handy to reproduce issues or just as a copy/paste template for other tests

BoD (approved):
👍

martinbonnin:
Moving this to draft as it has a side effect of subscribing to the store asynchronously now and therefore is subject to race conditions. This is something we've seen in the past and would need to be revisited.

martinbonnin:
Marking this 2 years old PR as "ready for review". There are still CI flakes that I believe are not related to this PR and will address separately

BoD (approved):
🎉

PR #4321, opened 2022-08-05
First drop for data builders

martinbonnin:
Simple use case is in tests:

```kotlin
    val data = GetStuffQuery.Data {
      nullableInt = null
      nonNullableInt = 42

      this["alias"] = 50
    }

    assertEquals(null, data.nullableInt)
    assertEquals(42, data.nonNullableInt)
    assertEquals(50, data.alias)
```

Still needed:
* custom scalars support
* java codegen
* default values
* strict mode
* use in our own tests
* (maybe) cleanup the `IrType2` stuff


BoD (approved):
Looks good! 🚀

PR #4323, opened 2022-08-08
[Data Builders] add support for custom scalars

martinbonnin:
Custom scalars are a bit weird because:
* Our adapters know only about Json so theorically only strings, boolean, numbers, objects and arrays
* But Kotlin callsites want to specify them as Kotlin classes like `Date`, `GeoPoint`, etc...

This PR does a few tricks to make this work:
1. introduce a new "ANY" token in the `JsonReader` API. This allows a `MapJsonReader` to read any value from Map as-is though `UnsafeAdapter`
2. Use `UnsafeAdapter` always for custom scalars whose adapter is registered at runtime. This saves the need to pass an instance of `CustomScalarAdapters` around
3. For custom scalars whose adapter is registered at compile time, the value is serialized before being put in the `__fields` map so that the adapters can read it

It's not super pretty but it works. And I can't really see a better way without adding more abstraction layers that will certainly add a small performance penalty in the nominal case.

See https://github.com/apollographql/apollo-kotlin/issues/4016

### DIFF
--- a/apollo-api/api/apollo-api.api
+++ b/apollo-api/api/apollo-api.api
@@ -245,6 +245,13 @@ public final class com/apollographql/apollo3/api/BooleanExpressions {
 	public static final fun variable (Ljava/lang/String;)Lcom/apollographql/apollo3/api/BooleanExpression;
 }
 
+public final class com/apollographql/apollo3/api/BuilderProperty {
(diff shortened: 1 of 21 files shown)


BoD (approved):
Looks good! 👍

PR #4324, opened 2022-08-08
Version catalog

BoD:
Replace our (beloved?) `dependencies.gradle` with the more modern [version catalog](https://docs.gradle.org/current/userguide/platforms.html).


apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/SourceDirectorySetTests.kt
```diff
@@ -1,13 +0,0 @@
...
-import com.apollographql.apollo3.gradle.util.TestUtils
-import com.apollographql.apollo3.gradle.util.generatedChild
-import org.gradle.testkit.runner.TaskOutcome
-import org.junit.Assert
-import org.junit.Test
-import java.io.File
-
-class SourceDirectorySetTests {
-
-}
```
martinbonnin, line 13: 🗡️ 👍

apollo-gradle-plugin/testProjects/groovy/build.gradle
```diff
@@ -1,16 +0,0 @@
...
-  mavenCentral()
-}
-
-apollo {
-  // Make sure Action<> can be called from Groovy
-  service("main") {
-    // Make sure groovy can find the no-arg overload
-    packageNamesFromFilePaths()
-  }
-}
```
martinbonnin, line 16: Why removing that tests? I think a significant part of the user base is using Groovy and our test coverage there isn't so great.

apollo-gradle-plugin/testProjects/mismatchedVersions/settings.gradle.kts
```diff
@@ -1,2 +1,10 @@
...
\ No newline at end of file
+rootProject.name = "mismatchedVersions"
+
+dependencyResolutionManagement {
+  versionCatalogs {
+    create("libs") {
+      from(files("../../../gradle/libs.versions.toml"))
+    }
+  }
+}
```
martinbonnin, line 10: A bit unfortunate that we have to copy paste this block all over the place now but there's no silver bullet 🤷

gradle/libs.versions.toml
```diff
@@ -0,0 +1,142 @@
...
+apollo-plugin = { group = "com.apollographql.apollo3", name = "apollo-gradle-plugin", version.ref = "apollo" }
+apollo-runtime = { group = "com.apollographql.apollo3", name = "apollo-runtime", version.ref = "apollo" }
+apollo-rx2 = { group = "com.apollographql.apollo3", name = "apollo-rx2-support", version.ref = "apollo" }
+apollo-mockserver = { group = "com.apollographql.apollo3", name = "apollo-mockserver", version.ref = "apollo" }
+apollo-mppUtils = { group = "com.apollographql.apollo3", name = "apollo-mpp-utils", version.ref = "apollo" }
+apollo-normalizedCache = { group = "com.apollographql.apollo3", name = "apollo-normalized-cache", version.ref = "apollo" }
+apollo-normalizedCache-incubating = { group = "com.apollographql.apollo3", name = "apollo-normalized-cache-incubating", version.ref = "apollo" }
+apollo-normalizedCache-sqlite = { group = "com.apollographql.apollo3", name = "apollo-normalized-cache-sqlite", version.ref = "apollo" }
+apollo-normalizedCache-sqlite-incubating = { group = "com.apollographql.apollo3", name = "apollo-normalized-cache-sqlite-incubating", version.ref = "apollo" }
```
martinbonnin, line 60: Nit: What do you say about we stick to kebab-case every where instead of a mix here? Doesn't have to be done now. but as we move forward?

tests/jvmoverloads/build.gradle.kts
```diff
@@ -1,12 +1,12 @@
 plugins {
-  id("org.jetbrains.kotlin.jvm")
+  kotlin("jvm")
```
martinbonnin, line 2: :old_man_yells_at_gradle: I personally think this is too much. I like the consistency of `id("")` over the (maybe?) type safety of `java` or `kotlin("jvm")`. The `kotlin("jvm")` especially is weird because it behaves differently in the plugins or dependencies block 🤷 

I am definitely not dying on this one hill though :).

martinbonnin (commented):
CI complains because there are still a bunch of references to dependencies.gradle but looks cool 👍

apollo-gradle-plugin/testProjects/groovy/build.gradle
BoD: I had the impression that it was not used, because it's not referenced by any test, but maybe I'm missing something?

apollo-testing-support/build.gradle.kts
```diff
@@ -11,30 +11,30 @@ kotlin {
         api(projects.apolloApi)
         api(projects.apolloRuntime)
         api(projects.apolloMockserver)
-        api(groovy.util.Eval.x(project, "x.dep.kotlinCoroutines"))
-        implementation(groovy.util.Eval.x(project, "x.dep.atomicfu").toString()) {
+        api(libs.kotlinx.coroutines)
+        implementation(libs.atomicfu.get().toString()) {
           because("We need locks in TestNetworkTransportHandler (we don't use the gradle plugin rewrite)")
```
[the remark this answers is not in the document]
BoD: Yeah exactly, a bit annoying, but the available signatures are:
```kotlin
fun implementation(dependencyNotation: Any): Dependency?
fun implementation(dependencyNotation: String, configure: ExternalModuleDependency.() -> Unit): ExternalModuleDependency
fun <T : Dependency> implementation(dependency: T, configure: T.() -> Unit): T
fun implementation(dependencyNotation: String, configure: Closure<*>)
```

and the lambda we want is the one on `ExternalModuleDependency`, the second one

gradle/libs.versions.toml
BoD: Yup, not a big deal, but my reasoning was that the `-` are used to break up the generated fields and so it makes sense for them to be logically grouped, and for instance `apollo.normalized.cache` doesn't make much sense because `normalized` is not a thing. But we can keep `normalizedcache` in lowercase though!

tests/jvmoverloads/build.gradle.kts
BoD: Ahaha no worries, consistency is always a good thing! I'll put back the `id`!

Note: you can also use the version catalog for plugins with `alias(libs.plugins.xyz)` but that only works if the version is specified, which is not what we want.

apollo-gradle-plugin/testProjects/groovy/build.gradle
martinbonnin: Looks like something that I wrote quickly to verify something and was too lazy to add a test for... I'd be in favor of keeping it and adding a single tests that calls `./gradlew --help` or so, just making sure the build script compiles. I can clean up my own mess and do, just let me know.

PR #4328, opened 2022-08-10
Introduce `FakeResolver`

martinbonnin:
Introduce `FakeResolver`:

```kotlin
interface FakeResolver {
  fun resolveLeaf(context: FakeResolverContext): Any
  fun resolveListSize(context: FakeResolverContext): Int
  fun resolveMaybeNull(context: FakeResolverContext): Boolean
  fun resolveTypename(context: FakeResolverContext): String
}
```

It's now possible to create a fake model with just one line:

```kotlin
val data = GetFooQuery.Data()
```

See https://github.com/apollographql/apollo-kotlin/issues/4016


tests/data-builders/src/test/kotlin/test/DataBuilderTest.kt
```diff
@@ -113,4 +118,83 @@ class DataBuilderTest {
     assertEquals(43, data.long2?.value)
     assertEquals(44, data.long3)
   }
+
+  @Test
+  fun fakeValues() {
```
BoD, line 123: 😍

BoD (approved):
🎉

PR #4329, opened 2022-08-10
enableCompatibilityMetadataVariant

martinbonnin:
Okio 3.2.0 supports enableCompatibilityMetadataVariant so looks like we can too 🎉


BoD (approved):
Nice! 👍

PR #4330, opened 2022-08-11
reflective Data.toJson()

martinbonnin:
Add `Data.toJson` for native/JS


apollo-annotations/src/appleMain/kotlin/com/apollographql/apollo3/annotations/ApolloAdaptableWith.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo3.annotations
+
+import kotlin.reflect.AssociatedObjectKey
+import kotlin.reflect.ExperimentalAssociatedObjects
+import kotlin.reflect.KClass
+
+@OptIn(ExperimentalAssociatedObjects::class)
```
martinbonnin, line 7: This is using experimental Kotloin stdlib features but since it's only for native/JS, I think it's ok

apollo-annotations/src/appleMain/kotlin/com/apollographql/apollo3/annotations/ApolloAdaptableWith.kt
```diff
@@ -0,0 +1,10 @@
...
+@AssociatedObjectKey
```
BoD, line 8: TIL! Very cool feature! 👍

BoD (approved):
👏

apollo-annotations/src/appleMain/kotlin/com/apollographql/apollo3/annotations/ApolloAdaptableWith.kt
martinbonnin: Not sure I want to know how it's implemented 😅 ...
... but it's useful!

Issue #4016, opened 2022-04-11
Schema based test builders


martinbonnin:
When working with Jetpack Compose `@Preview`, faking data is useful in the `main` source set while currently, the test builders are only generated for the `test` source set by default. This can be overriden but it's not straightforward. Also because the current test builders are `responseBased`, they generate a fair amount of code, not to mention that specific code is generated for each and every operation.

Instead, we could generate `schemaBased` test builders only once and add them to the main source set. For a query like this:

```graphql
query HeroName {
  hero {
    name
  }
}
```

generate a fake model like this:
```kotlin
// pseudo code, not sure this can work exactly like this
val data = buildFakeData<HeroName> {
  hero = humanHero {
    name = "Luke"
    // Here it's not really type safe but that's certainly OK for fake models
    height = 180
  }
}
```

We could even enable `strict mode` so that errors are caught at runtime: see https://github.com/apollographql/apollo-kotlin/issues/3344

Todo:
* [x] custom scalars support
* [x] default values
* [x] add reflective toJson for native/Js 
* [x] use in our own tests
* [x] java codegen
* [x] update doc
* [ ] (maybe) strict mode
* [ ] (maybe) [kotlin-faker](https://github.com/serpro69/kotlin-faker) resolver
* [ ] (maybe) cleanup the IrType2 stuff

martinbonnin:
Current proposal for generated code:

```kotlin
class HeroQuery {
  class Data(val hero: Hero) {

  }
  class Hero(val name: String)

  companion object {
    fun Data(block: QueryBuilder.() -> Unit): Data {
      // Code goes here
    }
  }  
}

class AliasedProperty(val alias: String, val map: MutableMap<String, Any?>) {
  infix fun to(value: Any) {
    map[alias] = value
  }
}


@Suppress("PropertyName")
abstract class ObjectBuilder {
  protected val __map = mutableMapOf<String, Any?>()
  
  var __typename: String by __map

  fun alias(alias: String): AliasedProperty {
    return AliasedProperty(alias, __map)
  }
}

class QueryBuilder: ObjectBuilder() {
  var hero: Character? by __map

  fun build(): Query {
    return Query(__map)
  }
}

fun buildQuery(block: QueryBuilder.() -> Unit): Query {
  val builder = QueryBuilder()
  builder.__typename = "Query"

  builder.block()

  return builder.build()
}

class Query(map: Map<String, Any?>): Map<String, Any?> by map

[...]

PR #4331, opened 2022-08-11
use Data builders in our own tests, part 1

martinbonnin:
Dogfooding the data builders, I realized 2 things:

* We'd need `Executable.Data.toJson` instead of just `Operation.Data.toJson` (for fragments), same for Data builders
* Being able to specify the runtime adapters in `buildQuery()` would be nice

I'll add the two points above before continuing the dog fooding

see #4016

### DIFF
--- a/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/mockserver.kt
+++ b/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/mockserver.kt
@@ -1,9 +1,12 @@
 package com.apollographql.apollo3.testing
 
+import com.apollographql.apollo3.api.AnyAdapter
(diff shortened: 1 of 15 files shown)


apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/mockserver.kt
```diff
@@ -19,3 +22,40 @@ fun <D : Operation.Data> MockServer.enqueue(
...
+
+  enqueue(MockResponse.Builder()
+      .statusCode(statusCode)
+      .body(response)
+      .delayMillis(delayMs)
+      .build())
+}
+
+
+fun MockServer.enqueueData(
```
BoD, line 44: 👍

BoD (approved):
Definitely improves the tests' readability!

Issue #3531, opened 2021-11-09
Move the build system to convention plugins


martinbonnin:
Like described in https://developer.squareup.com/blog/herding-elephants/, move our `subprojects {}` blocks to convention plugins so that we have a better modeling of our build.

BoD:
Done in #4333

PR #4333, opened 2022-08-11
Move the build system to convention plugins

BoD:
See #3531


apollo-gradle-plugin-external/build.gradle.kts
```diff
@@ -29,3 +30,10 @@ gradlePlugin {
 gr8 {
   removeGradleApiFromApi()
 }
+
+// Override convention plugin behavior
```
martinbonnin, line 34: Make `allWarningsAsError` part of the extension?

build-logic/src/main/kotlin/Mpp.kt
```diff
@@ -69,7 +69,7 @@ fun Project.okioNodeJs(): String {
   return "com.squareup.okio:okio-nodefilesystem:$okioVersion"
 }
 
-fun KotlinMultiplatformExtension.configureAppleTargets(vararg presetNames: String) {
+private fun KotlinMultiplatformExtension.configureAppleTargets(presetNames: Set<String>) {
```
martinbonnin, line 72: Just curious why no vararg anymore?

settings.gradle.kts
```diff
@@ -14,9 +14,12 @@ includeBuild("build-logic")
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 
 pluginManagement {
+  includeBuild("build-logic")
+
   repositories {
     mavenCentral()
     google()
+    gradlePluginPortal()
```
martinbonnin, line 22: Do you mind adding a content block?

```kotlin
      content {
        includeModule("foo", "bar")
      }
```

`gradlePluginPortal()` transparently redirects to other repos so I'd trust the other repos more than `gradlePluginPortal()`

martinbonnin (approved):
A couple of questions but looks very good 👍 . Thanks for diving into this 💙

build-logic/src/main/kotlin/com/apollographql/apollo3/buildlogic/plugin/LibraryConventionPlugin.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    fun mpp(action: Action<MppConfiguration>) {
+      action.execute(mppConfiguration)
+      project.configureMppDefaults(
+          withJs = mppConfiguration.withJs.getOrElse(true),
+          withLinux = mppConfiguration.withLinux.getOrElse(true)
+      )
+    }
+
+    fun javaModuleName(javaModuleName: String) {
```
BoD, line 51: Had to make this one a function instead of a Property in the end, otherwise it was (on jvm modules only!) called too early when the value of the property wasn't available uet.

martinbonnin (approved):

PR #4340, opened 2022-08-12
Enable Gradle's build cache

BoD:
Enable Gradle's build cache. 

I manually disabled it on the CI for the push and tag workflows, to minimize the risk of non-reproducible builds there.

### DIFF
--- a/tests/gradle.properties
+++ b/tests/gradle.properties
@@ -6,3 +6,7 @@ kotlin.mpp.stability.nowarn=true
 xcodeproj=samples/multiplatform/kmp-ios-app/kmp-ios-app.xcodeproj
 
 android.useAndroidX=true
+
+# Enable the build cache
+# See https://docs.gradle.org/current/userguide/build_cache.html
+org.gradle.caching=true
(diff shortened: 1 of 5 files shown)


tests/gradle.properties
```diff
@@ -6,3 +6,7 @@ kotlin.mpp.stability.nowarn=true
 xcodeproj=samples/multiplatform/kmp-ios-app/kmp-ios-app.xcodeproj
 
 android.useAndroidX=true
+
+# Enable the build cache
+# See https://docs.gradle.org/current/userguide/build_cache.html
+org.gradle.caching=true
```
BoD, line 12: Not sure this is needed or if the one in the root project is enough.

martinbonnin: It's technically an included build and not the root project so I would expect it to be required (but I'm not sure).

martinbonnin (approved):
Almost sounds too easy 😎 . I guess we'll see how that goes.

tests/gradle.properties
BoD: Confirmed: it's needed.

```
./gradlew -p tests :enums:check
./gradlew -p tests :enums:clean
./gradlew -p tests :enums:check
```

- Without it, tests run again
- With it: `Task :enums:test FROM-CACHE`

martinbonnin: 👍 . That being said, running tests from cache seems... a good way to avoid seeing things that break... But we'll see...

PR #4341, opened 2022-08-12
Add Java Data Builders

martinbonnin:
Add Java builders. Current experience looks like this:

```java
    GetAnimalQuery.Data data = GetAnimalQuery.buildData(
        factory.buildQuery()
            .animal(
                factory.buildLion()
                    .species("LionSpecies")
                    .roar("Rooooaaarr")
                    .build()
            )
            .build()
    );
```

The `factory` part is a bit unfortunate but that's the only way I found to pass `customScalarAdapters` around

See #4016


BoD (approved):
🚀

Issue #4320, opened 2022-08-04
Add a flag to disable the fieldsCanMerge validation on disjoint types 


martinbonnin:
Given this schema:

```graphql
  type Query {
    foo: PropertyValue
  }
  union PropertyValue = PropertyValueBoolean | PropertyValueInt

  type PropertyValueBoolean {
    name: String!
    value: Boolean
  }
  type PropertyValueInt {
    name: String!
    value: Int
  }
```

and this query:

```graphql
  query GetFoo {
    foo {
      ...PropertyValue
    }
  }
  fragment PropertyValue on PropertyValue {
    __typename
    ... on PropertyValueBoolean {
        value
    }
    ... on PropertyValueInt {
        value
    }
  }
```

Validation fails with:

```
`value` cannot be merged with `value`: they have different shapes. 
```

This is expected from reading [the spec](https://spec.graphql.org/draft/#sel-GALRFFfCAACGBssG) although it's not clear what the benefit of this validation is on completely disjoint types and it makes the generated models more akward to use. 

Other users have bumped into the same issue in other projects like [here](https://github.com/graphql/graphql-js/issues/53#issuecomment-1077737208)

We could have a Gradle flag that specifically disables this validation:

```kotlin
apollo {
  fieldOnDisjointTypesMustMerge = false
}
```

More context:
* Spec pull request: https://github.com/graphql/graphql-spec/pull/162
* Slack discussion: https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1659536084099869

pt2121:
Can I give this a try?

martinbonnin:
Very sorry for the delay @pt2121 ! Thanks for looking into this. I'm reviewing the PR now, I'll follow up there.

martinbonnin:
This is now available in 3.6.0

PR #4342, opened 2022-08-14
Add a flag to disable fieldsCanMerge validation on disjoint types

pt2121:
re #4320

Basically, I skip `fieldsInSetCanMerge` if the `fieldOnDisjointTypesMustMerge` flag is `true`.
Please let me know if that's the expectation.

### DIFF
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
@@ -2,7 +2,6 @@
 
 package com.apollographql.apollo3.ast
 
-import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.annotations.ApolloInternal
 import com.apollographql.apollo3.ast.internal.ExecutableValidationScope
@@ -27,9 +26,9 @@ fun BufferedSource.toSchema(filePath: String? = null): Schema = parseAsGQLDocume
  * See [parseAsGQLDocument] and [validateAsExecutable] for more granular error reporting
  */
 @ApolloExperimental
-fun BufferedSource.toExecutableDefinitions(schema: Schema, filePath: String? = null): List<GQLDefinition> = parseAsGQLDocument(filePath)
+fun BufferedSource.toExecutableDefinitions(schema: Schema, filePath: String? = null, fieldOnDisjointTypesMustMerge: Boolean = false): List<GQLDefinition> = parseAsGQLDocument(filePath)
     .valueAssertNoErrors()
-    .validateAsExecutable(schema)
+    .validateAsExecutable(schema, fieldOnDisjointTypesMustMerge)
     .valueAssertNoErrors()
 
 /**
@@ -88,12 +87,14 @@ fun GQLDocument.validateAsSchemaAndAddApolloDefinition(): GQLResult<Schema> {
 /**
  * Validates the given document as an executable document.
  *
+ * @param fieldOnDisjointTypesMustMerge if true, the FieldsInSetCanMerge validation is not run.
+ *
  * @return  a [GQLResult] containing the operation and fragment definitions in 'value', along with any potential issues
  */
 @ApolloExperimental
-fun GQLDocument.validateAsExecutable(schema: Schema): GQLResult<List<GQLDefinition>> {
+fun GQLDocument.validateAsExecutable(schema: Schema, fieldOnDisjointTypesMustMerge: Boolean): GQLResult<List<GQLDefinition>> {
   val fragments = definitions.filterIsInstance<GQLFragmentDefinition>().associateBy { it.name }
-  val issues = ExecutableValidationScope(schema, fragments).validate(this)
+  val issues = ExecutableValidationScope(schema, fragments, fieldOnDisjointTypesMustMerge).validate(this)
   return GQLResult(definitions, issues)
 }
 
@@ -104,6 +105,7 @@ fun GQLDocument.validateAsExecutable(schema: Schema): GQLResult<List<GQLDefiniti
 fun GQLFragmentDefinition.inferVariables(
     schema: Schema,
     fragments: Map<String, GQLFragmentDefinition>,
-) = ExecutableValidationScope(schema, fragments).inferFragmentVariables(this)
+    fieldOnDisjointTypesMustMerge: Boolean
+) = ExecutableValidationScope(schema, fragments, fieldOnDisjointTypesMustMerge).inferFragmentVariables(this)
 
 
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
@@ -26,9 +26,9 @@ fun BufferedSource.toSchema(filePath: String? = null): Schema = parseAsGQLDocume
  * See [parseAsGQLDocument] and [validateAsExecutable] for more granular error reporting
  */
 @ApolloExperimental
-fun BufferedSource.toExecutableDefinitions(schema: Schema, filePath: String? = null, fieldOnDisjointTypesMustMerge: Boolean = false): List<GQLDefinition> = parseAsGQLDocument(filePath)
+fun BufferedSource.toExecutableDefinitions(schema: Schema, filePath: String? = null, fieldsOnDisjointTypesMustMerge: Boolean = true): List<GQLDefinition> = parseAsGQLDocument(filePath)
     .valueAssertNoErrors()
-    .validateAsExecutable(schema, fieldOnDisjointTypesMustMerge)
+    .validateAsExecutable(schema, fieldsOnDisjointTypesMustMerge)
     .valueAssertNoErrors()
 
 /**
@@ -87,14 +87,14 @@ fun GQLDocument.validateAsSchemaAndAddApolloDefinition(): GQLResult<Schema> {
 /**
  * Validates the given document as an executable document.
  *
- * @param fieldOnDisjointTypesMustMerge if true, the FieldsInSetCanMerge validation is not run.
+ * @param fieldsOnDisjointTypesMustMerge Whether fields with different shape are disallowed to be merged in disjoint types.
  *
  * @return  a [GQLResult] containing the operation and fragment definitions in 'value', along with any potential issues
  */
 @ApolloExperimental
-fun GQLDocument.validateAsExecutable(schema: Schema, fieldOnDisjointTypesMustMerge: Boolean): GQLResult<List<GQLDefinition>> {
+fun GQLDocument.validateAsExecutable(schema: Schema, fieldsOnDisjointTypesMustMerge: Boolean = true): GQLResult<List<GQLDefinition>> {
   val fragments = definitions.filterIsInstance<GQLFragmentDefinition>().associateBy { it.name }
-  val issues = ExecutableValidationScope(schema, fragments, fieldOnDisjointTypesMustMerge).validate(this)
+  val issues = ExecutableValidationScope(schema, fragments, fieldsOnDisjointTypesMustMerge).validate(this)
   return GQLResult(definitions, issues)
 }
 
@@ -105,7 +105,7 @@ fun GQLDocument.validateAsExecutable(schema: Schema, fieldOnDisjointTypesMustMer
 fun GQLFragmentDefinition.inferVariables(
     schema: Schema,
     fragments: Map<String, GQLFragmentDefinition>,
-    fieldOnDisjointTypesMustMerge: Boolean
-) = ExecutableValidationScope(schema, fragments, fieldOnDisjointTypesMustMerge).inferFragmentVariables(this)
+    fieldsOnDisjointTypesMustMerge: Boolean
+) = ExecutableValidationScope(schema, fragments, fieldsOnDisjointTypesMustMerge).inferFragmentVariables(this)
 
 
--- a/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
+++ b/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
@@ -87,7 +87,8 @@ fun GQLDocument.validateAsSchemaAndAddApolloDefinition(): GQLResult<Schema> {
 /**
  * Validates the given document as an executable document.
  *
- * @param fieldsOnDisjointTypesMustMerge Whether fields with different shape are disallowed to be merged in disjoint types.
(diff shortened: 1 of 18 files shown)


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -189,7 +189,14 @@ class Options(
      */
     val addJvmOverloads: Boolean = false,
     val addTypename: String = defaultAddTypename,
-    val requiresOptInAnnotation: String? = defaultRequiresOptInAnnotation
+    val requiresOptInAnnotation: String? = defaultRequiresOptInAnnotation,
+
+    /**
+     * Whether to merge fields in disjoint types without the FieldsInSetCanMerge validate.
+     *
+     * If true, disable the FieldsInSetCanMerge validation.
```
BoD, line 197: I think the meaning of the flag has been inverted: 
- `fieldsOnDisjointTypesMustMerge = true` means the validation is run (the default)
- `fieldsOnDisjointTypesMustMerge = false` means the validation is disabled

```suggestion
     * Whether fields with different shape are disallowed to be merged in disjoint types.
     * 
     * Note: setting this to `false` relaxes the standard GraphQL [FieldsInSetCanMerge](https://spec.graphql.org/draft/#FieldsInSetCanMerge()) validation which may still be 
     * run on the backend.
     *
     * See also [issue 4320](https://github.com/apollographql/apollo-kotlin/issues/4320)
     *
     * Default: true
```

(Also should it be `fieldsOnDisjointTypesMustMerge`? I think plural makes more sense.)

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gqloperationdefinition.kt
```diff
@@ -12,4 +12,5 @@ fun GQLOperationDefinition.rootTypeDefinition(schema: Schema) = when (operationT
 fun GQLOperationDefinition.validate(
     schema: Schema,
     fragments: Map<String, GQLFragmentDefinition>,
-) = ExecutableValidationScope(schema, fragments).validateOperation(this)
\ No newline at end of file
+    fieldOnDisjointTypesMustMerge: Boolean,
```
BoD, line 15: To avoid a breaking change, add a default value for `fieldsOnDisjointTypesMustMerge` and `@JvmOverloads`

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
```diff
@@ -88,12 +87,14 @@ fun GQLDocument.validateAsSchemaAndAddApolloDefinition(): GQLResult<Schema> {
 /**
  * Validates the given document as an executable document.
  *
+ * @param fieldOnDisjointTypesMustMerge if true, the FieldsInSetCanMerge validation is not run.
+ *
  * @return  a [GQLResult] containing the operation and fragment definitions in 'value', along with any potential issues
  */
 @ApolloExperimental
-fun GQLDocument.validateAsExecutable(schema: Schema): GQLResult<List<GQLDefinition>> {
+fun GQLDocument.validateAsExecutable(schema: Schema, fieldOnDisjointTypesMustMerge: Boolean): GQLResult<List<GQLDefinition>> {
```
BoD, line 95: A default value for the param would probably make sense here too.

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/ValidationTest.kt
```diff
@@ -5,54 +5,72 @@ import com.apollographql.apollo3.ast.parseAsGQLDocument
...
 import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
 import java.io.File
+import kotlin.test.assertTrue
 
 @Suppress("UNUSED_PARAMETER")
-@RunWith(Parameterized::class)
-class ValidationTest(name: String, private val graphQLFile: File) {
-  private val separator = "\n------------\n"
+@RunWith(Enclosed::class)
```
BoD, line 20: Didn't know about `Enclosed`, but at least on my side the new test doesn't seem to run when running `./gradlew :apollo-compiler:test`. Maybe put it in its own file instead?

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gqloperationdefinition.kt
BoD: Then you can run `./gradle apiDump` to generate the compatibility report and push it (the CI checks it).

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -507,6 +510,9 @@ internal class ExecutableValidationScope(
   }
 
   private fun fieldsInSetCanMerge(fieldsWithParent: List<FieldWithParent>) {
+    if (fieldOnDisjointTypesMustMerge) {
```
BoD, line 513: I think we're disabling more than what you need here, maybe we could move this a bit deeper? In `sameResponseShapeRecursive` and `haveSameResponseShape`?

BoD (commented):
Thank you for this!

A few comments, but looks good overall.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
pt2121: Whoops, sorry, the comment is wrong. I will fix these.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gqloperationdefinition.kt
pt2121: Thanks for the suggestions. I'll update

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/ValidationTest.kt
pt2121: My bad, I ran the validation test from my IDE but didn't try from command line.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
pt2121: Yeah, that makes sense. 
I was confused by the issue title - it made me think that we want to disable this whole validation.

pt2121:
Thanks for the feedback! I will look into them tonight.

apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -542,6 +542,14 @@ interface Service {
   @ApolloExperimental
   val requiresOptInAnnotation: Property<String>
 
+  /**
+   * Whether to merge fields in disjoint types without the FieldsInSetCanMerge validate.
+   * If true, disable the FieldsInSetCanMerge validation.
+   *
+   * Default value: false
+   */
```
BoD, line 550:
```suggestion
  /**
   * Whether fields with different shape are disallowed to be merged in disjoint types.
   *
   * Note: setting this to `false` relaxes the standard GraphQL [FieldsInSetCanMerge](https://spec.graphql.org/draft/#FieldsInSetCanMerge()) validation which may still be
   * run on the backend.
   *
   * See also [issue 4320](https://github.com/apollographql/apollo-kotlin/issues/4320)
   *
   * Default: true.
   */
```

pt2121: Fixed

BoD (approved):
Thanks a lot! This looks good 👍

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/api.kt
```diff
@@ -88,12 +87,14 @@ fun GQLDocument.validateAsSchemaAndAddApolloDefinition(): GQLResult<Schema> {
 /**
  * Validates the given document as an executable document.
  *
+ * @param fieldsOnDisjointTypesMustMerge Whether fields with different shape are disallowed to be merged in disjoint types.
```
martinbonnin, line 90: Suggestion to rephrase the comment from negative voice to positive. My brain tends to struggle with negations:

```suggestion
 * @param fieldsOnDisjointTypesMustMerge set to false to relax the standard GraphQL [FieldsInSetCanMerge](https://spec.graphql.org/draft/#FieldsInSetCanMerge()) 
 * and allow fields of different types at the same Json path as long as their parent types are disjoint.
```

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -44,10 +44,13 @@ import com.apollographql.apollo3.ast.sharesPossibleTypesWith
 /**
  * @param fragmentDefinitions: all the fragments in the current compilation unit.
  * This is required to check the type conditions as well as fields merging
+ *
+ * @param fieldsOnDisjointTypesMustMerge Whether fields with different shape are disallowed to be merged in disjoint types.
```
martinbonnin, line 48: For the sake of consistency (albeit this is private API so doesn't matter as much):
```suggestion
 * @param fieldsOnDisjointTypesMustMerge set to false to relax the standard GraphQL [FieldsInSetCanMerge](https://spec.graphql.org/draft/#FieldsInSetCanMerge()) 
 * and allow fields of different types at the same Json path as long as their parent types are disjoint.
```

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/ValidationTest.kt
```diff
@@ -31,7 +26,7 @@ class ValidationTest(name: String, private val graphQLFile: File) {
       if (parseResult.issues.isNotEmpty()) {
         parseResult.issues
       } else {
-        parseResult.valueAssertNoErrors().validateAsExecutable(schema = schema!!).issues + if (graphQLFile.name == "capitalized_fields_disallowed.graphql") {
+        parseResult.valueAssertNoErrors().validateAsExecutable(schema = schema!!, fieldsOnDisjointTypesMustMerge = true).issues + if (graphQLFile.name == "capitalized_fields_disallowed.graphql") {
```
martinbonnin, line 29: Nit: true is the default already:

```suggestion
        parseResult.valueAssertNoErrors().validateAsExecutable(schema = schema!!).issues + if (graphQLFile.name == "capitalized_fields_disallowed.graphql") {
```

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/FieldsOnDisjointTypesMustMergeTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import okio.source
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import java.io.File
+
+class FieldsOnDisjointTypesMustMergeTest {
+  private val graphQLFile = File("src/test/validation/operation/fields_on_disjoint_types_must_merge/different_shapes.graphql")
+
+  @Test
+  fun testEnableFieldsOnDisjointTypesMustMerge() = checkExpected(graphQLFile) { schema ->
```
martinbonnin, line 18: This test is run twice:
- once here
- once in `ValidationTest`

Generally speaking, `src/test/validation/` is "owned" by `ValidationTest` so I would just use the `ValidationTest` infrastructure. Something like this (assuming you rename `different_shapes` to `merging_allowed`:

```kotlin
        val mustMerge = graphQLFile.name != "merging_allowed.graphql"
        parseResult.valueAssertNoErrors().validateAsExecutable(schema = schema!!, fieldsOnDisjointTypesMustMerge = mustMerge).issues + if (graphQLFile.name == "capitalized_fields_disallowed.graphql") {
```

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/ValidationTest.kt
martinbonnin: For the record (not for this PR), we could move the test to https://github.com/google/TestParameterInjector. It's pretty handy (but not multiplatform)

martinbonnin (commented):
A few comments but looks good to me overall. Let me know what you think.

On a separate note, reviewing this, I realized we have a bug in the validation logic and something like this doesn't throw a validation error:

```graphql
type Query {
  value: Int
  value2: Int
}
```

```graphql
{
  value
  value: value2
}
```

A server should (rightfully) forbid this. Since this is permissive at the moment, I don't want to [break other people workflows](https://xkcd.com/1172/) by making this more strict but this is certainly something to revisit for v4

pt2121:
Martin, thanks for the review. I'll look into them this weekend.

apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
pt2121: Fixed

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/FieldsOnDisjointTypesMustMergeTest.kt
pt2121: I updated the test - added `merging_allowed` which is the same as `different_shapes` but it's used to validate the case where `fieldsOnDisjointTypesMustMerge` is `false`. let me know if I understand your suggestion correctly

martinbonnin: Yep, all good, thanks!

martinbonnin:
Thanks for the contribution! I just launched the CI tests. I'll merge once everything is ✅

pt2121:
Cool, thanks for a small ticket for a new contributor :slightly_smiling_face:

pt2121:
Ah, test failed somehow. :sweat_smile: 
It looks Ok when I ran it on my local. let me recheck

martinbonnin:
That's a flake, I've seen it before (see also https://github.com/apollographql/apollo-kotlin/issues/3920). I just relaunched the CI, it should pass this time

martinbonnin:
It's all merged. Thanks again for the contribution!

PR #4344, opened 2022-08-16
Enable Gradle Enterprise

### LINKED ISSUE #4335: enable Gradle enterprise
Enabling Gradle Enterprise unlocks a few things for the project:

1. More metrics and faster build times in general
2. Shared network cache enabling maintainers and contributors to complete a build within seconds

Goooler:
Closes #4335

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -23,3 +23,14 @@ pluginManagement {
   }
 }
 
+plugins {
+  `gradle-enterprise`
+}
+
+gradleEnterprise {
+  buildScan {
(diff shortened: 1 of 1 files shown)


BoD:
Thanks! 

Just a heads up that we are exploring if we'll be able to obtain a [non free](https://gradle.com/pricing/) license of Gradle Enterprise.

Goooler:
ok

Goooler:
@martinbonnin Can we enable build scan publishing for now? I believe it might be useful for us.

martinbonnin:
@Goooler I think so, let me circle back internally to double check we're OK with the terms & conditions but it should be possible 👍

Issue #3694, opened 2021-12-13
[umbrella issue] Java Runtime


martinbonnin:
Now that `apollo-runtime` brings `kotlinx.coroutines` as a dependency, it's not as easy to work with interceptors, etc.. in Java because everything is a `Flow` or `suspend` function.

This issue is to gather interest about writing an alternative runtime that can reuse a lot of the common parts (`apollo-api`, normalization, cache, etc...) but expose a coroutine-free runtime. 

* [x] #4388
* [x] #4389
* [x] #4392
* [ ] #4394
* [x] #4390
* [x] #4391
* [ ] #4395
* [ ] #4396
* [x] #4393
* [x] #4397

kicktipp:
I would love to see better java support. At least it should be documented how to use the lib with java. I tried yesterday to upgrade to 3.0 but failed.

MatthewFrench:
Same, I'm seeing problems using apollo Optional.Absent and graphql query errors like `Type parameter '...Data' is not within its bound; should implement 'com.apollographql.apollo3.api.Operation.Data'`.

martinbonnin:
Hi @MatthewFrench ! Do you have sample code/a reproducer? Maybe we can fix this one without implementing a separate coroutine-free Java runtime.

BoD:
Hi @cristian1980 👋 

Currently the recommended way to use Apollo Kotlin from Java is through the RxJava adapters, which are described [here](https://www.apollographql.com/docs/kotlin/advanced/rxjava). Basically you would do something like:

```java
Single<ApolloResponse<GetAppConfigQuery.Data>> queryResponse = Rx2Apollo.single(
        apolloClient.query(new GetAppConfigQuery())
);
queryResponse.subscribe( 
        response -> {
          // Use response
        },
        throwable -> {
          // Handle error
        }
);
```

brendanlong:
I'm using Rx3Apollo and it works for me in gradle but I get the same error in Intellij:

<img width="1099" alt="Screen Shot 2022-05-05 at 6 51 15 PM" src="https://user-images.githubusercontent.com/1447206/167049436-87e6bd9b-f273-4921-891f-c7ae6dbd2c7a.png">

<img width="1143" alt="Screen Shot 2022-05-05 at 6 51 28 PM" src="https://user-images.githubusercontent.com/1447206/167049474-79e38492-9f87-4519-9512-5d8eb6b8d05c.png">

[...]

PR #4354, opened 2022-08-24
Apollo-java runtime explorations

BoD:
In this exploration, an `apollo-java` module is added.

In this approach the module depends on `apollo-runtime` (but `implementation` only so it doesn't "leak")
  - pros: avoid re-writing some amount of code (and keep future changes in sync)
  - cons: we pull kotlin and coroutines dependencies

The code is mostly a bunch of facades around Kotlin code, using callbacks instead of coroutines/Flows. Sometimes a blocking version also exists, e.g. `ApolloCall.executeBlocking` (naming to be discussed!). 

The facades are written in Java when possible:
- avoids a buch of IntelliJ bugs (e.g. [KTIJ-22127](https://youtrack.jetbrains.com/issue/KTIJ-22127/)) related to usage of generics in Kotlin dependencies
- nicer for users who navigate to the source to see a familiar language (although they won't see anything interesting, and will see Kotlin when going deeper) 

The module also depends on `apollo-normalized-cache-sqlite`:
- pros: cache-specific extensions on builders can be used in a "natural" way
- cons: 
  - dependency on cache even when not used
  - less modular (e.g. can't depend on `incubating` instead)
  - can't use `implementation` here because we want e.g. `FetchPolicy` to be visible. So now we expose 2 `ApolloStore`, you need to import the right one 🙁 (but maybe we could "facade" more things to avoid that)

Lots of things are missing (e.g. Interceptors)!

Misc open questions:
- for one-off calls (e.g. `execute`) should we look into a `Future` based API instead of a specific callback?
- for Flow like calls (e.g. `watch`) should we do something with Rx (like we do now with adapters), or are specific callbacks good / preferrable? Should we look into [Java's Flow](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) (never used it, didn't even know this existed until recently)

Related to #3694.


martinbonnin:
A bunch of early comments/thoughts

> In this approach the module depends on apollo-runtime (but implementation only so it doesn't "leak")
> pros: avoid re-writing some amount of code (and keep future changes in sync)
> cons: we pull kotlin and coroutines dependencies

My gut feeling there is that there is not much logic in `apollo-runtime` and keeping it around would add a lot of friction (in addition to the added dependency)

> Lots of things are missing (e.g. Interceptors)!

That's what I'm talking about. Designing a callback based API for interceptor, passing a non-suspend `HttpEngine`, having builders work without extension functions, etc... is going to be a lot of working around basic stuff...

> Sometimes a blocking version also exists, e.g. ApolloCall.executeBlocking (naming to be discussed!).

Do we need blocking?

> for Flow like calls (e.g. watch) should we do something with Rx (like we do now with adapters)

I think RxJava adapters make a lot of sense. Callback are widely compatible, RxJava/WebFlux/Flow/etc... is for everything else (including blocking calls like above) so we'd be better duplicating it.

> The facades are written in Java when possible:

Yay for Java 👍 

> The module also depends on apollo-normalized-cache-sqlite
> pros: cache-specific extensions on builders can be used in a "natural" way
> cons:
> dependency on cache even when not used
> less modular (e.g. can't depend on incubating instead)

That's a pain but without extension functions, I don't think we can do without this. My 2 cents here would be to design  `apollo-runtime-java` without cache at first. Since it's mostly going to be used on the backend, I'm not sure how much caching is required there.

>can't use implementation here because we want e.g. FetchPolicy to be visible. So now we expose 2 ApolloStore, you need to import the right one 🙁 (but maybe we could "facade" more things to avoid that)

If we decide we need a client cache for Java, we duplicate `ApolloStore` also. The only thing that stays is `apollo-api` everything else is duplicated. I know this duplicates some code but TBH, it's mostly glue that I don't think contains that much logic.

BoD:
A bit of follow-up: I tend to agree with all of this :). With this "front-end" approach, the only benefit is to avoid rewriting some components (+ having to "backport" future fixes and features). But if the code in question is not that big / complex (especially true if we skip the cache for now) rewriting is not a huge deal. Writing façades/wrappers is also far from ideal, it's a lot of boilerplate, and as a Java developer it's not nice to see Kotlin when browsing the sources. And there's the coroutines dependency + odd stacktraces in exceptions.

BoD:
Closing for now, as we're going with the other approach of rewriting (instead of writing wrappers) for the new runtime

Issue #3885, opened 2022-02-21
@defer: opt-out batching for Queries containing @defer


BoD:
At least for now we won't support batching in a `@defer` scenario, so let's automatically opt-out any operation using `@defer` from batching.

BoD:
To implement this we would need a way to know if an operation is using `@defer`, at run-time. Here's a related issue #3985 which is about having the ability to do so.

BoD:
Revisiting/reframing  this one a bit:
- We don't know **if** or **how** a backend will support both features at the same time
- If they do in the future we can implement support for it (or at least, throw a clear error message if it's not possible to use both at the same time) 
- For now I don't think we should do anything. Closing the ticket for now and a specific one can be open later if @defer + batching is a thing.

PR #4356, opened 2022-08-25
Add directives to CompiledSelection / disable batching when @defer is present

BoD:
This may be "over the top" just to disable batching 😅. But considering this could also be useful for [other use cases](https://github.com/apollographql/apollo-kotlin/issues/3985), I gave it a go. 

Relates to #3885

Opening as a draft, and I'll be totally 👍 with closing it if the sentiment is that it's not worth merging 😊

### DIFF
--- a/apollo-api/api/apollo-api.api
+++ b/apollo-api/api/apollo-api.api
@@ -284,10 +284,17 @@ public final class com/apollographql/apollo3/api/CompiledCondition {
 	public fun toString ()Ljava/lang/String;
 }
 
+public final class com/apollographql/apollo3/api/CompiledDirective {
+	public fun <init> (Ljava/lang/String;Ljava/util/List;)V
+	public final fun getArguments ()Ljava/util/List;
+	public final fun getName ()Ljava/lang/String;
+}
+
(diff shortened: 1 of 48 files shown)


martinbonnin:
I'd be completely ok with this throwing an error to the user so it feels a bit overkill to grow the `CompiledField` class for that specific use case. 

Something I'm contemplating at the moment is changing `operation_document` to something that's easier to parse at runtime and get the selections from that (think binary XML ala Android manifest). What we're doing at the moment is that we embed the operation document twice in the apps:

* once as a string (operation_document)
* once as bytecode (selections)

We're currently making the assumption that bytecode is easier to use/faster at runtime so we're trading off space for that but this is something we have never actually verified.

BoD:
> changing operation_document to something that's easier to parse at runtime and get the selections from that (think binary XML ala Android manifest). 

I see what you mean, very interesting!

About the narrower issue of batching vs defer, I'll have a look at if we can throw a descriptive error, which I agree is enough.

BoD:
Closing this for now, it was interesting to work on and gave us an idea of a possible alternative way to go, but not worth merging for now.

PR #4357, opened 2022-08-25
Fix JS tests passing even when they should fail

BoD:
The issue was that the `Promise` wasn't returned by `runTest` following a refact in #4287.


apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/internal/runTest.kt
```diff
@@ -12,36 +11,10 @@ import kotlin.coroutines.EmptyCoroutineContext
...
-    runBlockingOrPromise(context) {
-      before()
-      try {
-        block()
-      } finally {
-        after()
-      }
-    }
-  }
-}
```
martinbonnin, line 42: Could you make that function return `Any` instead ? (`Unit` on JVM/native and `Promise<T>` on JS)

martinbonnin (approved):

apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/internal/runTest.kt
BoD: It looks like JUnit doesn't like it, it wants the tests methods to return `void` (so we could make it work with `fun test() { runTest {} }` instead of `fun test() = runTest {}` but that's a bit annoying

martinbonnin: 👍 sounds good

PR #4358, opened 2022-08-25
Fix #2818

### LINKED ISSUE #2818: `ApolloStore#writeAndPublish` overwrites cache entries when using nested fragments

eduardb:
Following the suggestion from [here](https://github.com/apollographql/apollo-kotlin/issues/2818#issuecomment-750851549):
> A quick fix might be to do the merge in RealResponseWriter so that we do not overwrite the entries in buffer. The different values will come from different ResponseField but maybe it's not a big issue.

Also added an integration test based on the [reproducer](https://github.com/martinbonnin/apollo-android-samples/blob/main/reproducers/2818-fragment-overwrites/src/main/kotlin/main.kt), as suggested by @BoD (thank you!).

### DIFF
--- a/apollo-integration/src/test/java/com/apollographql/apollo/FragmentOverwritesTest.kt
+++ b/apollo-integration/src/test/java/com/apollographql/apollo/FragmentOverwritesTest.kt
@@ -1,0 +1,94 @@
+package com.apollographql.apollo
+
+import com.apollographql.apollo.Utils.immediateExecutor
+import com.apollographql.apollo.Utils.immediateExecutorService
+import com.apollographql.apollo.Utils.mockResponse
+import com.apollographql.apollo.api.Operation
+import com.apollographql.apollo.api.ResponseField
+import com.apollographql.apollo.cache.normalized.CacheKey
+import com.apollographql.apollo.cache.normalized.CacheKeyResolver
+import com.apollographql.apollo.cache.normalized.lru.EvictionPolicy
+import com.apollographql.apollo.cache.normalized.lru.LruNormalizedCacheFactory
+import com.apollographql.apollo.coroutines.await
+import com.apollographql.apollo.fetcher.ApolloResponseFetchers
+import com.apollographql.apollo.integration.fragmentoverwrites.HomeQuery
+import com.apollographql.apollo.integration.fragmentoverwrites.fragment.SectionFragment
+import com.google.common.truth.Truth.assertThat
+import kotlinx.coroutines.runBlocking
+import okhttp3.Dispatcher
+import okhttp3.OkHttpClient
+import okhttp3.mockwebserver.MockWebServer
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+class FragmentOverwritesTest {
+
+    private lateinit var apolloClient: ApolloClient
+
+    @get:Rule
+    val server = MockWebServer()
+
+    private val cacheKeyResolver = object : CacheKeyResolver() {
+        override fun fromFieldArguments(field: ResponseField, variables: Operation.Variables): CacheKey {
+            return CacheKey.NO_KEY
+        }
+
+        override fun fromFieldRecordSet(field: ResponseField, recordSet: Map<String, Any>): CacheKey {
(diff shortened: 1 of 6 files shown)


BoD (approved):
Looks good to me! Thanks a lot 🙏

apollo-integration/src/test/java/com/apollographql/apollo/FragmentOverwritesTest.kt
```diff
@@ -0,0 +1,94 @@
...
+                )
+            ).execute()
+
+            val cacheResponse = apolloClient.query(HomeQuery())
+                .toBuilder()
+                .responseFetcher(ApolloResponseFetchers.CACHE_ONLY)
+                .build()
+                .await()
+
+            assertThat(cacheResponse.data?.home?.sectionA?.name).isEqualTo("modifiedSectionName")
```
martinbonnin, line 91: Can you also add a check that the fragment is correctly written? 

```suggestion
            assertThat(cacheResponse.data?.home?.sectionA?.name).isEqualTo("modifiedSectionName")
            assertThat(cacheResponse.data?.home?.fragments?.sectionFragment?.sectionA?.imageUrl).isEqualTo("modifiedUrl")
```

martinbonnin (commented):
A small comment but looks good to me, thanks!

apollo-integration/src/test/java/com/apollographql/apollo/FragmentOverwritesTest.kt
eduardb: Done in [0c04a41](https://github.com/apollographql/apollo-kotlin/pull/4358/commits/0c04a4174adcb11b66c82122faff6ad683df500c). Thanks!

martinbonnin:
Thanks for the contribution!

PR #4359, opened 2022-08-26
Allow custom scalars targets in FakeResolver return types

martinbonnin:
Previously, compile-time adapters required `resolveLeaf` to return a Json-like value. This break the symmetry with runtime adapters and is more cumbersome to use. 

To allow this, we now generate a `customScalarAdapters` property that contains all compile-time adapters that we can use to automatically serialize when needed.

Random thoughts: the compile-time vs run-time difference make all this code pretty cumbersome. In the future, we should think of ways to make everything a compile-time thing. Maybe with KSP:

```kotlin
@ApolloAdapter // annotation that will signal the adapter to use
object MyLongAdapter: Adapter<MyLong> {
  override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): MyLong {
    return MyLong(reader.nextString()!!.toLong())
  }

  override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: MyLong) {
    writer.value(value.value.toString())
  }
}
```


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -154,6 +154,12 @@ sealed class CompiledNamedType(val name: String) : CompiledType() {
   override fun leafType() = this
 }
 
+/**
+ * A GraphQL scalar type that is mapped to a Kotlin. This is named "Custom" for historical reasons
+ * but is also used for builtin scalars
+ * 
+ * TODO v4: rename this to ScalarType
```
martinbonnin, line 161: I started adding "TODO v4" for stuff that will need to be reworked. I don't think we had any marker previously?

BoD: Good idea!
(I don't think we did)

BoD (approved):
Very cool! 👍

PR #4360, opened 2022-08-26
fixed missing inflate function

engdorm:
I try to go over the tutorial 
step 8 didn't worked
after some debug 
I saw that this function is missing 
but exist in main branch 
so I added it to the tutorial


martinbonnin (approved):
Thanks for sending this!

PR #4361, opened 2022-08-29
Port Java builder code from Apollo Android

### LINKED ISSUE #4312: Generate Java builders for operations
### Use case
At Indeed, we are heavy consumers of GraphQL APIs from Java in backend code. We have used Apollo Android in the past to consume GraphQL APIs. Apollo Android had support for generating convenient builders for operations, variables, and responses. This made it easy to construct operation inputs, and also responses (for unit tests).

Support for Java builders isn't present in Apollo Kotlin. This means that large, complex operation variables and responses have to be constructed using the `new` syntax, which is unwieldy and painful. We've have some engineers revert back to Apollo Android just because of this reason.

We would like for Apollo Kotlin to generate Java builder code for operations, variables, and responses.

dwbutler:
Ports Java compiler builder code from Apollo Android.

Resolves #4312


apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
```diff
@@ -18,6 +19,14 @@ sealed class Optional<out V> {
   object Absent : Optional<Nothing>()
 
   companion object {
+    @JvmStatic
+    fun <T> absent(): Optional<T> = Absent
```
dwbutler, line 23: The generated Java code references `Optional.absent()` and `Optional.fromNullable`. Were these methods removed deliberately? Can `Optional.Absent` be used from Java?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
```diff
@@ -0,0 +1,15 @@
+package com.apollographql.apollo3.api.internal
+
+/**
+ * Represents an operation that accepts a single input argument, mutate it and returns no result
+ *
+ * @param <T> the type of the input to the operation
+</T> */
+interface Mutator<T> {
```
dwbutler, line 8: Referenced from the [generated toBuilder methods](https://github.com/apollographql/apollo-kotlin/pull/4361/files#diff-a5874eb3e26a618a146c7a7ceb6337778e2a398d8f0f9fecda79737ca26712f8R152) and ported from [Mutator.java](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-api/src/jvmMain/java/com/apollographql/apollo/api/internal/Mutator.java).

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -206,7 +206,8 @@ object ApolloCompiler {
             generatedSchemaName = options.generatedSchemaName,
             flatten = options.flattenModels,
             scalarMapping = options.scalarMapping,
-            generateDataBuilders = options.generateDataBuilders
+            generateDataBuilders = options.generateDataBuilders,
```
dwbutler, line 209: @martinbonnin How does the `generateDataBuilders` option differ from the ported `generateModelBuilder` option?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
```diff
@@ -18,6 +19,14 @@ sealed class Optional<out V> {
...
+
+    @JvmStatic
+    fun <T> fromNullable(nullableReference: T?): Optional<T> {
+      return nullableReference?.let { Present(it) } ?: absent()
+    }
```
martinbonnin, line 28: This seems to duplicate `presentIfNotNull` below. Is this a matter of adding `@JvmStatic` on `presentIfNotNull` (not sure how doable this is without breaking ABI compatibility)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
martinbonnin: > Were these methods removed deliberately?

`fromNullable` was renamed to `presentIfNotNull` to make it more explicit. 

> Can Optional.Absent be used from Java?

You can do 

```java
Optional optional = Optional.Absent.INSTANCE;
```

I can see how that's cumbersome though so `Optional.absent()` looks good to me 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
```diff
@@ -0,0 +1,89 @@
...
+
+import com.apollographql.apollo3.compiler.codegen.java.JavaAnnotations
+import com.apollographql.apollo3.compiler.codegen.java.JavaClassNames
+import com.squareup.javapoet.ClassName
+import com.squareup.javapoet.CodeBlock
+import com.squareup.javapoet.ParameterizedTypeName
+import com.squareup.javapoet.TypeName
+import com.squareup.javapoet.WildcardTypeName
+
+fun TypeName.isList() =
```
dwbutler, line 11: These extension methods were ported from [apollo/compiler/testUtils.kt](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/testUtils.kt)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
martinbonnin: `generateDataBuilders` generate data based on the schema and not the models. This is especially useful for fragments:

Assuming something like this:

```graphql
query GetHero {
   hero {
      name
      ...droidFragment
   } 
}

fragment droidFragment on Droid {
   name
   primaryFunction
}
```

The code for each solution is going to look like (not sure about the exact details but that should give the high level overview):

 `generateDataBuilders`:
```java
    BuilderFactory factory = BuilderFactory.DEFAULT
    GetHeroQuery.Data data = GetHeroQuery.buildData(
        factory.buildQuery()
            .hero(
                factory.buildDroid()
                    .name("R2-D2")
                    .primaryFunction("translation")
                    .build()
                )
            )
            .build()
      );
```

`generateModelBuilders`:
```java
    GetHeroQuery.Data data = GetHeroQuery.builder()
       .hero(
          Hero.builder()
                .__typename("Droid")
                .name("R2-D2")
                .droidDetails(
                   DroidDetails.builder()
                      .__typename("Droid")
                      .name("R2-D2")
                      .primaryFunction("translation")
                      .build()
                 )
                .build()
       )
       .build()
```

`generateDataBuilders` has some initial overhead (the factory) but will be simpler as more fragments are involved

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
```diff
@@ -0,0 +1,89 @@
...
+  return (this as ParameterizedTypeName)
+    .typeArguments
+    .first()
+    .let { if (it is WildcardTypeName) it.upperBounds.first() else it }
+}
+
+fun TypeName.isOptional(expectedOptionalType: ClassName? = null): Boolean {
+  val rawType = (this as? ParameterizedTypeName)?.rawType ?: this
+  return if (expectedOptionalType == null) {
+    rawType == JavaClassNames.Optional || rawType == JavaClassNames.Input
```
dwbutler, line 24: Compared to the original code, only `com.apollographql.apollo3.api.Optional` and `com.apollographql.apollo3.api.Input` are supported. (Input is deprecated so I'm not sure whether it's actually needed here.)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
dwbutler: I see, it's a more fluent interface. Is this the right time to update documentation BTW? Or should that be a separate MR?

martinbonnin: Documentation should go in a separate MR so that we can hold the documentation updates until the version is actually released.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -20,7 +20,7 @@ import com.apollographql.apollo3.compiler.singularize
  * Inputs should always be GraphQL identifiers and outputs are valid Kotlin/Java identifiers.
  */
 internal abstract class CodegenLayout(
-    ir: Ir,
+    val ir: Ir,
```
dwbutler, line 23: I exposed this at one point to get access to the list of enums and custom scalars. I believe I ended up not needing to do this so I will revert.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
martinbonnin: This doesn't seem to be used in any test. Can you modify or add a CodegenTest for this?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
martinbonnin: Can you make it `internal` ?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
```diff
@@ -42,5 +48,35 @@ internal class InputObjectBuilder(
           .makeDataClassFromParameters(fields.map {
             it.toNamedType().toParameterSpec(context)
           })
+          .addBuilder()
           .build()
+
+  private fun TypeSpec.Builder.addBuilder(): TypeSpec.Builder {
```
dwbutler, line 54: Ported from [InputTypeSpecBuilder#addBuilder](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/InputTypeSpecBuilder.kt#L56)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
martinbonnin: Maybe `Input` is not needed?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
```diff
@@ -244,5 +247,39 @@ internal class OperationBuilder(
         context.resolver.resolveOperationSelections(operation.name)
     )
   }
-}
 
+  private fun TypeSpec.Builder.addBuilder(context: JavaContext): TypeSpec.Builder {
```
dwbutler, line 251: Ported from [OperationTypeSpecBuilder#addBuilder](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/OperationTypeSpecBuilder.kt#L228)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
```diff
@@ -0,0 +1,240 @@
...
+import com.squareup.javapoet.CodeBlock
+import com.squareup.javapoet.FieldSpec
+import com.squareup.javapoet.MethodSpec
+import com.squareup.javapoet.ParameterSpec
+import com.squareup.javapoet.ParameterizedTypeName
+import com.squareup.javapoet.TypeName
+import com.squareup.javapoet.TypeSpec
+import javax.lang.model.element.Modifier
+
+internal class Builder(
```
dwbutler, line 27: Ported from [BuilderTypeSpecBuilder](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/BuilderTypeSpecBuilder.kt)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
```diff
@@ -0,0 +1,240 @@
...
+      .addMethods(fieldSetterWithMutatorMethodSpecs())
+      .addMethod(buildMethod())
+      .build()
+  }
+
+  private fun builderFields(): List<FieldSpec> {
+    fun valueCode(value: IrValue, type: TypeName): CodeBlock {
+      return when (value) {
+        is IrEnumValue -> CodeBlock.of("\$T.safeValueOf(\$S)", type, value.value)
+        else -> value.codeBlock()
```
dwbutler, line 51: The new `codeBlock()` method allowed the original code to be greatly simplified. Only problem is `IrEnumValue.codeBlock()` doesn't work because it doesn't have access to the type name.

martinbonnin: I think this all goes aways because we don't want default values in Kotlin code?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/model/ModelBuilder.kt
```diff
@@ -103,4 +113,43 @@ internal class ModelBuilder(
       }
     }
   }
+
+  private fun TypeSpec.addBuilder(context: JavaContext): TypeSpec {
```
dwbutler, line 117: Ported from [testUtils#withBuilder](https://github.com/apollographql/apollo-kotlin/blob/66cd44629e374b667d16ee831f5a4d42cc38c50b/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/testUtils.kt#L364). Moved here since now it's only needed in one place, in this class.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
dwbutler: I'll get rid of `fromNullable` and expose `presentIfNotNull` as `@JvmStatic`.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
```diff
@@ -244,5 +247,39 @@ internal class OperationBuilder(
...
+          fieldDefaultValues = emptyMap(),
+          fieldJavaDocs = emptyMap(),
+          context = context
+        ).build()
+      )
+    }
+
+    operation.variables
+      .map {
+        context.layout.propertyName(it.name.decapitalizeFirstLetter()) to context.resolver.resolveIrType(it.type)
```
martinbonnin, line 270: GraphQL is case sensitive so we need to respect the variable case to avoid clashes:
```suggestion
        context.layout.propertyName(it.name) to context.resolver.resolveIrType(it.type)
```

```graphql
query GetHero($id: String!, $ID: String!) {
  hero1: hero(id: $id) { name }
  hero2: hero(id: $ID) { name }
}
```

This is a silly example and unlikely to happen in variables but this is happening for real on field names and enum values so it's more consistent to use the GraphQL case for this.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
dwbutler: I actually didn't find any tests for it in Apollo Android, but I will check again. It seems to be not very well documented how it's supposed to work.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
dwbutler: Might as well! [java.util.Optional(https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) (the implementation I am most familiar with) defines `of` and `ofNullable`

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
```diff
@@ -244,5 +247,39 @@ internal class OperationBuilder(
         context.resolver.resolveOperationSelections(operation.name)
     )
   }
-}
 
+  private fun TypeSpec.Builder.addBuilder(context: JavaContext): TypeSpec.Builder {
+    addMethod(Builder.builderFactoryMethod())
+
+    val operationClassName = ClassName.get("", simpleName)
```
martinbonnin, line 254: JavaPoet certainly doesn't need to import the symbol in this file so it's certainly ok to pass `""` but since we have `packageName` available, this will make the code a bit more robust:

```suggestion
    val operationClassName = ClassName.get(packageName, simpleName)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
dwbutler: Makes sense. None of the original methods were internal, and I don't imagine these extensions methods would get used outside of the compiler, but might as well do it.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
dwbutler: I think you're right. The generated code won't use `Input` anywhere.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
martinbonnin: From the javadoc of `of`:

```
Returns an Optional with the specified present non-null value.
```

That's really not what we want because we want to allow `present(null)` so I'd stick with:

```
Optional.absent()
Optional.present(value: T)
Optional.presentIfNotNull(value:T)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
dwbutler: True. Eventually. 😃 My thought was to port over everything intact, and then later rip out all default value code everywhere in the whole project.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
martinbonnin: If there are no tests for this, I'd just leave it out. It's always easier to add features later than removing something that's illy specified.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/TypeName.kt
martinbonnin: Exactly! This is not a huge issue at the moment but the compiler might become transitively included more and more and we don't want to pollute the namespace too much.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
martinbonnin: Sounds like a plan!

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
dwbutler: Gotcha. Apollo Android was doing this, but I will change it.

dwbutler: There are probably other places this is happening, then.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
martinbonnin: Something else about `data builders` is that they work with `FakeResolver` to provide default values for model fields that haven't been set.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
```diff
@@ -42,5 +48,35 @@ internal class InputObjectBuilder(
...
+    if (inputObject.fields.isEmpty()) {
+      return this
```
martinbonnin, line 56: For operations, you're adding a default empty `Builder` while here (input objects), you're not adding any builder at all. I don't mind either solution but the choice should be consistent

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
dwbutler: Does `Optional.present(null)` make sense given that the parameter is defined to be not null? I think throwing a NullPointerException makes sense. (that's what `java.util.Optional#of` does)

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
dwbutler: From the perspective of someone migrating from Apollo 2 to 3, if I were using `toBuilder` I would expect it to still be there in Apollo 3, even if it wasn't very well documented.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
dwbutler: I agree that's weird, but that's ported straight from Apollo Android. IMHO there should always be a builder there, so I'll see about changing that.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
martinbonnin: > given that the parameter is defined to be not null? 

The parameter is definitely nullable. If we wanted to enforce non-null, we'd put a `Any` upper bound:

```kotlin
// enforce non-null value (which we don't want)
fun <T: Any> present(value: T): Optional<T> = Present(value)

// allow null (which we want)
fun <T> present(value: T): Optional<T> = Present(value)
```

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
martinbonnin: >  if I were using `toBuilder`

Given that it was not tested, I'm not sure how many users use it. What's more, one can use `toBuilder()` without any parameter right? 

```java
MyQuery.Data data = MyQuery.Data.builder().[...].build()

data.toBuilder()
  .name("foo") // change name
  .build()
```

Not sure where `Mutator` is required up there?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
martinbonnin: Make sense 👍 . Agree on `Builder` all the times, let's make all the things consistent.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
dwbutler: I found the original PR: https://github.com/apollographql/apollo-kotlin/pull/680

At the time, you asked the author to add integration tests, but I didn't see any!

I guess it's your decision whether to support this functionality in Apollo 3.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
dwbutler: I made this change, and now I know why this was never noticed - there is no test for it. 🤦  I can sort of see why - an empty input type is unusual.

dwbutler: I don't think it's legal to have an input with no fields:
image
So maybe this is an edge case not worth worrying about too much.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/OperationBuilder.kt
dwbutler: I found and fixed one more place this was happening, with input fields.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
dwbutler: Ah, my misunderstanding of generics in Kotlin I guess. I thought `T?` would allow null, and `T` would not.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
martinbonnin: > I guess it's your decision whether to support this functionality in Apollo 3. 

I think it's fine to leave it out for now. It's always easier to add it later than remove it.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
martinbonnin: Good catch. As it turns out, it's completely invalid to have an input object with no input field. See [the spec](https://spec.graphql.org/draft/#sec-Input-Objects.Type-Validation):

```
An Input Object type must define one or more input fields.
```

So nevermind my initial comment. Input objects with no input fields should never happen. Maybe just add a comment in the source, make sure we don't fall into that trap again?

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Optional.kt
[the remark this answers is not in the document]
martinbonnin: I guess the rationale is that `T` in itself can be of any Kotlin type, including nullable ones. It allows to do stuff like this:

```
fun <T:Any> assertNotNull(value: T?): T
```

So `T` without the `Any` upper bound can be nullable but once you put the `Any` upper bound, you have to type `T?` to make it nullable

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
dwbutler: Interesting. Yeah I'll do that.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
dwbutler: I went ahead and removed all the code that assigns a default value to builder fields so we can have a clean PR to merge into main.

apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/Mutator.kt
dwbutler: Okay, I'll leave `toBuilder` but remove mutator support.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
dwbutler: Okay I reverted my change and added a comment.

apollo-api/src/commonTest/kotlin/test/OptionalTest.kt
```diff
@@ -2,12 +2,19 @@ package test
...
 import com.apollographql.apollo3.api.Optional
 import kotlin.test.Test
+import kotlin.test.assertFails
 import kotlin.test.assertIs
 
 class OptionalTest {
+  @Test
+  fun presentTest() {
+    assertIs<Optional.Present<*>>(Optional.present("some value"))
+    assertIs{ Optional.present(null) }
```
martinbonnin, line 12:
```suggestion
    assertIs<Optional.Present<*>>(Optional.present(null))
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/JavaAnnotations.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo3.compiler.codegen.java
+
+import com.squareup.javapoet.AnnotationSpec
+import org.jetbrains.annotations.NotNull as JetBrainsNonNull
+import org.jetbrains.annotations.Nullable as JetBrainsNullable
+
+object JavaAnnotations {
```
martinbonnin, line 7:
```suggestion
internal object JavaAnnotations {
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/model/ModelBuilder.kt
```diff
@@ -103,4 +112,42 @@ internal class ModelBuilder(
...
+      return this
+    } else {
+      val builderVariable = JavaClassNames.Builder.simpleName().decapitalizeFirstLetter()
+      val builderClass = ClassName.get("", JavaClassNames.Builder.simpleName())
+      val toBuilderMethod = MethodSpec.methodBuilder(Builder.TO_BUILDER_METHOD_NAME)
+        .addModifiers(Modifier.PUBLIC)
+        .returns(builderClass)
+        .addStatement("\$T \$L = new \$T()", builderClass, builderVariable, builderClass)
+        .addCode(fields
+          .map { CodeBlock.of("\$L.\$L = \$L;\n", builderVariable, it.name, it.type.unwrapOptionalValue(it.name)) }
```
martinbonnin, line 129: Is there a specific reason you call `unwrapOptionalValue` here? Looks like it's only required for `Optional<T>` values which are not generated right now (see also https://github.com/apollographql/apollo-kotlin/issues/4337)

Even if `Optional` were generated, I think it's fine to have the builder field have the same type as the model? It's more consistent? 


```java
  public static class Animal {
    public Optional<String> name;
  }

  public static final class Builder {
    private Optional<String> name;
  }

```

martinbonnin, line 129: You'll need to escape the property name:

```suggestion
          .map { CodeBlock.of("\$L.\$L = \$L;\n", builderVariable, context.layout.propertyName(it.name), it.type.unwrapOptionalValue(it.name)) }
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/model/ModelBuilder.kt
```diff
@@ -103,4 +112,42 @@ internal class ModelBuilder(
...
+        it.typeSpecs.find { it.name == JavaClassNames.Builder.simpleName() } != null
+      }.map { ClassName.get("", it.name) }
+
+      return toBuilder()
+        .addMethod(toBuilderMethod)
+        .addMethod(Builder.builderFactoryMethod())
+        .addType(
+          Builder(
+            targetObjectClassName = ClassName.get("", name),
+            fields = fields.map { it.name to it.type.unwrapOptionalType() },
```
martinbonnin, line 145: Same as above: I don't think you need to unwrap here?

martinbonnin, line 145: Also same as above: escape the property name:
```suggestion
            fields = fields.map { context.layout.propertyName(it.name) to it.type.unwrapOptionalType() },
```

apollo-api/src/commonTest/kotlin/test/OptionalTest.kt
```diff
@@ -2,12 +2,19 @@ package test
...
+  }
+
   @Test
   fun presentIfNotNullTest() {
     assertIs<Optional.Present<*>>(Optional.presentIfNotNull("some value"))
     assertIs<Optional.Absent>(Optional.presentIfNotNull(null))
```
martinbonnin, line 18: Nice test! 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/model/ModelBuilder.kt
martinbonnin: I guess it boils down to whether we want to be able to pass null to a previously non-null builder:

```java
Data.Builder builder = new Data.Builder();

builder.name(Optional.of("Luke"));

// ... some code

// If we want to be able to do this ultimately, we'll need to keep the  `Optional` wrapper
builder.name(Optional.ofNullable(null))

```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
```diff
@@ -0,0 +1,183 @@
...
+      }
+      .filter { (_, type) ->
+        if (type.isList()) {
+          buildableTypes.contains(type.listParamType())
+        } else {
+          buildableTypes.contains(type)
+        }
+      }
+      .map { (fieldName, fieldType) ->
+        fieldSetterWithMutatorMethodSpec(fieldName, fieldType)
```
martinbonnin, line 108: Is that `Mutator` code still used?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
```diff
@@ -42,5 +43,29 @@ internal class InputObjectBuilder(
...
+          .addBuilder()
           .build()
+
+  private fun TypeSpec.Builder.addBuilder(): TypeSpec.Builder {
+    if (inputObject.fields.isEmpty()) {
+      // The GraphQL spec doesn't allow an input with no fields
+      return this
+    } else {
+      val builderFields = inputObject.fields.map {
+        context.layout.escapeReservedWord(it.name) to context.resolver.resolveIrType(it.type)
```
martinbonnin, line 55:
```suggestion
        context.layout.propertyName(it.name) to context.resolver.resolveIrType(it.type)
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
```diff
@@ -42,5 +43,29 @@ internal class InputObjectBuilder(
...
+      }
+      val javaDocs = inputObject.fields
+        .filter { !it.description.isNullOrBlank() }
+        .associate { context.layout.escapeReservedWord(it.name) to it.description!! }
```
martinbonnin, line 59:
```suggestion
        .associate { context.layout.propertyName(it.name) to it.description!! }
```

tests/java-tests/src/test/java/test/OperationTest.java
```diff
@@ -13,9 +12,9 @@
 public class OperationTest {
   @Test
   public void canUseDefaultFunctions() throws IOException {
-    GetRandomQuery query = new GetRandomQuery();
+    GetRandomQuery query = GetRandomQuery.builder().build();
 
-    GetRandomQuery.Data data = new GetRandomQuery.Data(42);
+    GetRandomQuery.Data data = GetRandomQuery.Data.builder().random(42).build();
```
martinbonnin, line 17: 👍

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/model/ModelBuilder.kt
dwbutler: TBH I just copied this code from Apollo Android without trying to understand it too much. 😅 I'll remove it and see if it has any effect.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/Builder.kt
dwbutler: Good catch, I removed it

martinbonnin (commented):
A few comments/nitpickings on the style. Given that [we're aiming for a better way to address 
 that use case](https://github.com/apollographql/apollo-kotlin/issues/4346#issuecomment-1232175611), I'll close this PR but feel free to continue the discussion here if anything is unclear

martinbonnin:
Closing as it was merged with https://github.com/apollographql/apollo-kotlin/pull/4364. Thanks again!

Issue #4346, opened 2022-08-17
Generate possibleTypes enum for GQLInterface and GQLUnion


chalermpong:
### Use case

This is an alternative solution for my previous problem #4272.

Handling GQLInterface and GQLUnion will be easier if the generate code provides information about possible types.

For example,
```graphql
# schema
interface ThingWithID {
  id: ID!
}
type Human implements ThingWithID {
  id: ID!
  name: String!
}
type Droid implements ThingWithID {
  id: ID!
  model_name: String!
}
type Query {
  something: ThingWithID
}

# Operation
query TestOperation {
  something {
    id
    ... on Human {
      name
    }
    ... on Droid {
      model_name
    }
  }
}
```
The current way to handle something (using operationBased codegen model):
```kotlin
// current way of doing things
when {
  data.something?.onHuman != null -> {
    data.something.onHuman.name
  }
  data.something?.onDroid != null -> {
    data.something.onDroid.model_name
  }
}
```
The new way to handle something:
```kotlin
// new way of doing things
when (something.__typeEnum) {
  ThingWithID.PossibleTypes.Human -> {
    // something.onHuman?.name
  }
  ThingWithID.PossibleTypes.Droid -> {
    // something.onDroid?.name
  }
  ThingWithID.PossibleTypes.UNKNOWN__ -> {
    // Handle unknown type
  }
}
```
This new code is easier to maintain from the following benefits:
- IDE can help generating code to handle all possible case
- IDE can help detecting missing case when new types are added in schema

### Describe the solution you'd like
In generated GQLInterface/GQLUnion class, add `enum class PosssibleTypes`.
```kotlin
public class ThingWithID {
  public enum class PossibleTypes {
    Human("Human"),
    Droid("Droid"),
    /**
     * Auto generated constant for unknown enum values
     */
    UNKNOWN__("UNKNOWN__"),
    ;

    public val rawValue: String = rawValue

    public companion object {
      public fun safeValueOf(rawValue: String): ThingWithID =
          values().find { it.rawValue == rawValue } ?: UNKNOWN__
    }
  }
}
```
In generated operation code, add a compute field `val __typeEnum`.
```kotlin
public class TestOperation() : Query<TestOperation.Data> {
  public data class Data(
    public val something: Something?,
  ) : Query.Data {
    public data class Something(
      public val __typename: String
    ) {
      // Added compute field
      public val __typeEnum: ThingWithID.PossibleTypes
[...]

PR #4362, opened 2022-08-30
Generating possible types enum for Interface and Union (#4346)

chalermpong:
Hi

Please review my code for generating possible types enum here. 
I just finish only operationBased codegen model Kotlin.


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/InterfacePossibleTypeEnumBuilder.kt
```diff
@@ -0,0 +1,38 @@
...
+    private val iface: IrInterface
+) : CgFileBuilder {
+  private val layout = context.layout
+  private val packageName = layout.typePackageName()
+
+  private val simpleName = possibleTypesEnumTypeName(iface.name)
+  private val compiledTypeName = "${layout.compiledTypeName(iface.name)}${Identifier.PossibleTypes}"
+
+  override fun prepare() {
+    context.resolver.registerMapType(simpleName, ClassName(packageName, compiledTypeName))
```
martinbonnin, line 21: As discussed in https://github.com/apollographql/apollo-kotlin/issues/4346#issuecomment-1233231498, you can either not register anything or create your own `registerPossibleTypes` here

apollo-compiler/src/test/graphql/com/example/possible_types_enum/kotlin/operationBased/possible_types_enum/type/FriendPossibleTypes.kt.expected
```diff
@@ -0,0 +1,28 @@
...
+//
+package com.example.possible_types_enum.type
+
+import kotlin.String
+
+/**
+ * Auto generated enum class for enumerable possible type of Friend
+ */
+public enum class FriendPossibleTypes(
+  public val rawValue: String,
```
martinbonnin, line 14: I don't think `rawValue` is needed here?

martinbonnin: Nevermind, it's needed to lookup the correct enum value at runtime

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinCodeGen.kt
```diff
@@ -119,11 +121,13 @@ internal class KotlinCodeGen(
         .filter { !context.resolver.canResolveSchemaType(it.name) }
         .forEach { iface ->
           builders.add(InterfaceBuilder(context, iface, generateDataBuilders))
+          builders.add(InterfacePossibleTypeEnumBuilder(context, iface))
```
martinbonnin, line 124: I would have made the enum a nested class of the main interface to not clutter the namespace too much:

```kotlin
public class Friend {
  public enum class FriendPossibleTypes(
     public val rawValue: String,
  } { /* ... */ }


  public companion object {
    public val type: UnionType = UnionType("Friend", Human.type, Human2.type)
  }
}
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/PossibleTypesEnumCommon.kt
```diff
@@ -0,0 +1,98 @@
...
+        .addParameter("rawValue", KotlinSymbols.String)
+        .build()
+
+private fun enumSpec(typename: String) = TypeSpec.anonymousClassBuilder()
+    .addSuperclassConstructorParameter("%S", typename)
+    .build()
+
+private fun possibleTypesEnumSafeValueOfFunSpec(compiledName: String): FunSpec {
+  return FunSpec.builder(safeValueOf)
+      .addParameter("rawValue", String::class)
```
martinbonnin, line 63: We avoid reflection in the compiler because it doesn't survive R8 relocation in the Gradle plugin

```suggestion
      .addParameter("rawValue", KotlinSymbols.String)
```

martinbonnin (commented):
A few comments/nitpickings on the style. Given that [we're aiming for a better way to address 
 that use case](https://github.com/apollographql/apollo-kotlin/issues/4346#issuecomment-1232175611), I'll close this PR but feel free to continue the discussion here if anything is unclear

PR #4363, opened 2022-08-31
pin `deferSpec` to 20220824

martinbonnin:


abernix (approved):
LGTM and aligns with https://github.com/apollographql/router/pull/1652 and its issue https://github.com/apollographql/router/issues/1648.

Do you need to do anything nice for the user to account for an [HTTP 406](https://github.com/apollographql/router/issues/1659)?

martinbonnin:
> Do you need to do anything nice for the user to account for an https://github.com/apollographql/router/issues/1659?

I don't think so? This will only happen if the router makes a breaking change and bumps to, say, `20220915` and stops supporting `20220824`. In that case users will have to catch the `406` and the app will most likely break but it's not that different from other server errors. What's more, the router already includes an actionable message in the error (`"the router received a query with the @defer directive but the client does not accept multipart/mixed HTTP responses. To enable @defer support, add the HTTP header 'Accept: multipart/mixed; deferSpec=20220824'"` so I think it's best if the client doesn't do anything.

abernix:
Sounds good to me and I believe I agree with you but was just pushing the conversation around to pressure test it a bit. 😉

PR #4364, opened 2022-09-01
Java builders 2

martinbonnin:
This is a follow up from https://github.com/apollographql/apollo-kotlin/pull/4361 with some minor cleanups/tweaks

Thanks @dwbutler for tackling this!!


martinbonnin:
Thanks again @dwbutler

Issue #4272, opened 2022-07-16
Sealed interfaces for polymorphic selection sets


chalermpong:
### Question
I have tried responseBased model which currently generates code as sealed interface. But I have a problem that the generated code was so large that my project never finish compiling. Right now I can use only operationBased model.

With operationBased model, the query like this one:
```graphqls
interface ThingWithID {
    id: ID!
}

type Human implements ThingWithID {
    id: ID!
    name: String!
}

type Droid implements ThingWithID {
    id: ID!
    model_name: String!
}

type Query {
    something: ThingWithID
}
```
will generate the following code:
```kotlin
public data class Data(
    public val something: Something?,
  ) : Query.Data

  public data class Something(
    public val __typename: String,
    public val id: String,
    /**
     * Synthetic field for inline fragment on Human
     */
    public val onHuman: OnHuman?,
    /**
     * Synthetic field for inline fragment on Droid
     */
    public val onDroid: OnDroid?,
  )

  public data class OnHuman(
    public val name: String,
  )

  public data class OnDroid(
    public val model_name: String,
  )
)
```
It is very difficult to handle all possible cases on multiple nullable fields. 

If this code was generated with responseBased model, it will be a sealed interface. All possible cases can be handled easily with `when` statement. 

Is it possible to generate code as a sealed interface like below? And will it against the concept of "operationBased".
```kotlin
sealed interface Something {
    val __typename: String
    val id: String

    data class OnHuman(
        override val __typename: String,
        override val id: String,
        val name: String,
   ): Something

    data class OnDroid(
        override val __typename: String,
        override val id: String,
        val model_name: String,
    ): Something

    data class OnOtherwise(
        override val __typename: String,
        override val id: String
    ): Something
}
```

martinbonnin:
Unfortunately, there's no silver bullet there. It's either `responseBased` and potentially long compilation times or `operationBased` and less type information. 

If we really wanted to, we could look into changing the codegen mode per-query/per-field but that will add significant complexity to the codebase (which is already quite complex) so I'm not sure there's a lot to be gained from that.

You can read [this blog post](https://www.apollographql.com/blog/mobile/kotlin/the-story-behind-apollo-kotlin-3-codegen/) for more details. If you want to share your schema and queries, I can take a looks at potential ways to reorganize things but to be honest, I'm not sure that'd be enough. You might be able to save a factor 2 or 3 but that's nothing compared to something that grows exponentially.

[...]

PR #4370, opened 2022-09-06
OperationBasedWithInterfaces

chalermpong:
Hi
I just finished implementing #4272. 

**Here is explanation how the new code working:**

Most of the changes are in `OperationBasedModelGroupBuilder.kt` 

`fun buildField`, 

I calculate all possible shapes (without shapes from `GQLFragmentSpread` field), filter out shapes that won't happened, and attached that information inside `OperationField`.

`OperationField.toModelGroup()`, 
1. If there are more than one possible shape, the result IrModelGroup will be a polymorphic model group which consists of the followings classes: 
- `AbstractionModel` (generated as sealed interface)
- `ShapedModel` (generated one class for one possible shape), 
- `FallbackModel` (for unknown __typename)
2. In `fun toProperty`, if the target field has more than one possible shape, the field's model type will be changed into `AbstractionModel` class.
3. When generating `ShapedModel`, I make nullable field become non-null field by evaluating whether field condition always true after considering shape's possible type.

**TODO**: Java codegen is currently not working because polymorphic adapter is not currently support in Java. I think I will ignore shape information in `OperationField.toModelGroup()` when target language = Java, and leave Java codegen untouched.


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
```diff
@@ -174,16 +175,24 @@ internal class OperationBasedModelGroupBuilder(
...
     }
 
     val selfPath = path + "." + info.responseName
-
+    val shapes = shapes(
+        schema,
+        allFragmentDefinitions,
+        selections.map { it.selection }.filter { it !is GQLFragmentSpread },
+        rawTypeName
+    ).filter { it.possibleTypes.isNotEmpty() }
```
martinbonnin, line 195: Out of curiosity, is that filtering anything? By definition, I would expect `shapes()` to only return shapes with non -empty possibleTypes

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
```diff
@@ -174,16 +175,24 @@ internal class OperationBasedModelGroupBuilder(
...
+          rawTypeName = rawTypeName
       )
     }
 
     val selfPath = path + "." + info.responseName
-
+    val shapes = shapes(
+        schema,
+        allFragmentDefinitions,
+        selections.map { it.selection }.filter { it !is GQLFragmentSpread },
```
martinbonnin, line 193: Can you elaborate on why you're filtering out fragment spreads? Feels like it'd be useful as well?  For an example, using the current code, 

```graphql
query TestQuery {
  hero {
    name
    ...DroidDetails
  }
}
```

Generates the below

```kotlin
   public data class Hero(
      public val __typename: String,
      public val name: String,
      public val droidDetails: DroidDetails?,
```

Feels like this is a place where `IHero` would have been handy as well?

martinbonnin (commented):
General comments:

* This will need to be behind an `opt-in` flag (`codegenModels.set("operationBased2")`) to avoid breaking all the users that currently rely on `operationBased` codegen
* Given that Java doesn't have sealed interfaces neither exhaustive `when` statements, I'd say Java codegen is largely out of scope

Now to the meat of the code, I think we'll need to clarify a bit what the expectations are in terms of codegen. As stated [here](https://github.com/apollographql/apollo-kotlin/pull/4370/files#r963666984), feels like this PR is missing a few interesting use cases. 

What would you say about adding a `design-docs/operationBased2` (until we find a better name for the feature 😅 ) with a few example of queries and the expected (pseudo) generated code?

On my side, I'll go over the compiler and add a bunch of KDoc/comments to try to better define some of the most common terms like shape/polymorphic/id/path/etc... How does that sound?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: Yes, it returns empty possibleTypes for `something` field below:

```graphql
query TestOperation {
  something {
    id
    ... on Human {
      name
    }
    ... on ThingWithName {
      name
    }
  }
}

# schema
type Query {
  something: ThingWithID
}
interface ThingWithID {
  id: ID!
}

interface ThingWithName {
  name: String!
}
type Human implements ThingWithID & ThingWithName {
  id: ID!
  name: String!
}
```
`shapes()` will return result below:
```
0 = {Shape@2727} Shape(typeSet=[ThingWithID], possibleTypes=[Droid])
1 = {Shape@2728} Shape(typeSet=[ThingWithID, Human], possibleTypes=[])
2 = {Shape@2729} Shape(typeSet=[ThingWithID, ThingWithName], possibleTypes=[Pet])
3 = {Shape@2730} Shape(typeSet=[ThingWithID, Human, ThingWithName], possibleTypes=[Human])
```

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: At first, I planed to include shapes from fragment spreads. 
```graphql
query TestOperation {
  something {
    ... ThingWithIDFragment
  }
}
fragment ThingWithIDFragment on ThingWithID {
  id
  ... on Human {
    name
  }
  ... on Droid {
    build_name
    model_name
  }
}
```
The operation above, there will be 2 shaped models which `thingWithIDFragment` field already shaped according possible type
```kotlin
public sealed interface ISomething {
  public val __typename: String,
  public val id: String,
  public val thingWithIDFragment: ThingWithIDFragment
}
// Shaped model: Human
public data class HumanSomething {
  public val __typename: String,
  public val id: String,
  public val thingWithIDFragment: HumanThingWithIDFragment // Shaped model of ThingWithIDFragment
}
// Shaped model: Droid
public data class DroidSomething {
  public val __typename: String,
  public val id: String,
  public val thingWithIDFragment: DroidThingWithIDFragment // Shaped model of ThingWithIDFragment
}
// Fallback model
public data class Something {
  public val __typename: String,
  public val id: String,
  public val thingWithIDFragment: ThingWithIDFragment
}
```
There are two reasons that made me decide not to do this.
After trying to implement, I found a problem with generated response adapter for `HumanSomething` and `DroidSomething`

Firstly, I couldn't make ImplementationAdapter work.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
martinbonnin: Right. That's for interfaces in responseBased codegen (https://github.com/apollographql/apollo-kotlin/blob/071039c17dd42b21ecbfe081f5d1b69363407fbd/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/ResponseBasedModelGroupBuilder.kt#L409)

We might want to add a parameter to avoid returning those.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: Also, I don't normally write operation like this:
```graphql
query TestQuery {
  hero {
    name
    ...DroidDetails
  }
}
```
I would rather write it like below. I think it is not too difficult to get `IHero` if user wants to use it.
```graphql
query TestQuery {
  hero {
    name
    ... on Droid {
      ...DroidDetails
    }
  }
}
```

martinbonnin: > The operation above, there will be 2 shaped models which thingWithIDFragment field already shaped according possible type

I think "shapes()" is not what we want here. It's a bad name that caught me off guard [earlier](https://github.com/apollographql/apollo-kotlin/pull/4370#discussion_r963947262) actually. The `shapes()` function as in `shapes.kt` not only returns "user shapes" which might have no possible types but it also traverses the whole selection to find all possible type.

In your example: 

```graphql
query TestOperation {
  something {
    ... ThingWithIDFragment
  }
}
```

`something` is of type `ThingWithID` and only contains one fragment spread whose type condition is also `ThingWithID`. So the user should always see 

```kotlin
data class Something {
  public val __typename: String,
  public val id: String,
  public val thingWithIDFragment: ThingWithIDFragment # non-nullable because always satified
}
```

My earlier example is different though:

```graphql
query TestQuery {
  hero {
    name
    ...DroidDetails
  }
}
```

`DroidDetails` will not always be returned so we'd need 2 classes and 1 interface:

```kotlin
sealed interface IHero {
  val name: String
  val DroidDetails: DroidDetails? = null
}
class CharacterHero: IHero {
  val name: String
  val DroidDetails: DroidDetails? = null
}
class DroidHero: IHero {
  val name: String
  val DroidDetails: DroidDetails
}
```

So we need to enumerate all possible combinations of fragments. Which is different from the `shapes` from `shapes.kt` which is itself different from the Json shapes

martinbonnin: > I would rather write it like below.

That's more verbose. It's fine if people prefer this form but I wouldn't make that a recommendation.

chalermpong: Anyway, it is just my opinion. If you think it's good to have, we can discuss more how to do it.

chalermpong:
> This will need to be behind an opt-in flag (codegenModels.set("operationBased2")) to avoid breaking all the users that currently rely on operationBased codegen
OK.

> Given that Java doesn't have sealed interfaces neither exhaustive when statements, I'd say Java codegen is largely out of scope
OK

> I think we'll need to clarify a bit what the expectations are in terms of codegen.
I would say it generates models according to interesting types (as I described [here](https://github.com/apollographql/apollo-kotlin/issues/4272#issuecomment-1188599288)). Do you think `inline fragment` is more official word for my "interesting types"?

> What would you say about adding a design-docs/operationBased2 (until we find a better name for the feature 😅 ) with a few example of queries and the expected (pseudo) generated code?
Yes, I would be really happy to do that.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: > I think "shapes()" is not what we want here. It's a bad name that caught me off guard https://github.com/apollographql/apollo-kotlin/pull/4370#discussion_r963947262 actually. The shapes() function as in shapes.kt not only returns "user shapes" which might have no possible types but it also traverses the whole selection to find all possible type.

Do you mean we should not apply shape to fragment spread? but only apply non-null to fragment when its type condition always satisfied? 

Like this example...
```graphql
query TestOperation {
  something {
    ... ThingWithNameFragment
  }
}
fragment ThingWithNameFragment on ThingWithName {
  name
  ... on Human {
    friend {
      id
    }
  }
}
# schema
type Query {
  something: ThingWithID
}
interface ThingWithID {
  id: ID!
}

interface ThingWithName {
  name: String!
}
type Human implements ThingWithID & ThingWithName {
  id: ID!
  name: String!
  friend: ThingWithID!
}
type Pet implements ThingWithID & ThingWithName {
  id: ID!
  name: String!
}
type Droid implements ThingWithID {
  id: ID!
}
```
Operation above will generate code shown below...
``` kotlin
// shapes() output
// 0 = {Shape@2731} Shape(typeSet=[ThingWithID], possibleTypes=[Droid])
// 1 = {Shape@2732} Shape(typeSet=[ThingWithID, ThingWithName], possibleTypes=[Pet])
// 2 = {Shape@2733} Shape(typeSet=[ThingWithID, ThingWithName, Human], possibleTypes=[Human])

sealed interface ISomething {
  val __typename: String
  val thingWithNameFragment: ThingWithNameFragment?
}

// Shape 1
data class ThingWithNameSomething(
  val __typename: String,
  val thingWithNameFragment: ThingWithNameFragment
)
// Shape 2
data class HumanThingWithNameSomething(
  val __typename: String,
  val thingWithNameFragment: ThingWithNameFragment
)

// Fallback model
data class Something(
  val __typename: String,
  val thingWithNameFragment: ThingWithNameFragment?
)
```
Is this what you suggest?

chalermpong: @martinbonnin I revised my code as my previous comment. Please check.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: Should I add a parameter now?

chalermpong:
> This will need to be behind an opt-in flag (codegenModels.set("operationBased2")) to avoid breaking all the users that currently rely on operationBased codegen

Added.

@martinbonnin  I have revised code as your comments. If everything is ok, I will start on design-docs tomorrow.

martinbonnin:
> @martinbonnin I have revised code as your comments. If everything is ok, I will start on design-docs tomorrow.

Thanks! Sorry I didn't have much time to look too much into this this week. Design docs and most importantly examples of how we want the codegen to work would be awesome as it's easier to read and discuss than the actual compiler implementation

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedModelGroupBuilder.kt
chalermpong: @martinbonnin I have the feeling that there is no point to generate `Shape 2` in my [previous comment](https://github.com/apollographql/apollo-kotlin/pull/4370#discussion_r964356393) because it is identical to `Shape1` . 

I think I have to find a way to filter out shapes from inline fragment inside fragment spread.

chalermpong:
Design doc added. But I will revise my code a bit as I described in [comment](https://github.com/apollographql/apollo-kotlin/pull/4370#discussion_r967688486).

martinbonnin:
@chalermpong what about nested fragments?

```graphql
{
  node {
    ... on Animal {
      species
      ... on WarmBlooded {
        temperature
      }
      ... on Pet {
        name
      }
    }
  }
}
```

what do we want in that case?

```kotlin
class Data(val node: INode)
sealed interfaceINode(val onAnimal: OnAnimal?)
class AnimalNode(val onAnimal: OnAnimal)
class OtherNode(val onAnimal: OnAnimal?)
```

So far so good but should `OnAnimal` be a sealed interface too? If not then we're having the same issue where one has to check for null on `WarmBlooded` to determine what to do:

```kotlin
val animal = data.node.onAnimal

if (animal?.onWarmBlooded != null) {
  // do something
}
if (animal?.onPet != null) {
  // do something
}
```

Ideally we'd want something like this:

```kotlin
sealed interface IOnAnimal(val onWarmBlooded: OnWarmBlooded?, val onPet: OnPet?)
class OnWarmBloodedAnimal(val onWarmBlooded: OnWarmBlooded, val onPet: OnPet?)
class OnPetAnimal(val onWarmBlooded: OnWarmBlooded?, val onPet: OnPet)
class OnPetWarmBloodedAnimal(val onWarmBlooded: OnWarmBlooded, val onPet: OnPet)
class OnOtherAnimal(val onWarmBlooded: OnWarmBlooded, val onPet: OnPet)
```

So that we could do 

```kotlin
val onAnimal = data.node.onAnimal

when (animal) {
  is OnWarmBloodedAnimal -> // do something
  is OnPetAnimal -> // do something
  is OnPetWarmBloodedAnimal -> // do something
  is OnOtherAnimal -> // do something
}
```

Writing all of this, sounds like it's a lot of noise. Is it ok to have "interfaces" for top level fields but not for nested fragments?

chalermpong:
@martinbonnin First of all. Thank you for being patient with my issue. I think I am the only one who cannot use `responseBased` codegen and needs this feature.

For `animal` query in your previous comment, here is the current output code from `operationBased2` model:
```kotlin
public data class Data(public val animal: IAnimal?)

// To shorten the code, I will skip val __typename: String which appears in every class
sealed interface IAnimal { val onAnimal: Animal.IOnAnimal }
data class WarmBloodedAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class PetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class WarmBloodedPetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class Animal(val onAnimal: IOnAnimal): IAnimal

public sealed interface IOnAnimal { 
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded?,
   val onPet: OnAnimal.OnPet?
}
data class WarmBloodedOnAnimal(
  val species: String, 
  val onWarmBlooded: OnAnimal.OnWarmBlooded,
  val onPet: OnAnimal.OnPet?
): IOnAnimal
data class PetOnAnimal(
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded?,
  val onPet: OnAnimal.OnPet,
): IOnAnimal
data class WarmBloodedPetOnAnimal(
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded,
  val onPet: OnAnimal.OnPet,
): IOnAnimal
data class OnAnimal(
  val species: String,
  val onWarmBlooded: OnWarmBlooded?,
  val onPet: OnPet?,
): IOnAnimal
```
So we can already do this...
```kotlin
when (val onAnimal = response.data!!.animal!!.onAnimal) {
   is TestOperation.Data.Animal.OnAnimal -> TODO()
   is TestOperation.Data.Animal.PetOnAnimal -> print(onAnimal.onPet.name)
   is TestOperation.Data.Animal.WarmBloodedOnAnimal -> TODO()
   is TestOperation.Data.Animal.WarmBloodedPetOnAnimal -> TODO()
}
```
For the current output code, I don't like `IAnimal` interface and its subclasses because they are all the same. It will do just fine with only `data class Animal`. Do you think I should revise this?
```kotlin
sealed interface IAnimal { val onAnimal: Animal.IOnAnimal }
data class WarmBloodedAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class PetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class WarmBloodedPetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
data class Animal(val onAnimal: IOnAnimal): IAnimal
```

Now, let review `animal` query but this time using named fragment:
```graphql
query TestOperation {
  animal {
    ... AnimalFragment
  }
}

fragment AnimalFragment on Animal {
  species
  ... on WarmBlooded {
    temperature
  }
  ... on Pet {
    name
  }
}
```
Here is the output code:
```kotlin
sealed interface IAnimal { val animalFragment: IAnimalFragment }
data class WarmBloodedAnimal(val animalFragment: IAnimalFragment): IAnimal
data class PetAnimal(val animalFragment: IAnimalFragment): IAnimal
data class WarmBloodedPetAnimal(val animalFragment: IAnimalFragment): IAnimal
data class Animal(val animalFragment: IAnimalFragment): IAnimal

sealed interface IAnimalFragment { 
  val species: String
  val onWarmBlooded: AnimalFragment.OnWarmBlooded?
  val onPet: AnimalFragment.OnPet?
}
data class WarmBloodedAnimalFragment(
  val species: String,
  val onWarmBlooded: AnimalFragment.OnWarmBlooded,
  val onPet: AnimalFragment.OnPet?
): IAnimalFragment
data class PetAnimalFragment(
  val species: String,
  val onWarmBlooded: AnimalFragment.OnWarmBlooded?,
  val onPet: AnimalFragment.OnPet
): IAnimalFragment
data class WarmBloodedPetAnimalFragment(
  val species: String,
  val onWarmBlooded: AnimalFragment.OnWarmBlooded,
  val onPet: AnimalFragment.OnPet
): IAnimalFragment
data class AnimalFragment(
  val species: String,
  val onWarmBlooded: AnimalFragment.OnWarmBlooded?,
  val onPet: AnimalFragment.OnPet?
): IAnimalFragment
```
Again, I don't like `IAnimal` interface and its subclasses. This is why I want to filter out possible shapes from inline fragments inside fragment spread.

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+}
+</pre></td>
+<td><pre lang="kotlin">
+public data class Data(public val animal: IAnimal?)
+
+sealed interface IAnimal { val onCat: Animal.OnCat? }
+data class CatAnimal(val onCat: Animal.OnCat): IAnimal
+data class Animal(val onCat: Animal.OnCat?): IAnimal
```
martinbonnin, line 114: For consistency with responseBased codegen:

```suggestion
sealed interface Animal { val onCat: Animal.OnCat? }
data class CatAnimal(val onCat: Animal.OnCat): Animal
data class OtherAnimal(val onCat: Animal.OnCat?): Animal
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+    ... on Pet {
+      name
+    }
+  }
+}
+</pre></td>
+<td><pre lang="kotlin">
+public data class Data(public val animal: IAnimal?)
+
+// Abstraction Model
```
martinbonnin, line 142:
```suggestion
// Interface Model
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+sealed interface IAnimal { val onAnimal: Animal.IOnAnimal }
+// 3 TypeSet Models
+data class WarmBloodedAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
+data class PetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
+data class WarmBloodedPetAnimal(val onAnimal: Animal.IOnAnimal): IAnimal
+// Fallback Model
+data class Animal(val onAnimal: IOnAnimal): IAnimal
```
martinbonnin, line 149:
```suggestion
sealed interface Animal { 
  val onWarmBlooded: Animal.OnWarmBlooded?
  val onPet: Animal.OnPet?
}
// 3 TypeSet Models
data class WarmBloodedAnimal(
  val onWarmBlooded: Animal.OnWarmBlooded
  val onPet: Animal.OnPet?  
): Animal
data class PetAnimal(
  val onWarmBlooded: Animal.OnWarmBlooded?
  val onPet: Animal.OnPet
): Animal
data class WarmBloodedPetAnimal(
  val onWarmBlooded: Animal.OnWarmBlooded
  val onPet: Animal.OnPet
): Animal
// Fallback Model
data class OtherAnimal(
  val onWarmBlooded: Animal.OnWarmBlooded?
  val onPet: Animal.OnPet?  
): Animal
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+      ... on Pet {
+        name
+      }
+    }
+  }
+}
+</pre></td>
+<td><pre lang="kotlin">
+
+// animal is not generated as polymorphic model because inline fragment ... on Animal is always assignable
```
martinbonnin, line 180:
```suggestion
// animal is not generated as polymorphic model because the `... on Animal` inline fragment type condition is always satisfied
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+data class WarmBloodedPetOnAnimal(
+  val species: String,
+  val onWarmBlooded: OnAnimal.OnWarmBlooded,
+  val onPet: OnAnimal.OnPet,
+): IOnAnimal
+data class OnAnimal(
+  val species: String,
+  val onWarmBlooded: OnWarmBlooded?,
+  val onPet: OnPet?,
+): IOnAnimal
```
martinbonnin, line 209:
```suggestion
public data class Data(public val animal: Animal?)
data class Animal(val onAnimal: OnAnimal)

public sealed interface OnAnimal { 
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded?,
  val onPet: OnAnimal.OnPet?
}
data class WarmBloodedOnAnimal(
  val species: String, 
  val onWarmBlooded: OnAnimal.OnWarmBlooded,
  val onPet: OnAnimal.OnPet?
): OnAnimal
data class PetOnAnimal(
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded?,
  val onPet: OnAnimal.OnPet,
): OnAnimal
data class WarmBloodedPetOnAnimal(
  val species: String,
  val onWarmBlooded: OnAnimal.OnWarmBlooded,
  val onPet: OnAnimal.OnPet,
): OnAnimal
data class OtherOnAnimal(
  val species: String,
  val onWarmBlooded: OnWarmBlooded?,
  val onPet: OnPet?,
): OnAnimal
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+data class WarmBloodedPetAnimalFragment(
+  val species: String,
+  val onWarmBlooded: AnimalFragment.OnWarmBlooded,
+  val onPet: AnimalFragment.OnPet
+): IAnimalFragment
+data class AnimalFragment(
+  val species: String,
+  val onWarmBlooded: AnimalFragment.OnWarmBlooded?,
+  val onPet: AnimalFragment.OnPet?
+): IAnimalFragment
```
martinbonnin, line 265: This is basically the same as example 7 with `OnAnimal` replaced with `AnimalFragment`?

```suggestion
public data class Data(public val animal: Animal?)
data class Animal(val animalFragment: AnimalFragment)

public sealed interface AnimalFragment { 
  val species: String,
  val onWarmBlooded: OnWarmBlooded?,
  val onPet: OnPet?
}
data class WarmBloodedAnimalFragment(
  val species: String, 
  val onWarmBlooded: OnWarmBlooded,
  val onPet:OnPet?
): AnimalFragment
data class PetAnimalFragment(
  val species: String,
  val onWarmBlooded: OnWarmBlooded?,
  val onPet: OnPet,
): AnimalFragment
data class WarmBloodedPetAnimalFragment(
  val species: String,
  val onWarmBlooded: OnWarmBlooded,
  val onPet: OnPet,
): AnimalFragment
data class OtherAnimalFragment(
  val species: String,
  val onWarmBlooded: OnWarmBlooded?,
  val onPet: OnPet?,
): AnimalFragment
```

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+<td><pre lang="kotlin">
+
+// animal is not generated as polymorphic model because inline fragment ... on Animal is always assignable
+public data class Data(public val animal: Animal?)
+data class Animal(val onAnimal: IOnAnimal)
+
+// onAnimal field has two nested inline fragments -> generated as Polymorphic model
+public sealed interface IOnAnimal { 
+  val species: String,
+  val onWarmBlooded: OnAnimal.OnWarmBlooded?,
```
martinbonnin, line 187: I think there's no need to namespace `OnWarmBlooded` here as it will be accessible in the `OnAnimal` scope. `val onWarmBlooded: OnWarmBlooded?` should work (I included it in the suggestion below)

design-docs/operationBased2.md
```diff
@@ -0,0 +1,269 @@
...
+    temperature
+  }
+  ... on Pet {
+    name
+  }
+}
+</pre></td>
+<td><pre lang="kotlin">
+
+// animal is not generated as polymorphic model because there is no inline fragment
```
martinbonnin, line 237: Can you elaborate why you only want this new codegen for inline fragments and not fragment spreads? With something like this:

```graphql
{
  animal {
    ...catFragment
    ...lionFragment
  }
}
```

it feels like the benefits would be the same as with 

```graphql
{
  animal {
    ... on Cat { ... }
    ... on Lion { ... }
  }
}
```

Any reason to only do the transformation for inline fragments?

design-docs/operationBased2.md
chalermpong: OK. 

At first, I thought that it would be more compatible with `operationBased` if I named classes like these. But I was wrong. Your suggestion is more compatible with `operationBased` codegen.

chalermpong:
@martinbonnin I revised code. Now it's ready for review.

martinbonnin:
@chalermpong I created a follow up PR there https://github.com/chalermpong/apollo-kotlin/pull/1/files to investigate. I juste added an integration test that fails on a conflicting declaration. I'll keep investigating and push my work over there. Ultimately, I'll merge in your branch.

martinbonnin:
I've updated the follow up PR with a change that mostly keeps the Ir as-is. The general idea is to:

1. generate nodes for each field, inline fragment and fragment definition just like for `operationBased` models
2. each of these nodes tracks its `parentTypes`, that is all the inline fragments type conditions that were traversed to reach this node + the type of the field itself
3. when generating modelGroups, compute the shapes depending on synthetic fields `parentTypes` (it is renamed to `buckets` here to work without GQLSelection/Ast)
4. if there are more than 1 shape/bucket, generate several model groups. For different models, patch the properties depending on the possible types.

design-docs/Codegen.md
```diff
@@ -6,9 +6,14 @@ For more complex queries, involving [merged fields](https://spec.graphql.org/dra
...
 * **operationBased**: the Kotlin models map the sent operation.
+* **experimental_operationBasedWithInterfaces**: based on **operationBased**, but expose more type information. 
 * **compat**: for compatibility with Apollo Kotlin 2.x
 
-`operationBased` will generate less code but will use more memory and expose less type information. `responseBased` will generate interfaces to access the models in a more polymorphic way. It will also store each merged field exactly once and have more efficient json parsing. That comes at the price of more generated code.
+`responseBased` will generate interfaces to access the models in a more polymorphic way. It will also store each merged field exactly once and have more efficient json parsing. That comes at the price of more generated code. 
+
+`operationBased` will generate less code but will use more memory and expose less type information.
+
+`experimental_operationBasedWithInterfaces` same as `operationBased` but will generate interfaces to access model similar to `responseBased`.
```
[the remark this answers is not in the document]
martinbonnin: I went with 

```
generates interfaces for selection sets that contain fragments to make it easier to use `when` statements.
```

Because it's very far from `responseBased` codegen actualy.

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/OperationBasedWithInterfacesModelGroupBuilder.kt
```diff
@@ -0,0 +1,435 @@
...
+import com.apollographql.apollo3.compiler.decapitalizeFirstLetter
+import com.apollographql.apollo3.compiler.escapeKotlinReservedWord
+
+/**
+ * Very similar to [OperationBasedModelGroupBuilder] except:
+ * - it doesn't support compat
+ * - it doesn't support `@include` and `@defer`
+ * - it generates interfaces for polymorphic selection sets
+ *
+ * For that last point, it starts just like [OperationBasedModelGroupBuilder] and once it has a list of nodes
```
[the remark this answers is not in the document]
martinbonnin: Classic 🤦

BoD (approved):
👍
Thanks for the contribution!

martinbonnin:
Thanks for the PR 💙

PR #4383, opened 2022-09-12
remove IllegalStateException from IDLE / TERMINATED state in terminate and responseCallback 

WilliamsDHI:
@martinbonnin Here is the PR in reference to this [thread](https://github.com/apollographql/apollo-kotlin/blob/92aa7b56b1bb268e95fdf93811dc61821d090bd6/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java#L366).

### DIFF
--- a/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
+++ b/apollo-runtime/src/main/java/com/apollographql/apollo/internal/RealApolloCall.java
@@ -363,8 +363,7 @@ public final class RealApolloCall<T> implements ApolloQueryCall<T>, ApolloMutati
         return Optional.fromNullable(originalCallback.get());
       case IDLE:
       case TERMINATED:
-        throw new IllegalStateException(
-            CallState.IllegalStateMessage.forCurrentState(state.get()).expected(ACTIVE, CANCELED));
+        return Optional.absent();
       default:
         throw new IllegalStateException("Unknown state");
     }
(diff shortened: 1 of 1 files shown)


martinbonnin:
Hi 👋 thanks for opening this. As this more of a workaround than an actual fix for the root cause, I'm a bit reluctant to merging this as-is, even if 2.x is in maintainance mode. What I can offer is to publish a `2.5.13-absent-callback` version with the fix to an alternate [repsy.io](https://repsy.io/) repository. This way you can test it and let us know how that works for you before we decide what to do with the `release-2.x` branch? How does that sound?

WilliamsDHI:
That sounds great and I think would serve our purposes. We are really just looking for a work-around until we have bandwidth to update and refactor our app to Apollo 3.

martinbonnin:
Sounds good. It's getting a bit late here but I'll do that first thing tomorrow.

WilliamsDHI:
As expected, no issues with this absent-callback version. We're using in develop now and will be pushing a hotfix soon to production with this version. Thanks again.

martinbonnin:
@WilliamsDHI any news from production? Is everything looking good?

WilliamsDHI:
@martinbonnin Because the crash has stabilized, we are not putting out a hot fix. This update will now go in production in a few more weeks with a different release. But so far things are working well in our test environment.

WilliamsDHI:
@martinbonnin The 2.5.14-absent-callback version has been in production for a week now and is looking very good. Please keep it up and we'll continue to use it. If the merge or another fix is made in the future to Apollo 2 then we'll update our version again.

martinbonnin:
Thanks for the update :). I'll merge this PR and it'll be part of the next 2.x release. 
Thanks again for all the investigations!

martinbonnin:
This is fixed in https://github.com/apollographql/apollo-kotlin/pull/4383 and will be released in 2.5.14

martinbonnin:
Not much activity these days in the `release-2.x` branch. I guess we could do one this week unless something comes up. I'll double check internally if we have anything else planned for 2.x

PR #4386, opened 2022-09-15
Update LoggingInterceptor documentation

harmittaa:
Added docs for LoggingInterceptor's log Level, as those were missing.

Used [OkHttp HttpLoggingInterceptor.kt](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L49) for reference.

Sample data from [apollo-kotlin-tutorial](https://github.com/apollographql/apollo-kotlin-tutorial).

`./gradlew dokkaHtml` executes locally without issues.

Sample screenshots:
image
Screenshot 2022-09-15 at 7 55 14

### DIFF
--- a/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/LoggingInterceptor.kt
+++ b/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/LoggingInterceptor.kt
@@ -10,7 +10,7 @@ import kotlin.jvm.JvmOverloads
 /**
  * An interceptor that logs requests and responses.
  *
- * @param level the level of logging. Caution: when uploading files, setting this to [Level.BODY] will cause the files to
+ * @param [level] the level of logging. Caution: when uploading files, setting this to [Level.BODY] will cause the files to
  * be fully loaded into memory, which may cause OutOfMemoryErrors.
  */
 class LoggingInterceptor(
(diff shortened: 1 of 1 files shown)


apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/LoggingInterceptor.kt
```diff
@@ -10,7 +10,7 @@ import kotlin.jvm.JvmOverloads
 /**
  * An interceptor that logs requests and responses.
  *
- * @param level the level of logging. Caution: when uploading files, setting this to [Level.BODY] will cause the files to
+ * @param [level] the level of logging. Caution: when uploading files, setting this to [Level.BODY] will cause the files to
```
BoD, line 13: [TIL](https://kotlinlang.org/docs/kotlin-doc.html#throws-class-exception-class)!

BoD (approved):
Thanks a lot! 🙏

martinbonnin (approved):

apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/LoggingInterceptor.kt
martinbonnin: So it's technically not changing anything, right?

harmittaa: > So it's technically not changing anything, right?

Nothing is changing with the brackets, both syntaxes are equal as is the effect on the docs. It just adds a bit of readability here when overviewing the docs inside the source code.

martinbonnin: Sounds good, thanks!

martinbonnin:
Thanks!

PR #4387, opened 2022-09-16
Defer with router tests

BoD:
Adds `@defer` tests running against the [Apollo Router](https://www.apollographql.com/docs/router/).

### DIFF
--- a/apollo-ast/src/main/resources/builtins.graphqls
+++ b/apollo-ast/src/main/resources/builtins.graphqls
@@ -135,5 +135,5 @@ directive @deprecated(
 
 directive @defer(
   label: String
-  if: Boolean
+  if: Boolean! = true
 ) on FRAGMENT_SPREAD | INLINE_FRAGMENT
--- a/tests/defer/router/subgraphs/computers/computers.js
+++ b/tests/defer/router/subgraphs/computers/computers.js
@@ -15,9 +15,14 @@ const resolvers = {
             return computers;
         },
         computer: (_, args, context) => {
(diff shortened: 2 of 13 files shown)


apollo-ast/src/main/resources/builtins.graphqls
```diff
@@ -135,5 +135,5 @@ directive @deprecated(
 
 directive @defer(
   label: String
-  if: Boolean
+  if: Boolean! = true
```
BoD, line 138: Not related to this PR, but noticed this was not up to date with the latest spec.

martinbonnin, line 138: Nit: technically a breaking change but I guess it's ok 👍

martinbonnin (approved):

tests/defer/router/subgraphs/computers/computers.js
```diff
@@ -0,0 +1,35 @@
...
+        computer: (_, args, context) => {
+            return computers.find(p => p.id === args.id);
+        }
+    },
+    Computer: {
+        errorField: (_, args, context) => {
+            throw new Error("Error field");
+        },
+        nonNullErrorField: (_, args, context) => {
+            throw new Error("Error field");
```
alessbell, line 26: Nit: in the reference implementation tests, the `nonNullErrorField` [attempts to return `null`](https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/execution/__tests__/defer-test.ts#L56-L58)

PR #4401, opened 2022-09-19
More compiler wording

martinbonnin:
Ongoing work to document the compiler a bit more:
- typeInScope -> parentType: because this is what the spec uses. I find it a bit misling because "parent" could refer to the "parent" field or to the "parent" selection set but consistency with the spec is better here.
- typeDefinitionInScope -> parentTypeDefinition: same reason
- leafType -> rawType: a leaf type could be understood as a non-composite type (enum or scalar). We actually use this in some places in our codebase. Using `rawType` makes it clear this is for a named type without any non-null/list wrapper type

Edit: didn't realize I had `id` -> `path` changes in there... removing them for now
Edit2: they should be all gone now


BoD (approved):
👍

PR #4402, opened 2022-09-19
Fix data builders + multi module

martinbonnin:
`__Schema` was generated twice. Many thanks to @agrosner for catching this 🙏 

Also use the same logic for `Pagination`


BoD (approved):
👍

Issue #4213, opened 2022-06-20
Java codegen: Add option to generate GraphQL enums as Java enums (opposite of Kotlin `sealedClassesForEnumsMatching`)


stengvac:
**Summary**
Enum declared in graphql schema is generated differently for Java and Kotlin changing with `generateKotlinModels.set(false)`. Is this by design?

**Version**
3.3.2

**Description**
I would expect enum in schema generated as `enum class` for Kotlin - done. Also as `enum` for Java - now enum is generad as class with static fields.

Currently:
```
public class JobState {
  public static EnumType type = new EnumType("JobState");

  public static JobState RUNNING = new JobState("RUNNING");

  public static JobState SUCCEEDED = new JobState("SUCCEEDED");

  public static JobState FAILED = new JobState("FAILED");

  public String rawValue;
 //rest ommited
}
```

martinbonnin:
Hi 👋 

The reason GraphQL enums are  mapped to Java classes (and not Java enums) is  to allow access to the `rawValue` that is not known at compile time for new enum values. Imagine your backend introduces a new `PAUSED` JobState. With a Java enum, the best we can do is map that to an `UNKNOWN` enum. But that doesn't tell if it's `PAUSED` or something else.

Now Kotlin has another parameter named `sealedClassesForEnumsMatching` to opt-in that behaviour but it was never implemented for Java for lack of time. I can see how that's confusing when comparing both codegens. We can certainly add this option when focusing on [the Java runtime](https://github.com/apollographql/apollo-kotlin/issues/3694) but there's a lot on the todo list before we can tackle this topic. Feel free to send a pull request though if you're intereseted and we'll look into it!

martinbonnin:
I've updated the title of this issue to make the expected outcome clearer. Let us know if you look into it. If not, we'll tackle this as part of the larger "Java Runtime" effort.

stengvac:
Hello, sorry for late response. 

I will try to have a look. Will see if I can manage to do it :)

martinbonnin:
Thanks 💜 

[...]

PR #4404, opened 2022-09-19
Add classesForEnumsMatching codegen option to generate enums as Java enums

BoD:
See #4213


tests/enums/src/main/graphql/java/schema.graphqls
```diff
@@ -0,0 +1,56 @@
...
+
+enum FooEnum {
+  # not renamed in extra.graphqls, will be renamed automatically
+  type,
+}
+
+enum FooClass {
+  # not renamed in extra.graphqls, will be renamed automatically
+  type,
+}
```
martinbonnin, line 56: Do you need to duplicate the schema? Looks like it's the same as the Kotlin one?

BoD: No, you're right! Updated.

martinbonnin (approved):

PR #4409, opened 2022-09-21
Build scripts: remove generated project and version catalog accessors

martinbonnin:
I've seen multiple occasions where the `R8` task was run over and over again making local development quite slow. After investigation, looks like it's due to typesafe accessors (see https://github.com/gradle/gradle/issues/22095). Not only projects but version catalogs too it seems. So it's a bit sad but this PR disables them. Because it's one thing less to generate/compile, it should also make the build a bit faster.

Also fix a couple of Gradle warnings in `apollo-gradle-plugin`


apollo-gradle-plugin/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 plugins {
   id("org.jetbrains.kotlin.jvm")
-  id("apollo.library")
   id("java-gradle-plugin")
+  id("apollo.library")
```
martinbonnin, line 4: "apollo.library" needs to be after "java-gradle-plugin" or it creates 2 overlapping maven publications

apollo-gradle-plugin/src/test/files/gradle/settings.gradle
```diff
@@ -3,7 +3,7 @@ rootProject.name = "testProject"
 dependencyResolutionManagement {
   versionCatalogs {
     libs {
-      from(files("../../../gradle/libs.versions.toml"))
+      from(files("../../../gradle/libraries.toml"))
```
martinbonnin, line 6: Because we can't add `golatac` before the `plugins {}` block is parsed, plugins can't use `golatac` and the Gradle tests are still using the generated accessors.

build-logic/build.gradle.kts
```diff
@@ -4,48 +4,50 @@ import org.jetbrains.kotlin.samWithReceiver.gradle.SamWithReceiverGradleSubplugi
...
+  id("net.mbonnin.golatac").version("0.0.2")
 }
 
 plugins.apply(SamWithReceiverGradleSubplugin::class.java)
 extensions.configure(SamWithReceiverExtension::class.java) {
   annotations(HasImplicitReceiver::class.qualifiedName!!)
 }
 
-group = "com.apollographql.apollo3"
+group = "com.apollographql.apollo3.build"
```
martinbonnin, line 15: This is so that our convention plugins are not taken into account by `checkApolloDuplicates`

BoD (approved):
Sad, but perf is more important! 🚀

PR #4413, opened 2022-09-27
🧽  Make the build output less noisy

martinbonnin:
Keeping the campsite clean. This shut downs a bunch of logs in the build output

### DIFF
--- a/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/KotlinCompiler.kt
+++ b/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/KotlinCompiler.kt
@@ -15,7 +15,6 @@ object KotlinCompiler {
       sources = kotlinFiles
 
       this.allWarningsAsErrors = allWarningsAsErrors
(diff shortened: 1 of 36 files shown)


apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/KotlinCompiler.kt
```diff
@@ -15,8 +15,9 @@ object KotlinCompiler {
       sources = kotlinFiles
 
       this.allWarningsAsErrors = allWarningsAsErrors
+      this.suppressWarnings
```
BoD, line 18:
```suggestion
```

BoD (approved):
✨

apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/KotlinCompiler.kt
martinbonnin: Thanks!

Issue #4398, opened 2022-09-19
Java codegen: nullability annotations


martinbonnin:

BoD:
Done in $4415

PR #4415, opened 2022-09-29
Java codegen: nullability annotations

BoD:
Adds options to output nullability annotations (JetBrains, Android, JSR305) on fields.

See #4398.


apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
```diff
@@ -52,17 +53,17 @@ internal class InputObjectBuilder(
       return this
     } else {
       val builderFields = inputObject.fields.map {
-        context.layout.propertyName(it.name) to context.resolver.resolveIrType(it.type)
+        FieldSpec.builder(context.resolver.resolveIrType(it.type).withoutAnnotations(), context.layout.propertyName(it.name))
```
BoD, line 56: Refactored `Builder` to take a list of `FieldSpec` instead of maps

apollo-compiler/src/test/graphql/com/example/__schema/java/operationBased/__schema/type/ColorInput.java.expected
```diff
@@ -114,44 +114,48 @@ public class ColorInput {
...
      */
     public Builder reviewRefInput(@Nullable ReviewRefInput reviewRefInput) {
-      this.reviewRefInput = Optional.presentIfNotNull(reviewRefInput);
+      this.reviewRefInput = reviewRefInput;
       return this;
     }
 
     public ColorInput build() {
-      return new ColorInput(red, green, blue, enumWithDefaultValue, reviewRefInput);
+      return new ColorInput(Optional.presentIfNotNull(red),
```
BoD, line 154: (Not really related to nullable annotations!). This changes how builders work. I think this way is a bit nicer because previously you would be required to call every method in the builder, otherwise it could lead to NPEs. 
With this change, not calling a method is equivalent to it being absent. WDYT?

apollo-compiler/build.gradle.kts
```diff
@@ -36,6 +32,12 @@ dependencies {
   testImplementation(golatac.lib("truth"))
   testImplementation(golatac.lib("kotlin.test.junit"))
   testImplementation(golatac.lib("google.testparameterinjector"))
+  testImplementation(project(":apollo-api-java")) {
+    because("For Java and Guava Optional adapter use in the Java generated code")
+  }
```
martinbonnin, line 37: I missed this in the last PR but this sounds a bit fishy. Because we relocate everything in the plugin, we must avoid reflection so I don't think we need this? Instead use `JavaClassNames.foo` like for other symbols?

BoD: How about `Java generated code references Java and Guava Optionals`?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/NamedType.kt
```diff
@@ -26,11 +26,13 @@ internal class NamedType(
...
 
 internal fun NamedType.toParameterSpec(context: JavaContext): ParameterSpec {
+  val irType = context.resolver.resolveIrType(type)
   return ParameterSpec
       .builder(
-          context.resolver.resolveIrType(type),
+          irType.withoutAnnotations(),
           context.layout.propertyName(graphQlName),
       )
+      .addAnnotations(irType.annotations)
```
martinbonnin, line 35: Is that a Java8 thing? (https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/ElementType.html#TYPE_USE)

Asked differently: does the generated code require Java8 compatibility? (I think it is fine if that's the case, just trying to wrap my head around the implications there and planning for future Java19 work 🙃 )

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
martinbonnin: Much better 👍 

Nit: should this class be named `BuilderBuilder`? 🙃

apollo-compiler/build.gradle.kts
martinbonnin: Sounds good 👍

apollo-compiler/src/test/graphql/com/example/__schema/java/operationBased/__schema/type/ColorInput.java.expected
martinbonnin: LGTM! Ping @dwbutler since I believe you're one of the first user of the Java operation builders. Does that work for you?

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/JavaResolver.kt
```diff
@@ -91,22 +107,22 @@ internal class JavaResolver(
...
-      }
+        resolveIrType(type.ofType).let { if (wrapNullableFieldsInOptional) unwrapFromOptional(it) else it.withoutAnnotations() }
+      }.addNonNullableAnnotation()
     }
 
     return when (type) {
       is IrNonNullType -> error("") // make the compiler happy, this case is handled as a fast path
       is IrOptionalType -> resolveIrType(type.ofType).boxIfPrimitiveType().wrapInOptional()
-      is IrListType -> ParameterizedTypeName.get(JavaClassNames.List, resolveIrType(type.ofType).boxIfPrimitiveType())
+      is IrListType -> ParameterizedTypeName.get(JavaClassNames.List, resolveIrType(type.ofType).withoutAnnotations().boxIfPrimitiveType())
```
martinbonnin, line 117: I guess this means no `List<@NonNull Integer>`? This is fine for now but newest versions of `org.jetbrains:annotation` should support `TYPE_USE` so it'd be nice to have it at some point. Make a follow up issue?

BoD: TIL! Didn't know this was allowed. Good idea! -> #4417

apollo-compiler/src/test/graphql/com/example/java_jetbrains_annotations/TestOperation.graphql
```diff
@@ -0,0 +1,20 @@
...
+  nonNullableInt(nullableInt: $nullableInt, nonNullableInt: $nonNullableInt, nonNullableIntWithDefault: $nonNullableIntWithDefault, nullableInput: $nullableInput, nonNullableInput: $nonNullableInput, nonNullableInputWithDefault: $nonNullableInputWithDefault)
+  nullableMyType {
+    nullableInt
+    nonNullableInt
+  }
+  nonNullableMyType {
+    nullableInt
+    nonNullableInt
+  }
+}
```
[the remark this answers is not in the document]
BoD: Yes good point!

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/file/InputObjectBuilder.kt
BoD: > should this class be named BuilderBuilder?

Actually, yes I think so! Will rename :)

apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/helpers/NamedType.kt
martinbonnin: Nice. I'd be curious to see what happens trying to compile the generated sources with Java7 ...

martinbonnin: [Tried it](https://github.com/apollographql/apollo-kotlin/tree/compile-against-java-1-7), didn't work:

```
    Compilation produced the following diagnostics:
    build/generated/test/hero_with_review/java/operationBased/com/example/hero_with_review/selections/TestQuerySelections.java:21: error: incompatible types: java.util.List<com.apollographql.apollo3.api.CompiledField> cannot be converted to java.util.List<com.apollographql.apollo3.api.CompiledSelection>
      private static List<CompiledSelection> __createReview = Arrays.asList(
```

martinbonnin: I guess that means `TargetLanguage.JAVA` should be `TargetLanguage.JAVA8`

martinbonnin (approved):
Nice PR 👍 Java codegen is looking lovelier than ever :).

apollo-compiler/src/test/graphql/com/example/__schema/java/operationBased/__schema/type/ColorInput.java.expected
BoD: Actually there's an issue when wanting to pass `null`. Will amend!

martinbonnin (approved):

apollo-compiler/src/test/graphql/com/example/__schema/java/operationBased/__schema/type/ColorInput.java.expected
dwbutler: That behavior makes sense to me!

PR #4419, opened 2022-09-30
Java codegen: Nullability annotations on generics

### LINKED ISSUE #4417: Java codegen: Nullability annotations on generics (e.g. List<@nullable Integer>)

BoD:
Sadly this could only be done with the JetBrains annotations

Resolves #4417


gradle/libraries.toml
```diff
@@ -12,7 +12,7 @@ cache = "2.0.2"
 dokka = "1.7.10"
 guava = "31.1-jre"
 javaPoet = "1.13.0"
-jetbrains-annotations = "13.0"
+jetbrains-annotations = "23.0.0"
```
martinbonnin, line 15: 🙌

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/JavaResolver.kt
```diff
@@ -149,6 +149,15 @@ internal class JavaResolver(
     }
   }
 
+  private fun TypeName.withoutAnnotationsForGenerics(): TypeName {
```
martinbonnin, line 152:
```suggestion
  /**
   * Only keep the annotations that support TYPE_USE targets for use 
   * in generics like `List<@NotNull String>` for an example
   */
  private fun TypeName.filterTypeUseAnnotation(): TypeName {
```

martinbonnin (approved):
That went smoothly 👍

PR #4424, opened 2022-09-30
`apollo-runtime-java`: mutations

BoD:
Also, fix a few generics related warnings

### DIFF
--- a/libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/interceptor/ApolloInterceptor.java
+++ b/libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/interceptor/ApolloInterceptor.java
@@ -1,24 +1,23 @@
 package com.apollographql.apollo3.runtime.java.interceptor;
 
 import com.apollographql.apollo3.api.ApolloRequest;
+import com.apollographql.apollo3.api.Operation;
 import com.apollographql.apollo3.runtime.java.ApolloCallback;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.concurrent.Executor;
-
 /**
- * ApolloInterceptor is responsible for observing and modifying the requests going out and the corresponding responses
- * coming back in. Typical responsibilities include adding or removing headers from the request or response objects,
- * transforming the returned responses from one type to another, etc.
+ * ApolloInterceptor is responsible for observing and modifying the requests going out and the corresponding responses coming back in.
+ * Typical responsibilities include adding or removing headers from the request or response objects, transforming the returned responses
+ * from one type to another, etc.
  */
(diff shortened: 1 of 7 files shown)


libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/interceptor/ApolloInterceptor.java
```diff
@@ -1,24 +1,23 @@
...
 import com.apollographql.apollo3.runtime.java.ApolloCallback;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.concurrent.Executor;
-
 /**
- * ApolloInterceptor is responsible for observing and modifying the requests going out and the corresponding responses
- * coming back in. Typical responsibilities include adding or removing headers from the request or response objects,
- * transforming the returned responses from one type to another, etc.
+ * ApolloInterceptor is responsible for observing and modifying the requests going out and the corresponding responses coming back in.
```
BoD, line 9: Woops it looks like we have different formatter settings :(.  Or maybe we should disable reformatting of Javadocs?

martinbonnin: Not sure I did reformat that. I copy/pasted it from v2 so hopefully it didn't go through the current reformating rules and everything is ok 👍

libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/interceptor/ApolloInterceptor.java
```diff
@@ -1,24 +1,23 @@
...
+ * Typical responsibilities include adding or removing headers from the request or response objects, transforming the returned responses
+ * from one type to another, etc.
  */
-public interface ApolloInterceptor {
+public interface ApolloInterceptor<D extends Operation.Data> {
```
martinbonnin, line 13: I'm curious how come that works? A single `ApolloInterceptor` typically sees different `D` so it cannot really be parametrized?

BoD: Woops you're probably right, I went too fast on this 😅. Currently the interceptors list is created inside `execute`, but later it should be in `ApolloClient` and not specific to one `D`.

BoD: I think it should be good now in 7177271d98a633e1fdc6f42a0ace3d1f1807ddbb

martinbonnin (approved):

Issue #4421, opened 2022-09-30
OPERATION_BASED_WITH_INTERFACES doesn't generate interface model as sealed interface


chalermpong:
### Summary
With `experimental_operationBasedWithInterfaces` codegen model, the interface model is not generated as `sealed interface` when it is inside another single model.

### Version
3.6.1-SNAPSHOT

### Steps to reproduce the behavior
Use this graphql
```graphql
# Schema
interface Animal {
  species: String!
  parent: Animal!
}

query TestOperation {
  animal {
    ... AnimalFragment
  }
}
fragment AnimalFragment on Animal {
  species
  parent {
    species
    ... on Lion {
      temperature
    }
    ... on Cat {
      name
    }
  }
}
```
The output code will be
```kotlin
public data class AnimalFragment(
  public val species: String,
  public val parent: Parent,
) : Fragment.Data {
  public interface Parent { // <-- Parent should be generated as sealed interface
    ...
  }
}
```

### Propose Fix
It is not generated as `sealed interface` because of this code:
```kotlin
// ModelBuilder
private val nestedBuilders = model.modelGroups.flatMap {
  it.models.map {
    ModelBuilder(
        context = context,
        model = it,
        superClassName = null,
        path = path + model.modelName,
        hasSubclassesInSamePackage = hasSubclassesInSamePackage, // <-- nestedBuilders use hasSubclassesInSamePackage from its parent model which is incorrect.
        adaptableWith = null
    )
  }
}
```
`hasSubclassesInSamePackage` should be calculated for each model instead of using parent's value.

chalermpong:
@martinbonnin Right now, `ModelBuilder` apply `hasSubclassesInSamePackage` to nested model groups. I think `hasSubclassesInSamePackage` should be calculateed for each nested model group.
```kotlin
//ModelBuilder.kt
private val nestedBuilders = model.modelGroups.flatMap { modelGroup ->
    
    // Reuse localInheritance from FragmentModelsBuilder.kt
    val localInheritance = modelGroup.models.any { it.isInterface } &&
        modelGroup.models.any { !it.isInterface }

    modelGroup.models.map {
      ModelBuilder(
          context = context,
          model = it,
          superClassName = null,
          path = path + model.modelName,
          hasSubclassesInSamePackage = it.isInterface && localInheritance,
          adaptableWith = null
      )
    }
```
[...]

PR #4427, opened 2022-09-30
Bugfix operationbased with interface not generated interface model as sealed interface

chalermpong:
Fix for #4421


chalermpong:
There is an unwanted side effect on responseBased codegen. Need to revise the solution.

chalermpong:
I revised the solution. Ready for review.

libraries/apollo-compiler/src/test/graphql/com/example/fragment_with_inline_fragment/kotlin/operationBased2/fragment_with_inline_fragment/TestQuery.kt.expected
```diff
@@ -1,124 +0,0 @@
-//
-// AUTO-GENERATED FILE. DO NOT MODIFY.
-//
-// This class was automatically generated by Apollo GraphQL version '$VERSION'.
-//
-package com.example.fragment_with_inline_fragment
```
martinbonnin, line 6: Thanks for removing these 🙏

martinbonnin (approved):
LGTM, thanks for fixing this!

@BoD looking at this, I believe we're missing a few `sealed` interfaces for `responseBased` fragment implementations. Any reason we do not pass `true` for [`FragmentBuilder`](https://github.com/apollographql/apollo-kotlin/blob/05f2a3295517fb7a64c2259c38b453a37f3c4d19/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/FragmentBuilder.kt#L50)?

BoD:
> @BoD looking at this, I believe we're missing a few sealed interfaces for responseBased fragment implementations. Any reason we do not pass true for [FragmentBuilder](https://github.com/apollographql/apollo-kotlin/blob/05f2a3295517fb7a64c2259c38b453a37f3c4d19/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/FragmentBuilder.kt#L50)?

@martinbonnin No, I can't think of a reason, this was probably an oversight. The generated code looks good with `true` here!

martinbonnin:
Alright, merging this and I'll make a follow up PR for `FragmentBuilder`

PR #4434, opened 2022-10-03
Enable sealed interfaces for fragment implementations

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/4427


BoD (approved):
👍

PR #4442, opened 2022-10-06
Java runtime: Subscriptions/WebSockets

BoD:


BoD:
Opening as a draft for now:
- looking for early feedback
- some TODOs left in the code
- more tests are needed
- GraphQLWS and AppSync subprotocols not implemented

libraries/apollo-runtime-java/api/apollo-runtime-java.api
```diff
@@ -60,3 +69,68 @@ public abstract interface class com/apollographql/apollo3/runtime/java/intercept
 	public abstract fun proceed (Lcom/apollographql/apollo3/api/ApolloRequest;Lcom/apollographql/apollo3/runtime/java/ApolloCallback;)V
 }
 
+public class com/apollographql/apollo3/runtime/java/internal/MapUtils {
+	public fun <init> ()V
+	public static fun entry (Ljava/lang/Object;Ljava/lang/Object;)Ljava/util/Map$Entry;
+	public static fun mapOf ([Ljava/util/Map$Entry;)Ljava/util/Map;
```
martinbonnin, line 75: Nitpicking: I have the feeling that `ImmutableMapBuilder` would be somewhat more idiomatic (but not sure). In all cases, can you add `com/apollographql/apollo3/runtime/java/internal/` to the ignored packages of apiDump so that it doesn't appear here?

libraries/apollo-runtime-java/api/apollo-runtime-java.api
```diff
@@ -9,22 +9,31 @@ public abstract interface class com/apollographql/apollo3/runtime/java/ApolloCal
...
+	public fun subscription (Lcom/apollographql/apollo3/api/Subscription;)Lcom/apollographql/apollo3/runtime/java/ApolloCall;
 }
 
 public class com/apollographql/apollo3/runtime/java/ApolloClient$Builder {
 	public fun <init> ()V
 	public fun addCustomScalarAdapter (Lcom/apollographql/apollo3/api/CustomScalarType;Lcom/apollographql/apollo3/api/Adapter;)Lcom/apollographql/apollo3/runtime/java/ApolloClient$Builder;
 	public fun addInterceptor (Lcom/apollographql/apollo3/runtime/java/interceptor/ApolloInterceptor;)Lcom/apollographql/apollo3/runtime/java/ApolloClient$Builder;
 	public fun addInterceptors (Ljava/util/List;)Lcom/apollographql/apollo3/runtime/java/ApolloClient$Builder;
+	public fun addWsHeader (Lcom/apollographql/apollo3/api/http/HttpHeader;)Lcom/apollographql/apollo3/runtime/java/ApolloClient$Builder;
+	public fun addWsHeaders (Ljava/util/List;)Lcom/apollographql/apollo3/runtime/java/ApolloClient$Builder;
```
martinbonnin, line 24: Would it make send to duplicate the Kotlin API here? (with NetworkTransport as an interface, public WebSocketNetworkTransport etc..?). Less cognitive load for us and for the users?

libraries/apollo-runtime-java/api/apollo-runtime-java.api
BoD: Good call!

libraries/apollo-runtime-java/api/apollo-runtime-java.api
BoD: I think probably yes :) Will amend.

martinbonnin: Turns out we have one already: https://github.com/apollographql/apollo-kotlin/blob/05f2a3295517fb7a64c2259c38b453a37f3c4d19/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ImmutableMapBuilder.kt#L6. (I thought it was a JDK thing but it's not, it's coming [from guava](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html) actually). Maybe reuse that? Or is it too verbose?

libraries/apollo-runtime-java/api/apollo-runtime-java.api
martinbonnin: Turns out we have one already: https://github.com/apollographql/apollo-kotlin/blob/05f2a3295517fb7a64c2259c38b453a37f3c4d19/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ImmutableMapBuilder.kt#L6. (I thought it was a JDK thing but it's not, it's coming [from guava](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html) actually). Maybe reuse that? Or is it too verbose?

martinbonnin (approved):

libraries/apollo-runtime-java/api/apollo-runtime-java.api
martinbonnin: Let's do this for v5 😂

martinbonnin (approved):

PR #4445, opened 2022-10-07
Pass protocols to WebSocket constructor in `JSWebSocketEngine`

baconz:
Of note, subscription-transport-ws (I know it is old...) will close any connection that does not include the protocol header.

See: https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/server.ts#L132

@BoD called this out here: https://github.com/apollographql/apollo-kotlin/issues/4339


BoD:
Thanks! 🙏 Looks good to me. Just to be sure, as we don't have many tests on the WebSocket+JS side, did you get hit by this on your project and that fixes it?

BoD (approved):

baconz:
Well, it wasn't working!! It was tough to get this tested because I can't get the project to build on apple silicon. It seems this is a known issue in Gradle-land and they're working on it (https://github.com/gradle/gradle/issues/19140).

Anyway, I got it to build on a Linux box and tested it out. Should all be working with my latest patch.

BoD:
> Well, it wasn't working!!

Thanks a lot for checking! 🙏 

> I can't get the project to build on apple silicon

Oh interesting, we're working on Apple silicon too but I don't think we've tried the toolchain auto-provisioning feature, and just installed a compatible JDK manually.

The fix looks good.

BoD (approved):

BoD:
Sorry for the delay! This will be in the next release, and can be also tried out with the [snapshots](https://github.com/apollographql/apollo-kotlin#snapshots).

baconz:
Sweet thanks for merging! I tried out the snapshot and things are working!

PR #4450, opened 2022-10-10
Move APQ test to the apropriate module

martinbonnin:
Many thanks @BoD for catching this. The good news is that almost no change was needed to move the test


BoD (approved):
💪

PR #4463, opened 2022-10-17
Java runtime: batching

### LINKED ISSUE #4391: Java runtime batching
Add batching support to ApolloJava

BoD:
Close #4391


BoD:
Marking as WIP for now, will introduce an HttpEngine interface, which will induce a refact of this one.

tests/java-client/src/test/java/test/BatchingTest.java
```diff
@@ -0,0 +1,394 @@
...
+  MockServer mockServer;
+  ApolloClient apolloClient;
+  private String mockServerUrl;
+
+  @Before
+  public void before() {
+    mockServer = new MockServer();
+
+    /*
+      Because url doesn't suspend on the JVM, we can just use the return value
```
martinbonnin, line 45: Oh my. Java friendly mockserver should be doable...

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -111,6 +99,11 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
     }
     if (sendNow) {
       executePendingRequests()
+    } else {
+      scope.launch {
+        delay(batchIntervalMillis)
+        executePendingRequests()
+      }
```
martinbonnin, line 106: This changes the behaviour. For a bunch of grouped requests, the latency is now always `batchIntervalMillis` while before it was on average `batchIntervalMillis/2`.

Of course it might be better for other patterns of requests I believe grouped requests is the typical use case, send n requests in a very short timeframe when opening a screen and then get all of them all at once. So not sure we want to change this?

libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/network/http/internal/BatchingHttpInterceptor.java
```diff
@@ -0,0 +1,256 @@
...
+      pendingRequestsSize = pendingRequests.size();
+    }
+    if (pendingRequestsSize >= maxBatchSize) {
+      executePendingRequests();
+    } else {
+      scheduleExecutePendingRequests();
+    }
+
+    try {
+      callback.onResponse(pendingRequest.getResponse());
```
martinbonnin, line 120: This blocks a thread. Could we pass `callback` to `PendingRequest` instead ?

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
BoD: Sorry for the lack of context! I changed this because in the current implementation, we would have a never-ending loop (the `while (true)` above).  I agree that it changes the behavior though. Maybe we can keep the loop but suspend until there's at least one pending request?

martinbonnin: > Maybe we can keep the loop but suspend until there's at least one pending request?

Yup, that'd work!

libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/network/http/internal/BatchingHttpInterceptor.java
```diff
@@ -0,0 +1,256 @@
...
+      if (exception != null) {
+        throw exception;
+      }
+      return response;
+    }
+  }
+
+  @Override public void intercept(@NotNull HttpRequest request, @NotNull HttpInterceptorChain chain, @NotNull HttpCallback callback) {
+    // Batching is enabled by default, unless explicitly disabled
+    String canBeBatchedHeader = HttpKt.get(request.getHeaders(), ExecutionOptions.CAN_BE_BATCHED);
```
martinbonnin, line 93: Java ergonomics of `Http` aren't great. I added it to https://github.com/apollographql/apollo-kotlin/issues/4171

BoD: Good catch!

libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/network/http/internal/BatchingHttpInterceptor.java
BoD: Good call! Simpler and works.

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/BatchingHttpInterceptor.kt
```diff
@@ -66,7 +66,8 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
 ) : HttpInterceptor {
   private val dispatcher = CloseableSingleThreadDispatcher()
   private val scope = CoroutineScope(dispatcher.coroutineDispatcher)
-  private val mutex = Mutex()
+  private val pendingRequestsMutex = Mutex()
+  private val loopMutex = Mutex(locked = true)
```
martinbonnin, line 70: The mutex doesn't need to be locked initially, right? 

Overall, it's a bit weird to use a `Mutex` here, I would have expected a `pthread_cond_t` or some similar MPP equivalent but I guess there's none? 

Also, back to the initial use case of "grouped queries", is this solving the issue?

```
- intercept query1
- unblocks loop immediatly
- query1 gets sent in a batch containing only one query
- intercept query2, query3, query n
- loop is waiting for `batchIntervalMillis`
- query2, query3 and query n are sent batched
```

martinbonnin: The more I think about this, the more I think your initial hunch of a task to be scheduled when `intercept()` is called was the good one, except of a fixed `batchIntervalMillis` delay, it should keep track of time windows and tune it accordingly?

BoD: Your example is right :) 

I went back to the scheduling but with an adjusting time window instead of a fixed one, I think the delay of         `batchIntervalMillis - (currentTimeMillis() % batchIntervalMillis) - 1` makes sense but LMK what you think.

martinbonnin: LGTM 👍

martinbonnin (approved):

PR #4468, opened 2022-10-18
👷Data Builders: make DefaultFakeResolver open and stateless

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/4453

### DIFF
--- a/libraries/apollo-api/api/apollo-api.api
+++ b/libraries/apollo-api/api/apollo-api.api
@@ -471,9 +471,8 @@ public final class com/apollographql/apollo3/api/CustomTypeValue$GraphQLString :
 	public fun <init> (Ljava/lang/String;)V
 }
 
-public final class com/apollographql/apollo3/api/DefaultFakeResolver : com/apollographql/apollo3/api/FakeResolver {
+public class com/apollographql/apollo3/api/DefaultFakeResolver : com/apollographql/apollo3/api/FakeResolver {
 	public fun <init> (Ljava/util/List;)V
-	public final fun getTypes ()Ljava/util/List;
 	public fun resolveLeaf (Lcom/apollographql/apollo3/api/FakeResolverContext;)Ljava/lang/Object;
 	public fun resolveListSize (Lcom/apollographql/apollo3/api/FakeResolverContext;)I
 	public fun resolveMaybeNull (Lcom/apollographql/apollo3/api/FakeResolverContext;)Z
@@ -619,7 +618,7 @@ public abstract interface class com/apollographql/apollo3/api/FakeResolver {
 }
 
 public final class com/apollographql/apollo3/api/FakeResolverContext {
-	public fun <init> (Ljava/util/List;Lcom/apollographql/apollo3/api/CompiledField;)V
+	public final fun getCacheKey ()Ljava/lang/String;
 	public final fun getMergedField ()Lcom/apollographql/apollo3/api/CompiledField;
 	public final fun getPath ()Ljava/util/List;
 }
--- a/libraries/apollo-api/api/apollo-api.api
+++ b/libraries/apollo-api/api/apollo-api.api
@@ -477,6 +477,7 @@ public class com/apollographql/apollo3/api/DefaultFakeResolver : com/apollograph
 	public fun resolveListSize (Lcom/apollographql/apollo3/api/FakeResolverContext;)I
 	public fun resolveMaybeNull (Lcom/apollographql/apollo3/api/FakeResolverContext;)Z
 	public fun resolveTypename (Lcom/apollographql/apollo3/api/FakeResolverContext;)Ljava/lang/String;
+	public fun stableIdForObject (Ljava/util/Map;Lcom/apollographql/apollo3/api/CompiledField;)Ljava/lang/String;
 }
 
 public final class com/apollographql/apollo3/api/DefaultUpload : com/apollographql/apollo3/api/Upload {
@@ -615,10 +616,11 @@ public abstract interface class com/apollographql/apollo3/api/FakeResolver {
 	public abstract fun resolveListSize (Lcom/apollographql/apollo3/api/FakeResolverContext;)I
 	public abstract fun resolveMaybeNull (Lcom/apollographql/apollo3/api/FakeResolverContext;)Z
 	public abstract fun resolveTypename (Lcom/apollographql/apollo3/api/FakeResolverContext;)Ljava/lang/String;
+	public abstract fun stableIdForObject (Ljava/util/Map;Lcom/apollographql/apollo3/api/CompiledField;)Ljava/lang/String;
 }
 
 public final class com/apollographql/apollo3/api/FakeResolverContext {
-	public final fun getCacheKey ()Ljava/lang/String;
+	public final fun getId ()Ljava/lang/String;
 	public final fun getMergedField ()Lcom/apollographql/apollo3/api/CompiledField;
 	public final fun getPath ()Ljava/util/List;
 }
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
@@ -6,8 +6,9 @@ import com.apollographql.apollo3.api.json.MapJsonReader
  * @property path the path to the element being resolved. [path] is a list of [String] or [Int]
  * @property mergedField the field being resolved
  */
-class FakeResolverContext(
+class FakeResolverContext internal constructor(
     val path: List<Any>,
+    val cacheKey: String,
     val mergedField: CompiledField,
 )
 
@@ -92,11 +93,12 @@ private fun buildFakeObject(
     typename: String,
     base: Map<String, Any?>,
     resolver: FakeResolver,
-    customScalarAdapters: CustomScalarAdapters
+    customScalarAdapters: CustomScalarAdapters,
 ): Map<String, Any?> {
   @Suppress("UNCHECKED_CAST")
   return buildFieldOfType(
       emptyList(),
+      "",
       CompiledField.Builder("data", CompiledNotNullType(ObjectType.Builder(typename).build()))
           .selections(selections)
           .build(),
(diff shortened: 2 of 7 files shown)


libraries/apollo-api/api/apollo-api.api
```diff
@@ -619,7 +618,7 @@ public abstract interface class com/apollographql/apollo3/api/FakeResolver {
 }
 
 public final class com/apollographql/apollo3/api/FakeResolverContext {
-	public fun <init> (Ljava/util/List;Lcom/apollographql/apollo3/api/CompiledField;)V
```
martinbonnin, line 622: This and the line above are 2 breaking changes.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -6,8 +6,9 @@ import com.apollographql.apollo3.api.json.MapJsonReader
  * @property path the path to the element being resolved. [path] is a list of [String] or [Int]
  * @property mergedField the field being resolved
  */
-class FakeResolverContext(
+class FakeResolverContext internal constructor(
     val path: List<Any>,
+    val cacheKey: String,
```
martinbonnin, line 11: I'm not a huge fan of adding the concept of cacheKey there but didn't find an alternative. 

GraphQL Java [exposes the parent `map` and path of types](https://www.graphql-java.com/documentation/data-fetching/#the-interesting-parts-of-the-datafetchingenvironment). That's enough to compute the current field parent cache key but it means that:

1. this has to be done for each leaf field
2. this has to be done for each parent up to the root

This is incredibly wasteful. The other option would be to make `FakeResolverContext` into a visitor pattern and it could then maintain its own stack in parallel to the recursion that's happening in `buildFakeObject` but that feels a bit too much details for an implementation.

martinbonnin:
Converted to draft because we need an API to accomodate non-`@typePolicy` usages

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -222,31 +250,25 @@ private fun buildFieldOfNonNullType(
...
+      "Int" -> context.cacheKey.hashCode() % 100
+      "Float" -> (context.cacheKey.hashCode() % 100000).toFloat() / 100.0
+      "Boolean" -> context.cacheKey.hashCode() % 2 == 0
       "String" -> {
         val index = context.path.indexOfLast { it is String }
-        context.path.subList(index, context.path.size).joinToString { it.toPathComponent() }
+        context.path.subList(index, context.path.size).joinToString(separator = "") { it.toPathComponent() }
       }
 
-      "ID" -> context.path.joinToString { it.toString() }
```
davidshepherd7, line 242: Is this deterministic if the same object (as determined by having the same cacheKey) is reached via a different path? Probably rare that this would matter since usually an ID is going to be the cacheKey for an object, but sometimes people might have schemas that include e.g. `user { friend_user_id }` instead of `user { friend { id } }`.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -222,31 +250,25 @@ private fun buildFieldOfNonNullType(
...
-      "Int" -> currentInt++
-      "Float" -> currentFloat++
-      "Boolean" -> (!currentBoolean).also { currentBoolean = it }
+      "Int" -> context.cacheKey.hashCode() % 100
+      "Float" -> (context.cacheKey.hashCode() % 100000).toFloat() / 100.0
+      "Boolean" -> context.cacheKey.hashCode() % 2 == 0
       "String" -> {
         val index = context.path.indexOfLast { it is String }
-        context.path.subList(index, context.path.size).joinToString { it.toPathComponent() }
+        context.path.subList(index, context.path.size).joinToString(separator = "") { it.toPathComponent() }
```
davidshepherd7, line 264: The index only appears in the default string if it's a list of strings right? So this doesn't have the same problem as with ID because a string can't have a cacheKey?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
martinbonnin: Good catch, yea, that's not deterministic. I'll fix this.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
martinbonnin: The important thing there is `context.path.indexOfLast { it is String }`. It will make the return value depend only on the name of the field. Back to the user case, it means it's going to be `"contactMobile"` or `address[0]` and this doesn't depend what path the user is. It's always going to be `"contactMobile"`. 

It'sa bit naive as an implementation. Ultimately I'd like to provide an implementation based on [kotlin-faker](https://serpro69.github.io/kotlin-faker/) but as far as determinism is concerned it should be good.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -48,6 +52,16 @@ interface FakeResolver {
    * @return a concrete type that implements the type in `context.mergedField.type.rawType()`
    */
   fun resolveTypename(context: FakeResolverContext): String
+
+  /**
+   * Use [stableIdForObject] to provide [FakeResolverContext.id]. You can then use [FakeResolverContext.id]
+   * to derive stable values in the resolveXyz() methods above. This way, you're guaranteed that a fake object will have
+   * the same field values no matter it's path in the query.
```
BoD, line 59:
```suggestion
   * the same field values no matter its path in the query.
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -222,31 +237,27 @@ private fun buildFieldOfNonNullType(
...
 }
 
-class DefaultFakeResolver(val types: List<CompiledNamedType>) : FakeResolver {
-  private var currentInt = 0
-  private var currentFloat = 0.0
-  private var currentBoolean = false
-
-  private var impl = mutableMapOf<String, Int>()
+/**
+ *
```
BoD, line 241: Nit: empty KDoc

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -266,14 +277,30 @@ class DefaultFakeResolver(val types: List<CompiledNamedType>) : FakeResolver {
...
+    if (keyFields.isNotEmpty()) {
+      return buildString {
+        append(obj["__typename"].toString())
+        keyFields.forEach {
+          append(obj[it].toString())
+        }
+      }
+    }
+
+    if (obj.containsKey("__stableId")) {
```
BoD, line 299: Should this test be done first? If for some reason one wants to override the stable id even when there's a typepolicy (not sure why though).

BoD (approved):
🚀

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
martinbonnin: Good call, I'll change this 👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
```diff
@@ -222,31 +237,38 @@ private fun buildFieldOfNonNullType(
...
 
   override fun resolveLeaf(context: FakeResolverContext): Any {
     return when (val name = context.mergedField.type.rawType().name) {
-      "Int" -> currentInt++
-      "Float" -> currentFloat++
-      "Boolean" -> (!currentBoolean).also { currentBoolean = it }
+      "Int" -> context.id.hashCode() % 100
+      "Float" -> (context.id.hashCode() % 100000).toFloat() / 100.0
+      "Boolean" -> context.id.hashCode() % 2 == 0
       "String" -> {
```
davidshepherd7, line 263: Did you consider adding `context.id.hashCode()` to the default strings?

That might be nice because it means that the generated strings are always unique to the object, so it might help to avoid some false negatives? e.g. I think [this assertion](https://github.com/apollographql/apollo-kotlin/pull/4468/files#diff-686a647ac7e0eb49657d51dd021f92adfae0b63e9a3540b668fb0e59ffdc507aR267) will always pass as it is now, even if there were two distinct objects there.

But, maybe it has other downsides that I'm not thinking of?

tests/data-builders-kotlin/src/main/graphql/schema.graphqls
```diff
@@ -18,6 +18,17 @@ type Query {
...
+  promo: Promo
+}
+
+type Promo {
+  product: Product!
+}
+
+type Product {
+  name: String!
+  price: Float!Ò
```
davidshepherd7, line 31: Is `Ò` a typo?

martinbonnin: This is 100% a typo. Fix is there: https://github.com/apollographql/apollo-kotlin/pull/4522. Thanks for the heads up!

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/fakeResolver.kt
davidshepherd7: Oh I meant more like `user.name = "name_8756512089"`, but yeah I'm also not sure if it's worth it anyway.

martinbonnin: Ah, I see, makes a lot of sense 👍. I'm a bit reluctant to introduce this now cause it might break existing tests but will keep it in mind for future evolutions.

PR #4471, opened 2022-10-20
Only generate used types when generateSchema is true

martinbonnin:
When using data builders, this was generating builders for all types in the schema, even the unused ones.

This is potentially breaking for anyone relying on unused types (not sure what the use case for this would be?). But there's a simple workaround if it's really needed. I'll start a changelog to make sure we do not forget about it


BoD (approved):
🚀

PR #4472, opened 2022-10-21
Data builders: only generate used fields

martinbonnin:
This is a follow up of this [slack discussion](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1666262315621899)

One of the main motivations behind data builders was to reduce the amount of generated code and make sure it's not subject to [exponential blowup](https://github.com/apollographql/apollo-kotlin/issues/3144).

Data builders generate code that is O(schemaSize) while test builders generate code that is O(sumOfResponseSizeOverAllOperations). In general this should be more predictable but if the schema is large, there is an upfront cost that make the data builders actually generate more code.

This pull request tracks field usages so that only fields that are used in operation are generated in builders.

@eduardb let me know what you think

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
@@ -354,11 +354,19 @@ interface Service {
   val generateApolloMetadata: Property<Boolean>
 
   /**
-   * A list of [Regex] patterns for input/scalar/enum types that should be generated whether they are used by queries/fragments
-   * in this module. When using multiple modules, Apollo Kotlin will generate all the types by default in the root module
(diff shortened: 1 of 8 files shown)


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -354,12 +354,20 @@ interface Service {
   val generateApolloMetadata: Property<Boolean>
 
   /**
-   * A list of [Regex] patterns for input/scalar/enum types that should be generated whether they are used by queries/fragments
-   * in this module. When using multiple modules, Apollo Kotlin will generate all the types by default in the root module
+   * A list of [Regex] patterns matching [schema coordinates](https://github.com/magicmark/graphql-spec/blob/add_field_coordinates/rfcs/SchemaCoordinates.md)
+   * for types and fields should be generated whether they are used by queries/fragments in this module.
```
BoD, line 358: Nit:
```suggestion
   * for types and fields that should be generated whether they are used by queries/fragments in this module or not.
```

BoD (approved):
👍

Issue #4026, opened 2022-04-13
⚙️  Low level compiler APIs


martinbonnin:
It has come up a few times that tweaking the compiler output is desired. This issue is to brainstorm about possible extension points for the compiler and collect use cases.

* Allowing to set import aliases on `FileSpec.Builder`: https://github.com/apollographql/apollo-kotlin/issues/4023
* Add prefixes to the generated classes: https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1649789764514569
* More?

One of the challenges is going to be how to expose that to Gradle. Because every Task input must be serializable, providing code is not easy. `OperationOutputGenerator` for an example uses a small workaround that the user has to manage a version by hand, which is not ideal for code that might change more often.

Another one is that this will certainly require keeping the KotlinPoet classes in the R8 configuration, which means it's more likely to clash with other buildscript dependencies.

BoD:
Another potential use-case: have the generated model implement `Serializable` or `Parcelable` so they can easily be passed around in Intents, or in a [navigation graph](https://developer.android.com/guide/navigation/navigation-pass-data).

martinbonnin:
I don't think the Parcelable use case](https://community.apollographql.com/t/android-jetpack-navigation-passing-arguments-between-fragments-with-codegen-classes/3309)) holds. Serializing a model usually requires some context (variables, scalar adapters, defer labels, etc...) so it's better addressed through serialization using fromJson/toJson instead of implementing Parcelable/Serializable

BoD:
Another potential use case: the ability to make the generated types `internal` or not depending on custom logic (see #4334)

AnkitPatanaik:
Another potential use case:  Adding default values to Apollo Models w/o writing a custom directive

i.e.
```
query MyQuery {
   name 
   nullable_name // this is a String? type
}

// response
{
  name: "ankit",
  nullable_name: null
}

// ideal response
{
  name: "ankit",
  nullable_name: "my_default_name"
}
```

BoD:
This can also be used to add getters/setters in Java models (#4210)

BoD:
[...]

PR #4474, opened 2022-10-24
Compiler hooks

BoD:
Tentative API.
Related to #4026.

### DIFF
--- a/libraries/apollo-compiler/api/apollo-compiler.api
+++ b/libraries/apollo-compiler/api/apollo-compiler.api
@@ -462,25 +462,49 @@ public final class com/apollographql/apollo3/compiler/codegen/kotlin/helpers/Dat
 public abstract interface class com/apollographql/apollo3/compiler/hooks/ApolloCompilerJavaHooks {
 	public abstract fun getVersion ()Ljava/lang/String;
 	public abstract fun overrideResolvedType (Lcom/apollographql/apollo3/compiler/codegen/ResolverKey;Lcom/squareup/javapoet/ClassName;)Lcom/squareup/javapoet/ClassName;
-	public abstract fun postProcessJavaFile (Lcom/squareup/javapoet/JavaFile;)Lcom/squareup/javapoet/JavaFile;
+	public abstract fun postProcessFiles (Ljava/util/Collection;)Ljava/util/Collection;
+}
+
+public final class com/apollographql/apollo3/compiler/hooks/ApolloCompilerJavaHooks$FileInfo {
+	public fun <init> (Lcom/squareup/javapoet/JavaFile;)V
+	public final fun component1 ()Lcom/squareup/javapoet/JavaFile;
+	public final fun copy (Lcom/squareup/javapoet/JavaFile;)Lcom/apollographql/apollo3/compiler/hooks/ApolloCompilerJavaHooks$FileInfo;
+	public static synthetic fun copy$default (Lcom/apollographql/apollo3/compiler/hooks/ApolloCompilerJavaHooks$FileInfo;Lcom/squareup/javapoet/JavaFile;ILjava/lang/Object;)Lcom/apollographql/apollo3/compiler/hooks/ApolloCompilerJavaHooks$FileInfo;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getJavaFile ()Lcom/squareup/javapoet/JavaFile;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
 }
 
 public abstract interface class com/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks {
 	public abstract fun getVersion ()Ljava/lang/String;
 	public abstract fun overrideResolvedType (Lcom/apollographql/apollo3/compiler/codegen/ResolverKey;Lcom/squareup/kotlinpoet/ClassName;)Lcom/squareup/kotlinpoet/ClassName;
-	public abstract fun postProcessFileSpec (Lcom/squareup/kotlinpoet/FileSpec;)Lcom/squareup/kotlinpoet/FileSpec;
+	public abstract fun postProcessFiles (Ljava/util/Collection;)Ljava/util/Collection;
+}
+
+public final class com/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks$FileInfo {
+	public fun <init> (Lcom/squareup/kotlinpoet/FileSpec;Z)V
+	public final fun component1 ()Lcom/squareup/kotlinpoet/FileSpec;
+	public final fun component2 ()Z
+	public final fun copy (Lcom/squareup/kotlinpoet/FileSpec;Z)Lcom/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks$FileInfo;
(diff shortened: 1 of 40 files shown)


libraries/apollo-gradle-plugin/rules.pro
```diff
@@ -13,10 +13,9 @@
 # For debug
 -keepattributes SourceFile,LineNumberTable
 
-# kotlinpoet uses EnumSetOf that makes a reflexive access to "values"
-# https://github.com/square/kotlinpoet/blob/9952ddcd5095a1fd09c86b9fb07faa347a4c04f0/kotlinpoet/src/main/java/com/squareup/kotlinpoet/PropertySpec.kt#L102
--keepclassmembers class com.squareup.kotlinpoet.KModifier {
-    public static **[] values();
+# Keep all of KotlinPoet to allow usage in Apollo compiler hooks
+-keep class com.squareup.kotlinpoet.** {
```
BoD, line 17: Not sure if it is possible to keep everything, but still relocate it?

tests/compiler-hooks/build.gradle.kts
```diff
@@ -0,0 +1,209 @@
...
+              hasTypeName = true
+              propertySpec.toBuilder()
+                  .addModifiers(KModifier.OVERRIDE)
+                  .build()
+            } else {
+              propertySpec
+            }
+          }
+          if (hasTypeName) {
+            // XXX addSuperinterface(ClassName.bestGuess(interfaceName)) doesn't compile for an unknown reason, but this is equivalent
```
BoD, line 163: I have no explanation. Using `addSuperinterface` triggers:

```
Overload resolution ambiguity: 
public final fun addSuperinterface(superinterface: KClass<*>, delegate: CodeBlock = ...): TypeSpec.Builder defined in com.squareup.kotlinpoet.TypeSpec.Builder
public final fun addSuperinterface(superinterface: TypeName, delegate: CodeBlock = ...): TypeSpec.Builder defined in com.squareup.kotlinpoet.TypeSpec.Builder
```

but that doesn't make sense to me. Maybe an issue related to r8?

libraries/apollo-gradle-plugin/rules.pro
martinbonnin: There's `-dontshrink` but it's global (we already have `-dontobfuscate`. There's a question whether we want to expose a relocated API. It's a bit unexpected for users. 
Maybe we could leverage [IsolationMode](https://docs.gradle.org/current/javadoc/org/gradle/workers/IsolationMode.html) to use kotlinpoet in an isolated classloader?

tests/compiler-hooks/build.gradle.kts
martinbonnin: That's because we relocate the stdlib and `KClass` is relocated so the `KClass` from the `addSuperinterface` method is a different one from the one in `build.gradle.kts`

All in all, I think it's ok to have users of this feature rely on the `external` plugin. This was we don't have to keep kotlinpoet and don't risk breaking existing consumers.

martinbonnin:
As explained above, I think it's ok to have this feature only if you're using the external plugin (maybe even detect it with reflection and throw if someone tries to use this with the shadowed plugin?). 

Looking forward, it'd be nice to use the hooks:
* internally for `generateAsInternal`
* externally, with `apollo-compiler-hooks` that exposes `RenamingHooks`, `DefaultToNullHooks`, `AddTypenameInterfaceHooks`

Moving even forward. There's a question whether this is the correct level of abstraction. It's convenient and somewhat optimized because all the `FileSpec` hold in memory but we could also expose kotlin files and apollo metadata, similar to what the Kotlin compiler is doing, and have users use PSI to parse/rewrite the source.

tests/compiler-hooks/build.gradle.kts
BoD: Good catch about KClass!

Ok to make it work with the `external` plugin, I agree that it's acceptable for this somewhat advanced feature

libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -457,6 +458,37 @@ public final class com/apollographql/apollo3/compiler/codegen/kotlin/helpers/Dat
 	public static final fun makeDataClassFromProperties (Lcom/squareup/kotlinpoet/TypeSpec$Builder;Ljava/util/List;)Lcom/squareup/kotlinpoet/TypeSpec$Builder;
 }
 
+public final class com/apollographql/apollo3/compiler/hooks/AddInternalCompilerHooks : com/apollographql/apollo3/compiler/hooks/DefaultApolloCompilerKotlinHooks {
```
martinbonnin, line 461: This should be internal

libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -457,6 +458,37 @@ public final class com/apollographql/apollo3/compiler/codegen/kotlin/helpers/Dat
...
+	public fun getVersion ()Ljava/lang/String;
+	public fun overrideResolvedType (Lcom/apollographql/apollo3/compiler/codegen/ResolverKey;Lcom/squareup/kotlinpoet/ClassName;)Lcom/squareup/kotlinpoet/ClassName;
+	public fun postProcessFileSpec (Lcom/squareup/kotlinpoet/FileSpec;)Lcom/squareup/kotlinpoet/FileSpec;
+}
+
+public abstract class com/apollographql/apollo3/compiler/hooks/DefaultApolloCompilerKotlinHooks : com/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks {
+	public fun <init> ()V
+	public fun overrideResolvedType (Lcom/apollographql/apollo3/compiler/codegen/ResolverKey;Lcom/squareup/kotlinpoet/ClassName;)Lcom/squareup/kotlinpoet/ClassName;
+	public fun postProcessFileSpec (Lcom/squareup/kotlinpoet/FileSpec;)Lcom/squareup/kotlinpoet/FileSpec;
+}
```
martinbonnin, line 490: I think most of that can be internal too?

tests/compiler-hooks/README.md
```diff
@@ -0,0 +1,16 @@
...
+- `DefaultNullValuesHooks` - Adds a default `null` value to data class fields that are nullable
+- `TypeNameInterfaceHooks` - Adds a super interface to models that expose `__typename`
+- `PrefixNamesHooks` - Prefix generated class names with the specified prefix. Shows how to make
+  `postProcessFileSpec` and `overrideResolvedType` work together.
+
+## Gradle plugin
+
+To use `compilerKotlinHooks` you must use `id("com.apollographql.apollo3.external")` instead of the
+usual `id("com.apollographql.apollo3")`. This is because this one includes the full dependency to KotlinPoet that the
+hooks api depends on.
```
martinbonnin, line 16:
```suggestion
To use `compilerKotlinHooks` you must use `id("com.apollographql.apollo3.external")` instead of the
usual `id("com.apollographql.apollo3")`. This is because the external plugin doesn't relocate the KotlinPoet dependency that the hooks api depends on.
```

martinbonnin: Can we make the plugin throw if we detect that someone is using `shadowed` + `hooks`?

libraries/apollo-compiler/api/apollo-compiler.api
martinbonnin: > I meant for DefaultApolloCompilerKotlinHooks to be public as an easy way to implement the hooks  

Fair enough. 

>can't be internal but can be @ApolloInternal 

👍

martinbonnin (approved):

PR #4476, opened 2022-10-26
Rename `generateModelBuilder` to `generateModelBuilders` and add test

martinbonnin:
For consistency with other options (`generateDataBuilders`, `generateOptionalOperationVariables`, `generateKotlinModels` etc.. that all use a plural)


BoD (approved):
👍

PR #4485, opened 2022-11-02
Java runtime: make ApolloClient implement Closeable

BoD:
Also unrelated: make `BackpressureStrategy.BUFFER` the default in Rx adapters

### DIFF
--- a/tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt
+++ b/tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt
@@ -9,6 +9,7 @@ import com.apollographql.apollo3.api.ExecutionOptions.Companion.CAN_BE_BATCHED
 import com.apollographql.apollo3.api.json.jsonReader
 import com.apollographql.apollo3.mockserver.MockServer
 import com.apollographql.apollo3.mockserver.enqueue
+import com.apollographql.apollo3.mpp.currentTimeMillis
 import com.apollographql.apollo3.testing.internal.runTest
 import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
@@ -64,6 +65,7 @@ class QueryBatchingTest {
         .serverUrl(mockServer.url())
         .httpBatching(batchIntervalMillis = 300)
         .build()
+    alignWithCurrentTime(300)
 
     val result1 = async {
       apolloClient.query(GetLaunchQuery()).execute()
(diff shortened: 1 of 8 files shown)


martinbonnin (approved):

tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt
```diff
@@ -256,4 +261,9 @@ class QueryBatchingTest {
     assertFalse(request.headers.keys.contains("query1+query2-different-value"))
     assertFalse(request.headers.keys.contains(CAN_BE_BATCHED))
   }
+
+  private suspend fun alignWithCurrentTime(batchIntervalMillis: Int) {
+    // Pending requests in BatchingHttpInterceptor are scheduled with a delay that depends on the current time. Align with it to ensure the tests are reproducible
+    delay(batchIntervalMillis - currentTimeMillis() % batchIntervalMillis - 1)
```
martinbonnin, line 267: This sounds fishy that we have to duplicate the interceptor logic. Could we write the test differently so that it doesn't need to know the implementation details of the interceptor?

BoD: Hmmm not sure, as the implementation now depends on the current time, so the test can't assume to know when the pending requests will be executed. On the other hand, inside the interceptor we could align with its creation time to help testing.

martinbonnin: Put it differently: this was already an issue before your changes, right? 
Then I guess it's ok. I was afraid this was the result of a behaviour change but looks like it's not?

BoD: It was not an issue previously because the loop would tick every `batchIntervalMillis`, offset by the instantiation time (so: starting at the same time as the test). Now it still "ticks" on `batchIntervalMillis`, but the offset is based on the current time 😅. So I've offset the test with a delay to align, but I agree that it's a bit odd.

Instead we could do this in the Interceptor to match the previous behavior and have predictable tests:
```diff
+ private val creationTime = currentTimeMillis()
...
- delay(batchIntervalMillis - (currentTimeMillis() % batchIntervalMillis) - 1)
+ delay(batchIntervalMillis - ((currentTimeMillis() - creationTime) % batchIntervalMillis) - 1)
```

What do you think?

martinbonnin (approved):

tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt
martinbonnin: Interesting, I like it 👍

martinbonnin (approved):

PR #4486, opened 2022-11-02
🐘  Use `registerJavaGeneratingTask` 

martinbonnin:
Use `variant.registerJavaGeneratingTask` instead of `kotlin.sourceSets["main"].kotlin.serDir()` so that the generated sources are marked as "generated" and not picked up by lint (by default). 

This doesn't use the new AGP variant APIs just yet since they're incubating and also change the codegen task output dir so it's a potentially more intrusive change. 

Closes https://github.com/apollographql/apollo-kotlin/issues/4478


BoD (approved):
🚀

PR #4494, opened 2022-11-07
Add usedCoordinates configuration and nuse it to automatically compute the used coordinates

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/4039

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateUsedCoordinatesAggregateTask.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateUsedCoordinatesAggregateTask.kt
@@ -1,0 +1,33 @@
+package com.apollographql.apollo3.gradle.internal
+
+import com.apollographql.apollo3.compiler.toUsedCoordinates
+import com.apollographql.apollo3.compiler.writeTo
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
+
+@CacheableTask
+abstract class ApolloGenerateUsedCoordinatesAggregateTask : DefaultTask() {
+  @get:OutputFile
+  @get:Optional
+  abstract val outputFile: RegularFileProperty
+
+  @get:InputFiles
+  @get:PathSensitive(PathSensitivity.RELATIVE)
+  abstract val incomingUsedCoordinates: ConfigurableFileCollection
+
+  @TaskAction
+  fun taskAction() {
+    val coordinates = incomingUsedCoordinates.files.fold(setOf<String>()) { acc, n -> acc + n.toUsedCoordinates() }
+
+    // Merge all files into a single list
+    coordinates.writeTo(outputFile.get().asFile)
+  }
+}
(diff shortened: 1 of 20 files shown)


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -580,7 +732,17 @@ abstract class DefaultApolloExtension(
...
+      task.alwaysGenerateTypesMatching.set(project.provider {
+        val jsonFiles = if (service.usedCoordinates != null) {
+          listOf(service.usedCoordinates!!)
+        } else {
+          usedCoordinatesConsumerConfiguration.files
+        }
+        jsonFiles.map { it.toUsedCoordinates() }
+            .fold(service.alwaysGenerateTypesMatching.getOrElse(emptySet())) { acc, new ->
+              acc + new
+            }
```
BoD, line 744: (Nit again) this is the same as `service.alwaysGenerateTypesMatching.getOrElse(emptySet()) + jsonFiles.map { it.toUsedCoordinates() }`?

tests/multi-module-3/root/README.md
```diff
@@ -0,0 +1,3 @@
+A multi-module repo to test __Schema and Pagination args
```
BoD, line 1:
```suggestion
A multi-module repo to test `apolloUsedCoordinates` / `apolloSchema`
```

BoD (approved):

BoD:
🚀

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateUsedCoordinatesAggregateTask.kt
```diff
@@ -0,0 +1,33 @@
...
+  @get:Optional
+  abstract val outputFile: RegularFileProperty
+
+  @get:InputFiles
+  @get:PathSensitive(PathSensitivity.RELATIVE)
+  abstract val incomingUsedCoordinates: ConfigurableFileCollection
+
+  @TaskAction
+  fun taskAction() {
+    val coordinates = incomingUsedCoordinates.files.fold(setOf<String>()) { acc, n -> acc + n.toUsedCoordinates() }
```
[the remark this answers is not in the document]
martinbonnin: It's not.  `val coordinates = incomingUsedCoordinates.files.map { it.toUsedCoordinates() }` returns a `List<Set<String>>` and not a `Set<String>` like the "fold" version. Let's make a puzzler about it? 😅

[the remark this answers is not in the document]
BoD: oooh right! So it's the same as `incomingUsedCoordinates.files.flatMap{ it.toUsedCoordinates() }.toSet()`, right? 😁 (I'm not saying it's clearer than `fold`)

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
martinbonnin: Same thing. I think that'd return a `Set<Any>` ? Because it would try to add a `Iterable<Set<String>>` to a `Set<String>`

BoD: Makes sense!

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateUsedCoordinatesAggregateTask.kt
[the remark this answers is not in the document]
martinbonnin: Yup, I think that'd work. Although if there are a lot duplicate coordinates, the fold version will (most likely?) allocate a bit less because it builds the set "greedily" while the `flatMap` creates a big list first. Maybe I'm wrong though. Just my hunch about it...

PR #4496, opened 2022-11-07
Suppress warning for deprecated input fields

martinbonnin:
Same as for [enum values](https://github.com/apollographql/apollo-kotlin/pull/4470), suppress warnings in generated code for input fields as the user has no control over them.


BoD (approved):
👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/InputAdapter.kt
```diff
@@ -30,6 +31,11 @@ internal fun List<NamedType>.inputAdapterTypeSpec(
       .addSuperinterface(KotlinSymbols.Adapter.parameterizedBy(adaptedTypeName))
       .addFunction(notImplementedFromResponseFunSpec(adaptedTypeName))
       .addFunction(writeToResponseFunSpec(context, adaptedTypeName))
+      .apply {
+        if (this@inputAdapterTypeSpec.any { it.deprecationReason != null }) {
+          addAnnotation(suppressDeprecationAnnotationSpec)
+        }
```
BoD, line 37: Would it make sense to also add OptIn if any field is marked as such, like in #4470?

martinbonnin: Yep, good catch 👍

martinbonnin: done

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -590,7 +590,7 @@ interface Service {
    * - AnnotationTarget.PROPERTY
    * - AnnotationTarget.CLASS
    *
-   * Default: "com.apollographql.apollo3.annotations.ApolloRequiresOptIn"
+   * Default: "none"
```
martinbonnin, line 593: I think this was an oversight?

BoD: Oops yes, good catch!

BoD (approved):

libraries/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt
```diff
@@ -125,28 +125,9 @@ class CodegenTest {
...
-              "input_object_variable_and_argument",
-              "mutation_create_review",
-              "mutation_create_review_semantic_naming",
-              "named_fragment_inside_inline_fragment",
-              "nested_conditional_inline",
-              "optional",
-              "root_query_inline_fragment",
-              "union_inline_fragments",
-              "unique_type_name",
-              "variable_default_value",
```
martinbonnin, line 149: No more warnings 🙌 (this was already the case before, I just removed this).

I left "deprecated_merged_field" and "deprecation" because it's user-actionable. If the queries are modified, the warnings go away

PR #4509, opened 2022-11-14
Introduce HttpFetchPolicyContext

martinbonnin:
- add a test and mention in the migration guide
- introduce HttpFetchPolicyContext

This postpones the moment when the `HttpFetchPolicy` is turned into a `HTTP header`  allowing to set a default `HttpFetchPolicy` on the `ApolloClient.Builder`:

```kotlin
        // Default to NetworkOnly
        .httpFetchPolicy(HttpFetchPolicy.NetworkOnly)
        .serverUrl(mockServer.url())
        .httpCache(dir, Long.MAX_VALUE)
        .build()

val response = apolloClient.query(GetRandomQuery())
          // it's possible to set HTTP headers without losing the default HttpFetchPolicy
          .addHttpHeader("foo", "bar")
          .execute()        
```
See https://github.com/apollographql/apollo-kotlin/issues/4497

### DIFF
--- a/docs/source/migration/3.0.mdx
+++ b/docs/source/migration/3.0.mdx
@@ -626,7 +626,7 @@ val response = apolloClient.query(request)
                       .execute()
 ```
 
-Do make it consistent with the normalized cache, the default `httpFetchPolicy` is now `HttpFetchPolicy.CacheFirst`. To keep the same behaviour as 2.0, use `HttpFetchPolicy.NetworkOnly`.
+To make it consistent with the normalized cache, the default `httpFetchPolicy` is now `HttpFetchPolicy.CacheFirst`. To keep the same behaviour as 2.0, use `HttpFetchPolicy.NetworkOnly`.
 
 ### Optional values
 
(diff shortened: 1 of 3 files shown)


docs/source/migration/3.0.mdx
```diff
@@ -626,6 +626,8 @@ val response = apolloClient.query(request)
                       .execute()
 ```
 
+Do make it consistent with the normalized cache, the default `httpFetchPolicy` is now `HttpFetchPolicy.CacheFirst`. To keep the same behaviour as 2.0, use `HttpFetchPolicy.NetworkOnly`.
```
BoD, line 629:
```suggestion
To make it consistent with the normalized cache, the default `httpFetchPolicy` is now `HttpFetchPolicy.CacheFirst`. To keep the same behaviour as 2.0, use `HttpFetchPolicy.NetworkOnly`.
```

BoD (approved):
👍

docs/source/migration/3.0.mdx
martinbonnin: Thanks!

PR #4513, opened 2022-11-16
Kotlin-1.8.0

martinbonnin:
Update Kotlin Gradle Plugin to 1.8.0. Shouldn't change anything for JVM consumers because of

```
      apiVersion = "1.5"
      languageVersion = "1.5"
      jvmTarget = "1.8"
```

Native/JS consumers will probably need to use 1.8


BoD (approved):
🚀

PR #4514, opened 2022-11-17
🐘 Build scripts: centralized repositories declaration

martinbonnin:
Centralize all repositories declaration in `gradle/repositories.gradle.kts`. This way, it's easier to add/remove `mavenLocal` or or Kotlin EAP repos.


libraries/apollo-gradle-plugin/testProjects/java-only/build.gradle.kts
```diff
@@ -1,13 +1,11 @@
-import com.apollographql.apollo3.gradle.api.ApolloExtension
+apollo
```
BoD, line 1:
```suggestion
```

BoD (approved):
✨

Issue #4523, opened 2022-11-21
generateApolloSources fails with NoSuchMethodError after 3.6.2 -> 3.7.1 update


MyDogTom:
### Summary
After update to 3.7.1 `myModule:generateApolloSources` task fails with 
```
Caused by: java.lang.NoSuchMethodError: 'com.apollographql.relocated.okio.InputStreamSource com.apollographql.relocated.okio.Okio.source(java.io.File)'
```

On 3.6.2 compilation succeeds.

### Version
3.7.1

### Logs

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':myModule:generateMyServiceApolloSources'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:142)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:140)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:128)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
[...]

PR #4528, opened 2022-11-23
Improve relocation in gradle plugin

BoD:
- Change relocated package to `com.apollographql.apollo3.relocated` to avoid clashes when using v2 in the same project (related to #4523)
- Use `kotlinx.serialization` in `apollo-ast` instead of `moshi` to be able to relocate `apollo-ast` fully while not dropping the adapters

### DIFF
--- a/gradle/repositories.gradle.kts
+++ b/gradle/repositories.gradle.kts
@@ -14,6 +14,7 @@ pluginManagement {
           includeModule("me.champeau.gradle", "japicmp-gradle-plugin")
           includeModule("com.gradle.publish", "plugin-publish-plugin")
           includeModule("com.github.ben-manes", "gradle-versions-plugin")
+          includeModule("org.jetbrains.kotlin.plugin.serialization", "org.jetbrains.kotlin.plugin.serialization.gradle.plugin")
         }
       }
       @Suppress("DEPRECATION")
--- a/libraries/apollo-ast/api/apollo-ast.api
+++ b/libraries/apollo-ast/api/apollo-ast.api
@@ -1226,6 +1226,8 @@ public final class com/apollographql/apollo3/ast/internal/SchemaValidationScopeK
 }
 
 public final class com/apollographql/apollo3/ast/introspection/IntrospectionSchema {
+	public static final field Companion Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Companion;
+	public synthetic fun <init> (ILcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema;Lkotlinx/serialization/internal/SerializationConstructorMarker;)V
 	public fun <init> (Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema;)V
 	public final fun component1 ()Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema;
 	public final fun copy (Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema;)Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema;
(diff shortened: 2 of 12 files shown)


libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -1234,9 +1236,28 @@ public final class com/apollographql/apollo3/ast/introspection/IntrospectionSche
...
+	public fun deserialize (Lkotlinx/serialization/encoding/Decoder;)Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema;
+	public synthetic fun deserialize (Lkotlinx/serialization/encoding/Decoder;)Ljava/lang/Object;
+	public fun getDescriptor ()Lkotlinx/serialization/descriptors/SerialDescriptor;
+	public fun serialize (Lkotlinx/serialization/encoding/Encoder;Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema;)V
+	public synthetic fun serialize (Lkotlinx/serialization/encoding/Encoder;Ljava/lang/Object;)V
+	public fun typeParametersSerializers ()[Lkotlinx/serialization/KSerializer;
+}
+
+public final class com/apollographql/apollo3/ast/introspection/IntrospectionSchema$Companion {
+	public final fun serializer ()Lkotlinx/serialization/KSerializer;
```
martinbonnin, line 1255: Is there an option to hide this from the public API?

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -1,87 +1,94 @@
...
       ) : Type()
 
-      @TypeLabel("INTERFACE")
-      @JsonClass(generateAdapter = true)
+      @SerialName("INTERFACE")
+      @Serializable
       data class Interface(
           override val name: String,
           override val description: String?,
-          val kind: String,
```
martinbonnin, line 60: This is technicaly a breaking change but I think it's ok. I don't think many folks call the constructor outside the serialization.

libraries/apollo-ast/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
   antlr
   id("org.jetbrains.kotlin.jvm")
   id("apollo.library")
-  id("com.google.devtools.ksp")
+  kotlin("plugin.serialization")
```
martinbonnin, line 7: Nit. We're using the `id("")` notation everywhere else

```suggestion
  id("org.jetbrains.kotlin.plugin.serialization")
```

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -147,51 +153,36 @@ data class IntrospectionSchema(
   }
 }
 
-/**
- *
- */
+@OptIn(ExperimentalSerializationApi::class)
+private val json: Json by lazy {
+  Json {
+    ignoreUnknownKeys = true
```
martinbonnin, line 159: Curious why this is needed. We might have to tweak the introspection query if we're fetching too many fields

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -147,51 +153,36 @@ data class IntrospectionSchema(
...
+    classDiscriminator = "kind"
+    explicitNulls = false
+  }
+}
+
+@OptIn(ExperimentalSerializationApi::class)
```
martinbonnin, line 165: I don't think it's ok to use this. If a future version changes the signatures, they will break at runtime.

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/schema_to_introspection.kt
```diff
@@ -56,7 +56,7 @@ private class IntrospectionSchemaBuilder(private val schema: Schema) {
         name = name,
         description = description,
         fields = fields.map { it.toSchemaField() },
-        interfaces = implementsInterfaces.map { IntrospectionSchema.Schema.Type.Interface(kind = "INTERFACE", name = it, description = null, fields = null, possibleTypes = null, interfaces = null) }.ifEmpty { null },
+        interfaces = implementsInterfaces.map { IntrospectionSchema.Schema.Type.Interface(name = it, description = null, fields = null, possibleTypes = null, interfaces = null) }.ifEmpty { null },
```
martinbonnin, line 59: 👍

libraries/apollo-gradle-plugin/rules.pro
```diff
@@ -19,12 +19,31 @@
...
+-if @kotlinx.serialization.Serializable class ** {
+    public static ** INSTANCE;
+}
+-keepclassmembers class <1> {
+    public static <1> INSTANCE;
+    kotlinx.serialization.KSerializer serializer(...);
+}
+# @Serializable and @Polymorphic are used at runtime for polymorphic serialization.
+-keepattributes RuntimeVisibleAnnotations,AnnotationDefault
+# End kotlinx-serialization rules
```
martinbonnin, line 46: I'm curious if you can use explicit serializers to remove most (or all?) the reflection? See also https://github.com/Kotlin/kotlinx.serialization/issues/1348

libraries/apollo-gradle-plugin/rules.pro
```diff
@@ -54,11 +73,10 @@
 -dontusemixedcaseclassnames
 # Keep class names to make debugging easier
 -dontobfuscate
--repackageclasses com.apollographql.relocated
+-repackageclasses com.apollographql.apollo3.relocated
```
martinbonnin, line 76: 👍

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
BoD: It appears that certain fields are always fetched but are used only in some of the type kinds, e.g. `inputFields` only relevant for `InputObject`. 
(Tangent: I never thought about this before, but how come the introspection schema is not "strongly typed"? 😅)

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
BoD: But we bundle and relocate the code, so we should be immune to changes? (OTOH I'm ok to avoid it)

libraries/apollo-gradle-plugin/rules.pro
BoD: Yes, referencing `.serializer()` explicitly works! Good catch 👍
I initially had the impression that calling the generic method was equivalent since it calls `.serializer()`, but it's actually a different one

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
martinbonnin: We relocate in `apollo-gradle-plugin` but some users might use `apollo-ast` outside the Gradle context so I'd feel safer avoid this for the time being.

BoD: That makes sense 👍

libraries/apollo-ast/api/apollo-ast.api
BoD: I asked a question [here](https://github.com/Kotlin/kotlinx.serialization/issues/2108)

martinbonnin (approved):

gradle/repositories.gradle.kts
```diff
@@ -14,6 +14,7 @@ pluginManagement {
           includeModule("me.champeau.gradle", "japicmp-gradle-plugin")
           includeModule("com.gradle.publish", "plugin-publish-plugin")
           includeModule("com.github.ben-manes", "gradle-versions-plugin")
+          includeModule("org.jetbrains.kotlin.plugin.serialization", "org.jetbrains.kotlin.plugin.serialization.gradle.plugin")
```
martinbonnin, line 17: Sorry I missed that one. Do we need this? The serialization plugin should be on maven central these days: https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/plugin/serialization/org.jetbrains.kotlin.plugin.serialization.gradle.plugin/

martinbonnin: I removed it again in https://github.com/apollographql/apollo-kotlin/pull/4531/files, see what breaks 👀

BoD: Version 1.6.10 (which we use during idea sync) is not on Maven central, but now I'm no longer sure this is needed because I removed it on my machine and it still works 😅

martinbonnin: Maybe you have configuration cache enabled 😅 . Ok I'll add it again in #4531 with a comment to remove it at some point

PR #4531, opened 2022-11-24
Do not lie in comments

martinbonnin:
Turns out only `settings.gradle.kts` have a special handling of `pluginManagement {}` so we can do listOf(...) here


BoD (approved):
😅

PR #4534, opened 2022-11-25
Make `canBeBatched` and `httpHeaders` orthogonal

martinbonnin:
Similar to https://github.com/apollographql/apollo-kotlin/pull/4509, store `canBeBatched` separately from HTTP headers until the very last moment to avoid them overwriting each other.


BoD (approved):
🎯

PR #4537, opened 2022-11-28
Fix item wrongly removed from http cache when error in subscriptions

BoD:
Issue encountered [here](https://community.apollographql.com/t/apollo-v3-http-cache-isnt-working-not-able-to-debug-the-caching-process-after-migrating-from-2-x-x-to-3-x-x/5101).

For the http cache, we add 2 interceptors:
1. an `HttpInterceptor` that deals with caching
2. an `ApolloInterceptor` which cleans up the cache if there was an error

The cache key is computed in 1. and used in 2. But 1. is executed for queries and mutations only, whereas 2. is also executed for subscriptions. When a subscription resulted in an error, an arbitrary (the most recent one) cache key would be incorrectly cleared from the cache. The proposed fix is to do this cleanup only for queries and mutations.

### DIFF
--- a/libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/HttpCacheExtensions.kt
+++ b/libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/HttpCacheExtensions.kt
@@ -22,6 +22,7 @@ import com.apollographql.apollo3.network.http.HttpInterceptorChain
 import com.apollographql.apollo3.network.http.HttpNetworkTransport
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import java.io.File
 
@@ -117,16 +118,19 @@ fun ApolloClient.Builder.httpCache(
       )
           .run {
             if (request.operation is Query<*> || request.operation is Mutation<*>) {
-              val cacheKey = synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey.remove(request.requestUuid.toString()) }
               catch { throwable ->
                 // Revert caching of responses with errors
+                val cacheKey = synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey[request.requestUuid.toString()] }
                 cacheKey?.let { cachingHttpInterceptor.cache.remove(it) }
                 throw throwable
               }.onEach { response ->
                 // Revert caching of responses with errors
+                val cacheKey = synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey[request.requestUuid.toString()] }
                 if (response.hasErrors()) {
                   cacheKey?.let { cachingHttpInterceptor.cache.remove(it) }
                 }
+              }.onCompletion {
+                synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey.remove(request.requestUuid.toString()) }
               }
             } else {
               this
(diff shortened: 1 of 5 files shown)


martinbonnin:
The root problem seems to be that `cacheKey` is global to an `ApolloClient` and not a request. There is still a bug if two request run concurrently. They might revert the wrong cache key.

BoD:
Good point! Hmm, we could store the `cacheKey` as a header on the http response, which can be retrieved inside the `.onEach {` part, but not sure how to deal with the `catch {` part 🤔

martinbonnin:
> not sure how to deal with the `catch {` part 🤔

Mmmmm right. The issue is that the cache key is computed at the HTTP layer but we're reading it at the GraphQL layer. 

You could maintain a map `GraphQL request uuid` : `HTTP cache key` in the HTTP interceptor, just after computing it. Then you can read this map in the `catch {}` block.

libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/HttpCacheExtensions.kt
```diff
@@ -72,16 +73,19 @@ fun ApolloClient.Builder.httpCache(
       directory = directory,
       maxSize = maxSize,
   )
-  var cacheKey: String? = null
+  val apolloRequestToCacheKey = mutableMapOf<String, String>()
+  var apolloRequestUuid: Uuid? = null
```
martinbonnin, line 77: This is still racy. This code can be accessed from multiple threads. `requestUuid` needs to be passed down to the interceptor and `apolloRequestToCacheKey` protected by a lock

libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/HttpCacheExtensions.kt
```diff
@@ -74,18 +74,19 @@ fun ApolloClient.Builder.httpCache(
...
-  var apolloRequestUuid: Uuid? = null
   return addHttpInterceptor(object : HttpInterceptor {
     override suspend fun intercept(request: HttpRequest, chain: HttpInterceptorChain): HttpResponse {
       val cacheKey = CachingHttpInterceptor.cacheKey(request)
-      apolloRequestToCacheKey[apolloRequestUuid!!.toString()] = cacheKey
+      val requestUuid = request.headers.valueOf(CachingHttpInterceptor.REQUEST_UUID_HEADER)!!
+      synchronized(apolloRequestToCacheKey) {
+        apolloRequestToCacheKey[requestUuid] = cacheKey
+      }
       return chain.proceed(request.newBuilder().addHeader(CachingHttpInterceptor.CACHE_KEY_HEADER, cacheKey).build())
```
martinbonnin, line 84: Filter out `REQUEST_UUID_HEADER` here so that it's not seen by the backend?

BoD: Good catch!

martinbonnin (approved):

martinbonnin (approved):

sam43:
Hi @BoD,
Hope you are doing well. I have tested the `v3.7.2-SNAPSHOT` version, by creating the `.jar` version. It works perfectly fine, I will be waiting for the next release for the library to integrate into my project which's release is kinda blocked right now! But no worries at all. 
I have a request, As I have discovered the bug can you please add me to the ticket so that I might be a contributor as well, this will boost my GitHub profile even more. Thanks for your support, regards.

BoD:
Hi @sam43, thanks a lot for the feedback, glad that the fix works for you! We've just released v3.7.2 which includes the fix. Thanks again for reporting the issue 🙏

PR #4543, opened 2022-12-01
Build scripts: bump gr8 to 0.7

martinbonnin:
This enables Gradle build cache for GR8 tasks, meaning no more waiting for `shadowR8Jar` 🤞


BoD (approved):
👍

PR #4547, opened 2022-12-02
Use jdk11 for building build-logic

martinbonnin:
This way, the toolchain is set to jdk11 for all builds.

Jdk8 for M1 is not present in some distributions. See https://github.com/apollographql/apollo-kotlin/pull/4492


BoD (approved):
👍

PR #4552, opened 2022-12-02
🐘  Workaround for failing tests failing themselves without details

martinbonnin:
Make it possible to fail tests again! Without this, you'd get the error below

```
Receiver class org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest$Executor$execute$1 does not define 
or inherit an implementation of the resolved method 'abstract void failure(java.lang.Object, 
org.gradle.api.tasks.testing.TestFailure)' of interface org.gradle.api.internal.tasks.testing.TestResultProcessor.
```

There was a breaking change in Gradle APIs. See https://youtrack.jetbrains.com/issue/KT-54634/MPP-Test-Failure-causes-KotlinJvmTestExecutorexecute1-does-not-define-failure#focus=Comments-27-6691278.0-0 for more details


BoD (approved):
🐘

Issue #4557, opened 2022-12-06
"Companion" enum value causes compile error


ddolny:
### Summary

When graphql schema contains enum class with "Companion" value, code does not compile.

### Version

3.7.1

### Steps to reproduce the behavior

1. Create schema with enum that contains "Companion" entry 
2. Try to generate apollo classes.

I've tried to generate sealed class instead of enum, but still companion object is generated and clash of the names is present

### Logs

```
Conflicting declarations: public companion object, enum entry Companion
```

BoD:
Hi 👋

Thank you for reporting this! We'll have a look at escaping it automatically. In the meantime you can already instruct the codegen to name the enum value differently to avoid the conflict: create a file named `extra.graphqls` next to your schema file, and use the `@targetName` directive inside, like this:

```graphql
extend enum MyEnum {
  Companion @targetName(name: "_Companion")
}

```

BoD:
A fix is made to automatically escape `Companion` to `Companion_`. This will be in the next release, in the meantime it can be tried out with the [snapshots](https://github.com/apollographql/apollo-kotlin#snapshots).

BoD:
The fix is now available in v3.7.3.

PR #4558, opened 2022-12-06
Escape "Companion" in enum value names

BoD:
Related to #4557


martinbonnin:
Should it be reported to KotlinPoet? Same as https://github.com/square/kotlinpoet/issues/1183?

martinbonnin:
> Should it be reported to KotlinPoet? Same as https://github.com/square/kotlinpoet/issues/1183?

Nevermind, the issue only happens because we are also generating `Companion`, right? So it's not like it's a general thing 👍 . 
LGTM!

martinbonnin (approved):

PR #4564, opened 2022-12-08
Un-break Gradle configuration cache in multi module cases

BoD:
[This change](https://github.com/apollographql/apollo-kotlin/pull/4527/files?w=1#diff-0a0d324d0e4fd61cf803cac9fb66b12de1fddad8de4338bdfda444e064017079R380) broke the configuration cache by using `project` at execution time which is forbidden. This introduces a field instead.


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateSourcesTask.kt
```diff
@@ -213,6 +213,9 @@ abstract class ApolloGenerateSourcesTask : DefaultTask() {
   @get:Input
   abstract val projectName: Property<String>
 
+  @get:Input
+  abstract val projectPath: Property<String>
```
martinbonnin, line 217: Looks like `projectName` isn't used anymore. Can you remove it?

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateSourcesTask.kt
```diff
@@ -377,7 +380,7 @@ abstract class ApolloGenerateSourcesTask : DefaultTask() {
       ApolloMetadata(
           commonMetadata = if (rememberCommonMetadata) commonMetadata else null,
           compilerMetadata = outputCompilerMetadata,
-          moduleName = project.path,
+          moduleName = projectPath.get(),
```
martinbonnin, line 383: 👍

martinbonnin (changes requested):
I think `projectName` can be removed. Looks good otherwise 👍

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateSourcesTask.kt
BoD: It was still used to be passed to Options.moduleName but that was unused too so I removed it :)

Issue #4566, opened 2022-12-09
Url encoding with GET http method


julienbanse:
### Summary

When ApolloClient is set to use GET http method, generated urls are not correctly encoded for native part.
The object `NSUrl` doesn't accept urls with spaces and cause an exception `kotlin.NullPointerException` in `DefaultHttpEngine`

The crash doesn't happen in android part.

Test with release 3.6.2 and 3.7.x

### Version

3.7.2

### Steps to reproduce the behavior

**gradle part :**

```
apollo {
    service("apolloClient") {
        packageName.set("com.test.api")
        codegenModels.set("responseBased")
        generateSchema.set(true)
        generateApolloMetadata.set(true)
        generateFragmentImplementations.set(true)
    }
}
```
I don't know if it usefull.

**init the client :**

```
ApolloClient
        .Builder()
        .serverUrl(url)
        .httpMethod(HttpMethod.Get)
        .build()
```

and run a request from iOS side

### Logs

```
kotlin.NullPointerException
  at 0  PMApiContent            0x1061a65f3    kfun:kotlin.Throwable#<init>(){} + 99 (/opt/buildAgent/work/5f69639f351c4725/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:29:21)
  at 1  PMApiContent            0x10619e937    kfun:kotlin.Exception#<init>(){} + 91 (/opt/buildAgent/work/5f69639f351c4725/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:21:28)
  at 2  PMApiContent            0x10619ec03    kfun:kotlin.RuntimeException#<init>(){} + 91 (/opt/buildAgent/work/5f69639f351c4725/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:32:28)
  at 3  PMApiContent            0x10619eecf    kfun:kotlin.NullPointerException#<init>(){} + 91 (/opt/buildAgent/work/5f69639f351c4725/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:43:28)
  at 4  PMApiContent            0x1061d2483    ThrowNullPointerException + 131 (/opt/buildAgent/work/5f69639f351c4725/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/native/internal/RuntimeUtils.kt:15:11)
...
```

BoD:
[...]

PR #4567, opened 2022-12-09
Escape spaces when url encoding, for Apple

BoD:
Related to #4566


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -199,7 +199,7 @@ class DefaultHttpRequestComposer(
         }
         append(it.key.urlEncode())
```
martinbonnin, line 200: Should we do it here too?

BoD: 👍 there's no reason not to!

martinbonnin: Well, since this is common code, we're hoping that Android/JS can deal with the '+' sign like iOS 🤞 . If not, we'll have to expect/actual this. I'm not sure what the internet is saying about that. Is it generally accepted?

BoD: I've always assumed `+` was how to treat spaces in URLs but apparently [it's complicated](https://stackoverflow.com/a/47188851/15695) 😅.

Maybe we should use `%20` instead to stay on the safe side? Also I'll move the test to `commonMain` so at least we know that it's not rejected by JVM and JS.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/internal/UrlEncode.kt
```diff
@@ -3,19 +3,16 @@ package com.apollographql.apollo3.api.http.internal
...
  */
-internal fun String.urlEncode(
-    spaceToPlus: Boolean = false
-): String = buildString {
-  this@urlEncode.forEach {
-    when {
-      it in RESERVED_CHARS -> append(it.percentEncode())
-      spaceToPlus && it == ' ' -> append('+')
-      else -> append(it)
+internal fun String.urlEncode(): String = buildString {
```
martinbonnin, line 11: Can you put a link to this PR, the stackoverflow question and the RFC so we can track this later if needed? 

Same thing in `HttpRequest.url`, can you add Javadoc mentionning that it expect a valid url as defined by https://www.rfc-editor.org/rfc/rfc3986#section-2 ?

martinbonnin (approved):

martinbonnin:
Double checking: this change is transparent for Android, JVM and JS users, right? No need to put any mention in the changelog?

BoD:
> Double checking: this change is transparent for Android, JVM and JS users, right? No need to put any mention in the changelog?

Correct ✅. OkHttp and JS already escaped space to `%20` automatically.

PR #4569, opened 2022-12-09
🐘  Make it easier to customize the deployed targets

martinbonnin:
Refactor code so that we can easily activate/deactivate native targets and/or JS


BoD (approved):
👍

PR #4573, opened 2022-12-12
Fix providing linker flags to the Kotlin compiler with KGP 1.8

martinbonnin:
Same as https://github.com/cashapp/sqldelight/pull/3671/files/f5f277367f99aa63224a323b1f33a640904ea4c5

I briefly considered keeping the old code for backward compat with KGP 1.5, 1.6 or so but given this is only used in native, it should be pretty safe to assume users are running latest KGP (they need 1.7 at the moment to consume the artifacts anyway)


BoD (approved):
👍

PR #4588, opened 2022-12-16
Update CONTRIBUTING.md and design docs

martinbonnin:
Removing some of the Builders stuff because it was useful during the transition but it loses some relevance these days. It's still there in the history if needed


BoD (approved):
🧹

PR #4589, opened 2022-12-16
Add GraphQLWsProtocol.Factory.webSocketPayloadComposer

### LINKED ISSUE #4575: Support extensions in graphql-ws subscribe message not only for Apollo Persistent Query
### Use case

At the moment, apollo persistent query can be added to graphql-ws subscribe msg 'extensions' field.

 
```kotlin
interface SubscribeMessage {
  id: '<unique-operation-id>';
  type: 'subscribe';
  payload: {
    operationName?: string | null;
    query: string;
    variables?: Record<string, unknown> | null;
    extensions?: Record<string, unknown> | null;
  };
}
```

Could we add arbitrary extensions map?

For ex, add a subscribeExtensions parameter to GraphQLWsProtocol.Factory?  

```kotlin
class Factory(
      private val connectionPayload: suspend () -> Map<String, Any?>? = { null },
      private val pingIntervalMillis: Long = -1,
      private val pingPayload: Map<String, Any?>? = null,
      private val pongPayload: Map<String, Any?>? = null,
      private val connectionAcknowledgeTimeoutMs: Long = 10_000,
      private val frameType: WsFrameType = WsFrameType.Text,
  )
```

martinbonnin:
Closes #4575

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/GraphQLWsProtocol.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/GraphQLWsProtocol.kt
@@ -5,7 +5,6 @@ import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.*
 import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.Operation
-import com.apollographql.apollo3.api.http.DefaultHttpRequestComposer
 import com.apollographql.apollo3.api.http.DefaultWebSocketPayloadComposer
 import com.apollographql.apollo3.api.http.WebSocketPayloadComposer
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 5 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/GraphQLWsProtocol.kt
```diff
@@ -147,22 +168,75 @@ class GraphQLWsProtocol(
...
    */
-  class Factory(
-      private val connectionPayload: suspend () -> Map<String, Any?>? = { null },
-      private val pingIntervalMillis: Long = -1,
-      private val pingPayload: Map<String, Any?>? = null,
-      private val pongPayload: Map<String, Any?>? = null,
-      private val connectionAcknowledgeTimeoutMs: Long = 10_000,
-      private val frameType: WsFrameType = WsFrameType.Text,
-  ) : WsProtocol.Factory {
+  class Factory constructor() : WsProtocol.Factory {
```
martinbonnin, line 172: I'm considering renaming `WsProtocol.Factory` to `WsProtocol.Builder` at some point since they more or less serve the same purpose

BoD (approved):
👍

Issue #4596, opened 2022-12-22
MultipartReader$Companion.readHeaders throws Unexpected header


ramanandsingh85:
### Version

com.apollographql.apollo3: 3.7.3,  graphql-yoga: 3.1.1

### Summary

Graphql Server 3.x is returning invalid multipart response. Following header is added by the android apollo kotlin sdk which seems the sole cause:

-H "Accept:multipart/mixed; deferSpec=20220824, application/json"

It is clear from logs that five hyphens are appended at the end of response which making multipart rader throw exception:

Unexpected header: -----

### Steps to reproduce the behavior

Use any graphql query with the above version set and it will throw an error.

It was working in android kotlin sdk 2.x

### Logs

```
-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  --> POST https://api.xxx-staging.live/graphql
2022-12-22 17:13:12.382 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  Content-Type: application/json
2022-12-22 17:13:12.382 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  Content-Length: 147
2022-12-22 17:13:12.382 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  X-APOLLO-OPERATION-ID: 530c832b59f59d578fbc1e895e4ff9777ccad79dcb221af71bc961da07f3a281
2022-12-22 17:13:12.382 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  X-APOLLO-OPERATION-NAME: tipItemList
2022-12-22 17:13:12.382 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  Accept: multipart/mixed; deferSpec=20220824, application/json
2022-12-22 17:13:12.383 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  {"operationName":"tipItemList","variables":{},"query":"query tipItemList { tipItemList { amount animation_asset_link created_at id image name } }"}
2022-12-22 17:13:12.383 29892-29946 okhttp.OkHttpClient     staging.xxxx.app                 I  --> END POST (147-byte body)
2022-12-22 17:13:12.816 29892-30129 okhttp.OkHttpClient     staging.xxxx.app                 I  <-- 101 Switching Protocols https://api.xxxshop-staging.live/graphql (1008ms)
[...]

PR #4598, opened 2022-12-23
Allow having an empty last part in multipart

BoD:
Related to #4596.
See also https://github.com/dotansimha/graphql-yoga/issues/2243

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
@@ -109,20 +109,19 @@ class MultipartReader constructor(
     afterBoundaryLoop@ while (true) {
       when (source.select(afterBoundaryOptions)) {
         0 -> {
-          // "\r\n": We've found a new part.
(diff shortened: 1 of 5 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/internal/MultipartReader.kt
```diff
@@ -110,6 +110,14 @@ class MultipartReader constructor(
       when (source.select(afterBoundaryOptions)) {
         0 -> {
           // "\r\n": We've found a new part.
+          if (source.select(Options.of("--$boundary--".encodeUtf8())) == 0) {
```
martinbonnin, line 113: Can this be done at the same time as the other select? Looks like creating `Options` [could be expensive](https://medium.com/@jerzy.chalupski/okio-options-ce8f3ac1584f) so I wouldn't want to create too many?

martinbonnin (commented):
LGTM modulo the select thing. Maybe let's wait a bit see what graphql-yoga has to say about it?

martinbonnin (approved):
It's nice to be [liberal about what we accept](https://en.wikipedia.org/wiki/Robustness_principle#:~:text=In%20computing%2C%20the%20robustness%20principle,liberal%20in%20what%20you%20accept%22.)

Issue #4602, opened 2023-01-03
DiskLruCache is not main-safe - performs file i/o on the current thread.


prashanOS:
### Version

3.7.3

### Summary

Currently, DiskLruCache + HttpCacheExtensions leads to an Android StrictMode DiskReadViolation (which occurs when file I/O is performed on the main thread) when entries are removed from the cache:

Example from HttpCacheExtensions:
`cacheKey?.let { cachingHttpInterceptor.cache.remove(it) }`

DiskLruCache performs file I/O and therefore either needs to internally make sure this work is happening on a background thread or consumers (in this case HttpCacheExtensions) must enforce this. Neither is being done right now.

If you don't have it set up already, would highly recommend turning on StrictMode in CI for your Android project so that you can catch these issues in testing!

### Steps to reproduce the behavior

1. Turn on DiskLruCache.
2. Enable Android StrictMode with disk read violations set to fatal.
3. Make a query that will return a response with errors to trigger cache removal.
4. See crash.

### Logs

```
Process: <omitted>, PID: 19851
java.lang.RuntimeException: StrictMode ThreadPolicy violation
        at android.os.StrictMode$AndroidBlockGuardPolicy.onThreadPolicyViolation(StrictMode.java:1875)
	at android.os.StrictMode$AndroidBlockGuardPolicy.lambda$handleViolationWithTimingAttempt$0$android-os-StrictMode$AndroidBlockGuardPolicy(StrictMode.java:1789)
	at android.os.StrictMode$AndroidBlockGuardPolicy$$ExternalSyntheticLambda1.run(Unknown Source:6)
	at android.os.Handler.handleCallback(Handler.java:942)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7872)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
Caused by: android.os.strictmode.DiskReadViolation
	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1658)
	at libcore.io.BlockGuardOs.open(BlockGuardOs.java:269)
[...]

PR #4606, opened 2023-01-05
Http cache: don't access the disk from the main thread in error case

BoD:
Related to #4602.

Deleting cache entries in the error case was done on the thread calling `execute()` (normally, the main thread). This now happens on the the client's dispatcher (by default, `dispatchers.IO`)


martinbonnin (approved):

martinbonnin:
Good catch 👍

PR #4610, opened 2023-01-06
Add deprecation warnings

martinbonnin:
Add messages during Gradle execution when any of those is used:

- `useVersion2Compat()`
- `generateTestBuilders`
- `useSchemaPackageNameForFragments`
- `codegenModels.set("compat")`

There is no way to disable the warning besides actually migrating, which I think is ok.

### DIFF
--- a/libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
+++ b/libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
@@ -6,6 +6,7 @@ import com.apollographql.apollo3.ApolloCall
 import com.apollographql.apollo3.ApolloClient
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_0_0
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince.Version.v3_7_5
 import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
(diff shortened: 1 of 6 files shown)


BoD (approved):
👍

martinbonnin:
Heads up I'm moving this to draft to put any other future deprecations at the same time. This way we can properly signal a "deprecation" maintenance release in the changelog

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -323,6 +326,8 @@ fun <T> MutableExecutionOptions<T>.doNotStore(doNotStore: Boolean) = addExecutio
  *
  * Default: false
  */
+@Deprecated("Will be removed in v4, where this is the default behavior", ReplaceWith(""))
+@ApolloDeprecatedSince(v3_7_5)
 fun <T> MutableExecutionOptions<T>.emitCacheMisses(emitCacheMisses: Boolean) = addExecutionContext(
```
martinbonnin, line 331: @BoD sorry I missed this but this doesn't work. On v3 replacing with `""` is not a solution. We need to keep this and start deprecating with v4

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -299,6 +302,8 @@ fun <T> MutableExecutionOptions<T>.doNotStore(doNotStore: Boolean) = addExecutio
  *
  * Default: false
  */
+@Deprecated("Will be removed in v4, where this is the default behavior", ReplaceWith(""))
+@ApolloDeprecatedSince(v3_7_5)
 fun <T> MutableExecutionOptions<T>.emitCacheMisses(emitCacheMisses: Boolean) = addExecutionContext(
```
martinbonnin, line 307: Same here

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: Oh yeah you're right deprecating it here is very helpful. No worries!

Issue #4603, opened 2023-01-04
Data builders: generate builders for unknown types


BoD:
### Use case

Currently the data builders are only generated for all types known in the schema at build time.

When testing the case of receiving a new type (server's schema is more recent than the schema used at build time), it can be handy to also have the ability to build unknown types. With the responseBased codegen that would be the `Other` types of the generated sealed interfaces.

### Describe the solution you'd like

Have the codegen generate builders for the interfaces, taking the `__typename` as a parameter.

martinbonnin:
Possible implementation:

```kotlin
    val data = GetNodeQuery.Data {
      // use 'Other' in the name
      // the typename is mandatory because we can't get it from the schema
      node = buildOtherNode("SomeNewType") {
        id = "42"
      }
    }
```

martinbonnin:
This is now available in [version 3.7.4](https://github.com/apollographql/apollo-kotlin/releases/tag/v3.7.4)

leinardi:
Awesome, thanks! :+1:

PR #4612, opened 2023-01-09
Add data builders for unknown interface and union types

BoD:
Related to #4603


.idea/runConfigurations/CodegenTest.xml
```diff
@@ -16,7 +16,7 @@
       </option>
       <option name="taskNames">
         <list>
-          <option value=":libraries:apollo-compiler:test" />
+          <option value=":apollo-compiler:test" />
```
martinbonnin, line 19: 👍

tests/data-builders-kotlin/src/test/kotlin/test/DataBuilderTest.kt
```diff
@@ -80,6 +84,20 @@ class DataBuilderTest {
     assertEquals("Rooooaaarr", data.animal.onLion?.roar)
   }
 
+  @Test
+  fun unknownInterfaceChildTest() {
```
martinbonnin, line 88: Why not `unknownInterfaceChildTest` ?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -147,14 +147,26 @@ internal abstract class CodegenLayout(
     return "${name.capitalizeFirstLetter()}Builder"
   }
 
-  fun mapName(name: String): String {
+  fun unknownBuilderName(name: String): String {
```
martinbonnin, line 150:
```suggestion
  fun otherBuilderName(name: String): String {
```

tests/data-builders-kotlin/src/test/kotlin/test/DataBuilderTest.kt
BoD: Arh naming 😅 - the interface is not unknown, it's its child which is. But for unions this naming doesn't make sense I think (I don't think Cat is a "child" of the Feline union?)

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -147,14 +147,26 @@ internal abstract class CodegenLayout(
...
+    return "Other${name.capitalizeFirstLetter()}Builder"
+  }
+
+  fun objectMapName(name: String): String {
     return "${name.capitalizeFirstLetter()}Map"
   }
 
-  fun builderFunName(name: String): String {
+  fun unknownMapName(name: String): String {
```
martinbonnin, line 158: I don't think we need to distinguish between `OtherAnimalMap` and just `AnimalMap`. 

I would keep "other" for the public API (i.e. the `buildOtherAnimal`) because it's special (it requires typename). For everything else, I'd keep things symmetrical and just use `${graphqlType}Map` (so `AnimalMap` here for an example). Thoughts?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
martinbonnin: Actually, same remark as just below. Just use `builderName` and keep everything symmetrical?

tests/data-builders-kotlin/src/test/kotlin/test/DataBuilderTest.kt
martinbonnin: I see. I usually refer to `implementation`/`super interface` instead of `child`/`parent` for this kind of stuff. (And `member`/`union`):

* `Cat` is an implementation of the `Animal` interface
* `Animal` is a super interface of `Cat`
* `Cat` is a member of the `Feline` union
* `Feline` is the union containing `Cat`

So for the test I guess `otherInterfaceImplementation()` ?

BoD: Yes that makes sense! 👍

martinbonnin (changes requested):
Naming appart, LGTM!

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
martinbonnin: Makes a lot of sense 👍 
Then my only objection there is to rename all `unkown` to `other`

martinbonnin (approved):

martinbonnin:
Cherry-pick to `release-3.x`?

PR #4615, opened 2023-01-10
Import wip IntelliJ plugin module

BoD:
First drop for IntelliJ plugin, work in progress.

(Unrelated) also use latest versions of actions in workflows.


intellij-plugin/.gitignore
```diff
@@ -0,0 +1 @@
+mockJDK
```
martinbonnin, line 1: I never know how multiple .gitignore play together... Can we put this in the main one?

martinbonnin (approved):
🎉

PR #4628, opened 2023-01-11
runtime: Always put the APQ interceptor last in the interceptors list

### LINKED ISSUE #4617: toFlow() and .watch() result in IllegalArgumentException: Flow has more than one element
### Version

3.7.3

### Summary

When using cache and network, both:

- `apollo.query(makeQuery()).fetchPolicy(FetchPolicy.CacheAndNetwork).toFlow()`
- `apollo.query(makeQuery()).fetchPolicy(FetchPolicy.CacheAndNetwork).watch()`

result in `java.lang.IllegalArgumentException: Flow has more than one element`.

If I use `FetchPolicy.NetworkOnly` no exception is thrown and the data is collected as expected. Am I using the `FetchPolicy` feature correctly? I assumed the flow would emit twice, once for the cache and one more time for the network. Likewise I assumed that `.watch()` will continuously emit whenever the cache updates, such as after a mutation in another part of the app that updates any of the data related to the original query.

I'm also noticing that after deleting the app and rerunning it, the sqlite database in the Device File Explorer is populated with data from a very old version of the schema, when the schema and data types were completely different. Could this be related? EDIT: For this cleaning and rebuilding the project fixed it. The flow exception still remains.

martinbonnin:
Closes #4617

### DIFF
--- a/libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/ApolloClient.java
+++ b/libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/ApolloClient.java
@@ -171,6 +171,7 @@ public class ApolloClient implements Closeable {
     private WebSocket.Factory webSocketFactory;
     private Executor executor;
     private List<ApolloInterceptor> interceptors = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


BoD (approved):
👍

PR #4633, opened 2023-01-13
IJ plugin: use Gradle's continuous mode

BoD:
Don't observe GQL files ourselves, instead let Gradle do it.
Also, observe project root changes to dynamically determine if current project is an "Apollo project" or not.


martinbonnin (approved):
🚀

PR #4634, opened 2023-01-16
Update JS dependencies

martinbonnin:
Call yarn again to update the lock file:

```
./gradlew rmbuild && rm -rf kotlin-js-store-1.7 && ./gradlew kotlinNpmInstall
```

This shuts down a [dependabot warning](https://github.com/apollographql/apollo-kotlin/security/dependabot/40). I'm pretty sure that warning was harmless as the CVE was about a denial of service but the vulnerable package (`minimatch`) was only used for tests. So someone could theorically DDOS your dev machine if they had control over it which means that anyways they could DDOS more easily than using the Kotlin/JS test runner.

Anyways, that should remove one warning.


BoD (approved):
👍

PR #4635, opened 2023-01-16
Publish IJ plugin snapshots

BoD:
Publish the IJ plugin to the Maven central snapshots repository.

This should make it available when adding this custom repository in the plugins settings: `https://s01.oss.sonatype.org/content/repositories/snapshots/com/apollographql/apollo-intellij-plugin/<version>-SNAPSHOT/apollo-intellij-plugin-<version>-SNAPSHOT.xml`.

### DIFF
--- a/intellij-plugin/build.gradle.kts
+++ b/intellij-plugin/build.gradle.kts
@@ -15,8 +15,8 @@ plugins {
 
 group = properties("pluginGroup")
 
-// Use the global version defined in the root project
-version = project.findProperty("VERSION_NAME").toString()
+// Use the global version defined in the root project + snapshot suffix if from the CI
+version = properties("VERSION_NAME") + if (System.getenv("COM_APOLLOGRAPHQL_IJ_PLUGIN_SNAPSHOT").toBoolean()) ".${properties("snapshotVersion")}" else ""
 
 repositories {
   mavenCentral()
@@ -41,7 +41,7 @@ intellij {
 
 // Configure Gradle Changelog Plugin - read more: https://github.com/JetBrains/gradle-changelog-plugin
 changelog {
-  version.set(properties("VERSION_NAME"))
+  version.set(project.version.toString())
   groups.set(emptyList())
 }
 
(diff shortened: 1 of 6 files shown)


intellij-plugin/build.gradle.kts
```diff
@@ -148,3 +149,47 @@ tasks.register("downloadMockJdk") {
 tasks.named("test").configure {
   dependsOn("downloadMockJdk")
 }
+
+tasks.register("generateUpdatePluginsXml") {
+  val filePath = "build/publications/updatePlugins.xml"
+  val pluginName = properties("pluginName")
+  val version = properties("VERSION_NAME")
```
martinbonnin, line 156: You need to mark this as input or else the task might be always up-to-date

intellij-plugin/build.gradle.kts
```diff
@@ -148,3 +149,47 @@ tasks.register("downloadMockJdk") {
 tasks.named("test").configure {
   dependsOn("downloadMockJdk")
 }
+
+tasks.register("generateUpdatePluginsXml") {
```
martinbonnin, line 153: Are tasks created like this cacheable by default?

intellij-plugin/build.gradle.kts
```diff
@@ -148,3 +149,47 @@ tasks.register("downloadMockJdk") {
...
+        password = System.getenv("SONATYPE_NEXUS_PASSWORD")
+      }
+    }
+  }
+
+  publications {
+    create<MavenPublication>("default") {
+      artifactId = properties("pluginName")
+      artifact(tasks.named("signPlugin"))
+      artifact(tasks.named("generateUpdatePluginsXml"))
```
martinbonnin, line 192: Shouldn't this publish the plugin .jar as well?

intellij-plugin/build.gradle.kts
BoD: Woops didn't think about the cache, good question, I'll 👀

intellij-plugin/build.gradle.kts
BoD: It looks like this only publishes the .zip (the output of `signPlugin`), which is all that we need?

martinbonnin: Looks like it 👍

intellij-plugin/build.gradle.kts
BoD: Nope, you must [ask for it explicitly](https://docs.gradle.org/current/userguide/build_cache.html#using_the_runtime_api), TIL!

martinbonnin (approved):

Issue #4640, opened 2023-01-18
JavaScript name (name) generated for this declaration clashes with another declaration: val name: String


ScottPierce:
### Version

3.7.1

### Summary

```graphql
mutation CreateCustomer($name: String!, $storeId: Int!) {
    createCustomer(input: {customer: {storeId: $storeId, name: $name}}) {
        customer {
            id
            storeId
            name
        }
    }
}
```

```Kotlin
public data class CreateCustomerMutation(
  public val name: String,
  public val storeId: Int,
) : Mutation<CreateCustomerMutation.Data> {
  public override fun id(): String = OPERATION_ID

  public override fun document(): String = OPERATION_DOCUMENT

  public override fun name(): String = OPERATION_NAME
```

### Steps to reproduce the behavior

Create a mutation with an external parameter with the name `name`. This will conflict with the parameter by the name of `name`.

The solution would be to change the name of the function `name()` to something less common, like `operationName()`, or `mutationName()`, etc. Another alternative would be to set the JsName annotation on the base class: `@JsName("mutationName")`. This will allow the same Kotlin api, but when compiled to JS, would actually use a different function for JS.

### Logs

```
JavaScript name (name) generated for this declaration clashes with another declaration: val name: String
```

BoD:
Thanks for reporting this! Will have a look, in the meantime can you give the parameter a different name?

BoD:
Thanks for the tip about @JsName, I've done just that [here](https://github.com/apollographql/apollo-kotlin/pull/4643).

ScottPierce:
Awesome! Thanks for the fast turn around - I wonder if it makes sense to do the same thing with document() while you are in there, although document is probably less common.

BoD:
Thank you! That's a good point! I've thought about it but figured we should really name them with a reserved name, like `__operation` to avoid any clashes - but didn't want to break the symmetry with `id`, nor change its js name (not sure if that would be a breaking change). On the other hand, since there is an easy workaround, I didn't worry too much about it 😅 Maybe we can rename these fields in v4 (breaking changes allowed).

BoD:
This has been released as part of [v3.7.5](https://github.com/apollographql/apollo-kotlin/releases/tag/v3.7.5)

PR #4643, opened 2023-01-19
Add a @JsName annotation to Operation.name()

BoD:
A fix for #4640.

I was wondering why nobody had this problem before, especially with `id` (a common argument name) - turns out we've had `@JsName` on `id()` since JS was [introduced](https://github.com/apollographql/apollo-kotlin/pull/3208/files?w=1#diff-3dea12462b10a56e86f12d23dcdb8bcd533f438da4cbabc4248c31a127d2bff6R34).


martinbonnin (approved):

ScottPierce:
I wonder if it makes sense to do the same thing with `document()` while you are in there, although document is probably less common.

PR #4644, opened 2023-01-19
Gradle 8.0-rc-2

martinbonnin:
~On hold for https://github.com/JetBrains/gradle-intellij-plugin/issues/1272~


libraries/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/util/TestUtils.kt
```diff
@@ -66,6 +66,15 @@ object TestUtils {
       appendLine("}")
       appendLine()
 
+      append("""
+        java {
+          toolchain { 
+            languageVersion.set(JavaLanguageVersion.of(8))
+          }
+        }
+      """.trimIndent())
```
martinbonnin, line 75: Gradle 8 uses the Gradle JSK as toolchain by default now. See https://slack-chats.kotlinlang.org/t/8496136/i-m-confused-about-https-kotlinlang-org-docs-gradle-configur

PR #4645, opened 2023-01-19
Data builders: support for `@skip` and `@include`

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/4639

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ObjectBuilder.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ObjectBuilder.kt
@@ -9,10 +9,6 @@ abstract class ObjectBuilder(override val customScalarAdapters: CustomScalarAdap
 
   var __typename: String by __fields
 
-  infix fun String.to(value:Any?) {
-    __fields[this] = value
-  }
-
   operator fun set(key: String, value: Any?) {
     __fields[key] = value
   }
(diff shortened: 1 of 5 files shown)


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ObjectBuilder.kt
```diff
@@ -9,6 +9,10 @@ abstract class ObjectBuilder(override val customScalarAdapters: CustomScalarAdap
 
   var __typename: String by __fields
 
+  infix fun String.to(value:Any?) {
```
martinbonnin, line 12: This overrides the default `to` that creates a `Pair`. I think it's ok because it reads really well and is only available in the context of an `ObjectBuilder` but we can brainstorm other names.

BoD: Not a strong preference, but I think I'd rather keep the `this["xxx"]` syntax - I find it slightly less "magic" and, having only 1 way to do it is probably better

martinbonnin: Fair enough 👍 I'll revert to the existing `this["foo"]` syntax

BoD (approved):
👍

PR #4653, opened 2023-01-24
IJ Plugin: compatibility for platform version 2023.1

BoD:


BoD:
Marking as draft for the time being: GraphQL plugin with same upper version is not yet on the Marketplace, therefore the verifier complains.

intellij-plugin/gradle.properties
```diff
@@ -9,7 +9,7 @@ pluginRepositoryUrl=https://github.com/apollographql/apollo-kotlin
 # for insight into build numbers and IntelliJ Platform versions.
 pluginSinceBuild=212
 # This can be kept empty to mean no upper bound, but it's recommended to set it to whatever was tested
-pluginUntilBuild=223.*
+pluginUntilBuild=231.*
```
martinbonnin, line 12: Out of curiosity, what happens if we leave that blank?

BoD: It's possible to leave it empty (no upper bound) but not recommended by the JetBrains team :). The downside is that we'll need to re-publish with each new release of the platform (which I think should happen 3 times a year), for the time being it's not really an issue for us, but could be annoying in the future.

martinbonnin: I see 👀 . Sounds like a good thing to automate...

martinbonnin (approved):

PR #4660, opened 2023-02-01
Split ApolloCompiler in multiple steps

martinbonnin:
This is mostly a refactoring. No behaviour change is intended. This splits the `ApolloCompiler` into multiple steps:

* `buildIrOperations`
* `buildOperationOutput`
* `buildIrSchema`
* `writeJava`/`writeKotlin`

`IrOperations` and `IrSchema` are now different classes so that ultimately we can serialize `IrOperations` and reuse it in multi-module scenarios


BoD (approved):
⚙️ 🚀

PR #4662, opened 2023-02-01
Mark most deprecated symbols with ERROR deprecation

martinbonnin:
This is a separate PR so we can easily cherry-pick it to apply it in the 3.x release branch. This way, users will have an error message instead of a removed symbol


libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -773,15 +773,6 @@ public final class com/apollographql/apollo3/compiler/introspection/Introspectio
...
 
-public final class com/apollographql/apollo3/compiler/introspection/IntrospectionSchemaJsonAdapter : com/squareup/moshi/JsonAdapter {
-	public fun <init> (Lcom/squareup/moshi/Moshi;)V
-	public fun fromJson (Lcom/squareup/moshi/JsonReader;)Lcom/apollographql/apollo3/compiler/introspection/IntrospectionSchema;
-	public synthetic fun fromJson (Lcom/squareup/moshi/JsonReader;)Ljava/lang/Object;
-	public fun toJson (Lcom/squareup/moshi/JsonWriter;Lcom/apollographql/apollo3/compiler/introspection/IntrospectionSchema;)V
-	public synthetic fun toJson (Lcom/squareup/moshi/JsonWriter;Ljava/lang/Object;)V
-	public fun toString ()Ljava/lang/String;
-}
-
```
martinbonnin, line 784: That's binary breaking but I haven't found a way to add the `@Suppress("DEPRECATION_ERROR")` annotation to Moshi generated files.

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -662,16 +647,45 @@ abstract class DefaultApolloExtension(
...
+          customScalarsMapping.put("Date", "java.util.Date")
+          or
+          customScalarsMapping.set(mapOf("Date" to "java.util.Date"))
+          
+          With:
+          
+          mapScalar("Date", "java.util.Date")
+                    
+          If you have several scalars, call mapScalar() several times.
+        """.trimIndent()
```
martinbonnin, line 666: Behaviour change: using this will now trigger an error because from Groovy the deprecation warning/errors are not displayed

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -129,13 +129,34 @@ abstract class DefaultApolloExtension(
...
+
+      if (hasApolloBlock) {
         val packageNameLine = if (defaultService.packageName.isPresent) {
           "packageName.set(\"${defaultService.packageName.get()}\")"
         } else {
           "packageNamesFromFilePaths()"
         }
-        it.logger.warn("""
+        error("""
             Apollo: using the default service is deprecated and will be removed in a future version. Please define your service explicitly:
```
BoD, line 160: I guess the message should be adjusted if it errors:
```suggestion
            Apollo: using the default service is no longer allowed. Please define your service explicitly:
```

BoD (approved):
🧹

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
martinbonnin: Right! I went with

```
Apollo: using the default service isdeprecated. Please define your service explicitly:
```

PR #4665, opened 2023-02-01
Update gradle plugin configuration doc

adubovkin:
The existing configuration results in a gradle error with a cryptic message. After diffing around I realized that `schemaFiles` is not a property but rather a `ConfigurableFileCollection` so we should use a `setFrom` method instead and provide the list of paths


martinbonnin (approved):
Good catch, thanks !

PR #4666, opened 2023-02-02
allWarningsAsErrors 

martinbonnin:
Enable `allWarningsAsErrors` except for:
* the `requires-opt-in-deprecated` integration test that has some warnings in generated code that we can't suppress (we should maybe allow to customize this?)
* the `compile${foo}KotlinMetadata` that fail to find SQLiter libs and other things. Given that HMPP is pretty much under work, I found it easier to ignore them 

That's mostly it. Everything else is warning free at the price of some `@Suppress("UNUSED_VARIABLE")` in the integration tests


BoD (approved):
👍

PR #4670, opened 2023-02-06
IJ plugin: fetch tooling model from project

BoD:
For now the information in the models is not used, but should be the basis to share that with the GraphQL plugin.


intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/gradle/GradleToolingModelService.kt
```diff
@@ -0,0 +1,161 @@
...
+          logw(t, "Couldn't fetch Gradle project model")
+          null
+        } finally {
+          gradleCancellation = null
+        }
+      } ?: return@submit
+
+      // We're only interested in projects that apply the Apollo plugin - and thus have the codegen task registered
+      val allApolloGradleProjects: List<GradleProject> = (rootGradleProject.children + rootGradleProject)
+          .filter { it.tasks.any { it.name == CODEGEN_GRADLE_TASK_NAME } }
```
martinbonnin, line 83: This creates coupling in the unlikely event that we rename "generateApolloSources". Could you ask the model for every project and "just" fail on those that do not have the Apollo plugin installed?

BoD: The issue is that we iterate on subprojects to get the model, and if you have a project with a huge amount of modules, that will be impractical (and unneeded if most of them are not Apollo).
It's a bit unfortunate that there's no way to know what plugins are applied from `GradleProject` directly.

Agreed that the coupling to the task name would best be avoided though.

martinbonnin: >  that will be impractical 

impractical as in "slow"?

BoD: As in "so slow that it's unusable" :) (I'm thinking of projects with hundreds or even thousands of modules).

martinbonnin: 👍

martinbonnin (approved):

PR #4674, opened 2023-02-09
Remove deprecation on connectToAndroidSourceSet because alternative have issues too

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/4669

* `registerJavaGeneratingTask` fails on incremental build in unit tests.
* `connectToAndroidSourceSet` fails to mark generated files as "generated" for lint purposes

Between both issues, it feels better to have something that compiles. This can be overriden from Gradle files if needed.


BoD (approved):
👍

PR #4678, opened 2023-02-09
SchemaDownlader: Update to download deprecated input fields

ndwhelan:
Added `(includeDeprecated: true)` as a params to `inputFields` in the `fragment` portion of the GraphQL query to download a schema.

### Current Behavior

When the `downloadApolloSchema` Gradle task is run, input fields with the `@deprecated` tag **ARE NOT** downloaded.

### Expected Change in Behavior

When the `downloadApolloSchema` Gradle task is run, input fields with the `@deprecated` tag **ARE** downloaded.

### DIFF
--- a/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
+++ b/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
@@ -199,7 +199,7 @@ object SchemaDownloader {
         isDeprecated
         deprecationReason
       }
-      inputFields(includeDeprecated: true {
+      inputFields(includeDeprecated: true) {
         ...InputValue
       }
       interfaces {
@@ -221,6 +221,8 @@ object SchemaDownloader {
       description
       type { ...TypeRef }
       defaultValue
+      isDeprecated
+      deprecationReason
     }
 
     fragment TypeRef on __Type {
--- a/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
+++ b/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
@@ -199,7 +199,7 @@ object SchemaDownloader {
         isDeprecated
         deprecationReason
       }
-      inputFields {
+      inputFields(includeDeprecated: true {
         ...InputValue
       }
       interfaces {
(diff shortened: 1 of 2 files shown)


libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
```diff
@@ -199,7 +199,7 @@ object SchemaDownloader {
         isDeprecated
         deprecationReason
       }
-      inputFields {
+      inputFields(includeDeprecated: true {
```
martinbonnin, line 202: Looks like you're missing a closing parenthesis here

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
```diff
@@ -199,7 +199,7 @@ object SchemaDownloader {
...
         ...InputValue
```
martinbonnin, line 203: You'll also want to update the `InputValue` fragment to include `isDeprecated` and `deprecationReason`

ndwhelan:
Wow, that is embarrassing. Sorry for that. I need to actually pull the code
and do this properly instead of just using the GitHuh editor. Thank you for
the advice.

martinbonnin:
Hi @ndwhelan how can we help here? Want us to get this through the finish line?

ndwhelan:
Apologies. Numerous high priority things came up for me this week. I'd like
to try to finish it this weekend.

If you think it's preferable that you take it on at this point, though, I
certainly would not have any problem with that.

Best,

Nick

martinbonnin:
No worries at all! Take your time, just wanted to make sure this doesn't fall through the cracks.

ndwhelan:
I've updated things from the PR, and updated a test. The tests are passing, but I wasn't really convinced it wouldn't return the whole thing regardless of my query in SchemaDownloader. I want to say it was, because it would fail if I put bad syntax in, but I'm not 100% sure. Any feedback on the test or improving that would be appreciated. I should be able to resolve it promptly if necessary.

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
ndwhelan: Done.

martinbonnin (approved):
This all looks very good to me, Thanks! 🙏

BoD (approved):
👍

BoD:
This has been released as part of [v3.7.5](https://github.com/apollographql/apollo-kotlin/releases/tag/v3.7.5)

PR #4681, opened 2023-02-13
stay on AGP 7.4.0-beta02 or IJ sync fails

martinbonnin:
A more recent AGP was needed to support Gradle 8 ([see this PR](https://github.com/apollographql/apollo-kotlin/pull/4644)) but 7.4.0 is a bit too much for IJ


BoD (approved):
🍭

Issue #4003, opened 2022-04-07
ApolloCall execute / toFlow / exception handling improvements


BoD:
### Current situation
`ApolloCall` exposes 2 APIs:
- `suspend execute(): ApolloResponse` 
- `toFlow(): Flow<ApolloResponse>`

`execute()` will throw an `ApolloException` in case of a network, http, or cache error (can also be `ApolloCompositeException` in case of network _and_ cache errors)

It will also throw an exception if the underlying Flow contains more than 1 element (ex: `CacheAndNetwork` or with `@defer`)

Users need to handle 2 kinds of errors:
- network/http errors with a try catch
- GraphQL errors by looking at `.errors`

(With `ApolloResponse.dataAssertNoErrors` this can be centralized in one try catch)

`toFlow()` will throw an exception either as soon as it is encountered, or assembled in `ApolloCompositeException` after some responses have been emitted, depending on the case. 

Cache misses are either emitted with a null `data` or not emitted at all (configurable with `emitCacheMisses`).

### Suggested change
- Add an `exception: ApolloException?` field to `ApolloResponse`
- Never throw `ApolloException` in `toFlow()` : instead, `ApolloResponse` are emitted with a null `data` and non-null `exception`
- To be discussed: should we also not throw in `execute()`? Not sure of the most logical behavior to adopt with `CacheFirst` and a cache miss + network error for instance. No exception thrown but return an ApolloResponse with a null data and an `ApolloCompositeException` in `exception`? 

Benefits:
- Handling network/http errors and GraphQL errors is done in a similar way: look at `exception` or `errors`
- More predictable behavior of `toFlow()`: cache misses / network errors can always be emitted, the caller can decide to ignore them or not by looking at `exception`, depending on the need
- This can potentially simplify our code in some places

### More info
- This would be a breaking change.
- A comment that prompted this idea is [here](https://github.com/apollographql/apollo-kotlin/pull/3987#discussion_r841627202).

martinbonnin:
This is now merged in `main`. It's still time to make adjustments if needed though. Please go to [the RFC](https://github.com/apollographql/apollo-kotlin/issues/4711) for comments

PR #4685, opened 2023-02-14
Do not throw in Flows

BoD:
See #4003.

- Do not throw in HttpNetworkTransport and `WebSocketNetworkTransport`
- Adjust `ApolloCacheInterceptor`, `FetchPolicyInterceptor`, and `ApolloCall,execute()` accordingly
- `ThrowingXyzInterceptor`s, and a `throwApolloException` parameter on `toFlow` are added to get the old behavior back

Missing:
- [x] Apply same changes inside normalized cache incubating package
- [ ] Documentation update
- [x] Java runtime: I think it'd make sense to remove `ApolloCallback.onFailure` and only keep `onResponse` since reponse can have an exception
- [x] Some bits were removed so we should deprecate them first (in #4610)


tests/defer/src/commonTest/kotlin/test/DeferNormalizedCacheTest.kt
```diff
@@ -370,7 +374,10 @@ class DeferNormalizedCacheTest {
                     emit(networkResponse as ApolloResponse<D>)
                   }
                   delay(10)
-                  throw ApolloNetworkException("Network error")
+                  emit(ApolloResponse.Builder(requestUuid = uuid, operation = query, data = null)
```
martinbonnin, line 377: Can you add a comment that `NetworkTransport.execute` must not throw? (which I believe to be the new contract?)

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -52,47 +55,86 @@ private constructor(
...
+            it.withHttpInfo(request.requestUuid, httpResponse, millisStart)
+          }
+      )
     }
+        .catch { e ->
+          if (e is ApolloException) {
+            emitAll(errorResponse(request.operation, e))
+          } else {
+            throw e
+          }
```
martinbonnin, line 98: Can we remove this `catch` and instead only try/catch around to throwing methods? `parseJsonResponse` here I believe?

martinbonnin: Also add a comment to `parseJsonResponse`:

```kotlin
 * @throws IOException if reading [jsonReader] fails
 * @throws JsonEncodingException if the data is not valid json
 * @throws JsonDataException if the data is not of the expected type
 * @throws JsonParseException for other cases  
```

libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/TestNetworkTransport.kt
```diff
@@ -16,40 +20,45 @@ import kotlinx.coroutines.flow.flow
...
       // "Emulate" a network call
       yield()
 
-      val response = lock.withLock { queue.removeFirstOrNull() } ?: error("No more responses in queue")
-      if (response is TestResponse.NetworkError) throw ApolloNetworkException("Network error queued in QueueTestNetworkTransport")
-
       @Suppress("UNCHECKED_CAST")
-      val apolloResponse = (response as TestResponse.Response).response as ApolloResponse<D>
-
+      val apolloResponse = lock.withLock { queue.removeFirstOrNull() } as ApolloResponse<D>? ?: error("No more responses in queue")
```
martinbonnin, line 34: We shouldn't throw in `NetworkTransport.execute`. `"No more responses in queue"` should be in `apolloResponse.exception`

libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/TestNetworkTransport.kt
```diff
@@ -59,26 +68,26 @@ class QueueTestNetworkTransport : NetworkTransport {
...
+  private val operationsToResponses = mutableMapOf<Operation<out Operation.Data>, ApolloResponse<out Operation.Data>>()
 
   override fun <D : Operation.Data> execute(request: ApolloRequest<D>): Flow<ApolloResponse<D>> {
-    val response = lock.withLock { operationsToResponses[request.operation] }
-        ?: error("No response registered for operation ${request.operation}")
-    if (response is TestResponse.NetworkError) throw ApolloNetworkException("Network error queued in QueueTestNetworkTransport")
     @Suppress("UNCHECKED_CAST")
-    val apolloResponse = (response as TestResponse.Response).response as ApolloResponse<D>
+    val apolloResponse = lock.withLock { operationsToResponses[request.operation] } as ApolloResponse<D>?
+        ?: error("No response registered for operation ${request.operation}")
```
martinbonnin, line 76: Same here, we shouldn't throw in `NetworkTransport.execute`

tests/integration-tests/src/commonTest/kotlin/test/CancelTest.kt
```diff
@@ -46,7 +47,7 @@ class CancelTest {
     val apolloClient = ApolloClient.Builder().serverUrl(mockServer.url()).normalizedCache(MemoryCacheFactory()).build()
 
     val job = launch {
-      apolloClient.query(EpisodeHeroNameQuery(Episode.EMPIRE)).executeCacheAndNetwork().toList()
+      apolloClient.query(EpisodeHeroNameQuery(Episode.EMPIRE)).fetchPolicy(FetchPolicy.CacheAndNetwork).toFlow().toList()
```
martinbonnin, line 50: 👍

libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/TestNetworkTransport.kt
BoD: I was wondering about this one! My thinking was: if this happens, it's a programming mistake, and so it might be good that it throws? Not a strong opinion though, I'm also 👍 to emit a response with exception.

martinbonnin: Fair point, Let's go with throwing on programming mistakes 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -83,8 +85,10 @@ class ApolloCall<D : Operation.Data> internal constructor(
    *                    println("order received: ${it.data?.order?.id"})
    *                  }
    * ```
+   * @param throwApolloExceptions whether to throw [ApolloException]s (e.g. network errors, cache misses) encountered in responses,
+   * rather than emitting them in [ApolloResponse.exception]. This was the behavior in 3.x and is kept here as a convenience for migration.
    */
-  fun toFlow(): Flow<ApolloResponse<D>> {
+  fun toFlow(throwApolloExceptions: Boolean = false): Flow<ApolloResponse<D>> {
```
martinbonnin, line 91: Can this be on `ApolloClient.Builder()` so that one doesn't have to pass it to all `toFlow` calls?

BoD: Yes, but I think it could be slightly confusing because `execute()` would still throw.

martinbonnin: Mmmm to me the current contract of `execute` is something like this:

```
  /**
   * Retrieves a single value from the given ApolloCall, ignoring any cache miss or network error
   * @throws ApolloException if the call returns zero or multiple valid GraphQL responses
   */
  suspend fun execute(): ApolloResponse<D> {
```

It could stay exactly the same in 4.x?

BoD: I think that makes sense. I'm just wondering if a user setting `throwApolloExceptions` to `false` on ApolloClient would expect `execute()` to follow it, and thus would expect it to return an ApolloResponse with a non null exception instead of throwing? But maybe clarifying this in `throwApolloExceptions`'s KDoc is enough to avoid confusion.

martinbonnin: I see `throwApolloExceptions = true` as the "exceptional" case. I think we're also most likely going to need a `foldFetchExceptions`. All in all, my favorite solution would look like this:

```kotlin
/**
 * whether exceptions such as cache miss or other ApolloException should throw
 *
 * If true, the call site must catch ApolloException. This was the behaviour in Apollo Kotlin 3
 *
 * Default: false
 */
fun ApolloClient.Builder.throwOnException(throwOnException: Boolean) { }

/**
 * whether to fold the fetch exception in [CacheInfo]
 *
 * Policies such as [FetchPolicy.CacheFirst] or [FetchPolicy.NetworkFirst] are often used to retrieve a
 * single response. In such cases, set [foldFetchExceptions] to true to avoid emitting the first
 * exception (if any) and read it from [ApolloResponse.cacheInfo]?.cacheMissException and [ApolloResponse.cacheInfo]?.networkException
 *
 * Default: false
 */
fun ApolloClient.Builder.foldFetchExceptions(foldFetchExceptions: Boolean) { }

/**
 * Retrieves a single value from the given ApolloCall, ignoring any cache miss or network error
 * @throws ApolloException if the call returns zero or multiple valid GraphQL responses
 */
suspend fun ApolloCall.execute(): ApolloResponse<D> { }
```

If you want 3.x behaviour, set `throwOnException` and `foldFetchExceptions` to true. We can certainly make them deprecated from day 1 on v4. `execute()` could stay as a handy way to get a single value (also because it's not too complicated to keep around). `ThrowingFetchPolicyInterceptors.kt` could go away. WDYT?

BoD: Yes, that looks good to me 👍

libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/TestNetworkTransport.kt
```diff
@@ -16,40 +20,45 @@ import kotlinx.coroutines.flow.flow
...
+        override fun adapter(): Adapter<Operation.Data> = throw UnsupportedOperationException()
+        override fun serializeVariables(
+            writer: JsonWriter,
+            customScalarAdapters: CustomScalarAdapters,
+        ) = throw UnsupportedOperationException()
+
+        override fun rootField(): CompiledField = throw UnsupportedOperationException()
+      }, requestUuid = uuid4(), data = null)
+          .exception(ApolloNetworkException("Network error queued in QueueTestNetworkTransport"))
+          .build()
```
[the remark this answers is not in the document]
BoD: Wait actually I don't have access to an Operation in `QueueTestNetworkTransport`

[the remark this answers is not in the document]
martinbonnin: Not in `request.operation`?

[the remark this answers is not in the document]
BoD: Oh you're right of course! 😅 ✅

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ExecutionOptions.kt
```diff
@@ -81,4 +90,15 @@ interface MutableExecutionOptions<T> : ExecutionOptions {
...
+
+  /**
+   * Configures whether exceptions such as cache miss or other [ApolloException] should throw, instead of being emitted in
+   * [ApolloResponse.exception].
+   *
+   * If true, the call site must catch [ApolloException]. This was the behavior in Apollo Kotlin 3.
+   *
+   * Default: false
+   */
+  @Deprecated("Provided as a convenience to migrate from 3.x, will be removed in a future version")
```
BoD, line 102: Actually now I'm wondering if it should be deprecated as it could be handy in certain cases. For instance with subscriptions, when receiving an error message, in the doc we show [how to use retryWhen](https://www.apollographql.com/docs/kotlin/advanced/websocket-errors#websocket-errors-with-single-consumers), which no longer works, but would with `throwOnException(true)`.

BoD: OTOH it's easy for a project to implement it themselves, so let's not clutter the API surface

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -33,6 +33,14 @@ private constructor(
     @JvmField
     val errors: List<Error>?,
 
+    /**
+     * An [ApolloException] if a valid GraphQL response wasn't received or `null` if a valid GraphQL response was received.
+     * For example, `exception` is non null if there is a network failure or cache miss.
+     * If `exception` is non null, [data] and [errors] will be null.
```
martinbonnin, line 39: Maybe add a cross reference to `errors`? (`see also [errors]`)

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -44,8 +42,9 @@ enum class FetchPolicy {
   /**
    * Try the cache, if that failed, try the network.
    *
-   * An [ApolloCompositeException] is thrown if the data is not in the cache and the network call failed.
-   * If coming from the cache 1 value is emitted, otherwise 1 or multiple values can be emitted from the network.
+   * 1 value from the cache is emitted first (with a non null [ApolloResponse.exception] if it's a cache miss),
+   * and then, if it's a cache miss, 1 or multiple values from the network are emitted (with a non null [ApolloResponse.exception] if there
+   * was a network error).
```
martinbonnin, line 47:
```suggestion
   * This [FetchPolicy] emits one or several [ApolloResponse]s. 
   * Cache misses and network errors have [ApolloResponse.exception] set to a non-null [ApolloException]
```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -54,31 +53,31 @@ enum class FetchPolicy {
   /**
    * Only try the cache.
    *
-   * A [CacheMissException] is thrown if the data is not in the cache, otherwise 1 value is emitted.
+   * 1 value is emitted, with a non null [ApolloResponse.exception] if it's a cache miss.
```
martinbonnin, line 56:
```suggestion
   * This [FetchPolicy] emits one [ApolloResponse].
   * Cache misses have [ApolloResponse.exception] set to a non-null [ApolloException].
```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -54,31 +53,31 @@ enum class FetchPolicy {
...
   CacheOnly,
 
   /**
    * Try the network, if that failed, try the cache.
    *
-   * An [ApolloCompositeException] is thrown if the network call failed and the data is not in the cache.
-   * If coming from the network 1 or multiple values can be emitted, otherwise 1 value is emitted from the cache.
+   * 1 or multiple values from the network are emitted (with a non null [ApolloResponse.exception] if there was a network error),
+   * and then, if there was a network error, 1 value from the cache is emitted (with a non null [ApolloResponse.exception] if it's a cache
+   * miss).
```
martinbonnin, line 65:
```suggestion
   * This [FetchPolicy] emits one or several [ApolloResponse]s. 
   * Cache misses and network errors have [ApolloResponse.exception] set to a non-null [ApolloException]
```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -54,31 +53,31 @@ enum class FetchPolicy {
...
    */
   NetworkFirst,
 
   /**
    * Only try the network.
    *
-   * An [ApolloException] is thrown if the network call failed, otherwise 1 or multiple values can be emitted.
+   * 1 or multiple values are emitted, with a non null [ApolloResponse.exception] if there was a network error.
```
martinbonnin, line 72:
```suggestion
   * This [FetchPolicy] emits one or several [ApolloResponse]s. Several [ApolloResponse]s
   * may be emitted if your [NetworkTransport] supports it. For an example with `@defer`
   * Network errors have [ApolloResponse.exception] set to a non-null [ApolloException]
```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -54,31 +53,31 @@ enum class FetchPolicy {
...
   NetworkOnly,
 
   /**
    * Try the cache, then also try the network.
    *
-   * If the data is in the cache, it is emitted, if not, no exception is thrown at that point. Then the network call is made, and if
-   * successful the value(s) are emitted, otherwise either an [ApolloCompositeException] (both cache miss and network failed) or an
-   * [ApolloException] (only network failed) is thrown.
+   * 1 value from the cache is emitted first (with a non null [ApolloResponse.exception] if it's a cache miss),
+   * and then, 1 or multiple values from the network are emitted (with a non null [ApolloResponse.exception] if there was a network error).
```
martinbonnin, line 80:
```suggestion
   * This [FetchPolicy] emits two or more [ApolloResponse]s. 
   * Cache misses and network errors have [ApolloResponse.exception] set to a non-null [ApolloException]
```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: Tried simplifying and uniformizing the wording. Let me know what you think

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -148,25 +147,14 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
 /**
  * Gets the result from the network, then observes the cache for any changes.
  * [fetchPolicy] will control how the result is first queried, while [refetchPolicy] will control the subsequent fetches.
- * Network and cache exceptions are ignored by default, this can be changed by setting [fetchThrows] for the first fetch and [refetchThrows]
- * for subsequent fetches (non Apollo exceptions like `OutOfMemoryError` are always propagated).
- *
- * @param fetchThrows whether to throw if an [ApolloException] happens during the initial fetch. Default: false
- * @param refetchThrows whether to throw if an [ApolloException] happens during a refetch. Default: false
```
martinbonnin, line 155: That's a behaviour change, can you note it in the migration guide?

martinbonnin: No need to be exhaustive/nitpick the wording but let's keep track of things that changed in there?

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -286,6 +237,22 @@ fun <T> MutableExecutionOptions<T>.refetchPolicyInterceptor(interceptor: ApolloI
     RefetchPolicyContext(interceptor)
 )
 
+/**
+ * Configures whether to fold the fetch exception in [CacheInfo].
+ *
+ * Policies such as [FetchPolicy.CacheFirst] or [FetchPolicy.NetworkFirst] are often used to retrieve a single response. In such cases,
+ * set [foldFetchExceptions] to true to avoid emitting the first exception (if any) and read it from
+ * [ApolloResponse.cacheInfo]?.cacheMissException and [ApolloResponse.cacheInfo]?.networkException.
```
martinbonnin, line 245:
```suggestion
 * [ApolloResponse.cacheInfo]?.cacheMissException and [ApolloResponse.cacheInfo]?.networkException instead.
```

libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/HttpCacheExtensions.kt
```diff
@@ -120,15 +119,10 @@ fun ApolloClient.Builder.httpCache(
...
+              onEach { response ->
                 // Revert caching of responses with errors
                 val cacheKey = synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey[request.requestUuid.toString()] }
-                cacheKey?.let { cachingHttpInterceptor.cache.remove(it) }
-                throw throwable
-              }.onEach { response ->
-                // Revert caching of responses with errors
-                val cacheKey = synchronized(apolloRequestToCacheKey) { apolloRequestToCacheKey[request.requestUuid.toString()] }
-                if (response.hasErrors()) {
+                if (response.hasErrors() || response.exception != null) {
```
martinbonnin, line 125: Much nicer 🤩

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
BoD: Yes, this is better! 👍

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -33,85 +34,81 @@ val CacheOnlyInterceptor = object : ApolloInterceptor {
...
-        }
-      }.singleOrNull()
-
-      if (cacheResponse != null) {
+      ).single()
+      if (cacheResponse.exception == null) {
         emit(cacheResponse)
         return@flow
+      } else if (!request.foldFetchExceptions) {
+        emit(cacheResponse.newBuilder().isLast(false).build())
```
martinbonnin, line 62: Can this be handled in a more "central" place? I'm thinking something like this:

```kotlin

internal val FetchPolicyRouterInterceptor = object : ApolloInterceptor {
    if (request.operation !is Query) {
      // Subscriptions and Mutations do not support fetchPolicies
      return chain.proceed(request)
    }

    if (!request.foldFetchExceptions) {
       // fast path
       return request.fetchPolicyInterceptor.intercept(request, chain)
    }
    return flow {
      val foldedExceptions = mutableListOf<ApolloException>()
      var hasEmitted = false

      request.fetchPolicyInterceptor.intercept(request, chain)
          .collect {
            if (!hasEmitted && it.exception != null) {
              // remember to send the exception later
              foldedExceptions.add(it.exception!!)
              return@collect
            }
            emit(
                it.newBuilder()
                    .cacheInfo(
                        it.cacheInfo!!.newBuilder()
                            .cacheMissException(foldedExceptions.filterIsInstance<CacheMissException>().firstOrNull())
                            .networkException(foldedExceptions.firstOrNull { it !is CacheMissException })
                            .build()
                    )
                    .build()
            )
            hasEmitted = true
          }

      if (hasEmitted) {
        // If we haven't emitted anything, send a composite exception
        val first = if(!foldedExceptions.isEmpty()) foldedExceptions.firstOrNull() else null
        val second = if(foldedExceptions.size > 1) foldedExceptions.get(1) else null
        emit(
            ApolloResponse.Builder(request.operation, request.requestUuid, null)
                .exception(ApolloCompositeException(first, second))
                .build()
        )
      }
    }
  }
}

```

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
martinbonnin: I pushed a change to use the same wording in the non-incubating normalized cache (and also remove a non-breaking space that made it through the GitHub UI)

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -93,16 +106,34 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
+   * For subscriptions or operations using `@defer`, you usually want to use [toFlow] instead to listen to all values.
+   *
+   * @throws ApolloException if the call returns zero or multiple valid GraphQL responses.
    */
   suspend fun execute(): ApolloResponse<D> {
-    return toFlow().single()
+    val responses = toFlow().toList()
+    val (errors, successes) = responses.partition { it.exception != null }
+    if (successes.size > 1) {
+      throw ApolloException("The operation returned multiple items, use .toFlow() instead of .execute()")
```
martinbonnin, line 126: 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -93,16 +106,34 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
+    } else if (successes.isEmpty()) {
+      if (errors.size == 1) {
+        throw errors[0].exception!!
+      } else if (errors.size > 1) {
+        throw ApolloCompositeException(errors[0].exception, errors[1].exception)
+      } else {
+        throw ApolloException("The operation did not emit any item, check your interceptor chain")
+      }
```
martinbonnin, line 134: Should we honour `throwOnException` in `execute()` as well? Seems weird to have an API that requires catching exceptions and not the other one.

martinbonnin: Also reminder to update [the doc page for error handling](https://www.apollographql.com/docs/kotlin/essentials/errors)

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/AutoPersistedQueryInterceptor.kt
```diff
@@ -50,7 +50,12 @@ internal class AutoPersistedQueryInterceptor(
           emit(response.withAutoPersistedQueryInfo(false))
         }
         isPersistedQueryNotSupported(response.errors) -> {
-          throw AutoPersistedQueriesNotSupported()
+          emit(
+              ApolloResponse.Builder(request.operation, request.requestUuid, null)
+                  .exception(AutoPersistedQueriesNotSupported())
+                  .build()
+          )
+          return@flow
```
martinbonnin, line 58: 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/NetworkTransport.kt
```diff
@@ -1,12 +1,20 @@
...
 import kotlinx.coroutines.flow.Flow
 
 interface NetworkTransport {
 
+  /**
+   * Execute a request.
+   *
+   * @param request the request to execute.
+   * @return a flow of responses.
+   * [ApolloException]s must not be thrown from the flow. Instead, they must be emitted as [ApolloResponse.exception].
```
martinbonnin, line 16:
```suggestion
   * This function must not throw in normal circumstances. Instead, return an [ApolloResponse] with a non
   * null [ApolloResponse.exception].
   * It may throw if the error is the result of a programming error like wrongly configured interceptors.
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
BoD: Yes, I was bothered by this because it needs to throw anyway here - but I agree it's a programming error, whereas for the regular case it makes sense to follow `throwOnException`. 

This change is more impactful I think (assuming `execute` is used a lot more than `toFlow`)! 
I don't think a migration helper could be made to automatically refactor code to be `if (response.exception != null)` instead of `try {} catch(e: ApolloException {}` (or maybe only for simple cases) but it could probably at least detect it and issue a warning.

I still think it's the right move: 
- it could be surprising to have `ApolloReponse.exception` but it being essentially useless because the exception is thrown instead
- generally speaking, throwing is not cool
- one can still set `throwOnException` to false (either globally on a specific call) (albeit it's deprecated)
- one can do `response.exception?.let { throw it }` or write an extension that does it

martinbonnin: I'm confused are you suggesting to keep as is and throw on network error (even if `throwOnException  == false`)?

BoD: No, I'm agreeing we should throw if it's true, and not throw if it's false 😊
Also recognizing that since it impacts execute(), and it's false by default, it's somewhat a "bigger" change. Regardless, it makes sense to me.

martinbonnin: giphy

libraries/apollo-api/api/apollo-api.api
```diff
@@ -633,6 +641,7 @@ public abstract interface class com/apollographql/apollo3/api/MutableExecutionOp
 	public abstract fun httpMethod (Lcom/apollographql/apollo3/api/http/HttpMethod;)Ljava/lang/Object;
 	public abstract fun sendApqExtensions (Ljava/lang/Boolean;)Ljava/lang/Object;
 	public abstract fun sendDocument (Ljava/lang/Boolean;)Ljava/lang/Object;
+	public abstract fun throwOnException (Ljava/lang/Boolean;)Ljava/lang/Object;
```
martinbonnin, line 644: Question: what about adding `throwOnException()` only on `ApolloClient.Builder` (and not `MutableExecutionOptions`)? Feels a bit overkill to be able to override that behaviour at the `ApolloCall`/`ApolloRequest` level and it limits the API surface changes?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -28,11 +28,23 @@ private constructor(
 
     /**
      * GraphQL [operation] execution errors returned by the server to let client know that something has gone wrong.
-     * This can either be null or empty depending on what your server sends back
+     * This can either be null or empty depending on what your server sends back.
+     *
+     * See also [exception] for exceptions happening before a valid GraphQL response could be received.
```
martinbonnin, line 33: Re-reading this. The GraphQL errors are not only from execution. Server validation errors will also appear here (although they should "never happen" because we also validate client side). Also feels like this should mention partial data. 

Tentative change:

```
[GraphQL errors](https://spec.graphql.org/October2021/#sec-Errors) returned by the server to let client know that something has gone wrong.

If no GraphQL error was raised, [errors] is null. Else it's a non-empty list of errors indicating where the error(s) happened.

Note that because GraphQL allows partial data, it is possible to have both [data] non null and [errors] non null.

See also [exception] for exceptions happening before a valid GraphQL response could be received.
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -71,10 +83,10 @@ private constructor(
   @get:JvmName("dataAssertNoErrors")
   val dataAssertNoErrors: D
```
martinbonnin, line 84: Where do we stand on this API? Since we're moving most of the API to be non-throwing, should we deprecate it? And maybe offer `Result` like helpers instead?

martinbonnin: Also I think there's a soft consensus that `val` should not throw.

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -286,6 +238,22 @@ fun <T> MutableExecutionOptions<T>.refetchPolicyInterceptor(interceptor: ApolloI
...
+ * Policies such as [FetchPolicy.CacheFirst] or [FetchPolicy.NetworkFirst] are often used to retrieve a single response. In such cases,
+ * set [foldFetchExceptions] to true to avoid emitting the first exception (if any) and read it from
+ * [ApolloResponse.cacheInfo]?.cacheMissException and [ApolloResponse.cacheInfo]?.networkException instead.
+ *
+ * This was the behavior in Apollo Kotlin 3.
+ *
+ * Default: false
+ */
+@Deprecated("Provided as a convenience to migrate from 3.x, will be removed in a future version", ReplaceWith(""))
+fun <T> MutableExecutionOptions<T>.foldFetchExceptions(foldFetchExceptions: Boolean) = addExecutionContext(
```
martinbonnin, line 253: Same as `throwOnException`? Make it on `ApolloClient.Builder` only?

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -33,83 +33,72 @@ val CacheOnlyInterceptor = object : ApolloInterceptor {
...
-        }
-      }.map { response ->
+      ).map { response ->
         response.newBuilder()
             .cacheInfo(
                 response.cacheInfo!!
                     .newBuilder()
-                    .cacheMissException(cacheException as? CacheMissException)
+                    .networkException(response.exception)
+                    .cacheMissException(cacheResponse.exception as? CacheMissException)
```
martinbonnin, line 70: Is that still required now that the exceptions are set in `FetchPolicyRouterInterceptor`?

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/FetchPolicyInterceptors.kt
```diff
@@ -33,83 +33,72 @@ val CacheOnlyInterceptor = object : ApolloInterceptor {
...
-          cacheException = throwable
-        } else {
-          throw throwable
-        }
-      }.singleOrNull()
-
-      if (cacheResponse != null) {
-        emit(cacheResponse)
+      ).single()
+      emit(cacheResponse.newBuilder().isLast(cacheResponse.exception == null).build())
```
martinbonnin, line 57: This `isLast` + `single` thing feels a bit "off". Let's revisit in a future PR (see also [this one](https://github.com/apollographql/apollo-kotlin/pull/4319))

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -68,13 +77,16 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
+        .throwOnException(throwOnException)
   }
 
   /**
-   * Returns a cold Flow that produces [ApolloResponse]s for this [ApolloCall].
+   * Returns a cold Flow that produces [ApolloResponse]s from this [ApolloCall].
    * Note that the execution happens when collecting the Flow.
    * This method can be called several times to execute a call again.
    *
+   * Any [ApolloException]s will be emitted in [ApolloResponse.exception] and will not be thrown, unless [throwOnException] is set to true.
```
martinbonnin, line 88: Simpler? 

```suggestion
   * The returned [Flow] does not throw unless [throwOnException] is set to true.
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -93,16 +106,36 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
+        errors[1].newBuilder()
+            .exception(ApolloCompositeException(errors[0].exception, errors[1].exception))
+            .build()
+      } else {
+        throw ApolloException("The operation did not emit any item, check your interceptor chain")
+      }
+    } else {
+      return successes.first()
+    }
   }
```
martinbonnin, line 140: Suggestion: use `when`, lift `return` outside of the branches and sort the `when` cases by number of items

```suggestion
    return when(successes.size) {
      0 -> {
        when(errors.size) {
          0 -> throw ApolloException("The operation did not emit any item, check your interceptor chain")
          1 -> errors.first()
          else -> {
            errors[1].newBuilder()
                .exception(ApolloCompositeException(errors[0].exception, errors[1].exception))
                .build()
          }
        }
      }
      1 -> {
        successes.first()
      }
      else -> {
        throw ApolloException("The operation returned multiple items, use .toFlow() instead of .execute()")
      }
    }
```

martinbonnin: Also might be worth adding an `ApolloCompositeException` that takes > 2 suppressed exceptions to account for this case?

libraries/apollo-normalized-cache-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -148,25 +148,14 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
...
- * for subsequent fetches (non Apollo exceptions like `OutOfMemoryError` are always propagated).
- *
- * @param fetchThrows whether to throw if an [ApolloException] happens during the initial fetch. Default: false
- * @param refetchThrows whether to throw if an [ApolloException] happens during a refetch. Default: false
  */
-@JvmOverloads
-fun <D : Query.Data> ApolloCall<D>.watch(
-    fetchThrows: Boolean = false,
-    refetchThrows: Boolean = false,
-): Flow<ApolloResponse<D>> {
```
martinbonnin, line 161: Keep the overload with `DeprecationLevel.ERROR`?

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
```diff
@@ -28,7 +27,6 @@ import com.apollographql.apollo3.interceptor.ApolloInterceptor
 import com.apollographql.apollo3.interceptor.ApolloInterceptorChain
 import com.apollographql.apollo3.mpp.currentTimeMillis
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.catch
```
martinbonnin, line 31: There's a `maybeWriteToCache` in this file that certainly needs to check for `response.exception` before writing to the cache?

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -93,16 +106,36 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
   /**
-   * A shorthand for `toFlow().single()`.
-   * Use this for queries and mutation to get a single [ApolloResponse] from the network or the cache.
-   * For subscriptions, you usually want to use [toFlow] instead to listen to all values.
+   * Retrieve a single [ApolloResponse] from this [ApolloCall], ignoring any cache misses or network errors.
+   *
+   * Use this for queries and mutations to get a single value from the network or the cache.
+   * For subscriptions or operations using `@defer`, you usually want to use [toFlow] instead to listen to all values.
+   *
+   * @throws ApolloException if the call returns zero or multiple valid GraphQL responses.
```
martinbonnin, line 120: Trying to wrap my head around this:

* throwOnException = false
  * foldFetchException = false 
    * cache first + cache error + network success -> returns the network response, no need to have info about the cache miss
    * cache first + cache error + network error -> returns response.exception = ApolloCompositeException() with both exceptions
  * foldFetchException = true
    * cache first + cache error + network success -> returns the network response, cache miss is in response.cacheInfo.cacheMissException
    * cache first + cache error + network error -> returns response.exception = ApolloCompositeException() with both 
* throwOnException = true
  * foldFetchException = false 
    * cache first + cache error + network success -> returns the network response, no need to have info about the cache miss
    * cache first + cache error + network error -> throws ApolloCompositeException() with both exceptions
* throwOnException = true
  * foldFetchException = true 
    * cache first + cache error + network success -> returns the network response, cache miss is in response.cacheInfo.cacheMissException
    * cache first + cache error + network error -> throws ApolloCompositeException() with both exceptions

Did I get this right? 

I'm starting to question the need for a separate `foldFetchException`. We can keep it internally of course but given this is only for backward compat, there's no real need to decouple both parameters?

BoD: This looks 💯 correct to me. 

And yes, good point, as a user migrating to v4 this would probably be more confusing than needed!. A single flag to replicate the v3 behavior is probably better. Now the difficult question: how to name the flag? 😅

`foldAndThrowExceptions`?

libraries/apollo-api/api/apollo-api.api
BoD: Yeah good point, this is probably overkill. Will keep it confined to `ApolloClient`  👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
martinbonnin: Something else I just realized is one is in `apollo-runtime` and the other in `apollo-normalized-cache` (fun times!)

Maybe make it explicit about v3 compat? 

* apollo-runtime: `throwOnExceptions()`?
* apollo-normalized-cache: `useV3ExceptionHandling()` ?  (calls `throwOnExceptions()`)

I'm not 100% sold TBH. I think we'll need to check the whole story in Confetti/tutorial/other repos to double check the ergonomics later on.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
BoD: Yeah, the goal was to help centralizing error handling, but now you have it anyway by checking `.errors` and `.exception` without a need to throw. Ok to deprecate it! 🧹

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
BoD: What about `useV3ExceptionHandling()` only?

martinbonnin: In both artifacts? Feels dangerous because it's very easy to import the wrong one.

BoD: No sorry I meant only in apollo-runtime, and it could control both behaviors. Or am I missing something?

martinbonnin: That would force exposing the parameter outside of `apollo-runtime` (for `apollo-normalized-cache` to be able to read it). Works for me with some `@ApolloInternal` annotation?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
martinbonnin: As discussed, let's wait a bit before we deprecate

BoD: Reverted it and opened #4696

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt
```diff
@@ -121,28 +124,32 @@ fun ApolloClient.Builder.store(store: ApolloStore, writeToCacheAsynchronously: B
       .writeToCacheAsynchronously(writeToCacheAsynchronously)
 }
 
+@Deprecated(level = DeprecationLevel.ERROR, message = "Exceptions no longer throw", replaceWith = ReplaceWith("watch()"))
+@ApolloDeprecatedSince(v4_0_0)
+@Suppress("UNUSED_PARAMETER")
+fun <D : Query.Data> ApolloCall<D>.watch(
+    fetchThrows: Boolean,
+    refetchThrows: Boolean,
+) = watch()
```
martinbonnin, line 133: maybe throw instead? It's better to have a clear error than not knowing where to look?

BoD: It's `DeprecationLevel.ERROR` so you shouldn't be able to call this, but I was thinking of the case where the code was compiled with 3.x but is running with 4.x. On the other hand this will surely not behave as expected in any case, so crashing early with a clear message may actually be better in fact 👍

martinbonnin (approved):
🚀

PR #4693, opened 2023-02-17
More doc updates about not throwing exceptions

BoD:
See #4685

### DIFF
--- a/docs/source/essentials/errors.mdx
+++ b/docs/source/essentials/errors.mdx
@@ -10,26 +10,21 @@ Whenever you execute a GraphQL operation with Apollo Kotlin (or any other GraphQ
 
 ## Network errors
 
-When calling `execute()`, network errors throw an `ApolloException`. To handle them, wrap your query in a `try/catch` block:
+Network errors are surfaced in the `ApolloResponse.exception` field.
 
 ```kotlin
-try {
-  val response = apolloClient.query(query).execute()
-} catch(exception: ApolloException) {
+val response = apolloClient.query(query).execute()
+if (response.exception != null) {
   // handle exception here
 }
 ```
 
-When using `toFlow()` (e.g. with [subscriptions](../essentials/subscription)) and `watch()` (with the [normalized cache](../caching/query-watchers)), network errors are emitted in `ApolloResponse.exception`.
+This is also true when calling `toFlow()` (e.g. with [subscriptions](../essentials/subscription)) and `watch()` (with the [normalized cache](../caching/query-watchers)).
 
 ```kotlin
-apolloClient.query(query).toFlow().collect { response ->
+apolloClient.subscription(subscription).toFlow().collect { response ->
   if (response.exception != null) {
     // handle network errors
-  } else if (response.hasErrors()) {
-    // handle GraphQL errors
-  } else {
-    // no errors, handle data
   }
 }
 ```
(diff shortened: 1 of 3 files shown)


docs/source/essentials/errors.mdx
```diff
@@ -10,26 +10,21 @@ Whenever you execute a GraphQL operation with Apollo Kotlin (or any other GraphQ
...
 
 ```kotlin
-apolloClient.query(query).toFlow().collect { response ->
+apolloClient.subscription(subscription).toFlow().collect { response ->
   if (response.exception != null) {
     // handle network errors
-  } else if (response.hasErrors()) {
-    // handle GraphQL errors
-  } else {
-    // no errors, handle data
```
martinbonnin, line 32: Why remove that part?

docs/source/migration/4.0.mdx
```diff
@@ -3,5 +3,79 @@ title: Migrating to Apollo Kotlin 4.0
...
+> Note: this page is a work in progress.
+
 * Requires Kotlin Gradle Plugin 1.6 (from 1.5)
 * Dropped `enableCompatibilityMetadataVariant`
+
+## Error handling
+
+### Exception surfacing
+
+In v3, errors like network errors, cache misses, and WebSocket protocol errors were surfaced by throwing exceptions in `ApolloCall.execute()` and in Flows (`ApolloCall.toFlow()`, `ApolloCall.watch()`). This was not ideal because it was a difference in how to handle GraphQL errors vs other errors. Moreover, throwing terminates a Flow and consumers would have to handle re-collection.
```
martinbonnin, line 15:
```suggestion
In v3, non-GraphQL errors like network errors, cache misses, and WebSocket protocol errors were surfaced by throwing exceptions in `ApolloCall.execute()` and in Flows (`ApolloCall.toFlow()`, `ApolloCall.watch()`). This was not ideal because it was a difference in how to handle GraphQL errors vs other errors. Moreover, throwing terminates a Flow and consumers would have to handle re-collection.
```

docs/source/migration/4.0.mdx
```diff
@@ -3,5 +3,79 @@ title: Migrating to Apollo Kotlin 4.0
...
+
+In v4, a field `ApolloResponse.exception` has been added and these errors are now surfaced by returning (for `execute()`) or emitting (for Flows) an `ApolloResponse` with a non null `exception` instead of throwing it. This allows consumers to handle different kinds of errors at the same place, and it avoids Flows from being terminated. 
+
+To ease the migration to v4, the v3 behavior can be restored by setting `ApolloClient.Builder.throwOnException()` to `true`.
```
martinbonnin, line 19:
```suggestion
To ease the migration to v4, the v3 behavior can be restored by calling `ApolloClient.Builder.throwOnException(true)`.
```

docs/source/essentials/errors.mdx
BoD: It's just that we talk about GraphQL errors in the next section. It's spoiling it! :)

martinbonnin: Fair enough. Ultimately, we'll want to be opinionated about how to handle errors and use the same pattern everywhere for SEO purposes...

martinbonnin (approved):
📖

PR #4695, opened 2023-02-20
"Pre-deprecate" dataAssertNoErrors

BoD:


BoD:
As discussed, let's not touch this in v3

PR #4696, opened 2023-02-21
Deprecate dataAssertNoErrors

BoD:
Moving this on a dedicated PR for now to try it out


BoD:
Reminder to update the migration guide if we do merge this

BoD:
Let's do this:
- in v4
  - mark `dataAssertNoErrors` as deprecated (warning level), because the cosmetics are not great
  - add `fun dataOrThrow()` to replace it
  - (wait for the refact of `exceptions` that surfaces GraphQL errors before doing this)
- in v3
  - nothing

PR #4697, opened 2023-02-21
Update AGP and Gradle

martinbonnin:
Fix for https://issuetracker.google.com/issues/252848749

```
java.lang.NoSuchMethodError: 'org.gradle.internal.buildoption.BuildOption$Value org.gradle.api.internal.StartParameterInternal.getIsolatedProjects()'
```


BoD (approved):
👍

PR #4698, opened 2023-02-21
Fix cacheKeyForObject doc

martinbonnin:
`leafType()` is deprecated and `rawType().name` might give the name of the interface and not the concrete type

Many thanks @trod-123 for catching this


BoD (approved):
👍

PR #4700, opened 2023-02-22
Fix normalizing data when using `@include` or `@skip` with default values

martinbonnin:
For queries like this:

```graphql
query GetAnimal($skipIf: Boolean = true) {
  animal {
    species @skip(if: $skipIf)
  }
}
```
Because the normalizer didn't know about the default value, it would try to normalize null to the cache and fail because species is of non-nullable type.

Note to myself to add backward compat `copy` functions when backporting this to v3


BoD (approved):
🔧

Issue #4699, opened 2023-02-22
Deprecated arguments missing from introspection query


BoD:
### Version

3.7.4

### Summary

Deprecated field arguments are not included when retrieving a schema.

`(includeDeprecated: true)` is missing [here](https://github.com/apollographql/apollo-kotlin/blob/7cfc0c3cfb066768b680385e3eee3e6a2bee05e8/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt#L193).

BoD:
Initially reported [here](https://community.apollographql.com/t/download-deprecated-parameters-using-introspection/5784), thanks [Tomino](https://community.apollographql.com/u/Tomino)!

BoD:
This has been released as part of [v3.7.5](https://github.com/apollographql/apollo-kotlin/releases/tag/v3.7.5)

PR #4702, opened 2023-02-22
Include deprecated arguments and directives in introspection

BoD:
See #4699


martinbonnin (approved):
🚀

martinbonnin (changes requested):
As hinted by @calvincestari (Thanks!), we'll need to make `includeDeprecated` on args and inputFields optional as they're only supported in the draft version of the spec and some servers might not support it.

libraries/apollo-cli/src/main/kotlin/com/apollographql/apollo3/cli/DownloadSchemaCommand.kt
```diff
@@ -39,7 +40,8 @@ internal class DownloadSchemaCommand: CliktCommand() {
         registryUrl = registryUrl,
         schema = File(schema),
         headers = headersMap,
-        insecure =insecure,
+        insecure = insecure,
+        includeDeprecatedInputValuesAndArguments = includeDeprecatedInputValuesAndArguments,
```
martinbonnin, line 44: This should be either `includeDeprecatedInputValues` or `includeDeprecatedInputFieldsAndArguments`

martinbonnin:
🐶 🥫  suggestion: could we recursively use apollo-kotlin to build a typesafe introspection query to be used by apollo-kotlin? Maybe have 2 versions of the schema, one for the current spec and one for the draft?

libraries/apollo-cli/src/main/kotlin/com/apollographql/apollo3/cli/DownloadSchemaCommand.kt
BoD: Oh yes good catch! I'll go with`includeDeprecatedInputFieldsAndArguments` (not a big fan of the long name but at least it's precise)

BoD:
I tried to use apollo-kotlin but couldn't use the 'draft spec' introspection types with it 😿. 

Instead I updated the built-ins so it will possible in the future after this is published in a version. Also removed the restriction preventing from overriding the introspection types in a project's schema, so if this happens again in the future we'll be able to make it work.

Also, removed the flag and instead try the 'draft spec' query first, and fallback to the 'October2021 spec' query if it failed.

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -13,30 +13,31 @@ import okio.source
 import java.io.File
 
 @Serializable
-private data class IntrospectionSchemaEnvelope(
+private class IntrospectionSchemaEnvelope(
```
martinbonnin, line 16: Might be worth doing a separate PR for this? I think we'll want `equals`/`hashCode`/`toString` implementations to keep working? Also we can then do it for `gql.kt` at the same time

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrSchemaBuilder.kt
```diff
@@ -38,9 +38,6 @@ object IrSchemaBuilder {
       val typeDefinition = schema.typeDefinition(name)
 
       when {
-        typeDefinition.isBuiltIn() && typeDefinition !is GQLScalarTypeDefinition-> {
-          // do nothing, introspection types are in apollo-api
-        }
```
martinbonnin, line 43: If you're doing this, you'll have to remove `CompiledSchemaType` and friends [there](https://github.com/apollographql/apollo-kotlin/blob/a4d5f82888218fddf1d70ee9bdbbfe17befb04a2/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt#L378)

And also the shortcuts in codegen [there](https://github.com/apollographql/apollo-kotlin/blob/b0cac343c06c591273df779b4c423b3422a9af05/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinResolver.kt#L188). 

I'm all for it but if we want to be able to backport this to 3.x, maybe do this in a separate PR?

BoD: I think I'm gonna revert this after all, it breaks `generateSchema` and not sure it's worth it

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
martinbonnin: `toString()` is nice for debugging. Not sure about equals/hashCode

martinbonnin:
* When converting from introspection Json -> SDL, we should keep the introspection type definitions, directive definitions and scalar definitions 
* Providing a schema with type definitions should override the builtins.graphql one
* For backward compat reason, we're still allowing users to pass a schema without introspection type definitions
* To do in separate PR:
  * remove data classes
* investigate if we can remove `CompilesSchemaType` and others

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrSchemaBuilder.kt
BoD: This didn't work when using enums from the built-in introspection (so either `__TypeKind` or `__DirectiveLocation`. This fixes it, but not sure if it's odd.

martinbonnin: I guess that's OK for 3.x given we have a hack already for scalars. Certainly something to revisit for 4.x though

BoD: Yeah I think this PR should be OK to merge in 3.x. I'll open other PRs for 4.x only (with some breaking changes).

libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -1248,15 +1248,21 @@ public final class com/apollographql/apollo3/ast/introspection/IntrospectionSche
 
 public final class com/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema {
 	public static final field Companion Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$Companion;
-	public synthetic fun <init> (ILcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$QueryType;Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$MutationType;Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$SubscriptionType;Ljava/util/List;Lkotlinx/serialization/internal/SerializationConstructorMarker;)V
+	public synthetic fun <init> (ILcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$QueryType;Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$MutationType;Lcom/apollographql/apollo3/ast/introspection/IntrospectionSchema$Schema$SubscriptionType;Ljava/util/List;Ljava/util/List;Lkotlinx/serialization/internal/SerializationConstructorMarker;)V
```
martinbonnin, line 1251: For the record, this serialization constructor isn't really part of the public API as it looks like it can only be called from this module. (example decompiled file [here](https://gist.github.com/martinbonnin/df5805b3357369184f81a9939cec2584))

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/introspection_to_schema.kt
```diff
@@ -286,10 +336,6 @@ private class GQLDocumentBuilder(private val introspectionSchema: IntrospectionS
 @ApolloExperimental
 fun IntrospectionSchema.toGQLDocument(filePath: String? = null): GQLDocument = GQLDocumentBuilder(this, filePath)
     .toGQLDocument()
-    /**
-     * Introspection already contains builtin types like Int, Boolean, __Schema, etc...
-     */
-    .withoutBuiltinDefinitions()
```
martinbonnin, line 292: 👍

libraries/apollo-ast/src/main/resources/builtins.graphqls
```diff
@@ -1,6 +1,6 @@
 # Built in scalar and introspection types from:
 # - https://spec.graphql.org/June2018/#sec-Scalars
-# - https://spec.graphql.org/June2018/#sec-Schema-Introspection
+# - https://github.com/graphql/graphql-spec/blob/afc0a35d271ba9502c3c68aeda6e6c6fbc223774/spec/Section%204%20--%20Introspection.md
```
martinbonnin, line 3: Can you add a comment here ? 

```
# In theory the user needs to provide the builtin definitions because we cannot know in advance what
# version of the spec they are using neither if they have extended any of the introspection types.
# This file is a fallback to make a best effort guess in case the user didn't provide these definitions at the 
# risk of potentially validating invalid queries.
```

libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/introspection/duplicate.json
```diff
@@ -1512,7 +1512,8 @@
               },
               "defaultValue": null
             }
-          ]
+          ],
+          "isRepeatable": false
```
martinbonnin, line 1516: This is (yet another) question: I think we'll want to keep being able to read "old" json schemas. Without this, we're taking the risk of having codegen fail for a lot of users that are using current (or older) spec without `"isRepeatable"`

libraries/apollo-compiler/src/test/graphql/com/example/introspection_query/kotlin/operationBased/introspection_query/Introspection.kt.expected
```diff
@@ -0,0 +1,217 @@
...
+import com.example.introspection_query.fragment.IntrospectionCommon
+import com.example.introspection_query.selections.IntrospectionSelections
+import kotlin.Any
+import kotlin.Boolean
+import kotlin.Int
+import kotlin.String
+import kotlin.Unit
+import kotlin.collections.List
+
+public class Introspection() : Query<Introspection.Data> {
```
martinbonnin, line 27: Where are these files coming from?

martinbonnin (changes requested):
The same way we need to support querying both versions from introspection, we'll need to be able to read both versions of Jsons. Maybe it's "just" a matter of setting default values for `isRepeatable` and others potentially missing fields.

libraries/apollo-compiler/src/test/graphql/com/example/introspection_query/kotlin/operationBased/introspection_query/Introspection.kt.expected
BoD: Woops added by mistake, good catch!

libraries/apollo-ast/api/apollo-ast.api
martinbonnin: https://github.com/Kotlin/kotlinx.serialization/issues/2209

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -238,7 +238,16 @@ fun BufferedSource.toIntrospectionSchema(origin: String = ""): IntrospectionSche
...
     val introspectionSchemaEnvelope = json.decodeFromString(IntrospectionSchemaEnvelope.serializer(), this.readUtf8())
-    introspectionSchemaEnvelope.data ?: introspectionSchemaEnvelope.__schema?.let { IntrospectionSchema(it) }
+    introspectionSchemaEnvelope.data ?: introspectionSchemaEnvelope.__schema?.let { schema ->
+      IntrospectionSchema(IntrospectionSchema.Schema(
+          queryType = schema.queryType,
+          mutationType = schema.mutationType,
+          subscriptionType = schema.subscriptionType,
+          types = schema.types,
+          // Old introspection json (pre `April2016`) may not have the `locations` field, in which case the list will be empty, which is invalid. Exclude those directives.
+          directives = schema.directives.filter { directive -> directive.locations.isNotEmpty() }
```
martinbonnin, line 248: Have you ever seen such a schema? 

This is certainly the way to go for 3.x because [we ignore unknown directives](https://github.com/apollographql/apollo-kotlin/blob/05f2a3295517fb7a64c2259c38b453a37f3c4d19/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/internal/ValidationCommon.kt#L133). But 4.x should certainly require a post 2016 schema

libraries/apollo-gradle-plugin/testProjects/convertSchema/schemas/schema-legacy.json
```diff
@@ -0,0 +1,778 @@
+{
+  "__schema": {
```
martinbonnin, line 2: Can we use something that's more precise than legacy to name this schema? For an example using a version from the [spec website](https://spec.graphql.org/): `schema-april-2016-spec.json` ?

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
BoD: > Have you ever seen such a schema?
No, except in the json of our own tests (this didn't cause any issue because the whole `directives` list was ignored)

libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/introspection/duplicate.json
martinbonnin: Can you revert that part? I'd like to keep our test suite unchanged unless we make some behaviour change that we need to document either in the CHANGELOG or migration guide (but I don't think this is the case here?)

BoD: 👍 makes sense

libraries/apollo-gradle-plugin/testProjects/buildCache/src/main/graphql/starwars/schema.json
```diff
@@ -5815,7 +5815,13 @@
...
               "defaultValue": null
             }
-          ]
+          ],
+          "locations": [
+            "FIELD",
+            "FRAGMENT_SPREAD",
+            "INLINE_FRAGMENT"
+          ],
+          "isRepeatable": false
```
martinbonnin, line 5824: Same here, can we keep the test suite unchanged ?

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
BoD: I can remove the `= emptyList()` part on directives for v4 then - this will throw at parse time.

martinbonnin: Yea I think that's OK. It's not the best error but it should not happen too much hopefully 🤞

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -144,9 +157,66 @@ data class IntrospectionSchema(
...
+        mutationType: MutationType? = this.mutationType,
+        subscriptionType: SubscriptionType? = this.subscriptionType,
+        types: List<Type> = this.types,
+    ) = copy(
+        queryType = queryType,
+        mutationType = mutationType,
+        subscriptionType = subscriptionType,
+        types = types,
+        directives = emptyList(),
+    )
```
martinbonnin, line 219: I'm guessing this is a binary compatibility method? If yes, the problem is that it shadows the default `.copy` and gets resolved from `normalize()` for an example. 

Maybe make it hidden? 

```suggestion
    @Deprecated("For binary compatibility only.", level = DeprecationLevel.HIDDEN)
    fun copy(
        queryType: QueryType = this.queryType,
        mutationType: MutationType? = this.mutationType,
        subscriptionType: SubscriptionType? = this.subscriptionType,
        types: List<Type> = this.types,
    ) = copy(
        queryType = queryType,
        mutationType = mutationType,
        subscriptionType = subscriptionType,
        types = types,
        directives = emptyList(),
    )
```

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
```diff
@@ -168,7 +238,16 @@ fun BufferedSource.toIntrospectionSchema(origin: String = ""): IntrospectionSche
...
   return try {
     val introspectionSchemaEnvelope = json.decodeFromString(IntrospectionSchemaEnvelope.serializer(), this.readUtf8())
-    introspectionSchemaEnvelope.data ?: introspectionSchemaEnvelope.__schema?.let { IntrospectionSchema(it) }
+    introspectionSchemaEnvelope.data ?: introspectionSchemaEnvelope.__schema?.let { schema ->
+      IntrospectionSchema(IntrospectionSchema.Schema(
+          queryType = schema.queryType,
+          mutationType = schema.mutationType,
+          subscriptionType = schema.subscriptionType,
+          types = schema.types,
+          // Old introspection json (pre `April2016`) may not have the `locations` field, in which case the list will be empty, which is invalid. Exclude those directives.
```
martinbonnin, line 247:
```suggestion
          // Old introspection json (pre `April2016`) may not have the `locations` field, in which case the list will be empty, which is invalid. Exclude those directives.
          // Validation doesn't validate the unknown directives (yet). A future version may want to fail here and enforce proper validation.
```

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/introspection_to_schema.kt
```diff
@@ -274,6 +280,50 @@ private class GQLDocumentBuilder(private val introspectionSchema: IntrospectionS
...
+        sourceLocation = sourceLocation,
+        description = description,
+        name = name,
+        arguments = args.map { it.toGQLInputValueDefinition() },
+        locations = locations.map { it.toGQLDirectiveLocations() },
+        repeatable = isRepeatable,
+    )
+  }
+
+  private fun IntrospectionSchema.Schema.Directive.Argument.toGQLInputValueDefinition(): GQLInputValueDefinition {
```
martinbonnin, line 295: Note for 4.x: I'm almost sure `Directive.Argument` is the same as `Field.Argument` (and since we're going to use generated model, we can certainly share a fragment there)

libraries/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/ConvertSchemaTests.kt
```diff
@@ -74,12 +73,31 @@ class ConvertSchemaTests {
...
-          greeting: String
-        }
-      """.trimIndent())
+      assertjThat(to.readText())
+          .isEqualTo(File(dir, "schemas/schema.sdl").readText())
     }
   }
+
+  @Test
+  fun `convert from legacy Json to SDL works`() {
```
martinbonnin, line 82:
```suggestion
  fun `convert from october-2015 JSON to SDL works`() {
```

libraries/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/ConvertSchemaTests.kt
```diff
@@ -74,12 +73,31 @@ class ConvertSchemaTests {
...
+    TestUtils.withTestProject("convertSchema") { dir ->
+      // schema-legacy.json doesn't have:
```
martinbonnin, line 84:
```suggestion
      // schema-october-2015.json doesn't have:
```

libraries/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/DownloadSchemaTests.kt
```diff
@@ -140,6 +142,27 @@ class DownloadSchemaTests {
...
+
+      mockServer.takeRequest().body.readUtf8().let {
+        assertTrue(it.contains("inputFields(includeDeprecated: true)"))
+        assertTrue(it.contains("args(includeDeprecated: true)"))
+      }
+      mockServer.takeRequest().body.readUtf8().let {
+        assertFalse(it.contains("inputFields(includeDeprecated: true)"))
+        assertFalse(it.contains("args(includeDeprecated: true)"))
+      }
+      assertEquals(schemaString1, File(dir, "src/main/graphql/com/example/schema.json").readText())
```
martinbonnin, line 161: 👍

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
```diff
@@ -44,26 +44,37 @@ object SchemaDownloader {
...
-        introspectionSchema = introspectionSchemaJson.toIntrospectionSchema()
+        introspectionSchemaJson = try {
+          downloadIntrospection(
+              endpoint = endpoint,
+              headers = headers,
+              insecure = insecure,
+              includeDeprecatedInputFieldsAndArguments = true,
+          )
+        } catch (e: Exception) {
+          // Maybe the server doesn't support deprecated input fields / arguments, try without them
```
martinbonnin, line 69: Note for 4.x: we can do better than this and only retry if the server returned `response.errors`. This will avoid falling back to an "older" introspection if there is a transient network error.

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/introspection/IntrospectionSchema.kt
BoD: Exactly - good catch! 👍

martinbonnin: Mmm looks like that failed in `apiCheck` because the method is marked as 'synthetic'. I _think_ this is ok though cause it's still binary compatible.

martinbonnin (approved):

PR #4704, opened 2023-02-24
Don't use data classes in IntrospectionSchema and gql

BoD:


martinbonnin:
So no `toString()`? It's easier to add them later on than remove them so works for me, I just wanted to double check.

martinbonnin (approved):

PR #4707, opened 2023-02-26
[Build scripts] Bump toolchain to Java 17

martinbonnin:
Bump toolchain to Java17 except Gradle tests that still run Java11 for testing with older Gradle versions

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
 import org.jetbrains.kotlin.samWithReceiver.gradle.SamWithReceiverExtension
 import org.jetbrains.kotlin.samWithReceiver.gradle.SamWithReceiverGradleSubplugin
 
@@ -50,9 +51,15 @@ dependencies {
   runtimeOnly(golatac.lib("kotlinx.binarycompatibilityvalidator"))
 }
 
+// Keep in sync with CompilerOptions.kt
 java {
-  // Keep in sync with CompilerOptions.kt
   toolchain.languageVersion.set(JavaLanguageVersion.of(17))
+}
+tasks.withType(JavaCompile::class.java).configureEach {
+  options.release.set(11)
+}
+tasks.withType(KotlinJvmCompile::class.java).configureEach {
+  kotlinOptions.jvmTarget = "11"
 }
 
 gradlePlugin {
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -55,7 +55,7 @@ dependencies {
 java {
   toolchain.languageVersion.set(JavaLanguageVersion.of(17))
 }
-tasks.withType(JavaCompile::class.java).configureEach {
+tasks.withType<JavaCompile>().configureEach {
   options.release.set(11)
 }
 tasks.withType(KotlinJvmCompile::class.java).configureEach {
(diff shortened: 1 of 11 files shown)


build-logic/build.gradle.kts
```diff
@@ -50,9 +51,15 @@ dependencies {
...
 
+// Keep in sync with CompilerOptions.kt
 java {
-  // Keep in sync with CompilerOptions.kt
-  toolchain.languageVersion.set(JavaLanguageVersion.of(11))
+  toolchain.languageVersion.set(JavaLanguageVersion.of(17))
+}
+tasks.withType(JavaCompile::class.java).configureEach {
+  options.release.set(11)
+}
```
lwasyl, line 60: FYI there is [Kotlin DSL syntax](https://gradle.github.io/kotlin-dsl-docs/api/org.gradle.kotlin.dsl/org.gradle.api.-domain-object-collection/index.html) with nice Kotlin generics. I'm not using .kts but I think it's available there
```suggestion
tasks.withType<JavaCompile>().configureEach {
  options.release.set(11)
}
```

build-logic/src/main/kotlin/CompilerOptions.kt
```diff
@@ -59,6 +60,15 @@ fun Project.configureJavaAndKotlinCompilers() {
   allWarningsAsErrors(true)
 }
 
+fun Project.configureTests(jvmVersion: Int) {
+  tasks.withType(Test::class.java).configureEach {
+    val javaToolchains = this@configureTests.extensions.getByName("javaToolchains") as JavaToolchainService
```
lwasyl, line 65: also here you could probably use `configure<JavaToolchainService> { }` syntax (https://gradle.github.io/kotlin-dsl-docs/api/org.gradle.kotlin.dsl/org.gradle.api.plugins.-extension-aware/configure.html)

e: ah I see it's `build-logic`. FYI you can also use Kotlin DSL extensions by adding a regular dependency  `implementation(org.gradle.kotlin.dsl.ProjectExtensionsKt.gradleKotlinDsl(project))`, and `"org.jetbrains.kotlin.plugin.sam.with.receiver"` plugin (needed to avoid the `it`s in lambdas)

lwasyl (approved):

build-logic/src/main/kotlin/CompilerOptions.kt
martinbonnin: I'm on team `no kotlin-dsl`. I know I'm the outlier there but there are enough different ways to do the same things in Gradle that I don't need a `fun String.invoke(dependencyNotation: Any)` in the classpath. 

That's a long story but I believe a good proportion of the pain Kotlin devs have using Kotlin build scripts is because we're trying to hard to make it look like a DSL when in fact it's not and the logic behind is imperative. The order of calls matters, etc..

BoD (approved):
🔧

build-logic/src/main/kotlin/CompilerOptions.kt
lwasyl: Oh, I'm similar camp but I guess with slightly different approach — I'm 100% against `gradle.kts`  and IMO if only `gradleKotlinDsl()` API existed it'd just be a nice sugar striking a good balance between `val check by tasks.existing` and using pure Java APIs and manual casting, and it would be a nice addition to build logic writte in Kotlin. Agreed it sometimes goes too far though, `String.invoke` _is_ :unamused: 

Anyway, just wanted to share that `gradleKotlinDsl` dependency trick because it took me a while to find it :slightly_smiling_face:

PR #4709, opened 2023-02-27
Runtime: Allow to register interceptors after the APQ interceptor

martinbonnin:
I wanted to register an interceptor that logs the APQ errors:

```graphql
 class ApqLoggingInterceptor: ApolloInterceptor {
      return chain.proceed(request).onEach {
        if (isPersistedQueryNotFound(it.errors)) {
          println("PersistedQuery not found")
        }
        if (isPersistedQueryNotSupported(it.errors)) {
          println("PersistedQuery not supported")
        }
      }
    }
}
```

Unfortunately, it's not possible to add it **after** the APQ interceptor because of https://github.com/apollographql/apollo-kotlin/pull/4628/commits/6f45f201a2359401beb87684c36ea96f57cf6037
This PR reverts most of it and instead throws an error early


BoD (approved):
👍

PR #4713, opened 2023-02-27
Remove builtin Compiled types

BoD:
Followup of #4702


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -374,30 +372,6 @@ fun resolveVariables(value: Any?, variables: Executable.Variables): Any? {
...
-@JvmField
-val CompiledInputValueType = ObjectType.Builder("__InputValue").build()
-
-@SharedImmutable
-@JvmField
-val CompiledEnumValueType = ObjectType.Builder("__EnumValue").build()
-
-@SharedImmutable
-@JvmField
-val CompiledDirectiveType = ObjectType.Builder("__Directive").build()
```
martinbonnin, line 399: Feels bad to remove something that has not been deprecated. Can we make those `@Deprecated` with level `ERROR` and a nice error message?

Also: we should write somewhere our deprecations rules like:

1. symbol becomes deprecated with `WARNING` level
2. after a minimum of x months (x = ~6?), it becomes deprecated with `ERROR` (so that it's still binary compatible)
3. at next major release (when?) it becomes removed.

martinbonnin (changes requested):
Please add a note to `migration/4.0.mdx` to remember the __Schema class changed package name

martinbonnin (approved):

PR #4722, opened 2023-03-02
Enable gradle-home-cache-cleanup

Goooler:
To reduce cache sizes and on CI.

https://github.com/gradle/gradle-build-action/blob/main/README.md#removing-unused-files-from-gradle-user-home-before-saving-to-cache


martinbonnin (approved):
Thanks!

PR #4725, opened 2023-03-02
Add ApolloClient.ignorePartialData

BoD:


docs/source/migration/4.0.mdx
```diff
@@ -61,14 +61,27 @@ client.subscription(MySubscription()).toFlow().collect { response ->
...
+With GraphQL it is possible to receive partial data (both `data` and `errors` are filled), but clients sometimes don't want to handle it and prefer to consider any errors as fatal. To help with this, an option `ApolloClient.ignorePartialData` has been added to force `ApolloResponse.data` to be `null` even if it wasn't null in the response, when there are errors. This can simplify error handling at the call site, e.g.:
+
+```kotlin
+val response = apolloClient.query(MyQuery()).execute()
+if (response.data == null) {
+  // There were network or GraphQL error(s)
+} else {
+  // No errors
+}
+```
```
martinbonnin, line 75: This doesn't belong in the migration guide but more in the error handling doc. It's "just" a new feature, nothing breaking.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -96,6 +96,8 @@ private constructor(
 
   fun hasErrors(): Boolean = !errors.isNullOrEmpty()
 
+  fun isSuccessful(): Boolean = exception == null && !hasErrors()
```
martinbonnin, line 99: I'm still not sure about this one. Is a "partial response" successful or not? I don't want to hurt their feelings 😅 . A partial success is still some sort of success....

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -150,19 +152,41 @@ private constructor(
...
   }
 
+  private fun <D : Operation.Data> ApolloResponse<D>.withNullData() =
+      ApolloResponse.Builder(operation = operation, requestUuid = requestUuid, data = null)
+          .errors(errors)
+          .exception(exception)
+          .extensions(extensions)
+          .addExecutionContext(executionContext)
+          .isLast(isLast)
+          .build()
```
martinbonnin, line 188: Don't we have `ApolloResponse.newBuilder()`? If we don't we should add. it

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -243,6 +267,29 @@ private constructor(
       this.useV3ExceptionHandling = useV3ExceptionHandling
     }
 
+    private var ignorePartialData: Boolean? = null
```
martinbonnin, line 270: Can we make this on `MutableExecutionContext`? I'd like to be able to override this per-query

docs/source/migration/4.0.mdx
BoD: Oops, agreed!

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
BoD: There's one but you can't change `data` with it 😅 But I can add it actually.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -132,10 +132,18 @@ private constructor(
       this.extensions = extensions
     }
 
+    fun operation(operation: Operation<D>) = apply {
+      this.operation = operation
+    }
```
martinbonnin, line 137: I don't think it makes sense to allow changing the `operation`. Once you have a `ApolloResponse<D>`, the `D` type parameter is bound to the operation so this cannot work:

```kotlin
val response = ApolloResponse<MyQuery.Data>(myQuery, uuid, data)

val myMytation = MyMutation()

// Expecting Operation<MyQuery.Data>, not MyMutation.Data
response.newBuilder().operation(myMutation)
```

BoD: Agreed 👍

PR #4726, opened 2023-03-02
Move to kotlinx.serialization

martinbonnin:
The compiler is one tiny step closer to being multiplatform!
Moshi is still around for the http-cache (because it plays nice with okio) and benchmarks (because we benchmark against it)

One nice side effect is that we can run the main build without KSP


BoD (approved):
👍

PR #4727, opened 2023-03-03
Enable build scans

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/4344#issuecomment-1451905249

Co-authored-by: Goooler<wangzongler@gmail.com>

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -19,7 +19,16 @@ pluginManagement {
 }
 
 plugins {
+  `gradle-enterprise`
   id("org.gradle.toolchains.foojay-resolver-convention") version "0.4.0"
 }
 
 apply(from = "./gradle/repositories.gradle.kts")
+
+gradleEnterprise {
+  buildScan {
+    termsOfServiceUrl = "https://gradle.com/terms-of-service"
+    termsOfServiceAgree = "yes"
+    publishAlways()
+  }
+}


settings.gradle.kts
```diff
@@ -19,7 +19,16 @@ pluginManagement {
 }
 
 plugins {
+  `gradle-enterprise`
```
Goooler, line 22:
```suggestion
  id("com.gradle.enterprise") version "3.12.4"
```

Goooler (approved):

settings.gradle.kts
martinbonnin: Thanks!

martinbonnin:
@Goooler do you know if command line arguments are captured by build scans? We're currently passing secrets [here](https://github.com/apollographql/apollo-kotlin/blob/97be096ef37dd365494cb4b55a086a9bbe65c1b8/.github/workflows/tag.yml#L24). I wouldn't want that to be in the build scans.

Goooler:
https://docs.gradle.com/enterprise/gradle-plugin/#captured_information

martinbonnin:
Yea I saw that, I just couldn't find it anywhere in the build scan....

Goooler:
From my experience, not.

martinbonnin:
Agreed, doesn't look like it is but the doc is confusing... I asked the question in the [Gradle slack](https://gradle-community.slack.com/archives/CFW1EQHNE/p1677839052150859). Let's see if we can get confirmation. If not, we'll need to move this configuration somewhere else.

martinbonnin:
@Goooler so we discussed this internally and decided to only enable the build scan on pull requests. As the CI is also doing releases and what not, this felt safer. And it felt like having build insights about sonatype upload speed wasn't really needed anyways. Let us know otherwise!

Goooler (approved):

BoD (approved):
👍

PR #4737, opened 2023-03-06
apollo-ast ergonomics

martinbonnin:
While we're on the topic of error handling, it felt like `getOrThrow()` was more Kotlin-y. No strong opinion, I'm opening this PR for discussion mainly, I'm OK to let it go 🎶

### DIFF
--- a/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/GQLResult.kt
+++ b/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/GQLResult.kt
@@ -1,5 +1,7 @@
 package com.apollographql.apollo3.ast
 
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
+
 
 /**
  * The result of a parsing or validation operation. It's tri-state:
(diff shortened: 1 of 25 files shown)


libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/GQLResult.kt
```diff
@@ -10,24 +10,29 @@ package com.apollographql.apollo3.ast
...
 ) {
 
   init {
-    check (value != null || issues.containsError()) {
+    check(value != null || issues.containsError()) {
       "Apollo: GQLResult must contain a value or an error"
     }
   }
 
+  @Deprecated("Use getOrThrow instead", replaceWith = ReplaceWith("getOrThrow"))
```
BoD, line 24: Add an `@ApolloDeprecatedSince`?

BoD (approved):
Yup, I like it 👍

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/GQLResult.kt
martinbonnin: Ayayay good catch!

PR #4738, opened 2023-03-07
Add multipart subscriptions test

martinbonnin:


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -276,4 +321,10 @@ private constructor(
       )
     }
   }
+
+  private companion object {
+    private const val EMPTY = 0
+    private const val PAYLOAD = 1
+    private const val OTHER = 2
```
martinbonnin, line 328: TIL: you still need those fields to be private even if the companion itself is private. I'm guessing because there are JVM static fields on `HttpNetworkTransport`, not sure...

martinbonnin: For the record, this is not a binary-compatibility-validator issue. Adding `const` makes the static field public:

```kotlin
class Test {
  private companion object {
    // generates a `public static final String toto = "Hello";`
    const val field = "Hello"
  }
}
```

martinbonnin: https://youtrack.jetbrains.com/issue/KT-57196

PR #4741, opened 2023-03-08
Be strict about non-lazy APIs that cannot be changed after service creation

martinbonnin:
This is a reflection on https://github.com/apollographql/apollo-kotlin/issues/4720. 

The `Service` properties fall in 2 parts:

1. the ones that change the task graph (like `generateApolloMetadata` or `outputDirConnection`) at configuration time
2. the lazy ones that are only used at task execution time

Putting both of them in the same `Service` was certainly a mistake albeit one that's hard to undo now. The properties from 1. should all be var and not `Property<>` and should certainly be in a separate block:

```kotlin
apollo {
  service("foo") {
    generateApolloMetadata = true
    packageName.set("com.example")
    outputDirConnection {
      connectToKotlinSourceSet("test")
    }

    options {
      mapScalar("Date", ...)
      generateDataBuilders.set(true)
      generateAsInternal.set(true) 
      ....
    }
  }
}
```

It's too late now to change this API but this PR:
* forbids any type-1 property to be changed after the fact
* wraps a few more type-2 properties in `provider {}` so they can be changed later on if needed


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultService.kt
```diff
@@ -116,12 +128,20 @@ abstract class DefaultService @Inject constructor(val project: Project, override
...
+    }
+
     this.operationOutputAction = action
   }
 
   var outputDirAction: Action<in Service.DirectoryConnection>? = null
 
   override fun outputDirConnection(action: Action<in Service.DirectoryConnection>) {
+    check(!registered) {
+      "Apollo: outputDirAction {} cannot be configured outside of a service {} block"
```
BoD, line 142:
```suggestion
      "Apollo: outputDirConnection {} cannot be configured outside of a service {} block"
```

BoD (approved):
🐘

PR #4750, opened 2023-03-10
Introduce a dynamic JSON parser for JS

baconz:
This change introduces a new `HttpEngine` that uses the browser's `fetch` and `json()` calls directly so that all of the JSON parsing happens in native code. The resulting `dynamic` is then passed to a new `DynamicJsJsonReader` for use by the adapters.

Of note, I also tried a version of this where I modified the existing ktor engine to call `JSON.parse()` directly with the byte buffer in the response. I found that approach to be roughly 1/2 as fast as using the native `json()` api.

I put the new http engine behind a flag, so it shouldn't break any existing behavior.

There are still performance issues with the `Call` creation, but this change speeds up the response parsing by ~4-5x.

See https://github.com/apollographql/apollo-kotlin/issues/4732

### DIFF
--- a/libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
+++ b/libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
@@ -6,10 +6,35 @@ import com.apollographql.apollo3.api.json.internal.toDoubleExact
 import com.apollographql.apollo3.api.json.internal.toIntExact
 import com.apollographql.apollo3.api.json.internal.toLongExact
 import com.apollographql.apollo3.exception.JsonDataException
-import kotlin.jvm.JvmOverloads
+
+@Suppress("NOTHING_TO_INLINE")
+private inline fun objectEntries(json: dynamic) = js("Object").entries(json).unsafeCast<Array<Array<Any>>>()
 
 /**
- * A [JsonReader] that reads data from a regular [Map<String, Any?>]
+ * By wrapping each iterator we can use introspection to determine what kind of iterator we have in the stack
+ *
+ */
+sealed interface IteratorWrapper {
+  val iterator: Iterator<*>
+
+  /**
+   * A wrapped iterator of the standard non-entry types
+   *
+   * @property iterator
+   */
+  data class StandardIterator(override val iterator: Iterator<*>) : IteratorWrapper
+
+  /**
+   * A wrapped entry of object entries (pairs of [String, Any?]) that represent the members
+   * of the container.
+   *
+   * @property iterator
+   */
+  data class EntryIterator(override val iterator: Iterator<Array<Any>>) : IteratorWrapper
+}
+
+/**
+ * A [JsonReader] that reads data from a javascript dynamic
  *
  * Map values should be any of:
  * - String
@@ -18,8 +43,8 @@ import kotlin.jvm.JvmOverloads
  * - Long
  * - JsonNumber
  * - null
- * - Map<String, Any?> where values are any of these values recursively
- * - List<Any?> where values are any of these values recursively
+ * - dynamic where values are any of these values recursively
+ * - Array<Any?> where values are any of these values recursively
  *
  * Anything else is undefined
  *
@@ -28,11 +53,10 @@ import kotlin.jvm.JvmOverloads
  * @param root the root [Map] to read from
  * @param pathRoot the path root to be prefixed to the returned path when calling [getPath]. Useful for [buffer].
  */
-class MapJsonReader
-@JvmOverloads
+class DynamicJsJsonReader
 constructor(
-    val root: Map<String, Any?>,
-    private val pathRoot: List<Any> = emptyList(),
+    val root: dynamic,
+    private val pathRoot: Array<Any> = emptyArray(),
 ) : JsonReader {
 
   private var peekedToken: JsonReader.Token
@@ -55,8 +79,8 @@ constructor(
   /**
    * The current object memorized in case we need to rewind
    */
-  private var containerStack = arrayOfNulls<Map<String, Any?>>(MAX_STACK_SIZE)
-  private val iteratorStack = arrayOfNulls<Iterator<*>>(MAX_STACK_SIZE)
+  private var containerStack = arrayOfNulls<Any?>(MAX_STACK_SIZE)
+  private val iteratorStack = arrayOfNulls<IteratorWrapper>(MAX_STACK_SIZE)
   private val nameIndexStack = IntArray(MAX_STACK_SIZE)
 
   private var stackSize = 0
@@ -68,7 +92,7 @@ constructor(
 
   private fun anyToToken(any: Any?) = when (any) {
     null -> JsonReader.Token.NULL
-    is List<*> -> JsonReader.Token.BEGIN_ARRAY
+    is Array<*> -> JsonReader.Token.BEGIN_ARRAY
     is Map<*, *> -> JsonReader.Token.BEGIN_OBJECT
     is Int -> JsonReader.Token.NUMBER
     is Long -> JsonReader.Token.LONG
@@ -76,7 +100,13 @@ constructor(
     is JsonNumber -> JsonReader.Token.NUMBER
     is String -> JsonReader.Token.STRING
     is Boolean -> JsonReader.Token.BOOLEAN
-    else -> JsonReader.Token.ANY
+    else -> {
+      if (objectEntries(any).isNotEmpty()) {
+        JsonReader.Token.BEGIN_OBJECT
+      } else {
+        JsonReader.Token.ANY
+      }
+    }
   }
 
   /**
@@ -90,19 +120,19 @@ constructor(
       return
     }
 
-    val currentIterator = iteratorStack[stackSize - 1]!!
+    val currentIteratorWrapper = iteratorStack[stackSize - 1]!!
 
     if (path[stackSize - 1] is Int) {
       path[stackSize - 1] = (path[stackSize - 1] as Int) + 1
     }
 
-    if (currentIterator.hasNext()) {
-      val next = currentIterator.next()
+    if (currentIteratorWrapper.iterator.hasNext()) {
+      val next = currentIteratorWrapper.iterator.next()
       peekedData = next
 
-      peekedToken = when (next) {
-        is Map.Entry<*, *> -> JsonReader.Token.NAME
-        else -> anyToToken(next)
+      peekedToken = when (currentIteratorWrapper) {
+        is IteratorWrapper.EntryIterator -> JsonReader.Token.NAME
+        is IteratorWrapper.StandardIterator -> anyToToken(next)
       }
     } else {
       peekedToken = if (path[stackSize - 1] is Int) {
@@ -118,7 +148,7 @@ constructor(
       throw JsonDataException("Expected BEGIN_ARRAY but was ${peek()} at path ${getPathAsString()}")
     }
 
-    val currentValue = peekedData as List<Any?>
+    val currentValue = peekedData as Array<*>
 
     check(stackSize < MAX_STACK_SIZE) {
       "Nesting too deep"
@@ -126,7 +156,7 @@ constructor(
     stackSize++
 
     path[stackSize - 1] = -1
-    iteratorStack[stackSize - 1] = currentValue.iterator()
+    iteratorStack[stackSize - 1] = IteratorWrapper.StandardIterator(currentValue.iterator())
     advanceIterator()
   }
 
@@ -149,8 +179,7 @@ constructor(
       "Nesting too deep"
     }
     stackSize++
-    @Suppress("UNCHECKED_CAST")
-    containerStack[stackSize - 1] = peekedData as Map<String, Any?>
+    containerStack[stackSize - 1] = peekedData.asDynamic()
 
     rewind()
   }
@@ -185,12 +214,13 @@ constructor(
       throw JsonDataException("Expected NAME but was ${peek()} at path ${getPathAsString()}")
     }
     @Suppress("UNCHECKED_CAST")
-    val data = peekedData as Map.Entry<String, Any?>
+    val data = peekedData as Array<Any>
 
-    path[stackSize - 1] = data.key
-    peekedData = data.value
-    peekedToken = anyToToken(data.value)
-    return data.key
+    val (key, value) = data
+    path[stackSize - 1] = key
+    peekedData = value
+    peekedToken = anyToToken(value)
+    return key as String
   }
 
   override fun nextString(): String {
@@ -379,7 +409,7 @@ constructor(
   override fun rewind() {
     val container = containerStack[stackSize - 1]
     path[stackSize - 1] = null
-    iteratorStack[stackSize - 1] = container!!.iterator()
+    iteratorStack[stackSize - 1] = IteratorWrapper.EntryIterator(objectEntries(container!!).iterator())
     nameIndexStack[stackSize - 1] = 0
     advanceIterator()
   }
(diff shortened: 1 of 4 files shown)


libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
```diff
@@ -0,0 +1,64 @@
...
+ * - Double
+ * - Long
+ * - JsonNumber
+ * - null
+ * - `dynamic` where values are any of these values recursively
+ * - Array<Any?> where values are any of these values recursively
+ *
+ * Anything else is undefined
+ *
+ * TODO Note: We can probably safely remove the non-JS primitives (ie Long, Double, JsonNumber)
```
martinbonnin, line 28: I think JS objects support those (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#data_structures_and_types). Kotlin seems to map JsonNumber (our BigInt) to `Double` but I guess this might change.

```kotlin
       val o = js("{ float: 3.14, int: 3.0, bigint: 922337203685477580700 }")

        println(o.float::class.simpleName) // "Double"
        println(o.int::class.simpleName) // "Int"
        println(o.bigint::class.simpleName) // "Double"
```

libraries/apollo-api/api/apollo-api.api
```diff
@@ -1090,40 +1097,14 @@ public abstract interface class com/apollographql/apollo3/api/json/JsonWriter :
 	public abstract fun value (Z)Lcom/apollographql/apollo3/api/json/JsonWriter;
 }
 
-public final class com/apollographql/apollo3/api/json/MapJsonReader : com/apollographql/apollo3/api/json/JsonReader {
-	public static final field Companion Lcom/apollographql/apollo3/api/json/MapJsonReader$Companion;
+public final class com/apollographql/apollo3/api/json/MapJsonReader : com/apollographql/apollo3/api/json/MapishJsonReader {
```
martinbonnin, line 1100: Pretty sure this is a breaking change we can't backport to 3.x. Are you ok living in 4.x land (`main`) for now? We were planning to start releasing alphas in late April but we could do a `-dev.1` if needed.

martinbonnin: Also +1 on the naming 😃 !

libraries/apollo-mpp-utils/src/jvmMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
```diff
@@ -14,7 +14,7 @@ actual fun currentTimeFormatted(): String {
 }
 
 actual fun currentThreadId(): String {
-  return Thread.currentThread().id.toString()
+  return Thread.currentThread().threadId().toString()
```
martinbonnin, line 17: Can we do that? This is a [Java19 only](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Thread.html) API. Is the Kotlin compiler smart enough to replace it? I would actually have expected the compiler to error here 🤔

martinbonnin: Interestingly, it fails at runtime:

```kotlin
test.ThreadTests[jvm] > cacheIsNotReadFromTheMainThread[jvm] FAILED
    java.lang.NoSuchMethodError: 'long java.lang.Thread.threadId()'
```

That should be caught at compile time. I'll investigate this...

martinbonnin: Alright, that's another instance of https://youtrack.jetbrains.com/issue/KT-57077/1.8.20-RC-243-shows-Java-19-warnings-even-if-configured-with-Java-17-toolchain, I merged main with Kotlin `1.8.20-RC` and this is correctly caught at compile time

martinbonnin: Latest IntelliJ EAPs should detect the toolchain and not show the symbol strikedthrough (in the IDE) but for some reason it's not working: https://youtrack.jetbrains.com/issue/KTIJ-24844

I reverted that change, it should compile now

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/FastBrowserJsHttpEngine.kt
```diff
@@ -0,0 +1,95 @@
...
+ * ~1/2 the speed of the native JS apollo client.
+ *
+ */
+class FastBrowserJsHttpEngine : HttpEngine {
+  private var disposed = false
+  private val controller = AbortController()
+
+  init {
+    if (!PlatformUtils.IS_BROWSER) {
+      throw RuntimeException("FastHttpEngine is not available outside of the browser.")
```
martinbonnin, line 69: Sounds like this [could work](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch) on some versions of node too?

libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  override fun getArrayIterator(container: Any): Iterator<Any?> {
+    @Suppress("UNCHECKED_CAST")
+    return (container as Array<Any?>).iterator()
+  }
+
+  override fun getMapIterator(container: dynamic): Iterator<Map.Entry<String, Any?>> {
+    return objectKeys(container).map { key -> DynamicMapEntry(key, container[key]) }.iterator()
```
martinbonnin, line 62: Pretty sure we can optimize by not going through `Iterator` and saving some `DynamicMapEntry` but that certainly means duplicating some code.

martinbonnin:
Thanks for opening this 🙏 ! A few specific comments above.  

In general, I wonder how much want to make this a core runtime API, i.e. make `ApolloClient` able to work with non-byte responses. There is a lot of functionality that relies on having bytes in `HttpResponse` (batching, multipart, HTTP cache, mockserver, maybe more). My guess is it's a bit early to go all in. 

Same thing with `MapishJsonReader`. As much as I love the name 💙 , it introduce a new abstract class in something that is a pretty low level API and I'm not 100% sure the implications of that just yet and I don't want to block this PR for bikeshedding about API design. Plus I _think_ copy/pasting would allow some JS-only optimizations.

All in all, can we start small by:
* reverting the `apollo-runtime` specific parts for now? (you'll still be able to use the dynamic reader, see below)
* leaving `MapJsonReader` untouched and duplicate the code in `DynamicJsJsonReader`?

> There are still performance issues with the `Call` creation,

If `Call` is slow, you can bypass it completely. See [this page](https://www.apollographql.com/docs/kotlin/advanced/no-runtime/) about using models without apollo-runtime.

You would do something like this:

```kotlin
val jsObject = fetch(/*...*/)

val jsonReader = DynamicJsJsonReader(jsObject)
val response = MyQuery().parseJsonResponse(jsonReader, customScalarAdapters)
```

You won't be able to use any of HTTP caching, batching, etc... but I guess this is OK?

> Of note, I also tried a version of this where I modified the existing ktor engine to call `JSON.parse()` directly with the byte buffer in the response. I found that approach to be roughly 1/2 as fast as using the native `json()` api.

Not sure what "native" is these days 😅 . What do you mean by `using the native json() api.`?

libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
baconz: Oh, interesting. Ok! I can remove the TODO.

libraries/apollo-api/api/apollo-api.api
baconz: 😄 I cribbed it from Kotlin which internally has an `isArrayish()`.

Yes, we're fine waiting for 4.x. We don't plan on releasing anything until June-ish and can work off of SNAPSHOTs until then.

libraries/apollo-mpp-utils/src/jvmMain/kotlin/com/apollographql/apollo3/mpp/utils.kt
baconz: Interesting. I was running 19 and without that change the Kotlin compiler failed on a deprecation warning.

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/FastBrowserJsHttpEngine.kt
baconz: Yeah, I think it could be fairly easily extended to work generally on node, but figured better leave that until somebody needs it.

libraries/apollo-api/src/jsMain/kotlin/com/apollographql/apollo3/api/json/DynamicJsJsonReader.kt
baconz: Ah, yes indeed. From looking at the profiles it doesn't seem like we're spending a ton of time there. Most of the time is spent in the data class construction. This change speeds up the parsing step by ~15x, so I'm not sure we need to squeeze more out of it.

Happy to make the change though if you think it is worth the tradeoff.

baconz:
All of those comments make sense. I'll take a stab at another version!

> Not sure what "native" is these days 😅 . What do you mean by using the native json() api.

I'm talking about `Response.json()` (https://developer.mozilla.org/en-US/docs/Web/API/Response/json). With this API the response never reaches the JS runtime until it has been parsed, so the parse is truly "native" (C++ code).

With all of that said, what would you think about providing an `expect` utility function that wraps all of this functionality?

To your point about it being early to go all-in on KMM/KJS, I don't want to make its failure a self-fulfilling prophecy. We should make it easy for others to come in and get something moderately performant up and running without understanding too many library internals.

I'm imagining maybe a new library like `apollo-js-wrapper` like:

```kotlin
expect suspend fun <D : Query.Data> wrapQuery(query: Query<D>, apolloClient: ApolloClient): ApolloResponse<D>
expect suspend fun <D : Mutation.Data> wrapMutation(mutation: Mutation<D>, apolloClient: ApolloClient): ApolloResponse<D>
```

The default implementaiton could just be `apolloClient.execute()` and the JS implementation could be something like:

```kotlin
suspend fun fetchToJson(url: String, requestInit: RequestInit): dynamic {
  // basically what I had in `FastBrowserJsHttpEngine`
}

suspend fun <D : Query.Data> wrapQuery(query: Query<D>, apolloClient: ApolloClient): ApolloResponse<D> {
  val body = buildJsonString {
    query.composeJsonRequest(this, customScalarAdapters)
  }
  val d = fetchToJson(...)
  // error handling etc
  val reader = DynamicJsJsonParser(d)
  return query.parseFromJson(reader)
}

 ```

File: libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/FastBrowserJsHttpEngine.kt (hunk 4 above)
Comment (martinbonnin) [reply libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/FastBrowserJsHttpEngine.kt:69]: Makes sense 👍

File: libraries/apollo-mpp-utils/src/jvmMain/kotlin/com/apollographql/apollo3/mpp/utils.kt (hunk 3 above)
Comment (martinbonnin) [reply libraries/apollo-mpp-utils/src/jvmMain/kotlin/com/apollographql/apollo3/mpp/utils.kt]: We've been running the Kotlin `-RC` versions and the point you branched from had a regression in the Kotlin compiler (KT-57077). It's all good now!

**Comment by martinbonnin** [comment]
> I'm talking about Response.json()

👍 Thanks for the link!

> The default implementaiton could just be apolloClient.execute() and the JS implementation could be something like:

```
suspend fun <D : Query.Data> wrapQuery(query: Query<D>, apolloClient: ApolloClient): ApolloResponse<D> {
  val body = buildJsonString {
    query.composeJsonRequest(this, customScalarAdapters)
  }
  val d = fetchToJson(...)
  // error handling etc
  val reader = DynamicJsJsonParser(d)
  return query.parseFromJson(reader)
}
```

My issue is that this breaks the `ApolloClient` contract. If there's an `ApolloClient` parameter, the caller can expect that stuff like `executionContext`, APQs, timeouts, interceptors, HTTP cache, normalized cache and more get honoured. The `ApolloClient` API is a lot more than HTTP fetch and bypassing everything for JS sounds like a lot of future troubles.

>  We should make it easy for others to come in and get something moderately performant up and running without understanding too many library internals.

For what it's worth, the APIs used in the [no-runtime doc](https://www.apollographql.com/docs/kotlin/advanced/no-runtime/) are all public APIs. If anything, I think I'd like to use this opportunity to improve the ergonomics of that. It might be useful for other non-JS users too.

All in all, thanks a lot for looking into this! This is cool stuff! Sorry it's the end of a big week here and my brain is a bit fried  😄 I don't want us to commit to something that's going to cost in maintenance in the long run. Let me recharge a bit over the week end and maybe I'll see this with more clarity on Monday!

**Comment by martinbonnin** [comment]
> I like the simplicity of that! I guess the issue becomes configurability 

Agreed. 

> 1. Introduce a new HttpConfiguration object that we pass into the functions where we can put things like headers/cookies (maybe even interceptors some day!?).

Not a fan of this. That'll end up duplicating most of `ApolloClient`.

> 2. Make these extension functions on top of HttpEngine (or pass an HttpEngine). The trouble there is that we're back to our old friend HttpRespnse, and needing to customize it for dynamic.

As you found out, I'm a bit wary of touching `HttpReponse` as that changes the contract of a lot of things (no more bytes-only). 


**New proposal!**: Make it work at the `NetworkTransport` layer? This API is very small:

```kotlin
interface NetworkTransport {

  /**
   * Execute a request.
   *
   * This function must not throw in normal circumstances. Instead, emit an [ApolloResponse] with a non null [ApolloResponse.exception].
   * It may throw if the error is the result of a programming error like wrongly configured interceptors.
   *
   * @param request the request to execute.
   * @return a flow of responses.
   */
  fun <D : Operation.Data> execute(
      request: ApolloRequest<D>,
  ): Flow<ApolloResponse<D>>

  fun dispose()
}
```

It could use its own HTTP fetching mecanism that can read `request.headers` and configure the underlying HTTP code? 

From the caller perspective, it's an extra call:

```kotlin
                           // a network transport that consumes JS objects as dynamic instead
                           // of reading them byte by byte to make parsing faster on JS
                           .useDynamicNetworkTransport("https://example.com/graphql)
                           .build()
```

Anyone trying to use `HttpInterceptors` would get an exception because we check it [there](https://github.com/apollographql/apollo-kotlin/blob/23eaeba2ac8cf0331aaed2bb55a9ed56e3348147/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt#L497). Same thing for HTTP cache, etc... these will all be forbidden at client construction time.

Of course that means you'll have to go through the price of making an `ApolloCall` but sounds like this is OK?

Or (New Proposal #3) if we say we don't want to use `ApolloClient`, then doing the HTTP call can be the responsability of the caller and then we add easy API to parse the response:


```kotlin
val operation = MyQuery(myVariables)

// toRequestBody() is in Apollo Kotlin
val postBody: String = operation.toRequestBodyString() 

// executeHttpCall() is outside Apollo Kotlin
val networkResponse: dynamic = executeHttpCall() 

// toApolloResponse() is in Apollo Kotlin
// if response is a BufferedSource, it uses the existing JsonReaders
// if response is a dynamic (not sure how to detect this reliably, maybe it's just the fallback case), it uses the faster JsonReader
val apolloResponse = networkResponse.toApolloResponse(operation, customScalarAdapters)
```

martinbonnin:
@baconz many thanks for updating this 💙 
As discussed async, this PR now contains the minimal changes to support `DynamicJsJsonReader` and we decided to postpone the bigger `apollo-runtime` changes to a later point where we have a better understanding of what makes JS parsing so slow.

Thanks again for looking into this! @BoD can you take a look?

martinbonnin (approved):

BoD (approved):
👍 Very cool. Thanks a lot @baconz!

martinbonnin:
Thanks!

PR #4763, opened 2023-03-16
Add release-3.x to GH workflows

BoD:


.github/workflows/docs-publish.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - release-2.x
+      - release-3.x
```
martinbonnin, line 8: I think what this does is that it launches  a netlify build on pushes to 3.x so we don't have to wait for something to change on main. And the netlify build checks out `release-3.x` alongside `main` and `release-2.x`. It's good to have it in all cases 👍

.github/workflows/push.yml
```diff
@@ -2,7 +2,9 @@ name: push
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
+      - release-3.x
```
martinbonnin, line 7: 👍

martinbonnin (approved):

BoD:
Looks like `tests/kotlin-js-store-1.7/yarn.lock` was out of date. Not really related to this PR, but build fails without it.

PR #4768, opened 2023-03-17
Multipart subscriptions: set the correct accept header

martinbonnin:


libraries/apollo-api/api/apollo-api.api
```diff
@@ -864,7 +864,6 @@ public final class com/apollographql/apollo3/api/http/ByteStringHttpBody : com/a
 public final class com/apollographql/apollo3/api/http/DefaultHttpRequestComposer : com/apollographql/apollo3/api/http/HttpRequestComposer {
 	public static final field Companion Lcom/apollographql/apollo3/api/http/DefaultHttpRequestComposer$Companion;
 	public static final field HEADER_APOLLO_OPERATION_ID Ljava/lang/String;
-	public static final field HEADER_APOLLO_OPERATION_NAME Ljava/lang/String;
```
martinbonnin, line 867: FWIW, I'm team "drop const" in public APIs now. If a newer version of the lib is pulled during dependency resolution, I want  the new value of constants to stay in sync, it'd be quite surprising else.

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -289,8 +292,17 @@ private constructor(
       this.exposeErrorBody = exposeErrorBody
     }
 
+    fun addHttpHeader(name: String, value: String) = apply {
+      headers.add(HttpHeader(name, value))
+    }
+
     fun httpHeaders(headers: List<HttpHeader>) = apply {
-      interceptors.add(HeadersInterceptor(headers))
+      // In case this builder comes newBuilder() remove any existing interceptor
```
BoD, line 300:
```suggestion
      // In case this builder comes from newBuilder(), remove any existing interceptor
```

BoD (approved):
👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -36,11 +36,16 @@ class DefaultHttpRequestComposer(
     val operation = apolloRequest.operation
     val customScalarAdapters = apolloRequest.executionContext[CustomScalarAdapters] ?: CustomScalarAdapters.Empty
 
-    val requestHeaders = listOf(
-        HttpHeader(HEADER_APOLLO_OPERATION_ID, operation.id()),
-        HttpHeader(HEADER_APOLLO_OPERATION_NAME, operation.name()),
-        HttpHeader(HEADER_ACCEPT_NAME, HEADER_ACCEPT_VALUE),
-    ) + (apolloRequest.httpHeaders ?: emptyList())
+    val requestHeaders = mutableListOf<HttpHeader>().apply {
```
BoD, line 39: You probably know it already but just in case - there's `buildList {}` :)

martinbonnin: Nope, we're compatible with very old kotlin-stdlib. We can't have nice things!

BoD: 😞  Oh well 😅

Issue #4780, opened 2023-03-21
DiskReadViolation when using the Disk Normalized Cache


agrosner:
### Version

3.7.4

### Summary

when injecting dagger code to initialize `ApolloClient` like:
```kotlin

 @Provides
    @PerApplication
    fun memoryDiskApollloClient(
        apolloClient: ApolloClient,
        @MemoryCache memoryCacheFactory: NormalizedCacheFactory,
        @DiskCache diskCacheFactory: NormalizedCacheFactory,
        cacheKeyGenerator: CacheKeyGenerator
    ): ApolloClient = apolloClient.newBuilder()
        .logCacheMisses(
            getCacheMissLogger(PlatformApolloContextualInfo.FederatedClientMemoryAndDiskCacheMiss)
        )
        .normalizedCache(
            memoryCacheFactory
                .chain(diskCacheFactory),
            cacheKeyGenerator = cacheKeyGenerator
        )
        .build()
```

We receive a disk read violation error when in strict mode. 

### Steps to reproduce the behavior

Standard initialize with dagger or on main thread.

### Logs

<img width="1101" alt="Screenshot 2023-03-21 at 9 48 47 AM" src="https://user-images.githubusercontent.com/2117859/226631626-1939a6fb-71ae-425d-9a0e-8e4004651ca8.png">

martinbonnin:
Fixed in 3.8.0

PR #4781, opened 2023-03-21
Move cache creation to be lazy

BoD:
Related to #4780

### DIFF
--- a/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
+++ b/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
@@ -36,7 +36,11 @@ internal class DefaultApolloStore(
 
   override val changedKeys = changedKeysEvents.asSharedFlow()
 
-  private val cache: OptimisticCache = OptimisticCache().chain(normalizedCacheFactory.createChain()) as OptimisticCache
+  // Keeping this as lazy to avoid accessing the disk at initialization which usually happens on the main thread
+  private val cache: OptimisticCache by lazy {
+    OptimisticCache().chain(normalizedCacheFactory.createChain()) as OptimisticCache
+  }
+
   private val lock = Lock()
 
   override suspend fun publish(keys: Set<String>) {
--- a/tests/integration-tests/src/commonTest/kotlin/test/NormalizedCacheThreadingTest.kt
+++ b/tests/integration-tests/src/commonTest/kotlin/test/NormalizedCacheThreadingTest.kt
@@ -1,0 +1,40 @@
+package test
+
+import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.cache.normalized.api.MemoryCacheFactory
+import com.apollographql.apollo3.cache.normalized.api.NormalizedCache
+import com.apollographql.apollo3.cache.normalized.api.NormalizedCacheFactory
+import com.apollographql.apollo3.cache.normalized.normalizedCache
+import com.apollographql.apollo3.integration.normalizer.CharacterNameByIdQuery
+import com.apollographql.apollo3.mpp.currentThreadName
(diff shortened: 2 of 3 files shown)


tests/integration-tests/src/commonTest/kotlin/test/NormalizedCacheThreadingTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    var cacheCreateThreadName: String? = null
+    val apolloClient = ApolloClient.Builder()
+        .networkTransport(QueueTestNetworkTransport())
+        .normalizedCache(object : NormalizedCacheFactory() {
+          override fun create(): NormalizedCache {
+            cacheCreateThreadName = currentThreadName()
+            return MemoryCacheFactory().create()
+          }
+        }).build()
+    assertNull(cacheCreateThreadName)
```
martinbonnin, line 32: Nice test!

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
```diff
@@ -36,7 +36,11 @@ internal class DefaultApolloStore(
 
   override val changedKeys = changedKeysEvents.asSharedFlow()
 
-  private val cache: OptimisticCache = OptimisticCache().chain(normalizedCacheFactory.createChain()) as OptimisticCache
+  // Keeping this as lazy to avoid accessing the disk at initialization which usually happens on the main thread
+  private val cache: OptimisticCache by lazy {
+    OptimisticCache().chain(normalizedCacheFactory.createChain()) as OptimisticCache
```
martinbonnin, line 41: Is lazy thread safe? I remember reading something or some rumour saying `lazy` was generating a lot of bytecode. Might be worth making it a lateninit and use regular locks?

BoD: By default `by lazy` uses a [synchronized](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt#L69) block on the JVM and on native it [looks](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/concurrent/Lazy.kt#L16) like it's using a Lock.

For the bytecode yes it can be "verbose" (👀  [this](https://youtu.be/Ta5wBJsC39s?t=1173) from 2019!)

Pushed a commit doing this manually LMKWYT!

martinbonnin: Looks probably ok to keep `lazy` here according to Chet guidance. It's quite an heavy object and allocation is done only once so both work for me!

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/DefaultApolloStore.kt
```diff
@@ -36,7 +36,20 @@ internal class DefaultApolloStore(
...
+  private var _cache: OptimisticCache? = null
+  private val cache: OptimisticCache
+    get() {
+      if (_cache == null) {
+        lock.write {
+          if (_cache == null) {
+            _cache = OptimisticCache().chain(normalizedCacheFactory.createChain()) as OptimisticCache
+          }
+        }
+      }
```
martinbonnin, line 49: That doesn't work though. The whole `if {}` needs to be locked or you could create 2 caches

martinbonnin: Ah never mind, there's double if !

martinbonnin (approved):

Issue #4784, opened 2023-03-22
Fragments are skipped when the if condition is a variable with a false default value


BoD:
### Version

3.7.5

### Summary

When using `@skip(if: $var)` on a fragment where $var has a `false` default value, the fragment should not be skipped in the parser, but it is.

### Steps to reproduce the behavior

```graphql
query SkipFragmentWithDefaultToFalse($skipIf: Boolean = false) {
  animal {
    ...dogFragment @skip(if: $skipIf)
  }
}
```

```kotlin
val operation = SkipFragmentWithDefaultToFalseQuery()
val response = operation.parseJsonResponse("""
  {
    "data": {
      "animal": {
        "__typename": "Dog",
        "barf": "ouaf"
      }
    }
  }
""")
println(response.dataAssertNoErrors)
```

Prints:
```
Data(animal=Animal(__typename=Dog, dogFragment=null))
```

but should print:
```
Data(animal=Animal(__typename=Dog, dogFragment=DogFragment(barf=ouaf)))
```

Thanks to @rohandhruva for reporting this!

martinbonnin:
Fixed in 3.8.0

PR #4785, opened 2023-03-22
Consider variable default values with @skip/@include/@defer

BoD:
Related to #4784

In the response adapters, we generate lines like this:

```kotlin
if (not(variable("skipIf")).evaluate(customScalarAdapters.adapterContext.variables(), __typename, customScalarAdapters.adapterContext, null)) {
  // Read fields for the fragment
}
```

but `customScalarAdapters.adapterContext.variables()` was empty, since it only considered variables that are not optional, or optional *and* present.

with this approach we include the optionals that are absent and have a default value (booleans only)


BoD:
Just pushed:
- use variable adapters instead of introducing a new `variables()` method (and use a flag that piggybacks on `CustomScalarAdapters.AdapterContext`)
- revert what was done in #4700 for the case of the normalizer - the new approach works for it too

martinbonnin (approved):
👏

BoD:
⚠️ Just a reminder that tweaks will be needed to avoid breaking changes when backporting to 3.x

PR #4792, opened 2023-03-24
Backport #4785 (skip-variable-default-false) to v3 branch

BoD:
Backport #4785 to v3 branch


martinbonnin (approved):

BoD:
Following [discussion](https://github.com/apollographql/apollo-kotlin/pull/4794#discussion_r1147552732), I reverted the revert (😅) and added a test, LMK if you had something different in mind

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Executables.kt
```diff
@@ -2,19 +2,15 @@
...
 
 @Suppress("UNCHECKED_CAST")
 fun <D : Executable.Data> Executable<D>.variables(customScalarAdapters: CustomScalarAdapters): Executable.Variables {
-  val valueMap = MapJsonWriter().apply {
-    beginObject()
-    serializeVariables(this, customScalarAdapters.serializeVariablesWithDefaultBooleanValues())
-    endObject()
-  }.root() as Map<String, Any?>
-  return Executable.Variables(valueMap)
+  return variables(customScalarAdapters, true)
```
martinbonnin, line 13: SHould be a `false` here to not send the `false` over the network? 

```suggestion
  return variables(customScalarAdapters, false)
```

BoD: For the network, `serializeVariables` is called directly (with a `CustomScalarAdapters` that has it to false), right?

martinbonnin: If you're using `apollo-runtime` yes but if you're not, I don't know what people are doing, I'd keep the behaviour without booleans for public methods. It'd be surprising to have default booleans in there

BoD: Ah yes that makes sense! So one last (hopefully!) tweak.

martinbonnin (approved):

PR #4793, opened 2023-03-24
Fix minSdk for compose support

BoD:
Publishing snapshots [failed](https://github.com/apollographql/apollo-kotlin/actions/runs/4510211540/jobs/7940894794) because of it. Not sure why I'm not seeing it locally though 🤔


libraries/apollo-compose-paging-support/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ android {
   compileSdk = golatac.version("android.sdkversion.compile").toInt()
 
   defaultConfig {
-    minSdk = golatac.version("android.sdkversion.min").toInt()
+    minSdk = 21
```
martinbonnin, line 18: Put this in librairies.toml, next to the other ones?

martinbonnin:
>  Not sure why I'm not seeing it locally though

If you're building from the IDE, AGP injects the minSdk of your connected device for some tasks. Might be an explanation, you can disable it in settings: Optimize build for target device
Screenshot 2023-03-24 at 12 36 52

BoD:
No I'm on the CLI 🤔! That's a cool setting to know about though 👍!

PR #4800, opened 2023-03-27
With ignorePartialData, make data and exception mutually exclusive

martinbonnin:


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -182,10 +182,10 @@ private constructor(
         .let { flow ->
           if (request.ignorePartialData == true) {
             flow.map { response ->
-              if (response.data != null && response.hasErrors()) {
-                response.newBuilder().data(null).build()
-              } else {
-                response
+              when {
+                response.hasErrors() -> response.newBuilder().data(null).exception(ApolloException("Partial data. Read errors to get more details.")).build()
```
BoD, line 186: This contradicts the KDoc of `exception` which says `If exception is non null, data and errors will be null`. Maybe that's ok and we can remove this stipulation from the doc?

Also nit, for the message, data could have been null in which case it not really "partial"?

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -182,10 +182,10 @@ private constructor(
...
+                response.data == null -> response.newBuilder().data(null).exception(ApolloException("Not data found")).build()
```
BoD, line 187:
```suggestion
                response.data == null -> response.newBuilder().exception(ApolloException("No data found")).build()
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: Something else I considered was doing `ApolloGraphQLException(val errors: List<Error>)` but then I'm contradicting myself from https://github.com/apollographql/apollo-kotlin/issues/4787

* 4787 is about taking something that is modeled as an error (`ApqNotSupportedException`) and turning into normal operation (`ApolloResponse. serverApqEnabled `)
* 4800 would be about taking something that is modeled as normal operation (`ApolloResponse.errors`) and turning it into an error `ApolloGraphQLError`

So if we're doing it for errors because it's convenient, we might as well do it for APQs

BoD: Yes that looks good!

martinbonnin: But I'm not sure it's any better. The [Confetti code](https://github.com/joreilly/Confetti/blob/fcfcf0d7e42f69416d1e3278e1f68016e0272fa2/shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SessionsViewModel.kt#L85) looks like this:

```kotlin
        if (
            sessionsData == null || bookmarksData == null
        ) {
            // Here one has to remember to check exceptions and errors
            // also the (unlikely) case where data is null (or not sent) would not produce any log
            // which might come as a surprise.
            bookmarksResponse.exception?.printStackTrace()
            sessionsResponse.exception?.printStackTrace()
            bookmarksResponse.errors?.let { println(it) }
            sessionsResponse.errors?.let { println(it) }
            return SessionsUiState.Error
        }
```

BoD: Sorry I missed that comment before!

Thinking more about it, I actually like the `ApolloGraphQLException(val errors: List<Error>)` you mentioned before would be cool, but maybe it would warrant renaming `ignorePartialData` to `treatGraphQLErrorsAsException` or something similar? If you set it to true, it means you never want to deal with GraphQL errors, even if there's partial data. So if they happen it's an exception.

```
when {
  response.hasErrors() -> response.newBuilder().data(null).exception(ApolloGraphQLException(response.errors)).build()
  response.data == null -> response.newBuilder().exception(ApolloException("No data found")).build()
  else -> response  
}
```

Maybe? 😅

martinbonnin: Yea, I like this better too 👍

martinbonnin:
Superseded by #4836

PR #4804, opened 2023-03-27
Add % to reserved characters to encode in URL

slinstacart:
Currently the library is not encoding the '%' character in GET parameters, i.e if I send a parameter string of '80%' this isn't parsed properly since it needs to be encoded to '%25'

This is a bug in UrlEncode.kt where it looks like it was introduced here in this [PR](https://github.com/apollographql/apollo-kotlin/commit/aa17db584e0b7d6f273a3894237882fd2770d04a#diff-c07ae4e7ec62cfe5701f2fbaf63b5122f0505d001d01671b8a0a60f6c0600184) which was originally encoding every character.


martinbonnin:
Just in time for next release, thanks!

PS: we might want to switch to a full-blown uri library like https://github.com/eygraber/uri-kmp/

martinbonnin (approved):

BoD (approved):
Thanks! 👍

PR #4806, opened 2023-03-27
Allow custom SqlDriver

hbmartin:
Implementation for discussion at https://github.com/apollographql/apollo-kotlin/issues/4504

Publicly exposes `SqlNormalizedCacheFactory` constructors which take a `SqlDriver` argument.

### DIFF
--- a/libraries/apollo-normalized-cache-sqlite-incubating/build.gradle.kts
+++ b/libraries/apollo-normalized-cache-sqlite-incubating/build.gradle.kts
@@ -49,13 +49,13 @@ kotlin {
 
     findByName("jvmMain")?.apply {
       dependencies {
-        implementation(golatac.lib("sqldelight.jvm"))
+        api(golatac.lib("sqldelight.jvm"))
       }
     }
 
     findByName("appleMain")?.apply {
       dependencies {
-        implementation(golatac.lib("sqldelight.native"))
+        api(golatac.lib("sqldelight.native"))
       }
     }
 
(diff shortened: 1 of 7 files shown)


libraries/apollo-normalized-cache-sqlite-incubating/build.gradle.kts
```diff
@@ -49,13 +49,13 @@ kotlin {
 
     findByName("jvmMain")?.apply {
       dependencies {
-        implementation(golatac.lib("sqldelight.jvm"))
+        api(golatac.lib("sqldelight.jvm"))
```
martinbonnin, line 52: Can you try adding `api("app.cash.sqldelight:runtime")` to the `commonMain` source set instead of each individual platform? `SqlDriver` is in the `runtime` artifact (which is added automatically by the plugin) so there should be no need to leak other things.

martinbonnin (approved):

libraries/apollo-normalized-cache-sqlite-incubating/build.gradle.kts
hbmartin: @martinbonnin Done! Used the version catalog + golatac for the reference, does that work?
re: incubating, I think it makes sense? Happy to put it wherever you think best.

martinbonnin: Thanks for this, looks good 👍 . FWIW, the golatac stuff is a workaround for https://github.com/gradle/gradle/issues/22095. We've seen some tasks being rerun due to version catalog accessors. Maybe it's better in newer versions of Gradle, I'll double check.

Re: incubating, incubating is easier for me because the API is still experimental so we can change it more easily. Let's start with this and if people ask for it in the main artifact, I'll backport.

Issue #4744, opened 2023-03-09
Use Apollo Kotlin inside Apollo Kotlin


martinbonnin:
### Use case

In order to validate introspection queries we could dogfood Apollo Kotlin from the Gradle plugin

PR #4808, opened 2023-03-28
Use Apollo to download introspection schema

BoD:
See #4744


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -375,37 +375,37 @@ fun resolveVariables(value: Any?, variables: Executable.Variables): Any? {
   }
 }
 
-@Deprecated("Introspection types are now generated like other types. Use the generated class instead.", level = DeprecationLevel.ERROR)
+@Deprecated("Introspection types are now generated like other types. Use the generated class instead.")
```
martinbonnin, line 378: Why not ERROR any more?

BoD: Woops sorry I should have commented yes! This is because these are still used in the code generated by 3.x. With the error level it wouldn't compile. With this change, it compiles with a warning (I had to use `allWarningsAsErrors(false)` because of that - which is a shame but can be removed later when we use 4.x).

libraries/apollo-gradle-plugin/src/test/kotlin/com/apollographql/apollo3/gradle/test/DownloadSchemaTests.kt
```diff
@@ -162,6 +162,34 @@ class DownloadSchemaTests {
     }
   }
 
+  @Test
+  fun `schema is downloaded correctly when server doesn't support deprecated input fields and arguments nor isRepeatable on directives`() {
+    withSimpleProject(apolloConfiguration = apolloConfiguration) { dir ->
```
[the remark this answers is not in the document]
BoD: good point! 👍

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,52 @@
...
+  }
+  service("graphql-Draft") {
+    packageName.set("com.apollographql.apollo3.tooling.graphql.draft")
+    sourceFolder.set("graphql/Draft")
+    generateAsInternal.set(true)
+  }
+  service("apollo") {
+    packageName.set("com.apollographql.apollo3.tooling.apollo")
+    sourceFolder.set("apollo")
+    generateAsInternal.set(true)
```
martinbonnin, line 40: Nit what about "platformApi" instead of "apollo" ? We have so much "apollo" everywhere...

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,52 @@
...
+
+// Using a published version of the plugin but project dependencies, so checkApolloVersions would complain
+tasks.configureEach {
+  if (name == "checkApolloVersions") {
+    enabled = false
+  }
+}
+
+// Code generated with 3.x uses classes which are deprecated in current apollo-api
+allWarningsAsErrors(false)
```
martinbonnin, line 52: Can you use the 3.x runtime instead?

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaHelper.kt
```diff
@@ -38,7 +36,29 @@ internal object SchemaHelper {
...
+    val composer = DefaultHttpRequestComposer(endpoint)
+    val apolloRequest = ApolloRequest.Builder(query)
+        .httpHeaders(headers.map { HttpHeader(it.key, it.value) })
+        .build()
+    val httpRequest = composer.compose(apolloRequest)
+    val httpEngine = DefaultHttpEngine(newOkHttpClient(insecure))
+    val httpResponse = runBlocking { httpEngine.execute(httpRequest) }
+    val bodyStr = httpResponse.body?.use {
+      it.readUtf8()
+    }
```
martinbonnin, line 54:
```suggestion
    val bodyStr = httpResponse.body?.string()
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
martinbonnin: See below, this should work if using the same runtime version as the version of the codegen

BoD: Yes looks like it's working 🎉 (seems to be confusing IJ though, but maybe it's a fluke)

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaHelper.kt
BoD: Can't find this one, is it an extension in Apollo?

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,52 @@
 plugins {
   id("org.jetbrains.kotlin.jvm")
   id("apollo.library")
+
+  // Depend on a published version of the plugin to avoid a "chicken and egg" problem
+  id("com.apollographql.apollo3").version("3.7.5")
```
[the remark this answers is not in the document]
BoD: Not sure how to do it in plugin declaration :( Is it possible with golatac?

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaHelper.kt
martinbonnin: Nevermind, I though we were in OkHttp land 👍 (although we might put it in ApolloClient 👀 )

libraries/apollo-tooling/build.gradle.kts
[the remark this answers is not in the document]
martinbonnin: I think putting it next to `kotlin-gradle-plugin` in `build-logic/build.gradle.kts` should work

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,63 @@
...
+    sourceFolder.set("graphql/October2021")
+    generateAsInternal.set(true)
+  }
+  service("graphql-Draft") {
+    packageName.set("com.apollographql.apollo3.tooling.graphql.draft")
+    sourceFolder.set("graphql/Draft")
+    generateAsInternal.set(true)
+  }
+  service("platformApi") {
+    packageName.set("com.apollographql.apollo3.tooling.platformApi")
```
martinbonnin, line 40: Maybe use lower case for package names? Not sure if it is required but I find upper case surprising.

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,63 @@
...
+  if (name == "checkApolloVersions") {
+    enabled = false
+  }
+}
+
+// Within the 'tests' project (a composite build), dependencies are automatically substituted to use the project's one.
+// But here we want to keep the published version of apollo-api when compiling the generated models.
+// So, disable the substitution rules (see https://docs.gradle.org/current/userguide/composite_builds.html#deactivate_included_build_substitutions).
+configurations.all {
+  resolutionStrategy.useGlobalDependencySubstitutionRules.set(false)
```
martinbonnin, line 62: Add to every library inside the convention plugin?

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -1,16 +1,63 @@
...
+    generateAsInternal.set(true)
+    mapScalarToKotlinString("GraphQLDocument")
+    introspection {
+      endpointUrl.set("https://graphql.api.apollographql.com/api/graphql")
+      schemaFile.set(file("src/main/graphql/platformApi/schema.graphqls"))
+    }
+  }
+}
+
+// Using a published version of the plugin but project dependencies, so checkApolloVersions would complain
```
martinbonnin, line 51:
```suggestion
// We're using project(":apollo-compiler") and the published "apollo-runtime" which do not have the same version
// TODO: exclude apollo-compiler from the version check as it's not used at runtime.
```

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/InspectionQueryTest.kt
```diff
@@ -100,16 +99,115 @@ class IntrospectionQueryTest {
...
+            }
+          }
+        }
+      }
+    }""".trimIndent()
+
   private fun String.normalized(): String {
     return Buffer().writeUtf8(this).use {
-      it.parseAsGQLDocument(null).getOrThrow().toUtf8()
+      it.parseAsGQLDocument(null).getOrThrow().toUtf8().trimIndent()
```
martinbonnin, line 205: Is the trimIndent() required here?

martinbonnin: Ah, I see below why it's required. I think if it's called `normalize`, the expectation is to normalize both the expected and actual results. If we want to compare formating and whitespaces, we should call this function "format". But this sounds more fragile

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/SchemaDownloaderTests.kt
```diff
@@ -0,0 +1,184 @@
...
+    mockServer.takeRequest().body.utf8().let {
+      assertFalse(it.contains("inputFields(includeDeprecated: true)"))
+      assertFalse(it.contains("args(includeDeprecated: true)"))
+      assertTrue(it.contains("isRepeatable"))
+    }
+    mockServer.takeRequest().body.utf8().let {
+      assertFalse(it.contains("inputFields(includeDeprecated: true)"))
+      assertFalse(it.contains("args(includeDeprecated: true)"))
+      assertFalse(it.contains("isRepeatable"))
+    }
```
martinbonnin, line 181: Nice test!

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaHelper.kt
```diff
@@ -38,7 +36,29 @@ internal object SchemaHelper {
...
+    check(httpResponse.statusCode in 200..299 && bodyStr != null) {
+      "Cannot get schema from $endpoint: ${httpResponse.statusCode}:\n${bodyStr ?: "(empty body)"}"
+    }
+    return bodyStr
```
martinbonnin, line 58: Parse `bodyStr` here to make sure the server response is valid?

BoD: It's done upstream in `SchemaDownload.download()`, but I agree it makes sense to do it here 👍

BoD: Good point! Used the generated parser in the last commit.

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/InspectionQueryTest.kt
BoD: Yeah it didn't really make sense 😅 Just pushed comparing parsed with original json, but ignoring spaces

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaDownloader.kt
```diff
@@ -158,12 +158,6 @@ object SchemaDownloader {
       headers: Map<String, String>,
       insecure: Boolean,
   ): String {
-    val responseString = SchemaHelper.executeQuery(
-        query = DownloadSchemaQuery(graphID = graph, variant = variant),
-        endpoint = endpoint,
-        headers = headers + mapOf("x-api-key" to key),
-        insecure = insecure
-    )
```
martinbonnin: good catch

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/IntrospectionQueryTest.kt
```diff
@@ -99,115 +99,16 @@ class IntrospectionQueryTest {
...
       it.parseAsGQLDocument(null).getOrThrow().toUtf8().trimIndent()
     }
   }
 
+  private fun String.normalized() = this.trimIndent().replace(Regex("\\s+"), " ")
+
   @Test
   fun canUseTheSameIntrospectionQueryAsVersion3() {
-    assertEquals(normalizedV3IntrospectionQuery, v3IntrospectionQuery.normalized())
+    assertEquals(v3IntrospectionQuery.normalized(), v3IntrospectionQuery.asParsedGQLDocument().normalized())
```
martinbonnin, line 112: I'm confused. this compares v3 to v3. Shouldn't it include `IntrospectionQuery.QUERY_DOCUMENT` somewhere?

BoD: Yeah this tests that the v3 query can still be parsed but that's not very useful. If we compare the v3 introspection query to IntrospectionQuery.OPERATION_DOCUMENT, they're a bit different though (order of the fragments + some added `__typename`).

BoD: The test should make more sense now 😅. The v3 query is the same as the June2018 after normalization.

martinbonnin (approved):

PR #4809, opened 2023-03-28
Streaming logging interceptor

martinbonnin:
Useful for multipart subscriptions, defer, etc.. I didn't do the "request body" part because I'm missing a test case but it should probably be done by wrapping the sink at some point too.


BoD (approved):
Very cool! 👍

PR #4813, opened 2023-03-29
Make apollo-compiler not depend on apollo-api

BoD:
This duplicates `BooleanExpression` (minus the bits only needed at runtime).


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -0,0 +1,137 @@
+package com.apollographql.apollo3.compiler.ir
```
martinbonnin, line 1: Package name doesn't match file path

martinbonnin (approved):
I like this change. I think ultimately we should try to optimize the `apollo-api` Boolean expression to minimize the runtime impact. I think you should already be able to remove some of the functionality in `apollo-api` like `simplify`?

BoD:
> I think you should already be able to remove some of the functionality in apollo-api like simplify?

Yes there's a bunch of unused functions but they're public 😅 I don't expect many people use them though - but least we can deprecate them!

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/BooleanExpression.kt
```diff
@@ -12,20 +14,25 @@ import kotlin.reflect.KClass
  * other that may also contain possibleTypes
  */
 sealed class BooleanExpression<out T : Any> {
-  /**
-   * This is not super well defined but works well enough for our simple use cases
-   */
+  @Deprecated(message = "Should not be used anymore", level = DeprecationLevel.ERROR)
```
martinbonnin, line 17: Maybe something more actionable, just in case?

```suggestion
  @Deprecated(message = "This was only used in internal API and shouldn't have been part of the public API. If you needed this, please open an issue.", level = DeprecationLevel.ERROR)
```

martinbonnin (approved):

libraries/apollo-tooling/build.gradle.kts
```diff
@@ -55,3 +55,10 @@ tasks.configureEach {
     enabled = false
   }
 }
+
+// Within the 'tests' project (a composite build), dependencies are automatically substituted to use the project's one.
+// But here we want to keep the published version of apollo-api when compiling the generated models.
+// So, disable the substitution rules (see https://docs.gradle.org/current/userguide/composite_builds.html#deactivate_included_build_substitutions).
+configurations.all {
+  resolutionStrategy.useGlobalDependencySubstitutionRules.set(false)
+}
```
martinbonnin, line 64: Leftover from a previous rebase?

BoD: woops good catch! Removed the faulty commit

PR #4817, opened 2023-03-30
Use Apollo Kotlin in more places

BoD:
- Use the public platform API to download and upload schemas
  - Uploading a schema can now go to a subgraph
- Still use the internal API for registering operations, but use Apollo Kotlin instead of "manual GraphQL"


libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaUploader.kt
```diff
@@ -11,45 +14,87 @@ object SchemaUploader {
       key: String,
       graph: String?,
       variant: String = "current",
+      subgraph: String? = null,
+      revision: String? = null,
       headers: Map<String, String> = emptyMap(),
```
martinbonnin, line 19: FWIW, this is used in Confetti [here](https://github.com/joreilly/Confetti/blob/main/backend/service-graphql/src/main/kotlin/dev/johnoreilly/confetti/backend/DefaultApplication.kt#L124). Will be interesting to see if still working.

martinbonnin (approved):

PR #4834, opened 2023-04-04
Update dependencies

martinbonnin:
Update most dependencies except `compile-testing` (because it breaks on Java annotations) and `uuid` (because it pulls Kotlin 1.8 stdlib)


BoD (approved):
🚀

Issue #4833, opened 2023-04-04
 3.8.0 introduced behavior change in `addHttpHeader`


ychescale9:
### Version

3.8.0

### Summary

https://github.com/apollographql/apollo-kotlin/pull/4754 seems to have introduced a behavior change:

https://github.com/apollographql/apollo-kotlin/blob/d8191287004b98c3468d7f0fdc82caac0fb94793/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt#L66-L68

as the headers are added to the private `additionalHttpHeaders` instead of `httpHeaders`, we can no longer inspect the current `httpHeaders` in an `ApolloCall` and make further changes to the headers based on the current headers:

```kt
// `this` is ApolloCall<T>
val header = httpHeaders?.find { it.name == headerName }
if (header == null) {
    addHttpHeader(headerName, headerValues)
} else {
    val newValue = header.value.plus(",$headerValues")
    httpHeaders = httpHeaders?.replace(headerName, newValue)
}
```

Using `httpHeaders(listOf(HttpHeader(HEADER_KEY, HEADER_VALUE)))` instead as the document suggested works, but the behavior change caught by our tests is a bit unexpected.

BoD:
Hi!

Thanks for reporting this! Indeed, this part of the code has changed in the context of [this issue](https://github.com/apollographql/apollo-kotlin/issues/4753).

Agreed that this is unexpected, and it would make sense for the additional headers to be part of what  `ApolloCall.httpHeaders` returns. I'll take a look at a fix!

BoD:
The fix was released as part of v3.8.1.

PR #4835, opened 2023-04-04
Add ApolloCall.replaceClientHttpHeaders

BoD:
Adds a flag `replaceClientHeaders`, if `true` `httpHeaders` replace the client’s, otherwise they’re added on top of them. 

The default is `false` because that’s the most useful use-case (challenge this assumption if disagreed!). This is a behavior change from 3.x (albeit since 3.8.0 the behavior change is lesser since `addHttpHeader` already adds instead of replacing).(Let’s note this change in the migration guide)

When merging this to `3.x`: 

- make the default `true`
- automatically set `replaceClientHeaders` to `false` when calling `addHttpHeader`

Related to #4833


libraries/apollo-runtime/api/apollo-runtime.api
```diff
@@ -24,6 +25,7 @@ public final class com/apollographql/apollo3/ApolloCall : com/apollographql/apol
 	public synthetic fun httpMethod (Lcom/apollographql/apollo3/api/http/HttpMethod;)Ljava/lang/Object;
 	public fun ignorePartialData (Ljava/lang/Boolean;)Lcom/apollographql/apollo3/ApolloCall;
 	public synthetic fun ignorePartialData (Ljava/lang/Boolean;)Ljava/lang/Object;
+	public final fun replaceClientHttpHeaders (Ljava/lang/Boolean;)Lcom/apollographql/apollo3/ApolloCall;
```
martinbonnin, line 28: 🚲 🏡 : `ignoreApolloClientHttpHeaders(true)` ?

```kotlin
apolloClient.query(myQuery).addHttpHeader("key", "value")
              .ignoreApolloClientHttpHeaders(true)                             
              .execute()
```

martinbonnin: Or just `ignoreClientHttpHeaders` if we aim for brevity

BoD: I'm 👍  with `ignoreApolloClientHttpHeaders`

BoD: Sorry I changed my mind 😅 I think `ignore` may be confusing, as it may be interpreted as the ApolloClient headers will be ignored even if this call doesn't have its own headers set, which is not true. WDYT?

martinbonnin: The same with `replaceClientHttpHeaders(true)`. If the call has no headers (headers = null), what should `replaceClientHttpHeaders(true)` do?

Feels like we should handle `call.headers = null` as `call.headers = emptyList()` and overwrite the client headers if that's the case?

martinbonnin:
Yea fair enough might work

libraries/apollo-runtime/api/apollo-runtime.api
BoD: > Feels like we should handle call.headers = null as call.headers = emptyList() and overwrite the client headers if that's the case?

Yes I think it's a good idea 👍. Removes an ambiguity.

I'm ok with `ignore` then (thought of another option: `override`)

martinbonnin: Not a fan of `override` because it can be understood as "I can override individual header names" like the `"key"` example below

martinbonnin (approved):

PR #4836, opened 2023-04-04
Error handling take #n

martinbonnin:
New take on error handling, supersedes #4800 and #4696

No more `ignorePartialData`, the reasonning is that checks should be done on `data != null`:

```kotlin
val response = apolloClient.query(query).execute()

// having to create an intermediate variable is unfortunate but needed to have smart cast
// See https://youtrack.jetbrains.com/issue/KT-8819/
val data = response.data
if (data != null) {
  // do something with data
} else {
  // do something with exception
}
```

If you prefer throwing, use `dataOrThrow()`:

```kotlin
// data is non-null
val data = apolloClient.query(query).execute().dataOrThrow()
```

Details

- Revert "hide setters, they should have been hidden from day1 (#4769)"
- Revert "Add ApolloClient.ignorePartialData (#4725)"
- make ApolloException sealed
- deprecate dataAssertNoError
- make sure `data == null` implies `exception ! = null`
- update apiDump


libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
```diff
@@ -64,7 +65,7 @@ internal class ApolloCacheInterceptor(
     if (request.doNotStore) {
       return
     }
-    if (response.exception != null) {
+    if (response.data == null) {
```
martinbonnin, line 68: That's the only behaviour change I needed to do

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/RegisterOperations.kt
```diff
@@ -269,7 +269,7 @@ object RegisterOperations {
       runBlocking { call.execute() }
     } catch (e: ApolloHttpException) {
       val body = e.body?.use { it.readUtf8() } ?: ""
-      throw ApolloException("Cannot push operations: (code: ${e.statusCode})\n$body", e)
+      throw Exception("Cannot push operations: (code: ${e.statusCode})\n$body", e)
```
martinbonnin, line 272: Changed this to not leak `apollo-api` to the caller.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -78,48 +78,107 @@ private constructor(
...
   @get:JvmName("dataAssertNoErrors")
+  @Deprecated(message = "Use dataOrThrow instead", replaceWith = ReplaceWith("dataOrThrow()"))
   val dataAssertNoErrors: D
     get() {
       return when {
         exception != null -> throw exception
-        hasErrors() -> throw ApolloException("The response has errors: $errors")
-        else -> data ?: throw ApolloException("The server did not return any data")
+        hasErrors() -> throw DefaultApolloException("The response has errors: $errors")
+        else -> data ?: throw DefaultApolloException("The server did not return any data")
```
BoD, line 114:
```suggestion
        else -> data
```
`hasErrors()` or `data == null` can't be true cause that would mean `exception != null`

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -157,39 +154,20 @@ private constructor(
             // canBeBatched(apolloRequest.canBeBatched)
             addHttpHeader(ExecutionOptions.CAN_BE_BATCHED, apolloRequest.canBeBatched.toString())
           }
-          if (apolloRequest.ignorePartialData != null) {
-            ignorePartialData(apolloRequest.ignorePartialData)
-          }
-          if (apolloRequest.useV3ExceptionHandling != null) {
-            useV3ExceptionHandling(apolloRequest.useV3ExceptionHandling)
-          }
```
BoD, line 165: This should be kept right?

BoD: (maybe a test is missing around that, I feel this should have broken something (my bad probably! I introduced useV3ExceptionHandling))

BoD (approved):
A few remarks, but I like this change! 👍

docs/source/essentials/errors.mdx
```diff
@@ -90,55 +98,68 @@ The server will send the following response:
...
+```graphql
+extend type Query @nonnull(fields: "person")
 ```
 
-Note that the safe call is still required on `film` because this field has a nullable type in the GraphQL schema. There are ways to override this in codegen. For more details, learn about [the `@nonnull` directive](../advanced/nonnull).
+Alternatively, you can check for errors in an interceptor and force data to null whenever a GraphQL error happens: 
+
 
-Another option is to call `ignorePartialData(true)`. This will force `ApolloResponse.data` to be null even if it wasn't in the response, whenever there are errors. This can also simplify error handling:
+If you don't want to handle partial responses, you can simplify your error handling logic. `ApolloResponse.dataAssertNoErrors` returns a non-nullable `data` and throws if there are any errors. This way, you can handle all errors in a single `catch {}` block:
```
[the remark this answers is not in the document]
martinbonnin: Yup 👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloRequest.kt
```diff
@@ -50,17 +48,14 @@ private constructor(
   ) : MutableExecutionOptions<Builder<D>> {
     private var requestUuid: Uuid = uuid4()
     override var executionContext: ExecutionContext = ExecutionContext.Empty
-      private set
```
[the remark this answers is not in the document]
martinbonnin: Right, I reverted a bit too much

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
martinbonnin: 👍 made it call into `dataOrThrow()`

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: Yikes, thanks for noticing

BoD (approved):

Issue #4759, opened 2023-03-16
Automatic "compat" -> "operationBased" migration


martinbonnin:
Add a way to transform compat codegen into operationBased codegen automatically
```[tasklist]
- [ ] Add ability to run this on specific modules
```

BoD:
Available in the current version

PR #4840, opened 2023-04-05
[IJ plugin] Add a helper to migrate from compat to responseBased codegen

BoD:
Adds a "refactoring processor":
- remove references to `.fragments` synthetic field
- rename `AsXyz` models to `OnXyz` and `asXyz` synthetic fields to `onXyz`
- remove `codegenModels.set("compat")` from gradle conf

Sorry this PR is big, it moves code around because we used to have only 1 processor, now there are 2 + an abstract one.

Also grouped them in an Apollo menu:

Screenshot 2023-04-05 at 15 47 31

Related to #4759


martinbonnin (approved):
LGTM!

PR #4841, opened 2023-04-05
Mention HTTP headers changes in migration doc

BoD:
Also update error handling section


docs/source/migration/4.0.mdx
```diff
@@ -40,9 +48,18 @@ try {
...
+    is ApolloGraphQLException -> {
+      // Handle GraphQL errors
+    }
+    else -> {
+      // Handle network error
+    }
+  }
+} else {
+  // No errors
+  val data = response.data
```
martinbonnin, line 61: What about doing the opposite and check `data` first so that it can be smart cast? :

```kotlin
val data = response.data
if (data ! = null) {
  // do something with data here
} else {
  // do something with exception here
}
```

BoD: 👍 flipped it

martinbonnin (approved):

PR #4849, opened 2023-04-06
Publishing: fix Gradle portal publishing and apollo-gradle-plugin-external

martinbonnin:
* Gradle portal publishing requires a websiteUrl
* `apollo-gradle-plugin-external` requires javadoc & source (which is not the case for `apollo-gradle-plugin` because the portal publishing plugin does it.


BoD (approved):
👏

PR #4855, opened 2023-04-10
Issue 4854: accept connectionPayload lambda instead of static auth

AlexHartford:
Fixes https://github.com/apollographql/apollo-kotlin/issues/4854

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/AppSyncWsProtocol.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/AppSyncWsProtocol.kt
@@ -22,140 +22,140 @@ class AppSyncWsProtocol(
     webSocketConnection: WebSocketConnection,
     listener: Listener,
 ) : WsProtocol(webSocketConnection, listener) {
+    
+  private var authorization: Map<String, Any?>? = null
 
-    private var authorization: Map<String, Any?>? = null
+  override suspend fun connectionInit() {
+    val message = mutableMapOf<String, Any?>(
+        "type" to "connection_init",
+    )
 
-    override suspend fun connectionInit() {
-        val message = mutableMapOf<String, Any?>(
-            "type" to "connection_init",
-        )
+    sendMessageMapText(message)
+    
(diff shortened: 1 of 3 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/AppSyncWsProtocol.kt
```diff
@@ -17,18 +17,22 @@ import kotlinx.coroutines.withTimeout
  * A [WsProtocol] for https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html
  */
 class AppSyncWsProtocol(
-    private val authorization: Map<String, Any?>,
     private val connectionAcknowledgeTimeoutMs: Long,
+    private val connectionPayload: suspend () -> Map<String, Any?>? = { null },
```
BoD, line 21: Looks like the compilation fails because the constructor is referenced by the Factory below

BoD:
Thanks a lot for the PR! Would it be possible to make this a new constructor, but also keep the old one with the static map, for backwards compatibility? 🙏

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/AppSyncWsProtocol.kt
AlexHartford: woops, my bad. I was cutting/pasting changes because my editor wanted to change all the code styling.

AlexHartford:
> Thanks a lot for the PR! Would it be possible to make this a new constructor, but also keep the old one with the static map, for backwards compatibility? 🙏

Done! Let me know if there's anything else I can do.

BoD:
Sorry, I should have said earlier, since `AppSyncWsProtocol` is public it also needs a secondary constructor for backwards compat. You can run `./gradlew apiDump` and check the resulting file in `apollo-runtime/api/apollo-runtime.api`, you should see only additions, no deletions or changes. (Please also commit that file, as we keep it for tracking).

Looks good otherwise!

BoD (approved):
Thanks a lot!

PR #4865, opened 2023-04-16
Make the README build badge report on latest main branch status

caarmen:
For the build badge in the README:
* Use an image which reflects the status of the ~CI~ push workflow only for the main branch
* Put a link that opens the workflows directly on the ~CI~ push workflow for the main branch

I think it makes more sense for the badge to only reflect the status of the `main` branch. It's "normal" for PR builds to fail (sometimes 😅 ), and I assume we don't want the main README to be red in that case.

Also: make the badge refer to the `push.yml` workflow. Looks like the `ci.yml` workflow ("CI") was moved to `push.yml` in ad095f4f54f8f552d36dec626f045fb69f56df15 (At least in that commit the `ci.yml` was deleted and `push.yml` was created.)

With these changes the badge is green now 💚 😅


martinbonnin (approved):
Thanks!

BoD (approved):
🙏

PR #4867, opened 2023-04-17
Attempt at enabling renovate

martinbonnin:


martinbonnin:
Merging without waiting for CI because merging is the only way to test

PR #4880, opened 2023-04-18
GEGP: add retries for tests

martinbonnin:
Add support for Gradle enterprise [test retries](https://docs.gradle.com/enterprise/flaky-test-detection/#gradle)


BoD (approved):
So cool! Should help with the flakiness

Issue #3815, opened 2022-01-25
API cleanup: use "scalar" instead of "custom scalar"


BoD:
It is now possible to specify a type and adapter for built-in scalars (#3748) so the terminology "custom scalar" could be simply "scalar". The class `CustomScalarAdapters` (and related functions `customScalarAdapters` `addCustomScalarAdapter` on `ApolloClient.Builder`) can be renamed.

Since this is a breaking change, keeping it for the next major release.

See also #3813 where we discuss replacing it by or embedding it in an `AdapterContext`.

There are also a few other symbols:
- com.apollographql.apollo3.api.CustomScalarType
- com.apollographql.apollo3.compiler.codegen.kotlin.file.CustomScalarBuilder
- com.apollographql.apollo3.compiler.ir.IrCustomScalar

PR #4882, opened 2023-04-18
Refact: custom scalar adapters renames (with typealias)

BoD:
Related to #3815 

- Rename `CustomScalarAdapters` → `ScalarAdapter`, `CustomScalarType` → `ScalarType`
- Split `Adapter` in 3:
    - `Adapter`: for scalars, implemented by users. 2 methods: `fromJson`/ `toJson` (no more `ScalarAdapters` parameter).
    - `DataAdapter`:  for generated data models, used by the codegen. 2 methods: `deserializeData(DeserializeDataContext)` / `serializeData(SerializeDataContext)`
    - `VariablesAdapter`: for variables, used by the codegen. Single method: `serializeVariables(SerializeVariablesContext)`
- An `AdapterToDataAdapter` is a bridge for the codegen

Breaking changes:

- Users’ `Adapter` implementations will break since the interface changed.
    - We could keep the old methods, marked as deprecated + default implementations of the new one, that calls the old one. But that’s not nice, as doing so would allow to implement `Adapter` and not override any methods and it would still compile, which is confusing.
    - Another option is to deprecate `Adapter` and introduce e.g. `ScalarAdapter`, and keep `addCustomScalarAdapter` on `ApolloClient.Builder` that converts the old `Adapter`  into the new `ScalarAdapter`.
    - Or, only rely on an IJ plugin migration helper.
- `CustomScalarAdapters` → `ScalarAdapter`
    - I think we can introduce a (deprecated) typealias?


BoD:
Separated the .expected to a specific commit to make it a tad easier to review.

Setting to draft while I'm looking at minimizing breaking changes.

BoD:
Closed in favor of #4895

Issue #4775, opened 2023-03-20
WebSockets: allow changing the serverUrl of WebSocketNetworkTransport


martinbonnin:
### Use case

Follow up from https://github.com/apollographql/apollo-kotlin/pull/4128#issuecomment-1476403441

martinbonnin:
The fix is available in 3.8.2

PR #4885, opened 2023-04-20
Allow changing the serverUrl of WebSocketNetworkTransport

Oleur:
Related to #4775.

Allow to dynamically set the websocket server URL in case we need to pass a new server URL with a new auth when the socket is reopen.

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
@@ -197,6 +197,7 @@ private constructor(
     private var httpExposeErrorBody: Boolean? = null
     private var webSocketEngine: WebSocketEngine? = null
     private var webSocketReopenWhen: (suspend (Throwable, attempt: Long) -> Boolean)? = null
+    private var webSocketReopenServerUrl: (suspend () -> String)? = null
 
     init {
       failOnNativeIfLegacyMemoryManager()
@@ -316,11 +317,26 @@ private constructor(
 
     /**
      * The url of the GraphQL server used for WebSockets
+     * Use this function or webSocketServerUrl((suspend () -> String)) but not both.
      *
      * See also [subscriptionNetworkTransport] for more customization
      */
     fun webSocketServerUrl(webSocketServerUrl: String) = apply {
       this.webSocketServerUrl = webSocketServerUrl
+    }
+
+    /**
+     * Configure dynamically the url of the GraphQL server used for WebSockets.
+     * Use this function or webSocketServerUrl(String) but not both.
+     *
+     * @param webSocketServerUrl a function returning the new server URL.
+     * This function will be called every time a WebSocket is opened. For example, you can use it to update your
+     * auth credentials in case of an unauthorized error.
+     *
+     * It is a suspending function, so it can be used to introduce delay before setting the new server URL.
+     */
+    fun webSocketServerUrl(webSocketServerUrl: (suspend () -> String)) = apply {
+      this.webSocketReopenServerUrl = webSocketServerUrl
     }
 
     /**
@@ -527,6 +543,9 @@ private constructor(
         check(webSocketReopenWhen == null) {
           "Apollo: 'webSocketReopenWhen' has no effect if 'subscriptionNetworkTransport' is set"
         }
+        check(webSocketReopenServerUrl == null) {
+          "Apollo: 'webSocketReopenServerUrl' has no effect if 'subscriptionNetworkTransport' is set"
+        }
         subscriptionNetworkTransport!!
       } else {
         val url = webSocketServerUrl ?: httpServerUrl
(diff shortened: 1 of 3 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -57,12 +57,13 @@ import okio.use
...
 class WebSocketNetworkTransport
 private constructor(
-    private val serverUrl: String,
+    private val serverUrl: String?,
     private val headers: List<HttpHeader>,
     private val webSocketEngine: WebSocketEngine = DefaultWebSocketEngine(),
     private val idleTimeoutMillis: Long = 60_000,
     private val protocolFactory: WsProtocol.Factory = SubscriptionWsProtocol.Factory(),
     private val reopenWhen: (suspend (Throwable, attempt: Long) -> Boolean)?,
+    private val reopenServerUrl: (suspend () -> String)?,
```
martinbonnin, line 66: Nit: no need to add a second parameter there, you can only keep the suspend one and set it from the builder:

```kotlin
    fun build(): WebSocketNetworkTransport {
      if (serverUrl != null && reopenServerUrl != null) error("You can't have a static and a dynamic serverUrl at the same time")
      if (serverUrl == null && reopenServerUrl == null) error("No serverUrl specified")
      return WebSocketNetworkTransport(
          { serverUrl ?: reopenServerUrl!!.invoke() },
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -353,11 +354,24 @@ private constructor(
...
 
     fun serverUrl(serverUrl: String) = apply {
       this.serverUrl = serverUrl
     }
 
+    /**
+     * Configure dynamically the server URL.
+     *
+     * @param serverUrl a function return the new server URL to use when reopening the [WebSocketNetworkTransport] containing
+     * the new auth credentials for example in case of an unauthorized error with the web socket.
```
martinbonnin, line 367:
```suggestion
     * @param serverUrl a function returning the new server URL.
     * This function will be called every time a WebSocket is opened. For example, you can use it to update your
     * auth credentials in case of an unauthorized error.
```

martinbonnin:
A few comments but LGTM overall. You'll also need to update the public API dump with `./gradlew apiDump` since this introduces new API

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
Oleur: Oh yes way better! I'll apply the changes :)

martinbonnin (approved):

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
```diff
@@ -347,14 +347,27 @@ private constructor(
...
     private var idleTimeoutMillis: Long? = null
     private var protocolFactory: WsProtocol.Factory? = null
     private var reopenWhen: (suspend (Throwable, attempt: Long) -> Boolean)? = null
 
     fun serverUrl(serverUrl: String) = apply {
+      this.serverUrl = { serverUrl }
+    }
+
+    /**
+     * Configure dynamically the server URL.
```
BoD, line 362: Nit
```suggestion
     * Configure the server URL dynamically.
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
```diff
@@ -550,6 +568,9 @@ private constructor(
                 if (webSocketReopenWhen != null) {
                   reopenWhen(webSocketReopenWhen)
                 }
+                if (webSocketReopenServerUrl != null) {
+                  serverUrl(webSocketReopenServerUrl)
```
BoD, line 572: Should we throw here if `webSocketServerUrl != null && webSocketReopenServerUrl != null`?

BoD (commented):
Thanks! A few minor comments.

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloClient.kt
martinbonnin: This is handled in `WebSocketNetworkTransport.Builder` but this reminds me we should add `webSocketReopenServerUrl` to the `copy()` function below

Oleur: Yup I was actually on it :)

martinbonnin (approved):

BoD (approved):
👍

PR #4896, opened 2023-04-23
[IJ plugin] Fix crash when some tooling models could not be fetched

BoD:


martinbonnin (approved):

martinbonnin:
Do we know the reason why some tooling models could not be fetched?

BoD:
> Do we know the reason why some tooling models could not be fetched?

When opening a project sometimes we receive a "Gradle has synced" event several times, which triggers the fetching of tooling models. If one is already ongoing, we cancel it first, which will result in an exception. But now that I'm writing this, I'm thinking we may be able to detect that case and not go further. I'll see if this can be improved a bit :) Thanks!

PR #4900, opened 2023-04-24
Share Gradle Tests Home with the main build

martinbonnin:
This PR tells Gradle instrumentation tests to reuse `~/.gradle` on both local and CI machines. In CI, the `gradle-build-action` uses init scripts to intercept build scans and display them nicely in the Github Actions UI. We don't want to do this for integration tests so this PR uses a system property to bypass them. It's not the cleanest way but seems to be working.

* [With shared home](https://ge.apollographql.com/s/ckcpgn7mukubk): 2min 6
* [Without shared home](https://ge.apollographql.com/s/vfmhjdjxrtzkc): 4min 12

### DIFF
--- a/libraries/apollo-gradle-plugin/build.gradle.kts
+++ b/libraries/apollo-gradle-plugin/build.gradle.kts
@@ -1,5 +1,3 @@
-import com.android.build.gradle.internal.tasks.factory.dependsOn
-
 plugins {
   id("org.jetbrains.kotlin.jvm")
   id("java-gradle-plugin")
--- a/libraries/apollo-gradle-plugin/build.gradle.kts
+++ b/libraries/apollo-gradle-plugin/build.gradle.kts
@@ -1,3 +1,5 @@
+import com.android.build.gradle.internal.tasks.factory.dependsOn
+
 plugins {
   id("org.jetbrains.kotlin.jvm")
   id("java-gradle-plugin")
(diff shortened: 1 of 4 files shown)


BoD (approved):
🚀

BoD (approved):

libraries/apollo-gradle-plugin/src/test/kotlin/util/TestUtils.kt
```diff
@@ -176,8 +176,15 @@ object TestUtils {
         .forwardStdOutput(output)
         .forwardStdError(error)
         .withProjectDir(projectDir)
+        // Runs in-process, saves a bit of time
         .withDebug(true)
-        .withArguments("--stacktrace", *args)
+        .withArguments(
+            "-g",
+            // Reuse the main build home directory
```
runningcode, line 183: I would recommend using a copy of this directory. I am afraid that sharing this directory with the test process could lead to a deadlock.

martinbonnin: The thing is local machines. `~/.gradle/caches/modules-2` is pretty large on local machines. On my machine, it's currently 15GB and I don't really want to copy that every time I run the Gradle integration tests. 
I thought several Gradle runs could share the same Gradle home?

libraries/apollo-gradle-plugin/build.gradle.kts
```diff
@@ -113,6 +113,32 @@ configure<PublishingExtension> {
   }
 }
 
+tasks.register("wrapInitScripts") {
```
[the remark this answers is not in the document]
martinbonnin: Yup, it's these ones: https://github.com/gradle/gradle-build-action/tree/main/src/resources/init-scripts

libraries/apollo-gradle-plugin/src/test/kotlin/util/TestUtils.kt
```diff
@@ -176,8 +176,15 @@ object TestUtils {
         .forwardStdOutput(output)
         .forwardStdError(error)
         .withProjectDir(projectDir)
+        // Runs in-process, saves a bit of time
```
runningcode, line 179: can you tell me more about this comment and what it does?

martinbonnin:
Merging this and we'll see how it goes 🤞

PR #4907, opened 2023-04-27
First pass at an option to generate response based code with jsExport

baconz:
This is meant as a first pass. Happy to go back and modify the approach if things don't look good.

This allows us to performantly fetch data on the JS side and then unsafeCast() to our response based generated models, which, in turn, allows us to use the same code to handle JS-based responses that we use to parse non-JS based responses, and obviates the need for apollo-jsei.

The big changes here are:

1. Put `JsExport` on all top level declarations and suppress `NON_EXPORTABLE_TYPE` because it is irrelevant. The important thing is the `JsExport` tells the compiler to generate the classes with unmangled property names. This allows us to cast directly from dynamic to the class.
2. Use `Array` as the list type because JS will parse lists as `Array`. Arrays are invariant, so we need to add `out` on arrays of interfaces.
3. Currently remove accessors when adding `JsExport` because `Companion` objects are illegal in exported classes (see TODO below).
4. Add a new `unsafeCastOrCast()` convenience method that allows the caller to access interfaces regardless of whether they're calling from JS or non-JS code.
5. Do not nest Fragment interfaces because nested interfaces are illegal on nested classes.
6. Use `String` instead lieu of enums for properties. enums are non-exportable, but we still generate the classes, so (similar to custom scalars) it is up to the caller to create the enum instances as needed.

Notes:

 * The `JsExport` only makes sense on response based codegen since it will be in a shape that matches the parsed JSON. I'm not sure of the best way to enforce this -- should we just throw an error if the user did not request `responseBased` codegen, but set `JsExport = true`?
 * Custom adapters do not work with `JsExport`. We should probably enforce this (similar to above).
 * Extension functions on generated code break on JS since we are casting a raw JS object and not actually instantiating a class. I'm not sure of a way to warn users about this beyond documenting it.
  * `generateAsInternal = true` does not work with `JsExport`, since the compiler ends up giving the `internal` modifier precedence, and thus mangling the variable names. Similar to above, we need a check that it is disabled.

TODO:

 * Currently not generating accessors in jsExports mode. We will need to have another pass that generates top-level extension functions which call through to unsafeCastOrCast().
 * Update tests
 * Update documentation


baconz:
@martinbonnin please hold off on reviewing  if you haven't started already. I'm still seeing some variable mangling on the interfaces and trying to track it down.

baconz:
Ok, I think this is ready for review. I tried to document all of the gotchas in the main description. I have it working on my fairly complex internal project, and it works for a few test projects as well.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapters.kt
```diff
@@ -42,6 +42,30 @@ class ListAdapter<T>(private val wrappedAdapter: Adapter<T>) : Adapter<List<@Jvm
   }
 }
 
+inline fun <reified T> arrayFromJson(wrappedAdapter: Adapter<T>, reader: JsonReader, customScalarAdapters: CustomScalarAdapters): Array<T> {
```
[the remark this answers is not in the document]
martinbonnin: Also interesting that inline functions are not considered by https://github.com/Kotlin/binary-compatibility-validator 👀 
I guess that makes sense because changes to inline function do not change the API but could still break source compatibility.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/JsExport.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.apollographql.apollo3.compiler.codegen.kotlin.KotlinContext
+import com.apollographql.apollo3.compiler.codegen.kotlin.KotlinSymbols
+import com.apollographql.apollo3.compiler.codegen.maybeFlatten
+import com.apollographql.apollo3.compiler.ir.IrModel
+import com.apollographql.apollo3.compiler.ir.IrModelGroup
+import com.squareup.kotlinpoet.AnnotationSpec
+import com.squareup.kotlinpoet.TypeSpec
+
+internal val suppressNonExportableType = AnnotationSpec.builder(KotlinSymbols.Suppress)
+  .addMember("%S", "NON_EXPORTABLE_TYPE")
```
martinbonnin, line 13: Can you add a comment there? My understanding is that we're making the top level query exportable but what we actually want is the nested models to be exportable. The query being exported is just a side effect of Kotlin not supporting the annotation on nested classes?

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateSourcesBase.kt
```diff
@@ -111,6 +112,10 @@ abstract class ApolloGenerateSourcesBase : DefaultTask() {
   @get:Optional
   abstract val addJvmOverloads: Property<Boolean>
 
+  @get:Internal
+  @get:Optional
+  abstract val jsExport: Property<Boolean>
```
martinbonnin, line 117: Can this be `@ApolloExperimental` ?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumResponseAdapterBuilder.kt
```diff
@@ -65,6 +65,7 @@ internal class EnumResponseAdapterBuilder(
 
     return TypeSpec
         .objectBuilder(layout.enumResponseAdapterName(name))
+        .maybeAddJsExport(context)
```
martinbonnin, line 68: Is this needed since enums can't be exported to JS? Or do I miss something there?

martinbonnin (commented):
Thanks for diving into this!

I left a few comments below as well as an integration test [here](https://github.com/baconz/apollo-kotlin/pull/1).

Re: custom adapters, I think they work with the same limitations as other `@JsExport` code, i.e. all properties mush be mappable to JS (see `Point` in the [integration test](https://github.com/baconz/apollo-kotlin/pull/1/files#diff-306343a91d5c4f5da9e23e79cebf09e3b3911dee0b7b916bfbe1327ee1d21613R9)). 
Also +1 to your comments about failing early on bad configurations (`internal`, `operationBased`, etc.. are not supported)

Overall, it's adding a bunch of complexity to the compiler. We can maybe rework a bit of the internals. Having to tunnel `jsExport` down to `resolveIrType()` feels a bit wrong but can be done in a subsequent PR. The biggest feedbacks at this point are:

1. keep the current behaviour of accessors in `non-jsExport` mode, they should return null if the cast fail.
2. investigate the accessors in the `jsExport` case. You have the use case so you'll know better about the ergonomics but it feels like we could have them behave exactly like the current accessors using `__typename` instead of the Kotlin type system.

baconz:
Thanks for all of the comments @martinbonnin! I will address these tonight (PDT). A couple of questions:

 * Should I add `jsExport` as a dimension to `CodegenTest` or is that overkill?
 * For the options validation (eg making sure not to specify `internal`) do you think that should live `KotlinCodeGen#write` or somewhere else? I didn't see any other validation type stuff in there yet.
  * Thoughts on the best place to document this? Should I add a section to the `responseBased` page?
  * Nice catch on `as` vs `as?` in `unsafeCastOrCast`. I can't see a good way to make the behavior match for JS vs non-JS without adding something like an `__expectedTypename` property to the generated code. If we did that we could write:
  
  ```kotlin
  actual inline fun <reified T> Any.unsafeCastOrCast(): T? {
    val obj = this.asDynamic()
    return if (obj["__typename"] == obj["__expectedTypename"]) {
      this.unsafeCast<T>()
    } else {
      null
    }
}
```

That feels like a pretty significant change to the API surface, though, and I'm not sure we'd want to do it here? If you think it looks ok, I can give it a shot.

I will say that for my use-case I'm fine if the behavior is different, so I'm inclined to just leave it with different behavior for now.

**Comment by martinbonnin** [comment]
> Should I add `jsExport` as a dimension to CodegenTest or is that overkill?

You can add a new folder and set the option just for this folder like for some other options (see [here](https://github.com/apollographql/apollo-kotlin/blob/21d45e88a606add6c939090241aa595e68ba72c4/libraries/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/CodegenTest.kt#L258)). 

I'm fine without `apollo-compiler` unit tests too if there are integration tests (in the `tests` composite build)

> For the options validation (eg making sure not to specify internal) do you think that should live KotlinCodeGen#write or somewhere else? 

You can do so in the Gradle Plugin ([here](https://github.com/apollographql/apollo-kotlin/blob/287e4fe916555a0def53d2dcd3e65a169015f112/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultService.kt#L261) is where `operationBased` is enforced for Java for an example).

> Thoughts on the best place to document this? Should I add a section to the responseBased page?

Would be cool to add a dedicated "Interoperability with JS" in the advanced section explaining the rationale behind the `jsExport` parameter, how to use it and the different limitations. You can link to `responseBased` from there (actually the `responseBased` page has changed in main so you will also save you some conflicts doing a dedicated page.

> I can't see a good way to make the behavior match for JS vs non-JS without adding something like an __expectedTypename property to the generated code.

It's even worse than that because a given model might match multiple `__typename`. For an example, `OtherAnimal`, can be `Turtle`, `Lion`, `Snake`, etc... so the codegen will need to have this list somewhere.

> I will say that for my use-case I'm fine if the behavior is different, so I'm inclined to just leave it with different behavior for now.

👍 works for me. So the contract of `unsafeCastOrCast` is something like so:

```kotlin
/**
 * Uses `unsafeCast` on JS or a regular `as` cast on other platforms
 */
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapters.kt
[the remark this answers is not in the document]
baconz: Ah, went to make this change and remembered why I didn't make them private to begin with:

> Public-API inline function cannot access non-public-API

If I make `arrayFromJson` or `arrayToJson` private, the inline extension function won't be able to access them.

I guess that sort of makes sense.

The solution I found is to make them private member functions of the anonymous object. That will hide them from the public API and satisfies the compiler.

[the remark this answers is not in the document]
martinbonnin: Nice trick 👍

baconz:
I addressed everything except the `CodegenTest`. I could not find a good way to make `kotlin-compile-testing` use the KMM plugin, and thus `assertCompiles` bailed with:

```
Declaration annotated with '@OptionalExpectation' can only be used in common module sources
```

when it tried to process the `JsExport` annotation.

I tried a bunch of things (including shifting paths around to include `commonMain`, but to no avail.

Another notable change is that I removed `unsafeCastOrCast` from the normal (non-jsExport) generated code. I can't see a reason to put it there. I still think it is nice to include the extension as a convenience, but can also just remove it completely.

docs/source/advanced/js-interop.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+If you would prefer to use `operationBased` models, and performance is not _as_ critical, you can
+use `DynamicJsJsonReader`. `DynamicJsJsonReader` works with a javascript object that is already parsed on the JS side.
+
+In JS reading response byte by byte from a byte array like okio is doing incurs a lot of overhead because Kotlin
+uses `Long` indices in its Arrays, and Longs do not have a JS implementation. 
+
+By bypassing this reading, `DynamicJsJsonReader` allows faster reading of responses while still keeping the full Kotlin type information. 
+
+In testing we've seen a ~2x performance boost on JS platforms using this parser vs ~10x with the `JsExport` approach.
```
martinbonnin, line 215: Would be cool to publish the benchmark for that if you have them around.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/unsafeCastOrCast.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Uses `unsafeCast` on JS or a regular `as?` cast on other platforms. Note that
+ * this means that JS platforms will crash at runtime if the cast fails, whereas
+ * other (statically) typed targets will return a null value.
+ *
+ * @return
+ */
+inline fun <reified T> Any.defaultUnsafeCastOrCast(): T? {
+  return this as? T
```
martinbonnin, line 13: I think it makes more sense to default to `as` here? Because:

1. Kotlin refers to `as` as "unsafe" cast in its [doc](https://kotlinlang.org/docs/typecasts.html#unsafe-cast-operator)
2. More generally, if there's a disconnect between `__typename` and the shape of the response, it's an error and it's best to catch it as soon as possible?

martinbonnin:
Many thanks for all the work on this! 

> I addressed everything except the CodegenTest. I could not find a good way to make kotlin-compile-testing use the KMM plugin, and thus assertCompiles bailed with [...]

Agreed. Let's leave `CodegenTest` out of it for now. Do you want to open an issue on `kotlin-compile-testing` for this ? As more and more projects target multiplatform, it will become more common..

> Another notable change is that I removed unsafeCastOrCast from the normal (non-jsExport) generated code. I can't see a reason to put it there.

Agreed 👍 

> I still think it is nice to include the extension as a convenience, but can also just remove it completely.

Works for me. My main and last question is [this one](https://github.com/apollographql/apollo-kotlin/pull/4907#discussion_r1195722244).

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/unsafeCastOrCast.kt
martinbonnin: All in all, I think I would go with something like so: 

```
@ApolloExperimental
inline fun <reified T> Any.apolloUnsafeCast(): T 
```

Or do you explicitely need the `as?` behaviour on non-JS?

docs/source/advanced/js-interop.mdx
```diff
@@ -0,0 +1,234 @@
+---
+title: JS Interoperability
+---
```
martinbonnin, line 3: I reshuffled this file a bit, I hope you don't mind.

docs/source/advanced/js-interop.mdx
baconz: I can run some official benchmark.js benchmarks of `DynamicJsJsonReader` vs `BufferedSourceJsonReader` this afternoon so we have some hard numbers.

Getting a benchmarks for the `jsexport` approach is a little harder because you need to account for the `.json()` call which happens asynchronously. Let me see what I can do about that.

docs/source/advanced/js-interop.mdx
baconz: It looks great!

baconz:
I opened a ticket in `kotlin-compile-testing` here: https://github.com/tschuchortdev/kotlin-compile-testing/issues/377

docs/source/advanced/js-interop.mdx
martinbonnin: Nice!
Kotlin/JS is direly lacking real life benchmarks and I feel this is as close as it gets to one. As Kotlin/JS gets more attention, this would be useful data points for future optimizations.

baconz:
Also meant to link this to https://github.com/apollographql/apollo-kotlin/issues/4728

martinbonnin (approved):

docs/source/advanced/js-interop.mdx
```diff
@@ -11,7 +11,7 @@ workloads (like those found in the Kotlin JSON parsing codepaths) can be slow.
 
 To work around this, Apollo Kotlin provides two alternative solutions to work with JS faster:
 
-* `jsExport` can be up to ~10x faster but does support the Kotlin type system nor `operationBased` codegen.
+* `jsExport` can be up to ~10x faster but does not support the Kotlin type system nor `operationBased` codegen.
```
martinbonnin, line 14: Woops, thanks!

docs/source/advanced/js-interop.mdx
baconz: I did some benchmarking using a big polymorphic query result from the Github API.

The idea was to compare:

 * `JSON.parse` + `unsafeCast` (`JsExporrt`)
 * `JSON.parse` + `DynamicJsJSONReader`
 * `BufferedSourceJsonReader`

I ran the tests on Kotlin 1.8.21 on Chrome 112 on my Macbook Pro M1.

The results are actually more dramatic than what we had in the doc:

```
parse with js export 40327.72686447989 ops/sec
parse with js export 68 runs
parse with dnymaic reader 9989.38589840788 ops/sec
parse with dnymaic reader 54 runs
parse with buffer reader 394.15146896515483 ops/sec
parse with buffer reader 63 runs
```

So dynamic is ~25x faster than buffered, and `JsExport` is ~4x faster than Dynamic.

* [The query](https://github.com/baconz/apollo-kotlin-js-performance/blob/benchmark/shared/src/commonMain/graphql/SearchQuery.graphql)
* [The setup](https://github.com/baconz/apollo-kotlin-js-performance/blob/benchmark/shared/src/jsMain/kotlin/com/baconz/apolloperformance/GithubSearch.kt)
 * [The benchmarking code](https://github.com/baconz/apollo-kotlin-js-performance/blob/benchmark/shared/src/commonMain/resources/index.html#L12)

baconz: I can clean up that repo and publish it as a benchmarking tool for this, or we could include a snapshot of those numbers. Thoughts?

docs/source/advanced/js-interop.mdx
```diff
@@ -0,0 +1,234 @@
+---
+title: JS Interoperability
```
BoD, line 2: I got to say, this doc is really great! Thanks a lot @baconz + @martinbonnin!

BoD (approved):

libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -139,8 +139,8 @@ public final class com/apollographql/apollo3/compiler/JavaNullable$Companion {
 }
 
 public final class com/apollographql/apollo3/compiler/KotlinCodegenOptions {
-	public fun <init> (Lcom/apollographql/apollo3/compiler/TargetLanguage;Ljava/util/List;ZZLcom/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks;ZLjava/lang/String;)V
-	public synthetic fun <init> (Lcom/apollographql/apollo3/compiler/TargetLanguage;Ljava/util/List;ZZLcom/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks;ZLjava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lcom/apollographql/apollo3/compiler/TargetLanguage;Ljava/util/List;ZZLcom/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks;ZLjava/lang/String;Z)V
+	public synthetic fun <init> (Lcom/apollographql/apollo3/compiler/TargetLanguage;Ljava/util/List;ZZLcom/apollographql/apollo3/compiler/hooks/ApolloCompilerKotlinHooks;ZLjava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
martinbonnin, line 143: Note to myself: this is a binary change. Since we're targeting 4.0 for this PR and the Gradle plugin is the main consumer of this, it should be fine. 
If someone is reading this in a few months, it means I was wrong and I'm sorry!

docs/source/advanced/js-interop.mdx
martinbonnin: For the record, I'm trying to get the deploy preview approved so we can see the doc in real conditions. I'm not too familiar with netlify but we'll get there!

baconz: I just pushed up the benchmark section! LMK what you think.

martinbonnin:
For the record, I was waiting for access to netlify to trigger the doc preview but this is taking longer than expected. Merging the PR and I'll do follow up PRs if docs need any more tweaks.

Thanks again @baconz for the awesome work and benchmarks 💙 !

PR #4916, opened 2023-05-03
Enable remote build cache

martinbonnin:


settings.gradle.kts
```diff
@@ -44,3 +44,15 @@ gradleEnterprise {
...
+
+
+buildCache {
+  local {
+    isEnabled = true
+    isPush = true
+  }
+  remote(gradleEnterprise.buildCache) {
+    isEnabled = true
+    isPush = true
```
runningcode, line 56: We recommend that you only push to the remote build cache only from CI builds.

PR #4924, opened 2023-05-05
Issue templates: add a freeform option

martinbonnin:
Give the option of freedom! I always end up editing the issue after the fact. If we end up getting too many low quality issues again, we can always put the bars again? 

Screenshot 2023-05-05 at 11 26 11

Screenshot 2023-05-05 at 11 26 15


BoD (approved):
Looks good! Yes we'll see if there's any impact on the quality of bug reports :)

PR #4926, opened 2023-05-05
Add a test for running with AGP 8.2

martinbonnin:
This PR is a bit larger than what I would have expected because it required splitting tests between Java 11 (older Gradle versions do not like Java 17) and Java 17 (new AGP versions do not like Java 11)

Most of the changes are renames though and the tldr is that we're compatible with AGP 8.2 out of the box 🎉 .

There is still one weird thing that IntelliJ doesn't recognized the generated sources folder as a source set in my tests :(. This has been a pretty fragile thing historically, might require some fine tuning.


BoD (approved):
Ultra nice!

PR #4929, opened 2023-05-05
Cloud Router test

martinbonnin:


tests/websockets/src/jvmTest/kotlin/CloudRouterTest.kt
```diff
@@ -0,0 +1,33 @@
...
+// Ignored because it depends on a remote server
+@Ignore
+class CloudRouterTest {
+  @Test
+  fun simple() = runTest {
+    val apolloClient = ApolloClient.Builder()
+        .serverUrl("")
+        .subscriptionNetworkTransport(
+            HttpNetworkTransport.Builder()
+                .serverUrl("https://main--subs-testing.staging.cloud.starstuff.dev/graphql")
```
calvincestari, line 20: I remember Hugh saying that this staging endpoint wouldn't stick around so if this is a CI job it'll start failing at some point.

martinbonnin: It's mostly as a reference for my future self. This test is ignored (see the `@Ignore` annotation on top of this class). I like keeping them around cause it's handy to get back to them, uncomment the `@Ignore` and try stuff. But they're not part of any CI test.

PR #4933, opened 2023-05-09
Run Gradle tests in CI

martinbonnin:
This is leftovers from https://github.com/apollographql/apollo-kotlin/pull/4926. Because the tasks are now `testJava11` and `testJava17`, the CI must run `allTests` and not `test`


BoD (approved):
👍

PR #4943, opened 2023-05-12
🐛 extending an enum value removes description

Jephuff:
The merging of the enum value extensions is using the extending enum value as a base and merging directives, but this means if the base had a description we lose that. I think switching this to using the existing value as the base which might have the description and copying it while merging in the extending directives gets the expected behaviour.

For example, a schema like this
```
enum Enum {
  "description"
  VALUE
}

extend enum Enum {
  VALUE @directive
}
```

The result is 
```
enum Enum {
   VALUE @directive
}
```

But I believe the result should be 
```
enum Enum {
  "description"
   VALUE @directive
}
```


BoD (approved):
👍

martinbonnin (approved):
Thanks for catching this 🙏

PR #4961, opened 2023-05-16
GE: remove Built-By attribute in generated jars

martinbonnin:
The "Built-By" jar attribute is different depending your machine and creates build cache misses


BoD (approved):
👍

PR #4965, opened 2023-05-17
[IJ plugin] Navigate to enum declaration

BoD:
https://github.com/apollographql/apollo-kotlin/assets/372852/5331af1c-9789-4771-86f9-12aa597e9331


martinbonnin (approved):
Very cool 🤩 

Small feedback, there's a lot of GraphQL gutter icon. I think I would keep it for operation myself. Then you can click, that takes you to the operation and you can run it from the IDE.

PR #4966, opened 2023-05-17
Doc: clarify that operationBased codegen is the recommendation

martinbonnin:


docs/source/advanced/response-based.mdx
```diff
@@ -1,18 +1,18 @@
...
 
-> 💡 For more in-depth information on each codegen, see [this design document](https://github.com/apollographql/apollo-android/blob/main/design-docs/Codegen.md).
+The Json response may have a different shape than your GraphQL operation though. This is the case when using merged fields or fragments. If you want to access your Kotlin properties as they are in the JSON response, Apollo Kotlin provides a `responseBased` codegen that matches the JSON response. GraphQL fragments are represented as Kotlin interfaces, so you can access their fields with Kotlin code like `(data.hero as Droid).primatryFunction`. Because it maps to the JSON response, the `responseBased` models have the property of allowing JSON streaming and/or mapping to dynamic JS objects. But because GraphQL is a very expressive language, [it's also easy to create a GraphQL query that generate a very large JSON response](https://github.com/apollographql/apollo-kotlin/issues/3144). **For this reason and other [limitations](#responsebased-codegen-limitations), we recommend using `operationBased` codegen by default**.
```
BoD, line 13:
```suggestion
The Json response may have a different shape than your GraphQL operation though. This is the case when using merged fields or fragments. If you want to access your Kotlin properties as they are in the JSON response, Apollo Kotlin provides a `responseBased` codegen that matches the JSON response. GraphQL fragments are represented as Kotlin interfaces, so you can access their fields with Kotlin code like `(data.hero as Droid).primaryFunction`. Because it maps to the JSON response, the `responseBased` models have the property of allowing JSON streaming and/or mapping to dynamic JS objects. But because GraphQL is a very expressive language, [it's also easy to create a GraphQL query that generate a very large JSON response](https://github.com/apollographql/apollo-kotlin/issues/3144). **For this reason and other [limitations](#responsebased-codegen-limitations), we recommend using `operationBased` codegen by default**.
```

docs/source/advanced/response-based.mdx
```diff
@@ -231,3 +231,11 @@ As a convenience, the `responseBased` codegen generates methods with the name pa
 val primaryFunction = hero1.droidFields().primaryFunction
 val height = hero2.humanFields().height
 ```
+
+## `responseBased` codegen limitations
+
+1. Because GraphQL is a very expressive language, [it's easy to create a GraphQL query that generate a very large JSON response](https://github.com/apollographql/apollo-kotlin/issues/3144). If you're using a lot of nested fragments, the generated code size can grow exponentially with the nesting level. We have seen relatively small GraphQL queries breaking the JVM limits like [maximum method size](https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-4.html#jvms-4.7.3).
+2. When using fragments, data classes must be generated for each operation where the fragments are used. To avoid name clashes, the models are nested and this comes with two side effects:
+   a. The resulting `.class` file can be very long, breaking the [256 default maximum file name on MacOS](https://apple.stackexchange.com/questions/86611/does-os-x-enforce-a-maximum-filename-length-or-character-restriction).
```
BoD, line 239:
```suggestion
   a. The resulting `.class` file name can be very long, breaking the [256 default maximum file name on MacOS](https://apple.stackexchange.com/questions/86611/does-os-x-enforce-a-maximum-filename-length-or-character-restriction).
```

BoD (approved):
👍

Note: I've recently renamed the file from `response-based-codegen.mdx` to `codegen-methods.mdx` and so I added this in the `_redirects` file:

```
/advanced/response-based-codegen/ /docs/kotlin/advanced/codegen-methods/
```

so I think you can just rename the file to `response-based-codegen.mdx` again and remove the redirect.

(Also reminder to merge this to the 3.x branch as well)

martinbonnin:
> so I think you can just rename the file to response-based-codegen.mdx again and remove the redirect.

Or make the redirect in the other direction?

BoD:
> Or make the redirect in the other direction?

Yes that works too

docs/source/_redirects
```diff
@@ -17,4 +17,4 @@
 /tutorial /docs/kotlin/tutorial/00-introduction
 /advanced/ui-tests/ /docs/kotlin/testing/ui-tests/
 /advanced/test-builders/ /docs/kotlin/testing/test-builders/
-/advanced/response-based-codegen/ /docs/kotlin/advanced/codegen-methods/
+/advanced/codegen-methods/ /docs/kotlin/advanced/response-based-codegen/
```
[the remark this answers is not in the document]
martinbonnin: 👍

PR #4972, opened 2023-05-22
Draft: Fix fan out in websocket network transport

baconz:
What about something like this as a fix for https://github.com/apollographql/apollo-kotlin/issues/4805 ?

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/WebSocketNetworkTransport.kt
@@ -17,6 +17,7 @@ import com.apollographql.apollo3.internal.isDeferred
 import com.apollographql.apollo3.internal.transformWhile
 import com.apollographql.apollo3.network.NetworkTransport
 import com.apollographql.apollo3.network.ws.internal.Command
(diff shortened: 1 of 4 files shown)


martinbonnin:
Thanks for looking into this. Would be nice to have some integration tests on this. I'll see what I can do.

martinbonnin:
I just added an integration test. It's all timing based so I hope it will not create too many flakes but on the other hand, since this is all implementation details, I don't really see another way to to it. Let me now what you think.

BoD (approved):
This looks good! Thanks!

martinbonnin:
Merging this cause I want to include it in an upcoming maintainance release. Let me know if you have any concerns @baconz

martinbonnin (approved):

baconz:
@martinbonnin sorry I was late to the party on this one.

Merging sounds good, and I think a bunch of folks here will be happy to get their hands on this!

I think the test could be improved by having `reopenWhen` emit to a flow that triggers the server to start again after at least 2 attempts. We could assert that the flow sees only one emission in the face of a server outage.

I tried to get that working, but TBH couldn't figure out how to get the server running on my machine.

martinbonnin:
> couldn't figure out how to get the server running on my machine

The server starts automatically from the integration test. If you open the `apollo-kotlin/tests` folder in IntelliJ, you should even be able to click the small green triangle to run it.

PR #4975, opened 2023-05-22
Fix Gradle annotation mismatch

BoD:
Related to [this failure](https://github.com/apollographql/apollo-kotlin/actions/runs/5045383276/jobs/9049587948#step:5:2630)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -63,6 +63,7 @@ tasks.register("ciPublishRelease") {
 tasks.register("ciTestsGradle") {
   description = "Execute the Gradle tests (slow)"
   dependsOn(":apollo-gradle-plugin:allTests")
+  dependsOn(":apollo-gradle-plugin-external:validatePlugins")
 }
 
 tasks.register("ciTestsNoGradle") {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -63,6 +63,7 @@ tasks.register("ciPublishRelease") {
 tasks.register("ciTestsGradle") {
   description = "Execute the Gradle tests (slow)"
   dependsOn(":apollo-gradle-plugin:allTests")
+  dependsOn(":apollo-gradle-plugin-external:validatePlugins")
```
martinbonnin, line 66: Might be worth doing this instead? 

```
dependsOn(":apollo-gradle-plugin:build")
```

I would expect `validatePlugins` to be a dependency of `build` so adding the dependency on `build` is a shortcut to "build and test everything" ? 

IIRC, the only reason this wasn't made for `ciTestsNoGradle` is to **not** execute the iOS sim tests

martinbonnin (approved):

build.gradle.kts
martinbonnin: We might want to partition the tests. Something like:

```
tasks.register("ciTestsGradle") {
  description = "Execute the Gradle tests (slow)"
  dependsOn(":apollo-gradle-plugin:build")
  dependsOn(":apollo-gradle-plugin-external:build")
}

tasks.register("ciTestsNoGradle") {
  subprojects {
    if (name != "apollo-gradle-plugin" && name != "apollo-gradle-plugin-external") {
      dependsOn(tasks.matching { it.name == "test" })
    }
    dependsOn(tasks.matching { it.name == "jvmTest" })
    dependsOn(tasks.matching { it.name == "jsIrTest" })
    dependsOn(tasks.matching { it.name == "macosX64Test" })
    dependsOn(tasks.matching { it.name == "apiCheck" })
  }
}
```

But this is more involved so 👍 to merge as is.

PR #4984, opened 2023-05-24
Dokka: configure theming in the root project too

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/pull/4298#issuecomment-1560710333


BoD (approved):
👍

PR #4986, opened 2023-05-25
Release 3.8.2

martinbonnin:
Cherry-picks from `main` up to `d11b6b1273bb64f03d49c2c766516f03c01e8af1`

Few comments:

* #4823 is not pulled in because `SchemaUploader` is using a different API on 3.x
* #4839 because that's a behaviour change (request headers replace client headers on 3.x, they will be additive in 4.x)
* 3.x is still using Kotlin 1.8.10, we might want to bump that at some point but I didn't touch any of that


BoD (approved):
🚀

PR #4997, opened 2023-05-30
Use Kotlin 1.9.0

BoD:


build-logic/src/main/kotlin/Node.kt
```diff
@@ -17,10 +17,8 @@ fun Project.configureNode() {
     )
   }
 
-  // See https://youtrack.jetbrains.com/issue/KT-47215
```
BoD, line 20: `disableGranularWorkspaces()` no longer exists, but it looks like the original issue KT-47215 is also no longer relevant

tests/defer/src/jsTest/kotlin/util/HttpUtil.kt
```diff
@@ -1,6 +0,0 @@
-package util
-
-import http.OutgoingHttpHeaders
```
BoD, line 3: This import was invalid it seems, but this whole class is unused anyway

tests/defer/src/jsTest/kotlin/util/JsUtil.kt
```diff
@@ -1,8 +1,11 @@
 package util
 
-import tsstdlib.Symbol
```
BoD, line 3: This import was invalid, not sure why it didn't cause a problem before

libraries/apollo-gradle-plugin/testProjects/multiplatform/build.gradle.kts
```diff
@@ -26,18 +26,3 @@ apollo {
         packageNamesFromFilePaths()
     }
 }
-
-// See https://youtrack.jetbrains.com/issue/KT-56019
-val myAttribute = Attribute.of("com.apollographql.test", String::class.java)
```
martinbonnin, line 31: I think we have more instances of this in `tests`

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkConditionalFragment.kt
```diff
@@ -21,6 +21,7 @@ internal fun checkConditionalFragments(definitions: List<GQLDefinition>): List<I
     when (it) {
       is GQLOperationDefinition -> scope.checkConditionalFragments(it.selectionSet.selections)
       is GQLFragmentDefinition -> scope.checkConditionalFragments(it.selectionSet.selections)
+      else -> {}
```
martinbonnin, line 24: Interesting. Is that triggering an error now or is it a warning we've had forever?

tests/defer/src/jsTest/kotlin/test/DeferWithHelixTest.kt
```diff
@@ -1,285 +0,0 @@
...
-import kotlin.random.Random
-import kotlin.test.Ignore
-import kotlin.test.Test
-import kotlin.test.assertEquals
-
-@Ignore
-// TODO These tests are temporarily ignored as the latest version of graphql-js
-// uses a package format which is incompatible with Kotlin/JS.
-// See https://youtrack.jetbrains.com/issue/KT-12784
-class DeferWithHelixTest {
```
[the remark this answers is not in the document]
BoD: There were errors related to JS dependencies in this area and it took me a while to figure out it was related to `disableGranularWorkspaces` (see above). My first attemt was to get rid of the issue altogether (as these tests have been ignored for a while anyway). Now looking at KT-12784 it is possible that we could re-enable them actually 🤔 I'll undelete them and have a look at if we can unignore them!

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/checkConditionalFragment.kt
BoD: Not sure why but now if I remove this now it compiles fine - I'm pretty sure it didn't while I was working on this PR 🤨 There is a warning in the IDE only though.

martinbonnin: The plot thickens... 🕵️ Not a big deal, was just curious. Keeping the `else` branch is better 👍

libraries/apollo-gradle-plugin/testProjects/multiplatform/build.gradle.kts
BoD: Actually removing the ones in tests still fails. But this one produces `Configuration with name 'releaseFrameworkIosFat' not found.` and removing this resolves that.

martinbonnin: Interesting, looks like [it's not yet fixed with 1.9](https://youtrack.jetbrains.com/issue/KT-55751/MPP-Gradle-Consumable-configurations-must-have-unique-attributes#focus=Comments-27-7362740.0-0). Something else to try in the -RC

martinbonnin: Added a comment there: https://youtrack.jetbrains.com/issue/KT-55751/MPP-Gradle-Consumable-configurations-must-have-unique-attributes#focus=Comments-27-7417705.0-0

tests/defer/src/jsTest/kotlin/test/DeferWithHelixTest.kt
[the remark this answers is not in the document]
BoD: I undeleted them, but AFAICT using the latest version of GraphqL-JS (ESM) still doesn't work - added a comment on KT-12784.

build-logic/src/main/kotlin/Node.kt
BoD: Using a dev version of the RC fixes it.

martinbonnin (commented):
I was going to comment that we should merge this but I'm having second thoughts on the K/N binary format. I wouldn't want to force all K/N users on 1.9-RC so my vote goes to waiting for stable before merging.

BoD:
>I was going to comment that we should merge this but I'm having second thoughts on the K/N binary format. I wouldn't want to force all K/N users on 1.9-RC so my vote goes to waiting for stable before merging.

Ah yeah it's true about the native binaries.

Yes, makes sense to wait, and it shouldn't be too long before the stable anyway.

martinbonnin (approved):

PR #4999, opened 2023-05-31
[IJ plugin] Add GQL -> generated Kotlin navigation

BoD:
Cmd+b on an Operation, a Fragment, a field, an enum, or input type allows to go to the related generated code.

https://github.com/apollographql/apollo-kotlin/assets/372852/a3d5f800-e388-40d7-8187-80cd41bcdd23


martinbonnin (approved):
Nice! One nitpick: it could be argued that clicking a fragment spread could expose both options like fields:

* Go to GraphQL fragment definition (in the .graphql file)
* Go to Kotlin definition (in the .kt file)

Also would be nice to have a visual clue that one of them leads to a .graphql file and the other to a .kt file.

BoD:
> clicking a fragment spread could expose both options
Good point, I'll have a look!

> Also would be nice to have a visual clue that one of them leads to a .graphql file and the other to a .kt file (or is it part of https://github.com/apollographql/apollo-kotlin/pull/4980#event-9391155516?).

💯 the video doesn't show it because I hadn't rebased yet, but the GraphQL destinations have a GraphQL icon now.

PR #5029, opened 2023-06-19
Use a relative path for the mockJDK

martinbonnin:
See also https://github.com/JetBrains/gradle-intellij-plugin/issues/1376#issuecomment-1594323024


BoD (approved):
👍

Issue #5035, opened 2023-06-21
Android Studio Plugin - Navigate to Query Gutter Icon Missing


agrosner:
### Version

3.8.1 (apollo), 4.0.0-alpha.2-SNAPSHOT.20 (plugin)

### Summary

In the example provided in [the announcement](https://www.apollographql.com/blog/mobile/kotlin/announcing-the-apollo-kotlin-plugin-for-android-studio-and-intellij-idea/), you can navigate from a query to its corresponding GraphQL file operation via gutter icons that link you there. 

Android Studio Hedgehog | 2023.1.1 Canary 7
Build #AI-231.9011.34.2311.10238683, built on June 1, 2023
Runtime version: 17.0.6+0-17.0.6b829.9-10027231 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.4
GC: G1 Young Generation, G1 Old Generation
Memory: 16384M
Cores: 10
Metal Rendering is ON
Registry:
    external.system.auto.import.disabled=true
    debugger.watches.in.variables=false
    debugger.new.tool.window.layout=true
    ide.text.editor.with.preview.show.floating.toolbar=false
    ide.instant.shutdown=false
    ide.experimental.ui=true

Non-Bundled Plugins:
    detekt (2.0.0)
    com.intellij.lang.jsgraphql (4.0.1)
    com.apollographql.ijplugin (4.0.0-alpha.2-SNAPSHOT.20)
    com.nbadal.ktlint (0.12.0)
    com.jetbrains.kmm (0.6.0(231)-5)
    com.developerphil.adbidea (1.6.10)
    org.sonarlint.idea (8.3.0.71062)

In our project we return a query from an interface implementation like:
```kotlin
class SomeQueryMapper : QueryMapper<SomeQueryExperience, SomeQuery.Data, SomeQuery> {

    override fun map(exp: SomeQueryExperience): SomeQuery =
        SomeQuery()
}
```

The gutter icons do not show up for this instance. We will send this query off to a client outside of the file / not at the source call here. My guess is the plugin doesn't consider this case? In this situation we just go to the generated code.

### Steps to reproduce the behavior

1. Create a class or function that just returns a new query
2. Observe no gutter icon

BoD:
Thanks for reporting!

[...]

PR #5037, opened 2023-06-23
[IJ plugin] Make navigation more robust when gql elements are lowercase

BoD:
Fix for #5035


martinbonnin (approved):
That's technically not 100% correct because 2 operations can have different cases `GetFoo` and `GetFOO` are 2 different operations for an example. It just happens that on MacOS the filesystem is case insensitive and therefore operation files will clobber each other but it's an issue on linux for an example. 

I'll defer to you as whether this is a real issue or not

BoD:
Totally right! I made the comparison a bit less lenient :)

martinbonnin (approved):

PR #5045, opened 2023-06-28
[IJ plugin] Add API keys section in settings

BoD:
Add a GraphOS / API Keys section to the settings.

https://github.com/apollographql/apollo-kotlin/assets/372852/6b044c61-33a6-48ae-93a6-4e3bade69b65


martinbonnin:
UX feedback: I'd separate `project` into `project` & `service` and rename `Service` name to `Apollo Graph`

* **Gradle project**: dropdown, defaults to the first project that applies the apollo plugin sorted lexicographically
* **Gradle service**: dropdown, defaults to the first service in the selected project, sorted lexicographically
* **Apollo Key**: text edit, hint to `service:$Graph:$key`
* **Apollo Graph**: remove? And only support Graph keys? TBC that we can do that.

BoD:
Updated, now there are 4 fields.

I kept user keys and graph name field, because I have my own example where I don't have enough rights to access API keys for a certain graph, so I must use my user API key.

Also used "GraphOS" instead of "Apollo" (but we'll probably get some product feedback later about the best wording to use).

martinbonnin:
Makes a ton of sense 👍 I liked Apollo Key because the setup in a lot of places is: export an environment variable named `APOLLO_KEY` but GraphOS API Key works too 👍 Something to bikeshed with the GraphOS team.

martinbonnin (approved):

PR #5047, opened 2023-06-29
Add multiplatform apollo-ast

martinbonnin:
This PR adds an alternative  hand written recursive descent parser side by side with [Antlr](https://www.antlr.org/). The lexer and parser are based on @pcarrier [graphed](https://github.com/pcarrier/graphed) project with some tweaks to make it map the `GQLFoo` class hierarchy and inspiration from the [graphql-js implementation](https://github.com/mjmahone/graphql-js/blob/8724e22f9e89224dd2669d6fc0c4b7fe34c2c8dd/src/language/parser.ts).

This has several advantages:

1. It's Kotlin multiplatform. 
2. It allows more fine grained control over the custom grammar logic (things like `const`, `list`, lookahead, etc...)
3. It is smaller and faster (See benchmarks below)

And a few drawbacks:

1. ~no streaming support, the whole document is loaded as a `String` right now~ Nevermind, [antlr buffers everything internally as well.](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/Java/src/org/antlr/v4/runtime/CodePointBuffer.java#L218)
2. no error recovery. First parsing error is terminal

I'm hoping the advantages outweigh the drawbacks. Also the drawbacks can be fixed if they prove too problematic.

## Benchmarks

You can run benchmarks with `./gradlew :apollo-ast:benchmark`. On my M1 macbook pro it's twice faster. Running `parseAsGQLDocument` on all files from `apollo-compiler/src/test/graphqls`:

```
jmh summary:
Benchmark              Mode  Cnt    Score    Error  Units
Benchmark.antlrTest   thrpt   10  328.142 ±  8.754  ops/s
Benchmark.parserTest  thrpt   10  771.173 ± 18.674  ops/s
```

In terms of size, running `./gradlew :apollo-ast:jar`:

* At `3441362ba45072821cb82fc13895608fb7cbcac5` (with Antlr): 650 kB (not counting `antlr-runtime`)
* At `7d838f5b7badf7a23012fc6901ca9391e27962ed`(with manual parser): 593 kB

This still feels way too large but at least it's better.


docs/source/migration/4.0.mdx
```diff
@@ -199,6 +199,10 @@ The normalized cache must be configured before the auto persisted queries, confi
 
 The AST classes (`GQLNode` and subclasses) as well as `Introspection` classes are not data classes anymore (see https://github.com/apollographql/apollo-kotlin/pull/4704/).
 
+`GQLSelectionSet` and `GQLArguments` are deprecated and removed from `GQLField`, `GQLInlineFragment` and other constructors. Use `.selections` directly
+
+`GQLInlineFragment.typeCondition` is now nullable to account for inline fragments who inherit their type condition
```
martinbonnin, line 204: These are placeholders as we're still alpha and things are changing fast. I'll revisite the migration guide globally when we're a bit more stable.

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
```diff
@@ -17,15 +17,15 @@ private class CheckKeyFieldsScope(
 
 fun checkKeyFields(operation: GQLOperationDefinition, schema: Schema, allFragmentDefinitions: Map<String, GQLFragmentDefinition>) {
   val parentType = operation.rootTypeDefinition(schema)!!.name
-  CheckKeyFieldsScope(schema, allFragmentDefinitions).checkField("Operation(${operation.name})", operation.selectionSet.selections, parentType)
+  CheckKeyFieldsScope(schema, allFragmentDefinitions).checkField("Operation(${operation.name})", operation.selections, parentType)
```
martinbonnin, line 20: This is a breaking change. `GQLSelectionSet` and `GQLArguments` are removed. There are a lot of places in the spec where there are list of symbols surrounded by `{ }` or `()`. Sometimes we had a node for them (like `GQLSelectionSet` in `GQLField.selectionSet`) but sometimes not (like `GQLObjectTypeDefinition.fields`).

Removing the intermediate node makes the callsite cleaner (like here) but looses the `sourceLocation` for the opening brace/parenthesis. We could also have gone the other direction and add extra intermediate nodes everywhere so it's all a tradeoff.

I went with "smaller/cleaner callsites" because this is what makes sense here I think as apollo-compiler doesn't really care about the position of the braces/parenthesis. Also we're already stripping whitespace and comments so it felt somewhat consistent.

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
```diff
@@ -92,23 +92,23 @@ private fun CheckKeyFieldsScope.collectFields(
           return@flatMap emptyList()
         }
 
-        collectFields(it.selectionSet.selections, it.typeCondition.name, implementedTypes)
+        collectFields(it.selections, it.typeCondition?.name ?: parentType, implementedTypes)
```
martinbonnin, line 95: This is the other breaking change. Our antlr parser didn't support inline fragments without a type condition. See https://spec.graphql.org/draft/#sel-GAFbfLABAB_F3kG:

```graphql
query inlineFragmentNoType($expandedInfo: Boolean) {
  user(handle: "zuck") {
    id
    name
    ... @include(if: $expandedInfo) {
      firstName
      lastName
      birthday
    }
  }
}
```

This is now supported by the parser but I'm considering throwing in the compiler because it makes the naming of fragment classes a bit awkward. Any thoughts here?

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/gql.kt
```diff
@@ -726,22 +744,24 @@ class GQLSchemaExtension(
...
-      operationTypesDefinition.join(writer, separator = "")
-      unindent()
-      write("}\n")
+      if (operationTypeDefinitions.isNotEmpty()) {
+        write("{\n")
+        indent()
+        operationTypeDefinitions.join(writer, separator = "")
+        unindent()
+        write("}\n")
+      }
```
martinbonnin, line 753: It is valid for a `GQLSchemaExtension` to not have any `operationTypeDefinitions`, for an example to add a directive to the schema:

```graphql
extend schema @foo
```

But it is not valid to have empty braces (`{}` ) in the grammar. Same with other type extensions below

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/check_key_fields.kt
BoD: Unless I missed it, I think we never got any ask or complaint from users wanting to do this? (Although that doesn't necessary mean nobody encountered it, but maybe the workaround is easy enough that nobody reached out).

In addition to `@include`, one place where this can be nice is to use `@defer`.

But for the name... `fragment1`, `fragment2`, etc.?

In any case I find it reasonable to keep it unsupported and at least it's cool that with this new parser we have the ability to support it, now or later.

martinbonnin: > we never got any ask or complaint from users wanting to do this

Exactly. "Don't fix it if it ain't broke"

> one place where this can be nice is to use `@defer`

Indeed. Well, it's always easier to add than remove so I'll throw for now and if we get requests for this we can always add it.

martinbonnin: check added in https://github.com/apollographql/apollo-kotlin/pull/5047/commits/84e30bad7b648f1648bb8f4e664afcfd347f73ea

martinbonnin:
This should be ready for review. I've updated the initial description with benchmarks. It's looking good :)

libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/AntlrParseTest.kt
```diff
@@ -18,11 +18,11 @@ class AntlrParseTest {
       ab bc
       """.trimIndent()
           .buffer()
-          .parseAsGQLDocument()
+          .parseAsGQLDocument(useAntlr = false)
```
BoD, line 21: Name of the class should be changed I guess

BoD (approved):
Looks good 👍

And the benchmark results are 🚀

libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/AntlrParseTest.kt
martinbonnin: Yep, I have a follow up PR with more testing, trying to catch some lexer errors. I'll rename in that PR.

PR #5048, opened 2023-06-29
[IJ plugin] Fetch field latencies

BoD:
Latencies are fetched:
- when the Gradle tooling models change
- when the API key settings change
- every 12 hours

https://github.com/apollographql/apollo-kotlin/assets/372852/cd65a368-5c07-453c-ae9c-fd2a41508304


martinbonnin (approved):
LGTM. 2 comments:

* Might be worth adding a "manual sync" button? At least for troubleshooting.
* Might be worth double checking we're not going to DDOS the GraphOS API. I think it's fine but I guess this should be put on someone's radar?

PR #5063, opened 2023-07-05
[apollo-ast] Add more tests and rewrite the lexer to use bytes instead of strings

martinbonnin:
Fixes a bunch of edge-cases in the parsing of unicode/numbers. An unfortunate side effect is that reading bytes is actually slower than reading the full string

with this PR:

```
Benchmark              Mode  Cnt    Score   Error  Units
Benchmark.antlrTest   thrpt   10  336.409 ± 4.076  ops/s
Benchmark.parserTest  thrpt   10  600.569 ± 4.646  ops/s
```

vs previously:

```
Benchmark              Mode  Cnt    Score    Error  Units
Benchmark.antlrTest   thrpt   10  328.142 ±  8.754  ops/s
Benchmark.parserTest  thrpt   10  771.173 ± 18.674  ops/s
```

Not 100% sure why. My guess is that we still have to decode utf8 char by char in the lexer while the JVM can certainly do it faster in native land or so. Or maybe it's all this `Byte` arithmetic that is super slow, not sure. 

More tests are coming but I need to shuffle the source sets to introduce JVM-only source sets (for `TestParameterInjector`) and since this PR is quite big already, I'll make a separate one.


libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/LexerTest.kt
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    scanFirst(",,,foo,,,").apply {
+      assertName("foo")
+    }
+  }
+
+  @Test
+  fun lexesBlockString() {
+    builtinDefinitions()
```
BoD, line 118: I'm confused by this one

BoD (approved):
Impressive! And very nice tests 👍

libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/LexerTest.kt
martinbonnin: It's something I wrote during dev to try out something but it's not relevant indeed. Removed, thanks!

PR #5064, opened 2023-07-05
[apollo-ast] add endLine/endColumn

martinbonnin:
This PR adds `GQLNode.endLine` and `GQLNode.endColumn`. Also because `SourceLocation` adds some overhead, there is now an option to skip it if it's not going to be used. This means `sourceLocation` became nullable in the whole codebase but this was actually more or less the case because we had `SourceLocation.UNKNOWN` in a bunch of places already

The `endLine`/`endColumn` is inclusive. For this GraphQL document:

```graphql
{
  countries(first: 100) {
    code
  }
}
```

The 'countries" `GQLField` will have (including the selections):

* `line = 2`
* `column = 3`
* `endLine = 4`
* `endColumn = 3`

@mayakoneval would that fit your use case?

### DIFF
--- a/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
+++ b/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
@@ -38,3 +38,5 @@ class SourceLocation(
     val UNKNOWN = SourceLocation(-1, -1, -1, -1, null)
   }
 }
+
+fun SourceLocation?.pretty() = this?.pretty() ?: "(unknown location)"
--- a/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
+++ b/libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
@@ -3,13 +3,13 @@ package com.apollographql.apollo3.ast
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 
 /**
- * @param line the line number where the symbol starts, starting at 1
+ * @param line the line where the symbol starts, starting at 1
  *
- * @param column the column in the current line where the symbol starts, starting at 1
+ * @param column the column where the symbol starts, starting at 1
  *
- * @param endLine the line number where the symbol end, inclusive, starting at 1
+ * @param endLine the line where the symbol ends, inclusive, starting at 1
  *
- * @param endColumn the column in the current line where the symbol end, inclusive, starting at 1
+ * @param endColumn the column where the symbol ends, inclusive, starting at 1
  * *
  * @param filePath The path to the document containing the node
  * Might be null if the document origin is not known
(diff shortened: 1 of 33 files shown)


mayakoneval:
This is perfect for my use case. As long as its consistent I am happy. 

For the definitions: 

`type Something { 
   fieldA: String
 }`
 
 if you wanted the end for `fieldA` I assume it would be line 2, column 2 -> line 2, column 16
 
 yes?

martinbonnin:
> if you wanted the end for fieldA I assume it would be line 2, column 2 -> line 2, column 16

Almost! Because column starts at 1, if you have 2 spaces indentation, the second line will start at char #3

All it all, it's `2:3` -> `2:16`

Added a test [there](https://github.com/apollographql/apollo-kotlin/blob/849be30a2b43b51fcfe305d71f5f738a883e09ed/libraries/apollo-ast/src/test/kotlin/com/apollographql/apollo3/graphql/ast/test/ParserTest.kt#L100)

mayakoneval:
perfect! Thanks so much

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
```diff
@@ -3,16 +3,22 @@ package com.apollographql.apollo3.ast
...
 /**
- * @param line the line number of the source location, starting at 1
+ * @param line the line number where the symbol starts, starting at 1
  *
- * @param column the position in the current line, starting at 1
+ * @param column the column in the current line where the symbol starts, starting at 1
  *
+ * @param endLine the line number where the symbol end, inclusive, starting at 1
+ *
+ * @param endColumn the column in the current line where the symbol end, inclusive, starting at 1
```
BoD, line 12: Nit:

```suggestion
 * @param column the column in `line` where the symbol starts, starting at 1
 *
 * @param endLine the line number where the symbol ends, inclusive, starting at 1
 *
 * @param endColumn the column in `endLine` where the symbol ends, inclusive, starting at 1
```

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/exception.kt
```diff
@@ -5,23 +5,23 @@ import java.io.IOException
...
 )) {
 
   companion object {
-    private fun formatForIdea(sourceLocation: SourceLocation, description: String): String {
+    private fun formatForIdea(sourceLocation: SourceLocation?, description: String): String {
       // Idea understands a certain format and makes logs clickable
       // It's not 100% clear where this is specified but it at least works with
       // 2020.3
-      return "e: ${sourceLocation.pretty()}: $description"
+      return "e: ${sourceLocation?.pretty()}: $description"
```
BoD, line 19: We could have a `fun SourceLocation?.pretty()` that returns `"(unknown location)"`?

BoD (approved):
🚀

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/exception.kt
martinbonnin: Good idea, added [here](https://github.com/apollographql/apollo-kotlin/pull/5064/commits/d041d48f43266cab7270686142298745e287b8a7)

libraries/apollo-ast/src/main/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
BoD: Even better 👍

PR #5068, opened 2023-07-06
[Runtime] throw ApolloNetworkException instead of ApolloParseException on IOException

martinbonnin:
Low hanging fruit, closes https://github.com/apollographql/apollo-kotlin/issues/2783.

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
@@ -33,6 +33,7 @@ import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
+import okio.EOFException
 import okio.IOException
 import okio.use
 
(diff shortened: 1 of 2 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -100,14 +102,25 @@ private constructor(
...
-      )
+    val apolloException = when(throwable) {
+      is ApolloException -> {
+        // This happens for malformed JSON
+        throwable
+      }
+      is IOException -> {
+        // This happens when JsonReader returns an IO error
+        ApolloNetworkException(
+            message = "IOError while parsing the HTTP network response",
```
BoD, line 113: Nitpick
```suggestion
            message = "IOException while parsing the HTTP network response",
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpNetworkTransport.kt
```diff
@@ -100,14 +103,32 @@ private constructor(
...
+      is ApolloException -> {
+        // This happens for malformed JSON
+        throwable
+      }
+      is EOFException -> {
+        // This happens for premature end of file
+        ApolloParseException(
+            message = "EOFException while parsing the HTTP network response",
+            cause = throwable
+        )
```
martinbonnin, line 116: This is an interesting one. It could be argued that EOFException is a network exception if the stream is closed too early

* If there is a `Content-Length` this doesn't happen because OkHttp has a specific `java.net.ProtocolException` for this
* I'm not 100% sure about chunked encoding
* Also not 100% clear about the other engines (Ktor, etc...)

martinbonnin:
Turns out it's not low hanging fruits and this raises more questions. Since it's not critical, I'll close this until we refine the exception handling of `HttpEngine`

PR #5078, opened 2023-07-10
[apollo-ast] Switch back the parser to Strings and add more tests

martinbonnin:
I changed my mind and moved the parser back to Strings. Contrary to responses, the GraphQL documents are going to be found as Strings most of the times. Either as bytecode Strings or if read from the network as Json strings. Also it's faster:

```
jmh summary:
Benchmark               Mode  Cnt    Score   Error  Units
Benchmark.antlrTest    thrpt   10  346.757 ± 3.759  ops/s
Benchmark.graphqlJava  thrpt   10  232.713 ± 1.821  ops/s
Benchmark.parserTest   thrpt   10  844.182 ± 8.621  ops/s
```

This PR also moves the tests to `commonTest` and adds a bunch of new Lexer tests


libraries/apollo-ast/src/commonTest/kotlin/com/apollographql/apollo3/graphql/ast/test/LexerTest.kt
```diff
@@ -0,0 +1,634 @@
...
+      assertEquals(1, line)
+      assertEquals(2, column)
+    }
+    expectLexerException("01.23") {
+      assertEquals("Invalid number, unexpected digit after 0: '1'", message)
+      assertEquals(1, line)
+      assertEquals(2, column)
+    }
+    expectLexerException("+1") {
+      assertEquals("Unexpected symbol '+' (0x2b)", message)
```
BoD, line 529: I assumed this was valid, TIL!

BoD (approved):
👍

PR #5081, opened 2023-07-10
Enable build scans again

martinbonnin:
Sorry I got confused about the wording but turns out both are required. 

1. is a Build Scan only and determines whether to publish or not
2. is Gradle Enterprise only and adds another condition on top of 1.

This reverts commit 775c1d869f4a85f9e4cd62cfbde4a663a4acc7e6.


martinbonnin:
Merging quickly since it's a revert of a very small patch

PR #5086, opened 2023-07-10
Remove golatac

martinbonnin:
Now that we have an optimized build, re-enable version catalog accessors, see if we still have those out-of-date issues in the composite builds.


BoD (approved):
👍

PR #5088, opened 2023-07-11
[4.0 cleanups] Remove DefaultImpls everywhere

martinbonnin:
Because we target JVM 8+ everywhere now, we can use default interface methods. 

This is a breaking ABI change but mostly source compatible (except for Java callers that would manually reference `DefaultImpls` but this was never a public API in the first place and for Kotlin 1.6.10- callers that need to opt-in jvm-default)

See https://github.com/apollographql/apollo-kotlin/issues/4171


gradle/libraries.toml
```diff
@@ -95,7 +95,7 @@ gr8 = { group = "com.gradleup", name = "gr8-plugin", version = "0.8" }
 # to avoid leaking the Kotlin stdlib on the classpath
 #
 # Keep in sync with MIN_GRADLE_VERSION
-gradle-api-min = { group = "dev.gradleplugins", name = "gradle-api", version = "5.6" }
+gradle-api-min = { group = "dev.gradleplugins", name = "gradle-api", version = "6.8" }
```
martinbonnin, line 98: 6.8 is the first Gradle version with Kotlin 1.4 support. 

Interestingly, Kotlin 1.3 can read 1.6 metadata but fails on `@JvmDefault` interfaces because there's also a per-class `versionRequirements` ([gist](https://gist.github.com/martinbonnin/c46a4b44fd3c4d5db6e737cc60dedd4e))

BoD (approved):
🧹

PR #5096, opened 2023-07-13
Fix defer with router tests

BoD:
[Sometimes](https://github.com/apollographql/apollo-kotlin/actions/runs/5538687166/jobs/10118285481#step:8:330), but not always, `localhost` can't be resolved by node - I don't know why.
This only happens with a recent version of node (recently changed by [this change](https://github.com/apollographql/apollo-kotlin/pull/5090/files?w=1#diff-131d3dcf6023f5c7633513f53d03510c711d3a4f4e3f44f1a1237554e8f505dfL29)), and I could only reproduce with the defer with router tests, but not with the other tests which also use `localhost` from `MockServer.url()` - but I changed it there too just in case, and to please the nodejs gods.

### DIFF
--- a/libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
+++ b/libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
@@ -50,7 +50,7 @@ actual class MockServer actual constructor(override val mockServerHandler: MockS
   }.listen()
 
   override suspend fun url() = url ?: suspendCoroutine { cont ->
-    url = "http://localhost:${server.address().unsafeCast<AddressInfo>().port}/"
+    url = "http://127.0.0.1:${server.address().unsafeCast<AddressInfo>().port}/"
     server.on("listening") { _ ->
       cont.resume(url!!)
     }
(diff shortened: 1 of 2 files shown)


martinbonnin (approved):

martinbonnin:
This is peculiar 🤔 . Does not give any extra reason why `localhost` does not work? DNS, route, timeout, anything else? 

If needed we can re-introduce the old nodejs version.

BoD:
Yeah it is very odd. I see this in node's logs:
```
NET 99031: connect: find host localhost
NET 99031: connect: dns options { family: undefined, hints: 1024 }
HTTP 99031: sockets localhost:4002: 1 1
HTTP 99031: write ret = true
HTTP 99031: outgoing message end.
(node:99031) Warning: Setting the NODE_DEBUG environment variable to 'http' can expose sensitive data (such as passwords, tokens and authentication headers) in the resulting log.
(Use `node --trace-warnings ...` to show where the warning was created)
NET 99031: _read
NET 99031: _read wait for connection
NET 99031: afterConnect
NET 99031: destroy
NET 99031: close
NET 99031: close handle
HTTP 99031: SOCKET ERROR: connect ECONNREFUSED ::1:4002 Error: connect ECONNREFUSED ::1:4002
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
``` 

Actually maybe I misinterpreted it, could it be a IPv6 issue instead?

martinbonnin:
> Actually maybe I misinterpreted it, could it be a IPv6 issue instead?

Looks like it. In that case forcing IPv4 is the good fix 👍

PR #5099, opened 2023-07-17
[IJ plugin] v3 -> v4 migration: deprecations/renames

BoD:
A few symbols that have been renamed or replaced by something else. There may be a few more coming next.


martinbonnin (approved):
👍

PR #5101, opened 2023-07-18
Move serialization outside of cache lock

ashare80:
Lock blocks reads and writes during processing of data before and after accessing cache unnecessarily.


libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/CacheDataTransformer.kt
```diff
@@ -0,0 +1,53 @@
+package com.apollographql.apollo3.cache.normalized.api.internal
+
+import com.apollographql.apollo3.api.CompositeAdapter
+import com.apollographql.apollo3.api.CustomScalarAdapters
+import com.apollographql.apollo3.api.Executable
+import com.apollographql.apollo3.api.fromJson
+import com.apollographql.apollo3.api.json.MapJsonReader
+import com.apollographql.apollo3.cache.normalized.api.CacheKey
+
+data class CacheDataTransformer<D: Executable.Data>(
```
martinbonnin, line 10:
```suggestion
internal class CacheDataTransformer<D: Executable.Data>(
```

martinbonnin (approved):
One minor suggestion, LGTM otherwise, Thanks for the contribution!

BoD (approved):
👍 Thanks!

BoD:
I missed this before but indeed `CacheDataTransformer` cannot be `internal` because now `fun <D : Executable.Data> Executable<D>.readDataFromCache()` returns it (not caught at build time because we don't track the api in the incubating module).

To keep the return type of `readDataFromCache` we could:
- add `.toData(customScalarAdapters)`
- use `readInternal` instead of `readDataFromCache` inside `readDataFromCacheInternal` (it would have to be `@ApolloInternal` instead of `private` then)

martinbonnin:
Thanks again for the contribution! @BoD @ashare80, I pushed a few changes to:

* keep non incubating artifacts in sync
* keep backward compat and use `@ApolloInternal` to make the symbols we don't want to expose harder to use. 

All in all, I think the `Executable.readDataFromCache` shouldn't have been part of the public API because they're not locking the store. Something to revisit before shipping v4.

ashare80:
@martinbonnin Awesome thanks!

PR #5102, opened 2023-07-18
apollo-ast: fix column of block strings

martinbonnin:
The first column was computed against the last line of the block string


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/Parser.kt
```diff
@@ -758,27 +747,27 @@ internal class Parser(src: String, val withSourceLocation: Boolean, val filePath
...
+          "schema" -> parseSchemaDefinition(start)
+          "scalar" -> parseScalarTypeDefinition(start)
+          "type" -> parseObjectTypeDefinition(start)
+          "interface" -> parseInterfaceTypeDefinition(start)
+          "union" -> parseUnionTypeDefinition(start)
+          "enum" -> parseEnumTypeDefinition(start)
+          "input" -> parseInputObjectTypeDefinition(start)
+          "directive" -> parseDirectiveDefinition(start)
+          "query", "mutation", "subscription" -> parseOperationDefinition(start)
+          "fragment" -> parseFragmentDefinition(start)
```
martinbonnin, line 770: Another small fix: include the potential description in the definitions.

BoD (approved):
👍

PR #5103, opened 2023-07-18
start end

martinbonnin:
- replace endLine/endColumn by start/end
- update API dump

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
@@ -3,22 +3,26 @@ package com.apollographql.apollo3.ast
 import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 
 /**
+ * @param start the offset where the symbol starts, inclusive, starting at 0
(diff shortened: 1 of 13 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/SourceLocation.kt
```diff
@@ -35,7 +39,22 @@ class SourceLocation(
   companion object {
     @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
     @Deprecated("SourceLocation is now nullable and this is replaced by null", ReplaceWith("null"), level = DeprecationLevel.ERROR)
-    val UNKNOWN = SourceLocation(-1, -1, -1, -1, null)
+    val UNKNOWN = SourceLocation.forPath(null)
+
+    /**
+     * Constructs a [SourceLocation] that only contains a filePath for the moments when we're constructing nodes programmatically
+     * but still want to carry around the path of the original nodes for debugging purposes.
+     * TODO: I'm not sure how much this is helping vs confusing. Why might want to remove that and just set a null sourceLocation in those cases
```
BoD, line 47:
```suggestion
     * TODO: I'm not sure how much this is helping vs confusing. We might want to remove that and just set a null sourceLocation in those cases
```

BoD (approved):
👍

PR #5104, opened 2023-07-18
More Apollo AST APIs

martinbonnin:
Add:
* `String.paseAsGQLDocument(ParserOptions)` because there's no need writing everything to utf-8 just to read it back to utf-16 (same for  `parseAsGQLValue`, `parseAsGQLType`, `parseAsGQLSelections`)
* `File.parseAsGQLDocument(ParserOptions)` because it can populate the `filePath` automatically 
* `Path.parseAsGQLDocument(ParserOptions)` because it can populate the `filePath` automatically 

I didn't add `File.parseAsGQLValue(ParserOptions)` and the likes because I don't think it makes much sense to have a file that only contains a value. And it's always easier to add than remove if we need.

These APIs are introduced as Stable. The `BufferedSource` equivalent are still experimental because I'm not sure how much we want to commit on `BufferedSource` (vs kotlinx-io or maybe just exposing String APIs)


BoD (approved):
Very cool!

PR #5106, opened 2023-07-18
Fix "no schema found" error message

martinbonnin:
Before that the error message was this:

```
"Schema(s) found: But none of them contain type definitions."
```


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -60,6 +60,10 @@ object ApolloCompiler {
       it.toSchemaGQLDocument()
     }
 
+    if (schemaDocuments.isEmpty()) {
+      error("No schema found. Apollo needs a `.graphqls` or a `.json` schema.")
```
BoD, line 64:
```suggestion
      error("No schemas found. Apollo needs a `.graphqls` or a `.json` schema.")
```

But we should confirm with a native english speaker to be sure 😅

BoD (approved):
👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
martinbonnin: Ping @Meschreiber. Do you know what's the correct english there?

Meschreiber: @martinbonnin , either could be correct depending on the context. If we're looking for multiple schemas and didn't find any, we can stick with "No schemas found."

martinbonnin: We're looking for only 1 "main" schema. There could other "extra.graphqls" files but most often, they should only contain type extensions, not type definitions.

Meschreiber: Got it. If we only care about finding the one main schema, "No schema found." is more accurate.

martinbonnin: Thank you 🙏 . @BoD I'll stick with the current version but I'm happy to discuss it with Carm over risotto 😄

BoD: Hahaha, with pleasure! 😂

PR #4663, opened 2023-02-01
Remove deprecated symbols < 3.7.0

martinbonnin:
I kept two because I haven't found good replacements:

* `HttpResponse.body(bodyString: ByteString)` 
* `DirectoryConnection.connectToAndroidSourceSet` 

All the rest is gone...

PR #5109, opened 2023-07-18
[IJ plugin] v3 -> v4 migration: deprecations/renames, part 2

BoD:
That should be all for the deprecations/renames from #4662 / #4663


martinbonnin (approved):
Let's hope we don't add too many more 🤞

PR #5114, opened 2023-07-21
[IJ plugin] v3 -> v4 migration: Gradle conf

BoD:
- enclose service configuration in a `service("service") {}` block
- `mapScalar` instead of `customScalarsMapping.set` / `.put`
- multi module: `apollo`/`service`/`dependsOn` instead of `dependencies`/`apolloMetadata`
- and a few renames

Also add the Migrate actions to Tools / Apollo (in addition to Refactor / Apollo) for increased visibility

### DIFF
--- a/intellij-plugin/src/test/testData/migration/v3-to-v4/multiModule_after.gradle.kts
+++ b/intellij-plugin/src/test/testData/migration/v3-to-v4/multiModule_after.gradle.kts
@@ -1,0 +1,23 @@
+dependencies {
+  implementation(project(":schema"))
+}
+
+apollo {
+  service("service1") {
+    packageName.set("com.example.service1")
+    dependsOn(project(":schema"))
+  }
+
+  service("service2") {
+    packageName.set("com.example.service2")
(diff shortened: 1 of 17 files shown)


intellij-plugin/src/test/testData/migration/v3-to-v4/multiModule_after.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+
+apollo {
+  service("service1") {
+    packageName.set("com.example.service1")
+    dependsOn(project(":schema"))
+  }
+
+  service("service2") {
+    packageName.set("com.example.service2")
+    dependsOn(project(":schema"))
```
martinbonnin, line 13: The multi-module + multi-service is a bit weird because before you could have matadata that was used by only one service while now I think it's going to fail if it doesn't find a schema for the given service in the `:schema` module. There's no way to guess just by looking at the Gradle config of a single project though so this is fine for me 👍

martinbonnin (approved):

PR #5117, opened 2023-07-24
Gradle plugin: make generateSourcesDuringGradleSync opt-in now that we have the IJ plugin

martinbonnin:
I've been burned a couple of times where the sync was failing because of bad .graphql files. Feels wrong to have no autocomplete in Gradle files because of some GraphQL files. Plus we have the IJ plugin to do it now.


BoD (approved):
👍

PR #5118, opened 2023-07-24
Initial CCN support

martinbonnin:
Initial CCN support. APIs are under a `@ApolloExperimental` flag for now so we can change if needed but the codegen uses it by default. I'm not super keen on adding a configuration option for that. I'll dig into the different versioning options that exist in GraphQL


libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -846,8 +846,9 @@ public abstract interface class com/apollographql/apollo3/ast/NodeTransformer {
 public final class com/apollographql/apollo3/ast/ParserOptions {
 	public static final field Companion Lcom/apollographql/apollo3/ast/ParserOptions$Companion;
 	public fun <init> ()V
-	public fun <init> (ZZZ)V
-	public synthetic fun <init> (ZZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (ZZZZ)V
+	public synthetic fun <init> (ZZZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun getAllowClientControlledNullability ()Z
```
martinbonnin, line 851: Might want to turn that into a Builder now that I think about it 🤔 . The more I do APIs, the more I like Builders...

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/Parser.kt
```diff
@@ -256,14 +216,71 @@ internal class Parser(src: String, private val withSourceLocation: Boolean, val
 
     val arguments = parseArguments(const = false)
     val directives = parseDirectives(const = false)
+    var nullability: GQLNullability? = null
+
+    if (allowClientControlledNullability) {
+      nullability = parseNullability()
```
BoD, line 222: Shouldn't this be before `parseDirectives`?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gql.kt
```diff
@@ -1493,14 +1494,94 @@ private fun List<GQLArgument>.writeArguments(writer: SDLWriter) {
   join(writer, prefix = "(", separator = ", ", postfix = ")")
 }
 
-class GQLField(
+
+@ApolloExperimental
+sealed interface GQLNullability : GQLNode
+
+@ApolloExperimental
+class GQLRequired(override val sourceLocation: SourceLocation? = null) : GQLNullability {
```
BoD, line 1502: This is very much nitpicking but would `GQLNonNullable` / `GQLNullable` be more consistent with `GQLNullability`, and the terms in the spec (well the spec as it is currently!)

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gqltype.kt
```diff
@@ -81,3 +83,45 @@ internal fun GQLType.isOutputType(typeDefinitions: Map<String, GQLTypeDefinition
...
+    return this
+  } else if (this !is GQLListType && nullability !is GQLListNullability) {
+    return this
+  } else {
+    error("")
+  }
+}
+
+@ApolloExperimental
+fun GQLType.withCcn(nullability: GQLNullability?): GQLType {
```
BoD, line 109: Nitpicking again, but we don't use "ccn" before so maybe `withNullability` / `withListNullability` would be slightly more consistent?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/Parser.kt
martinbonnin: Good catch 🙏

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gql.kt
martinbonnin: The language is a bit mixed in the RFC currently. There is also required/optional [here](https://github.com/graphql/graphql-wg/blob/main/rfcs/ClientControlledNullability.md#definitions)

I like required/optional. An `optional` field has a nullable designator. 
I like that I can use 2 different words for why a field type is nullable.

libraries/apollo-ast/api/apollo-ast.api
martinbonnin: Builder is in the place

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gql.kt
martinbonnin: I've been staring at this for the last 15min and I'm not sold on anything yet. We currently have:

* designator
* nullability
* requiredStatus

I'm currently thinking of something like `modifers`:

```
sealed interface GQLTypeModifier : GQLNode
class GQLListTypeModifier : GQLTypeModifier
class GQLRequiredTypeModifier : GQLTypeModifier
class GQLOptionalTypeModifier : GQLTypeModifier
```

I like that this maps with types because at this end of the day, the `GQLType` is the source of truth for most (all?) matters and what we're effectively doing is `val finalType = field.type.applyModifiers(field.typeModifier)`

I updated this PR to track the current state of the grammar. Let's bike shed this in an upcoming meeting?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gqltype.kt
martinbonnin: PR updated. Although at the moment I really like the idea of:

```kotlin
fun GQLType.applyModifier(modifier: GQLTypeModifier?): GQLType 
```

But maybe I've just been doing too much Jetpack Compose 😄

martinbonnin: Although `apply` kind of hints that the instance is modified in-place so maybe more 

```kotlin
fun GQLType.withModifier(modifier: GQLTypeModifier?): GQLType 
```

BoD (approved):
This looks good! 👍

PR #5120, opened 2023-07-24
Add `./gradle ft`

martinbonnin:
`ft` is for `fast test`. It'll run the JVM tests except the Gradle one and is faster to iterate during development


BoD (approved):
👍

PR #5121, opened 2023-07-24
Move apollo stuff outside of apollo ast

martinbonnin:
Things like key fields and typename additions do not belong in apollo-ast


BoD (approved):
👍

Issue #5126, opened 2023-07-26
Exist basic scalar type in downloaded schema and got Schema errors


sonatard:
### Question

<img width="781" alt="スクリーンショット 2023-07-26 13 11 38" src="https://github.com/apollographql/apollo-kotlin/assets/1586910/8a13fadc-376c-4059-8d06-c2a52bb2721d">

When I download the Schema with `gradlew :app:downloadApolloSchema` , the basic scalar type(ID, Int, Float...) exists in Schema file. Probably because of this, Android Studio is throwing an error, but how should I deal with this?

There is no problem with the operation such as building or sending requests.

Also, I sometimes write code to parse GraphQL schemas in Go, but I get the same error there and cannot parse it. Should the Go GraphQL parser be made not to throw an error, or should the schema not include a basic scalar type, which is correct?

martinbonnin:
Hi 👋 . This is a known [issue in the GraphQL IntelliJ plugin](https://github.com/JetBrains/js-graphql-intellij-plugin/issues/665). These types needs to be in the schema because there exists several versions of them (deprecated input fields, repeated directives, etc... have been introduced more recently). And we need to know what version is used to do proper validation.

martinbonnin:
It's affecting validation and/or introspection. Take the [latest builtin `__InputValue` type definition](https://spec.graphql.org/draft/#sel-GAJXNFAD0HAAC8IBmuX):

```graphql
type __InputValue {
  name: String!
  description: String
  type: __Type!
  defaultValue: String
  isDeprecated: Boolean!
  deprecationReason: String
}
```

And write this query:

```graphql
{
  __schema {
    types {
      inputFields {
        isDeprecated
      }
    }
  }
}
```

It will validate well for Apollo Kotlin because `isDeprecated` is in the schema. Now if you try to execute this against an [older version of the spec](https://spec.graphql.org/October2016/#sec-The-__InputValue-Type) that doesn't know about deprecated input fields, this query will error all the time. This is what we want to avoid by validating the queries client side.

sonatard:
> When representing a GraphQL schema using the type system definition language, all built-in scalars must be omitted for brevity.
> https://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars

@martinbonnin Hi! I found this section in GraphQL Draft spec.
It seems to me that built-in scalars should not exist in the schema file. What do you think?

martinbonnin:
[...]

PR #5131, opened 2023-07-26
[IJ Plugin] Don't report redefinitions of built-in types as errors

BoD:
See https://github.com/JetBrains/js-graphql-intellij-plugin/issues/665
I'll also open a PR to the the GraphQL plugin so hopefully we can remove this from ours.

See also #5126


martinbonnin (approved):
Very cool 🎉 . I guess scalar redefinitions are technically errors but doesn't hurt to be a bit lenient here.

PR #5134, opened 2023-07-28
[IJ plugin] Improve compat->operationBased migration

BoD:
Handle some more cases, after testing it on https://github.com/HedvigInsurance/android.

It's still not perfect of course (e.g, removed `__typename` are untouched as there's no accurate way to know if they should be removed) but should still save a lot of time/manual work.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -7,7 +7,6 @@
       </value>
     </option>
     <option name="RIGHT_MARGIN" value="140" />
-    <option name="ENABLE_SECOND_REFORMAT" value="true" />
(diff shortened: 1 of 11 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -7,7 +7,6 @@
       </value>
     </option>
     <option name="RIGHT_MARGIN" value="140" />
-    <option name="ENABLE_SECOND_REFORMAT" value="true" />
```
BoD, line 10: This has been coming back since I switched to IJ 2023.2, LMK if this causes problems.

martinbonnin (approved):

PR #5138, opened 2023-07-30
Add minVersion

Meschreiber:
This PR replaces uses of "available in version ..." with the new [<MinVersion> component](https://apollograph.slack.com/archives/C0721M2F6/p1690243600979049?thread_ts=1690218529.773819&cid=C0721M2F6).

It's likely not comprehensive but at least replaces previous mentions of versions. (Searched for "available", "version", and "<=" in .mdx files.)

### DIFF
--- a/docs/source/fetching/defer.mdx
+++ b/docs/source/fetching/defer.mdx
@@ -1,6 +1,7 @@
 ---
 title: 'Using the @defer directive in Apollo Kotlin'
 description: 'Fetch slower schema fields asynchronously'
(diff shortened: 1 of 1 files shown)


martinbonnin:
Tangential to that PR, we should start adding `@since` to the [KDoc API docs](https://www.apollographql.com/docs/kotlin/kdoc/index.html). I've added an issue for tracking [here](https://github.com/apollographql/apollo-kotlin/issues/5144)

Issue #3796, opened 2022-01-19
Support Ktor for all KMM Targets as optional


ProVir:
Hi!

Now there are good DefaultHttpEngine and they are enough for most tasks.

But having KtorHttpEngine as an optional solution would be great!

Example - the application uses GraphQL and REST API, I would like to have a common basis for forming clients, especially when the host is shared with the same requirements for headers, logging and other things.

There is such an implementation for JS, it seems it can be adapted and made universal.

BoD:
Hi!

Indeed it looks like the code that already exists in the JavaScript target could be generalized.

One thing we don't have yet is the WebSocket side of this - but it would be great to have it since it is currently missing from the JavaScript target. That could prove trickier to implement however!

A thought: if we went ahead with this, it would certainly be in the form of an independent module (so projects can opt-in), and keep OkHttp/NSURLRequest by default. A drawback being that you now potentially depend on 2 HTTP libraries, even though only one is used.

martinbonnin:
+1 for the separate module for `apollo-engine-ktor`. 

For users that do not want to pull `OkHttp` in the classpath, a short term option is to exclude it using Gradle. I think that'd work.  If that becomes a bigger issue, I guess we could do `apollo-runtime-core` or something like this without the `OkHttp` dependency

Nailik:
Note on this request.
For Android theoretically the JS Http Engine could be used
https://github.com/apollographql/apollo-kotlin/blob/c6a16054c57e9c3b2c2c93f895d3169b37186d8a/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt

On iOS this results in an issue with the okio.Buffer because it is freezed.
You get an `kotlin.native.concurrent.InvalidMutabilityException: mutation attempt of frozen okio.Buffer@1ab1928` inside `/apollo3/api/json/BufferedSourceJsonReader.kt:706:12`
maybe the freeze occures inside the GCDWorker
https://github.com/apollographql/apollo-kotlin/blob/c6a16054c57e9c3b2c2c93f895d3169b37186d8a/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/internal/GCDWorker.kt


<details>
  <summary>Full Error output</summary>
  
  ```
[...]

PR #5142, opened 2023-08-02
Support Ktor Engine

yt8492:
Add to apollo-runtime/commonMain: 
- KtorHttpEngine
- KtorWebsocketEngine
- Extension for builder

implements for #3796


BoD:
Thanks a lot for the contribution! 🙏

A few thoughts:

- We will want to move this to a dedicated module, to avoid having `apollo-runtime` depending on Ktor
- It would be nice if we could reuse `KtorHttpEngine` /  `KtorWebSocketEngine`for Js instead of it being duplicated (not 100% sure how to do this - it depends on how we proceed for the 1st item above)
- We may need a few more tests (in particular around subscriptions/websockets)

We can help or take care of these of course!

martinbonnin:
> It would be nice if we could reuse KtorHttpEngine / KtorWebSocketEnginefor Js instead of it being duplicated

I've been thinking about this, it's a pain because of a circular module dependency :(. `HttpEngine` is defined in `apollo-runtime`. So if we want to have `apollo-runtime` depend on `apollo-engine-ktor`, we also need `apollo-engine-ktor` to depend on `apollo-runtime` :/. We _could_ introduce a `apollo-runtime-api` but it's quite the change. I'd prefer duplicating the implementations for the time being.

gradle/libraries.toml
```diff
@@ -30,11 +30,11 @@ kotlin-plugin-min = "1.6.0"
 kotlin-plugin = "1.9.0"
 kotlin-plugin-max = "1.9.0"
 kotlin-stdlib = "1.6.21"
-kotlinx-coroutines = "1.6.4"
+kotlinx-coroutines = "1.7.3"
 kotlinx-datetime = "0.4.0"
 kotlinx-serialization-runtime = "1.5.0"
 ksp = "1.9.0-1.0.11"
-ktor = "2.2.2"
+ktor = "2.3.2"
```
BoD, line 37: Turns out 2.3.3 was just released
```suggestion
ktor = "2.3.3"
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt
```diff
@@ -0,0 +1,75 @@
...
+import io.ktor.http.HttpHeaders
+import io.ktor.util.flattenEntries
+import okio.Buffer
+import kotlin.coroutines.cancellation.CancellationException
+
+class KtorHttpEngine(
+    private val client: HttpClient,
+): HttpEngine {
+
+  var disposed = false
```
BoD, line 23: Can probably be private

yt8492:
@BoD 
Thank you for reviewing my changes! I applied your suggestion to this pull request.

BoD:
Thank you so much @yt8492, this looks good!

Meanwhile, I've found a few issues with the WebSocketEngine that made a few tests fails, but it looks like I just managed to fix the last one. I will rebase my changes on your branch and push them here.

BoD (approved):
👍

martinbonnin (approved):
Thanks!

PR #5147, opened 2023-08-03
Update KotlinPoet

martinbonnin:
2 behaviour changes:

* We now need to specify the return type of functions with expression body (https://github.com/square/kotlinpoet/pull/1559)
* Imports are now automatically aliased (https://github.com/square/kotlinpoet/pull/1355)


BoD (approved):
Nice! The generated code is definitely a bit nicer to read

PR #5148, opened 2023-08-03
Allow explicit CCN syntax

martinbonnin:
Sync with https://github.com/graphql/graphql-spec/pull/895#discussion_r1279607241

Now all of these are allowed (and equivalent):

```
  list
  list[]
```

```
  list!
  list[]!
```

```
  matrix[!]
  matrix[[]!]
```


BoD (approved):
👍

PR #5158, opened 2023-08-04
Add a test for field names that have the same name as an enum type

martinbonnin:
Another low hanging fruit
Closes https://github.com/apollographql/apollo-kotlin/issues/1156


BoD (approved):
👍

PR #5161, opened 2023-08-07
JSON: Remove limitation on the JSON nesting

martinbonnin:
If the JSON is too nested, an OutOfMemory exception will be thrown.

Closes https://github.com/apollographql/apollo-kotlin/issues/5159


BoD (approved):
🚀

PR #5162, opened 2023-08-07
Add `mergeExtensions` and `toFullSchemaGQLDocument`

### LINKED ISSUE #4920: apollo-ast: support merging definitions without validation
### Use case

```
val parsedDocument = Buffer().writeUtf8(sdl).parseAsGQLDocument().valueAssertNoErrors()
val normalizedDefinitions = normalizeFederation(parsedDocument)
val mergedDefinitions = // merge definitions
```

_No response_

martinbonnin:
* `GQLDocument.mergeExtensions()` to merge extensions without doing schema validation (closes #4920)
* `GQLDocument.toFullSchemaGQLDocument()` to generate a verbose GQL document that merges the extensions and also contains the current builtin types as supported by Apollo Kotlin (see also https://github.com/graphql/graphql-spec/pull/1037)
* Hide most of the introspection classes from the public API (see https://github.com/apollographql/apollo-kotlin/issues/4171)

This pull request unfortunately contains a bunch of both binary and source breaking changes. Since `apollo-ast` was not advertised as much as `apollo-api` and `apollo-runtime`, I'm hoping this is OK. I took this opportunity to tweak a bit the public API:

* `okio` is not part of stable the public API anymore (it still is experimentally available though)
* It's not possible to go from `String`/`File` to `Schema` directly anymore, you have to go through `GQLDocument`:

```kotlin
// Before
Buffer().writeUtf8("type Query { foo: Int }").toSchema()

// After
"type Query { foo: Int }".toGQLDocument().toSchema()
```

More comments inline


build-logic/src/main/kotlin/Common.kt
```diff
@@ -5,7 +5,7 @@ import org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest
 fun Project.commonSetup() {
   pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
     tasks.register("ft") {
-      if (this@commonSetup.name != "apollo-gradle-plugin") {
+      if (this@commonSetup.name !in setOf("apollo-gradle-plugin", "intellij-plugin")) {
```
martinbonnin, line 8: Removed intellij-plugin from `ft` because they're a tiny bit slow but also very prone to flakes. I've had them fail on multiple occasions.

libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -11,7 +11,20 @@ public final class com/apollographql/apollo3/ast/ApolloParser {
...
+	public static synthetic fun toGQLNullability$default (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;ILjava/lang/Object;)Lcom/apollographql/apollo3/ast/GQLNullability;
+	public static final fun toGQLSelections (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;)Ljava/util/List;
+	public static synthetic fun toGQLSelections$default (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;ILjava/lang/Object;)Ljava/util/List;
+	public static final fun toGQLType (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;)Lcom/apollographql/apollo3/ast/GQLType;
+	public static synthetic fun toGQLType$default (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;ILjava/lang/Object;)Lcom/apollographql/apollo3/ast/GQLType;
+	public static final fun toGQLValue (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;)Lcom/apollographql/apollo3/ast/GQLValue;
+	public static synthetic fun toGQLValue$default (Ljava/lang/String;Lcom/apollographql/apollo3/ast/ParserOptions;ILjava/lang/Object;)Lcom/apollographql/apollo3/ast/GQLValue;
```
martinbonnin, line 27: Directly promoted to stable.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/api.kt
```diff
@@ -19,16 +21,6 @@ import okio.use
...
 
-/**
- * Parses the source to a [Schema], throwing on parsing or validation errors.
- *
- * See [parseAsGQLDocument] and [validateAsSchema] for more granular error reporting
- */
-@ApolloExperimental
-fun BufferedSource.toSchema(filePath: String? = null): Schema = parseAsGQLDocument(filePath).getOrThrow().validateAsSchema().getOrThrow()
-
-fun String.toSchema(): Schema = parseAsGQLDocument().getOrThrow().validateAsSchema().getOrThrow()
```
martinbonnin, line 30: Breaking change

BoD (approved):
👍

build-logic/src/main/kotlin/Common.kt
```diff
@@ -37,6 +37,8 @@ private fun Project.configureTestAggregation() {
       attribute(USAGE_ATTRIBUTE, objects.named(Usage::class.java, "apolloTestAggregation"))
     }
   }
+  // Hide this from the 'assemble' task
+  configuration.setVisible(false)
```
martinbonnin, line 41: This is a fix for a build regression introduced in https://github.com/apollographql/apollo-kotlin/pull/5181 that made `:apollo-gradle-plugin:testJava11` depend on `:apollo-*:test`

libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -125,7 +138,6 @@ public final class com/apollographql/apollo3/ast/GQLDirectiveLocation : java/lan
 public final class com/apollographql/apollo3/ast/GQLDocument : com/apollographql/apollo3/ast/GQLNode {
 	public static final field Companion Lcom/apollographql/apollo3/ast/GQLDocument$Companion;
 	public fun <init> (Ljava/util/List;Lcom/apollographql/apollo3/ast/SourceLocation;)V
-	public fun <init> (Ljava/util/List;Ljava/lang/String;)V
```
martinbonnin, line 128: Breaking change but keeping both constructor is painful because `GQLDocument(definitions, null)` becomes ambiguous.

martinbonnin:
I reshuffled the introspection bits again into 2 pieces:

* `introspection_reader` contains a lenient model of introspection and tries to make a best effort guess when some fields are absent. When recoverable (description missing, directive defintions missing, etc...) a log is printed. When unrecoverable (field definitions, arguments, enumValues, etc...) an error is thrown
* `introspection_writer` contains a strict model of the current introspection schema we support.

It's a bit verbose but that duplication makes it easier to handle the edge cases of different servers having different version of introspection

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/introspection_reader.kt
```diff
@@ -0,0 +1,556 @@
...
+
+  /**
+   * This assumes the case where `deprecationReason == null && isDeprecated` is not valid
+   * which is unclear from reading the spec as the below seems allowed:
+   *
+   * ```
+   * type Query {
+   *   foo: Int @deprecated(reason: null)
+   * }
+   * ```
```
BoD, line 389: Could we handle this case by passing `isDeprecated` + `deprecationReason`?

libraries/apollo-ast/src/jvmTest/kotlin/com/apollographql/apollo3/graphql/ast/test/IntrospectionTest.jvm.kt.kt
```diff
@@ -0,0 +1,54 @@
+package com.apollographql.apollo3.graphql.ast.test
```
BoD, line 1: File name has `.kt` twice

libraries/apollo-gradle-plugin/src/test/kotlin/util/TestUtils.kt
```diff
@@ -38,7 +38,7 @@ object TestUtils {
       block(dest)
     } finally {
       // Comment this line if you want to keep the directory around during development
-      dest.deleteRecursively()
+      //dest.deleteRecursively()
```
BoD, line 41:
```suggestion
      dest.deleteRecursively()
```

BoD (approved):
👍

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/introspection_reader.kt
martinbonnin: The whole IR is based on the premise that `deprecationReason != null` implies `isDeprecated = true`. Kotlin `@Deprecated` annotation also requires a message. I'm still unclear what the use case is for `@deprecated(reason: null)`. I'll add a comment there

PR #5164, opened 2023-08-07
Use compilations instead of multiple mpp targets for java codegen tests

martinbonnin:
Inspired from https://youtrack.jetbrains.com/issue/KT-61007

This introduces an intermediate `kotlinCodegenTest` source set that is common to everything except `javaCodegen`. Our sourceSet graph for tests is now like this:

mermaid-diagram-2023-08-07-195957


martinbonnin:
Build scans

* Before
  * apollo-kotlin https://ge.apollographql.com/s/ezzplv3ttifbw
  * apollo-tests https://ge.apollographql.com/s/oat2hvd4xf2ge
* After
  * apollo-kotlin https://ge.apollographql.com/s/iyaarjqicguka
  * apollo-tests https://ge.apollographql.com/s/5npnxfyyupr74

There are more native tasks executed [here](https://ge.apollographql.com/s/oat2hvd4xf2ge/timeline?toggled=WyIwIl0&view=by-type#0), will dig into it

martinbonnin:
We used to add targets for both `macosArm64` and `macosX64` while tests were running only on one of those (but compilation can be run for both). By only adding the host target, we should save a bit of time.

martinbonnin:
[1131 tests executed](https://ge.apollographql.com/s/7c6rrmgs6nkl2/console-log?page=4#L3637)

[macosX64 tests are executed here](https://ge.apollographql.com/s/7c6rrmgs6nkl2/timeline?toggled=WyIxNDEiXQ&view=by-type#141)

Merging this.

PR #5178, opened 2023-08-10
Persisted queries doc

Meschreiber:
Currently the PQ updates are only appearing on the v4 kotlin docs: https://www.apollographql.com/docs/kotlin/v4/advanced/persisted-queries/

Since v3.8.2 also supports PQs they should appear in v3 (https://www.apollographql.com/docs/kotlin/advanced/persisted-queries) as well.


Meschreiber:
Whoops. Closing in favor of https://github.com/apollographql/apollo-kotlin/pull/5179

PR #5181, opened 2023-08-11
[Infra] Count tests in CI

martinbonnin:
When refactoring the KMP setup, it's easy to miss some targets and accidentally remove some tests. This PR prints the number of executed tests at the end of `ciBuild` so we can keep track of this and a `build/apolloTestAggregation.txt` file that shows all the test class that we can diff on if some tests are missing.

Also it's a cool statistic to keep track of. As of today, there are 1130 (tests) + 1650 (apollo-kotlin) = 2780 tests


BoD (approved):
👍

PR #5187, opened 2023-08-16
Add WebSocketMockServer and tests for WebSocketEngine 

BoD:


libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/ws/NSURLSessionWebSocketEngine.kt
```diff
@@ -41,10 +49,16 @@ actual class DefaultWebSocketEngine(
 
   actual constructor() : this(
       webSocketFactory = { request, connectionListener ->
+        // We must not use the main thread's queue for the delegateQueue
```
BoD, line 52: That's the fix for the WebSocket not working on Apple (regression introduced in #4287)

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
```diff
@@ -0,0 +1,39 @@
...
+  val events: Flow<WebSocketEvent>
+
+  suspend fun sendText(sessionId: String, text: String)
+  suspend fun sendBinary(sessionId: String, binary: ByteArray)
+  suspend fun sendClose(sessionId: String, reasonCode: Short? = null, reasonMessage: String? = null)
+  fun stop()
+}
+
+@ApolloExperimental
+expect fun WebSocketMockServer(port: Int = Random.nextInt(10000, 20000)): WebSocketMockServer
```
martinbonnin, line 39: Yay for constructor fun!

martinbonnin, line 39: Can we let the system allocate a free port instead of allocating a random one ourselves?

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt
```diff
@@ -27,7 +27,7 @@ actual class DefaultHttpEngine constructor(private val connectTimeoutMillis: Lon
     expectSuccess = false
     install(HttpTimeout) {
       this.connectTimeoutMillis = this@DefaultHttpEngine.connectTimeoutMillis
-      this.socketTimeoutMillis = this@DefaultHttpEngine.readTimeoutMillis
+      this.requestTimeoutMillis = this@DefaultHttpEngine.readTimeoutMillis
```
martinbonnin, line 30: Can we document what the different options do here? Also make sure the comment [here](https://github.com/apollographql/apollo-kotlin/blob/f1d5fd10730b8c011da83d17e1580600d29374c3/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/http/HttpEngine.kt#L33) is still up to date?

tests/engine/src/commonTest/kotlin/GzipTest.kt
```diff
@@ -50,5 +47,12 @@ class HttpEngineTest {
...
+    engine.dispose()
   }
-}
\ No newline at end of file
+
+  @Test
+  fun gzipTestDefault() = gzipTest(DefaultHttpEngine())
+
+  @Test
+  fun gzipTestKtor() = gzipTest(KtorHttpEngine())
```
martinbonnin, line 57: Any objection to have 2 test tasks that substitute the engine constructor? Like we do for integration tests and Java vs Kotlin models? I can try in a follow up PR if you're up for it

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@ApolloExperimental
+interface WebSocketMockServer {
+  @ApolloExperimental
+  sealed class WebSocketEvent {
+    @ApolloExperimental
+    class Connect(val sessionId: String, val headers: Map<String, String>) : WebSocketEvent()
+
+    @ApolloExperimental
+    class TextFrame(val sessionId: String, val text: String) : WebSocketEvent()
```
martinbonnin, line 15: Naming bikeshedding: I thought a Text message was split accross multiple frames and therefore this would need to be called `TextMessage` (or just `Text`)? Similarly, `WebSocketEvent` could be named `WebSocketMessage`?

tests/engine/src/commonTest/kotlin/WebSocketEngineTest.kt
```diff
@@ -0,0 +1,177 @@
...
+      webSocketServer.sendText(sessionId, "server->client")
+      val message = connection.receive()
+      assertEquals("server->client", message)
+
+      connection.close()
+      val closeEvent = awaitItem()
+      assertTrue(closeEvent is WebSocketEvent.Close)
+
+      cancelAndIgnoreRemainingEvents()
+    }
```
martinbonnin, line 47: Nice test!

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  fun start()
+  fun url(): String
+
+  val events: Flow<WebSocketEvent>
+
+  suspend fun sendText(sessionId: String, text: String)
+  suspend fun sendBinary(sessionId: String, binary: ByteArray)
+  suspend fun sendClose(sessionId: String, reasonCode: Short? = null, reasonMessage: String? = null)
+  fun stop()
```
martinbonnin, line 35: Let's not repeat the `ApolloClient` mistake. Call this `close` and implement `Closeable`?

martinbonnin (commented):
A few comment but LGTM overall. A nice building block for future WebSockets improvements

tests/engine/src/commonTest/kotlin/GzipTest.kt
BoD: Yes that would be nice, duplicating all test methods is annoying. Not so sure about how to do it though 😅 so you can go ahead if you have ideas 🙏

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
BoD: Turns out, yes :) Passing 0 does that.

libraries/apollo-mockserver/build.gradle.kts
```diff
@@ -37,13 +37,29 @@ kotlin {
...
+          exclude(group = "com.apollographql.apollo3", module = "apollo-mockserver")
         }
         implementation(project(":apollo-runtime")) {
           because("We need HttpEngine for SocketTest")
         }
       }
     }
   }
+
+  val commonAppleJvmMain = sourceSets.create("commonAppleJvmMain").apply {
```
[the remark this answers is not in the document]
BoD: Done in 0bae9cf2cf5b63ee121abe66ee6d68eab6568e74. Not 100% sure about the test sourceset dependencies though!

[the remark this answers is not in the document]
martinbonnin: Let's see :)

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt
```diff
@@ -17,10 +17,16 @@ import io.ktor.util.flattenEntries
 import io.ktor.utils.io.CancellationException
 import okio.Buffer
 
-
+/**
+ * @param connectTimeoutMillis The connection timeout in milliseconds. The connection timeout is the time period in which a client should establish a connection with a server.
+ * @param readTimeoutMillis The request timeout in milliseconds. The request timeout is the time period required to process an HTTP call: from sending a request to receiving a response.
+ */
 actual class DefaultHttpEngine constructor(private val connectTimeoutMillis: Long, private val readTimeoutMillis: Long) : HttpEngine {
```
martinbonnin, line 24: While we're at it, do you mind renaming `readTimeoutMillis` to `requestTimeoutMillis`? Ultimately, I'd like us to support `readTimeoutMillis` because:

* I think it's a better parameter because it does not depend on the response size
* `requestTimeout` can be emulated quite easily using `withTimeout` while `readTimeoutMillis` actually requires low-level configuration

martinbonnin: So all in all it'd be:

* `connectTimeout`: the initial TCP handshake timeout
* `readTimeout`: the socket timeout on subsequent reads
* `requestTimeout` (not very useful): the overall timeout of the request

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
BoD: In `events` however you can receive "frames" (which I agree it could make sense to rename "messages"), but also other things that are not really messages  (connect, close, error) IMO.

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt
BoD: I think renaming `readTimeoutMillis` to `requestTimeoutMillis` is breaking right? 

Also, in my tests, `readTimeoutMillis` didn't seem to work actually 😅 But I'll try again to confirm.

In any case I think it makes sense to add a constructor with the 3 parameters? Give more power to the user.

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/WebSocketMockServer.kt
martinbonnin: Gotcha 👍 

I was expecting an API similar to MockWebserver where we don't have anything like `Connect` or `Error`. All we can get is the HTTP request. 

Similarly for WebSockets, I would expect to be able to record the initial HTTP handshake request as well as subsequent Close/Ping/Pong control frames (or messages because these messages are always a single frame) and data messages (not individual frames).

Just my 2 cents, feel free to discard.

libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/http/KtorHttpEngine.kt
martinbonnin: > I think renaming readTimeoutMillis to requestTimeoutMillis is breaking right?

Indeed, source breaking, which unfortunately isn't tracked by [kbcv](https://github.com/Kotlin/binary-compatibility-validator). 

> In any case I think it makes sense to add a constructor with the 3 parameters? Give more power to the user.

I'd vote for only 2: `connectTimeout` and `readTimeout`. `requestTimeout` belongs in higher layers using [withTimeout](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html). That is if `readTimeout` is working as expected of course.

BoD: I just checked and `socketTimeout` doesn't work on Ktor JS. I'll open an issue. In the meantime we could use `socketTimeout` for Jvm and Apple, but `requestTimeout` for JS?

martinbonnin: I think so? If you're talking about the expect/actual constructor that takes a simple `timeoutMillis` parameter then it's fine to map it to `socketTimeout` on Jvm/Apple and `requestTimeout` on JS as long as it's properly documented.

Then for overloaded (non-actual) constructors, my preference would go to `connectTimeout` and `readTImeout` and leave `requestTimeout` out of it.

martinbonnin (approved):

PR #5191, opened 2023-08-17
Schema Nullability Extensions

martinbonnin:
Add Schema Nullability Extensions. This is similar to a `@nonnull` schema directive but works with lists. 

Given the following SDL:

```graphql
# schema.graphqls
type Query {
  random: Int
  list: [String]
  required: Int!
}
```

You can extend it like so:

```graphql
# extra.graphqls
extend type Query {
  # make random non-nullable
  random: Int!
  # make list and list items non-nullable
  list: [String!]!
  # make required nullable
  required: Int
  # add a new field
  new: Float
}
```

### DIFF
--- a/libraries/apollo-ast/api/apollo-ast.api
+++ b/libraries/apollo-ast/api/apollo-ast.api
@@ -838,6 +838,10 @@ public final class com/apollographql/apollo3/ast/IssueKt {
 	public static final fun containsError (Ljava/util/List;)Z
 }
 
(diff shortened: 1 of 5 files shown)


BoD (approved):

martinbonnin:
> this isn't allowed in the spec, right?

Indeed, this is not. allowed. Only new fields are allowed. This is an Apollo addition.

PR #5192, opened 2023-08-17
Fix `Optional<V>.getOrThrow()` when V is nullable

martinbonnin:
The following would throw :(

```kotlin
val optional = Optional.present<String?>(null)
val value = optional.getOrThrow()

```


BoD (commented):
👍

PR #5194, opened 2023-08-17
Deprecate addTypename = "ifFragments" 

martinbonnin:
Keep the default to `"ifFragments` for now as changing to `"always"` has a bunch of implications, especially in tests. Will change the default once we have a more compelling Data Builders story


BoD (approved):
👍

Issue #5172, opened 2023-08-10
[IJ/AS plugin] Ignore id field in Unused field inspection


sonatard:
### Use case

I want to ignore the id field in Unused field inspection.
This is because the id field is retrieved for cache updates, but it is very often not accessed from Kotlin.

### Describe the solution you'd like

id: ID is ignore, Specify the field name to ignore or Specify the type to ignore

BoD:
Hi! 

That's a good point, I think there might be a lot of cases where you need to select fields only for the cache to be up to date, but they won't be used in the Kotlin code and thus highlighted. This shouldn't be specific to id fields though?

You can suppress the inspection with `# noinspection ApolloUnusedField` on top of an operation / fragment. Unfortunately at the moment this doesn't work on individual fields though.

Maybe can you elaborate with your concrete example? Thanks!

sonatard:
> This shouldn't be specific to id fields though?
Yes. The mutation response is often not used directly by Kotlin, and is mostly for updating the cache.

> Maybe can you elaborate with your concrete example? Thanks!

The IDs such as lines.nodes.id, productVariant.id, and product.id are not referenced from Kotlin.

```graphql
query StorefrontAddCartLine($cartID: ID!, $countryCode: CountryCode!, $languageCode: LanguageCode!) @inContext(country: $countryCode, language: $languageCode) {
    cart(id: $cartID) {
        id
        totalQuantity
        lines(first: 100) {
            nodes {
                id # only for cache
                quantity
                merchandise {
                    ... on ProductVariant {
                        id # only for cache
                        product {
                            id # only for cache
                            tags
                        }
                    }
                }
            }
        }
    }
}
```

BoD:
Are you [declaring](https://www.apollographql.com/docs/kotlin/caching/declarative-ids) your id fields with `@typePolicy`? If yes you should be able to omit the `id` fields in your selections, they will still automatically be selected under the hood.

sonatard:
I didn't know that the cache would be updated even if the id field was omitted. It's wonderful news!

Can the id field be omitted in the same way with Programmatic cache IDs? Defining Declarative cache IDs for many types is cumbersome, so I would like to achieve this with Programmatic cache IDs if possible.

BoD:
[...]

PR #5197, opened 2023-08-18
[IJ/AS plugin] Add options to ignore fields when reporting unused fields

BoD:
https://github.com/apollographql/apollo-kotlin/assets/372852/d2252c77-379d-432e-a04a-d3bb65137260

Related #5172


sonatard:
Hi! Is it possible to share this setting with the team?

BoD:
> Hi! Is it possible to share this setting as a config file with the team?

Hi @sonatard yes if your inspections profile is set to the project (I think it's the default), then this will automatically be saved in `.idea/inspectionProfiles/` which you can keep in git so everybody on the project has the same. This can also be done one-time by exporting/importing inspection profiles from the settings.

sonatard:
Thanks for your work!

PR #5199, opened 2023-08-21
Enable configuration cache

martinbonnin:
Most of the issues we had were side effects of Dokka not supporting configuration cache (see https://youtrack.jetbrains.com/issue/KT-60311/).

Dokka is invoked from the publish task for `apollo-gradle-plugin:testJavaXY`. In order to remove that dependency, I duplicated the publications in all projects. There is now additional "${Target}Test" publications that do not include Kdoc.

Because all the publications are now duplicated, relying on `publishAllPublicationsTo${RepoName}Repository` is an error because it creates overlapping files. Instead, there are `publishMainTo${RepoName}` and `publishTestTo${RepoName}`:

* Instead of `publishToMavenLocal`, call `publishMainToMavenLocal`
* Instead of `publishAllPublicationsToOssStagingRepository`, call `publishMainToOssStaging`
* Instead of `publishAllPublicationsToPluginTestRepository`, call `publishTestToPluginTest`

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -29,7 +29,7 @@ jobs:
           ./gradlew --stop
       - run: |
           # Build the libs
-          ./gradlew publishAllPublicationsToPluginTestRepository -x dokkaHtml
+          ./gradlew publishTestToPluginTest
       - run: |
           # Build the benchmark apks
           ./gradlew -p benchmark assembleRelease assembleReleaseAndroidTest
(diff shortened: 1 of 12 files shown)


martinbonnin:
Closing because of https://youtrack.jetbrains.com/issue/KT-61359
At the moment, I'd rather wait ~10-20s for configuration to happen than put too many workarounds in the build. Will revisit in a bit. If someone feels like configuration time is too long and this is important, please leave a message

PR #5204, opened 2023-08-23
Bump uuid and okio

martinbonnin:
We now require Kotlin 1.8+

See https://github.com/apollographql/apollo-kotlin/issues/5046#issuecomment-1689093251


BoD (approved):
🚀

Issue #5171, opened 2023-08-09
Consider migrating generated code to use Enum.entries instead of Enum.values() for Kotlin 1.9 and onwards


StylianosGakis:
### Use case

With this [KEEP](https://github.com/Kotlin/KEEP/blob/master/proposals/enum-entries.md)
 now being released for Kotlin 1.9, it seems like it's always a good idea to use `entries` instead of `values()`
 
This discussion started here for reference https://slack-chats.kotlinlang.org/t/13888266/as-i-was-looking-into-migrating-from-values-to-entries-for-a#349eb4bd-1ea5-44d2-998a-9e151a7f4624

Just as Martin mentioned in this discussion, this should probably be configurable to still support clients using Kotlin versions that are older than 1.9

### Describe the solution you'd like

The generated code does not have any `values()` references, but instead uses `entries` for all enums where `values()` was used before.

PR #5208, opened 2023-08-24
Use entries instead of values() when using Kotlin 1.9

BoD:
Fix for #5171

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo3.compiler.codegen.kotlin.file
 
+import com.apollographql.apollo3.compiler.TargetLanguage
 import com.apollographql.apollo3.compiler.codegen.Identifier
 import com.apollographql.apollo3.compiler.codegen.Identifier.UNKNOWN__
 import com.apollographql.apollo3.compiler.codegen.kotlin.CgFile
@@ -79,10 +80,22 @@ internal class EnumAsEnumBuilder(
         .addKdoc("Returns all [%T] known at compile time", selfClassName)
         .maybeSuppressDeprecation(enum.values)
         .maybeAddOptIn(context.resolver, enum.values)
-        .returns(KotlinSymbols.Array.parameterizedBy(selfClassName))
+        .returns(
+            if (context.isTargetLanguageVersionAtLeast(TargetLanguage.KOTLIN_1_9)) {
+              KotlinSymbols.List.parameterizedBy(selfClassName)
+            } else {
+              KotlinSymbols.Array.parameterizedBy(selfClassName)
+            }
+        )
         .addCode(
             CodeBlock.builder()
-                .add("return·arrayOf(\n")
+                .add(
+                    if (context.isTargetLanguageVersionAtLeast(TargetLanguage.KOTLIN_1_9)) {
+                      "return·listOf(\n"
+                    } else {
+                      "return·arrayOf(\n"
+                    }
+                )
                 .indent()
                 .add(
                     values.map {
@@ -97,12 +110,13 @@ internal class EnumAsEnumBuilder(
   }
 
   private fun IrEnum.safeValueOfFunSpec(): FunSpec {
+    val entries = if (context.isTargetLanguageVersionAtLeast(TargetLanguage.KOTLIN_1_9)) "entries" else "values()"
     return FunSpec
         .builder("safeValueOf")
         .addParameter("rawValue", String::
         class)
         .returns(selfClassName)
-        .addStatement("return·values().find·{·it.rawValue·==·rawValue·} ?: $UNKNOWN__")
+        .addStatement("return·$entries.find·{·it.rawValue·==·rawValue·} ?: $UNKNOWN__")
         .build()
   }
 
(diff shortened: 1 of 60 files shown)


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -79,10 +80,22 @@ internal class EnumAsEnumBuilder(
         .addKdoc("Returns all [%T] known at compile time", selfClassName)
         .maybeSuppressDeprecation(enum.values)
         .maybeAddOptIn(context.resolver, enum.values)
-        .returns(KotlinSymbols.Array.parameterizedBy(selfClassName))
+        .returns(
+            if (context.isTargetLanguageVersionAtLeast(TargetLanguage.KOTLIN_1_9)) {
```
BoD, line 84: Not sure about this.
- Pro: consistency with `entries`
- Con: it's a breaking change

BoD: In fact it's not that consistent, because it should be named `knownEntries` ... but that's "even more breaking" 😅

martinbonnin: Add `knownEntries` and deprecated `knownValues` ?

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/KotlinPluginFacade.kt
```diff
@@ -7,25 +7,38 @@ import com.apollographql.apollo3.gradle.api.kotlinMultiplatformExtension
...
     null -> {
-      // User didn't specify a version: defaults to the Kotlin plugin's version
-      // Commented for now as the only possible outcome is to target 1.5
-      // val majorMinor = project.getKotlinPluginVersion()!!.take(3)
-      TargetLanguage.KOTLIN_1_5
+      // User didn't specify a version: use apiVersion and languageVersion to choose the default
+      // Fallback to the Kotlin plugin version if apiVersion or languageVersion are not set
+      val pluginVersion = project.getKotlinPluginVersion().substringBeforeLast(".")
+      val kotlinOptions = project.tasks.withType(KotlinCompile::class.java).firstOrNull()?.kotlinOptions
+          ?: project.tasks.withType(KotlinNativeCompile::class.java).firstOrNull()?.kotlinOptions
```
martinbonnin, line 30: Taking `firstOrNull()` here is technically not correct. There could be distinct compilations with distinct `languageVersion`. This is an edge case, I know but still feels like we're making thing quite complex for little actual gain. 

What about we just take the KGP version and anyone wanting to override that can define the setting manually. They should find out the error quickly.

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/KotlinPluginFacade.kt
```diff
@@ -7,25 +7,38 @@ import com.apollographql.apollo3.gradle.api.kotlinMultiplatformExtension
...
+      val apiVersion = (kotlinOptions?.apiVersion ?: pluginVersion).split(".").map { it.toInt() }
+      val languageVersion = (kotlinOptions?.languageVersion ?: pluginVersion).split(".").map { it.toInt() }
+      // To use Enum.entries we need languageVersion >= 1.9 and apiVersion >= 1.8
+      if ((languageVersion[0] > 1 || languageVersion[1] >= 9) && (apiVersion[0] > 1 || apiVersion[1] >= 8)) {
```
martinbonnin, line 34: There is `KotlinVersion.isAtLeast()`

martinbonnin (approved):

martinbonnin (commented):
A few comments but looks good overall. Would be nice to have the diff in one of the CodegenTest (anyone) so that we can track the change just by looking at the Github history

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -70,45 +71,46 @@ internal class EnumAsEnumBuilder(
...
+        .addProperty(knownEntriesPropertySpec())
         .addFunction(knownValuesFunSpec())
+        .addFunction(safeValueOfFunSpec())
         .build()
   }
 
   private fun IrEnum.knownValuesFunSpec(): FunSpec {
     return FunSpec.builder(Identifier.knownValues)
         .addKdoc("Returns all [%T] known at compile time", selfClassName)
+        .addAnnotation(deprecatedAnnotation("Use knownEntries instead").toBuilder().addMember("replaceWith·=·ReplaceWith(%S)", "this.knownEntries").build())
```
martinbonnin, line 83: I think you don't need the `"this"`?

```suggestion
        .addAnnotation(deprecatedAnnotation("Use knownEntries instead").toBuilder().addMember("replaceWith·=·ReplaceWith(%S)", "knownEntries").build())
```

martinbonnin: Interesting. TIL!

martinbonnin (approved):

Issue #5207, opened 2023-08-24
Add options to not generate data classes


martinbonnin:
### Description

Data classes have a runtime overhead in terms of bytecode size. Especially, `toString()`, `componentN()` and `copy()` are not always used. It would be nice to be able to skip them. 

There are compiler plugins to deal with this like [Poko](https://github.com/drewhamilton/Poko) and/or [DCG](https://github.com/facebookincubator/dataclassgenerate) but we could probably do this without a compiler plugin by generating the methods as needed.

Ideally this is configurable:

```kotlin
apollo {
  service("service") {
    // Generates equals/hashCode
    modelMethods.set(listOf("equalsHashCode"))
    // Also generates toString
    modelMethods.set(listOf("equalsHashCode", "toString"))
    // Only generates copy
    modelMethods.set(listOf("copy"))
    // Generates data classes
    modelMethods.set(listOf("data"))
  }
}
```

BoD:
Available in [4.0.0-beta.1](https://github.com/apollographql/apollo-kotlin/releases/tag/v4.0.0-beta.1).

PR #5212, opened 2023-08-28
Add `generateMethods` options to control which methods are generated on "data" classes

baconz:
See https://github.com/apollographql/apollo-kotlin/issues/5207

This adds support for generated method configuration. Users can choose which methods are generated on "classes that hold data" -- they can either generate them as data classes, or with some combination of:

 * toString
 * copy
 * equals/hashCode


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -426,6 +426,22 @@ interface Service {
    */
   val generateOptionalOperationVariables: Property<Boolean>
 
+  /**
+   * Specifies which methods will be auto generated on operations, models, fragments and input objects.
+   *
+   * Pass a list of any of the following:
+   *
+   * - "equalsHashCode" generates `equals` and `hashCode` methods that will compare generated class properties
```
baconz, line 434: Note that `componentN` is not included as an option here. I think we can add it later if somebody asks for it.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/DataClass.kt
```diff
@@ -62,17 +65,193 @@ fun TypeSpec.Builder.makeDataClassFromProperties(properties: List<PropertySpec>)
...
+  val name = build().name ?: return this
+  fun printPropertiesTemplate() =
+      "$name(" + propertySpecs
+            .excludeInternalProperties()
+            .joinToString(",") { "${it.name}=\$${it.name}" } + ")"
+
+  fun methodCode() =
+      CodeBlock.builder()
+          .beginControlFlow("if (%L == null)", MEMOIZED_TO_STRING_VAR)
+          .add("%L = %P", MEMOIZED_TO_STRING_VAR, printPropertiesTemplate())
```
baconz, line 233: I'm not sure how the performance of this compares to Kotlin's implementation which uses `irConcat`.  https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DataClassMembersGenerator.kt#L207

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/DataClass.kt
```diff
@@ -62,17 +65,193 @@ fun TypeSpec.Builder.makeDataClassFromProperties(properties: List<PropertySpec>)
...
+      .addParameter("other", KotlinSymbols.Any.copy(nullable = true))
+      .returns(KotlinSymbols.Boolean)
+      .addCode(methodCode(build().name))
+      .build())
+}
+
+fun TypeSpec.Builder.withHashCodeImplementation(): TypeSpec.Builder = apply {
+  fun hashPropertyCode(property: PropertySpec) =
+      CodeBlock.builder()
+          .addStatement("${Identifier.__h} *= 31")
```
baconz, line 179: This matches the Kotlin implementation: https://github.com/JetBrains/kotlin/blob/master/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/DataClassMembersGenerator.kt#L235

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/DataClass.kt
```diff
@@ -62,17 +65,193 @@ fun TypeSpec.Builder.makeDataClassFromProperties(properties: List<PropertySpec>)
...
-    .returns(KotlinSymbols.Boolean)
-    .addStatement("return·other != null·&&·other::class·==·this::class")
-    .build()
-
-private fun hashCodeFunSpec() = FunSpec.builder(Identifier.hashCode)
-    .addModifiers(KModifier.OVERRIDE)
-    .returns(KotlinSymbols.Int)
-    .addStatement("return·this::class.hashCode()")
-    .build()
+private const val MEMOIZED_HASH_CODE_VAR: String = "__hashCode"
```
baconz, line 256: I added memoization because the Java implementation had it. I added a check that the properties are immutable for safety.

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt
```diff
@@ -426,6 +426,22 @@ interface Service {
...
+   * - "toString" generates a method that will print a pretty string representing the data in the class
+   * - "copy" (Kotlin only) generates a method that will copy the class with named parameters for
+   * - "dataClass" (Kotlin only and redundant with all other methods) generates the class as a [data class](https://kotlinlang.org/docs/data-classes.html)
```
baconz, line 437: `dataClass` felt a little clearer than `data`, which sounded like a method called `data()` to me. I'm happy to go back to the naming in #5207 if you feel strongly.

libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -101,6 +101,21 @@ public final class com/apollographql/apollo3/compiler/ExpressionAdapterInitializ
...
+	public static final field EQUALS_HASH_CODE Lcom/apollographql/apollo3/compiler/GeneratedMethod;
+	public static final field TO_STRING Lcom/apollographql/apollo3/compiler/GeneratedMethod;
+	public static fun valueOf (Ljava/lang/String;)Lcom/apollographql/apollo3/compiler/GeneratedMethod;
+	public static fun values ()[Lcom/apollographql/apollo3/compiler/GeneratedMethod;
+}
+
+public final class com/apollographql/apollo3/compiler/GeneratedMethod$Companion {
+	public final fun defaultsFor (Lcom/apollographql/apollo3/compiler/TargetLanguage;)Ljava/util/List;
+	public final fun fromName (Ljava/lang/String;)Lcom/apollographql/apollo3/compiler/GeneratedMethod;
+}
```
martinbonnin, line 117: This should probably be internal too.

martinbonnin (commented):
Thanks for looking into this! A few comments but looks good otherwise. 

There are 2 tests failing:
* The `MetadataTest` because they need to pass a value for the `generateMethod` parameter
* The `compiler-hooks` integration test because if fails to add the `GQL` prefix. TBH this is probably a bad design in the `CompilerHooks` APIs but the weird thing is that the default is changed. This Kotlin class shouldn't contain any `equals` method.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/DataClass.kt
```diff
@@ -62,17 +65,193 @@ fun TypeSpec.Builder.makeDataClassFromProperties(properties: List<PropertySpec>)
...
+        .build()
+  }
+
+  check(propertySpecs.all { !it.mutable }) {
+    "Cannot generate hashCode function for class with mutable properties, memoization would fail"
+  }
+
+  if (propertySpecs.isNotEmpty()) {
+    addProperty(
+        PropertySpec.builder(MEMOIZED_HASH_CODE_VAR, KotlinSymbols.Int.copy(nullable = true), KModifier.PRIVATE)
```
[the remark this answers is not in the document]
baconz: I suppose we don't. The Java version was memoized, so I thought I would keep this implementation consistent with that. I've removed the memoization for now.

baconz:
> The compiler-hooks integration test because if fails to add the GQL prefix. TBH this is probably a bad design in the CompilerHooks APIs but the weird thing is that the default is changed. This Kotlin class shouldn't contain any equals method. Edit: it contains equals because the NodeQuery class doesn't have any parameter and can't be a data class. It's the implementation of .equals that changed:

Oof. This is brutal -- I did not know about compiler hooks. I can't see a way to make the hooks work with the generated methods since copy, equals, and toString all need the name of the type when they are generated. If the type name changes, those methods also need to be regenerated, and I can't see a clean way to do that.

I can return to the old (`::class`-based) implementation of `equals` for cases where we are generating a "data class" with no properties -- that would make the test pass, but if a user switches to any of the generated methods with hooks enabled things will break.

We could disallow the use of post-processing hooks when `generateMethods` is anything other than `dataClass`, but that feels unsatisfying.

I was thinking of submitting a ticket to Kotlin Poet and asking if they could provide a way to dynamically get the name of the current type via a `%` variable or something. It feels like that would be a useful feature.

libraries/apollo-compiler/api/apollo-compiler.api
baconz: Unfortunately I think it needs to be public because `DefaultService` lives in a different module

martinbonnin: Aahh yes, good call. We have this `@ApolloInternal` annotation that we use in such cases. Not perfect but still something.

martinbonnin:
I think I found a way by using the resolver all the times (I pushed https://github.com/apollographql/apollo-kotlin/pull/5212/commits/981446e373585b0118eb38d53567300eab5a1e5a to your branch, I hope you don't mind). It's a bit of a kludge but it's working. 

> I can return to the old (::class-based) implementation of equals for cases where we are generating a "data class" with no properties

Do you mind doing this anyway? It creates much more concise generated code. The generated code is not something you read every day but it's a nice bonus if it's palatable.

tests/generated-methods/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+//  }
+//}
+
+
+tasks.withType(KotlinCompile::class.java).configureEach {
+  kotlinOptions {
+    apiVersion = "1.9"
+    languageVersion = "1.9"
+  }
+}
```
martinbonnin, line 37: We run tests using Kotlin 1.6 because this is what we claim compatibility with. `enum` is the only exception because it has code to support 1.9 `entries`. Since this test is not about `entries`, I'd rather have it use the minimum Kotlin version:

```suggestion
apollo {
  service("service") {
    packageName.set("generatedMethods")
    languageVersion.set("1.5")
    generateMethods.set(listOf("toString", "equalsHashCode", "copy"))
  }
}
```

tests/generated-methods/src/main/graphql/schema.graphqls
```diff
@@ -0,0 +1,16 @@
+type Query {
+  bigCat: BigCat!
+}
+
+type BigCat {
+  species: Species!
+  name: String!
+  age: Int!
+  isMean: Boolean!
+  numberZebrasEaten: Int
```
martinbonnin, line 10: 🙀 Wildlife has no mercy... 😅

docs/source/advanced/plugin-configuration.mdx
```diff
@@ -205,6 +205,35 @@ by default).
 
 With the [Android Studio plugin](../testing/android-studio-plugin), you can also go to <kbd>Tools</kbd> | <kbd>Apollo</kbd> | <kbd>Download schema</kbd> which acts as a shortcut to these tasks.
 
+## Generated methods
+By default all Kotlin models, operations, fragments, and input objects are generated as data classes. This means that the compiler will
+auto-generate `toString`, `equals` `hashCode`, `copy` and `componentN` for most generated classes. If you don't think you need all of those
+auto-generated methods, and/or you are worried about the size of the generated code, you can configure the Apollo compiler to generate none
+or a subset of the data class methods. To do this, set `generateMethods` to a list of the methods you need. The available methods are:
```
martinbonnin, line 212: Reminder to myself to quantify this, maybe using Confetti to get a rough idea of the expected improvement.

docs/source/advanced/plugin-configuration.mdx
```diff
@@ -205,6 +205,35 @@ by default).
 
 With the [Android Studio plugin](../testing/android-studio-plugin), you can also go to <kbd>Tools</kbd> | <kbd>Apollo</kbd> | <kbd>Download schema</kbd> which acts as a shortcut to these tasks.
 
+## Generated methods
+By default all Kotlin models, operations, fragments, and input objects are generated as data classes. This means that the compiler will
```
martinbonnin, line 209: "compiler" in these docs often refer to the GraphQL compiler, explicit that it's the Kotlin compiler here:

```suggestion
By default all Kotlin models, operations, fragments, and input objects are generated as data classes. This means that the Kotlin compiler will
```

libraries/apollo-compiler/src/test/graphql/com/example/measurements
```diff
@@ -2,244 +2,11 @@
...
 Test:                                                                                      Total LOC:
-aggregate-all                                                                                  193357
-aggregate-kotlin-responseBased                                                                  61239
-aggregate-kotlin-operationBased                                                                 40535
+aggregate-all                                                                                    1326
+aggregate-kotlin-responseBased                                                                    628
+aggregate-kotlin-operationBased                                                                     0
 aggregate-kotlin-compat                                                                             0
-aggregate-java-operationBased                                                                   91583
+aggregate-java-operationBased                                                                     698
```
martinbonnin, line 9: This is probably the value for a single test. Do you mind regenerating for all tests?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/DataClass.kt
```diff
@@ -62,17 +67,186 @@ fun TypeSpec.Builder.makeDataClassFromProperties(properties: List<PropertySpec>)
...
+      .addCode(methodCode())
+      .build()
+  )
+}
+
+internal fun TypeSpec.Builder.withToStringImplementation(className: ClassName): TypeSpec.Builder {
+  fun printPropertiesTemplate() =
+      "${className.simpleName}(" + propertySpecs
+            .excludeInternalProperties()
+            .joinToString(",") { "${it.name}=\$${it.name}" } + ")"
```
martinbonnin, line 233: Adding a space makes the toString match the Kotlin version
```suggestion
            .joinToString(", ") { "${it.name}=\$${it.name}" } + ")"
```

martinbonnin (approved):
A few final comments but looks good otherwise thanks for diving into this!

martinbonnin:
Thanks! Looking forward to trying it out IRL!

PR #5223, opened 2023-09-05
Add tests for empty objects in last chunk

martinbonnin:


bnjjj:
related to https://github.com/apollographql/router/pull/3748
Thanks for your help

Issue #5230, opened 2023-09-12
`useV3ExceptionHandling` should not throw `ApolloGraphQLException`


baconz:
### Version

4.0.0-alpha.3

### Summary

The behavior in V3 was that parse/network/http exceptions always threw, but a response with `errors` did not throw.

If you set `useV3ExceptionHandling` all exceptions, including `ApolloGraphQLException` will throw, which means that responses that would not have thrown in V3, will throw in V4 with `useV3ExceptionHandling == true`.

### Logs

```
(Your logs here)
```

PR #5231, opened 2023-09-12
useV3ExceptionHandling should only throw when there are no errors populated

baconz:
Closes: #5230


BoD (approved):
Good catch!  Thanks for the fix🙏

martinbonnin (approved):
Thanks!

PR #5234, opened 2023-09-12
Tweak the urlencode algorithm

martinbonnin:
Looks like urls can only contain ascii chars so any non-ascii char makes `NSURL(String)`  return null.

Invert the percent encoding logic. Instead of only encoding the reserved chars, we now percent encode everything but the unreserved ones.


BoD (approved):
💯

PR #5242, opened 2023-09-18
Suppress Kotlin warnings in generated code

martinbonnin:
For the cases where the deprecations are not easily fixable. Many thanks @bryanstern for raising this!

See https://github.com/apollographql/apollo-kotlin/issues/2079


BoD (approved):
👍

Issue #5239, opened 2023-09-14
[IJ/AS plugin] Analytics


bignimbus:
In order to gain insight into our users, if they **opt-in,** in the interest of addressing and anticipating their needs, we would like to collect data **anonymously**. 

Examples of information we are interested in:

- which Apollo Kotlin components are used, and their version
- version of Android (AGP, minSdk, targetSdk, Compose, …) and Kotlin
- which features / options of the Apollo Kotlin Gradle plugin are used
- number of Apollo Kotlin modules in the project
- etc.

The exact list of what is collected will be publicly available when this feature is implemented.
```[tasklist]
### Tasks
- [x] Initialize and save instance id
- [x] Opt-in dialog and setting
- [x] Backend setup
- [x] Networking
- [x] Gather properties
- [x] Gather events
- [x] Scheduling
- [x] Documentation
```

PR #5246, opened 2023-09-19
[IJ plugin] Telemetry: collect properties

BoD:
Related to #5239.
Collect some interesting properties. Some come from the IJ plugin directly (e.g. dependencies), some from the Gradle plugin and passed via the Tooling Model.

For now these are not sent anywhere.


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/ApolloGradleToolingModel.kt
```diff
@@ -22,6 +25,48 @@ interface ApolloGradleToolingModel {
...
+  interface TelemetryData {
+    val gradleVersion: String?
+    val androidMinSdk: Int?
+    val androidTargetSdk: Int?
+    val androidCompileSdk: String?
+    val androidAgpVersion: String?
+    val apolloServiceTelemetryData: List<ServiceTelemetryData>
+    val apolloGenerateSourcesDuringGradleSync: Boolean?
+    val apolloLinkSqlite: Boolean?
+    val apolloUseAntlr: Boolean?
```
martinbonnin, line 37: Late to the party but heads up this one is probably going away soon. Might not even be worth including it. Your call.

BoD: 👍 I'll remove it

PR #5252, opened 2023-09-22
Fix cyclic fragment detection

martinbonnin:
If the same nested fragment was visited several times in different subtrees, it was wrongly detected as a cycle


BoD (approved):
👍

PR #5254, opened 2023-09-23
Logging a single line response body

hbmartin:
**Problem**: A single line response body is not logged since it contains no `\n` and that char is used for terminating logging.
**Solution**: Use okio's null return on `buffer.readUtf8Line()` when the buffer is empty to terminate logging instead.
**Usecase**: AWS AppSync seems to always return single line response.


martinbonnin (approved):
Thanks!

BoD (approved):
👍

PR #5259, opened 2023-09-26
Tweaks for K2

BoD:
This makes a few changes necessary to make compilation with K2 pass - mostly changes for expect/actual syntax that is stricter.

To switch to K2:
- `gradle.properties` and `tests/gradle.properties`: `kotlin.experimental.tryK2=true`
- `CompilerOptions.kt`:
    - `languageVersion.set(KotlinVersion.KOTLIN_2_0)` instead of `KOTLIN_1_9`
    - `freeCompilerArgs.add("-Xsuppress-version-warnings")` as 2.0 is experimental (well maybe that won’t be needed anymore when we switch)
- Note: the Apollo compiler tests will fail because we're using [kotlin-compile-testing](https://github.com/tschuchortdev/kotlin-compile-testing) which currently uses Kotlin 1.8.0 which can't work with code compiled with 2.0. A PR is [here](https://github.com/tschuchortdev/kotlin-compile-testing/pull/388).

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrOperationsBuilder.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrOperationsBuilder.kt
@@ -787,7 +787,7 @@ internal fun GQLDirective.toDeferBooleanExpression(): BooleanExpression<BTerm>? 
 
   val labelArgumentValue = arguments.firstOrNull { it.name == "label" }?.value
   if (labelArgumentValue != null && labelArgumentValue !is GQLStringValue) throw IllegalStateException("Apollo: cannot pass ${labelArgumentValue.toUtf8()} to 'label' argument of 'defer' directive")
-  val label = (labelArgumentValue as GQLStringValue?)?.value
+  val label = labelArgumentValue?.value
   return when (ifArgumentValue) {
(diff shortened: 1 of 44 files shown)


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrOperationsBuilder.kt
```diff
@@ -787,6 +787,7 @@ internal fun GQLDirective.toDeferBooleanExpression(): BooleanExpression<BTerm>?
 
   val labelArgumentValue = arguments.firstOrNull { it.name == "label" }?.value
   if (labelArgumentValue != null && labelArgumentValue !is GQLStringValue) throw IllegalStateException("Apollo: cannot pass ${labelArgumentValue.toUtf8()} to 'label' argument of 'defer' directive")
+  @Suppress("USELESS_CAST", "KotlinRedundantDiagnosticSuppress")
   val label = (labelArgumentValue as GQLStringValue?)?.value
```
BoD, line 791: With K2, the cast is unnecessary and therefore produces a warning. With 1.9 the cast is needed.

martinbonnin (approved):
LGTM 👍 Long term we should remove the expect/actual classes and just expose a common interface and constructor functions. I think that would cover 90% of the cases here and simplify code a bunch.

PR #5263, opened 2023-09-27
Update apollo published

martinbonnin:
We've made a few breaking changes between alpha.1 and alpha.3 (DefaultImpls, Operation.serializeVariables mainly)


BoD (approved):
👍

PR #5281, opened 2023-10-04
Add apollo-execution and apollo-ksp

martinbonnin:
* `apollo-execution` implements the execution parts of the spec using provided resolvers
* `apollo-ksp` processes annotated Kotlin code and generates field resolvers from it

A work-in-progress example can be found [here](https://github.com/joreilly/Confetti/compare/main...no-reflection-server?body=&expand=1&title=switch+to+KSP+to+generate+the+server).

Opening as draft for now.

* [x] disable publishing as it's not ready for showtime yet
* [x] convert our tests to use it


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Adapters.kt
```diff
@@ -117,6 +118,22 @@ class ApolloOptionalAdapter<T>(private val wrappedAdapter: Adapter<T>) : Adapter
   }
 }
 
+@JvmName("-obj")
+fun <T> Adapter<T>.obj() = ObjectAdapter(this)
```
martinbonnin, line 122: Input objects are back to being `Adapter` instead of `CompositeAdapter`. This way, all input types implement  the `Adapter` interface. Also they were not using anything in `AdapterContext`

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt
```diff
@@ -732,7 +735,7 @@ internal class ExecutableValidationScope(
 
     forEach {
       if (it.name == null) {
-        issues.add(OtherValidationIssue(
+        issues.add(AnonymousOperation(
```
martinbonnin, line 738: This is an Apollo error, GraphQL is perfectly fine with it

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/CodegenSchema.kt
```diff
@@ -27,12 +28,14 @@ class CodegenSchema(
     val generateDataBuilders: Boolean,
 )
 
-internal class CodegenType(
+@ApolloInternal
+class CodegenType(
```
martinbonnin, line 32: Side effect of needing to have access to the fqcn of the input an enum types from the KSP processor to validate that they have the expected Kotlin type. Maybe there's a better way.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/CodegenLayout.kt
```diff
@@ -115,7 +119,7 @@ internal abstract class CodegenLayout(
   internal fun fragmentVariablesAdapterName(name: String) = fragmentName(name) + "_VariablesAdapter"
   internal fun fragmentSelectionsName(name: String) = regularIdentifier(name) + "Selections"
 
-  internal fun inputObjectName(name: String) = className(name)
+  fun inputObjectName(name: String) = className(name)
```
martinbonnin, line 122: Same side effect as above. Will revisit to see if we can keep `CodegenLayout` internal

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/file/EnumAsEnumBuilder.kt
```diff
@@ -63,7 +64,9 @@ internal class EnumAsEnumBuilder(
           values.forEach { value ->
             addEnumConstant(layout.enumAsEnumValueName(value.targetName), value.enumConstTypeSpec())
           }
-          addEnumConstant("UNKNOWN__", unknownValueTypeSpec())
+          if (withUnknown) {
+            addEnumConstant("UNKNOWN__", unknownValueTypeSpec())
+          }
```
martinbonnin, line 69: Fun fact, server types cannot reuse the client types because of this. While a client might receive an unknown enum, it is never the case for servers.

PR #5282, opened 2023-10-04
Allo to disable publication of modules

martinbonnin:
Disable publication of 

* apollo-execution
* apollo-ksp


build-logic/src/main/kotlin/api.kt.kt
```diff
@@ -0,0 +1,74 @@
+import org.gradle.api.Project
```
BoD, line 1: The file has `.kt.kt` suffix

build-logic/build.gradle.kts
```diff
@@ -63,14 +64,9 @@ tasks.withType(KotlinJvmCompile::class.java).configureEach {
...
-      id = "apollo.library"
-      implementationClass = "com.apollographql.apollo3.buildlogic.plugin.LibraryConventionPlugin"
-    }
-
-    register("apollo.test") {
-      id = "apollo.test"
-      implementationClass = "com.apollographql.apollo3.buildlogic.plugin.TestConventionPlugin"
+    register("build.logic") {
+      id = "build.logic"
+      implementationClass = "build.logic.Plugin"
```
BoD, line 69: The implementation class doesn't exist or am I missing something? (but how come applying the plugin works?)

BoD (commented):
`build-logic/src/main/kotlin/TestConventionPlugin.kt` is an empty file

build-logic/build.gradle.kts
martinbonnin: Sooooo.... I forgot to add the file to the commit. The fun stuff is that because the plugin is never applied, it still works. Gradle just needs the marker to resolve the .jar but the actual class is never loaded :). This is my new favorite Gradle hack! I'll add a comment.

BoD: Hahaha nice one. Ooooh yeah there was `apply false` indeed!

BoD (approved):

martinbonnin:
I updated Gradle to 8.4 which fails on https://github.com/JetBrains/gradle-intellij-plugin/issues/1469, reverting...

PR #5290, opened 2023-10-09
Fragment variables: fix false warning about unused variables

martinbonnin:
Many thanks to TheKeeperOfPie for catching this


BoD (approved):
👍

PR #5292, opened 2023-10-10
Make CodegenLayout internal again, use CodegenMetadata.resolveSchemaType instead

martinbonnin:
Fix for https://github.com/apollographql/apollo-kotlin/pull/5281/files#r1345525292

Also move the validation part out of ApolloProcessor

### DIFF
--- a/libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
+++ b/libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
@@ -1,0 +1,165 @@
+package com.apollographql.apollo3.ksp
+
+import com.apollographql.apollo3.annotations.ApolloInternal
+import com.apollographql.apollo3.ast.GQLEnumTypeDefinition
+import com.apollographql.apollo3.ast.GQLInputObjectTypeDefinition
+import com.apollographql.apollo3.ast.GQLInterfaceTypeDefinition
+import com.apollographql.apollo3.ast.GQLListType
+import com.apollographql.apollo3.ast.GQLNamedType
+import com.apollographql.apollo3.ast.GQLNonNullType
+import com.apollographql.apollo3.ast.GQLObjectTypeDefinition
+import com.apollographql.apollo3.ast.GQLScalarTypeDefinition
+import com.apollographql.apollo3.ast.GQLType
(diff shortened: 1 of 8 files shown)


libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
```diff
@@ -0,0 +1,165 @@
...
+            val expectedFQDN = codegenMetadata.resolveSchemaType(typeDefinition.name)?.toIrClassName() ?: error("Cannot resolve input ${typeDefinition.name}")
+            if (className != expectedFQDN) {
+              throw IncompatibleType("Input object type '${typeDefinition.name}' is mapped to '${expectedFQDN} but '${className.asString()} was found at ${ksTypeReference.location}")
+            }
+
+            IrInputObjectType(typeDefinition.name)
+          }
+
+          is GQLEnumTypeDefinition -> {
+            val expectedFQDN = codegenMetadata.resolveSchemaType(typeDefinition.name)?.toIrClassName() ?: error("Cannot resolve enum ${typeDefinition.name}")
```
martinbonnin, line 85: This is the actual change. We used to rely on `CodegenLayout` but this can be much better modeled using `CodegenMedata` instead.

libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
```diff
@@ -0,0 +1,165 @@
...
+            IrInputObjectType(typeDefinition.name)
+          }
+
+          is GQLObjectTypeDefinition, is GQLUnionTypeDefinition, is GQLInterfaceTypeDefinition -> {
+            /**
+             * Because of interfaces we do the lookup the other way around. Contrary to scalars, there cannot be multiple objects mapped to the same target
+             */
+            /**
+             * Because of interfaces we do the lookup the other way around. Contrary to scalars, there cannot be multiple objects mapped to the same target
+             */
```
BoD, line 99:
```suggestion
            /**
             * Because of interfaces we do the lookup the other way around. Contrary to scalars, there cannot be multiple objects mapped to the same target
             */
```

BoD (approved):
👍

PR #5298, opened 2023-10-12
[IJ plugin] Normalized cache viewer: UI

BoD:
This is the UI part only for now showing fake data.

https://github.com/apollographql/apollo-kotlin/assets/372852/cec2c4b2-d1b0-4304-8627-af579d14d64a

Missing: a filter for the records list.


martinbonnin (approved):
🤩

PR #5307, opened 2023-10-17
MockServer API cleanup

martinbonnin:
- Reorganise files
- Tweak symbol visibility
- Remove some `@ApolloExperimental`
- etc...

This is very binary breaking but mostly source compatible with the exception of multipart enqueue. 

Comments & details inline.


docs/source/testing/mocking-http-responses.mdx
```diff
@@ -26,7 +26,7 @@ val mockServer = MockServer()
 val apolloClient = ApolloClient.Builder().serverUrl(mockServer.url()).store(store).build()
 
 // Enqueue HTTP responses
-mockServer.enqueue("""{"data": {"random": 42}}""")
+mockServer.enqueueString("""{"data": {"random": 42}}""")
```
martinbonnin, line 29: I always found it confusing that you needed the import to have `enqueue` work (it resolves to the instance function by default). 

`enqueueString` makes it explicit and works better with `enqueueMultipart`

`MockServer.enqueue` is kept as `@DeprecationLevel.Error` for now

libraries/apollo-mockserver/api/apollo-mockserver.api
```diff
@@ -31,33 +45,25 @@ public final class com/apollographql/apollo3/mockserver/MockResponse$Builder {
 	public final fun statusCode (I)Lcom/apollographql/apollo3/mockserver/MockResponse$Builder;
 }
 
-public final class com/apollographql/apollo3/mockserver/MockServer : com/apollographql/apollo3/mockserver/MockServerInterface {
```
martinbonnin, line 34: `MockServerInterface` is no more. There is no more `expect class` because of [this tweet](https://twitter.com/Sellmair/status/1647932805647572994). There is only a `expect fun MockServer()`

libraries/apollo-mockserver/src/appleMain/kotlin/com/apollographql/apollo3/mockserver/AppleMockServer.kt
```diff
@@ -111,12 +111,16 @@ actual class MockServer(
...
+  }
+
   /**
-   * [MockServer] can only stop in between complete request/responses pairs
+   * [AppleMockServer] can only stop in between complete request/responses pairs
    * If stop() is called while we're reading a request, this might wait forever
    * Revisit once okio has native Timeout
    */
-  actual override suspend fun stop() {
+  override fun close() {
```
martinbonnin, line 123: `MockServer` now implements `Closeable` because it's quite convenient. The semantics arent really well defined yet (same as before). Some targets shut down synchronously, others asynchronously. Some shut down existing connections, other are more graceful. 

There is still `closeSynchronously` as `@ApolloExperimental` the time we fine tune this.

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/chunked.kt
```diff
@@ -0,0 +1,48 @@
+package com.apollographql.apollo3.mockserver
```
martinbonnin, line 1: No change in this file, this is just moving symbols around

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServerHandler.kt
```diff
@@ -0,0 +1,10 @@
+package com.apollographql.apollo3.mockserver
```
martinbonnin, line 1: No change in this file, this is just moving symbols around

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/http.kt
```diff
@@ -0,0 +1,91 @@
+package com.apollographql.apollo3.mockserver
```
martinbonnin, line 1: No change in this file, this is just moving symbols around

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/multipart.kt
```diff
@@ -0,0 +1,58 @@
...
+import kotlinx.coroutines.flow.mapNotNull
+import okio.Buffer
+import okio.ByteString
+
+
+private sealed interface Item
+private class BytesItem(val bytes: ByteString): Item
+private class DelayItem(val delayMillis: Long): Item
+
+internal class MultipartBodyImpl(private val boundary: String, private val partsContentType: String) : MultipartBody {
```
martinbonnin, line 16: The multipart API did change, more below.

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -33,11 +51,50 @@ interface MockServerInterface {
...
+      .body(string)
+      .delayMillis(delayMs)
+      .build())
+}
+
+@ApolloExperimental
+interface MultipartBody {
+  fun enqueuePart(bytes: ByteString, isLast: Boolean)
+  fun enqueueDelay(delayMillis: Long)
+}
```
martinbonnin, line 72: This is the new, breaking multipart API.

Instead of `MockServer.enqueueMultipart(List<String>, ...)`, there is now a `MockServer.enqueueMultipart(): MultipartBody` which allows to:

1. enqueue binary parts if we want to
2. handle delays in a more central place
3. hide `ChunkedResponse`: multipart responses are always chunked with this API. If no chunk encoding is required, another body builder can be used.

libraries/apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/EnqueueTest.kt
```diff
@@ -11,6 +12,7 @@ import kotlinx.coroutines.flow.flowOf
 import okio.ByteString.Companion.encodeUtf8
 import kotlin.test.Test
 
+@ApolloInternal
```
martinbonnin, line 15: Not 100% sure why we need those `@ApolloInternal` all around the place given that we Opt-In globally in `CompilerOptions.kt`. 

Without them, the code compiles without warning but the IDE is not happy

tests/defer/src/commonTest/kotlin/test/DeferNormalizedCacheTest.kt
```diff
@@ -75,7 +76,7 @@ class DeferNormalizedCacheTest {
         """{"incremental": [{"data":{"cpu":"386","year":1993,"screen":{"__typename":"Screen","resolution":"640x480"}},"path":["computers",0]}],"hasNext":true}""",
         """{"incremental": [{"data":{"isColor":false},"path":["computers",0,"screen"],"label":"a"}],"hasNext":false}""",
     )
-    mockServer.enqueueMultipart(jsonList)
+    mockServer.enqueueMultipart("application/json").enqueueStrings(jsonList)
```
martinbonnin, line 79: The parts content-type is now required because it felt like it belonged next to the data that is sent. If we wanted to make things even easier we could add a `enqueueJsonMultipart` or so but I don't think it's worth it?

tests/defer/src/commonTest/kotlin/test/DeferTest.kt
```diff
@@ -273,17 +274,15 @@ class DeferTest {
...
-    for ((index, json) in jsonList.withIndex()) {
-      val isLast = index == jsonList.lastIndex
-      chunkedResponse.send(
-          content = json,
-          isFirst = index == 0,
-          isLast = isLast,
-      )
+    jsonList.withIndex().forEach { (index, value) ->
+      multipartBody.enqueueDelay(delayMillis)
+      multipartBody.enqueuePart(value.encodeUtf8(), index == jsonList.lastIndex)
```
martinbonnin, line 279: Small win: we don't need to track `isFirst` anymore

libraries/apollo-mockserver/src/commonTest/kotlin/com/apollographql/apollo3/mockserver/test/EnqueueTest.kt
martinbonnin: Found it! https://github.com/apollographql/apollo-kotlin/pull/5308#discussion_r1361983482

BoD (approved):
👍

PR #5314, opened 2023-10-20
[IJ plugin] Cache viewer: pull file from attached devices

BoD:
https://github.com/apollographql/apollo-kotlin/assets/372852/fa7aff31-b7fa-4746-bb34-b00f0262cb7a


martinbonnin (approved):
Very cool 👏 

Nitpick possible UX improvement: instead of having a succession of popup menus which you can easily lose focus of if you're moving your mouse too fast, make a single popup or maybe modal that flattens everything like so:

```
Pixel_2XL_API_31:dev.johnoreilly.confetti:androidmakers2023.db
Pixel_2XL_API_31:dev.johnoreilly.confetti:kotlinconf2023.db
...
```

Possibly with a filter so you can easily find your db of choice

martinbonnin (approved):

PR #5315, opened 2023-10-20
Fix NSURL tests on recent apple OSes

martinbonnin:
Apple fixed NSURL to automatically percent encode invalid chars. See https://developer.apple.com/documentation/foundation/nsurl


BoD (approved):
👍

PR #5316, opened 2023-10-20
Make more of MockServer common code, only abstract the socket part

martinbonnin:
Instead of writing 3 different MockServer implementations, we now only have to implement 2 (lower level, smaller API surface) SocketServers. The JS one is using node js `net` APIs. The apple/JVM one is using `ktor-network`.


libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/SocketServer.js.kt
```diff
@@ -0,0 +1,95 @@
+package com.apollographql.apollo3.mockserver
+
+import Buffer
+import kotlinx.coroutines.channels.Channel
+import net.AddressInfo
+import net.createServer
+import okio.IOException
+import org.khronos.webgl.Int8Array
+import org.khronos.webgl.Uint8Array
+import org.khronos.webgl.set
```
lppedd, line 10: Probably it's better to ditch `kotlin-nodejs` (unmaintained) and instead rely on `kotlin-wrappers`.  
That means you'll be able to use `js.typedarrays.*` externals instead of WebGL stuff.

lppedd: Also see https://github.com/JetBrains/kotlin-wrappers/pull/2133 in case you go with `kotlin-wrappers`.

martinbonnin: TIL about kotlin-wrappers, this is much better indeed, thanks 🙏  Also allows us to remove the jcenter dependency 🎉

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/http.kt
```diff
@@ -74,18 +85,55 @@ internal fun readRequest(source: BufferedSource): MockRequest? {
...
   )
 }
+
+private fun String.trimEol() = this.trimEnd('\r', '\n')
+
+internal suspend fun writeResponse(response: MockResponse, version: String, writeData: WriteData) {
+  writeData("$version ${response.statusCode}\r\n".encodeToByteArray())
+
+  // We don't support 'Connection: Keep-Alive', so indicate it to the client
+  val headers = response.headers //+ mapOf("Connection" to "close")
```
BoD, line 129: Should this be commented?

BoD (approved):
Very cool! 

And `+482 −783` ❤️

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/http.kt
martinbonnin: It should as we now support keep-alive 🎉 . But those 2 comments are very much inconsistent. Will fix.

PR #5320, opened 2023-10-23
Add FieldNameGenerator for cases where paginationArgs is not enough

BoD:
Adds a `FieldNameGenerator` to allow to programmatically customize the way field names are generated (by default, `CompiledField.nameWithArguments` is used).

This is useful when a field takes an Input describing the pagination arguments, rather than several arguments, which can be ignored thanks to `@fieldPolicy`'s `paginationArgs`.

Implementation: quite a bit of plumbing as the generator needs to be passed around to `Normalizer` and to `ApolloResolver` via its context.


martinbonnin:
Moved this to draft so we're not tempted to merge it until we have more feedback

BoD:
Closing for now, we can revisit this later - ideally we'd like this to be battle-tested a bit.

Issue #5331, opened 2023-10-24
[IJ plugin] Quality of life improvements


eduardb:
### Version

4.0.0-beta.2

### Summary

1. The Pull from device -> Device -> All packages should probably be sorted alphabetically, I get a seemingly random order

1. For packages with more than one DB file, there are two DBs listed per row. I am rocking the new Android Studio UI, maybe related?
![image](https://github.com/apollographql/apollo-kotlin/assets/2426348/b7858dc3-821b-44fd-b04b-e1b61aa4b893)

1. If the Apollo cache DB filename doesn't have an extension, it doesn't show up in the list from "Pull from device". Would it be possible somehow to have it listed among the entries when pulling from the device, so one wouldn't have to manually pull the file, rename and then open it? I imagine looking at all the files in the `databases` directory, if there is a filename_without_extension and filename_without_extension-journal, that is a strong indicator that filename_without_extension is a DB. Or show all files in that directory and filter out known bad ones (`*-journal`, `*-wal`, `*-shm`, etc.) instead.

PR #5333, opened 2023-10-25
[IJ plugin] Cache viewer: "Pull from Device" modal dialog instead of action menu

### LINKED ISSUE #5329: [IJ plugin] Sometimes db files are not in the list from "Pull from device"
### Description

From the device explorer:

image (33)

From the Apollo plugin:

image (32)

`apollo.db` is missing

BoD:
Also, this should fix #5329 and #5331

https://github.com/apollographql/apollo-kotlin/assets/372852/6efdc6bd-7bec-426c-a377-8fd9b0a80d7a


martinbonnin (approved):
I like it but also I'll keep asking for a search filter + flat list in there so that I can use IDEA without the mouse 🤓 😁

BoD:
> I like it but also I'll keep asking for a search filter + flat list in there so that I can use IDEA without the mouse 🤓 😁

Actually the filter is supposed to work automagically (just start typing), but I just noticed -> it's buggy. I'll have a look.

BoD:
Fixed quick search. Also now the running apps are expanded by default - since most of the times in dev mode you'll have only one running.


https://github.com/apollographql/apollo-kotlin/assets/372852/a0164a7b-b795-4e17-a5e2-6fa0511ace1f

PR #5334, opened 2023-10-25
Add WebSocket support to the MockServer

martinbonnin:
Implement reading/writing websocket frames so the only expect/actual things we need is the Socket.

There are still a bunch of undefined behaviours around the semantics of closing a server (see https://youtrack.jetbrains.com/issue/KTOR-6406, also some JS tests are still disabled). But we're slowly getting there


martinbonnin:
I'm dropping the towel on the native failed test. Looks like a bug in `URLSession` to me. At least now [I know I'm not alone](https://developer.apple.com/forums/thread/679446).

BoD (approved):
🎉

PR #5335, opened 2023-10-25
Make test pass if plugin relocation is disabled

martinbonnin:
Locally I have `APOLLO_RELOCATE_JAR=false` in my env to iterate faster but some tests rely on that (for good reasons). It's not the stuff that changes a lot though so for the purpose of local iterations, allow to skip those tests, volkswagen CI style.


BoD (approved):
👍

PR #5347, opened 2023-11-03
Add apollo-debug-server library

BoD:
This adds a library `apollo-debug-server`, meant to be added to apps' debug builds only. Two APIs are then available: 
- `ApolloDebugServer.registerApolloClient(apolloClient: ApolloClient, id: String = "client")` 
- `ApolloDebugServer.unregisterApolloClient(apolloClient: ApolloClient)`.

When an ApolloClient is registered, a minimalist GraphQL server runs at a unix socket named with the scheme `apollo_debug_<app's package name>` (package name is retrieved through an [`InitializationProvider`](https://developer.android.com/topic/libraries/app-startup)).

Interested tools can then:
- list these named sockets with `adb cat /proc/net/unix` and look for the ones starting with `apollo_debug_`
- forward them to a local port with `adb forward tcp:<some port> localabstract:<the socket name>`, e.g. `adb forward tcp:12200 localabstract:apollo_debug_com.example.myapp`
- execute GraphQL queries on `http://localhost:<the chosen port>`, to get the list of normalized caches, and retrieve them

This mechanism is modeled on what the [Chrome Dev Tools do](https://github.com/GoogleChrome/lighthouse/blob/main/docs/readme.md) to inspect running WebViews, which is also used by [Stetho](https://github.com/facebookarchive/stetho/blob/main/stetho/src/main/java/com/facebook/stetho/Stetho.java#L448).

The module is multiplatform but only the Android target actually does anything (the other targets have no-op implementations) - allowing clients to call `registerApolloClient` in common code.

For GraphQL, this uses the `apollo-execution` incubating library which is currently not published, that's why it is shaded into the `aar`.

### DIFF
--- a/libraries/apollo-debug-server/build.gradle.kts
+++ b/libraries/apollo-debug-server/build.gradle.kts
@@ -1,0 +1,94 @@
+import com.android.build.gradle.tasks.BundleAar
+import org.jetbrains.kotlin.gradle.tasks.Kotlin2JsCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinNativeCompile
+
+plugins {
+  id("org.jetbrains.kotlin.multiplatform")
+  id("com.android.library")
+  alias(libs.plugins.apollo.published)
+  id("com.google.devtools.ksp")
+}
+
+apolloLibrary(
+    javaModuleName = "com.apollographql.apollo3.debugserver",
+    withLinux = false,
+)
+
+kotlin {
+  sourceSets {
+    findByName("commonMain")?.apply {
+      kotlin.srcDir("build/generated/ksp/metadata/commonMain/kotlin")
+
+      dependencies {
+        implementation(project(":apollo-normalized-cache"))
+
+        // apollo-execution is not published: we bundle it into the aar artifact
+        compileOnly(project(":apollo-execution"))
+      }
+    }
+
+    findByName("androidMain")?.apply {
+      dependencies {
+        implementation(libs.androidx.startup.runtime)
+      }
+    }
+  }
+}
+
+val shadow = configurations.create("shadow") {
+  isCanBeConsumed = false
+  isCanBeResolved = true
+}
+
+dependencies {
+  add("kspCommonMainMetadata", project(":apollo-ksp"))
+  add("kspCommonMainMetadata", apollo.apolloKspProcessor(file("src/androidMain/resources/schema.graphqls"), "apolloDebugServer", "com.apollographql.apollo3.debugserver.internal.graphql"))
+  add(shadow.name, project(":apollo-execution")) {
+    isTransitive = false
+  }
+}
+
+configurations.getByName("compileOnly").extendsFrom(shadow)
+
+android {
+  compileSdk = libs.versions.android.sdkversion.compile.get().toInt()
+  namespace = "com.apollographql.apollo3.debugserver"
+
+  defaultConfig {
+    minSdk = libs.versions.android.sdkversion.min.get().toInt()
+  }
+}
+
+// KMP ksp configuration inspired by https://medium.com/@actiwerks/setting-up-kotlin-multiplatform-with-ksp-7f598b1681bf
+tasks.withType<KotlinCompile>().configureEach {
+  dependsOn("kspCommonMainKotlinMetadata")
+}
+
+tasks.withType<Kotlin2JsCompile>().configureEach {
+  dependsOn("kspCommonMainKotlinMetadata")
+}
+
+tasks.withType<KotlinNativeCompile>().configureEach {
+  dependsOn("kspCommonMainKotlinMetadata")
+}
+
+tasks.configureEach {
+  if (name.endsWith("sourcesJar", ignoreCase = true)) {
+    dependsOn("kspCommonMainKotlinMetadata")
+  }
+}
+
+// apollo-execution is not published: we bundle it into the aar artifact
+val jarApolloExecution = tasks.register<Jar>("jarApolloExecution") {
+  archiveBaseName.set("apollo-execution")
+  from(provider {
+    shadow.files.map { zipTree(it) }
+  })
+}
+
+tasks.withType<BundleAar>().configureEach {
+  from(jarApolloExecution) {
+    into("libs")
+  }
+}
--- a/libraries/apollo-debug-server/src/androidMain/kotlin/com/apollographql/apollo3/debugserver/internal/server/Server.android.kt
+++ b/libraries/apollo-debug-server/src/androidMain/kotlin/com/apollographql/apollo3/debugserver/internal/server/Server.android.kt
@@ -1,0 +1,105 @@
+package com.apollographql.apollo3.debugserver.internal.server
+
+import android.net.LocalServerSocket
+import android.net.LocalSocket
+import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.debugserver.internal.graphql.GraphQL
+import com.apollographql.apollo3.debugserver.internal.initializer.ApolloDebugServerInitializer
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import java.io.BufferedReader
+import java.io.PrintStream
+import java.util.concurrent.Executors
+
+internal actual fun createServer(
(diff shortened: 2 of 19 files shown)


libraries/apollo-debug-server/src/androidMain/kotlin/com/apollographql/apollo3/debugserver/internal/server/Server.android.kt
```diff
@@ -0,0 +1,105 @@
...
+      while (true) {
+        val clientSocket = try {
+          localServerSocket.accept()
+        } catch (_: Exception) {
+          // Server socket has been closed (stop() was called)
+          break
+        }
+        launch { handleClient(clientSocket) }
+      }
+    }
```
martinbonnin, line 49: Does this need to be a coroutine? Is there any other suspending call than `ApolloStore.dump()`? This one really shouldn't be supend as it's synchronous. It's part of the APIs I want to change for v4. So all in all, unless there are other, I would just wrap the dump in `runBlocking {apolloStore.dump()}`

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
```diff
@@ -0,0 +1,219 @@
...
+    val graphQLRequestResult = Buffer().writeUtf8(jsonBody).parsePostGraphQLRequest()
+    if (graphQLRequestResult is GraphQLRequestError) {
+      return graphQLRequestResult.message
+    }
+    graphQLRequestResult as GraphQLRequest
+    val dumps = apolloClients.mapValues { (apolloClient, _) ->
+      apolloClient.apolloStore.dump()
+    }
+    val apolloDebugContext = ApolloDebugContext(apolloClients, dumps)
+    val graphQlResponse = executableSchema.execute(graphQLRequestResult, apolloDebugContext)
```
martinbonnin, line 56: Note to self: would be nice being able to pass arguments to the `Query()` root type instead of relying on executionContext

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
```diff
@@ -0,0 +1,219 @@
...
+  ) : ExecutionContext.Element {
+    override val key: ExecutionContext.Key<ApolloDebugContext> = Key
+
+    companion object Key : ExecutionContext.Key<ApolloDebugContext>
+  }
+}
+
+@ApolloObject
+internal class Query {
+  fun version() = 1
```
martinbonnin, line 74: GraphQL shouldn't need versioning? Or should it?🙃

libraries/apollo-debug-server/src/androidMain/resources/schema.graphqls
```diff
@@ -0,0 +1,32 @@
...
+    displayName: String!
+    recordCount: Int!
+}
+
+type NormalizedCache {
+    id: ID!
+    displayName: String!
+    clientDisplayName: String!
+    records: [KeyedRecord!]!
+}
```
martinbonnin, line 24: Let's make this GraphQL-y and merge `NormalizedCache` and `NormalizedCacheInfo`. 
I think that can remove the need for duplicating `displayName` and remove `clientDisplayName` completelly?

Listing clients:
```graphql
{
  clients {
    displayName
    normalizedCache {
      displayName
    }
  }
}
```

Getting records:

```graphql
{
  client(id) {
    normalizedCache {
      records
    }
  }
}
```

Maybe `NormalizedCache` doesn't even needs its own id?

libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+val shadow = configurations.create("shadow") {
+  isCanBeConsumed = false
+  isCanBeResolved = true
+}
+
+dependencies {
+  add("kspCommonMainMetadata", project(":apollo-ksp"))
+  add("kspCommonMainMetadata", apollo.apolloKspProcessor(file("src/androidMain/resources/schema.graphqls"), "apolloDebugServer", "com.apollographql.apollo3.debugserver.internal.graphql"))
```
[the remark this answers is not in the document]
BoD: That's how I understand it 👍

libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+  id("org.jetbrains.kotlin.multiplatform")
+  id("com.android.library")
+  alias(libs.plugins.apollo.published)
+  id("com.google.devtools.ksp")
+}
+
+apolloLibrary(
+    javaModuleName = "com.apollographql.apollo3.debugserver",
+    withLinux = false,
+)
```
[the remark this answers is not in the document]
BoD: I was thinking, if they're enabled it's a bit nicer for KMP apps, they can add the common dependency and not deal with expect/actual? Maybe not that useful actually.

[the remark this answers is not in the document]
martinbonnin: I think it gives the wrong impression that JS/native are supported while in fact not. Better fail early and not expose the symbols at all.

libraries/apollo-debug-server/src/androidMain/kotlin/com/apollographql/apollo3/debugserver/internal/server/Server.android.kt
BoD: Well, I just got reminded by my IDE that `runBlocking` doesn't exist in `commonMain` 😅

martinbonnin: IIRC if you remove the JS target, it will "work" again. By the time we add JS again, we should remove the suspend from `ApolloStore.dump`

BoD: >if you remove the JS target, it will "work" again

wow, indeed 🤯

BoD: That being said (⚠️ can of worms), would it be cool for `Resolver.resolve()` to be suspend, and allow `@ApolloObject` funs to be suspend?

martinbonnin: "probably" but this goes into data loader land and parallel execution questions. 
If a field is suspend, is it sharing its thread with other fields? If not I think it gives the wrong impression because you could as well call `runBlocking {}` inside the field logic (except for JS...). TLDR; might be useful but I don't really want to open that door at that point.

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/ApolloDebugServer.kt
```diff
@@ -0,0 +1,34 @@
...
+    if (apolloClients.isEmpty()) {
+      server?.stop()
+    } else {
+      if (server == null) {
+        server = createServer(apolloClients).apply {
+          start()
+        }
+      }
+    }
+  }
```
martinbonnin, line 33: This doesn't work if you have 2 ApolloClients:

```kotlin
ApolloDebugServer.registerApolloClient(client1, "client1")
// server is started
ApolloDebugServer.registerApolloClient(client2, "client2")
// nothing happens
```

BoD: Woops sorry I missed that comment 🙄 

But I think it works as expected? The server is already running, so no need to start it again. But it has a reference to the list of clients, so it will see it.

martinbonnin: Oooh I missed that the list was mutable. That's not super thread safe. Probably not going to be much a problem but if it ever becomes one, it'll be hell to investigate. Make the list an `AtomicReference<>` ?

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
martinbonnin: Done in https://github.com/apollographql/apollo-kotlin/pull/5352

PR #5348, opened 2023-11-03
[IJ plugin] Use apollo-debug-server to retrieve normalized caches

BoD:
https://github.com/apollographql/apollo-kotlin/assets/372852/a0e46b2b-cfa0-4602-bea6-131c8bff5be9


intellij-plugin/build.gradle.kts
```diff
@@ -219,6 +220,25 @@ dependencies {
...
 fun isSnapshotBuild() = System.getenv("COM_APOLLOGRAPHQL_IJ_PLUGIN_SNAPSHOT").toBoolean()
+
+apollo {
+  service("apolloDebug") {
+    packageName.set("com.apollographql.apollo3.debug")
+    schemaFile.set(file("../libraries/apollo-debug-server/src/androidMain/resources/schema.graphqls"))
+    introspection {
+      endpointUrl.set("http://localhost:12200/")
+      schemaFile.set(file("../libraries/apollo-debug-server/src/androidMain/resources/schema.graphqls"))
+    }
```
martinbonnin, line 235: Do we need introspection for this given that the schema is checked in source control?

martinbonnin (approved):
🚀

intellij-plugin/build.gradle.kts
BoD: Not really, but it's handy because it allows the IJ plugin to be configured with a URL, then we can execute queries in-IDE.

martinbonnin: 👀 oohoho I see 👍 . The `introspection {}` block is a bit misleading here, maybe we could rework that later... Anyways, cool stuff 👍

PR #5351, opened 2023-11-03
[apollo-ksp] Initial support for interfaces

martinbonnin:
Initial support for interfaces. The processor will look for descendant of Kotlin interfaces and make sure they are corresponds to the possible types of a given field:

```kotlin
interface Event

@ApolloObject
class MessageEvent(
    val channelId: String,
    val message: String
): Event

@ApolloObject
class AckEvent: Event {
  val ack: String = "ack"
}
```

### DIFF
--- a/libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
+++ b/libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
@@ -24,6 +24,7 @@ import com.apollographql.apollo3.compiler.ir.IrOptionalType
 import com.apollographql.apollo3.compiler.ir.IrScalarType
 import com.apollographql.apollo3.compiler.ir.IrType
 import com.apollographql.apollo3.compiler.resolveSchemaType
+import com.google.devtools.ksp.processing.KSPLogger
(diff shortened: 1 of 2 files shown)


libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
```diff
@@ -34,8 +35,25 @@ internal class ValidationScope(
...
+    private val logger: KSPLogger,
 ) {
+  private val possibleTypes = mutableMapOf<IrClassName, MutableSet<String>>()
+
+  init {
+    objectMapping.forEach { entry ->
+      val superTypes = entry.value.classDeclaration.superTypes.map { it.resolve().declaration.acClassName() }
+
+      superTypes.forEach {
+        logger.warn("${entry.value.className} extends $it")
```
BoD, line 48: Should this be a warning?

BoD (approved):
👍

libraries/apollo-ksp/src/main/kotlin/com/apollographql/apollo3/ksp/ksp-validation.kt
martinbonnin: 🤦 shouldn't be there

PR #5352, opened 2023-11-03
[apollo-execution] Allow to pass arguments to the root types

martinbonnin:
Add a `${ServiceName}ExecutableSchemaBuilder()` constructor function that sets up a pre-configured `ExecutableSchema.Builder`. 

This function takes a `() -> QueryType`, etc.. for root types so it's easier to pass arguments to the root instances.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -23,6 +23,9 @@
     <Properties>
       <option name="KEEP_BLANK_LINES" value="true" />
     </Properties>
+    <ScalaCodeStyleSettings>
+      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
+    </ScalaCodeStyleSettings>
     <XML>
       <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
     </XML>
--- a/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/ApolloDebugServer.kt
+++ b/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/ApolloDebugServer.kt
@@ -3,26 +3,34 @@ package com.apollographql.apollo3.debugserver
 import com.apollographql.apollo3.ApolloClient
 import com.apollographql.apollo3.debugserver.internal.server.Server
 import com.apollographql.apollo3.debugserver.internal.server.createServer
+import okio.withLock
+import java.util.concurrent.locks.ReentrantLock
 
 object ApolloDebugServer {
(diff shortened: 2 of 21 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -23,6 +23,9 @@
     <Properties>
       <option name="KEEP_BLANK_LINES" value="true" />
     </Properties>
+    <ScalaCodeStyleSettings>
+      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
+    </ScalaCodeStyleSettings>
```
martinbonnin, line 28: Not sure why this started happening. This is for scala so I do not expect too much impact. If this ends up moving too much, I'll dig into it.

BoD (approved):
Very cool! 🎉

martinbonnin (commented):
Can't approve my own PR but LGTM!

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/ApolloDebugServer.kt
```diff
@@ -3,26 +3,34 @@ package com.apollographql.apollo3.debugserver
...
   private val apolloClients = mutableMapOf<ApolloClient, String>()
   private var server: Server? = null
+  private val lock = ReentrantLock()
 
   fun registerApolloClient(apolloClient: ApolloClient, id: String = "client") {
-    if (apolloClients.containsKey(apolloClient)) error("Client '$apolloClient' already registered")
-    if (apolloClients.containsValue(id)) error("Name '$id' already registered")
-    apolloClients[apolloClient] = id
-    startOrStopServer()
+    lock.withLock {
```
martinbonnin, line 15: I don't think that works? This locks the writer but readers from the HTTP request thread won't use the lock and might iterate the list while it is modified

BoD: Ah you're right! I liked the `AtomicReference` idea but didn't like that it's JVM only. But maybe that's ok.

martinbonnin: [AtomicFu](https://github.com/Kotlin/kotlinx-atomicfu) has KMP versions.

BoD: Ah great! 🎉 
I think I'll use JVM `AtomicReference` to avoid adding the AtomicFu dependency, but with the knowledge that we can use it to support more targets.

PR #5353, opened 2023-11-03
Add apollo-debug-server

BoD:
This adds a library `apollo-debug-server`, meant to be added to apps' debug builds only. Two APIs are then available: 
- `ApolloDebugServer.registerApolloClient(apolloClient: ApolloClient, id: String = "client")` 
- `ApolloDebugServer.unregisterApolloClient(apolloClient: ApolloClient)`.

When an ApolloClient is registered, a minimalist GraphQL server runs at a unix socket named with the scheme `apollo_debug_<app's package name>` (package name is retrieved through an [`InitializationProvider`](https://developer.android.com/topic/libraries/app-startup)).

Interested tools can then:
- list these named sockets with `adb cat /proc/net/unix` and look for the ones starting with `apollo_debug_`
- forward them to a local port with `adb forward tcp:<some port> localabstract:<the socket name>`, e.g. `adb forward tcp:12200 localabstract:apollo_debug_com.example.myapp`
- execute GraphQL queries on `http://localhost:<the chosen port>`, to get the list of normalized caches, and retrieve them

This mechanism is modeled on what the [Chrome Dev Tools do](https://github.com/GoogleChrome/lighthouse/blob/main/docs/readme.md) to inspect running WebViews, which is also used by [Stetho](https://github.com/facebookarchive/stetho/blob/main/stetho/src/main/java/com/facebook/stetho/Stetho.java#L448).

The module is multiplatform but only the Android target actually does anything (the other targets have no-op implementations) - allowing clients to call `registerApolloClient` in common code.

For GraphQL, this uses the `apollo-execution` incubating library which is currently not published, that's why it is shaded into the `aar`.

(follow-up of #5347 which was closed by mistake)

### DIFF
--- a/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
+++ b/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
@@ -114,6 +114,10 @@ internal class GraphQLApolloClient(
   fun displayName() = displayName
 
   fun normalizedCaches(): List<NormalizedCache> = normalizedCaches
+
+  fun normalizedCache(id: String): NormalizedCache? {
+    return normalizedCaches.firstOrNull { it.id() == id }
+  }
 }
 
 @ApolloObject
--- a/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
+++ b/libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
@@ -1,0 +1,219 @@
+package com.apollographql.apollo3.debugserver.internal.graphql
+
+import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.annotations.ApolloAdapter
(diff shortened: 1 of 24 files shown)


libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
```diff
@@ -0,0 +1,191 @@
...
+import com.apollographql.apollo3.debugserver.internal.graphql.execution.ApolloDebugServerResolver
+import com.apollographql.apollo3.execution.ExecutableSchema
+import com.apollographql.apollo3.execution.GraphQLRequest
+import com.apollographql.apollo3.execution.GraphQLRequestError
+import com.apollographql.apollo3.execution.parsePostGraphQLRequest
+import kotlinx.coroutines.runBlocking
+import okio.Buffer
+import kotlin.reflect.KClass
+
+internal expect fun getExecutableSchema(): String
```
martinbonnin, line 28: Nit: rename to `getSchema()` as this one is not executable (it doesn't contain resolvers)

martinbonnin:
FWIW, I made a tentative simplification [there](https://github.com/apollographql/apollo-kotlin/commit/968dfe1512c118ef3e23e36d8fda53a8ca66dce4) (wildlly untested). Feel free to integrate or discard.

martinbonnin (approved):
🚀

PR #5371, opened 2023-11-15
Implement 2-step introspection

BoD:
1. execute a pre-introspection query used to get the server capabilities
2. execute the base introspection query augmented according to the capabilities

The Gradle plugin is no longer used for that part, but still present and used by the Apollo Platform API related operations (registry, persisted queries, field insights, telemetry).


libraries/apollo-tooling/api/apollo-tooling.api
```diff
@@ -37,15 +37,6 @@ public final class com/apollographql/apollo3/tooling/PublishOperationsSuccess :
 	public final fun getUpdated ()I
 }
 
-public final class com/apollographql/apollo3/tooling/SchemaDownloader$SpecVersion : java/lang/Enum {
```
BoD, line 40: Technically breaking! I think it was public by mistake.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/meta_introspection_reader.kt
```diff
@@ -0,0 +1,156 @@
...
+      if (inputValue.fields.orEmpty().any { it.name == "deprecationReason" }) {
+        add(InputValueDeprecatedReason)
+      }
+    }
+  }
+}
+
+private val json = Json {
+  // be robust to errors: [] keys
+  ignoreUnknownKeys = true
```
martinbonnin, line 118: What happens if a field errors (very unlikely to happen for an introspection query but 🤷 )?  Should we output an error instead?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/meta_introspection_reader.kt
```diff
@@ -0,0 +1,156 @@
...
+@Serializable
+private class MField(
+    val name: String,
+    val args: List<MInputValue>,
+)
+
+@Serializable
+private class MInputValue(
+    val name: String,
+)
```
martinbonnin, line 156: For the meta introspection query, you could use Apollo and generate all of that automatically.

libraries/apollo-tooling/src/main/resources/base-introspection.graphql
```diff
@@ -1,5 +1,10 @@
+# Base introspection query that will be augmented according to the server capabilities.
+# The capabilities are retrieved by executing the meta-introspection query next to this file.
+# The fields that can be added or modified are shown as comments.
```
martinbonnin, line 3: Can you add a comment ~what version of the spec this corresponds to?

libraries/apollo-tooling/src/main/resources/meta-introspection.graphql
```diff
@@ -0,0 +1,15 @@
...
+    types {
+      name
+      fields {
+        name
+        args {
+          name
+        }
+      }
+    }
+  }
```
martinbonnin, line 14: Download only the required types?

```graphql
query MetaIntrospectionQuery {
  __type(name: "__Schema") { ...typeDetails }
  __type(name: "__Type") { ...typeDetails }
  __type(name: "__Directive") { ...typeDetails }
  __type(name: "__Field") { ...typeDetails }
  __type(name: "__InputValue") { ...typeDetails }
}
```

It's not so much about the performance improvements but the fact that it conveys better what we're going to do with this data?

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/SchemaDownloaderTests.kt
```diff
@@ -9,21 +9,893 @@ import org.junit.Assert.assertTrue
 import org.junit.Test
 import java.io.File
 
-class SchemaDownloaderTests {
-  private lateinit var mockServer: MockServer
-  private lateinit var tempFile: File
+private const val metaSchemaJune2018Response = """
```
martinbonnin, line 12: Nit: can this be moved to separate test fixture files so that we have less scrolling to do?

libraries/apollo-tooling/src/test/kotlin/com/apollographql/apollo3/tooling/SchemaDownloaderTests.kt
```diff
@@ -138,23 +1023,53 @@ class SchemaDownloaderTests {
...
-    assertEquals(schemaString1, tempFile.readText())
+    assertEquals(schemaResponse, tempFile.readText())
   }
 
   @Test
-  fun `schema is downloaded correctly when server doesn't support deprecated input fields and arguments nor isRepeatable on directives`() = runTest(before = { setUp() }, after = { tearDown() }) {
-    mockServer.enqueueString(statusCode = 400)
-    mockServer.enqueueString(statusCode = 400)
-    mockServer.enqueueString(schemaString1)
+  fun `schema is downloaded correctly when server supports Draft spec as of 2023-11-15`() = runTest(before = { setUp() }, after = { tearDown() }) {
```
martinbonnin, line 1070: `@oneOf` is not in the draft yet so it's more like `server supports @oneOf`

libraries/apollo-tooling/api/apollo-tooling.api
martinbonnin: I think there is some value in allowing the user to get an introspection query for a subset of supported features. Spec version dates is probably the bad abstraction (`@oneOf` being an example) but allowing to pass options like [graphql-js is doing ](https://github.com/mjmahone/graphql-js/blob/8724e22f9e89224dd2669d6fc0c4b7fe34c2c8dd/docs-old/APIReference-Utilities.md#L111) would be nice.

martinbonnin (commented):
Looks great!  👍 

One last comment about bikeshedding the name 😃 . I have mixed feelings about "meta introspection query" as "meta" is quite the loaded term. The only suggestion I have in mind is "step1 introspection query" which isn't great either so I'm fine keeping this "meta" for now. Up to you.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/meta_introspection_reader.kt
BoD: Totally, I was in a "don't use Apollo" mindset, but it makes sense to use it here actually - the code will be simpler 🎉

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/introspection/meta_introspection_reader.kt
BoD: If you get an error here, something is very wrong, so it would make sense to fail early I think.

libraries/apollo-tooling/src/main/resources/meta-introspection.graphql
BoD: Didn't realize it was allowed! Will update 👍

libraries/apollo-tooling/api/apollo-tooling.api
BoD: In 75b0843611bb789b6cac6bfd4123ee995351e3d1 I made `IntrospectionCapability` public (and `@ApolloExperimental`) and added `SchemaDownloader.getIntrospectionQuery(capabilities: Set<IntrospectionCapability>): String`

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/IntrospectionCapability.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * `__InputValue.isDeprecated`, introduced in [Draft](https://spec.graphql.org/draft/)
+   */
+  InputValueIsDeprecated,
+
+  /**
+   * `__InputValue.deprecationReason`, introduced in [Draft](https://spec.graphql.org/draft/)
+   */
+  InputValueDeprecatedReason,
```
martinbonnin, line 55: I think these are all the same? Does it make sense to include one but not the others?

martinbonnin: More generally, the naming should be about "features" and not introspection names. Maybe `SpecifiedBy`, `OneOf`, `DeprecatedInputValues`, `RepeatableDirectives`, ...?

libraries/apollo-gradle-plugin/src/test-java11/kotlin/com/apollographql/apollo3/gradle/test/DownloadSchemaTests.kt
```diff
@@ -145,13 +173,16 @@ class DownloadSchemaTests {
   fun `download schema is never up-to-date`() {
 
     withSimpleProject(apolloConfiguration = apolloConfiguration) { dir ->
-      val mockResponse = MockResponse().setBody(schemaString1)
-      mockServer.enqueue(mockResponse)
+      val metaSchemaMockResponse = MockResponse().setBody(preIntrospectionResponse)
+      val schemaMockResponse = MockResponse().setBody(schemaString1)
+      mockServer.enqueue(metaSchemaMockResponse)
+      mockServer.enqueue(schemaMockResponse)
```
martinbonnin, line 179: `metaSchemaMockResponse` -> `preIntrospectionResponse`

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/IntrospectionCapability.kt
BoD: In 830615e670e4c00db468d70facae9c689028cfa4 I've renamed the enum to` GraphQLFeature`, and made it less granular.

PR #5378, opened 2023-11-17
Do not warn in generated code

martinbonnin:


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/helpers/KDoc.kt
```diff
@@ -135,3 +127,33 @@ internal fun <T: Annotatable.Builder<*>> T.maybeAddOptIn(
   val annotation = resolver.resolveRequiresOptInAnnotation() ?: return@applyIf
   addAnnotation(requiresOptInAnnotation(annotation))
 }
+
+/**
+ * Add suppressions for generated code.
+ * This is code the user has no control over and it should not generate warnings
+ */
+internal fun <T: Annotatable.Builder<*>> T.addSuppressions(
```
BoD, line 135: Contribution idea to Kotlin Poet!

BoD (approved):
👍

PR #5383, opened 2023-11-20
Expose ExecutionContext to HttpEngine and add OkHttp helpers

martinbonnin:
Makes it easier to create custom `HttpEngine` implementations and tag OkHttp requests. Makes it possible to do things like this: 

```kotlin
internal class MyHttpEngine: HttpEngine {
  private val okHttpClient = OkHttpClient()

  override suspend fun execute(request: HttpRequest): HttpResponse {
    val taggedRequest = request
        .toOkHttpRequest()
        .newBuilder()
        .tag(MyExecutionContext::class.java, request.executionContext[MyExecutionContext])
        .build()
    return okHttpClient.execute(taggedRequest).toApolloHttpResponse()
  }

  override fun dispose() {

  }
}
```


BoD (approved):
👍

PR #5384, opened 2023-11-20
Take default values into account when computing field keys

martinbonnin:
See the tests in [CacheArgumentTest](https://github.com/apollographql/apollo-kotlin/blob/3adf2835fcb2710bec0315266f26a9aa3673e35f/tests/cache-variables-arguments/src/test/kotlin/test/CacheArgumentTest.kt#L23). Allows to take defaultValues in cache keys into account for queries like so:

```graphql
query GetFoo($arg: Int! = 42) {
  foo(arg: $arg)
}
```

Also take this opportunity to fix `GQLIntValue` and `GQLFLoatValue` to make them support arbitrary precision numbers.

While these 2 things are not directly linked, they are related as `defaultValues` were the only places where we would use the `GQLIntValue` and `GQLFLoatValue` values.

Closes https://github.com/apollographql/apollo-kotlin/issues/5186

I'm not 100% certain yet about the exhaustivity and correctness of the coercion in different cases (see also https://github.com/graphql/graphql-spec/pull/793) but this should provide a good base to build on.


libraries/apollo-api/api/apollo-api.api
```diff
@@ -270,16 +270,17 @@ public abstract interface class com/apollographql/apollo3/api/BuilderScope {
 }
 
 public final class com/apollographql/apollo3/api/CompiledArgument {
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Object;ZZLkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Ljava/lang/String;Lcom/apollographql/apollo3/api/Optional;ZZLkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getName ()Ljava/lang/String;
-	public final fun getValue ()Ljava/lang/Object;
+	public final fun getValue ()Lcom/apollographql/apollo3/api/Optional;
```
martinbonnin, line 275: Breaking change.

libraries/apollo-api/api/apollo-api.api
```diff
@@ -305,7 +306,7 @@ public final class com/apollographql/apollo3/api/CompiledField : com/apollograph
 	public final fun getType ()Lcom/apollographql/apollo3/api/CompiledType;
 	public final fun nameWithArguments (Lcom/apollographql/apollo3/api/Executable$Variables;)Ljava/lang/String;
 	public final fun newBuilder ()Lcom/apollographql/apollo3/api/CompiledField$Builder;
-	public final fun resolveArgument (Ljava/lang/String;Lcom/apollographql/apollo3/api/Executable$Variables;)Ljava/lang/Object;
+	public final fun resolveArgument (Ljava/lang/String;Lcom/apollographql/apollo3/api/Executable$Variables;)Lcom/apollographql/apollo3/api/Optional;
```
martinbonnin, line 309: Same here. An argument with a variable value is possibly absent if that variable is absent as well and unless we change the signature, there's no way to account for that. We could go through deprecation cycles but given this is a low level API, I'm hoping this is fine.

libraries/apollo-api/api/apollo-api.api
```diff
@@ -351,7 +352,6 @@ public final class com/apollographql/apollo3/api/CompiledGraphQL {
 	public static final fun -notNull (Lcom/apollographql/apollo3/api/CompiledType;)Lcom/apollographql/apollo3/api/CompiledNotNullType;
 	public static final fun isComposite (Lcom/apollographql/apollo3/api/CompiledNamedType;)Z
 	public static final fun keyFields (Lcom/apollographql/apollo3/api/CompiledNamedType;)Ljava/util/List;
-	public static final fun resolveVariables (Ljava/lang/Object;Lcom/apollographql/apollo3/api/Executable$Variables;)Ljava/lang/Object;
```
martinbonnin, line 354: Another breaking change. This one was an accidental public I'd say

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
```diff
@@ -423,3 +411,22 @@ constructor(
...
+ * - Double
+ * - Long
+ * - JsonNumber
+ * - null
+ * - Map<String, ApolloJsonElement> where values are any of these values recursively
+ * - List<ApolloJsonElement> where values are any of these values recursively
+ *
+ * Anything else is undefined
+ */
+typealias ApolloJsonElement = Any?
```
martinbonnin, line 432: Interesting that this doesn't show in apiDump. I guess it's source-breaking only. 

In all cases, I wish I've done that earlier to be able to point in KDoc to what `Any?` can contain. Naming is very open to suggestions.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/adapter/VariablesAdapter.kt
```diff
@@ -60,25 +49,25 @@ private fun List<NamedType>.writeToResponseCodeBlock(context: KotlinContext): Co
...
-      builder.addStatement("$writer.name(%S)", graphQlName)
+    if (defaultValue != null) {
+      builder.beginControlFlow("else if ($withDefaultValues)")
+      builder.addStatement("$writer.name(%S)", name)
       builder.addStatement(
           "%M.toJson($writer, $customScalarAdapters, %L)",
-          KotlinSymbols.BooleanAdapter,
-          defaultValue.value,
+          KotlinSymbols.NullableAnyAdapter,
+          defaultValue.codeBlock(),
```
martinbonnin, line 70: This is the most important change. `serializeVariables` now serializes **all** the variables if asked to (instead of just the boolean variables before)

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/SelectionSetsBuilder.kt
```diff
@@ -79,19 +80,35 @@ internal class SelectionSetsBuilder(
...
+       * When passed explicitly, the argument values are coerced (but not their default value)
+       */
+      val userValue = operationArgument?.value?.coerceInExecutableContextOrThrow(schemaArgument.type, schema)
+      IrArgument(
+          name = schemaArgument.name,
+          value = (userValue ?: schemaArgument.defaultValue)?.toIrValue(),
+          isKey =  keyArgs.contains(schemaArgument.name),
+          isPagination = paginationArgs.contains(schemaArgument.name)
+      )
+    }
```
martinbonnin, line 104: This is the other important change. The `CompiledSelections` now contain even more schema information.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/json/MapJsonReader.kt
BoD: `ApolloJsonElement` sounds good to me 👍. I remember doing a similar thing (`typealias JsonMap = Map<String, Any?>`) in `DeferredJsonMerger` which made the code less messy.

BoD (approved):
👏

PR #5389, opened 2023-11-22
Support setting port for Apollo MockServer

sdfgsdfgd:
This PR adds support for optionally setting the port for the Apollo MockServer for kotlin.

### DIFF
--- a/libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
+++ b/libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
@@ -81,6 +81,7 @@ interface MockServer : Closeable {
     private var handlePings: Boolean? = null
     private var tcpServer: TcpServer? = null
     private var listener: Listener? = null
+    private var customPort: Int = 0
 
     fun handler(handler: MockServerHandler) = apply {
       this.handler = handler
@@ -90,9 +91,11 @@ interface MockServer : Closeable {
       this.handlePings = handlePings
     }
 
-    fun tcpServer(tcpServer: TcpServer) = apply {
+    fun tcpServer(tcpServer: TcpServer, port: Int = 0) = apply {
       this.tcpServer = tcpServer
+      this.customPort = port
     }
+
 
     fun listener(listener: Listener) = apply {
       this.listener = listener
@@ -100,10 +103,11 @@ interface MockServer : Closeable {
 
 
     fun build(): MockServer {
+      val server = tcpServer ?: TcpServer(customPort)
       return MockServerImpl(
           handler ?: QueueMockServerHandler(),
           handlePings ?: true,
-          tcpServer ?: TcpServer(),
+          server,
           listener
       )
     }
@@ -154,7 +158,12 @@ internal class MockServerImpl(
     }
   }
 
-  private suspend fun handleRequests(handler: MockServerHandler, socket: TcpSocket, listener: MockServer.Listener?, onRequest: (MockRequestBase) -> Unit) {
+  private suspend fun handleRequests(
+      handler: MockServerHandler,
+      socket: TcpSocket,
+      listener: MockServer.Listener?,
+      onRequest: (MockRequestBase) -> Unit
+  ) {
     val buffer = Buffer()
     val reader = object : Reader {
       override val buffer: Buffer
@@ -235,11 +244,20 @@ internal class MockServerImpl(
 }
 
 @JsName("createMockServer")
-fun MockServer(): MockServer = MockServerImpl(QueueMockServerHandler(), true, TcpServer(), null)
+fun MockServer(port: Int = 0): MockServer = MockServerImpl(
+    QueueMockServerHandler(),
+    true,
+    TcpServer(port),
+    null)
 
 @Deprecated("Use MockServer.Builder() instead", level = DeprecationLevel.ERROR)
 @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
-fun MockServer(handler: MockServerHandler): MockServer = MockServerImpl(handler, true, TcpServer(), null)
+fun MockServer(handler: MockServerHandler, port: Int = 0): MockServer =
+    MockServerImpl(
+        handler,
+        true,
+        TcpServer(port),
+        null)
 
 @Deprecated("Use enqueueString instead", ReplaceWith("enqueueString"), DeprecationLevel.ERROR)
 @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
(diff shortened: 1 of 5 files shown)


libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/TcpServer.js.kt
```diff
@@ -82,4 +82,4 @@ internal class NodeTcpServer : TcpServer {
   }
 }
 
-actual fun TcpServer(): TcpServer = NodeTcpServer()
+actual fun TcpServer(port: Int): TcpServer = NodeTcpServer()
```
BoD, line 85: The port could be passed to `NodeTcpServer`'s `listen` call.

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -235,11 +244,20 @@ internal class MockServerImpl(
...
+fun MockServer(port: Int = 0): MockServer = MockServerImpl(
+    QueueMockServerHandler(),
+    true,
+    TcpServer(port),
+    null)
 
 @Deprecated("Use MockServer.Builder() instead", level = DeprecationLevel.ERROR)
 @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
-fun MockServer(handler: MockServerHandler): MockServer = MockServerImpl(handler, true, TcpServer(), null)
+fun MockServer(handler: MockServerHandler, port: Int = 0): MockServer =
```
BoD, line 255: I think this change may not be binary compatible - to be sure, can you run `./gradlew apiDump` and add the changes to the PR. This will give a report of the changes to the public API, so we can check everything is additive only.

BoD (commented):
Thanks a lot!

martinbonnin:
Out of curiosity, what's the use case for this? 
When writing tests involving `MockServer`, you have an instance of `mockServer` available so it's not much harder to call `mockServer.url()` vs hardcoding a localhost adress like `"http://127.0.0.1:8080"`. Are there use cases that need a fixed port?

chris-hatton:
Hi @martinbonnin , I'm on @sdfgsdfgd 's team - our use case for this is running Automated UI tests in CI.  We wish to run Maestro UI testing against a Staging APK (identical to production except pointing to our Staging backend).  The UI tests will be supported by the Apollo Mock server responses.  We'll run the Apollo Mock server in a standalone JVM daemon application.  To get the Staging APK to talk to the local mock server without realising what's up, we'll run dnsmasq locally too, to map the staging URL to the local mock server, and poke the Android Emulator with an adb command to make dnsmasq the emulators DNS server :-) That takes care of the IP direction, but not the port.  We can imagine an even more complicated environmental solution combining reverse proxying with the port number known after running the Mock Server, but it would be preferable, and seems reasonable, to simply have control over which port the Mock Server starts on.

martinbonnin:
@chris-hatton thanks for all the context 🙏. 

Since your MockServer lives in a standalone JVM application, I'm guessing you don't use any of the `MockServer.enqueue` functions (since the application under test and server are not in the same process)?

The value of Apollo `MockServer` is twofold:

* it runs on JVM + native + JS
* it is scriptable with `MockServer.enqueue` & `MockServer.takeRequest`

If you're not using any of those, there are much more powerful solutions out there like http4k, ktor or even okhttp MockWebServer. 

All in all, that's a very reasonable change and if it's allowing you to keep using the `MockServer` API you are used to then I'm all for it. 
But if someone reading this is looking for a more general HTTP server then `MockServer` is probably not a good fit. To list a bunch of non-goals, performance, HTTP2 or TLS are probably never going to be implemented in Apollo `MockServer`

martinbonnin (approved):
Thanks!

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
martinbonnin: Given it's deprecated, I would even not touch this function at all. Or is the `Builder` too cumbersome @sdfgsdfgd ? (In which case we should probably not deprecated it)

sdfgsdfgd:
thank you for the approvals ! Could you guess an ETA for this change making it into a release ?

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
sdfgsdfgd: @martinbonnin  just to be sure, given the changes approved, are you expecting any changes here (it's probably due for removal anyways, I guess I was just keeping the code consistent across the board 🙂 )

martinbonnin: I tweaked things a bit in [this commit](https://github.com/apollographql/apollo-kotlin/pull/5389/commits/556eb62a48274206fdece28fdb274095a76b6c2b), will merge once the build passes

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -92,19 +92,23 @@ interface MockServer : Closeable {
       this.handlePings = handlePings
     }
 
-    fun tcpServer(tcpServer: TcpServer, port: Int = 0) = apply {
+    fun tcpServer(tcpServer: TcpServer) = apply {
```
martinbonnin, line 95: Moved `port` to a separate function as `tcpServer` has a port configured already.

martinbonnin:
Thanks for the contribution! [SNAPSHOTs](https://github.com/apollographql/apollo-kotlin/#snapshots) will be available once [CI passes](https://github.com/apollographql/apollo-kotlin/actions/runs/6963985352). We're due to a new beta soon so it should be out soon-ish.

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -235,11 +249,21 @@ internal class MockServerImpl(
 }
 
 @JsName("createMockServer")
-fun MockServer(): MockServer = MockServerImpl(QueueMockServerHandler(), true, TcpServer(), null)
+@JvmOverloads
+fun MockServer(port: Int = 0): MockServer = MockServerImpl(
```
martinbonnin, line 253: Heads up I have reverted this change because `MockServer(Int)` was a thing in 3.x and this would make tests fail without any build warning and/or error when upgrading to 4.0. 

Since we have a builder, configuring the port can be done with:

```kotlin
    val mockServer = MockServer.Builder()
        .tcpServer(TcpServer(port = port))
        .build()
```

If this is deemed too verbose, we can add `Builder.port()` as a shorthand but I'd like us to try without first since it's always easier to add than remove.

PR #5392, opened 2023-11-23
Rework the IrType hierarchy

martinbonnin:
Prep work for `@catch` and result types. Change the internal representation of `IrType` to remove `IrOptionalType` and `IrNonNullType`. Not only does it make it easier to reason about types but it considerably reduces the cardinality of cases we have to handle:

**Allowed**

* `T`
* `Nullable<T>`
* `Option<T>`
* `Option<Nullable<T>>`
* `Result<T>`
* `Result<Nullable<T>>`
* All `List` variations of the above

**Disallowed**

* `Nullable<Option<T>>`
* `Nullable<Result<T>>`
* `Nullable<Nullable<T>>`
* `Option<Option<T>>`
* `Option<Result<T>>`
* `Result<Option<T>>`
* `Result<Result<T>>`
* etc...

We need both `Option` and `Nullable` to distinguish `?` vs `Option` in Kotlin (see this [Arrow article](https://arrow-kt.io/learn/typed-errors/nullable-and-option/)). In java `Option` and `Nullable` might be represented using the same `Optional` depending the settings.

Most of this change was made with structural search and replace but some had to be done manually because there is no 1:1 mapping between the old IrTypes and the new ones. This is the case for `adapterInitializer` for an example.

### DIFF
--- a/libraries/apollo-compiler/api/apollo-compiler.api
+++ b/libraries/apollo-compiler/api/apollo-compiler.api
@@ -411,14 +411,20 @@ public final class com/apollographql/apollo3/compiler/ir/IrClassName {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class com/apollographql/apollo3/compiler/ir/IrEnumType : com/apollographql/apollo3/compiler/ir/IrNamedType, com/apollographql/apollo3/compiler/ir/IrType {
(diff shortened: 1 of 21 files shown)


BoD (approved):
👍

PR #5394, opened 2023-11-23
OneOf: codegen changes

BoD:


libraries/apollo-api/build.gradle.kts
```diff
@@ -16,6 +16,12 @@ kotlin {
       }
     }
 
+    findByName("jvmMain")?.apply {
+      dependencies {
+        compileOnly(libs.guava.jre)
+      }
+    }
```
martinbonnin, line 23: I have mixed feelings about pulling guava in `apollo-api`, even if it's compileOnly. `assertOneOf` isn't that huge. Could we generate it with kotlin poet instead?

martinbonnin: Or maybe just write it verbatim, no need for KotliPoet actually. Get a template and copy it to `packageName.utils.Assertions.kt`?

libraries/apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/Assertions.jvm.kt
```diff
@@ -0,0 +1,27 @@
+@file:JvmMultifileClass
+@file:JvmName("Assertions")
+
+package com.apollographql.apollo3.api
+
+@SafeVarargs
+@Suppress("NewApi")
```
martinbonnin, line 7: Is this safe?

BoD: Actually no 😱. Good catch! `Optional` itself isn't a problem since this method will only be called in projects that have it - however `isEmpty()` is level 33+. I'll use `isPresent()` instead which is 24+, same as `Optional` itself.

libraries/apollo-api/build.gradle.kts
BoD: Done in f5efb16aa992d6cb831d408fa21636349b9a0044

libraries/apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/Assertions.jvm.kt
martinbonnin: Add a comment in the code that it's safe to use `24+` because it's the same as `Optional` itself?

martinbonnin (approved):

PR #5395, opened 2023-11-23
[IJ Plugin] Add inspection for @oneOf input type constructor invocation

BoD:
Screenshot 2023-11-23 at 20 11 47

----
Note: this also changes the return types of `Optional.absent()` and `Optional.present()`.

### DIFF
--- a/intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
+++ b/intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
@@ -1,0 +1,8 @@
+<html>
+<body>
+Reports invalid constructor invocations of <code>@oneOf</code> input types.
+<p>
+    Exactly one field must be set, and it must be <code>Present</code>.
+</p>
+</body>
+</html>
--- a/intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
+++ b/intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
@@ -2,7 +2,7 @@
 <body>
 Reports invalid constructor invocations of <code>@oneOf</code> input types.
 <p>
-    Exactly one field must be set, and it must be <code>Present</code>.
+    Constructor of `@oneOf` input class must have exactly one <code>Present</code> argument.
 </p>
 </body>
 </html>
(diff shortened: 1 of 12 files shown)


libraries/apollo-api/src/commonTest/kotlin/test/OptionalTest.kt
```diff
@@ -37,7 +37,7 @@ class OptionalTest {
 
   @Test
   fun getOrThrowAbsent() {
-    val optional = Optional.absent<String?>()
+    val optional: Optional<String?> = Optional.absent()
```
BoD, line 40: Inference wasn't working here, I guess that means the `absent()` change is source-breaking in some cases :(

BoD: Yes it makes more sense like this ✅

intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
```diff
@@ -0,0 +1,8 @@
+<html>
+<body>
+Reports invalid constructor invocations of <code>@oneOf</code> input types.
+<p>
+    Exactly one field must be set, and it must be <code>Present</code>.
```
martinbonnin, line 5:
```suggestion
    Exactly one input field must be set, and it must be <code>Present</code>.
```

martinbonnin: Or maybe more like so?

```suggestion
    Constructor of `@oneOf` input class must have exactly one <code>Present</code> argument.
```

* Field is JVM/GraphQL. 
* Property is Kotlin
* Constructor takes arguments?

martinbonnin (approved):
A bit of naming bikeshedding but LGTM!

intellij-plugin/src/main/resources/inspectionDescriptions/ApolloOneOfInputCreation.html
BoD: Yup, much better!

intellij-plugin/src/main/resources/messages/ApolloBundle.properties
```diff
@@ -142,6 +142,10 @@ inspection.endpointNotConfigured.displayName=GraphQL endpoint not configured
 inspection.endpointNotConfigured.reportText=GraphQL endpoint not configured
 inspection.endpointNotConfigured.quickFix=Add introspection block
 
+inspection.oneOfInputCreation.displayName=OneOf Input Object creation issue
+inspection.oneOfInputCreation.reportText.wrongNumberOfArgs=<html><tt>@oneOf</tt> input must have exactly one field set
```
BoD, line 146:
```suggestion
inspection.oneOfInputCreation.reportText.wrongNumberOfArgs=<html><tt>@oneOf</tt> input class constructor must have exactly one argument
```

martinbonnin (approved):

Issue #5379, opened 2023-11-17
Allow the cache viewer to sort items more "intelligently" when  showing cache entries that have a number at the end


StylianosGakis:
### Use case

When looking at my cache entries in the IDE Plugin, in the window "Apollo Normalized Cache", my cache entries are sorted like this:
```
QUERY_ROOT
chat()
chat().messages.0
chat().messages.1
chat().messages.10
chat().messages.11
chat().messages.12
chat().messages.13
chat().messages.14
chat().messages.2
chat().messages.3
chat().messages.4
chat().messages.5
chat().messages.6
chat().messages.7
chat().messages.8
chat().messages.9
```
This is maybe not always a big deal, but in my case, working with reading chat messages from the cache, it would actually be quite convenient for the entries to be sorted exactly as they came into the cache to easier debug what is going on there.

### Describe the solution you'd like

The current sorting sorts per character, and that looks like it's sorting 1 -> 10 -> 2 instead of 1 -> 2 -> ... -> 10 -> 11 and so on.
For this to be fixed, the sorting would need to happen taking into consideration when numbers are involved and sorting those in a special way compared to other strings.

StylianosGakis:
Wow thanks a lot!

PR #5396, opened 2023-11-24
[IJ Plugin] Cache viewer: take numbers into account in key sorting

BoD:
Fix for #5379

Screenshot 2023-11-24 at 07 14 04

### DIFF
--- a/intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/normalizedcache/NormalizedCache.kt
+++ b/intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/normalizedcache/NormalizedCache.kt
@@ -31,11 +31,52 @@ data class NormalizedCache(
   )
 
   companion object {
+    /**
+     * Compares records by key, with `QUERY_ROOT` first, then by alphabetical order, taking numbers into account.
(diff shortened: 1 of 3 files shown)


intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/normalizedcache/NormalizedCache.kt
```diff
@@ -31,11 +31,52 @@ data class NormalizedCache(
...
+        key2 == "QUERY_ROOT" -> 1
+        else -> {
+          var idx1 = 0
+          var idx2 = 0
+          while (true) {
+            val c1 = key1[idx1]
+            val c2 = key2[idx2]
+            if (c1.isDigit() && c2.isDigit()) {
+              var number1 = 0
+              var number2 = 0
```
martinbonnin, line 52: Make them BigInteger so that we're overflow proof. Not a huge deal but I think there's a law saying that if something bad _may_ happen, that something _will_ happen? 🤷

martinbonnin: Oh well nevermind, it's only list indices (not ids or other ints) so yea, if we have more than INT_MAX list items, we'll probably have more pressing matters 👍

martinbonnin (approved):

PR #5402, opened 2023-11-27
[codegen] Throw a better exception than NullPointerException if a value is missing

martinbonnin:
This aligns the Kotlin codegen with the Java codegen. Also gives a more descriptive error than just "NullPointerException"


BoD (approved):
👍

PR #5403, opened 2023-11-27
[codegen] Go back to just one Adapter class

martinbonnin:
Go back to tunneling everything inside `CustomScalarAdapters`... 

This simplifies the code and with `@catch` coming up, both the composite and scalar adapters need the `errors` as a context. This could have been done by allocating a new adapter holding the `errors` if needed but that means we could not reuse adapters if we wanted to.


BoD (approved):
⚙️

PR #5408, opened 2023-11-28
[apollo-runtime] Allow no data and no exception in case of GraphQL errors

martinbonnin:
Revert filling `exception` in case of GraphQL errors. Using error aware parsing is a better solution for explicit exceptions on GraphQL errors


BoD (approved):
👍

PR #5410, opened 2023-11-28
[3.x] bump Gradle to 8.5

martinbonnin:


martinbonnin:
Nevermind, that requires too many changes, I'll downgrade my Java instead

PR #5412, opened 2023-11-29
[3.x] update Kotlin to 1.9

martinbonnin:
I needed Kotlin 1.9 to support new XCode command line options, which in turns needed other things.

~I think this is mostly transparent for users (languageVersion is still 1.5) but if we're ever doing a 3.x maintenance version, we should call this out in the changelog.~

Welp, I had to bump the Kotlin stdlib to 1.9.20 as well or `compileCommonMainMetadata` [fails](https://github.com/apollographql/apollo-kotlin/actions/runs/7031401469/job/19132961448?pr=5412). Definitely something to call out in the changelog. We now require Kotlin 1.8 to compile against apollo-kotlin (see https://github.com/apollographql/apollo-kotlin/pull/5412/commits/c1e24e16aa85e2f16291b50f97f81d48bc2ed300)


build-logic/src/main/kotlin/CompilerOptions.kt
```diff
@@ -56,7 +63,7 @@ fun Project.configureJavaAndKotlinCompilers(treatWarningsAsErrors: Boolean = fal
 private fun Project.treatWarningsAsErrors() {
   tasks.withType(KotlinCompile::class.java) {
     kotlinOptions {
-      allWarningsAsErrors = true
+      //allWarningsAsErrors = true
```
BoD, line 66: On purpose?

martinbonnin: Yes, sadly `languageVersion = 1.5` outputs a warning that we can't selectively disable. I'll add a comment

BoD (approved):
👍

martinbonnin:
Had to change a bunch of things and bump the kotlin-stdlib so it's not a transparent change but this is life. software evolves!

PR #5417, opened 2023-11-30
[apollo-runtime] fine grained exception + responses migration

martinbonnin:
Replace usev3ExceptionHandling with:

- `ApolloCall.executeV3()`
- `ApolloCall.toFlowV3()`

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo3
 
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.ExecutionContext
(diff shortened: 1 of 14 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -125,7 +126,26 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
+    return apolloClient.executeAsFlow(toApolloRequest(), ignoreApolloClientHttpHeaders = ignoreApolloClientHttpHeaders == true, false)
+  }
+
+  /**
+   * A version of [execute] that restores 3.x behaviour:
+   * - throw on fetch errors.
+   * - make `CacheFirst`, `NetworkFirst` and `CacheAndNetwork` policies ignore fetch errors.
+   * - throw ApolloComposite exception if needed.
+   */
+  @Deprecated("Use toFlow() and handle ApolloResponse.exception instead")
```
BoD, line 138: Should we add the `replaceWith`?

BoD (approved):
Looking good 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
martinbonnin: I prefer not to because it's not a "drop-in" replacement, the callsite needs to be changed. I don't want users to press Alt-Enter in automatic mode when going through deprecations.

BoD: Makes sense 👍

PR #5422, opened 2023-12-01
Add a few symbols as ERROR deprecation for the migration

martinbonnin:
A few symbols were removed without going through `ERROR` deprecation. Keep them around for a bit more time


BoD (approved):
👍

PR #5425, opened 2023-12-04
Remove MockServer(port) as it clashes with 3.x MockServer(timeout)

martinbonnin:
See [here](https://github.com/apollographql/apollo-kotlin/pull/5418/files#diff-79bdedd59fc1530b6d765978a0aa9a15053e0f614d681a2c338e8e15500119c0L57) for an example


BoD (approved):
👍

PR #5426, opened 2023-12-04
[api] remove some of the ApolloResponse.Builders

martinbonnin:
Right now we have:

1 `Builder(operation, uuid, data)`
2 `Builder(operation, uuid, exception)`
3 `Builder(operation, uuid, data, errors, extensions, exception)`

1 & 2 are somewhat misleading because they could imply that `data` & `exception` are exclusive which is not the case. 

3. is redundant with calling the different methods of the Builder.

Replace all 3 builders with:

* `Builder(operation, uuid)`


BoD (approved):
👍

PR #5433, opened 2023-12-05
Fix publishing

martinbonnin:
Workaround for https://github.com/gradle/gradle/issues/26132

Also revert the version to 4.0.0-beta.3, I'll push the tag manually


BoD (approved):
🪲

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo3/debugserver/internal/graphql/GraphQL.kt
```diff
@@ -114,11 +114,11 @@ internal class NormalizedCache(
...
 ) {
-  fun key() = record.key
+  fun key(): String = record.key
 
-  fun fields() = record.fields
+  fun fields(): Map<String, Any?> = record.fields
 
-  fun sizeInBytes() = record.sizeInBytes
+  fun sizeInBytes(): Int = record.sizeInBytes
 }
```
martinbonnin, line 122: Filed https://github.com/apollographql/apollo-kotlin/issues/5434 as a follow up

PR #5444, opened 2023-12-06
Add validation to check schema definitions are compatible with the bundled ones 

BoD:
Make sure the definitions are "semantically equal".

### DIFF
--- a/libraries/apollo-ast/api/apollo-ast.api
+++ b/libraries/apollo-ast/api/apollo-ast.api
@@ -768,8 +768,6 @@ public final class com/apollographql/apollo3/ast/GqldirectiveKt {
 }
 
 public final class com/apollographql/apollo3/ast/GqldocumentKt {
-	public static final field KOTLIN_LABS_VERSION Ljava/lang/String;
-	public static final field NULLABILITY_VERSION Ljava/lang/String;
 	public static final fun builtinDefinitions ()Ljava/util/List;
 	public static final fun kotlinLabsDefinitions (Ljava/lang/String;)Ljava/util/List;
 	public static final fun linkDefinitions ()Ljava/util/List;
--- a/libraries/apollo-ast/api/apollo-ast.api
+++ b/libraries/apollo-ast/api/apollo-ast.api
@@ -768,6 +768,8 @@ public final class com/apollographql/apollo3/ast/GqldirectiveKt {
 }
 
 public final class com/apollographql/apollo3/ast/GqldocumentKt {
+	public static final field KOTLIN_LABS_VERSION Ljava/lang/String;
+	public static final field NULLABILITY_VERSION Ljava/lang/String;
 	public static final fun builtinDefinitions ()Ljava/util/List;
 	public static final fun kotlinLabsDefinitions (Ljava/lang/String;)Ljava/util/List;
 	public static final fun linkDefinitions ()Ljava/util/List;
@@ -827,7 +829,7 @@ public abstract interface class com/apollographql/apollo3/ast/GraphQLIssue : com
 public abstract interface class com/apollographql/apollo3/ast/GraphQLValidationIssue : com/apollographql/apollo3/ast/GraphQLIssue {
 }
 
-public final class com/apollographql/apollo3/ast/IncompatibleDirectiveDefinition : com/apollographql/apollo3/ast/GraphQLValidationIssue {
+public final class com/apollographql/apollo3/ast/IncompatibleDefinition : com/apollographql/apollo3/ast/GraphQLValidationIssue {
 	public fun <init> (Ljava/lang/String;Ljava/lang/String;Lcom/apollographql/apollo3/ast/SourceLocation;)V
 	public fun getMessage ()Ljava/lang/String;
 	public fun getSourceLocation ()Lcom/apollographql/apollo3/ast/SourceLocation;
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
@@ -60,6 +60,17 @@ class IncompatibleDirectiveDefinition(
 }
 
 /**
+ * The enum definition is inconsistent with the expected one.
+ */
+class IncompatibleEnumDefinition(
+    enumName: String,
+    expectedDefinition: String,
+    override val sourceLocation: SourceLocation?,
+) : GraphQLValidationIssue {
+  override val message = "Unexpected '@$enumName' enum definition. Expecting '$expectedDefinition'."
+}
+
+/**
  * Fields have different shapes and cannot be merged
  *
  */
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
@@ -56,7 +56,7 @@ class IncompatibleDefinition(
     expectedDefinition: String,
     override val sourceLocation: SourceLocation?,
 ) : GraphQLValidationIssue {
-  override val message = "Unexpected '@$name' definition. Expecting '$expectedDefinition'."
+  override val message = "Unexpected '$name' definition. Expecting '$expectedDefinition'."
 }
 
 /**
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
@@ -49,25 +49,14 @@ class UnknownDirective @ApolloInternal constructor(
 }
 
 /**
- * The directive definition is inconsistent with the expected one.
+ * The definition is inconsistent with the expected one.
  */
-class IncompatibleDirectiveDefinition(
-    directiveName: String,
+class IncompatibleDefinition(
+    name: String,
     expectedDefinition: String,
     override val sourceLocation: SourceLocation?,
 ) : GraphQLValidationIssue {
-  override val message = "Unexpected '@$directiveName' directive definition. Expecting '$expectedDefinition'."
-}
-
-/**
- * The enum definition is inconsistent with the expected one.
- */
-class IncompatibleEnumDefinition(
-    enumName: String,
-    expectedDefinition: String,
-    override val sourceLocation: SourceLocation?,
-) : GraphQLValidationIssue {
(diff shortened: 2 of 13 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gqldirective.kt
```diff
@@ -122,26 +122,31 @@ private fun GQLValue?.toBoolean(): Boolean {
...
 
-private fun GQLValue?.toCatchTo(): CatchTo {
+private fun GQLValue?.toCatchTo(): CatchTo? {
   return when (this) {
     is GQLEnumValue -> when (this.value) {
       "NULL" -> CatchTo.NULL
       "RESULT" -> CatchTo.RESULT
       "THROW" -> CatchTo.THROW
-      else -> error("Unknown CatchTo value: ${this.value}")
+      else -> null
```
martinbonnin, line 131: For the record, we just discussed that and decided to move this to validation. Validation should check that the `@catch` directive definition is compatible with the one we're expecting, like for `@oneOf`

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/Issue.kt
```diff
@@ -59,6 +59,17 @@ class IncompatibleDirectiveDefinition(
...
+/**
+ * The enum definition is inconsistent with the expected one.
+ */
+class IncompatibleEnumDefinition(
+    enumName: String,
+    expectedDefinition: String,
+    override val sourceLocation: SourceLocation?,
+) : GraphQLValidationIssue {
+  override val message = "Unexpected '@$enumName' enum definition. Expecting '$expectedDefinition'."
+}
```
martinbonnin, line 71: Should we go for a generic `IncompatibleDefinition` instead? If we want to have the comparison code generic at some point?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -0,0 +1,187 @@
...
+import com.apollographql.apollo3.ast.GQLNonNullType
+import com.apollographql.apollo3.ast.GQLNullValue
+import com.apollographql.apollo3.ast.GQLObjectValue
+import com.apollographql.apollo3.ast.GQLStringValue
+import com.apollographql.apollo3.ast.GQLType
+import com.apollographql.apollo3.ast.GQLValue
+import com.apollographql.apollo3.ast.GQLVariableValue
+import com.apollographql.apollo3.ast.toUtf8
+
+internal fun GQLDirectiveDefinition.semanticEquals(other: GQLDirectiveDefinition): Boolean {
```
martinbonnin, line 22: 👍 
Also: so much semantic things these days 😄

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -0,0 +1,187 @@
...
+internal fun GQLEnumTypeDefinition.semanticEquals(definition: GQLEnumTypeDefinition): Boolean {
+  if (name != definition.name) {
+    return false
+  }
+
+  if (directives.size != definition.directives.size) {
+    return false
+  }
+
+  directives.sortedBy { it.name }.zip(definition.directives.sortedBy { it.name }).forEach { (a, b) ->
```
martinbonnin, line 129: [Directive order is significant](https://spec.graphql.org/draft/#sel-HAFjBRBAABABMw4b)

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
martinbonnin: Could that be done purely on `GQLNode`?

wildly untested pseudo code:

```kotlin
fun GQLNode.semanticEquals(other: GQLNode): Boolean {
  val equals = when (this) {
    is GQLObjectTypeDefinition -> other is GQLObjectTypeDefinition && name == other.name
    // more stuff
  }

  if (children.size != other.children.size) {
    return false
  }
  for (i in children.indices) {
    if (children[i].semanticEquals(other.children[i]) {
      return false
    }
  }
  return  true
}
```

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
BoD: Woops that makes sense. Good catch!

libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -768,6 +768,8 @@ public final class com/apollographql/apollo3/ast/GqldirectiveKt {
 }
 
 public final class com/apollographql/apollo3/ast/GqldocumentKt {
+	public static final field KOTLIN_LABS_VERSION Ljava/lang/String;
+	public static final field NULLABILITY_VERSION Ljava/lang/String;
```
martinbonnin, line 772: Can we hide that from public API?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -1,183 +1,200 @@
...
-  }
-
-  if (locations != other.locations) {
-    return false
-  }
-
-  if (repeatable != other.repeatable) {
-    return false
-  }
+internal fun GQLNode.semanticEquals(other: GQLNode): Boolean {
```
martinbonnin, line 24: Add a small KDoc? Especially mentioning that this doesn't check description and sourceLocation?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -1,183 +1,200 @@
...
+  when (this) {
+    is GQLDirectiveDefinition -> {
+      if (other !is GQLDirectiveDefinition) {
+        return false
```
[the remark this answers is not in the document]
BoD: It's handled later with `if (this is GQLNamed) {`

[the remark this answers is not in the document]
martinbonnin: I see 👍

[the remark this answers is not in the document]
martinbonnin: It's currently a bit weird though because the check is done twice for `GQLVariableValue` for an example. It implements `GQLNamed` but also makes the explicit check

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -0,0 +1,204 @@
...
+    is GQLArgument -> {
+      if (other !is GQLArgument) {
+        return false
+      }
+      if (name != other.name) {
+        return false
+      }
+    }
+
+    else -> {}
```
martinbonnin, line 175: We should fail with `TODO()` here in case we ever want to try that on other nodes.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
```diff
@@ -132,9 +134,33 @@ internal fun validateSchema(definitions: List<GQLDefinition>, requiresApolloDefi
...
+          if (!existing.semanticEquals(definition)) {
+            issues.add(IncompatibleDefinition(definition.name, definition.toSemanticSdl(), definition.sourceLocation))
+          }
+        }
+      }
+
+      is GQLEnumTypeDefinition -> {
+        val existing = typeDefinitions[definition.name]
+        if (existing != null) {
+          if (existing !is GQLEnumTypeDefinition || !existing.semanticEquals(definition)) {
```
martinbonnin, line 151:
```suggestion
          if (!existing.semanticEquals(definition)) {
```

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/semanticEquals.kt
```diff
@@ -21,7 +22,13 @@ import com.apollographql.apollo3.ast.GQLStringValue
 import com.apollographql.apollo3.ast.GQLVariableValue
 import com.apollographql.apollo3.ast.toUtf8
 
-internal fun GQLNode.semanticEquals(other: GQLNode): Boolean {
+/**
+ * Returns true if the two nodes are semantically equal, which ignores the source location and the description.
+ * Note that not all cases are implemented - currently [GQLEnumTypeDefinition] and [GQLDirectiveDefinition] are fully supported, and
```
martinbonnin, line 27: Feels like we should support `GQLInputTypeDefinition` and `GQLScalarTypeDefinition` as well given they can be directive arguments.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gqldirective.kt
```diff
@@ -135,13 +135,18 @@ private fun GQLValue?.toCatchTo(): CatchTo {
   }
 }
 
+private fun GQLDirective.isDefinedAndMatchesOriginalName(schema: Schema, originalName: String): Boolean {
+  return schema.directiveDefinitions.get(name) != null && schema.originalDirectiveName(name) == originalName
```
martinbonnin, line 139: Apologies for nitpicking this but I'd rather have the codegen crash than silently ignore missing directive definitions. 
All directive usage of `@catch` must be validated at that point so if we don't have a directive definition, it's a programming error that I'd rather catch (no pun intended) early

martinbonnin:
Can you update the title/description of this PR ?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/gqldirective.kt
BoD: Not sure why I had the impression it would still NPE without this, but it's not the case - just removed it 👍

martinbonnin (approved):

PR #5452, opened 2023-12-11
[apollo-tooling] bump bootstraped version of Apollo to `4.0.0-beta.3`

martinbonnin:
Eat our own `@catch` food.


BoD (approved):
👍

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaUploader.kt
```diff
@@ -114,14 +113,14 @@ object SchemaUploader {
...
+      when (val e = response.exception) {
         is ApolloHttpException -> {
           val body = e.body?.use { it.readUtf8() } ?: ""
           throw Exception("Cannot upload schema: (code: ${e.statusCode})\n$body", e)
         }
 
-        is ApolloGraphQLException -> {
-          throw Exception("Cannot upload schema: ${e.errors.joinToString { it.message }}")
+        null -> {
+          throw Exception("Cannot upload schema: ${response.errors?.joinToString { it.message }}")
```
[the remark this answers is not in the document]
BoD: It makes sense to me to to treat any errors as a fail here, so yes maybe `@catch(to: THROW)` would be cool. In that case we'd want `is ApolloGraphQLException` back.

BoD:
Field insights ✅

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/SchemaUploader.kt
```diff
@@ -67,27 +66,30 @@ object SchemaUploader {
...
       }
-    } else {
-      val code = data.graph?.uploadSchema?.code
-      val message = data.graph?.uploadSchema?.message
-      val success = data.graph?.uploadSchema?.success
-      check(success == true) {
-        "Cannot upload schema (code: $code): $message"
+
+      else -> {
+        throw Exception("Cannot upload schema: ${e.message}", e)
```
BoD, line 92: Should we make `ApolloGraphQLException.message` receive the GraphQL error's message? Currently it's a generic `"GraphQL error"`. (We land on this `else` case if `graph` isn't null but `uploadSchema` is.)

martinbonnin: Yes good idea 👍

BoD (approved):
✅

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/RegisterOperations.kt
```diff
@@ -255,30 +254,34 @@ object RegisterOperations {
...
-          throw Exception("Cannot push operations: ${e.errors.joinToString { it.message }}")
-        }
-
-        else -> {
-          throw Exception("Cannot push operations: ${e.message}", e)
-        }
+    if (data != null) {
+      val service = data.service
+      if (service == null) {
+        throw Exception("Cannot push operations: cannot find service '$graphID': ${response.errors?.joinToString { it.message }}")
```
martinbonnin, line 260: This is somewhat inconsistent. If we expect an error on null, we should make `service` `@semanticNonNull`. Do you remember what's happening in those cases?

martinbonnin:
Made another pass at it. Looks like we can factor the error handling. See `ApolloResponse.toException` ([source](https://github.com/apollographql/apollo-kotlin/blob/5414ee1717efe692f573745afda237ed4f6a3a3f/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/exceptions.kt#L14)). Let me know what you think

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/Telemetry.kt
```diff
@@ -39,24 +37,12 @@ object Telemetry {
...
-      }
-
-      else -> {
-        Result.failure(Exception("Cannot track Apollo Kotlin usage: ${e.message}", e))
-      }
+    if (response.data == null) {
+      throw response.toException("Cannot track Apollo Kotlin usage")
     }
+
+    return Result.success(Unit)
```
martinbonnin, line 45: Should we check `data.trackApolloKotlinUsage` for a specific value here?

BoD: `trackApolloKotlinUsage` is always `null` (the `Void` scalar)

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/RegisterOperations.kt
martinbonnin: Ah, it's the deprecated one, apologies. Let's keep this one as is for now but there is also the issue for `FieldInsights`, I'll comment there.

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/FieldInsights.kt
```diff
@@ -32,45 +30,30 @@ object FieldInsights {
...
-          FieldLatencies.FieldLatency(
-              parentType = parentType,
-              fieldName = fieldName,
-              durationMs = durationMs
-          )
-        } ?: emptyList())
-      }
+    val service = data.service
+    if (service == null) {
+      return FieldLatenciesResult.Error(cause = Exception("Cannot find service $serviceId: ${response.errors?.joinToString { it.message }}}"))
```
martinbonnin, line 41: Is this 
* semantic null => no need to display `response.errors`
* error null => we could make this `@semanticNonNull`

BoD: My hunch is semantic null but I have no proof 😅

martinbonnin: We can know for sure if semantic null is possible though, right? If passing a bad service id, do we get an error in the errors array?

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/Telemetry.kt
martinbonnin: I mapped `Void` to `Unit` [here](https://github.com/apollographql/apollo-kotlin/pull/5452/commits/edd88ade9a1a08c1aa9f68e915f72b007ded1383).

Funnily, `Void` can be nullable so it's a `null` of `null` 😄

martinbonnin: Unless `Void` is an error 👀

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/FieldInsights.kt
BoD: Just checked -> nothing in `errors` when passing a wrong service id.

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/Telemetry.kt
BoD: >Funnily, Void can be nullable so it's a null of null 😄

🤯

libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/RegisterOperations.kt
martinbonnin: Added some deprecation and comments [there](https://github.com/apollographql/apollo-kotlin/commit/79e68b16b33268cb854b66dc55627ac3fbcb7ad3)

BoD (approved):
> Made another pass at it. Looks like we can factor the error handling. See ApolloResponse.toException ([source](https://github.com/apollographql/apollo-kotlin/blob/5414ee1717efe692f573745afda237ed4f6a3a3f/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo3/tooling/exceptions.kt#L14)). Let me know what you think

Yes looks good 👍

martinbonnin:
I merged this PR so it doesn't become stale and we can move forward with other PRs. Reminder to ourselves to re-test again using the integration test since there's been a few changes since your last tests.

PR #5471, opened 2023-12-12
Update declarative-ids.mdx

FredLoh:
Call out that extras.graphql needs to be referenced in your schema configuration or it will be silently ignored!


martinbonnin (approved):
Thank you!

BoD (approved):

PR #5477, opened 2023-12-13
[IJ plugin] Don't suggest v4 migration until it is stable

BoD:


intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/refactoring/migration/v3tov4/ApolloV3ToV4MigrationProcessor.kt
```diff
@@ -24,12 +24,14 @@ import com.apollographql.ijplugin.refactoring.migration.v3tov4.item.UpdateThrowA
...
 import com.intellij.openapi.project.Project
 
-private const val apollo4LatestVersion = "4.0.0-beta.4"
-
 /**
  * Migrations of Apollo Kotlin v3 to v4.
  */
 class ApolloV3ToV4MigrationProcessor(project: Project) : ApolloMigrationRefactoringProcessor(project) {
+  companion object {
+    const val apollo4LatestVersion = "4.0.0-beta.4"
```
martinbonnin, line 32: You could use `com.apollographql.apollo3.compiler.APOLLO_VERSION` to have this match the published version automatically without having to bump this every time

intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/inspection/Apollo4AvailableInspection.kt
```diff
@@ -36,11 +37,15 @@ class Apollo4AvailableInspection : LocalInspectionTool() {
   // XXX kts files are not highlighted in tests
   private val buildGradleFileName = if (isUnitTestMode()) "build.gradle.kt" else "build.gradle.kts"
 
+  // Do not warn until 4.0 is stable (but keep enabled always in unit tests)
+  private val isEnabled = isUnitTestMode() || !ApolloV3ToV4MigrationProcessor.apollo4LatestVersion.contains('-')
```
martinbonnin, line 41: This doesn't work for SNAPSHOTs like `4.0.1-SNAPSHOT` which is OK if you maintain `apollo4LatestVersion` manually. Alternative would be using something like https://github.com/swiftzer/semver to do proper semver but that's more involved. Your call.

martinbonnin (approved):

intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/inspection/Apollo4AvailableInspection.kt
BoD: I'll keep as-is to very slightly (debatable!) err on Bill's razor's side

martinbonnin: Sounds good 👍 . Also, who's Bill? 👀

BoD: Our friend good ol' Bill Occam!

martinbonnin: Aahh William! Or is he known under 2 names? Which would be quite ironic 🙃

PR #5479, opened 2023-12-13
Debug server: don't crash when a client has no caches

BoD:
Thanks @rohandhruva for the report!

### DIFF
--- a/libraries/apollo-debug-server/build.gradle.kts
+++ b/libraries/apollo-debug-server/build.gradle.kts
@@ -24,9 +24,9 @@ kotlin {
 
       dependencies {
         implementation(project(":apollo-normalized-cache"))
-
-        api(project(":apollo-ast"))
-        api(project(":apollo-api"))
+        implementation(project(":apollo-normalized-cache-api"))
+        implementation(project(":apollo-ast"))
+        api(project(":apollo-runtime"))
       }
     }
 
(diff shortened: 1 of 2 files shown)


libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -24,9 +24,9 @@ kotlin {
 
       dependencies {
         implementation(project(":apollo-normalized-cache"))
-
-        api(project(":apollo-ast"))
-        api(project(":apollo-api"))
+        implementation(project(":apollo-normalized-cache-api"))
```
BoD, line 27: I don't think this one should be necessary (it's `api` in `:apollo-normalized-cache`), but I'm getting red inside IJ without it 🤷‍♀️

libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -24,9 +24,9 @@ kotlin {
...
+        implementation(project(":apollo-ast"))
```
BoD, line 28: This shouldn't have been `api`.

libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -24,9 +24,9 @@ kotlin {
...
+        api(project(":apollo-runtime"))
```
BoD, line 29: `ApolloClient` is in `apollo-runtime`, not `apollo-api`

martinbonnin (approved):

PR #5483, opened 2023-12-14
Add scheduled gradle plugin tests using the Platform API

### LINKED ISSUE #5466: Add platform API tests
Add test that run against a test graph in Studio. As those are hitting a stateful network service, we don't want to run them in each PR but maybe every night or so would be useful

BoD:
Resolves #5466

Opening as draft for now, waiting to create a dedicated account on Studio to use.


.github/workflows/platform-api-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef #v2.3.3
+
+      - name: Run Gradle Plugin tests against the Apollo Platform API
+        env:
+          COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_ENABLED: true
+          COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_KEY: ${{ secrets.COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_KEY }}
+          COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_GRAPH: ${{ secrets.COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_GRAPH }}
+          COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_SUBGRAPH: ${{ secrets.COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_SUBGRAPH }}
+          COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_PQ_LIST_ID: ${{ secrets.COM_APOLLOGRAPHQL_PLATFORM_API_TESTS_PQ_LIST_ID }}
```
martinbonnin, line 32: Is there a reason we prefix everything with `COM_APOLLO_GRAPHQL` here?

martinbonnin: +1 to simpler. 

I think I started the tradition with `COM_APOLLOGRAPHQL_PROFILE_ID` but the `COM_APOLLOGRAPHQL` was related to the name of the staging profile (`com.apollographql`) but it's not really needed here.

martinbonnin:
What about making this an integration test instead? Gradle tests are always a bit slower to run because:

* they publish a bunch of native publications to the local repo as a pre-requisite
* they run Gradle inside Gradle which sounds like it is slower (I haven't measured it)
* They're harder to run from the IDE

In general I'd keep Gradle integration tests for scenarios where:
* We need to test a specific version of Gradle
* We need to test up-to-date checks, build cache or other things that require instrumenting the outcome

In this case, we're only checking if the result is successful so integration test would work I think.

As a bonus, there is no need to tunnel the env variables anymore, they are directly accessible.

.github/workflows/platform-api-tests.yml
martinbonnin: I'd be okay with just `PLATFORM_API_TESTS_KEY` and the others in plain text (I don't think they are secret?)

Reduces the amount of logic that is in GitHub actions and easier to test locally.

.github/workflows/platform-api-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef #v2.3.3
+
+      - name: Run tests against the Apollo Platform API
+        env:
+          PLATFORM_API_TESTS_KEY: ${{ secrets.PLATFORM_API_TESTS_KEY }}
+        run: |
+          ./gradlew -p tests --no-daemon --console plain :platform-api:registerServiceApolloOperations
+          ./gradlew -p tests --no-daemon --console plain :platform-api:downloadServiceApolloSchemaFromRegistry
+          ./gradlew -p tests --no-daemon --console plain :platform-api:pushApolloSchema --key=$PLATFORM_API_TESTS_KEY --subgraph=subgraph1 --schema=platform-api/src/main/graphql/schema.graphqls --revision=1 --graphVariant=current
```
martinbonnin, line 36: Should we start adding `--no-daemon --console plain` in other places too?

Also nit: my favourite solution would be defining a task in `platform-api/build.gradle.kts`, `endToEndTest` or something like that so we have more logic in Kotlin files and less logic in yaml files.

martinbonnin (approved):

.github/workflows/platform-api-tests.yml
BoD: - `--console plain` is definitely useless, logs are already good in the other jobs that don't have it
- `--no-daemon` is probably only useful in jobs that have only one run, to save a few seconds. Here we have several, so it's probably actually slower for no good reason.

I'll remove

BoD: [Pushed](https://github.com/apollographql/apollo-kotlin/pull/5483/commits/7ef9a43e2c2d29e1ab9b2ca4eb20963bfc0b417b) a tweak to move the logic to `build.gradle.kts`. Uncommented temporarily `pull_request: branches: [ '*' ]` to make sure this works.

martinbonnin (approved):
Nice 👍

martinbonnin:
I [added some logs](https://github.com/apollographql/apollo-kotlin/pull/5483/commits/856919cecb1f39b1ac6786b8b0f373f0f6acbeb7) for consistency. Previously, only PQL would output something.

```
> Task :platform-api:downloadServiceApolloSchemaFromRegistry
Apollo: schema downloaded to /Users/mbonnin/git/apollo-kotlin/tests/platform-api/src/main/graphql/schema.graphqls

> Task :platform-api:pushApolloSchema
Apollo: schema pushed successfully

> Task :platform-api:registerServiceApolloOperations
Apollo: persisted query list uploaded successfully

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.5/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 20s
```

It can be argued whether those logs are really needed. If they are an issue, we should downgrade them to `info`

.github/workflows/platform-api-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+          java-version: 17
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@3fbe033aaae657f011f88f29be9e65ed26bd29ef #v2.3.3
+
+      - name: Run tests against the Apollo Platform API
+        env:
+          PLATFORM_API_TESTS_KEY: ${{ secrets.PLATFORM_API_TESTS_KEY }}
+        run: |
+          ./gradlew -p tests :platform-api:platformApiTests
```
BoD, line 34:
```suggestion
          ./gradlew --no-daemon -p tests :platform-api:platformApiTests
```

.github/workflows/platform-api-tests.yml
BoD: Added `--no-daemon` back since there's only one task again :)

martinbonnin: 👍 It'll run so fast now ⚡

PR #5484, opened 2023-12-14
Don't crash if ApolloDebugServerInitializer is not run (e.g. in unit tests)

BoD:
Thanks @rohandhruva for the report!


martinbonnin (approved):
👍

PR #5486, opened 2023-12-14
[apollo-cli] Publish apollo-cli-incubating again

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/5468

Since `apollo-cli-incubating` was published before, let's keep it around. Also explicit a bit more the stability of different artifacts until we figure out if we want more fine-grained versioning.


BoD (approved):
👍

PR #5489, opened 2023-12-15
[mockserver] Allow to set the content-type of String responses

martinbonnin:
A bunch of mockserver fixes:
* Allow to set the content-type of String responses
* Fix reading `content-length` (it should be case insensitive)
* More robust (albeit prone to false positive) handling of connection close


BoD (approved):
👍

martinbonnin:
I reverted the change to store headers lowercase as it could be surprising. As always, okhttp has a good API for [headers](https://github.com/square/okhttp/blob/10ccb88079514f9721692118679a994a0501e7ae/okhttp/src/jvmMain/kotlin/okhttp3/Headers.kt#L50) but I decided to not go all-in because it's a breaking change and we're not really a HTTP library. 

If we decide to make `mockserver` first class citizen, we should move it out of the repo and make the API better. Probably also share `HttpEngine` with the runtime.

BoD (approved):
👍

PR #5495, opened 2023-12-20
[IJ Plugin] Use recent version of slf4j to avoid a classloader issue

BoD:
Fixes:

```
java.lang.LinkageError: loader constraint violation: 
when resolving method 'org.slf4j.ILoggerFactory org.slf4j.impl.StaticLoggerBinder.getLoggerFactory()' the class loader 
com.intellij.ide.plugins.cl.PluginClassLoader @40ff4cd3 of the current class, org/slf4j/LoggerFactory, 
and the class loader 
com.intellij.ide.plugins.cl.PluginClassLoader @1280a557 for the method's defining class, org/slf4j/impl/StaticLoggerBinder, 
have different Class objects for the type org/slf4j/ILoggerFactory used in the signature 
(org.slf4j.LoggerFactory is in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @40ff4cd3, parent loader 'bootstrap'; 
org.slf4j.impl.StaticLoggerBinder is in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @1280a557, parent loader 'bootstrap') 
```

The slf4j version used by `org.xerial:sqlite-jdbc` was downgraded [here](https://github.com/xerial/sqlite-jdbc/pull/993). 

Although I'm not clear why this caused an issue for us, the IJ doc says _"A separate class loader is used to load the classes of each plugin. This allows each plugin to use a different library version, even if the same library is used by the IDE itself or by another plugin."_.


martinbonnin (approved):
"Fun with classloaders"!

PR #5497, opened 2023-12-21
Bump Kotlin to 2.0.0-Beta2

BoD:
Still missing the Compose compiler version from [here](https://androidx.dev/storage/compose-compiler/repository/), but doesn't seem to be a problem - or we can also wait more before merging.


martinbonnin (approved):

BoD:
> but doesn't seem to be a problem 

😬  I spoke too fast. Making this a draft then.

.gitignore
```diff
@@ -12,6 +12,7 @@
 # Generated files
 build
 /reports
+.kotlin
```
BoD, line 15: This seems to be new, some working files appear in `.kotlin`.

BoD: ah that's expected: https://kotlinlang.org/docs/whatsnew-eap.html#new-directory-for-kotlin-data-in-gradle-projects

martinbonnin (approved):

PR #5510, opened 2024-01-08
Dokka updates

martinbonnin:
- use dokkatoo
- use new branding
- fix CSS files (they changed at some point and our header is currently borked)
- add module doc

Screenshot 2024-01-08 at 10 55 11

Support for multiple versions is enabled but we don't have any previous kdoc for now. Plan is to store them in GCP or S2 and retrieve them during builds.

### DIFF
--- a/libraries/apollo-debug-server/README.md
+++ b/libraries/apollo-debug-server/README.md
@@ -1,0 +1,4 @@
+# Module apollo-debug-server
+
+apollo-debug-server is a server serves your normalized cache to the IntelliJ/Android Studio plugin. 
+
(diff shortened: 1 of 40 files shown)


libraries/apollo-debug-server/README.md
```diff
@@ -0,0 +1,4 @@
+# Module apollo-debug-server
+
+apollo-debug-server is a server serves your normalized cache to the IntelliJ/Android Studio plugin.
```
BoD, line 3:
```suggestion
apollo-debug-server is a server that serves your normalized cache to the IntelliJ/Android Studio plugin. 
```

PR #5512, opened 2024-01-08
Push Aggregate KDoc to MavenCentral in `apollo-kdoc` module

martinbonnin:
This way we can then retrieve the older versions to have them in the dropdown

### DIFF
--- a/build-logic/src/main/kotlin/Publishing.kt
+++ b/build-logic/src/main/kotlin/Publishing.kt
@@ -92,7 +92,7 @@ fun Project.configureDokka() {
   val project = this
   val kdocProject = project(":apollo-kdoc")
   kdocProject.configurations.all {
-    if (name == "dokkatoo") {
+    if (name == "dokkatoo" && project.name == "apollo-annotations") {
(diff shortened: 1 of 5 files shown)


martinbonnin:
I'll update this PR to push the aggregate doc to MavenCentral too.

build-logic/src/main/kotlin/Publishing.kt
```diff
@@ -80,25 +83,64 @@ fun Project.configureDokka() {
...
+  return dokkatoo
+}
+
+fun Project.configureDokka() {
+  configureDokkaCommon()
+  val project = this
+  val kdocProject = project(":apollo-kdoc")
+  kdocProject.configurations.all {
+    if (name == "dokkatoo" && project.name == "apollo-annotations") {
+      dependencies.add(kdocProject.dependencies.project(mapOf("path" to project.path)))
```
martinbonnin, line 96: Not sure how "project isolation compatible" this is. Problem for another day....

BoD (approved):
🎉

PR #5514, opened 2024-01-09
KDoc: Fix push CI job

martinbonnin:
I forgot to update the CI workflows


martinbonnin:
Landing quickly cause there's no other way to test

PR #5516, opened 2024-01-09
Make the compiler accept Json configuration files

martinbonnin:
The compiler can be used from different contexts:

* from the Gradle plugin (current)
* from a Gradle worker (needed for compiler APIs)
* from Maven

In order to make it easier to work in these different contexts, make the compiler accept JSON configuration files.

Instead of having conventions and default behaviours in the Gradle Plugin, they are now in the Compiler.


gradle/libraries.toml
```diff
@@ -144,7 +144,6 @@ kotlin-node = "org.jetbrains.kotlin-wrappers:kotlin-node:18.16.12-pre.634"
 kotlinx-serialization-plugin = { group = "org.jetbrains.kotlin", name = "kotlin-serialization", version.ref = "kotlin-plugin" }
 kotlinx-serialization-core = { group = "org.jetbrains.kotlinx", name = "kotlinx-serialization-core", version.ref = "kotlinx-serialization-runtime" }
 kotlinx-serialization-json = { group = "org.jetbrains.kotlinx", name = "kotlinx-serialization-json", version.ref = "kotlinx-serialization-runtime" }
-kotlinx-serialization-json-okio = { group = "org.jetbrains.kotlinx", name = "kotlinx-serialization-json-okio", version.ref = "kotlinx-serialization-runtime" }
```
martinbonnin, line 147: This is not stable so removing from the apollo-compiler API since users may start using the apollo-compiler API directly with this PR (some might have done so already but apollo-compiler was mostly used by the Gradle plugin so far)

libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -39,8 +39,31 @@ public final class com/apollographql/apollo3/compiler/CodegenMetadata$Companion
 
 public final class com/apollographql/apollo3/compiler/CodegenMetadataKt {
 	public static final fun schemaTypes (Lcom/apollographql/apollo3/compiler/CodegenMetadata;)Ljava/util/Set;
-	public static final fun toCodegenMetadata (Ljava/io/File;)Lcom/apollographql/apollo3/compiler/CodegenMetadata;
-	public static final fun writeTo (Lcom/apollographql/apollo3/compiler/CodegenMetadata;Ljava/io/File;)V
```
martinbonnin, line 43: Breaking changes everywhere in this file. I declare them OK cause the apollo-compiler API was not really meant to be accessed outside the Gradle Plugin.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -241,85 +282,40 @@ object ApolloCompiler {
...
-    val duplicates = irOperations.flatMapIndexed { index, irOperations1 ->
-      irOperations1.fragmentDefinitions.map { it.name to irOperationsFiles.get(index).path }
-    }.groupBy { it.first }.mapValues { it.value.map { it.second } }
-        .entries
-        .filter { it.value.size > 1 }
-        .flatMap { entry ->
-          entry.value.map {
-            "- ${entry.key}: $it"
-          }
-        }
```
martinbonnin, line 269: Turns out this was not really needed as the check is already done [here](https://github.com/apollographql/apollo-kotlin/blob/7c2c8eeb6421718bfa2ada8c7e025d86d8577977/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/ExecutableValidationScope.kt#L760). 

This means it's ok to have clashing fragments in sibling modules as long as they have different package names.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -118,18 +128,28 @@ object ApolloCompiler {
...
+    val scalarMapping = codegenSchemaOptions.scalarMapping
     checkScalars(schema, scalarMapping)
+
+    val codegenModels = codegenModels(codegenModels = codegenSchemaOptions.codegenModels, codegenSchemaOptions.targetLanguage)
+    @Suppress("NAME_SHADOWING")
+    val packageNameGenerator = packageNameGenerator ?: packageNameGenerator(
+        codegenSchemaOptions.packageName,
+        codegenSchemaOptions.packageNamesFromFilePaths,
+        packageNameRoots
+    )
```
martinbonnin, line 140: This is the prep work for moving `packageNameGenerator` to a `ServiceLoader`. We would still support `packageName` and `packageNamesFromFilePaths` as declarative options present in the JSON files that can be overriden by a compiler plugin if needed.

martinbonnin:
This should now be ready for review.  Initial description updated.

PR #5517, opened 2024-01-09
[IJ Plugin] Strip Apollo client directives before executing operations

### LINKED ISSUE #5507: 🧩 [IJ Plugin] Remove client only directives before sending the query to the server
Follow up from https://github.com/apollographql/apollo-kotlin/issues/5506

Right now, trying to execute a query that contains `@nonNull` or `@catch` fails because the server doesn't know the directive.

These directives are stripped in the compilation but this is an issue when running from the IDE plugin.

BoD:
Resolves #5507

https://github.com/apollographql/apollo-kotlin/assets/372852/fa0e2558-5eb7-4f65-ac35-1d5e44e47f6e


martinbonnin (approved):
👍

PR #5524, opened 2024-01-11
Lock down apollo-compiler API

martinbonnin:
Except for one tiny functional change, this is mostly moving things around & tweaking public visibility.

`ApolloCompiler` and all `FooOptions` are now public. 

Made some of `IrFoo` internal. Unfortunately, a lot has to be public for `ksp-processor` usage. Will dig more into this.

### DIFF
--- a/libraries/apollo-compiler/api/apollo-compiler.api
+++ b/libraries/apollo-compiler/api/apollo-compiler.api
@@ -1,6 +1,3 @@
-public final class com/apollographql/apollo3/compiler/-checkCapitalizedFields {
-}
-
 public abstract interface class com/apollographql/apollo3/compiler/AdapterInitializer {
 	public static final field Companion Lcom/apollographql/apollo3/compiler/AdapterInitializer$Companion;
 }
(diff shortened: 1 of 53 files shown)


libraries/apollo-compiler/api/apollo-compiler.api
```diff
@@ -9,17 +6,25 @@ public final class com/apollographql/apollo3/compiler/AdapterInitializer$Compani
 	public final fun serializer ()Lkotlinx/serialization/KSerializer;
 }
 
-public abstract interface class com/apollographql/apollo3/compiler/ApolloCompiler$Logger {
-	public abstract fun warning (Ljava/lang/String;)V
+public final class com/apollographql/apollo3/compiler/ApolloCompiler {
```
martinbonnin, line 9: Decided to make it public without an `@ApolloExperimental` marker. It's not like `apollo-compiler` has the same level of stability as `apollo-api`. We might need to break this at some point but I find this better than a bunch of noisy `@ApolloExperimental` annotations that would be basically required to use this API.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ApolloCompiler.kt
```diff
@@ -32,6 +31,13 @@ import com.apollographql.apollo3.compiler.codegen.java.JavaCodeGen
 import com.apollographql.apollo3.compiler.codegen.kotlin.KotlinCodeGen
 import com.apollographql.apollo3.compiler.hooks.ApolloCompilerJavaHooks
 import com.apollographql.apollo3.compiler.hooks.ApolloCompilerKotlinHooks
+import com.apollographql.apollo3.compiler.internal.addRequiredFields
+import com.apollographql.apollo3.compiler.internal.checkApolloInlineFragmentsHaveTypeCondition
+import com.apollographql.apollo3.compiler.internal.checkApolloReservedEnumValueNames
+import com.apollographql.apollo3.compiler.internal.checkApolloTargetNameClashes
+import com.apollographql.apollo3.compiler.internal.checkCapitalizedFields
+import com.apollographql.apollo3.compiler.internal.checkConditionalFragments
+import com.apollographql.apollo3.compiler.internal.checkKeyFields
```
martinbonnin, line 40: Moved to an internal package for the Java callers if any that don't read the Kotlin `internal` modifier.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -143,7 +140,6 @@ class CodegenSchemaOptions(
     val packageName: String? = null
 )
 
-@ApolloExperimental
```
martinbonnin, line 146: These are all the sam kind of symbols, make `ApolloCompiler` &`FooOptions` have all the same visibility

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/PackageNameGenerator.kt
```diff
@@ -25,13 +25,17 @@ interface PackageNameGenerator {
...
-  class Flat(val packageName: String) : PackageNameGenerator {
-    override fun packageName(filePath: String) = packageName
+  class Flat(private val packageName: String): PackageNameGenerator {
+    override fun packageName(filePath: String): String {
+      return packageName
+    }
+
     override val version: String
-      get() = "Flat-$packageName"
+      get() = error("this should only be called from the Gradle Plugin")
```
martinbonnin, line 34: This is the behaviour change

BoD (approved):
⚙️

PR #5535, opened 2024-01-15
Revive benchmarks again

martinbonnin:
We need to keep a low minSdk to run on the Firebase Test Labs devices.
Also move the benchmark logic to a Kotlin script outside of GitHub actions.


.github/workflows/benchmarks.yml
```diff
@@ -38,20 +38,10 @@ jobs:
         uses: google-github-actions/setup-gcloud@d51b5346f85640ec2aa2fa057354d2b82c2fcbce #v1.0.1
       - name: microbenchmarks
         uses: martinbonnin/run-benchmarks@ef9043b9a646a109f7381a4bf20f82ead5cbd382 #main
```
BoD, line 40: Need to call `run-benchmarks.main.kts` instead?

martinbonnin: Woops, right!

PR #5542, opened 2024-01-16
Invert memory vs SQL incubating benchmark

BoD:
This explains the surprising slowness of `CacheIncubatingTests.cache*Memory` and speed of `CacheIncubatingTests.cache*Sql` seen in #4231


martinbonnin (approved):
👍

PR #5547, opened 2024-01-19
Select all types in pre-introspection query

BoD:
`api.github.com` doesn't like querying too many introspection elements:

```graphql
{
  "errors": [
    {
      "type": "INTROSPECTION_LIMIT_EXCEEDED",
      "message": "Introspection fields may only be used 2 times, but some fields were used more than that: __Type.fields (5), __Field.args (5)"
    }
  ]
}
```

 We can be lenient in that case, and still try to execute the basic introspection query.

### DIFF
--- a/libraries/apollo-tooling/src/main/graphql/graphql/pre-introspection.graphql
+++ b/libraries/apollo-tooling/src/main/graphql/graphql/pre-introspection.graphql
@@ -1,14 +1,15 @@
 # Pre-introspection query allowing to retrieve the server's supported features.
 
 query PreIntrospectionQuery {
-  schema: __type(name: "__Schema") { ...TypeFields }
-  type: __type(name: "__Type") { ...TypeFields }
-  directive: __type(name: "__Directive") { ...TypeFields }
-  field: __type(name: "__Field") { ...TypeFields }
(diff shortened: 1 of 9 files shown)


martinbonnin (commented):
Can we report that to GitHub? This feels sketchy especially given a "regular" introspection query will return more data.

In the meantime, I would change the query to make GitHub happy:

```graphql
query PreIntrospectionQuery {
  __schema {
    types {
      ...TypeFields
    }
  }
}
```

This should reasonably work, right?

libraries/apollo-tooling/src/main/graphql/graphql/pre-introspection.graphql
```diff
@@ -1,14 +1,15 @@
...
-  schema: __type(name: "__Schema") { ...TypeFields }
-  type: __type(name: "__Type") { ...TypeFields }
-  directive: __type(name: "__Directive") { ...TypeFields }
-  field: __type(name: "__Field") { ...TypeFields }
-  inputValue: __type(name: "__InputValue") { ...TypeFields }
+  __schema {
+    types {
+      ...TypeFields
+    }
+  }
```
martinbonnin, line 8: Add a comment linking to this PR?

martinbonnin (approved):

PR #5552, opened 2024-01-22
[runtime] Fix disabling batching by default

martinbonnin:
See https://community.apollographql.com/t/operations-are-being-batched-even-when-batching-is-disabled-by-default/7145


BoD (approved):
👍

PR #5554, opened 2024-01-23
[compiler] Refactor CodegenLayout

martinbonnin:
_This is a refactoring change only, no functional change_

* Move `packageName` out of `CodegenSchema`. `packageName` is only used when writing the Java/Kotlin writing phase of compilation. 
* Simplify the `CodegenLayout` API. The package name layout is hardcoded as extension functions, there's no need to have it in the `CodegenLayout` API as it's never changed.
* Inline `escapeJavaReservedKeyword` usages. Those belong to the type builders where there is context about whether escaping is needed or not.
* Remove `escapeKotlinReserverdKeyword` as KotlinPoet is doing it.


BoD (approved):
👍

PR #5557, opened 2024-01-23
[compiler] Reorganize files

martinbonnin:
Reorganize files. Move builders to packages according to their purpose: `schema`, `operations`, `executableschema`. Also rename `CodeGen` => `Codegen` 

No functional change


BoD (approved):
👍

PR #5560, opened 2024-01-23
[compiler] Introduce JavaOutput, KotlinOutput, SourceOutput

martinbonnin:
`ApolloCompiler.xyz()` now returns `SourceOutput` that can be written to a `File` or to a KSP `CodeGenerator`. 

Make all serialization/deserialization functions public. The caller now has to call `writeTo()`/`toXyz()`. It's a bit more boilerplate all around but makes the `ApolloCompiler` API much more uniform.

Make `ApolloProcessor` use `ApolloCompiler`.


BoD (approved):
👍

PR #5562, opened 2024-01-24
Don't automatically add key fields to union selections

BoD:
Key fields were added to selections of unions, but that's not valid as unions have no fields (except `__typename`), and was crashing later on [here](https://github.com/apollographql/apollo-kotlin/blob/65ab70266f5c8363a20d200a3f16f7192e8e2e2b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/internal/checkKeyFields.kt#L85).

We can still check if selections have the key fields at least.

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
@@ -209,6 +209,7 @@ internal fun validateSchema(definitions: List<GQLDefinition>, requiresApolloDefi
 
   mergedScope.validateInterfaces()
   mergedScope.validateObjects()
+  mergedScope.validateUnions()
   mergedScope.validateInputObjects()
   mergedScope.validateCatch(mergedSchemaDefinition)
 
@@ -433,9 +434,30 @@ internal fun ValidationScope.validateRootOperationTypes(schemaDefinition: GQLSch
 }
 
 private fun ValidationScope.validateInterfaces() {
-  typeDefinitions.values.filterIsInstance<GQLInterfaceTypeDefinition>().forEach {
-    if (it.fields.isEmpty()) {
-      registerIssue("Interfaces must specify one or more fields", it.sourceLocation)
+  typeDefinitions.values.filterIsInstance<GQLInterfaceTypeDefinition>().forEach { i ->
+    if (i.fields.isEmpty()) {
+      registerIssue("Interfaces must specify one or more fields", i.sourceLocation)
+    }
+
+    i.implementsInterfaces.forEach { implementsInterface ->
+      val iface = typeDefinitions[implementsInterface] as? GQLInterfaceTypeDefinition
+      if (iface == null) {
+        registerIssue("Interface '${i.name}' cannot implement non-interface '$implementsInterface'", i.sourceLocation)
+      }
+    }
+
+    i.directives.forEach { directive ->
+      validateDirective(directive, i) {
+        issues.add(it.constContextError())
+      }
+    }
+
+    i.fields.forEach { gqlFieldDefinition ->
+      gqlFieldDefinition.directives.forEach { gqlDirective ->
+        validateDirective(gqlDirective, gqlFieldDefinition) {
+          issues.add(it.constContextError())
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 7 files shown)


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/internal/addRequiredFields.kt
```diff
@@ -75,16 +77,22 @@ private fun List<GQLSelection>.addRequiredFields(
...
     "always" -> isRoot
     else -> error("Unknown addTypename option: $addTypename")
   }
-  val requiredFieldNames = schema.keyFields(parentType).toMutableSet()
+  val requiredFieldNames = if (schema.typeDefinition(parentType) is GQLUnionTypeDefinition) {
+    // Can't select any fields on unions other than __typename
+    mutableSetOf()
+  } else {
+    schema.keyFields(parentType).toMutableSet()
+  }
```
martinbonnin, line 95: Should we allow `@typePolicy` on unions in the first place?

```
// This feels out of place. SomeUnion has no `id` field.
extend union SomeUnion @typePolicy(keyFields: "id") 
```

BoD: Yeah I was wondering about this :) We could forbid it too, it might be less confusing.

martinbonnin: For the record, we talked about this and decided to keep the directive definition untouched but add validation that the `keyFields` actually exists, which should make an error here.

BoD: Turns out this validation already existed but wasn't performed on interfaces nor unions.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/SchemaValidationScope.kt
```diff
@@ -433,9 +434,30 @@ internal fun ValidationScope.validateRootOperationTypes(schemaDefinition: GQLSch
...
+    i.directives.forEach { directive ->
+      validateDirective(directive, i) {
+        issues.add(it.constContextError())
+      }
+    }
+
+    i.fields.forEach { gqlFieldDefinition ->
+      gqlFieldDefinition.directives.forEach { gqlDirective ->
+        validateDirective(gqlDirective, gqlFieldDefinition) {
+          issues.add(it.constContextError())
```
martinbonnin, line 458: FWIW, this is where having a visitor pattern for validation would be useful. Maybe not right now but at some point.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo3/ast/internal/ValidationCommon.kt
```diff
@@ -241,15 +236,50 @@ internal fun ValidationScope.extraValidateTypePolicyDirective(directive: GQLDire
...
+    (connectionFieldsArg.value as GQLStringValue).value.parseAsGQLSelections().getOrThrow().forEach { selection ->
+      if (selection !is GQLField) {
+        registerIssue("Fragments are not supported in @$TYPE_POLICY directives", connectionFieldsArg.sourceLocation)
+      } else if (selection.selections.isNotEmpty()) {
+        registerIssue("Composite fields are not supported in @$TYPE_POLICY directives", connectionFieldsArg.sourceLocation)
+      } else {
+        val definition = fieldDefinitions.firstOrNull { it.name == selection.name }
+        if (definition == null) {
+          registerIssue("Field '${selection.name}' is not a valid connection field for type '$type'", connectionFieldsArg.sourceLocation)
+        }
```
martinbonnin, line 282: 👍  
Also, can this logic be factored in? Looks like it's very similar?

libraries/apollo-compiler/src/test/kotlin/com/apollographql/apollo3/compiler/keyfields/KeyFieldsTest.kt
```diff
@@ -125,15 +115,9 @@ class KeyFieldsTest {
...
-      checkKeyFields(operation, schema, emptyMap())
-      fail("An exception was expected")
-    } catch (e: SourceAwareException) {
-      check(e.message!!.contains("Field 'id' is not a valid key field for type 'Animal'"))
-      return
-    }
+    val issues = GQLDocument(schemaDefinitions, null).validateAsSchema().issues
+    check(issues.any { it.message.contains("Field 'id' is not a valid key field for type 'Animal'") })
+    check(issues.any { it.message.contains("Field 'version' is not a valid key field for type 'Node'") })
+    check(issues.any { it.message.contains("Field 'x' is not a valid key field for type 'Foo'") })
```
martinbonnin, line 121: Nit: I'd change the error to something like 

```
"No such field: 'Foo.x'"
```

I think there's no need to mention "key" since it's easy to see from the context and sourceLocation.

martinbonnin (changes requested):

martinbonnin (approved):

PR #5563, opened 2024-01-25
[tests] Keep test directories around

martinbonnin:
I always forget to uncommit that line and it doesn't hurt to keep directories around. Just do it inconditionally.


BoD (approved):
👍

PR #5573, opened 2024-01-29
Introduce SchemaLayout & OperationsLayout

martinbonnin:
(_Apologies for the initial title, clicked "submit" a bit too early_)

Introduce `SchemaLayout` and `OperationsLayout` that are specialized versions so that someone building the server code doesn't need to specify operations/fragments package names.

Also remove default values from options constructors. Given the number of options, it's way too easy to forget passing some. Instead, add `buildFoo()` functions that can be used for users of the `ApolloCompiler` API


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -168,208 +176,312 @@ class IrOptions(
...
+   * The target language to use to generate the sources
+   */
+  val targetLanguage: TargetLanguage?
+  /**
+   * Whether to decapitalize fields. This is useful if your schema has fields starting with an uppercase as
+   * it may create name clashes with models that use PascalCase
+   */
+  val decapitalizeFields: Boolean?
+  /**
+   * Whether to decapitalize fields. This is useful if your schema has fields starting with an uppercase as it may create name clashes
```
BoD, line 223:
```suggestion
   * Whether to suffix operation name with 'Query', 'Mutation' or 'Subscription'
```

BoD (approved):
👍

PR #5579, opened 2024-01-29
[🐘gradle-plugin] bump minimum required Gradle version

martinbonnin:
[AGP 8 requires Gradle 8](https://developer.android.com/build/releases/gradle-plugin) so go ahead and bump that version for Apollo Kotlin. It's not stricly required but moves the ecosystem and since v4 is going to stick around, will give us a bit of headroom.


BoD (approved):
👍

PR #5580, opened 2024-01-29
[🐘gradle-plugin] configuration cache and lazy properties for schema files

### LINKED ISSUE #5221: Implicit task dependency not working
### Version

3.8.2

### Summary

My team has a library that among other things, stitches a bunch of separate schema files into one.

We then take this generated schema file and feed it into the apollo plugin via a Provider. Typically a provider will create a implicit dependency on the task. 

Depending on which task I run I get:
```
Execution failed for task ':generateTest-serviceApolloSources'.
> java.io.FileNotFoundException: /Users/me/repros/apollo-kotlin-implicit-dep/build/newSchema/test.graphql
```

and sometimes it will show:
```
A problem was found with the configuration of task
'generateTestApolloSources' (type
'ApolloGenerateSourcesTask').
    
    Reason: Task
':generateTestApolloSources' uses this
output of task ':internalTask'
without declaring an explicit or implicit dependency. This can lead to
incorrect results being produced, depending on what order the tasks are
executed.
 
    
    For more information, please refer to
https://docs.gradle.org/8.2/userguide/validation_problems.html#implicit_dependency
in the Gradle documentation.
```

I've excluded some text just to avoid leaking anything, but you probably get the idea.

I believe the apollo plugin should be able to handle providers properly so that we don't need to explicitly declare the dependency.

Currently we need to do (pseudo code)
```
generateTestApolloSources dependsOn internalTask
generateTestApolloSchema dependsOn internalTask
generateTestApolloUsedCoordinates dependsOn internalTask
```

Thank you for you time and please let me know what you think!

I created a repro to show how it works.

You can't do much without the apollo plugin immediately failing.

https://github.com/dronda-t/apollo-implicit-dep-repro 

### Logs

_No response_

### LINKED ISSUE #5455: Build fails after schema file rename when Gradle configuration cache enabled
### Version

3.8.2 and 4.0.0-beta.3

### Summary

Currently build fails if you rename schema file and configuration cache was enabled. Take a look at the steps to reproduce issue below.

**My current workaround**
I have little bit investigated this and it seems the problem in the following code (schema files searched in a non-compatible way with configuration cache)
https://github.com/apollographql/apollo-kotlin/blob/84f0f9253765ff4bfd8f223539fefd24e8da0488/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt#L965-L970

To workaround this, I set `Service.schemaFiles` directly instead of relying on default behavior.
https://github.com/apollographql/apollo-kotlin/blob/84f0f9253765ff4bfd8f223539fefd24e8da0488/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/api/Service.kt#L83

```
apollo {
    service("main") {
        srcDir("src/main/graphql")
        schemaFiles.setFrom(
            fileTree("src/main/graphql") {
                include("**/*.graphqls")
            }
        )
    }
}
```

PS: since in my project I have multiple schema files, I cannot use just `Service.schemaFile` for single file

**Steps**
1) Create simple `my_lib` module and use default configuration for Apollo plugin. Place GQL schema file into main sources: `/Users/user/projects/sample/my_lib/src/main/graphql/schema.graphqls`
2) Build module with configuration cache enabled `./gradlew :my_lib:assembleDebug --configuration-cache`
3) Rename `schema.graphqls` to `schema2.graphqls`
4) Build module again with configuration cache enabled `./gradlew :my_lib:assembleDebug --configuration-cache`

**Actual result**
The second build fails with error
```
> Task :my_lib:generateMainApolloSources FAILED
Execution failed for task ':my_lib:generateMainApolloSources'.
> java.io.FileNotFoundException: /Users/user/projects/sample/my_lib/src/main/graphql/schema.graphqls (No such file or directory)

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
BUILD FAILED in 307ms
122 actionable tasks: 6 executed, 116 up-to-date
Configuration cache entry reused.
```

**Expected result**
Build must success

### Logs

_No response_

martinbonnin:
Rework the schema wiring to use `FileCollections` everywhere. 

Fixes #5455 
Fixes #5221

There is still an issue that `FileCollections` do not distinguish between "unset" and "empty". See https://github.com/gradle/gradle/issues/21752 so in the event that someone misconfigures their `schemaFiles` option, it might still fallback to `src/main/graphql/**.graphql` but oh well...


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -57,7 +56,7 @@ abstract class DefaultApolloExtension(
   internal fun getServiceInfos(project: Project): List<ApolloGradleToolingModel.ServiceInfo> = services.map { service ->
     DefaultServiceInfo(
         name = service.name,
-        schemaFiles = service.lazySchemaFiles(project),
+        schemaFiles = service.schemaFilesSnapshot(project),
```
martinbonnin, line 59: I never realized that but there is potential for failure here. Is `getServiceInfos()` depending on other tasks? If `schemaFiles` is generated by another task vs being always commited in the repo, the IJ plugin might see an absent file.

BoD: This is called directly inside `ApolloPlugin.apply`, after the creation of `ApolloExtension` - so not depending on any tasks. Not sure what we could do to make this scenario work though 🤔

BoD (approved):

PR #5582, opened 2024-01-29
[⚙️compiler] introduce InputFile

martinbonnin:
Our `packageNamesFromFilePaths()` require the normalized path (as in [PathSensitivy.RELATIVE](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/PathSensitivity.html#RELATIVE) normalization) in order to compute the package name.

I tried working around this for a while but I came to the conclusion that the proper solution is to make the normalized path a proper input to the compiler. Users that don't need to support `packageNamesFromFilePaths()` (like tests) can always use `files.toInputFiles()` to get a list of input files with empty normalized paths.


BoD (approved):
🐘

PR #5588, opened 2024-01-31
Introduce `SchemaLayout`, `OperationsLayout` and `ExecutableSchemaLayout`

martinbonnin:
Instead of the all-in-one codegen layout, expose specialized interfaces to different codegen packages. 

* `codegen.[java|kotlin].schema` uses `SchemaLayout`
* `codegen.[java|kotlin].executableschema` uses `ExecutableSchemaLayout`
* `codegen.[java|kotlin].operations` uses `OperationsLayout`

For convenience, `LayoutImpl` is unchanged and implements all 3 interfaces but it's internal only and doesn't leak in public API.

`ApolloCompiler.buildFoo()` now have 2 kind of options:

1. the "declarative" ones (`IrOptions`, `CodegenOptions`, etc...) are scalars that can be serialized in JSON option files.
1. the "turing complete" ones (`Layout`, `OperationOutputGenerator`, `ApolloCompilerHooks`, etc...) are code that are serialized as classpaths.

Note: this temporarily breaks passing a custom `packageNameGenerator` but turns out we have no test for this. I'll re-introduce that in a follow up PR


BoD (approved):
⚙️

PR #5591, opened 2024-02-02
[⚙️compiler] Add KDoc + remove hooks + API tweaks

martinbonnin:
Add KDoc + remove hooks + API tweaks. Especially, add a default implementation for `PackageNameGenerator.version` and `OperationOutputGenerator.version` so that implementors do not have to override it.


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
```diff
@@ -226,18 +229,85 @@ interface CommonCodegenOpt {
...
   val useSemanticNaming: Boolean?
   /**
-   * Which methods to auto generate on models, fragments, operations, and input objects
+   * Specifies which methods will be auto generated on operations, models, fragments and input objects.
+   *
+   * Pass a list of the following:
+   *
+   * - "equalsHashCode" generates `equals` and `hashCode` methods that will compare generated class properties
+   * - "toString" generates a method that will print a pretty string representing the data in the class
+   * - "copy" (Kotlin only) generates a method that will copy the class with named parameters for
```
BoD, line 252: Never noticed before, but this sentence is missing a few words?

BoD (approved):
🚀

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/Options.kt
martinbonnin: Added in https://github.com/apollographql/apollo-kotlin/pull/5591/commits/0246deff91d3e4785e6aad3a924a1ab6915936e2

PR #5599, opened 2024-02-05
[⚙️compiler] Remove `OperationOutputGenerator` from the `Plugin` API

martinbonnin:
I thought I did it already but turns out I didn't. 

This also allows to deprecate PackageNameGenerator and OperationOutputGenerator. All functionality can now be done with `Plugins`


BoD (approved):
👍

PR #5600, opened 2024-02-05
Publish the IJ plugin snapshots to the JetBrain Marketplace

BoD:
Decommission Repsy.

See also: https://github.com/apollographql/shortlinks/pull/32


.github/workflows/publish-ij-plugin-snapshot.yml
```diff
@@ -19,18 +19,11 @@ jobs:
...
-      - name: Increment snapshot version
-        run: |
-          git config user.name github-actions
-          git config user.email github-actions@github.com
-          ./scripts/increment-ij-plugin-snapshot.main.kts
-        env:
-          IJ_PLUGIN_SNAPSHOT: true
-          GH_TOKEN: ${{ github.token }}
-
       - name: Publish snapshot to Repsy
```
martinbonnin, line 22:
```suggestion
      - name: Publish snapshot to Jetbrains snapshots
```

intellij-plugin/build.gradle.kts
```diff
@@ -24,8 +25,8 @@ repositories {
 
 group = properties("pluginGroup")
 
-// Use the global version defined in the root project + snapshot suffix if from the CI
-version = properties("VERSION_NAME") + if (isSnapshotBuild()) ".${properties("snapshotVersion")}" else ""
+// Use the global version defined in the root project + dedicated suffix if building a snapshot from the CI
+version = properties("VERSION_NAME") + if (isSnapshotBuild()) ".${SimpleDateFormat("YYYY-MM-dd").format(Date())}" else ""
```
martinbonnin, line 29: Could we use the git commit instead or does that need to increment monotonically?

intellij-plugin/README.md
```diff
@@ -44,7 +44,7 @@ The plugin is published every Sunday in a specific **snapshots** repository. To
 configured:
 
 <kbd>Settings</kbd> > <kbd>Plugins</kbd> > <kbd>⚙</kbd>️ > <kbd>Manage Plugin
-Repositories</kbd> > <kbd>+</kbd> > <kbd>https://raw.githubusercontent.com/apollographql/apollo-kotlin/main/intellij-plugin/snapshots/plugins.xml</kbd>
+Repositories</kbd> > <kbd>+</kbd> > <kbd>https://go.apollo.dev/ij-plugin-snapshots</kbd>
```
martinbonnin, line 47: Reminder to call this out in the Changelog

intellij-plugin/build.gradle.kts
BoD: I'm pretty sure the publication date is used to check if a more recent version is used, so the git commit should work - but are you sure that'd be better? A date may be a bit more straightforward/may ease communication if someone reports an issue?

martinbonnin: Why not both? :)

martinbonnin: (cut the sha1 at 7 chars, like GH is doing to avoid adding too much entropy in this world?)

intellij-plugin/README.md
BoD: Good point! Also the deletion of the `snapshots/plugins.xml` file is not ideal, but it should at least show an error in the repo UI, letting users know that it's no longer valid.

intellij-plugin/build.gradle.kts
BoD: Ah both, good idea :) 👍

intellij-plugin/README.md
martinbonnin: I think we can use the [`next-docs` branch](https://github.com/apollographql/apollo-kotlin/pull/5578) to start drafting the changelog. Want to put an entry there?

BoD: perfect 👍

PR #5604, opened 2024-02-06
[⚙️compiler] Open up IR and add integration test for IR transform

martinbonnin:
Made `IrOperations` public (but `@ApolloExperimental` and they will probably stay that way a very long time)

Also add a test for IR transforms that flattens to a max depth of 3.


BoD (approved):
🚀

PR #5605, opened 2024-02-06
[🐘gradle-plugin] Fix a regression in alwaysGenerateTypesMatching

martinbonnin:
We were setting the default to `.*` for leaf modules which doesn't really make sense since the types are only generated in the schema module but had the unfortunate side effect of generating all types.


BoD (approved):
👍

PR #5606, opened 2024-02-06
[🐘 gradle-plugin] Add overload for dependsOn that tries to do cross project configuration

martinbonnin:
This is so that adding `dependsOn`:

```kotlin
apollo {
  service("service") {
    // ..
    dependsOn(project(":schema"))
  }
}
```

automatically adds the matching `isADependencyOf()`

This is all highly experimental but it's working in integration tests.

Also add more KDoc.


BoD (approved):
👍

Issue #4605, opened 2023-01-05
Memory cache lock contention


ynkm169:
### Use case

Hello, 
`private val lock = CacheLock()`

MemoryCache uses a lock that can be problematic. With Java concurrent data structures, lock stripping/lock splitting would make data structures such as ConcurrentHashMap very efficient. Android LRU cache internally hopefully is pretty efficient as well. 

However, with a synchronized lock like this, effectively the whole LRU cache is locked up allowing only 1 reader/writer at once.

Did some systracing. Launched 50 duplicate API calls, result shows there are around 60 RxCachedThreadS and 60 DefaultDispatch threads all contending for 1 monitor lock. 
<img width="1123" alt="Screen Shot 2023-01-04 at 6 15 10 PM" src="https://user-images.githubusercontent.com/1810298/210673974-c7f5b01a-5328-44a2-a0d4-8441b4048239.png">

### Describe the solution you'd like

Question
Is it possible to remove this lock all together and have some other work around? Possibly even sacrificing consistency for performance?

BoD:
Thank you for providing this, this is interesting data, and will dig into it.

Just a quick reaction to 
> it is not uncommon for dev to call an API to retrieve a list of items, and use flatmap to make a seperate API call for each item in the list

Not sure if you are using GraphQL already (if yes you can disregard my comment), but one of its advantage when compared to REST is the ability to avoid this, by having the fields you need at your disposal to be used in one query. Of course the ability to do it will depend on the server's schema, but in many cases you should be able to avoid some round trips.

ynkm169:
> Not sure if you are using GraphQL already
Yes, to provide more context, REST and GraphQL will co-exist for quite a while for us. Migration will take a lot of time.

ynkm169:
I wasn't sure if it is possible to build a LRU cache that's more performant than a synchronized LinkedHashMap. I Looks like [Guava cache](https://github.com/google/guava/blob/67350a341c5cfd5e9bebf86ba87687ab7ef602b2/guava/src/com/google/common/cache/LocalCache.java) is potentially more performant (hard to tell since Guava cache maybe better used for server than low power mobile devices. 

> The Least Recently Used page replacement algorithm was chosen due to its simplicity.
> The basic strategy is to subdivide the table among Segments, each of which itself is a
>    * concurrently readable hash table. The map supports non-blocking reads and concurrent writes
>    * across different segments.

BoD:
A few additional notes:
[...]

PR #5608, opened 2024-02-07
Cache lock changes

BoD:
The idea: instead of having a lock in `ApolloStore`, let caches handle their own locking, in the hope that this would reduce overall contention (see #4605).

Things to note:
- Each cache still uses a lock to at least ensure consistency between the different caches in the chain
- The access to SQLite appears to be guarded by a lock on its own, so no locking is needed in general. A lock is still used only in case there is a `nextCache`
- In the MemoryCache, the lock guards writes to the LRUCache and access to the next cache
- In incubating only

Added micro-benchmarks that start a few threads performing read and writes through ApolloStore in parallel (memory / sql / memory then sql).

It _looks like_ the bench performance is improved when comparing the results before/after the change (avg of 6 runs on a Samsung S10+):

|             | Memory     | Sql         | Memory then sql |
|-------------|------------|-------------|-----------------|
| Before      | 50,565,308 | 801,852,416 | 1,182,789,288   |
| After       | 45,090,843 | 708,763,083 | 900,559,012     |
| Improvement | 10.8%      | 11.6&       | 23.9%           |

### A few flame charts

Memory, before:
incubating-before-10-100-memory

Memory, after:
incubating-after-10-100-memory

Sql, before:
incubating-before-10-100-sql

Sql, after:
incubating-after-10-100-sql

### DIFF
--- a/benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
+++ b/benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
@@ -1,0 +1,98 @@
+package com.apollographql.apollo3.benchmark
+
+import androidx.benchmark.junit4.BenchmarkRule
+import androidx.benchmark.junit4.measureRepeated
+import androidx.test.platform.app.InstrumentationRegistry
+import com.apollographql.apollo3.api.json.jsonReader
+import com.apollographql.apollo3.api.parseJsonResponse
+import com.apollographql.apollo3.benchmark.Utils.dbName
+import com.apollographql.apollo3.benchmark.Utils.operationBasedQuery
+import com.apollographql.apollo3.benchmark.Utils.resource
+import com.apollographql.apollo3.benchmark.test.R
+import com.apollographql.apollo3.cache.normalized.incubating.ApolloStore
+import com.apollographql.apollo3.cache.normalized.incubating.api.CacheKeyGenerator
+import com.apollographql.apollo3.cache.normalized.incubating.api.CacheResolver
+import com.apollographql.apollo3.cache.normalized.incubating.api.FieldPolicyCacheResolver
+import com.apollographql.apollo3.cache.normalized.incubating.api.MemoryCacheFactory
+import com.apollographql.apollo3.cache.normalized.incubating.api.NormalizedCacheFactory
+import com.apollographql.apollo3.cache.normalized.incubating.api.TypePolicyCacheKeyGenerator
+import com.apollographql.apollo3.cache.normalized.incubating.sql.SqlNormalizedCacheFactory
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import java.lang.reflect.Method
+import java.util.concurrent.Executors
+
+class ApolloStoreIncubatingTests {
+  @get:Rule
+  val benchmarkRule = BenchmarkRule()
+
+  @Test
+  fun concurrentReadWritesMemory() {
+    concurrentReadWrites(MemoryCacheFactory())
+  }
+
+  @Test
+  fun concurrentReadWritesSql() {
+    Utils.dbFile.delete()
+    // Pass context explicitly here because androidx.startup fails due to relocation
+    val cacheFactory = SqlNormalizedCacheFactory(InstrumentationRegistry.getInstrumentation().context, dbName)
+    concurrentReadWrites(cacheFactory)
+  }
+
+  @Test
+  fun concurrentReadWritesMemoryThenSql() {
+    Utils.dbFile.delete()
+    val cacheFactory = MemoryCacheFactory().chain(SqlNormalizedCacheFactory(InstrumentationRegistry.getInstrumentation().context, dbName))
+    concurrentReadWrites(cacheFactory)
+  }
+
+  private fun concurrentReadWrites(cacheFactory: NormalizedCacheFactory) {
+    val apolloStore = createApolloStore(cacheFactory)
+    val query = operationBasedQuery
+    val data = query.parseJsonResponse(resource(R.raw.calendar_response_simple).jsonReader()).data!!
+    val threadPool = Executors.newFixedThreadPool(CONCURRENCY)
+    benchmarkRule.measureRepeated {
+      val futures = (1..CONCURRENCY).map {
+        threadPool.submit {
+          // Let each thread execute a few writes/reads
+          repeat(WORK_LOAD) {
+            apolloStore.writeOperation(query, data)
+            val data2 = apolloStore.readOperation(query)
+            Assert.assertEquals(data, data2)
+          }
+        }
+      }
+      // Wait for all threads to finish
+      futures.forEach { it.get() }
+    }
+  }
+
+  private fun createApolloStore(cacheFactory: NormalizedCacheFactory): ApolloStore {
+    return createApolloStoreMethod.invoke(
+        null,
+        cacheFactory,
+        TypePolicyCacheKeyGenerator,
+        FieldPolicyCacheResolver,
+    ) as ApolloStore
+  }
+
+
+  companion object {
+    private const val CONCURRENCY = 10
+    private const val WORK_LOAD = 5
+
+    /**
+     * There doesn't seem to be a way to relocate Kotlin metadata and kotlin_module files so we rely on reflection to call top-level
+     * methods
+     * See https://discuss.kotlinlang.org/t/what-is-the-proper-way-to-repackage-shade-kotlin-dependencies/10869
+     */
+    private val apolloStoreKtClass = Class.forName("com.apollographql.apollo3.cache.normalized.incubating.ApolloStoreKt")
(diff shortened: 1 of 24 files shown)


benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
```diff
@@ -0,0 +1,98 @@
...
+    val threadPool = Executors.newFixedThreadPool(CONCURRENCY)
+    benchmarkRule.measureRepeated {
+      val futures = (1..CONCURRENCY).map {
+        threadPool.submit {
+          // Let each thread execute a few writes/reads
+          repeat(WORK_LOAD) {
+            apolloStore.writeOperation(query, data)
+            val data2 = apolloStore.readOperation(query)
+            Assert.assertEquals(data, data2)
+          }
```
martinbonnin, line 63: Can you also add an integration test, that tests end to end using `apolloClient.query()` and the default dispatcher? (should be `Dispatchers.Default` IIRC). Are the results similar?

benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.Executors
+
+class ApolloStoreIncubatingTests {
+  @get:Rule
+  val benchmarkRule = BenchmarkRule()
+
+  @Test
+  fun concurrentReadWritesMemory() {
+    concurrentReadWrites(MemoryCacheFactory())
+  }
```
martinbonnin, line 33: Are those uploaded to datadog automagically using the [`run-benchmarks` script](https://github.com/apollographql/apollo-kotlin/blob/b5faf3cfe02989620f95730b1e68d5db361599d2/scripts/run-benchmarks.main.kts#L5) ? I would say so but I can't remember.

libraries/apollo-normalized-cache-api-incubating/src/appleMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/-cache-lock-apple.kt
```diff
@@ -1,5 +0,0 @@
-package com.apollographql.apollo3.cache.normalized.api.internal
-
-internal actual class CacheLock actual constructor() {
-  actual fun <T> lock(block: () -> T): T = block()
-}
```
martinbonnin, line 5: 👍

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/OptimisticCache.kt
```diff
@@ -14,115 +14,130 @@ import kotlin.reflect.KClass
...
   override fun loadRecord(key: String, cacheHeaders: CacheHeaders): Record? {
-    return try {
-      val nonOptimisticRecord = nextCache?.loadRecord(key, cacheHeaders)
-      nonOptimisticRecord.mergeJournalRecord(key)
-    } catch (ignore: Exception) {
-      null
+    return lock.read {
+      try {
+        val nonOptimisticRecord = nextCache?.loadRecord(key, cacheHeaders)
+        nonOptimisticRecord.mergeJournalRecord(key)
```
martinbonnin, line 24: Should this be a `lock.write {}`? (I expect `mergeJournalRecord` to modify data)

martinbonnin: Mmm nevermind, the returned `nonOptimisticRecord` isn't shared so all good 👍

libraries/apollo-normalized-cache-api-incubating/build.gradle.kts
```diff
@@ -15,6 +15,9 @@ kotlin {
         api(project(":apollo-mpp-utils"))
         implementation(libs.okio)
         api(libs.uuid)
+        implementation(libs.atomicfu.get().toString()) {
+          because("Use of ReentrantLock for Apple (we don't use the gradle plugin rewrite)")
```
martinbonnin, line 19: Might be worth actually using the plugin to remove our own `com.apollographql.apollo3.cache.normalized.api.internal.Lock` and make lock a no-op on JS (and probably save a few instructions on JVM/Native)?

libraries/apollo-normalized-cache-sqlite-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/SqlNormalizedCache.kt
```diff
@@ -17,66 +18,101 @@ class SqlNormalizedCache internal constructor(
...
-      recordDatabase.select(key)
-    } catch (e: Exception) {
-      // Unable to read the record from the database, it is possibly corrupted - treat this as a cache miss
-      apolloExceptionHandler(Exception("Unable to read a record from the database", e))
-      null
-    }
-    if (record != null) {
-      if (cacheHeaders.hasHeader(EVICT_AFTER_READ)) {
-        recordDatabase.delete(key)
+  // A lock is only needed if there is a nextCache
```
martinbonnin, line 21: Are we 100% certain of that? SQLDelight queries being thread safe no matter what underlying driver is used?

martinbonnin (commented):
looks good! I like that it's not changing any API so it's a net gain. 
I'm curious what the integration tests results are going to be.

benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
BoD: Do you mean still in a microbenchmark or something else?

martinbonnin: I think so? Same thing but closer to the real scenario of executing a query. Probably using MockWebServer or so. I know it adds more variance but it's also what users are actually doing.

benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
BoD: I _think_ so from what I read in the script 😅.

libraries/apollo-normalized-cache-sqlite-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/SqlNormalizedCache.kt
BoD: Not 100% certain alas. At least on native (MacOS) I made a small test that starts 2 threads, 1st starts a transaction, 2nd one also, I could see that the 2nd one was blocked until the 1st one finished. I am not sure if this is due to a lock somewhere in the driver, or if SQLite itself handles this. What I think I understand from [here](https://sqlite.org/isolation.html) is that SQLite handles this itself, but that's not 100% clear to me tbh.

benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
BoD: I'd say yes (and also yes for DataDog) Doesn't hurt to track it :) I'll have a look at the DD conf.

libraries/apollo-normalized-cache-api-incubating/build.gradle.kts
BoD: Actually the plugin is only useful for atomic, which we don't use (yet!), whereas we need locks to ensure consistency between the caches in the chain.

benchmark/microbenchmark/src/androidTest/java/com/apollographql/apollo3/benchmark/ApolloStoreIncubatingTests.kt
BoD: Added in c8cee0f74690b5fd78f3bd9b4bd76f2884f37548 benches that execute queries in parallel. The results are a bit questionable:

|             | Memory     | Sql           | Memory then sql |
|-------------|------------|---------------|-----------------|
| Before      | 44,201,994 | 1,047,540,153 | 41,335,067      |
| After       | 40,916,743 | 1,079,325,820 | 38,458,753      |
| Improvement | 7.4%       | -3%           | 6.9%            |

martinbonnin (approved):

PR #5610, opened 2024-02-07
Keep rx-support as DeprecationLevel.Error

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/5609, keep the symbols around until the next major to help with the migration

### DIFF
--- a/libraries/apollo-rx2-support/src/main/java/com/apollographql/apollo3/rx2/-Rx2Extensions.kt
+++ b/libraries/apollo-rx2-support/src/main/java/com/apollographql/apollo3/rx2/-Rx2Extensions.kt
@@ -1,0 +1,29 @@
+package com.apollographql.apollo3.rx2
+
+import com.apollographql.apollo3.ApolloCall
+import com.apollographql.apollo3.annotations.ApolloDeprecatedSince
+import com.apollographql.apollo3.api.ApolloResponse
+import com.apollographql.apollo3.api.Operation
+import io.reactivex.Flowable
+import io.reactivex.Scheduler
+import io.reactivex.Single
+import io.reactivex.annotations.CheckReturnValue
+import io.reactivex.schedulers.Schedulers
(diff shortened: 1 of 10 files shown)


libraries/apollo-rx2-support/src/main/java/com/apollographql/apollo3/rx2/-Rx2Extensions.kt
```diff
@@ -0,0 +1,29 @@
...
+import io.reactivex.Scheduler
+import io.reactivex.Single
+import io.reactivex.annotations.CheckReturnValue
+import io.reactivex.schedulers.Schedulers
+import kotlinx.coroutines.rx2.asCoroutineDispatcher
+import kotlinx.coroutines.rx2.asFlowable
+
+@JvmSynthetic
+@CheckReturnValue
+@Deprecated("use kotlinx-coroutines-rx2 directly", ReplaceWith("toFlow().asFlowable(scheduler.asCoroutineDispatcher()).firstOrError()"), level = DeprecationLevel.ERROR)
```
BoD, line 17: I like the `ERROR` level, I think it's a good compromise. Maybe we can also add a link to the migration guide:

```suggestion
/**
 * Since version 4 this wrapper should no longer be used. See also [the migration guide](https://www.apollographql.com/docs/kotlin/v4/migration/4.0#using-rxjava-artifacts-is-now-an-error).
 */
@Deprecated("use kotlinx-coroutines-rx2 directly", ReplaceWith("toFlow().asFlowable(scheduler.asCoroutineDispatcher()).firstOrError()"), level = DeprecationLevel.ERROR)
```

martinbonnin: I'd like to go through short links in error message so we can redirect easily if needed. I'll make a global pass at adding links once the documentation is in good shape.

PR #5621, opened 2024-02-13
Add more multipart tests

martinbonnin:
Add a test for `"payload": null`, see https://github.com/apollographql/router/issues/4634.

Also, do not crash on malformed multipart bodies and return Apollo.exception instead.


BoD (approved):
👍

martinbonnin:
Marking as ready for review matching https://github.com/apollographql/team-prometheus/blob/8804024bd7804f774caeb2dc8562297db5145432/rfcs/MultipartSubscriptionsTests.md#mapping-parts-to-client-events

This version is lenient to `payload: null`. We might want to fine tune more and refine the semantics of some messages but for the time being it should be an improvement of the existing:

* errors in the multipart protocol themselves now trigger proper `response.exception`
* unknown fields are now properly ignored.

libraries/apollo-api/api/apollo-api.api
```diff
@@ -989,6 +989,10 @@ public final class com/apollographql/apollo3/api/http/internal/UrlEncodeKt {
 	public static final fun urlEncode (Ljava/lang/String;)Ljava/lang/String;
 }
 
+public final class com/apollographql/apollo3/api/internal/ResponseParserKt {
+	public static final fun readErrors (Lcom/apollographql/apollo3/api/json/JsonReader;)Ljava/util/List;
```
BoD, line 993: Should it be `@ApolloInternal`?

BoD (approved):
👍

libraries/apollo-api/api/apollo-api.api
martinbonnin: I thought it could be useful in general but you're right, easier to remove the `@ApolloInternal` later if needed than add it.

PR #5623, opened 2024-02-14
[compiler] Add support for @catch on fieldDefinitions, interfaces and objects

martinbonnin:
Implementation of https://github.com/apollographql/specs/pull/48
See https://github.com/apollographql/specs/issues/47

`@catch` can now be set, by order or precedence:

At the schema level

```graphql
// All fields are modeled as FieldResult in Kotlin
extend schema @catch(to: RESULT)
```

At the field definition level

```graphql
type User {
  // email is modeled as FieldResult in Kotlin
  email @catch(to: RESULT)
}

// or if you don't own the schema
extend type User @catchField(name: "email", to: RESULT)
```

At the field level

```graphql
query GetEmail {
  user {
    // email is modeled as FieldResult in Kotlin for this query only
    email @catch(to: RESULT)
  }
}
```

Also took this opportunity to factor in the `@semanticNonNullField` handling in schema merging so that it's done in a central place.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -67,6 +67,8 @@
       </indentOptions>
     </codeStyleSettings>
     <codeStyleSettings language="kotlin">
+      <option name="CALL_PARAMETERS_WRAP" value="0" />
+      <option name="CALL_PARAMETERS_LPAREN_ON_NEXT_LINE" value="false" />
       <indentOptions>
         <option name="INDENT_SIZE" value="2" />
         <option name="CONTINUATION_INDENT_SIZE" value="4" />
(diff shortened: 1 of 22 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -67,6 +67,8 @@
       </indentOptions>
     </codeStyleSettings>
     <codeStyleSettings language="kotlin">
+      <option name="CALL_PARAMETERS_WRAP" value="0" />
+      <option name="CALL_PARAMETERS_LPAREN_ON_NEXT_LINE" value="false" />
```
martinbonnin, line 71: Not sure why we need this now. I'm attributing this to latest EAP but something to keep an eye on

BoD (approved):
👍

martinbonnin:
Ping @rohandhruva . If everything goes well, this should end up in the SNAPSHOTs soon.

rohandhruva:
Thank you so much, I'll keep an eye out for it!

PR #5624, opened 2024-02-15
Bump Kotlin to 2.0.0-Beta4

BoD:


martinbonnin (approved):
Noice 🚀
Billion dollar question: does it work with K2?

BoD:
> does it work with K2?

I think almost there but:
- 🔴 [KT-21846](https://youtrack.jetbrains.com/issue/KT-21846) is not resolved yet and chokes on `actual typealias ApolloTestResult = Promise<Any>`
- 🟠 [KT-65105](https://youtrack.jetbrains.com/issue/KT-65105/K2-Error-Member-overrides-different-Throws-filter) is not resolved yet, but a simple workaround is to not override `Executable` methods in `Operation` and `Fragment`
- 🟠 [KT-60741](https://youtrack.jetbrains.com/issue/KT-60741) is either back or has never been resolved. There's also a simple workaround but I'll try to create a minimum reproducer.
- 🟢 Need to bump `kotlin-plugin-min = "1.9.0"` for certain gradle tests to pass

BoD:
Actually, KT-21846 is [KT-60561](https://youtrack.jetbrains.com/issue/KT-60561) which has a workaround that works 🎉. With that I have integration tests passing on K2 ✅.

martinbonnin (approved):

PR #5625, opened 2024-02-16
Fix incorrect file name in tutorial documentation step 6

molundb:
There is no file called `LaunchListAdapter.kt`. It should be `LaunchList.kt`.


martinbonnin (approved):
Thanks!

Issue #5627, opened 2024-02-19
Allow Custom Extensions in buildPostBody Method


morux2:
### Use case

We are trying to verify queries by passing signatures via extensions instead of using AutoPersistedQuery. As mentioned in https://github.com/apollographql/apollo-kotlin/issues/2922, we tried using the [ByteStringHttpBody](https://www.apollographql.com/docs/kotlin/kdoc/apollo-api/com.apollographql.apollo3.api.http/-byte-string-http-body/index.html) method. However, there is no way to serialize variables since FileUploadAwareJsonWriter is  internal. We would like to call the original implementation as much as possible to keep up with future updates of apollo-kotlin.

<img width="600" alt="スクリーンショット 2024-02-19 16 02 49" src="https://github.com/apollographql/apollo-kotlin/assets/19369161/6690d788-3e67-4234-b877-fedc753fa8e5">

apollo-kotlin: 3.8.2

### Describe the solution you'd like

add extensions argument to [buildPostBody](https://www.apollographql.com/docs/kotlin/kdoc/apollo-api/com.apollographql.apollo3.api.http/-default-http-request-composer/-companion/build-post-body.html) method.

<img width="437" alt="スクリーンショット 2024-02-19 16 09 51" src="https://github.com/apollographql/apollo-kotlin/assets/19369161/3e45582d-35f1-49e6-bc2b-01381fdddf80">

martinbonnin:
Makes a lot of sense 👍 Thanks for the feature request.

An version of `buildPostBody` that allows writing extensions is available [here](https://github.com/apollographql/apollo-kotlin/pull/5630) (will backport in v3 in a bit).

If you are not doing [file uploads](https://www.apollographql.com/docs/kotlin), you can use the following for the time being:

```kotlin
class WithExtensionsHttpRequestComposer(private val serverUrl: String) : HttpRequestComposer {
  override fun <D : Operation.Data> compose(apolloRequest: ApolloRequest<D>): HttpRequest {

    val request = HttpRequest.Builder(HttpMethod.Post, serverUrl)
        .body(
            ByteStringHttpBody(
                "application/json",
                buildJsonByteString {
                  writeObject {
                    name("query")
                    value(apolloRequest.operation.document())
                    name("operationName")
                    value(apolloRequest.operation.name())
                    name("variables")
                    apolloRequest.operation.serializeVariables(this, apolloRequest.executionContext[CustomScalarAdapters]!!, false)
                    name("extensions")
[...]

PR #5630, opened 2024-02-20
[🏃runtime] Allow buildPostBody to write operation extensions

martinbonnin:
See #5627

Allow custom `HttpRequestComposer`s to write extensions


BoD (approved):
👍

PR #5636, opened 2024-02-21
[🐘 gradle-plugin] Use Gradle normalization instead of ours

martinbonnin:
We don't need internal `sourceRoots` as `FileCollection` rightfully includes the normalized path. 

This simplifies the code quite a bunch but is unfortunately blocked on https://github.com/gradle/gradle/issues/28147

Also this is a behaviour change for users using both `packageNameFromFilePaths` and `schemaFile` that can be mitigated by using `schemaFiles` instead:

```kotlin
    packageNameFromFilePaths()

    // replace
    schemaFile.set("src/main/graphql/com/example/schema.graphqls")

    // with
    schemaFiles.from(fileTree("src/main/graphql/").apply {
      include("com/example/schema.graphqls")
    })
```


martinbonnin:
I found a [workaround](https://github.com/apollographql/apollo-kotlin/pull/5636/commits/3412a6e31f3bc02eb39ae4a32dd9f6fef5f116df)

BoD (approved):
🐘

PR #5642, opened 2024-02-22
[🤖 CI] Upload diagnostics on push

martinbonnin:
The file name was wrong. The diagnostics.zip file is helpful to investigate CI failures


martinbonnin:
Merging quickly as it's the only way to test

PR #5644, opened 2024-02-22
[CI] disable CC when publishing to a remote repository.

martinbonnin:
Unbreaks pushing snapshots and kdoc

See https://youtrack.jetbrains.com/issue/KT-65879


BoD (approved):
👍

PR #5651, opened 2024-02-23
Use Store Cache and merge optimistic cache with Memory cache

BoD:
Replace our in-house LRUCache that required guarding with a lock with the `Cache` from the [Store project](https://github.com/MobileNativeFoundation/Store) (which is a KMP port of Guava's Cache).

Benchmarks after this change looked promising, but we still used a lock upstream inside `OptimisticCache` which reduced these gains.

To improve on that, this PR also merges the optimistic cache with the memory cache, and uses a thread-safe map (backed by `ConcurrentHashMap` on the JVM and a regular map behind a lock on other platforms) to store the `RecordJournal`s.

| CacheIncubatingTests | cacheOperationMemory | cacheResponseMemory | concurrentCacheOperationMemory | concurrentCacheResponseMemory |
|----------------------|----------------------|---------------------|--------------------------------|-------------------------------|
| Before               | 87,918,019           | 82,278,491          | 210,223,423                    | 213,267,240                   |
| After                | 80,766,616           | 80,221,426          | 42,095,053                     | 39,544,423                    |
| Improvement          | 8.13%                | 2.50%               | 79.98%                         | 81.46%                        |

| ApolloStoreIncubatingTests | concurrentReadWritesMemory |
|----------------------------|----------------------------|
| Before                     | 45,964,577                 |
| After                      | 18,559,416                 |
| Improvement                | 59.62%                     |

| CacheIncubatingIntegrationTests | concurrentQueriesTestNetworkTransportMemory |
|---------------------------------|---------------------------------------------|
| Before                          | 47,139,692                                  |
| After                           | 38,322,750                                  |
| Improvement                     | 18.70%                                      |

A few changes to get there:
- `NormalizedCache` is now an interface, and chaining is removed (`MemoryCache` still supports chaining)
- `OptimisticNormalizedCache` is now an interface, implemented by `MemoryCache` and `OptimisticNormalizedCacheWrapper` (new name for what `OptimisticCache` was)

The Store cache was copied to the project, but I'll see if we can depend on it instead - I had to make one addition: the  `getAllPresent()` fun, also the wasm target is missing - so opening in draft for now.


libraries/apollo-normalized-cache-api-incubating/src/jsMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/ConcurrentMap.js.kt
```diff
@@ -0,0 +1,3 @@
+package com.apollographql.apollo3.cache.normalized.api.internal
+
+internal actual fun <K, V> ConcurrentMap(): MutableMap<K, V> = CommonConcurrentMap()
```
martinbonnin, line 3: No need to pay the wrapper price?

```suggestion
internal actual fun <K, V> ConcurrentMap(): MutableMap<K, V> = mutableMapOf()
```

libraries/apollo-normalized-cache-api-incubating/src/wasmJsMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/ConcurrentMap.wasmJs.kt
```diff
@@ -0,0 +1,3 @@
+package com.apollographql.apollo3.cache.normalized.api.internal
+
+internal actual fun <K, V> ConcurrentMap(): MutableMap<K, V> = CommonConcurrentMap()
```
martinbonnin, line 3: No need to pay the wrapper price?

```suggestion
internal actual fun <K, V> ConcurrentMap(): MutableMap<K, V> = mutableMapOf()
```

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/OptimisticNormalizedCacheWrapper.kt
```diff
@@ -30,49 +29,39 @@ class OptimisticCache : NormalizedCache() {
...
+      var result: Boolean = wrapped.remove(cacheKey, cascade)
 
       val recordJournal = recordJournals[cacheKey.key]
       if (recordJournal != null) {
         recordJournals.remove(cacheKey.key)
         result = true
         if (cascade) {
           for (cacheReference in recordJournal.current.referencedFields()) {
-            result = result && remove(CacheKey(cacheReference.key), true)
+            result = result || remove(CacheKey(cacheReference.key), true)
```
martinbonnin, line 64: I was going to say this looks like it needs to be backported to non-incubating but sounds like the whole semantics of the return value need to be refined here, right?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/RecordWeigher.kt
```diff
@@ -23,9 +22,9 @@ internal object RecordWeigher {
 
   @JvmStatic
   fun calculateBytes(record: Record): Int {
-    var size = SIZE_OF_RECORD_OVERHEAD + record.key.utf8Size().toInt()
+    var size = SIZE_OF_RECORD_OVERHEAD + record.key.length
     for ((key, value) in record.fields) {
-      size += key.utf8Size().toInt() + weighField(value)
+      size += key.length + weighField(value)
```
martinbonnin, line 27: What's the reasoning for changing the weighter implementation?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/MemoryCache.kt
```diff
@@ -18,141 +20,242 @@ import kotlin.reflect.KClass
...
  */
 class MemoryCache(
+    private val nextCache: NormalizedCache? = null,
     private val maxSizeBytes: Int = Int.MAX_VALUE,
     private val expireAfterMillis: Long = -1,
-) : NormalizedCache() {
-  private val lock = Lock()
+) : OptimisticNormalizedCache {
+  // A lock is only needed if there is a nextCache
+  private val lock = nextCache?.let { Lock() }
```
martinbonnin, line 28: Do we still need our own `Lock`? Atomic fu says: ["JVM it is a typealias to the later class, erased on JS."](https://github.com/Kotlin/kotlinx-atomicfu?tab=readme-ov-file#locks)

tests/pagination/src/commonTest/kotlin/OffsetBasedWithPagePaginationTest.kt
```diff
@@ -233,13 +233,11 @@ class OffsetBasedWithPagePaginationTest {
 }
 
 internal fun assertChainedCachesAreEqual(apolloStore: ApolloStore) {
-  val hasNextCache = atomic(false)
-  apolloStore.accessCache { cache ->
-    // First cache is always OptimisticCache
-    hasNextCache.value = cache.nextCache!!.nextCache != null
+  val dump = apolloStore.dump().filterKeys {
+    // Ignore optimistic cache for comparison
+    it != OptimisticNormalizedCache::class && it != OptimisticNormalizedCacheWrapper::class
```
martinbonnin, line 238: What is the reason those might be diferent?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/RecordWeigher.kt
BoD: Yes sorry this is a totally unrelated to this PR! I just noticed this and thought in memory strings are stored either in utf-16 or some sort of compact form that's iso-latin-1 if possible (apparently [since JDK9](https://www.baeldung.com/java-9-compact-string) - I can't tell if this applies to Android). I think it's never utf-8. But `length*2` may be a better value if we think it's probably utf-16. 

And now the fun question: what is it on apple/js/wasm? 😅

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/MemoryCache.kt
martinbonnin: I remember that discussion now! 👍

tests/pagination/src/commonTest/kotlin/OffsetBasedWithPagePaginationTest.kt
BoD: `MemoryCache` uses `OptimisticNormalizedCache::class` as a key in its `dump()` fun. Maybe not the most intuitive thing?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/OptimisticNormalizedCacheWrapper.kt
BoD: yeah the `&&` didn't make much sense to me but if so the doc should also be updated to be precise:
```diff
- * @return `true` if a record with such key was successfully removed, `false` otherwise
+ * @return `true` if at least one record was successfully removed, `false` otherwise
```

tests/pagination/src/commonTest/kotlin/OffsetBasedWithPagePaginationTest.kt
martinbonnin: Gotcha 👍  Yea, not the best API. Add it to the list of APIs to bikeshed?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/OptimisticNormalizedCacheWrapper.kt
BoD: Not yet - I could open a dedicated ticket

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/internal/store/LocalCache.kt
```diff
@@ -0,0 +1,1977 @@
...
+        // eviction
+        private fun enqueueNotification(entry: ReferenceEntry<K, V>, cause: RemovalCause?) {
+            enqueueNotification(entry.key, entry.hash, entry.valueReference, cause)
+        }
+
+        private fun enqueueNotification(
+            key: K?,
+            hash: Int,
+            valueReference: ValueReference<K, V>?,
+            cause: RemovalCause?
```
martinbonnin, line 909: This gives me a bunch of "unused variables" warning. Can you add a suppression?

BoD:
Merging this for now. I've opened #5689 to circle back if we can use Store as a dependency instead of including it.

PR #5652, opened 2024-02-23
[3.x] Publish apollo-kdoc in 3.x too 

martinbonnin:
This way we can have both 3.x and 4.x kdocs at https://www.apollographql.com/docs/kotlin/kdoc/index.html


BoD (approved):
🎉

PR #5657, opened 2024-02-27
Migrate to org.jetbrains.intellij.platform v2.0.0

### LINKED ISSUE #5647: [IJ Plugin] Try the new IntelliJ Platform Gradle Plugin 2.0
- Doc: https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html
- Migration guide is here: https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-migration.html

BoD:
Resolves #5647.

Keeping as draft for now, while this is beta..

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Fetch Sources
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
 
-      # Setup Java 11 environment for the next steps
+      # Setup Java environment for the next steps
       - name: Setup Java
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 #v4.2.1
         with:
@@ -124,7 +124,7 @@ jobs:
           echo "::set-output name=version::$VERSION"
           echo "::set-output name=pluginVerifierHomeDir::~/.pluginVerifier"
 
-          ./gradlew :intellij-plugin:listProductsReleases # prepare list of IDEs for Plugin Verifier
+          ./gradlew :intellij-plugin:printProductsReleases # prepare list of IDEs for Plugin Verifier
 
       # Trigger code generation in test project, as the plugin's tests depend on it
       - name: Test project codegen
@@ -151,7 +151,7 @@ jobs:
 
       # Run Verify Plugin task and IntelliJ Plugin Verifier tool
       - name: Run Plugin Verification tasks
-        run: ./gradlew :intellij-plugin:runPluginVerifier -Pplugin.verifier.home.dir=${{ steps.properties.outputs.pluginVerifierHomeDir }}
+        run: ./gradlew :intellij-plugin:verifyPlugin -Pplugin.verifier.home.dir=${{ steps.properties.outputs.pluginVerifierHomeDir }}
 
       # Collect Plugin Verifier Result
       - name: Collect Plugin Verifier Result
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ package-lock.json
 
 # IntelliJ plugin
 intellij-plugin/mockJDK
+.intellijPlatform
--- a/.idea/runConfigurations/Run_IntelliJ_plugin.xml
+++ b/.idea/runConfigurations/Run_IntelliJ_plugin.xml
@@ -16,7 +16,7 @@
       </option>
       <option name="taskNames">
         <list>
-          <option value=":intellij-plugin:runIde" />
+          <option value=":intellij-plugin:runLocalIde" />
         </list>
       </option>
       <option name="vmOptions" />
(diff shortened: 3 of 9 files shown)


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -353,7 +352,7 @@ abstract class DefaultApolloExtension(
     return project.tasks.register(ModelNames.checkApolloVersions()) {
       val outputFile = BuildDirLayout.versionCheck(project)
 
-      it.inputs.property("allVersions", Callable {
+      it.inputs.property("allVersions", project.provider {
```
martinbonnin, line 355: Interesting... Was the issue that the callable is called eagerly?

BoD: Exactly! (Credits to @hsz 🙏). However this works locally but just noticed we still got the `ConcurrentModificationException` on the CI 😭

martinbonnin: IIRC `intellij-plugin` uses the published version of the plugin so we won't be able to tell before the next release if the fix actually worked.

hsz: We have snapshots. 🤫

BoD: > `intellij-plugin` uses the published version of the Gradle plugin

You're right! And my mistake - thought it was working locally but it's only because I ran with `--no-configuration-cache` :(

martinbonnin: I'd say the concurrent modification happens on the `configurations` collection because we're iterating all of them but `addAllLater` ([here](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/d95a86476b108a2899143586b987282e5dec0cc3/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt#L208)) may create new configurations while iterating ([here](https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/d95a86476b108a2899143586b987282e5dec0cc3/src/main/kotlin/org/jetbrains/intellij/platform/gradle/extensions/IntelliJPlatformDependenciesHelper.kt#L216))

Proposed fix to iterate a read-only copy of the configurations:

```diff
     private fun getDeps(configurations: ConfigurationContainer): List<String> {
-      return configurations.flatMap { configuration ->
+      // see https://github.com/apollographql/apollo-kotlin/pull/5657
+      val currentConfigurations = configurations.toList()
+      return currentConfigurations.flatMap { configuration ->
         configuration.dependencies
```

Means we won't check the newly created configurations for conflicting Apollo dependencies but it's probably fine.

martinbonnin: Of course this also requires a new release so just disabling the task inside `intellij-plugin` for now is the quickest solution.

martinbonnin: @hsz sorry, messages crossed. Regarding SNAPSHOTs, it's about Apollo releases. AKIJP (Apollo Kotlin IntelliJ Plugin) uses AKGP (Apollo Kotlin Gradle Plugin) . We could verify using Apollo snapshots but means we'd need a separate PR to check the plugin-ception 🙃 . 

All in all, my vote goes to disabling the check 😄

hsz: Yeah, I figured too late. Anyway, I'm around if you need me!

BoD: @martinbonnin unfortunately disabling the task doesn't work - the exception is thrown at configuration time - not 100% sure why, but looks like `getDeps` ends up being called during `org.gradle.configurationcache.ConfigurationCacheIO.writeConfigurationCacheState`.

martinbonnin: Mmmpff... Disable CC in that CI job?

BoD: Haha that should work! And I'll open an PR for your suggested fix above, so it can be in the next release, hopefully fixing the issue.

martinbonnin (approved):

hsz:
Congrats!

BoD:
@hsz Congrats to you! And thanks for all the support! 🙏

PR #5661, opened 2024-02-28
Suppress warnings

martinbonnin:
Suppress a bunch of warnings and hide a few Gradle symbols that should have been hidden from day1.

Most of the warnings were due to our own usage of deprecated APIs so it was mostly about adding `@Suppress`


BoD (approved):
👍

PR #5669, opened 2024-02-29
Clarify ApolloResponse exception vs errors

BoD:


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -10,12 +10,17 @@ import kotlin.jvm.JvmField
...
- * - data != null && exception != null: partial data with field errors
+ * - [data] contains the parsed data returned in the response
+ * - [errors] contains any GraphQL errors returned in the response
+ * - [exception] contains an exception if a GraphQL response wasn't received
+ *
+ * GraphQL responses can be partial responses, so it is valid to have both [data] != null and [errors] != null. The valid states are:
+ * - exception == null && data != null && errors == null: complete data with no error
+ * - exception == null && data != null && errors != null: partial data with field errors
+ * - exception == null && data == null && errors != null: no data, only field errors
+ * - exception != null && data == null && errors == null: no GraphQL response was received due to a network error or otherwise
```
martinbonnin, line 23: I think the case below can also happen today (it's probably not spec compliant but it's possible to forge a server that returns no data and no error)

```
* - exception == null && data == null && errors == null
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -10,12 +10,17 @@ import kotlin.jvm.JvmField
...
  *
- * Valid states are:
- * - data != null && exception == null: complete data with no error
- * - data == null && exception != null: no data, a network error or operation error happened
- * - data != null && exception != null: partial data with field errors
+ * - [data] contains the parsed data returned in the response
+ * - [errors] contains any GraphQL errors returned in the response
+ * - [exception] contains an exception if a GraphQL response wasn't received
+ *
+ * GraphQL responses can be partial responses, so it is valid to have both [data] != null and [errors] != null. The valid states are:
```
martinbonnin, line 19: Nit: 

```suggestion
 * GraphQL responses can contain partial data, so it is possible to have both [data] != null and [errors] != null. An [ApolloResponse] may be in any of these states:
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
BoD: I'm tempted to not highlight it since it should not happen in normal cases, devs should really not bother handling it. Wdyt?

martinbonnin: KDoc is about precision, not brevity. I think it's definitely ok to skip it in the docs but I would expect it in the KDoc.

BoD: Fair enough!

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -10,12 +10,18 @@ import kotlin.jvm.JvmField
...
- * - data == null && exception != null: no data, a network error or operation error happened
- * - data != null && exception != null: partial data with field errors
+ * - [data] contains the parsed data returned in the response
+ * - [errors] contains any GraphQL errors returned in the response
+ * - [exception] contains an exception if a GraphQL response wasn't received
+ *
+ * GraphQL responses can contain partial data, so it is possible to have both [data] != null and [errors] != null. An [ApolloResponse] may be in any of these states:
+ * - exception == null && data != null && errors == null: complete data with no errors
+ * - exception == null && data != null && errors != null: partial data with field errors
+ * - exception == null && data == null && errors != null: no data, only field errors
```
martinbonnin, line 22:
```suggestion
 * - exception == null && data != null && errors != null: partial data with GraphQL errors
 * - exception == null && data == null && errors != null: no data, only GraphQL errors
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt
```diff
@@ -10,12 +10,18 @@ import kotlin.jvm.JvmField
 import kotlin.jvm.JvmName
 
 /**
- * Represents a GraphQL response. GraphQL responses can be partial responses, so it is valid to have both data != null and exception != null
+ * Represents a GraphQL response.
```
martinbonnin, line 13: In retrospect, `ApolloResponse` could have been `ApolloResult` or so. I think this ship has sailed but we can be more precise about what `ApolloResponse` is here:

```suggestion
 * Represents a GraphQL response or an exception if no GraphQL response is received.
```

martinbonnin (approved):

PR #5670, opened 2024-03-01
Fix JS websocket throws an ISE on error on Safari and Firefox

baconz:
We're seeing the error event fire when the websocket closes as a result of a network error on Safari and Firefox. Since the socket has already been opened, the coroutine has already been resumed and this throws an IllegalStateException ("Already resumed").

I can submit this same patch up to ktor, since I think that's where this code came from originally.

### DIFF
--- a/libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt
+++ b/libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt
@@ -101,26 +101,40 @@ actual class DefaultWebSocketEngine : WebSocketEngine {
   private suspend fun WebSocket.awaitConnection(): WebSocket = suspendCancellableCoroutine { continuation ->
     if (continuation.isCancelled) return@suspendCancellableCoroutine
 
-    val eventListener = { event: Event ->
-      when (event.type) {
-        "open" -> continuation.resume(this)
-        "error" -> {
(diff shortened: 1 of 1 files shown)


BoD (approved):
Looks good to me 👍 Thanks!

martinbonnin (approved):
Good catch! 

Also [looks like that code is still in Ktor indeed](https://github.com/ktorio/ktor/blob/14c25646d52169101c4ccd5a052b38bf1993d967/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt#L119), a PR there would be great!

martinbonnin:
PS: how hard would it be to add a test for that? I _think_ KGP can do firefox tests with KarmaN

baconz:
I can take a stab at adding a test over the weekend!

baconz:
Hmm. I'm afraid I'm a bit stumped on this one. The challenge is that you need a websocket server running outside of the browser that dies (uncleanly) after a connection has been established.

I think the easiest way to do this would be to have a little node script that runs the websocket server, which you start up beforehand (eg a shell-based gradle task with a `dependsOn`), and then kill it on a timer. That feels a little brittle, but I think it would work.

Can you think of a better way? I found this [very old Karma plugin](https://github.com/weblogixx/karma-websocket-server), but I don't see an easy way to modify `KotlinKarma` to allow for custom plugins.

martinbonnin:
Urg, right. No mock server in the browser :/. Not sure it's worth the extra complexity then, especially since the fix is solid. I'm merging this as is as we can revisit browser tests that need a MockServer next time.

PR #5671, opened 2024-03-01
Use the streaming HttpEngine by default on Apple

### LINKED ISSUE #5659: Use StreamingNSURLSessionHttpEngine by default on Apple
`StreamingNSURLSessionHttpEngine` was introduced when working on the `@defer` support. It needed the new native memory manager to work, which is why we kept it as a separate implementation instead of modifying the existing `DefaultHttpEngine`.

Now that the NMM is used by default, we can make the switch.

BoD:
Resolves #5659


libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt
```diff
@@ -320,3 +303,20 @@ private class Pipe {
...
+private class DefaultDataTaskFactory : DataTaskFactory {
+  private var nsUrlSession: NSURLSession? = null
+
+  override fun dataTask(request: NSURLRequest, delegate: NSURLSessionDataDelegateProtocol): NSURLSessionDataTask {
+    if (nsUrlSession == null) nsUrlSession = NSURLSession.sessionWithConfiguration(
+        configuration = NSURLSessionConfiguration.defaultSessionConfiguration(),
+        delegate = delegate,
+        delegateQueue = null
+    )
+    return nsUrlSession!!.dataTaskWithRequest(request)
```
martinbonnin, line 320: This is prone to race conditions and also surprising because `delegate` would only be taken into account for the first call. What about

```kotlin
fun interface DataTaskFactory {
  fun dataTask(request: NSURLRequest): NSURLSessionDataTask
}

private class DefaultDataTaskFactory(private val delegate) : DataTaskFactory {
  // ...
}
```

libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt
```diff
@@ -48,33 +48,16 @@ import platform.posix.pthread_mutex_lock
...
- * as multiple threads mutate the internal state.
- *
- * That is why it is considered experimental and subject to change.
- *
- * Using the default [DefaultHttpEngine] will work when using `@defer`, but the payloads are buffered until the response is fully
- * received, thus losing the incremental nature of the response.
- */
-@ApolloExperimental
-class StreamingNSURLSessionHttpEngine(
+actual class DefaultHttpEngine(
```
martinbonnin, line 51: Can you remove the expect/actual classes to use top level constructor functions instead? (Can be made in a separate PR)

libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt
martinbonnin: Also more general comment about `fun interface`. We don't have many in the codebase and I think I'd prefer not adding too many. Regular interfaces are probably easier to evolve if we need to

martinbonnin: Ah but wait, we need to control the delegate so looks like we need a factory of factory instead 🙈

libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt
BoD: Will do it in the next PR (also to implement `Closeable` to replace `dispose()`)

libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt
```diff
@@ -320,3 +314,19 @@ private class Pipe {
...
+  private val nsUrlSession = NSURLSession.sessionWithConfiguration(
+      configuration = NSURLSessionConfiguration.defaultSessionConfiguration(),
+      delegate = delegate,
+      delegateQueue = null
+  )
+
+  override fun dataTask(request: NSURLRequest): NSURLSessionDataTask {
+    return nsUrlSession.dataTaskWithRequest(request)
+  }
+}
```
martinbonnin, line 332: Thinking about this more, do we even need `DataTaskFactory`? Can't we "just" have 

```
actual class DefaultHttpEngine : HttpEngine {
  constructor(
      timeoutMillis: Long = 60_000,
      configuration: NSURLSessionConfiguration?
  ) {
```

What is the use case for changing anythign else than the configuration?

BoD: I think we want to allow people to have a global `NSURLSession` for their app, like we do on the JVM with constructors that accept a `OkHttpClient` or `Call.Factory`.

But honestly it's hard to tell if it's actually useful without being an iOS dev 😅

martinbonnin: > I think we want to allow people to have a global NSURLSession for their app, like we do on the JVM with constructors that accept a OkHttpClient or Call.Factory.

Given we force the delegate, I don't think the `NSURLSession` can be reused by any other parts of the app so we might as well own it (also we're talking about closing it ourselves so if it is shared we def. don't want to close it and we'd have to refcount and what not)

Also looks like the iOS client allocates its own `URLsession`: https://github.com/apollographql/apollo-ios/blob/b434fb904b4695bcc84e83ca1cac132321bf5435/Sources/Apollo/URLSessionClient.swift#L64

BoD: Good points, yeah that sounds about right. Simply passing a `NSURLSessionConfiguration` is definitely a welcome simplification 👍

PR #5677, opened 2024-03-04
Don't assume a single emission in AutoPersistedQueryInterceptor

### LINKED ISSUE #5667: Experimental @defer support does not work with AutoPersistedQueryInterceptor
### Version

4.0.0-beta.4

### Summary

When using `@defer`, a `response` will have multiple parts. However, `AutoPersistedQueryInterceptor` uses `var response = chain.proceed(request).single()`, which will lead to an immediate `IllegalArgumentException`, because: `Flow has more than one element`. 

Enable the `AutoPersistedQueryInterceptor` and use any operation that has `@defer`. The failure should reproduce even if you use `client.query().toFlow()`. 

### Logs

_No response_

BoD:
Resolves #5667


martinbonnin (approved):
👍 
There's a bunch of "interesting" edge cases if the `PersistedQueriesNotFound` is followed by other items in the original `Flow`. I don't think we need to be too concerned but mentioning just in case you think of something?

BoD:
> There's a bunch of "interesting" edge cases if the PersistedQueriesNotFound is followed by other items in the original Flow. I don't think we need to be too concerned but mentioning just in case you think of something?

Yeah I think it should "never" happen (with quotes because odd things tend to happen 😅). Worst case, it will re-query with the body even if it's not strictly necessary - I think that's OK.

martinbonnin (approved):

PR #5678, opened 2024-03-04
Introduce incubating WebSocketNetworkTransport

martinbonnin:
This PR introduces an incubating `WebSocketNetwork`. Disclaimer: this is 100% untested. It's not finished yet but it's relatively self contained so I'm making a separate PR. 

This PR introduces incubating:

* `WebSocketNetworkTransport`: a `NetworkTransport` that is a lot more simple than the non-incubating one. The `execute()` method in particular is just a `callbackFlow`
* `WebSocketHolder`: a class responsible to hold the websocket
* `SubscribableWebSocket`: the glue between `WebSocket` and `WsProtocol`. This is where most of the state machine is
* `WsProtocol` and implementations for AppSync, GraphqlWs, etc.. This is mostly unchanged from the non-incubating version.

Missing:

* retry logic (~= reopenWhen)
* suspend serverUrl and headers for AppSync or other clients that want to change the server url
* connectivity indicators (is my socket connected?)
* lots of tests

### DIFF
--- a/libraries/apollo-api/api/apollo-api.api
+++ b/libraries/apollo-api/api/apollo-api.api
@@ -1290,9 +1290,13 @@ public final class com/apollographql/apollo3/exception/RouterError : com/apollog
 	public final fun getErrors ()Ljava/util/List;
 }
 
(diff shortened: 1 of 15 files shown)


BoD (approved):
✅

PR #5685, opened 2024-03-05
Add `ApolloClient.retryOnError(Boolean)`

martinbonnin:
So that took a while but I'm quite happy with this solution. It's a lot more flexible that the previous WebSocket only and in the end it's just a few lines of code meaning someone wanting to add custom logic shouldn't need copy/pasting too many things.

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/RetryOnErrorInterceptor.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/RetryOnErrorInterceptor.kt
@@ -1,8 +1,10 @@
 package com.apollographql.apollo3.interceptor
 
+import com.apollographql.apollo3.annotations.ApolloExperimental
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.Operation
+import com.apollographql.apollo3.api.Subscription
 import com.apollographql.apollo3.exception.ApolloException
 import com.apollographql.apollo3.exception.SubscriptionOperationException
 import com.benasher44.uuid.uuid4
@@ -15,8 +17,21 @@ import kotlinx.coroutines.flow.retryWhen
 import kotlin.math.pow
 import kotlin.time.Duration.Companion.seconds
 
-object RetryOnErrorInterceptor: ApolloInterceptor {
+@ApolloExperimental
+class RetryOnErrorInterceptor(shouldIntercept: ((ApolloRequest<*>) -> Boolean)? = null): ApolloInterceptor {
+  private val shouldIntercept: ((ApolloRequest<*>) -> Boolean) = shouldIntercept ?: {
+    when (it.retryNetworkErrors) {
+      true -> true
+      false -> false
+      null -> it.operation is Subscription
+    }
+  }
+
   override fun <D : Operation.Data> intercept(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
+    if (!shouldIntercept(request)) {
+      return chain.proceed(request)
+    }
+
     var first = true
     var attempt = 0
     return flow {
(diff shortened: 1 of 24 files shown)


tests/websockets/src/jvmTest/kotlin/incubating/RetryTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import sample.server.TagSubscription
+import sample.server.ZeroQuery
+import kotlin.test.assertIs
+
+class RetryTest {
+  @Test
+  fun retryIsWorking() = runTest(skipDelays = false) {
+    // Hopefully it's really a free port, if not this test will fail
+    val freePort = 8392
+    var sampleServer = SampleServer(freePort, "tag1")
```
BoD, line 26: We could extract the port from sampleServer.subscriptionsUrl(). But the hardcoded port is most probably fine.

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/RetryOnErrorInterceptor.kt
```diff
@@ -0,0 +1,71 @@
...
+      attempt++
+
+      delay(2.0.pow(attempt).seconds)
+      cause is RetryException
+    }
+  }
+}
+
+
+private fun ApolloException.isTerminalAndRecoverable(): Boolean {
```
BoD, line 61: Would it make sense to have this logic as a configurable lambda on `ApolloResponse`? So one could check for `exception` but also `errors` or even `data`.

BoD (approved):
Very nice! 🎉

tests/websockets/src/jvmTest/kotlin/incubating/RetryTest.kt
martinbonnin: Excellent point, done in https://github.com/apollographql/apollo-kotlin/pull/5685/commits/297950d86f4e5b21421b323b9a9d4bce26f24e9b

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/RetryOnErrorInterceptor.kt
martinbonnin: I have bikeshedded the API a bit:

* `retryNetworkErrors` -> `retryOnError` because it's shorter and pronounces better IMO
* There is now a required `defaultRetryOnError` in `RetryOnErrorInterceptor` so as to not break the symmetry of `RetryOnErrorInterceptor`

Overall I'm not sure yet how much retry logic belongs to user land vs something that we can provide with a bunch of parameters (the lambda you're proposing, + exponential backoff + Android Connectivity manager, maybe auth, etc...). My hunch is that it's certainly something that belongs to user land but maybe after some usage we can identify a minimal set of params that would serve the most use cases.

tldr; I would make `setRetryOnErrorInterceptor` API stable soon but wait a lot more for `RetryOnErrorInterceptor`

martinbonnin: I ended up changing thing again [here](https://github.com/apollographql/apollo-kotlin/pull/5690/files#diff-aa6dd8789c20afff5b1d558e73067359cd06b9313363cc9dc764a2c551611ac4R360). The `RetryOnErrorInterceptor` is now always installed by default. 

`ApolloClient.Builder` now takes a `retryOnError((ApolloRequest) -> Boolean)` to set a default `retryOnError` when not specified. By default it is false. You can see it in action  [here](https://github.com/apollographql/apollo-kotlin/pull/5690/files#diff-956eb6b81f65aa82149cac71ee464043a3c5b605f8bbe886417469241f3b48b9L28)

BoD: so retrying subscriptions is no longer the default, it needs to be explicit, right?

martinbonnin: It never has been. In that last [pr](https://github.com/apollographql/apollo-kotlin/pull/5690/files#diff-aa6dd8789c20afff5b1d558e73067359cd06b9313363cc9dc764a2c551611ac4R360), the `RetryOnErrorInterceptor` is always used and opt-in is made with `ApolloClient.Builder.retryOnError {}`while in previous attempts, opt-in was made with `ApolloClient.Builder.retryOnErrorInterceptor()`

BoD: All right - SGTM 👍

BoD (approved):

PR #5686, opened 2024-03-05
[build] Refactor in the Android configuration

martinbonnin:
Configure the android min version and compose version in `apolloLibrary {}`

### DIFF
--- a/build-logic/src/main/kotlin/Android.kt
+++ b/build-logic/src/main/kotlin/Android.kt
@@ -1,0 +1,33 @@
+import com.android.build.api.dsl.CommonExtension
+import org.gradle.api.Project
+
+fun Project.configureAndroid(
+    namespace: String,
+    androidOptions: AndroidOptions
+) {
+  plugins.apply("com.android.library")
+  extensions.findByName("android")?.apply {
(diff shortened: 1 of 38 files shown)


BoD (approved):
👍

build-logic/src/main/kotlin/Android.kt
```diff
@@ -0,0 +1,33 @@
...
+    this as CommonExtension<*, *, *, *, *>
+
+    compileSdk = getCatalogVersion("android.sdkversion.compile").toInt()
+    this.namespace = namespace
+
+    defaultConfig {
+      minSdk = if (androidOptions.withCompose) {
+        getCatalogVersion("android.sdkversion.min").toInt()
+      } else {
+        getCatalogVersion("android.sdkversion.compose.min").toInt()
```
msfjarvis, line 19: This looks incorrect, shouldn't the `compose.min` version be used when `androidOptions.withCompose` is true?

martinbonnin: 100% a mistake, thanks for catching this. Fix in https://github.com/apollographql/apollo-kotlin/pull/5699

PR #5687, opened 2024-03-06
[📖 docs] Add a note about `addTypename` in the cache docs

martinbonnin:


docs/source/caching/declarative-ids.mdx
```diff
@@ -56,13 +60,29 @@ You can specify _multiple_ key fields for an object if they're all required to u
...
+```
 
 > All of an object type's key fields must return a [scalar type](https://www.apollographql.com/docs/apollo-server/schema/schema/#scalar-types).
 
 > Note that the **key fields** specified this way will automatically be added during code generation to selections on the object type, since they are always needed to identify the object in the cache.
 > This means you don't need to include them in your queries.
 
+## Adding `__typename` to your operations
+
+In addition to the key fields, the declarative cache requires the `__typename` of each object by default. These typenames are not included by default as they make query larger for non-cache users. To avoid cache misses, add `__typename` to every selection using [addTypename](https://www.apollographql.com/docs/kotlin/kdoc/apollo-gradle-plugin-external/com.apollographql.apollo3.gradle.api/-service/add-typename.html):
```
martinbonnin, line 76: This Kdoc link could break in v5. It'd be nice to have per-branch permalink. I'll try to think of something but for the foreseeable future, this should be ok.

BoD (approved):
👍

PR #5692, opened 2024-03-07
🍎 Fix a NSUrlSession warning

martinbonnin:
I got this when running an app in Xcode:

Screenshot 2024-03-07 at 00 23 52


BoD (approved):
👍

PR #5694, opened 2024-03-07
Add MockServer.enqueueError() and MockServer.assertNoRequest()

martinbonnin:
2 utility functions that I wish I had while writing subscription tests

### DIFF
--- a/libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
+++ b/libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
@@ -281,6 +281,8 @@ fun MockServer.enqueueString(string: String, delayMs: Long = 0, statusCode: Int 
 fun MockServer.enqueueError(statusCode: Int) {
   enqueue(MockResponse.Builder()
       .statusCode(statusCode)
+      .body("")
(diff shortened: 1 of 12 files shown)


libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -251,38 +251,59 @@ fun MockServer(): MockServer = MockServerImpl(
...
       .body(string)
       .addHeader("Content-Type", contentType)
       .delayMillis(delayMs)
-      .build())
+      .build()
+  )
 }
 
-fun MockServer.enqueueGraphQLString(string: String) {
+fun MockServer.enqueueError(statusCode: Int = 200) {
```
BoD, line 281: Maybe the default should be 500?

```suggestion
fun MockServer.enqueueError(statusCode: Int = 500) {
```

BoD (approved):

libraries/apollo-mockserver/src/commonMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
martinbonnin: Urg 🤦  Good catch. It should be a manfatory parameter.

PR #5698, opened 2024-03-11
[CI flake] Hopefully fix WebSocketEngineTest

martinbonnin:
Looks like this happens when the host machine is under high load (e.g. compiling other modules, running compiler tests).

I'm not sure why but changing the executor moves the latency from ~1s to <100ms on my machine


martinbonnin:
> I wonder if we should make the same change in the HttpEngine

Good question. I'm not 100% about the fix yet. Let's merge this one and if we don't have CI flakes today, I'll do the same in the HttpEngine

martinbonnin:
[That didn't work](https://github.com/apollographql/apollo-kotlin/actions/runs/8243650192/job/22551354211?pr=5704)

PR #5703, opened 2024-03-11
Add test for using the incubating WebSocketNetworkTransport with JsExport

### LINKED ISSUE #5701: [websockets] Test incubating WebSocketNetworkTransport with @JsExport
### Description

Can we use `@JsExport` with websockets?

martinbonnin:
Closes #5701

This PR introduces `SubscriptionParser` (could probably reused outside of subscriptions) to allow callers to implement their own parsing, in this case using `JSON.parse()`


BoD (approved):
👍

PR #5707, opened 2024-03-12
[incubating-network-transport] Kdoc

martinbonnin:
No functional change. Just KDoc and visibility fixes

### DIFF
--- a/libraries/apollo-websocket-network-transport-incubating/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/incubating/OperationListener.kt
+++ b/libraries/apollo-websocket-network-transport-incubating/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/incubating/OperationListener.kt
@@ -1,3 +1,0 @@
-package com.apollographql.apollo3.network.ws.incubating
-
-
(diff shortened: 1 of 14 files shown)


libraries/apollo-websocket-network-transport-incubating/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/incubating/OperationListener.kt
```diff
@@ -1,37 +1,3 @@
 package com.apollographql.apollo3.network.ws.incubating
 
-import com.apollographql.apollo3.api.json.ApolloJsonElement
```
BoD, line 3: The file should be deleted right?

BoD (approved):
👍

libraries/apollo-websocket-network-transport-incubating/src/commonMain/kotlin/com/apollographql/apollo3/network/ws/incubating/OperationListener.kt
martinbonnin: Good catch. IDE leftover 🤷

PR #5719, opened 2024-03-13
Introduce "filesystem" sourceSet and use okio 3.9.0

martinbonnin:
Okio `3.9.0` introduces a source set that makes it possible to use `FileSystem.SYSTEM` without having to `expect`/`actual` for all targets 🎉  ([PR](https://github.com/square/okio/pull/1455))

Use this to reduce our `expect`/`actual` usage. There is now a `filesystem` sourceSet for all the targets that have a filesystem:

* JVM
* native
* but also JS in some cases (with `NodeJsFileSystem`)

The only target that is not included in practice is `wasmJs` for now.

Also move symbols that require a filesystem to the `filesystem` source set. This means it's not possible to call them by mistake on `wasmJs`. It's still possible to call them by mistake in the browser in which case they will fail at runtime.


BoD (approved):
👍

PR #5733, opened 2024-03-15
More KDoc and deprecate ApolloClient.Builder.addInterceptors()

### LINKED ISSUE #5723: Remove sendApqExtensions and sendDocument from `MutableExecutionOptions`
I don't think it makes sense to have a default at the `ApolloClient` layer?

Can probably remove `addInterceptors` too.

martinbonnin:
For consistency with other builders. There is `addFoo()` and `foos()` but no `addFoos()`

Closes #5723


BoD (approved):
👍

PR #5749, opened 2024-03-21
[CI flake] attempt at fixing watcher test flakes

martinbonnin:
- log when there is a lost event and increase buffer size
- increase timeouts and add assertNoElement

See https://github.com/apollographql/apollo-kotlin/issues/3920
See https://github.com/apollographql/apollo-kotlin/issues/2226

### DIFF
--- a/libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/channels.kt
+++ b/libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/channels.kt
@@ -1,10 +1,23 @@
 package com.apollographql.apollo3.testing
 
 import com.apollographql.apollo3.annotations.ApolloExperimental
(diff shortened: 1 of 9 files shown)


libraries/apollo-testing-support/src/commonMain/kotlin/com/apollographql/apollo3/testing/channels.kt
```diff
@@ -1,10 +1,23 @@
...
   receive()
 }
+
+@ApolloExperimental
+suspend fun <T> Channel<T>.assertNoElement(timeoutMillis: Long = 300): Unit {
+  try {
+    withTimeout(timeoutMillis) {
+      receive()
+    }
+    error("An item was received and no item was ")
```
BoD, line 19:
```suggestion
    error("An item was unexpectedly received")
```

BoD (approved):
👍

Issue #5735, opened 2024-03-18
Pagination: merge `nodes` field in addition to `edges`


BoD:
Currently it is possible to configure the (incubating) pagination system to handle Relay-style pagination by using `@TypePolicy(connectionFields: ...)` and `ConnectionFieldMerger`.

This will automatically merge the lists in `edges` into a single record, allowing the app code to watch it and update the UI.

In addition to the `edges` field, it is common for schemas to also have a `nodes` field (e.g. the GitHub and Shopify schemas), which simplifies a bit the usage of the API. `ConnectionFieldMerger` could handle this by merging `nodes` in addition to `edges`, if present.

Original question about this [here](https://github.com/apollographql/apollo-kotlin/issues/3593#issuecomment-2003872915).

martinbonnin:
Is there? (Or can we add?) an existing pull request or proposal to add this to the [Relay specification](https://github.com/facebook/relay/blob/main/website/spec/Connections.md)? Would be rad to add this but having some proposal before doing the implementation would be even rad-er!

sonatard:
I also hope that Nodes will be reflected in the Relay specification.

In the Relay specification, Edge cursors are not used, but even so, this is in the Note format and does not fundamentally address the issue.
https://github.com/facebook/relay/blob/main/website/spec/Connections.md?plain=1#L377-L381

Additionally, the terms "Relay Classic", "Relay Legacy", and "Relay Modern" appear in the Relay specification, but we do not understand what each of them refers to.

martinbonnin:
@sonatard do you want to open an issue in the Relay repo and we can follow up with the Relay team to clarify the terms and explore the possible addition of `nodes`?

sonatard:
Thank you for the suggestion. I plan to open two issues.

sonatard:
Create a following issue
https://github.com/facebook/relay/issues/4653

And I had actually created an issue about nodes in the past. I had forgotten about it until today.
https://github.com/facebook/relay/issues/3850

martinbonnin:
Good find! upvoted!

BoD:
Closing now as this is available in the incubating caching artifact. Don't hesitate to open tickets for comments/issues with this.

PR #5754, opened 2024-03-21
[Pagination] Support nodes in Connection types

BoD:
Fix for #5735.

### DIFF
--- a/libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/RecordMerger.kt
+++ b/libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/RecordMerger.kt
@@ -95,6 +95,16 @@ class FieldRecordMerger(private val fieldMerger: FieldMerger) : RecordMerger {
   }
 }
 
+/**
+ * A [RecordMerger] that merges lists in Relay style Connection types.
+ *
+ * It will merge the `edges` and `nodes` lists and update the `pageInfo` field.
+ *
+ * If the incoming data can't be merged with the existing data, it will replace the existing data.
+ *
+ * Note: although `nodes` is not a standard field in Relay, it is often used - see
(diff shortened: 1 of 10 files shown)


martinbonnin (commented):
Questions:
* Do we want to make a new version of https://github.com/apollographql/specs/blob/main/kotlin_labs/v0.2/kotlin_labs-v0.2.graphql ? 
* Is it ok that `connectionFields` applies to both standard connections and non-standard ones? The risk of name clash is probably low but still.

For those kinds of things that are iterating quickly and where tagging new versions of directives is cumbersome, what about a "draft" version like there is for the GraphQL spec?

libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/RecordMerger.kt
```diff
@@ -111,41 +111,95 @@ private object ConnectionFieldMerger : FieldRecordMerger.FieldMerger {
     return if (incomingBeforeArgument == null && incomingAfterArgument == null) {
```
martinbonnin, line 111: Add the link to the [Relay issue](https://github.com/facebook/relay/issues/3850) as a comment somewhere here?

BoD:
> Do we want to make a new version of [apollographql/specs@main/kotlin_labs/v0.2/kotlin_labs-v0.2.graphql](https://github.com/apollographql/specs/blob/main/kotlin_labs/v0.2/kotlin_labs-v0.2.graphql?rgh-link-date=2024-03-21T17%3A24%3A02Z) ?
> Is it ok that connectionFields applies to both standard connections and non-standard ones? The risk of name clash is probably low but still.

I think it's ok to keep only `connectionFields` - with this change, we merge `edges` _and_ `nodes` if present. So it's more like a extra thing that it does. If a `nodes` field is present and is not a list, it will be ignored, so I think it's not too risky.

I like the idea of having a `draft` version of the spec though - I can set this up.

Issue #3920, opened 2022-03-14
❄️  Flaky test: Timed out waiting for 500 ms


martinbonnin:
FWIW, looks like we have a flaky test. Usually re-running the CI is enough. I'm logging the error in this issue so that we can take action if this becomes too much of an issue.

```
test.WatcherTest[javaCodegen] > cacheOnlyFetchPolicy[javaCodegen] FAILED
    kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 500 ms
        at (Coroutine boundary.()
        at com.apollographql.apollo3.testing.ChannelsKt$receiveOrTimeout$2.invokeSuspend(channels.kt:9)
        at test.WatcherTest$cacheOnlyFetchPolicy$2.invokeSuspend(WatcherTest.kt:313)
        at com.apollographql.apollo3.testing.RunWithMainLoopJvmKt$runTest$3.invokeSuspend(runWithMainLoopJvm.kt:15)

        Caused by:
        kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 500 ms
            at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:186)
            at kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:156)
            at kotlinx.coroutines.EventLoopImplBase$DelayedRunnableTask.run(EventLoop.common.kt:497)
            at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:274)
            at kotlinx.coroutines.DefaultExecutor.run(DefaultExecutor.kt:69)
            at java.lang.Thread.run(Thread.java:750)
```

martinbonnin:
```
test.WatcherTest[javaCodegen] > cacheOnlyFetchPolicy[javaCodegen] FAILED
    kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 500 ms
        at (Coroutine boundary.()
        at com.apollographql.apollo3.testing.ChannelsKt$receiveOrTimeout$2.invokeSuspend(channels.kt:9)
        at test.WatcherTest$cacheOnlyFetchPolicy$2.invokeSuspend(WatcherTest.kt:313)
        at com.apollographql.apollo3.testing.RunWithMainLoopJvmKt$runTest$3.invokeSuspend(runWithMainLoopJvm.kt:15)

        Caused by:
[...]

PR #5755, opened 2024-03-21
[🗄️ cache] Make ApolloStore.publish() suspend

martinbonnin:
Made `publish` suspend so that the cache writers wait for the readers and no cache key event is lost. The buffer of size 64 is still there to smooth things out.

Incubating cache gets removed `publish` parameter.  
Stable cache gets overloads `writeFoo()` functions. The ones with a `publish` parameter should probably be deprecated but I'm not 100% sure yet.

Next steps: 
* benchmark thread hops vs no thread hops
* make `watch()` return a `FusibleFlow` so that the user has control of the buffering there.
* maybe deprecated `ApolloStore.writeFoo(publish)`

See #3920


tests/integration-tests/src/commonTest/kotlin/test/WatcherTest.kt
```diff
@@ -330,14 +331,17 @@ class WatcherTest {
...
+    val job = launch {
       apolloClient.query(query)
           .watch(null)
           .collect {
             channel.send(it.data)
           }
     }
 
+    // Because subscribe is called from a background thread, give some time to be effective
+    delay(500)
```
martinbonnin, line 343: That's an interesting one. If we want a guarantee that `.watch()` subscribes synchronously, we probably need to go back to "interceptors do their own dispatcher switch". I want to measure how costly the thread switch is.

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ApolloStore.kt
```diff
@@ -75,14 +75,30 @@ interface ApolloStore {
...
       operation: Operation<D>,
       operationData: D,
       customScalarAdapters: CustomScalarAdapters = CustomScalarAdapters.Empty,
       cacheHeaders: CacheHeaders = CacheHeaders.NONE,
       publish: Boolean = true,
   ): Set<String>
 
+  /**
+   * Write an operation data to the store and optionally publish changes of [Record] which have changed,
+   * that will notify any watcher that depends on these [Record] to re-fetch.
```
BoD, line 88:
```suggestion
   * Write an operation data to the store.
```

Or maybe do we want to highlight it a bit more? 

```suggestion
   * Write an operation data to the store. Call [publish] with the returned keys to notify any watcher that depends on these [Record] to re-fetch.
```

(same for `writeFragment`)

BoD (approved):
⚙️

libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ApolloStore.kt
martinbonnin: Good call out. I added a `@see publish` tag so hopefully it should be enough.

PR #5756, opened 2024-03-21
[CI flakes ❄️] Hopefully fix benchmarks

martinbonnin:
Increase the benchmark timeout. Looks like they are doing more work and thermal throttling kicks in, which, itself delays the test by 90s every time.

If this becomes too much of an issue/too expensive, we might want to decrease the work load instead.


BoD (approved):
👍

PR #5758, opened 2024-03-22
[apollo-ast] Enforce directive validation by default

martinbonnin:
We used to enforce validation for "new" client directives. Let's do the opposite instead and require a directive validation for all directives except the "old" ones that we lenienly accept for now


BoD (approved):
👍

PR #5761, opened 2024-03-25
[runtime] remove `androidx.startup` dependency

martinbonnin:
`androidx.startup` is not always available (unit tests for an example but also some other cases) and more generally, it's confusing to have the `NetworkMonitor` fail silently if `androidx.startup` doesn't work or if the `READ_NETWORK_STATE` permission is not granted.

In order to use `failFastIfOffline` and `waitForNetwork()`, you'll have to opt-in explicitely (I added some documentation [here](https://github.com/apollographql/apollo-kotlin/pull/5759/commits/0f72213251f1bc7b0ab55919f48e101a32139a49)):

```kotlin
  .serverUrl(...)
  // Android
  .networkMonitor(NetworkMonitor(context))
  // iOS
  .networkMonitor(NetworkMonitor())
```

It means you can't create a `NetworkMonitor` from `commonMain`, you'll have to expect/actual it but at this point, I think it's just more consistent.

Also take this opportunity to 

* initialize the `ConnectivityManager` on a background thread. See https://github.com/coil-kt/coil/pull/2104
* remove `ApolloClient.Builder.retryOnErrorInterceptor()`. The interceptor is added unconditionally. It can be disabled by setting `failfastOnError = false` and `retryOnError = false` (default)

See https://github.com/apollographql/apollo-kotlin/issues/5760
See https://github.com/apollographql/apollo-kotlin/pull/5720

- remove androidx.startup dependency
- initialize the ConnectivityManager in a background thread
- focus the RetryInterceptor on Network errors, remove ApolloClient.Builder.retryOnErrorInterceptor
- Add ensureUniqueUuids
- apiDump

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/NetworkObserver.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/network/NetworkObserver.kt
@@ -16,7 +16,9 @@ internal interface NetworkObserver: Closeable {
 }
 
 internal val NoOpNetworkObserver = object : NetworkObserver {
-  override fun setListener(listener: NetworkObserver.Listener) {}
+  override fun setListener(listener: NetworkObserver.Listener) {
+    listener.networkChanged(true)
+  }
 
   override fun close() {}
 }


build-logic/src/main/kotlin/Android.kt
```diff
@@ -21,6 +23,10 @@ fun Project.configureAndroid(
         getCatalogVersion("android.sdkversion.min").toInt()
       }
       testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+
+      if (this is LibraryVariantDimension) {
+        multiDexEnabled = true
```
BoD, line 28: Why do we need this?

BoD (approved):
👍

build-logic/src/main/kotlin/Android.kt
martinbonnin: The test APK grew over 64k methods 🤷 . 

Screenshot 2024-03-25 at 13 30 53
Screenshot 2024-03-25 at 13 31 06

Interestingly, it's not the case on those 2 screenshots (I think? but somehow two dex files are still required 🤷 )
In all cases, it's close enough that it may happen again. Plus AGP enables it by default if `minSdk > 21` anyways. 

No place for degrowth in dex-land 🌱 🙈

PR #5765, opened 2024-03-25
Simplify ApolloCall

martinbonnin:
`ApolloCall` doesn't need to duplicate all the `ApolloRequest` properties, it can just hold a mutable `ApolloRequest.Builder` because that's what it really is today. 

This also exposes `ignoreApolloClientHttpHeaders` on `ApolloRequest` and makes the API more consistent.

Also update the KDoc. Our KDoc for those symbols is still mostly empty though because of https://github.com/Kotlin/dokka/issues/3547

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
@@ -15,7 +15,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.toList
 
 /**
- * An [ApolloCall] is a thin class that binds an [ApolloRequest] with its [ApolloClient] and offers a fluent way to build an [ApolloRequest].
+ * An [ApolloCall] is a thin class that binds an [ApolloRequest] with its [ApolloClient]. It offers a fluent way to configure the [ApolloRequest].
  *
  * Contrary to an [ApolloRequest], an [ApolloCall] doesn't have a request id and a new request id is allocated every time [execute] or [toFlow] is called.
  *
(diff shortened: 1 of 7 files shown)


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -210,11 +171,12 @@ class ApolloCall<D : Operation.Data> internal constructor(
   /**
    * Retrieves a single [ApolloResponse] from this [ApolloCall].
    *
-   * Use this for queries and mutations to get a single value from the network or the cache.
-   * For subscriptions or operations using `@defer` that may return multiple values, use [toFlow] instead.
+   * [execute] calls [toFlow] and filters out cache or network errors to return a single success [ApolloResponse].
    *
-   * [execute] may fail due to an I/O error, a cache miss or other reasons. In that case, check [ApolloResponse.exception]:
+   * [execute] throws is more than one success [ApolloResponse] is returned, for an example, if [operation] is a subscription or a `@defer` query.
```
BoD, line 176:
```suggestion
   * [execute] throws if more than one success [ApolloResponse] is returned, for an example, if [operation] is a subscription or a `@defer` query.
```

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/ApolloCall.kt
```diff
@@ -15,121 +15,96 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.toList
 
 /**
- * An [ApolloCall] is a thin class that binds an [ApolloRequest] with its [ApolloClient].
+ * An [ApolloCall] is a thin class that binds an [ApolloRequest] with its [ApolloClient] and offers a fluent way to build an [ApolloRequest].
```
BoD, line 18: Feels a bit off but not sure. Maybe:
```suggestion
 * An [ApolloCall] is a thin class that binds an [ApolloRequest] with its [ApolloClient]. It offers a fluent way to configure the [ApolloRequest].
```

BoD (approved):
👍

Issue #5777, opened 2024-04-02
ApolloClient.Builder.okHttpClient() returns `null` instead of `this`


japhib:
### Version

main

### Summary

This code:
```java
new ApolloClient.Builder().okHttpClient(okHttpClient()).build()
```
throws a NullPointerException because the `okHttpClient` setter method on `ApolloClient.Builder` [returns `null` rather than `this`](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-runtime-java/src/main/java/com/apollographql/apollo3/runtime/java/ApolloClient.java#L232).

The javadoc for that method says this:
```
@return The {@link Builder} object to be used for chaining method calls
```
which makes it seem like a typo.

### Steps to reproduce the behavior

Run `new ApolloClient.Builder().okHttpClient(new OkHttpClient()).build()` and observe that a NullPointerException is thrown.

japhib:
PR: https://github.com/apollographql/apollo-kotlin/pull/5778

BoD:
Fixed in 3.8.4

PR #5778, opened 2024-04-02
#5777 fix ApolloClient.Builder.okHttpClient() returns `null` instead of `this`

japhib:
See https://github.com/apollographql/apollo-kotlin/issues/5777


BoD (approved):
Good catch 👍 Thanks!

PR #5784, opened 2024-04-08
Support configuring `ApolloClient` with lazily initialized `Call.Factory`.

ychescale9:
Resolves https://github.com/apollographql/apollo-kotlin/issues/5775

I've tested this locally:

**Eager `OkHttpClient` init:**

image

**Lazy `OkHttpClient` init:**

image

Creation of `ApolloClient` no longer crates an `OkHttpClient`.

image

~~The creation of `OkHttpClient` is now deferred to after the "App Startup" window (by Coil `ImageLoader`).~~

Correction: the `OkHttpClient` is still created within the "App Startup" window, but it's done in a background thread by Coil.

-----

The actual perf impact of using the lazy API obviously depends on each app.


BoD (approved):
Thanks a lot! This looks good!

martinbonnin (approved):
Nice! Thanks!

PR #5794, opened 2024-04-10
Move some tests to .websocket part 2/n

martinbonnin:
[SubscriptionTest.kt](https://github.com/apollographql/apollo-kotlin/compare/update-tests...update-tests-2?expand=1#diff-b937d74f77d02b4c7fa13e574a21c3adace591b8420fa2f001b17ca60c7aca65) is now moved to apollo-runtime.

2 nice benefits:

* tests are closer to the implementation, no need for included builds.
* no need for FakeWebSocketNetworkTransport. MockServer does everything we need :)

Moving forward, I think all our tests should use `MockServer`. There's still some work to replace the logic in `SampleServer` but we'll get there!


BoD (approved):
👍

PR #5797, opened 2024-04-10
[Codegen] Add CompiledArgumentDefinition

### LINKED ISSUE #5781: Codegen: rework how compiled field arguments are generated
Currently, the info about an argument being a key or paging arg is embedded into the `CompiledField`, which is not consistent. Instead let's introduce a `CompiledArgumentDefinition` for this

BoD:
- Introduce `CompiledArgumentDefinition` instead of storing schema argument details (isKey, isPagination) in `CompiledArgument`
- They are added as fields inside the `Type` class, with the form `__fieldName__argumentName`
- To limit the size of the generated code, track arguments that are used by operations and output only the definition for those
  - To do that, a wrapper class `UsedCoordinates` is introduced, to avoid manipulating maps directly.
  - The used coordinated are now tracked by the `CodegenTest` in `used-coordinates.json` files

- Behavior change: only the specified arguments values are set to the `CompiledArgument`, and not the default value from the schema. This makes the note saying `The argument defaultValues are not added to the name` in the Kdoc of `nameWithArguments()` true.
 
Resolves #5781


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrOperationsBuilder.kt
```diff
@@ -751,20 +745,80 @@ internal class IrOperationsBuilder(
...
         }
       }
     }
   }
+
+  private fun UsedCoordinates.putAllArguments(
+      parentType: String,
+      selections: List<GQLSelection>,
+      allFragmentDefinitions: Map<String, GQLFragmentDefinition>,
+  ) {
```
martinbonnin, line 777: If you end up keeping this, `allFragmentDefinitions` is in the context already, I think you can skip it here.

Also side note: yet another thing that would be easier to do with a proper traversal API.

libraries/apollo-compiler/src/test/graphql/com/example/antlr_tokens/java/operationBased/used-coordinates.json.expected
```diff
@@ -0,0 +1 @@
+{"typeToFieldsToArguments":{"TypeWithGraphQLKeywords":{"null":["fragment"]},"String":{},"Query":{}}}
```
martinbonnin, line 1: Can you output the JSON with indentation? Will make it easier to parse diffs.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrOperationsBuilder.kt
```diff
@@ -594,7 +608,12 @@ internal class IrOperationsBuilder(
        *   }
        * }
        */
-      usedFields.putType(first.type.rawType().name)
+      usedCoordinates.putType(first.type.rawType().name)
+
+      // When a field with arguments is selected, its parent type is referenced in the compiled selections
+      if (first.definitionHasArguments) {
+        usedCoordinates.putType(first.parentType)
```
martinbonnin, line 615: Could you add all the arguments here instead of calling `UsedCoordinates.putAllArguments` at the end of the process? You'll probably have to add `arguments` to the `CollectedField` but it's colocating everything.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrSchema.kt
```diff
@@ -62,6 +66,10 @@ internal data class IrObject(
     val deprecationReason: String?,
     val embeddedFields: List<String>,
     val mapProperties: List<IrMapProperty>,
+    /**
+     * Only contains fields that have arguments used in operations.
```
martinbonnin, line 70: I think it's OK to skip that comment since it's the whole premise of the IR that it matches the target and not the GraphQL schema.

We could add a more general comment how these things are decided in `IrOperationBuilder` and link from here  though.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/CompiledGraphQL.kt
```diff
@@ -412,50 +412,78 @@ class CompiledVariable(val name: String)
...
     )
   }
 }
 
+class CompiledArgument private constructor(
+    val definition: CompiledArgumentDefinition,
+    /**
+     * The compile-time value of that argument.
+     *
+     * Can be the defaultValue if no argument is defined in the operation.
```
martinbonnin, line 457: Didn't we go for  "no need for the schema `defaultValue`"?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/SelectionSetsBuilder.kt
```diff
@@ -81,28 +69,16 @@ internal class SelectionSetsBuilder(
...
-
-      val keyArgs = typeDefinition.keyArgs(name, schema)
-      val paginationArgs = typeDefinition.paginationArgs(name, schema)
-
-      /**
-       * When passed explicitly, the argument values are coerced (but not their default value)
-       */
-      val userValue = operationArgument?.value?.coerceInExecutableContextOrThrow(schemaArgument.type, schema)
+    val actualArguments = arguments.map { fieldArgument ->
+      val schemaArgument = fieldDefinition.arguments.first { it.name == fieldArgument.name }
```
martinbonnin, line 73: much simpler 👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/Resolver.kt
```diff
@@ -33,6 +33,7 @@ enum class ResolverKeyKind {
   Schema,
   CustomScalarAdapters,
   Pagination,
+  ArgumentDefinition,
```
martinbonnin, line 36: For later: we can use the coordinate to lookup the matching kotlin instead of adding a new kind each time.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/ir/IrSchema.kt
```diff
@@ -129,10 +129,17 @@ internal class IrCompositeType2(val name: String) : IrType2
...
+    val id: String,
     val name: String,
+    val propertyName: String,
     val isKey: Boolean,
     val isPagination: Boolean,
-)
+) {
+  companion object {
+    fun id(type: String, field: String, argument: String) = "$type.$field.$argument"
+    fun propertyName(fieldName: String, argumentName: String) = "__${fieldName}_${argumentName}"
```
martinbonnin, line 134: For later: we should come up with some rules about symbols. Are `__` to escape nameclashes or a signal that it is not meant to be called directly. It's not 100% clear at the moment

martinbonnin (approved):
👍

PR #5805, opened 2024-04-12
Let isFromCache be about the ApolloResponse

### LINKED ISSUE #5799: `isFromCache` is potentially confusing
### Version

3.x-4.0.0-beta.5

### Summary

`ApolloResponse` has an extension function, `isFromCache`. When using `FetchPolicy.CacheFirst`, if there's a cache miss, an `ApolloResponse` is emitted where `data == null` and `it.exception` is a `CacheMissException`. For this response, I would expect that `isFromCache` returns `false`. This is because it relies on `isCacheHit` being `true`, which is not the case for this response. [Link to code](https://github.com/apollographql/apollo-kotlin/blob/b018e29a7ec7eddb2d1206ce048890a3d2bfd53f/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt#L538)

I think it'd be ideal to make it return `true`, but I'm not sure if we can change this API now. Maybe we can introduce a new API, `isResponseFromCache` that does not rely on `cacheHit` being `true`, perhaps it returns `isCacheHit || it.exception is CacheMissException`?

_No response_

### Logs

```
(Your logs here)
```

BoD:
Resolves #5799

### DIFF
--- a/design-docs/Error handling changes in v4.md
+++ b/design-docs/Error handling changes in v4.md
@@ -1,0 +1,22 @@
+# Error handling changes in v4
+
+This document describes the changes made around error handling in v4. The purpose is to keep a log of the changes, in a
+kind of "decision record" way.
+
+## Broaden the meaning of `ApolloResponse.isFromCache`
+
+In v3, `ApolloResponse.isFromCache` returns `true` if the **data** comes from the cache (implementation:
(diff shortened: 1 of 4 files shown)


rohandhruva:
🙏 Thanks!

martinbonnin (approved):

PR #5806, opened 2024-04-12
Move some tests to .websocket part 4/4

martinbonnin:
I'll stop here for now. `tests/websocket` still contains:

* disabled tests against real servers (router, fullstack, appsync)
* one test for the current `.ws` websockets that is really to port over (`canResumeAfterGraphQLError`)
* cache + subscription tests

SampleServer is still around but I'll probably deprecate it at some point.


BoD (approved):
👍

PR #5809, opened 2024-04-15
Add DocumentTransform API

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/5807

PS: I also took a quick look at adding an API like 

```kotlin
fun write(schema: Schema, operations: List<GQLOperationDefinition>, outputDir: File) 
```

to allow contributing more files to the codegen but I ended up backtracking because the `GQLOperationDefinition` are not available in the codegen phase. All in all, I think this is better served by a custom Gradle task outside of the Apollo Gradle Plugin.


martinbonnin:
In light of https://github.com/apollographql/apollo-kotlin/issues/5807#issuecomment-2058442463, I'd vote for merging this one as experimental so we can play with it.

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/ApolloGenerateSourcesBaseTask.kt
```diff
@@ -72,10 +72,10 @@ fun ApolloGenerateSourcesBaseTask.layout(): LayoutFactory {
 fun ApolloGenerateSourcesBaseTask.requiresBuildscriptClasspath(): Boolean {
   if (packageNameGenerator != null || operationOutputGenerator != null) {
     if (packageNameGenerator != null) {
-      logger.lifecycle("Apollo: packageNameGenerator is deprecated, use Apollo compiler plugins instead")
+      logger.lifecycle("Apollo: packageNameGenerator is deprecated, use Apollo compiler plugins instead. See https://go.apollo.dev/ak-compiler-plugins for more details.")
```
BoD, line 75: Noice!

BoD (approved):
👍

Issue #5796, opened 2024-04-10
Make it impossible to pass as input some type which was generated only to preserve forwards compatibility but was not meant to be used as input


StylianosGakis:
### Use case

For a schema which contains these
```
type Mutation {
    someMutation(input: SomeInput!): Irrelevant!
}

input SomeInput {
    otherField: ID!
    enumInput: SomeEnum
}

enum SomeEnum {
    FOO,
    BAR,
}
```

Apollo-kotlin generates this code for `SomeEnum`:

```kotlin
public enum class SomeEnum(
  public val rawValue: String,
) {
  FOO("FOO"),
  BAR("BAR"),
  /**
   * Auto generated constant for unknown enum values
   */
  UNKNOWN__("UNKNOWN__"),
  ;

  public companion object {
    public val type: EnumType = EnumType("SomeEnum", listOf("FOO", "BAR"))

    public fun safeValueOf(rawValue: String): ChatMessageContext =
        values().find { it.rawValue == rawValue } ?: UNKNOWN__

    /**
     * Returns all [SomeEnum] known at compile time
     */
    public fun knownValues(): Array<SomeEnum> = arrayOf(
      FOO,
      BAR,
    )
  }
}
```

This makes it possible in user code to try and pass to a mutation this:
```
apolloClient.mutation(SomeMutation(input = SomeInput(otherField = ..., enumInput = SomeEnum.UNKNOWN__))
```

This however will always fail on the backend validation since this entry is not part of the real schema, and there is nothing to map it to.

### Describe the solution you'd like

The problem here is just that this is possible and the compiler is not warning us about trying to run something that would never work.

Optimally, this code should not compile in the first place.
If this turns out to be way too complicated, even a lint warning or something similar might be a good first step. But as of right now making this mistake is unfortunately something that can happen silently.

martinbonnin:
For the record, a possible solution is to use `sealedClassesForEnumsMatching.set(listOf(".*"))` (probably make it the default in v4) and have the `UNKNOWN__(rawValue)` constructor be annotated with an `@OptIn` annotation  so that it's harder to call from user code.

BoD:
Heads up that we added the @OptIn annotation to the unknown constructor in #5813. This should be available in the v4 snapshots shortly!

For now we haven't decided on making sealed classes for enums the default in v4.

BoD:
[...]

PR #5813, opened 2024-04-15
Make enum as sealed class Unknown constructor opt-in

BoD:
Related to #5796

This uses `@ApolloInternal`- should we introduce a dedicated annotation instead?

### DIFF
--- a/libraries/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloEnumConstructor.kt
+++ b/libraries/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloEnumConstructor.kt
@@ -1,0 +1,9 @@
+package com.apollographql.apollo3.annotations
+
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "Applied to the constructor of an enum's UNKNOWN__ value, as instantiating it is usually a mistake."
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CONSTRUCTOR)
+annotation class ApolloEnumConstructor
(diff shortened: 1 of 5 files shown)


martinbonnin:
> should we introduce a dedicated annotation instead?

I'd vote yes here to make the intent really clear. There might be use cases where opting-in is desirable and we'll guarantee the API while opting-in `@ApolloInternal` is always dangerous with no API guarantee. Maybe `@ApolloEnumConstructor`.

Ping @StylianosGakis any preference?

StylianosGakis:
I think a separate annotation makes sense yes. Exactly because it may be something that you may want to opt-in without it being the "wrong" thing to do. While using something ApolloInternal should always be the "wrong" thing to do.
This distinction would be welcome in my opinion!

libraries/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloEnumConstructor.kt
```diff
@@ -0,0 +1,9 @@
+package com.apollographql.apollo3.annotations
+
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
```
martinbonnin, line 4: Maybe WARNING?

```suggestion
    level = RequiresOptIn.Level.WARNING,
```

libraries/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo3/annotations/ApolloEnumConstructor.kt
```diff
@@ -0,0 +1,9 @@
...
+    message = "Applied to the constructor of an enum's UNKNOWN__ value, as instantiating it is usually a mistake."
```
martinbonnin, line 5:
```suggestion
    message = "The `UNKNOWN__` class represents GraphQL enums that are not present in the schema and whose `rawValue` cannot be checked at build time. You may want to update your schema instead of calling this constructor directly."
```

BoD: A lot nicer! 🙏

martinbonnin (approved):

PR #5814, opened 2024-04-15
Retry test: fix test flake

martinbonnin:
Threre was a race between:

```kotlin
mockServer.close()

mockServer.open()
// fails if the subscription retry tries to connect here.
mockServer.enqueueWebSocket()
```

Instead of closing the mockServer, we now close the TCP connection. This allows keeping a backlog of mock requests.


BoD (approved):
👍

PR #5815, opened 2024-04-15
❄️ Fix retry tests flake #2

martinbonnin:
When an individual flow completes early, we were mistaking the "complete" event with a "next" event and therefore sending a response for the bad operation id.

Also took this opportunity to lock things a bit more in `SubscribableSocket`


martinbonnin:
Merging quickly cause I want to rebase the other PRs on top of this.

PR #5821, opened 2024-04-16
Better messages and debuggability for ApolloCompilerPlugin

martinbonnin:
* fails if `plugin()` and `packageNameGenerator/operationOutputGenerator/operationIdGenerator` are set at the same time
* renames `Plugin` -> `ApolloCompilerPlugin`
* add links to https://go.apollo.dev/ak-compiler-plugins


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/serviceloader.kt
```diff
@@ -1,14 +1,19 @@
...
-  val plugins = ServiceLoader.load(Plugin::class.java).toList()
+internal fun apolloCompilerPlugin(warnIfNotFound: Boolean = false): ApolloCompilerPlugin? {
+  val plugins = ServiceLoader.load(ApolloCompilerPlugin::class.java).toList()
 
   if (plugins.size > 1) {
     error("Apollo: only a single compiler plugin is allowed")
   }
 
+  if (plugins.isEmpty() && warnIfNotFound) {
+    println("Apollo: a compiler plugin was added with `Service.plugin()` but could not be loaded by the ServiceLoader. Check your META-INF.services/com.apollographql.apollo3.compiler.ApolloCompilerPlugin file.")
```
BoD, line 14:
```suggestion
    println("Apollo: a compiler plugin was added with `Service.plugin()` but could not be loaded by the ServiceLoader. Check your META-INF/services/com.apollographql.apollo3.compiler.ApolloCompilerPlugin file.")
```

BoD (approved):
👍

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/serviceloader.kt
martinbonnin: Good catch 👍

PR #5823, opened 2024-04-17
Add compile time possibleTypes

### LINKED ISSUE #5808: Support for tree shaking in the __Schema.possibleTypes()
### Use case

We have configured apollo-kotlin gradle plugin with `alwaysGenerateTypesMatching.set(listOf(".*"))`. This is so that we can match against union or interface types which don't have explicit spreads but we can use the generated `.type.name` to match against `__typename`. 

Originally, we also had `generateSchema = true`, so that we could use `__Schema.possibleTypes()`, which is very useful for enumerating over interface or union types. However, because `possibleTypes()` operates at runtime, it's not possible for R8 or proguard to strip unused types from the android app APK.

Would it be possible to add support for being able to tree shake unused types out of the generated `__Schema` file? Maybe a middle ground is that we specify all the `possibleTypes` usages in our gradle configuration, and only those are generated, e.g. `__Schema.possibleTypesOfFoo()`.

_No response_

martinbonnin:
Fixes #5808

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinCodegen.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/KotlinCodegen.kt
@@ -203,7 +203,7 @@ internal object KotlinCodegen {
         builders.add(ObjectBuilder(context, irObject, generateDataBuilders))
       }
       if (generateSchema && context.resolver.resolve(ResolverKey(ResolverKeyKind.Schema, "")) == null) {
-        builders.add(SchemaBuilder(context, irSchema.irObjects, irSchema.irInterfaces, irSchema.irUnions, irSchema.irEnums))
+        builders.add(SchemaBuilder(context, codegenSchema.schema, irSchema.irObjects, irSchema.irInterfaces, irSchema.irUnions, irSchema.irEnums))
         builders.add(CustomScalarAdaptersBuilder(context, scalarMapping))
       }
 
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/schema/SchemaBuilder.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/kotlin/schema/SchemaBuilder.kt
@@ -1,12 +1,14 @@
 package com.apollographql.apollo3.compiler.codegen.kotlin.schema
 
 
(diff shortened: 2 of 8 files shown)


BoD (approved):
👍

rohandhruva:
Am I understanding this correctly: with this change, we'll generate `possibleTypes` for all the interfaces and unions in the schema, is that right? 

Can we "force" it to be generated for more types, including which are not interfaces or unions? This is purely for "symmetry", even if it just returns a single type.

```
when (myType) {
  in possibleVideoTypes -> { .. }
  in possibleGameTypes -> { .. }
  in possibleSingleType -> { .. }
  else -> { .. }
```

This "reads" better (IMO) even if `Single` has only one possible type. But, of course, if it creates a lot more complexity, it's probably not worth it.

martinbonnin:
> Am I understanding this correctly: with this change, we'll generate possibleTypes for all the interfaces and unions in the schema, is that right?

That is correct. Since the endgoal of this change was to avoid generating extra bytecode, I didn't include the objects because `possibleTypes` does not bring much value there. Not having it also signals the intent in the code. 

```kotlin
when (myType) {
  in possibleVideoTypes -> { .. }
  in possibleGameTypes -> { .. }
  in possibleSingleType -> { .. }
  else -> { .. }
```

Currently this is more going to look like 

```kotlin
when (myType) {
  in __Schema.videoPossibleTypes() -> { .. }
  in __Schema.gamePossibleTypes() -> { .. }
  in __Schema.singlePossibleTypes() -> { .. }
  else -> { .. }
```

You can import `_Schema.*` to remove the `__Schema.` but I'd rather keep `possibleTypes` in one word. We can use `vals` instead of `fun` though. So in the end something like so:


```kotlin
when (myType) {
  in videoPossibleTypes -> { .. }
  in gamePossibleTypes -> { .. }
  in singlePossibleTypes -> { .. }
  else -> { .. }
```

How does that sound? 
Also let me know about possibleTypes for objects. If you need them, it's easy to add.

rohandhruva:
> How does that sound?

Sounds perfect -- thank you for the consideration! I don't have a strong preference for `val` vs `fun` because I think they'd be about equivalent in this case. 

> Also let me know about possibleTypes for objects. If you need them, it's easy to add.

As you said, it probably doesn't make much sense. Let's skip it. My original suggestion was coming from the fact that we had a `possibleTypes()` function that accepted anything, so it was "easy" to use it for object types as well. But it's unnecessary.

martinbonnin:
PR to turn the fun into vals [here](https://github.com/apollographql/apollo-kotlin/pull/5826).

PR #5836, opened 2024-04-19
Fix links that depend on a redirect

Meschreiber:
Also updates "Apollo Studio" --> "GraphOS Studio" (the updated convention).


martinbonnin (approved):
✨

Issue #5834, opened 2024-04-19
SQL cache is unusably slow


edenman:
### Version

4.0.0-beta.4

### Summary

- New project, interacting with a old and very gnarly graphql schema. 
- Response is ~200kb
- We're observing the cache with a `CacheOnly` policy and `watch`
- When we make a network call, we're using `NetworkOnly` and then manually manipulating the response before writing it to the cache
- When we write a response to the cache, it takes *14* seconds for the `watch` to fire.  If I remove the `SqlNormalizedCacheFactory` that drops down to 3 seconds.

### Steps to reproduce the behavior

```
        ApolloClient.Builder().apply {
            interceptors(apolloInterceptors.toList())
            networkTransport(
                HttpNetworkTransport.Builder()
                    .httpEngine(DefaultHttpEngine(okHttpClient))
                    .httpRequestComposer(RequestComposer)
                    .build(),
            )
            normalizedCache(
                MemoryCacheFactory(10 * 1024 * 1024)
                    .chain(SqlNormalizedCacheFactory("$userId-apollo.db"))
            )
            writeToCacheAsynchronously(true)
            autoPersistedQueries()
        }.build()
```

BoD:
Hi! Thanks for reporting.

Can you tell us a bit more about the response manipulation you're doing, and how it is done? Are you using `doNotStore` so that the data is not written twice to the cache? Is the writing call itself taking a while, or is it only the watch being slow to be triggered (or both).

Also in general if you could share your (or part of your) schema, query and json payload, that would help a lot to have a look and try to understand what is happening (share privately to benoit.lubek@apollographql.com if that's preferable).

edenman:
- Correct, we're doing `doNotStore(true)` on the network query.  
- The response is a list of heterogeneous items
- The response includes some special caching instructions in the body, so we're manually processing those and then merging the existing items with the new ones, and then inserting a new FooResponse into the cache
- I'll send you the json payload via email

BoD:
Thanks a lot for having shared all the details. I could reproduce writing to the cache taking an unusually long time, due to a transaction per insert being used, instead of just one transaction for all of them, when the memory cache is used with chaining. This is fixed [here](https://github.com/apollographql/apollo-kotlin/pull/5840) and will be in the next v4 release (planned in the next few days).

BoD:
Released in 3.8.4

PR #5840, opened 2024-04-22
Sql cache: use a single transaction when using MemoryCache chaining

BoD:
A single transaction is used in `merge(records: Collection<Record>)`, but `merge(record: Record)` was called instead when using memory cache chaining, resulting in using 1 transaction per insert (significantly slower).

~~Also - enable WAL by default. This has no noticeable perf impact with the fix above (single transaction), but showed significant improvement without it.~~ (moved to its own PR)

Related to #5834

### DIFF
--- a/libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/MemoryCache.kt
+++ b/libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/MemoryCache.kt
@@ -98,15 +98,7 @@ class MemoryCache(
       return emptySet()
     }
     return lockWrite {
-      val oldRecord = loadRecord(record.key, cacheHeaders)
-      val changedKeys = if (oldRecord == null) {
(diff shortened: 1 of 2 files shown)


libraries/apollo-normalized-cache-api-incubating/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/api/MemoryCache.kt
```diff
@@ -115,7 +107,23 @@ class MemoryCache(
     if (cacheHeaders.hasHeader(ApolloCacheHeaders.DO_NOT_STORE)) {
       return emptySet()
     }
-    return lockWrite { records.flatMap { record -> merge(record, cacheHeaders, recordMerger) } }.toSet()
+    return lockWrite {
+      val changedKeys = records.flatMap { record -> internalMerge(record, cacheHeaders, recordMerger) }.toSet()
+      changedKeys + nextCache?.merge(records, cacheHeaders, recordMerger).orEmpty()
```
martinbonnin, line 112: 👍

martinbonnin (approved):

PR #5841, opened 2024-04-22
Fix benchmarks

BoD:
They were broken by #5772 and #5802, sorry about that!


martinbonnin (approved):
We could run benchmarks on each PR but I think it's fine if the benchmarks are occasionally broken. We do not always focus on benchmarking and that saves CI time.

PR #5843, opened 2024-04-22
SQL cache: enable WAL (Write Ahead Logging) by default

BoD:
WAL has shown performance gains when executing many independent inserts (see #5834).

### DIFF
--- a/libraries/apollo-normalized-cache-sqlite-incubating/src/androidMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/SqlNormalizedCacheFactory.kt
+++ b/libraries/apollo-normalized-cache-sqlite-incubating/src/androidMain/kotlin/com/apollographql/apollo3/cache/normalized/sql/SqlNormalizedCacheFactory.kt
@@ -20,7 +20,7 @@ actual class SqlNormalizedCacheFactory actual constructor(
   /**
    * @param [name] Name of the database file, or null for an in-memory database (as per Android framework implementation).
    * @param [factory] Factory class to create instances of [SupportSQLiteOpenHelper]
-   * @param [configure] Optional callback, called when the database connection is being configured, to enable features such as
+   * @param [configure] Optional callback, called when the database connection is being configured, to configure features such as
    *                    write-ahead logging or foreign key support. It should not modify the database except to configure it.
    * @param [useNoBackupDirectory] Sets whether to use a no backup directory or not.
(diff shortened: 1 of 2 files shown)


martinbonnin (commented):
No strong opinion here. Without a clear benchmark showing improvement, I'd probably err on the side of leaving things untouched. I'll leave it to you though.

BoD:
Running [this](https://github.com/apollographql/apollo-kotlin/blob/e40c86c5c17243f5152e4291adb01e2b1f32d5c7/benchmark/microbenchmark/src/androidTestStable/kotlin/com/apollographql/apollo3/benchmark/ApolloStoreTests.kt#L30) particular benchmark a few times, I can see a ~12% improvement when using WAL (same with the incubating version of the same test). Confidence is not ultra high however because I get a 8-15% variation between each test.

eduardb:
> Running [this](https://github.com/apollographql/apollo-kotlin/blob/e40c86c5c17243f5152e4291adb01e2b1f32d5c7/benchmark/microbenchmark/src/androidTestStable/kotlin/com/apollographql/apollo3/benchmark/ApolloStoreTests.kt#L30) particular benchmark a few times, I can see a ~12% improvement when using WAL (same with the incubating version of the same test). Confidence is not ultra high however because I get a 8-15% variation between each test.

Sorry for the drive-by, but we are currently focused on app performance improvements, so we are quite invested in these topics. Just curious, what device did you run the benchmark on? Could there have been external factors on the device that introduced the variation? 

Btw, there's also this great article from P-Y that could help: https://blog.p-y.wtf/statistically-rigorous-android-macrobenchmarks.

Also, using a rooted device should help, at least according to the documentation: https://developer.android.com/topic/performance/benchmarking/microbenchmark-overview#lock-clocks.

I hope this helps, and I am not being annoying xD

BoD:
Hey @eduardb no worries at all, quite the contrary! Advice on this topic is very much welcome and appreciated. This is also a hot topic for us currently.

Thanks for the article link, a very good one indeed. The spreadsheet template is promising. 

My methodology was a lot more naive: ran the test 6 times each (with/without WAL), discarded the lowest and highest values, and looked at the averages. 
 Benchmarks ran on a (non rooted) Samsung S10. My hunch is that cpu throttling is probably the cause of the variance. Airplane mode to avoid external apps waking up is probably a good advice too.

PR #5849, opened 2024-04-24
Fix backtick code formatting in ApolloCompilerPlugin docs

edenman:
Note sure what the `<Note>` tag does but it was breaking the backtick markdown formatting


martinbonnin:
Urg, good catch 👍 . `<Note>` is our own doc system equivalent of Github's `> [!NOTE]`. I'll ask around how to do this. We probably have the problem in other places too.

martinbonnin:
Turns out it's a MDX bug from the older version of MDX we're using. The fix is to leave an empty line:

```
<Note>

Using `backticks` here should work
</Note>
```

I just tried it out, will merge if it renders fine. Thanks for catching this!

Meschreiber (approved):

PR #5857, opened 2024-04-28
Clear current ApolloStore related interceptors when calling `.store()` on builder

rohandhruva:
I noticed a bug when using `.newBuilder().store()` on an existing `ApolloClient`. The intention was to set a different `ApolloStore()` using `builder.store(...)` on this client. But after `.build()` on this new client, I noticed that `.watch()` was not working when I wrote some data to the new `store.`

Turns out that we were ending up with 3 duplicated watchers, the ones that are added by `.store()` function. 

image

This PR fixes that problem by clearing the 3 store related watchers when using `.store()` on a builder. 

- Introduced a new `internal interface ApolloStoreInterceptor` to indicate these interceptors 
- In `ClientCacheExtensions`, clear them in `ApolloClient.Builder.store()` function
- This required introducing a new `removeInterceptor()` API

I am not sure if making `removeInterceptor()` a part of the builder API is desirable, but I couldn't think of a way to make this work without introducing such an API. Let me know if you have other approaches I can try for this.


BoD (approved):
Good catch! 🙏  We have other similar cases in the codebase, I'll follow-up with another PR to fix them.

BoD:
(I just added a commit to also do the same in the _incubating_ version)

martinbonnin (approved):

PR #5858, opened 2024-04-29
Remove existing interceptors from ApolloClient.Builder before adding new ones

BoD:
A follow-up to #5857, for `.httpCache()`, `.autoPersistedQueries()` and `.httpBatching()`.

### DIFF
--- a/libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
+++ b/libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
@@ -19,14 +19,14 @@ import java.time.Instant
 import java.time.format.DateTimeParseException
 
 class CachingHttpInterceptor internal constructor(
-    private val lruHttpCache: ApolloHttpCache,
+    private val lruHttpCache: ApolloHttpCache
 ) : HttpInterceptor {
 
   constructor(
       directory: File,
       maxSize: Long,
       fileSystem: FileSystem = FileSystem.SYSTEM,
-  ) : this(DiskLruHttpCache(fileSystem, directory, maxSize))
+  ): this(DiskLruHttpCache(fileSystem, directory, maxSize))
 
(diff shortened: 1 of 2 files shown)


libraries/apollo-http-cache/src/main/kotlin/com/apollographql/apollo3/cache/http/CachingHttpInterceptor.kt
```diff
@@ -152,11 +153,13 @@ class CachingHttpInterceptor internal constructor(
...
-    } catch (e: DateTimeParseException) {
-      0L
-    }
+    val servedDateMillis = response.headers.valueOf(CACHE_SERVED_DATE_HEADER)?.let {
+      try {
+        Instant.parse(it).toEpochMilli()
+      } catch (e: DateTimeParseException) {
+        0
+      }
+    } ?: 0
```
BoD, line 162: Actually I think the missing header can never happen except if my test

martinbonnin (commented):
This goes one step in the good direction but I think this API is still problematic. The fact that for an example, `ApolloClient.Builder.autoPersistedQueries()` is adding an interceptor under the hood is problematic. Code like [this](https://github.com/apollographql/apollo-kotlin/blob/7f4147fd8f294fa52214952403849352a7436bb3/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ClientCacheExtensions.kt#L122) is a code smell.

martinbonnin:
I'm thinking the order should always be:

* `NormalizedCacheInterceptor`
* `ApqInterceptor`
* `RetryInterceptor`
* `NetworkInterceptor`

? 

HTTP cache itself is a bit awkward because it spans 2 different domains (GraphQL + HTTP). From a quick cursory glance, it looks like it can be anywhere and is probably not going to disturb other `ApolloInterceptors`

BoD:
> I'm thinking the order should always be:
>
> - NormalizedCacheInterceptor
> - ApqInterceptor
> - RetryInterceptor
> - NetworkInterceptor

That sounds right to me 👍

martinbonnin (approved):

Issue #5852, opened 2024-04-25
Asynchronous caching and subsequent reads


rohandhruva:
### Question

We use asynchronous caching for almost all our operations. Are the cache writes vs subsequent reads guaranteed to be consistent or locked when using asynchronous caching? 

So for example:
- `client.query(foo)`, we got the response, but it's still writing to cache
- immediately follow it up with, say, `client.watch(foo)`

Is this second operation guaranteed to hit the cache (assuming it's the exact same operation)? Or can it miss the cache if the previous operation is not done writing to it? Or in other words, are cache reads locked against any ongoing writes?

And along similar lines, when using asynchronous caching, is there any way to get a signal that write to cache has completed? For example, if we use `.toFlow()`, we can get one `ApolloResponse<D>` followed by a completion event when it's done writing -- does that seem like something that can be done?

Thank you!

martinbonnin:
> Is this second operation guaranteed to hit the cache (assuming it's the exact same operation)? Or can it miss the cache if the previous operation is not done writing to it? Or in other words, are cache reads locked against any ongoing writes?

There is nothing like this. `writeToCacheAsynchronously` uses a separate coroutine in a separate scope so it's "fire & forget". It can take 1ms or 20s and subsequent reads happen concurrently.

> And along similar lines, when using asynchronous caching, is there any way to get a signal that write to cache has completed?

Nothing like this at the moment.

> does that seem like something that can be done?

We could swap the order of the two lines [here](https://github.com/apollographql/apollo-kotlin/blob/a8737d12d96089c0608f9878a47131f276ba5d4d/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt#L172). Meaning you could rely on the `Flow` terminating to know that the cache went through but that would remove the guarantee that "cache is written by the time you receive the event".

rohandhruva:
> We could swap the order of the two lines [here](https://github.com/apollographql/apollo-kotlin/blob/a8737d12d96089c0608f9878a47131f276ba5d4d/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt#L172). Meaning you could rely on the Flow terminating to know that the cache went through but that would remove the guarantee that "cache is written by the time you receive the event".

[...]

PR #5877, opened 2024-05-06
Complete flow only after cache writes have completed

rohandhruva:
This is a proposal PR that changes the `ApolloCacheInterceptor` flows to wait for the cache writes to complete before completing the flow. This is typically only useful when using `writeToCacheAsynchronously(true)`, so that you can get a signal of when the cache write has completed. 

Thinking about this a bit more, it gets a bit tricky when you use `execute()`: `singleSuccessOrException` uses `flow.toList()` which will always wait for flow completion by default. With this change, you effectively lose all benefits of asynchronous caching if you are also using `execute()`. 

Let me know your thoughts: maybe the `execute()` limitation means that we don't go ahead with this unless we find some workaround for `toList()`. 

Relates to #5852

### DIFF
--- a/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
+++ b/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/internal/ApolloCacheInterceptor.kt
@@ -28,20 +28,24 @@ import com.apollographql.apollo3.exception.apolloExceptionHandler
 import com.apollographql.apollo3.interceptor.ApolloInterceptor
 import com.apollographql.apollo3.interceptor.ApolloInterceptorChain
 import com.apollographql.apollo3.mpp.currentTimeMillis
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flatMapMerge
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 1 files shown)


BoD:
Thanks for this @rohandhruva !

Indeed as you noticed, the `execute()` case is problematic because the change would defeat the purpose of `writeToCacheAsynchronously`. I thought we could maybe get around that by using `isLast`, but unfortunately we can’t fully rely on it as there [can be false negatives](https://github.com/apollographql/apollo-kotlin/blob/b35c84399409ae4eb30707b7db16df02717d163b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/ApolloResponse.kt#L94).

Long term, maybe [Martin’s proposition](https://github.com/apollographql/apollo-kotlin/blob/main/design-docs/ApolloResult.md) of exposing events would help as we could emit an event to signal the end of async cache writes.

Back to your original question, your best bet would be to not use `writeToCacheAsynchronously` in specific cases - but I guess that’s probably what you currently do already?

rohandhruva:
> Back to your original question, your best bet would be to not use `writeToCacheAsynchronously` in specific cases - but I guess that’s probably what you currently do already?

Yes, that's what we do for now, especially for places that are only able to accept a single response. 

However, for UIs that are able to use a flow of responses, we would like the ability to get the response but still wait for the cache write, and then do something else in the `flow.onComplete { }` event with the guarantee that the cache was written to. That was the primary motivation behind this PR :) 

>  I thought we could maybe get around that by using `isLast`

That would be nice indeed :( Another option is that we can use `take(1)` for `execute()`, but we would need to hold the `singleSuccessOrException` logic somewhere in the stream of the flow, which does sound tricky. 

Let me think about this a bit more, and please let me know if you have any other suggestions I can try!

PR #5894, opened 2024-05-15
adding checks for json end_document in http transport

duzinkie:
the issue is discussed in: https://github.com/apollographql/apollo-kotlin/issues/5885

this is the recommended fix for the no http batching scenario


martinbonnin (approved):
Turns out v3 always consumes the full JSON reader (there's no `parseResponse` vs `toApolloResponse`) so this is good to go 👍 
Thanks again

PR #5895, opened 2024-05-15
adding checks for json end_document in http transport

duzinkie:
the issue is discussed in: https://github.com/apollographql/apollo-kotlin/issues/5885

this is the recommended fix for the no http batching scenario, for the main branch

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
@@ -45,10 +45,6 @@ internal object ResponseParser {
 
     jsonReader.endObject()
 
-    if (jsonReader.peek() != JsonReader.Token.END_DOCUMENT) {
-      println("Apollo: extra tokens after payload")
-    }
-
     return ApolloResponse.Builder(operation = operation, requestUuid = requestUuid ?: uuid4())
         .errors(errors)
         .data(data)
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
@@ -45,6 +45,10 @@ internal object ResponseParser {
 
     jsonReader.endObject()
 
+    if (jsonReader.peek() != JsonReader.Token.END_DOCUMENT) {
+      println("Apollo: extra tokens after payload")
+    }
+
     return ApolloResponse.Builder(operation = operation, requestUuid = requestUuid ?: uuid4())
         .errors(errors)
         .data(data)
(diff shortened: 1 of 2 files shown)


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
```diff
@@ -45,6 +45,10 @@ internal object ResponseParser {
 
     jsonReader.endObject()
 
+    if (jsonReader.peek() != JsonReader.Token.END_DOCUMENT) {
+      println("Apollo: extra tokens after payload")
+    }
```
martinbonnin, line 50: 2 things:

1. Can you move this to the `toApolloResponse()` function? I'd like to keep the possibility to read a single JSON object from `ResponseParser.parse` in the future
2. Let's make it an error? For this you can `throw JsonDataException("Expected END_DOCUMENT but was ${peek()}")`

martinbonnin:
Hi @duzinkie 👋 . Want to get this through the finish line?

duzinkie:
Hey @martinbonnin - sorry for the delays, I was on vacation - will be catching up on this this week.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/internal/ResponseParser.kt
duzinkie: I've made changes (hopefully) - as requested - lmk, and I'll push an equivalent to the 3.x branch.

Also - do you want the same applied to batching interceptor parsing - I'm guessing an exception should also be thrown there, not sure about whether code needs moving around?

martinbonnin: > I've made changes (hopefully) - as requested - lmk, and I'll push an equivalent to the 3.x branch.

Thanks, LGTM 👍 

> Also - do you want the same applied to batching interceptor parsing - I'm guessing an exception should also be thrown there, not sure about whether code needs moving around?

Yup, same exception, no need to move code around.

duzinkie: ok, Ive pushed a similar change to the other 3 prs.
not sure whether I should investigate integration test failures here?

martinbonnin: > ok, Ive pushed a similar change to the other 3 prs.

Thanks!

> not sure whether I should investigate integration test failures here?

Doesn't look related, I'll take a deeper look tomorrow

martinbonnin: Turns out it was related. `peek()` can throw but `toApolloResponse()` shouldn't. See https://github.com/apollographql/apollo-kotlin/pull/5895/commits/9ab202cc3536b505a8387f0e619db3660cf2ba62

I'll take it from there for this PR and the other 3. Thanks again for looking into this!

BoD (approved):

Issue #5901, opened 2024-05-16
A request with any enum having `rawValue = null` hangs forever


japhib:
### Version

4.0.0-beta.6

### Summary

Enums defined in the provided graphQL schema are defined as pseudo-enums, actually classes wrapping a String member called `rawValue`. While it's possible to use the provided enum constants from the class, you can also create your own, possibly passing in `null`.

Example enum in the graphql schema:
```graphql
enum StatusEnum {
    GOOD
    BAD
}
```

Generated `StatusEnum` class:

```java
public class StatusEnum {
  public static EnumType type = new EnumType("StatusEnum", Arrays.asList("GOOD", "BAD"));
  public static StatusEnum GOOD = new StatusEnum("GOOD");
  public static StatusEnum BAD = new StatusEnum("BAD");

  public String rawValue;

  public StatusEnum(String rawValue) {
    this.rawValue = rawValue;
  }

  public static StatusEnum safeValueOf(String rawValue) {
    switch(rawValue) {
      case "GOOD": return StatusEnum.GOOD;
      case "BAD": return StatusEnum.BAD;
      default: return new StatusEnum.UNKNOWN__(rawValue);
    }
  }
// ... lots of other stuff
}
```
A good usage of StatusEnum would be to use the constants such as `StatusEnum.GOOD`, or `StatusEnum.BAD`. Another valid use of `StatusEnum` would be to parse the string `"GOOD"` into `StatusEnum.GOOD` using `StatusEnum.safeValueOf("GOOD")` or even `new StatusEnum("GOOD")`.

An invalid, yet allowed, usage of StatusEnum is to pass in `null` to the constructor: `new StatusEnum(null)`. This isn't checked for invalidity at any point, even up through making the GraphQL call:

```java
var query = GetThingQuery.builder()
            // !!! Set the input enum to null !!!
            .something(new StatusEnum(null))
            .build()

// hangs forever!!
var result = Rx3Apollo.single(apolloClient.query(query), BackpressureStrategy.BUFFER).blockingGet();
```

When used in conjunction with `Rx3Apollo.single(...).blockingGet()`, a NullPointerException is thrown **in another thread** when attempting to serialize this value:
```
Exception in thread "Apollo Dispatcher" java.lang.NullPointerException: Parameter specified as non-null is null: method com.apollographql.apollo3.api.json.internal.FileUploadAwareJsonWriter.value, parameter value
[...]

PR #5904, opened 2024-05-17
Add a null-check to java enum safeValueOf

BoD:
Fix for #5901

- Also make the constructor of the class and of `UNKNOWN__` nullable
- Also call `super()` in `equals()` and `hashCode()`


libraries/apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/java/operationBased/case_sensitive_enum/type/Enum.java.expected
```diff
@@ -60,15 +61,15 @@ public class Enum {
         return true;
       }
       if (o instanceof UNKNOWN__) {
-        return true;
+        return super.equals(o);
```
martinbonnin, line 64: What is the rationale for this?

libraries/apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/java/operationBased/case_sensitive_enum/type/Enum.java.expected
```diff
@@ -60,15 +61,15 @@ public class Enum {
...
       }
       return false;
     }
 
     @Override
     public int hashCode() {
       if (!$hashCodeMemoized) {
-        int __h = 1;
+        int __h = super.hashCode();
```
martinbonnin, line 72: Same here. Do we really need that change?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/schema/EnumAsClassBuilder.kt
```diff
@@ -86,7 +86,7 @@ internal class EnumAsClassBuilder(
                 .returns(selfClassName)
                 .addCode(
                     CodeBlock.builder()
-                        .beginControlFlow("switch($rawValue)")
+                        .beginControlFlow("switch ($T.requireNonNull($rawValue))", JavaClassNames.Objects)
```
martinbonnin, line 89: TIL about `Objects.requireNonNull`. Should we use that instead of https://github.com/apollographql/apollo-kotlin/blob/39df250747de5e3028dc6c11076a2315984d05d5/libraries/apollo-api-java/src/main/java/com/apollographql/apollo3/api/java/Assertions.java#L4?

martinbonnin (commented):
A few comments. Also, can we do the same for Kotlin codegen?

libraries/apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/java/operationBased/case_sensitive_enum/type/Enum.java.expected
BoD: I noticed that without this, 2 unknown enums with a different rawValue would still be equal, since that field is not included in the comparison. The super version does include the field in the comparison.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/schema/EnumAsClassBuilder.kt
BoD: I was almost sure we had something like that but couldn't find it anymore 😅. Ours is a bit nicer since you can pass a message - so I'd keep it.

BoD:
> can we do the same for Kotlin codegen?

You're right! We can make the class constructor private. I'd keep the Kotlin `UNKNOWN__` constructor public though as it's a nice way to signal a warning when using it, thanks to the opt-in annotation. On the other hand, it's still possible to call `safeValueOf` without any warning... Maybe `safeValueOf` should be annotated too?

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/codegen/java/schema/EnumAsClassBuilder.kt
martinbonnin: Can you add a comment on ours? 

```
// A version of Objects.requireNonNull that allows a customized message
```

libraries/apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/java/operationBased/case_sensitive_enum/type/Enum.java.expected
martinbonnin: Shouldn't it compare `rawValue`s instead? I would expect `Object.equals()` to do referential equality only?

martinbonnin:
>  I'd keep the Kotlin UNKNOWN__ constructor public though as it's a nice way to signal a warning when using it, thanks to the opt-in annotation

I'd try to keep the symmetry with Java as much as possible.

>  On the other hand, it's still possible to call safeValueOf without any warning... Maybe safeValueOf should be annotated too?

I'd go for that. And rename the `@ApolloEnumConstructor` to `@ApolloUnknownEnum` or something like that.

libraries/apollo-compiler/src/test/graphql/com/example/case_sensitive_enum/java/operationBased/case_sensitive_enum/type/Enum.java.expected
BoD: You're totally right! I assumed equals was implemented in the base class but it's not, so calling super doesn't work. In the Kotlin version the equals/hashCode is ad-hoc, I'll do the same in Java.

BoD:
Turns out, actually in Kotlin we can't make the `UNKOWN__` constructor private, because unlike Java, outer classes can't access private constructors of inner classes. I kept the optin annotation there, and also applied it to `safeValueOf`.

martinbonnin (approved):

Issue #5887, opened 2024-05-15
[IJ Plugin] Warn when .graphqlrc files are present


BoD:
When using a v4 project, the location of the schema and operation files is automatically contributed to the GraphQL IJ Plugin by the Apollo plugin. Therefore `.graphqlrc` (and variants like `graphql.config.yml`, etc.) aren't necessary anymore, and in fact having  them in the project is probably confusing.

A warning explaining this, with a quick fix to delete the files could be implemented.

This could also be part of the migration 3.x -> 4.x

BoD:
Fix is released in 4.0.0-beta.7

PR #5908, opened 2024-05-20
[IJ Plugin] Add inspection to warn about the presence of a GraphQL config file

BoD:
Also remove GraphQL config files when  migrating v3 -> v4.

Fix for #5887 

Screenshot 2024-05-20 at 19 44 13
Screenshot 2024-05-20 at 19 43 56

### DIFF
--- a/intellij-plugin/src/main/resources/META-INF/plugin.xml
+++ b/intellij-plugin/src/main/resources/META-INF/plugin.xml
@@ -185,7 +185,7 @@
         groupKey="inspection.group.graphql.apolloKotlin"
         key="inspection.graphQLConfigFilePresent.displayName"
         enabledByDefault="true"
-        level="ERROR"
+        level="WARNING"
     />
 
     <annotator
--- a/intellij-plugin/src/main/resources/META-INF/plugin.xml
+++ b/intellij-plugin/src/main/resources/META-INF/plugin.xml
@@ -177,6 +177,39 @@
         level="WARNING"
     />
 
+    <!-- "GraphQL config file present" inspection -->
+    <!--suppress PluginXmlCapitalization, PluginXmlExtensionRegistration -->
+    <localInspection
+        implementationClass="com.apollographql.ijplugin.inspection.ApolloGraphQLConfigFilePresentInspection"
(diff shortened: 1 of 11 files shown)


intellij-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -173,6 +173,39 @@
...
 
+    <!-- "GraphQL config file present" inspection -->
+    <!--suppress PluginXmlCapitalization, PluginXmlExtensionRegistration -->
+    <localInspection
+        implementationClass="com.apollographql.ijplugin.inspection.ApolloGraphQLConfigFilePresentInspection"
+        groupPathKey="inspection.group.graphql"
+        groupKey="inspection.group.graphql.apolloKotlin"
+        key="inspection.graphQLConfigFilePresent.displayName"
+        enabledByDefault="true"
+        level="ERROR"
```
martinbonnin, line 184: If it is not needed but doesn't harm it should be a warning. 

Suggestion:

```
Apollo Kotlin 4 uses the Gradle tooling API to retrieve the GraphQL configuration and doesn't use the GraphQL config file.
```

If it's an actual error, ERROR is fine of course but sounds like there is a disconnect between the message and the level here.

intellij-plugin/src/main/resources/inspectionDescriptions/ApolloGraphQLConfigFilePresent.html
```diff
@@ -0,0 +1,15 @@
+<html>
+<body>
+Suggests to delete GraphQL config files on projects using Apollo Kotlin 4.
+<p>
+    It is no longer useful to have a GraphQL config file, as the Apollo plugin will automatically provide the location of your
+    project's operations and schema files to the GraphQL plugin, <br>
```
martinbonnin, line 6:
```suggestion
    It is no longer useful to have a GraphQL config file, as the Apollo plugin automatically provides the location of your
    project's operations and schema files to the GraphQL plugin, <br>
```

martinbonnin (approved):

PR #5916, opened 2024-05-23
Mark `ApolloGenerateIrOperationsTask` as Cacheable and sort `ir.json` entries

rohandhruva:
After enabling `generateApolloMetadata`, we started seeing gradle build cache misses. Non-stable output in `ir.json` and `metadata.json` was invalidating build cache for us. 

While it might be a good idea to find a way to make those deterministic, for now it seems like marking the `ApolloGenerateIrOperationsTask` as a `@CacheableTask` seems enough. I also tried doing the same for `ApolloGenerateOptionsTask`, but it was not quite as straightforward. 

For the future, it seems like there's [no native support](https://github.com/Kotlin/kotlinx.serialization/issues/121#issuecomment-1019916245) in kotlinx-serialization to generate deterministic output by, for example, sorting the json keys / lists before writing them to a file, but we could look into using a `JsonTransofrmingSerializer` to do something like that. 

(EDIT: I added back sorting of the keys in `ir.json`. It's not complete, though. Without the sorting, the cache key for `ApolloGenerateIrOperationsTask` can still be invalidated.)

Relates to #5917.

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/CodegenMetadata.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo3/compiler/CodegenMetadata.kt
@@ -28,7 +28,7 @@ class CodegenMetadata(
     }
     return CodegenMetadata(
         targetLanguage = targetLanguage,
-        entries = entries + other.entries
+        entries = (entries + other.entries).sortedBy { it.key.id }
     )
   }
 }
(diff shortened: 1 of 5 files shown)


martinbonnin:
I added a test and made the tasks cacheable [here](https://github.com/apollographql/apollo-kotlin/pull/5918). 

I _think_ kotlinx.serialization output should be predictable, same as most of the data structures we use in the codegen too (Kotlin uses `LinkedHashMap` and `LinkedHashSet` that have predictable ordering).

rohandhruva:
Superseded by PR#5918 and PR#5919.

PR #5921, opened 2024-05-27
Make enums as sealed classes generate sealed interfaces

BoD:
A follow-up to #5904, this makes the Kotlin version closer to the Java one: unknown enums can only be created through `safeValueOf`.


martinbonnin (commented):
Sealed interfaces are supported since Kotlin 1.5 so it should be ok

We might not need the OptIn marker after all. We added it initially because the `UNKNOWN__` constructor was public but we might want to wait a little bit with this usage see if we really need it.

This sadly doesn't solve the case where GraphQL enums are generated as Kotlin enums. We should probably make `sealedClassesForEnumsMatching.set(listOf(".*")` the default but can be done in a future PR.

This is source breaking change for anyone using the `UNKNOWN__()` constructor directly. For all callers, it is binary breaking. If you ship a library that exposes an Apollo generated enum, it is a binary breaking change.

martinbonnin (approved):

PR #5923, opened 2024-05-27
[gradle-plugin] Use lazy APIs

### LINKED ISSUE #5922: Apply com.apollographql.apollo3 plugin will break the dependencies in Kotlin Multiplatform according to IDEA
### Version

4.0.0-beta.6

### Summary

 When the plugin is applied, all dependencies used  in `commonMain` are not available in `commonTest`is applied

I created a repository with the documentation linked to the problem https://github.com/Hansanto/apollo-issue

### Logs

_No response_

martinbonnin:
Using `.any {}` was somewhat eagerly configuring things and that messed up KGP.

Fixes #5922

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
@@ -265,23 +265,24 @@ abstract class DefaultApolloExtension(
   }
 
   private fun maybeLinkSqlite() {
-    val doLink = when (linkSqlite.orNull) {
+    when (linkSqlite.orNull) {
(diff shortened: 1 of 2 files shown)


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo3/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -600,7 +601,7 @@ abstract class DefaultApolloExtension(
         codegenMetadataConsumerConfiguration.dependencies.add(it)
       }
 
-      val pending = pendingDownstreamDependencies.get(name)
+      val pending = pendingDownstreamDependencies.get(service.name)
```
martinbonnin, line 604: This is an unrelated fix that I discovered by serendipity for `isADependencyOf(bidirectional: true)`

BoD (approved):
👏

PR #5930, opened 2024-05-31
[runtime] Remove multidex dependency

martinbonnin:
Turns out we don't even need it for integration tests because we have minSdk >= 21. Thanks @edenman for catching this!


BoD (approved):
🧹

PR #5931, opened 2024-06-03
Bump languageVersion/apiVersion to Kotlin 2.0

BoD:
Also, I tried to enable KSP 2 but it crashes due to https://github.com/google/ksp/issues/1823 (which needs a fix that will be in Kotlin 2.1).

### DIFF
--- a/build-logic/src/main/kotlin/api.kt
+++ b/build-logic/src/main/kotlin/api.kt
@@ -1,15 +1,20 @@
-
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.jvm.tasks.Jar
 import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension
 import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin
 import org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask
 
 class AndroidOptions(
-    val withCompose: Boolean
+    val withCompose: Boolean,
+)
+
+class KotlinCompilerOptions(
+    val apiVersion: KotlinVersion = KotlinVersion.KOTLIN_2_0,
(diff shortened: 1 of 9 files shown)


libraries/apollo-debug-server/build.gradle.kts
```diff
@@ -123,3 +126,11 @@ dependencies {
     to.attribute(artifactType, ArtifactTypeDefinition.DIRECTORY_TYPE)
   }
 }
+
+// Can't use apiVersion KOTLIN_2_0 when using languageVersion KOTLIN_1_9, which is the case here because we're using KSP 1
```
martinbonnin, line 130: Where is `languageVersion` set to `KOTLIN_1_9`?  Is KSP overriding our default?

BoD: I'm not actually sure, but guessing yes, KSP must automatically switch to K1. Interestingly, if I set  `languageVersion.set(KotlinVersion.KOTLIN_1_9)` here, I'm getting an error:
```
Inconsistent settings for Kotlin source sets: 'jvmMain' depends on 'jvmCommonMain'
  'jvmMain': language version is KOTLIN_1_9
  'jvmCommonMain': language version is KOTLIN_2_0
  The language version of the dependent source set must be greater than or equal to that of its dependency.
```

build-logic/src/main/kotlin/api.kt
```diff
@@ -1,15 +1,20 @@
...
 import org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask
 
 class AndroidOptions(
-    val withCompose: Boolean
+    val withCompose: Boolean,
+)
+
+class KotlinCompilerOptions(
+    val apiVersion: KotlinVersion = KotlinVersion.KOTLIN_2_0,
+    val languageVersion: KotlinVersion = KotlinVersion.KOTLIN_2_0,
```
martinbonnin, line 17: Nit: is there any occurrence where we're not setting the same value for both parameters? I would probably have gone with a single parameter but 2 work too 👍

martinbonnin (approved):

build-logic/src/main/kotlin/api.kt
BoD: No you're totally right :) Changed it to a single `version` parameter

martinbonnin (approved):

PR #5934, opened 2024-06-03
Prepare apollo-testing-support to new MockServer

martinbonnin:
Moving `MockServer` to a separate repo/packageName would make it a breaking change for `apollo-testing-support`. 

Instead, `apollo-testing-support` is going to keep using the current (soon to be deprecated) `MockServer` for now. But we don't want to pull that in other parts of the repo.

This PR tries to contain the blast radius of MockServer. It makes most of `apollo-testing-support` deprecated:

* all the  things like `HostFileSystem`, `Channel.awaitItem()`, etc... were meant to be used from our own tests and the plan is to simply remove them.
* functions that have `MockServer` could be somewhat useful but there are ways to replace relatively easily.

### DIFF
--- a/libraries/apollo-runtime/src/commonTest/kotlin/test/network/NetworkMonitorTest.kt
+++ b/libraries/apollo-runtime/src/commonTest/kotlin/test/network/NetworkMonitorTest.kt
@@ -1,18 +1,21 @@
 package test.network
 
 import app.cash.turbine.test
+import com.apollographql.apollo3.ApolloClient
 import com.apollographql.apollo3.api.ApolloRequest
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.Operation
 import com.apollographql.apollo3.exception.ApolloNetworkException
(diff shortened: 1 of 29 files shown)


libraries/apollo-runtime/src/commonTest/kotlin/test/network/NetworkMonitorTest.kt
```diff
@@ -107,3 +111,19 @@ class NetworkMonitorInterceptor(private val networkMonitor: NetworkMonitor): Apo
...
+  MockServer().use { mockServer ->
+    ApolloClient.Builder()
+        .serverUrl(mockServer.url())
+        .apply(clientBuilder)
+        .build()
+        .use {apolloClient ->
+          MockServerTest(mockServer, apolloClient, this).block()
+        }
+  }
+}
```
martinbonnin, line 129: This code is duplicated a bunch of times. It's a bit sad but it's very trivial code so it shouldn't be too hard to maintain.

BoD (approved):
👍

PR #5935, opened 2024-06-03
[testing] Simplify our implementation of runTest

martinbonnin:
Always use `kotlinx.coroutines.test.runTest {}` as underlying mechanism for our tests. This removes our K2 workaround and simplifies the setup a bunch. 

Our own `runTest {}` is still there for:

* `before`/`after` callbacks which we should probably remove at some point. I started doing it but we have a loooooot of tests and ultimately gave up. Will refactor them progressively.
* never skipping the delays. We use delays in a few situations like batching/websockets. See https://github.com/Kotlin/kotlinx.coroutines/issues/3179


BoD (approved):
🧹👍

PR #5939, opened 2024-06-04
Use com.apollographql.mockserver.MockServer

martinbonnin:
2 main changes:

* `com.apollographql.apollo3.mockserver` -> `com.apollographql.mockserver`
* added `apollo-mockserver` as a dependency instead of relying on the transitive dependencies of `apollo-testing-support`


BoD (approved):
👍

PR #5954, opened 2024-06-07
Documentation tweaks and updates for v4

BoD:


docs/source/advanced/plugin-configuration.mdx
```diff
@@ -79,12 +77,14 @@ apollo {
 }
 ```
 
+
+> By default, Apollo Kotlin combines all `*.[graphqls|json|sdl]` files in `src/main/graphql/` (Android/JVM projects) or `src/commonMain/graphql/` (multiplatform projects).
+
```
martinbonnin, line 82: Do we need that here? That feels a bit out of place

docs/source/caching/normalized-cache.mdx
```diff
@@ -172,31 +176,29 @@ val response = apolloClient.query(query)
...
-      // ...
+    // ...
   }
 ```
+Note that cache misses **will** emit a response with a non null `.exception`, meaning that some of these policies can emit multiple values:
+- `CacheFirst` can emit 1 or 2 values
+- `NetworkFirst` can emit 1 or 2 values
+- `CacheAndNetwork` will emit 2 values
+
+For that reason you should use `.toFlow()` rather than `.execute()` when using these policies.
```
martinbonnin, line 193: Mixed feelings about this. `.execute()` works well for `CacheFirst` and `NetworkFirst` and AFAIK, it's one of the main usages today. 
`CacheAndNetwork` is different in that `execute()` doesn't work at all. Ultimately, we might want to deprecate `execute()` altogether but until we do, I wouldn't recommend against it for `CacheFirst` and `NetworkFirst`

docs/source/caching/normalized-cache.mdx
```diff
@@ -172,31 +176,29 @@ val response = apolloClient.query(query)
...
 
 As with `normalizedCache(NormalizedCacheFactory)`, `fetchPolicy(FetchPolicy)` is an extension function on `ApolloClient.Builder()`, so you need `apollo-normalized-cache` in your classpath for this to work.
 
 Because the normalized cache deduplicates data, it enables you to react to cache changes. You do this with `watchers` that listen for cache changes. [Learn more about query watchers.](./query-watchers/)
 
-## Specifying cache IDs
+## Specifying cache keys
```
martinbonnin, line 199: Sooooo the docs are quite heterogeneous about this.... 

Web has mostly "cache IDs", iOS has both "cache IDs" and "cache keys (not sure what the difference is). Given that most of the normalized cache documentation/content out there is for web (therefore using "cache IDs"), I would exercise caution before changing this.

docs/source/caching/normalized-cache.mdx
```diff
@@ -243,31 +245,23 @@ We're now caching two identical entries for the same `Author` object! This is un
...
-Call below to clear the cache of all entries, note that retrieving `apolloClient.apolloStore` will throw an exception if a cache has not been configured on the client.
-
-```kotlin
-try {
-    apolloClient.apolloStore.clearAll()
-} catch (e: IllegalStateException) {
-    println("Failed to clear Apollo normalized cache because none was found")
-}
-```
+Call `apolloClient.apolloStore.clearAll()` to clear the cache of all entries. Note that calling `apolloClient.apolloStore` will throw an exception if a cache has not been configured.
```
martinbonnin, line 267:
```suggestion
Call `apolloClient.apolloStore.clearAll()` to clear the cache of all entries. Note that calling `apolloClient.apolloStore` throws an exception if a cache has not been configured.
```

docs/source/config.json
```diff
@@ -39,14 +38,14 @@
       "Error handling": "/essentials/errors",
       "Custom scalars": "/essentials/custom-scalars",
       "Fragments": "/essentials/fragments",
-      "@defer support (experimental)": "/fetching/defer",
+      "@defer support": "/fetching/defer",
```
martinbonnin, line 41: I think we still want `@defer` to be experimental on mobile until the spec is merged

martinbonnin: Oh, I see below you removed the (experimental) from all left panel items. Is that on purpose?

docs/source/essentials/subscriptions.mdx
```diff
@@ -162,3 +165,9 @@ Like queries, subscriptions support partial responses with GraphQL errors, which
 Network errors terminate the `Flow`, and you need to retry to get new updates. Depending on the situation, retrying might open a new WebSocket or restart the subscription.
 
 See also [this section](../advanced/websocket-errors) about WebSocket errors handling.
+
+## Experimental WebSocket APIs
+
+Starting with 4.0.0, Apollo Kotlin provides new experimental WebSocket APIs that aim to replace the current implementation in the future.
+
+If you are interested in trying them out, please consult [the dedicated section](../advanced/experimental-websockets).
```
martinbonnin, line 173: 👍

martinbonnin (approved):
A very welcome improvement 👍  A few nitpick and one more fundamental question about naming of "cache keys". I would probably keep the docs untouched in the current state of things until we have reached some iOS/web/kotlin consensus.

docs/source/advanced/http-engine.mdx
```diff
@@ -4,15 +4,39 @@ title: Using a custom HTTP client
...
+
+And configure your `ApolloClient` to use the Ktor engine:
+
+```kotlin
+val client = ApolloClient.Builder()
+  .serverUrl("https://example.com/graphql")
+  // Create a new Ktor engine
+  .httpEngine(KtorHttpEngine())
+  // Or, if you want to reuse an existing Ktor client
+  .ktorClient(myKtorClient)
```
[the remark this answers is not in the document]
BoD: That's a good point. Removing ambiguity anywhere we can will probably be a win.

docs/source/advanced/plugin-configuration.mdx
BoD: Good catch, I probably pasted by mistake

docs/source/caching/normalized-cache.mdx
BoD: Totally right! Replacing with:

> Note: `.execute()` filters out cache or network errors to return a single success response, however since `CacheAndNetwork` can emit 2 success responses, you should use `.toFlow()` when using that policy.

docs/source/config.json
martinbonnin: 👍 web and ios don't have (experimental) in the menu. LGTM ✅

martinbonnin (approved):

PR #5963, opened 2024-06-13
Bump develocity to 3.17.5

BoD:


martinbonnin (approved):
👍 
Note to self: we also need to update a bunch of ENV variables that trigger warnings in CI

PR #5968, opened 2024-06-14
[❄️flakes] Robustify TCP server

martinbonnin:
I'm not 100% sure what to do in that case. Looks like this is happening when shutting down the server. For now I'm just ignoring the error.

See https://github.com/ktorio/ktor/pull/4076


BoD (approved):
👍

Issue #5964, opened 2024-06-13
Initializing the Normalized Cache with `CacheKeyResolver` & Relay Style Pagination Constructs.


ArjanSM:
### Question

Hi Team, 👋 
In continuation to the issue[#5666](https://github.com/apollographql/apollo-kotlin/issues/5666), I noticed that in the latest v4.0.0 Snapshot the normalized cache cannot be initialized using `CacheKeyResolver`s along with the `MetadataGenerator`, `RecordMerger`, `FieldNameGenerator`and `EmbeddedFieldsProvider`. 
Are there any plans to introduce an API which can initialize the Normalized Cache using the CacheKeyResolvers and the Relay Style Pagination constructs?

Thanks

BoD:
Hi!

In the incubating version we've started transitioning to `ApolloResolver` to replace `CacheResolver`. It is the same thing but with a slightly different API that will help with evolving the library in a non-breaking way. The version of `ApolloStore()` that accepts the `FieldNameGenerator` etc. is expecting an `ApolloResolver`.

However there is no `ApolloResolver` version of `CacheKeyResolver` - we should definitely add it. In the meantime you can transform your existing resolver into an `ApolloResolver` with something like:

```kotlin
class ApolloCacheKeyAdapter(private val cacheKeyResolver: CacheKeyResolver) : ApolloResolver {
  override fun resolveField(context: ResolverContext): Any? {
    return cacheKeyResolver.resolveField(variables = context.variables, parent = context.parent, parentId = context.parentKey, field = context.field)
  }
}
```

and pass `ApolloCacheKeyAdapter(yourExistingCacheKeyResolver)` to `ApolloStore()`.

BoD:
Done in https://github.com/apollographql/apollo-kotlin/pull/5970. Additional work on the incubating cache now happens in https://github.com/apollographql/apollo-kotlin-normalized-cache-incubating.

PR #5970, opened 2024-06-14
Add a CacheKeyApolloResolver based on ApolloResolver (incubating)

BoD:
Related to #5964.

Also tweaked the doc/kdoc a little bit.


martinbonnin (approved):
Oh my I forgot we had both `ApolloResolver` and `CacheResolver`... This is becoming quite hairy but besides that, LGTM

BoD:
> Oh my I forgot we had both ApolloResolver and CacheResolver... This is becoming quite hairy

Yeah 😅 It will probably make sense to break compatibility at some near point and when that comes, we can keep only one of them.

martinbonnin:
Yea, it's -incubating + `@ApolloExperimental`. At some point it's too much detail.

PR #5972, opened 2024-06-16
SqlNormalizedCacheFactory make sqldriver public to support Sqlcipher data encryption.

yuvaraj119:


BoD (approved):
Thanks a lot! 🙏 

I've update the API dump and added the API dependency to SqlDelight.

martinbonnin (approved):
See https://github.com/apollographql/apollo-kotlin/pull/4806 for the -incubating version

PR #5980, opened 2024-06-18
Cleanup mpp utils

martinbonnin:
* replace `currentTimeMillis()`  with `markNow()` where possible. 
* move `platform()` and `currentThreadId()` to `apollo-testing-support`. 

`apollo-mpp-utils` now only contains `currentTimeMillis()` that we should probably remove from `HttpInfos`


BoD (approved):
👍

PR #5987, opened 2024-06-21
Remove incubating compose support

BoD:
It now resides in `com.apollographql.compose:compose-support` and `com.apollographql.compose:compose-paging-support`


martinbonnin (approved):
🧹

PR #5989, opened 2024-06-24
Add ApolloClient.Builder.retryOnErrorInterceptor

### LINKED ISSUE #5982: RetryOnNetworkErrorInterceptor should be configurable
### Use case

We currently have our own interceptor to handle retries on network errors, but 
1) it also retries if the http status code is 5xx
2) we'd like the ability to configure the backoff duration

Not a huge deal (we can obviously keep using our own) but I'd like to switch to using NetworkMonitor+failFastIfOffline+retryOnError if possible

_No response_

martinbonnin:
Closes #5982

### DIFF
--- a/libraries/apollo-runtime/src/androidInstrumentedTest/kotlin/instrumented/NetworkMonitorTest.kt
+++ b/libraries/apollo-runtime/src/androidInstrumentedTest/kotlin/instrumented/NetworkMonitorTest.kt
@@ -6,6 +6,7 @@ import com.apollographql.apollo3.api.http.HttpResponse
 import com.apollographql.apollo3.exception.ApolloNetworkException
 import com.apollographql.mockserver.assertNoRequest
 import com.apollographql.mockserver.enqueueString
+import com.apollographql.apollo3.interceptor.RetryOnErrorInterceptor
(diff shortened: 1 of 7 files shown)


libraries/apollo-runtime/src/androidInstrumentedTest/kotlin/instrumented/NetworkMonitorTest.kt
```diff
@@ -37,7 +38,9 @@ class NetworkMonitorTest {
   @Test
   fun test() = mockServerTest(
       clientBuilder = {
-        networkMonitor(NetworkMonitor(InstrumentationRegistry.getInstrumentation().context))
+        retryOnErrorInterceptor(
+            RetryOnErrorInterceptor(NetworkMonitor(InstrumentationRegistry.getInstrumentation().context))
+        )
```
martinbonnin, line 43: I removed `ApolloClient.Builder.networkMonitor(NetworkMonitor)` as part of this PR. It makes the call site a bit more verbose but allow for better locality as the `NetworkMonitor` isn't used outside the `RetryOnErrorInterceptor`

BoD (approved):
👍

PR #5997, opened 2024-06-25
Point to the dedicated Java support doc

BoD:


docs/source/migration/4.0.mdx
```diff
@@ -203,8 +203,6 @@ Over the years, a lot of support functionality was added alongside `apollo-api`
 
 Moving forward, some artifacts are released separately and have new Maven coordinates, GitHub repositories and package names. Others are deprecated.
 
-For the initial v4 release, all the artifacts are kept with deprecation warnings. They will be removed in the future.
```
martinbonnin, line 206: Non-incubating artifacts are still around IIRC?

BoD: Yes except `apollo-cli`, but maybe it's OK to leave the sentence anyway.

martinbonnin: apollo-cli was `apollo-cli-incubating` in main (but maybe not in v3 in which case I can probably bring it back)

martinbonnin: Or maybe we just do "most artifacts are kept...."?

martinbonnin (approved):

PR #5999, opened 2024-06-26
Add bump kotlin shadow job

BoD:


martinbonnin:
Looks like this duplicates a lot of the "pr" workflow? Could you instead have a branch with the latest Kotlin version that gets recreated every night and triggers the "pr" workflow manually using a workflow_dispatch event? 

A question is whether we want the shadow branch to have a PR always open. Pro: makes us see when it fails. Con: creates noise in the PR list.

Also, what about fetching the latest version automatically? Possibly even using nightlies?

BoD:
> Looks like this duplicates a lot of the "pr" workflow? Could you instead have a branch with the latest Kotlin version that gets recreated every night and triggers the "pr" workflow manually using a workflow_dispatch event?

Good point! I like the idea of having a (draft) PR open so we can see if it's "red" :) But I agree it's a bit of noise too. I'll try that route, and if we're tired of seeing it, we can reevaluate this.

> what about fetching the latest Kotlin version automatically? Possibly even using nightlies?

👀  interesting idea, let me see...

martinbonnin:
> I'll try that route, and if we're tired of seeing it, we can reevaluate this.

Sounds good 👍 .

> what about fetching the latest Kotlin version automatically? Possibly even using nightlies?

Thinking more about this, I think nightlies make more sense if we want to send timely feedback. We could have 2 PRs but for the sake of simplicity, I think I'd prefer just one.

martinbonnin (approved):

PR #6001, opened 2024-06-26
Migration guide tweaks

BoD:
A few minor tweaks + a section about the package change.

### DIFF
--- a/docs/source/migration/4.0.mdx
+++ b/docs/source/migration/4.0.mdx
@@ -25,56 +25,7 @@ Apollo Kotlin 4 ships with a companion [Android Studio plugin](../testing/androi
 
 It automates most of the API changes but cannot deal with behavior changes like error handling. 
 
-We recommend using the plugin to automate the repetitive tasks but still go through this document for the details. 
-
-## Group id / plugin id / package name
-
-Apollo Kotlin 4 uses a new identifier for its package name, Gradle plugin id, and maven group id: **`com.apollographql.apollo`**.
-
-This also allows to run versions 4 and 3 side by side if needed.
-
-In most cases, you can update the identifier throughout your project by performing a find-and-replace and replacing `com.apollographql.apollo3` with `com.apollographql.apollo`.
-
-### Group id
-
-The maven group id used to identify Apollo Kotlin 4 [artifacts](https://repo1.maven.org/maven2/com/apollographql/apollo/) is `com.apollographql.apollo`:
-
-```kotlin
-// Replace:
-implementation("com.apollographql.apollo3:apollo-runtime:3.8.4")
-
-// With:
-implementation("com.apollographql.apollo:apollo-runtime:4.0.0")
-```
-
-### Gradle plugin id
-
-The Apollo Kotlin 4.0 Gradle plugin id is `com.apollographql.apollo`:
-
-```kotlin
-// Replace:
-plugins {
-  id("com.apollographql.apollo3") version "3.8.4"
-}
-
-// With:
-plugins {
-  id("com.apollographql.apollo") version "4.0.0"
-}
-```
-
-### Package name
-
-Apollo Kotlin 4 classes are imported from the `com.apollographql.apollo` package:
-
-```kotlin
-// Replace:
-import com.apollographql.apollo3.ApolloClient
-
-// With:
-import com.apollographql.apollo.ApolloClient
-```
-
+We recommend using the plugin to automate the repetitive tasks but still go through this document for the details.
 
 ## Error handling
 
--- a/docs/source/migration/4.0.mdx
+++ b/docs/source/migration/4.0.mdx
@@ -5,7 +5,7 @@ description: From 3.x
 
 <Note>
 
-**Version 4 is currently under development and this page itself is a work in progress.**
+**Version 4.0 is currently under development and this page itself is a work in progress.**
 
 </Note>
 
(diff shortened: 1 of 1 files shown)


docs/source/migration/4.0.mdx
```diff
@@ -5,31 +5,82 @@ description: From 3.x
...
 
-Apollo Kotlin 4 ships with a companion [Android Studio plugin](../testing/android-studio-plugin#migration-helpers) that automates most of the migration. 
+Apollo Kotlin 4.0 ships with a companion [Android Studio plugin](../testing/android-studio-plugin#migration-helpers) that automates most of the migration. 
 
-It automates most of the API changes but cannot deal with behaviour changes like error handling. 
+It automates most of the API changes but cannot deal with behavior changes like error handling. 
 
 We recommend using the plugin to automate the repetitive tasks but still go through this document for the details. 
 
+## Package name / group id / plugin id
```
BoD, line 30: Maybe this section should be merged with the "Maven artifacts reorganization" below.

docs/source/migration/4.0.mdx
```diff
@@ -5,31 +5,82 @@ description: From 3.x
 
 <Note>
 
-**Version 4 is currently under development and this page itself is a work in progress.**
+**Version 4.0 is currently under development and this page itself is a work in progress.**
```
martinbonnin, line 8: I think I'd stick with `Apollo Kotlin 4` and `Apollo Kotlin 3`. It's generic and will age better. 

In places where we don't want to repeat ourselves, I'd go with `3.x` and `4.x`. I don't see a strong reason to make a distinction between `3.0` and `3.x`? (or `4.0` and `4.x`?)

We don't technically have a `4.0` release. I guess it's ok in the url because we've always done that but would otherwise avoid it.

docs/source/migration/4.0.mdx
```diff
@@ -5,31 +5,82 @@ description: From 3.x
...
 
 </Note>
 
 Apollo Kotlin 3.0 was a major rewrite of Apollo in Kotlin multiplatform.
 
 Apollo Kotlin 4.0 focuses on tooling, stability and fixing some API regrets that came with 3.x.
 
-Because most of the common APIs stayed the same, we [kept the package name unchanged](https://github.com/apollographql/apollo-kotlin/issues/4710). Apollo Kotlin 4.0 removes some deprecated symbols. We strongly recommend removing deprecated usages before migrating to 4.0.  
+While most of the common APIs stayed the same, 4.0 isn't binary compatible with 3.x so we [changed the package name](https://github.com/apollographql/apollo-kotlin/issues/4710) to `com.apollographql.apollo`.
```
martinbonnin, line 16:
```suggestion
While most of the common APIs stayed the same, Apollo Kotlin 4 contains a few binary breaking changes. To account for that, and in order to be more future proof, we [changed the package name](https://github.com/apollographql/apollo-kotlin/issues/4710) to `com.apollographql.apollo`. 
```

martinbonnin: Also this is not done yet so only merge after the actual change.

docs/source/migration/4.0.mdx
martinbonnin: I'd say it's fine to have it prominently at the beginning. "Maven artifacts reorganization" is a lot more advanced APIs so I'm OK to have them in 2 different sections as it's probably 2 different audiences.

Nit: use the same order as what people are going to use when updating:

```suggestion
## Group id / plugin id / package name
```

martinbonnin (approved):

docs/source/migration/4.0.mdx
```diff
@@ -9,27 +9,78 @@ description: From 3.x
...
+Apollo Kotlin 4 removes some deprecated symbols. We strongly recommend removing deprecated usages before migrating to version 4.  
+
+If you are using a library that depends on Apollo Kotlin transitively, you need it to update to version 4 before you can update your own application to version 4.
 
 ## Automatic migration using the Android Studio plugin
 
 Apollo Kotlin 4 ships with a companion [Android Studio plugin](../testing/android-studio-plugin#migration-helpers) that automates most of the migration. 
 
-It automates most of the API changes but cannot deal with behaviour changes like error handling. 
+It automates most of the API changes but cannot deal with behavior changes like error handling.
```
martinbonnin, line 26: 👍 For the record, our style guide recommends American english so 👍

docs/source/migration/4.0.mdx
martinbonnin: Move to https://github.com/apollographql/apollo-kotlin/pull/6008

docs/source/migration/4.0.mdx
```diff
@@ -9,27 +9,29 @@ description: From 3.x
...
+Apollo Kotlin 4 focuses on tooling, stability and fixing some API regrets that came with 3.
 
-Because most of the common APIs stayed the same, we [kept the package name unchanged](https://github.com/apollographql/apollo-kotlin/issues/4710). Apollo Kotlin 4.0 removes some deprecated symbols. We strongly recommend removing deprecated usages before migrating to 4.0.  
+While most of the common APIs stayed the same, Apollo Kotlin 4 contains a few binary breaking changes. To account for that, and in order to be more future proof, we [changed the package name](https://github.com/apollographql/apollo-kotlin/issues/4710) to `com.apollographql.apollo`.
 
-If you are using a library that depends on Apollo Kotlin transitively, you need it to update to 4.x before you can update your own application to 4.0.
+Apollo Kotlin 4 removes some deprecated symbols. We strongly recommend removing deprecated usages before migrating to version 4.  
+
+If you are using a library that depends on Apollo Kotlin transitively, you need it to update to version 4 before you can update your own application to version 4.
```
martinbonnin, line 20: I forgot to move that part to next_docs. Note that that last line isn't true anymore under `com.apollographql.apollo`, I'll clean that up in a follow up PR

PR #6002, opened 2024-06-26
Nullability: remove `@ignoreErrors`, `@catchField` and add `@catchByDefault`

martinbonnin:
See https://github.com/apollographql/specs/pull/50

Bump nullability version to 0.4:

```graphql
extend schema @link(url: "https://specs.apollo.dev/nullability/v0.4", import: ["@semanticNonNull", "@catch", "@catchByDefault", "CatchTo"])

extend schema @catchByDefault(to: THROW) # or some other default
```

Allows to set `@catchByDefault` on operations/fragments:

```graphql
fragment QueryDetails on Query @catchByDefault(to: RESULT) {
  nonNull 
  nullable 
}

query Fragments @catchByDefault(to: NULL) {
  ...QueryDetails
  nonNull 
  nullable 
}
```

This unlocks scenarios like:

* `extend schema @catchByDefault(to: NULL)` to start exploring the directives
*  Add `@catch` to individual fields, get more comfortable with how it works
* When ready to do the big switch, change to `extend schema catch(to: THROW)` and (at the same time) add  `query GetFoo @catch(to: NULL) {}` on all operations/fragments (this should be a no-op)
* From this moment on, new queries written are `catch(to: THROW)` by default
* Remove `query GetFoo @catch(to: NULL) {}` progressively


BoD (approved):
👍

libraries/apollo-ast/api/apollo-ast.api
```diff
@@ -928,13 +928,13 @@ public final class com/apollographql/apollo3/ast/ReservedEnumValueName : com/apo
...
 	public static final field CATCH Ljava/lang/String;
-	public static final field CATCH_FIELD Ljava/lang/String;
+	public static final field CATCH_BY_DEFAULT Ljava/lang/String;
 	public static final field Companion Lcom/apollographql/apollo3/ast/Schema$Companion;
 	public static final field FIELD_POLICY Ljava/lang/String;
 	public static final field FIELD_POLICY_FOR_FIELD Ljava/lang/String;
 	public static final field FIELD_POLICY_KEY_ARGS Ljava/lang/String;
 	public static final field FIELD_POLICY_PAGINATION_ARGS Ljava/lang/String;
-	public static final field IGNORE_ERRORS Ljava/lang/String;
+	public static final field LINK Ljava/lang/String;
```
martinbonnin, line 937: This might be a BCV false positive as the matching symbol is annotated `@ApolloExperimental`. Probably because of `const`?

martinbonnin: Ah! I filed [this](https://github.com/Kotlin/binary-compatibility-validator/issues/90) 2 years ago 🙃

PR #6003, opened 2024-06-27
Kotlin nightlies: filter on 2.0.20, and trigger workflow

BoD:
The latest Kotlin dev was 2.0.20 yesterday but today it's 1.9.24 😅 - so filter on that. There are also some 2.1.0 versions but that fails to build with missing symbols, I think it's too early.

### DIFF
--- a/.github/workflows/bump-kotlin-nightlies.yml
+++ b/.github/workflows/bump-kotlin-nightlies.yml
@@ -18,3 +18,10 @@ jobs:
           git add .
           git commit -m "Bump Kotlin version"
           git push --force origin kotlin-nightlies
+      # Trigger a workflow manually because actions cannot trigger workflows to avoid endless loops
+      # https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
+      - name: Trigger pr workflow
+        run: |
+          gh workflow run pr --ref kotlin-nightlies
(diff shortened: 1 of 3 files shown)


martinbonnin:
> The latest Kotlin dev was 2.0.20 yesterday but today it's 1.9.24 😅 - so filter on that.

👍 

> There are also some 2.1.0 versions but that fails to build with missing symbols, I think it's too early.

Could it be a legit regression? 2.1.0 is supposed to be mostly backward compatible, right? Is the Kotlin team taking issues against 2.1.0 nightlies?

BoD:
> Could it be a legit regression? 2.1.0 is supposed to be mostly backward compatible, right? Is the Kotlin team taking issues against 2.1.0 nightlies?

All right I asked and it looks like for now it's too early to report issues on 2.1.

For the record the issue I got is that symbols from `platform.posix` in [here](https://github.com/apollographql/apollo-kotlin/blob/8db2bfc08fa61a54c7ec5994e405c140a54f9fc4/libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo3/network/http/DefaultHttpEngine.apple.kt#L39) were not found.

martinbonnin:
> All right I asked and it looks like for now it's too early to report issues on 2.1.

👍

martinbonnin (approved):

PR #6007, opened 2024-06-28
Kotlin Nightlies: Update apiDump

BoD:


.github/workflows/bump-kotlin-nightlies.yml
```diff
@@ -12,7 +12,9 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
       - name: Bump Kotlin version
         run: ./scripts/bump-kotlin-nightlies.main.kts
-      - name: Push changes to kotlin-nightlies branch
+      - name: Update API dump
+        run: ./gradlew apiDump
+      - name: Commit and push changes to kotlin-nightlies branch
```
martinbonnin, line 17: My favorite workflow here would be to push those changes manually to the nightly branch. 

This is something that needs to be reviewed. I want the worflow to fail if there is an apiDump change (or anything else breaks really). Then I checkout the nightly branch, make it work again. Decide if I need to file an issue to the Kotlin repo or if it's just a migration that has to be done and move on with my life. But I wouldn't want the robots to make that choice for me.

martinbonnin: We could decide to bake that in "bump-kotlin-nightlies.main.kts" (or more likely "make-compile-with-kotlin-2.0.20.main.kts") but it feels a bit out of place. I think I prefer the state to be in the branch compared to e.g. checking in diffs or doing regexes replacements

BoD: So I think we need the nightly branch to be [main] + [kotlin and ksp version bump] + [some commits we know we need to make it work (e.g. new apiDump)].

The CI job can be:
- checkout the `kotlin-nightlies` branch
- rebase on top of main
- bump kotlin/ksp versions
- commit/push
- execute PR workflow

martinbonnin: Yep, something like this. Of course the rebase might fail so it's yet another thing to handle but I think I'd prefer this over managing the state in the workflow file or other scripts.

PR #6012, opened 2024-06-28
Deprecate apollo-adapters

martinbonnin:
I had to remove `apollo-adapters` from integration-tests but I think we have test coverage in other places for that.


BoD (approved):
👍

PR #6013, opened 2024-06-28
Silence a few warnings

martinbonnin:
These are mostly warnings from the internal usages of `apollo-testing-support`. Once removed, we can keep them in a non-published module.


BoD (approved):
👍

PR #6027, opened 2024-07-04
[🐘gradle-plugin] Remix the multi-module configurations once again

martinbonnin:
* Introduce the `com.apollographql.direction` attribute and use it to disambiguate the downstream dependencies.
* Always create configurations in groups of 3: `scope, resolvable, consumable`.
* Hide internal API.
* Update migration guide.


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultApolloExtension.kt
```diff
@@ -64,14 +63,9 @@ abstract class DefaultApolloExtension(
...
-    }
-    if (existingService != null) {
-      existingService.isADependencyOf(project.rootProject.project(projectPath))
-    } else {
-      pendingDownstreamDependencies.compute(serviceName) { _, oldValue ->
-        oldValue.orEmpty() + projectPath
+    project.configurations.configureEach {
+      if (it.name == ModelNames.configuration(serviceName, ApolloDirection.Downstream, ApolloUsage.Ir, ConfigurationKind.DependencyScope)) {
+        it.dependencies.add(project.dependencies.project(mapOf("path" to projectPath)))
       }
```
martinbonnin, line 69: This is the actual fix. If `existingService` was not null, it's too late to call `isADependencyOf()`.

BoD (approved):
🚀

PR #6028, opened 2024-07-04
[📚docs] link directly to https://go.apollo.dev/ak-moved-artifacts

martinbonnin:
Be more specific when linking to the migration guide.


BoD (approved):
👍

PR #6031, opened 2024-07-08
[📚 docs] Fix experimental link

### LINKED ISSUE #5830: [Documentation] Defer docs "experimental" link results in a 404
### Description

In https://www.apollographql.com/docs/kotlin/fetching/defer/ if you click on the "experimental" link at the top, which currently links to "https://www.apollographql.com/docs/resources/release-stages/" results in a 404
image

Interestingly, if you just enter that link to your browser it does properly redirect to "https://www.apollographql.com/docs/resources/product-launch-stages/".
So something may be wrong with the way that redirects are handled in the docs themselves?

martinbonnin:
There is a disconnect between Gatsby and Netlify. Will be fixed in a future version of the docs framework but for now, let's just update the link.
Fixes #5830

### DIFF
--- a/docs/source/advanced/compiler-plugins.mdx
+++ b/docs/source/advanced/compiler-plugins.mdx
@@ -4,7 +4,7 @@ title: Apollo compiler plugins
 
 <ExperimentalFeature>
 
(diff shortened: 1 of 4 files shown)


BoD (approved):
👍

PR #6055, opened 2024-07-15
Undeprecate schemaFile

martinbonnin:
I changed my mind. There is a global expectation that there is a "main" schema file. Furthermore, deprecating this also had cascading impacts on `introspection {}` for which we don't have a good answer right now.


BoD (approved):
👍

martinbonnin:
Drawbacks: using `schemaFile.set()` ignore any `extra.graphqls` file.

PR #6060, opened 2024-07-16
[📚docs] Update evolution policy

martinbonnin:
* Relax maintainance window
* Explicit that there is a single maintainance branch per major version
* Add link to the Kotlin documentation about breaking changes
* Reword the source-breaking change paragraph to make it less sound like we're going to break everything all the time!

ping @apollographql/clients


BoD (approved):
👌

PR #6064, opened 2024-07-22
Add v4.0.0-rc.2 changelog

BoD:


CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 Change Log
 ==========
 
+# Version v4.0.0-rc.2
+
+We're on the road to v4 with this second release candidate which includes a few minor tweaks as well as one new feature.
```
martinbonnin, line 6: Feels like something needs to be said about `4.0.0-rc.2` using `4.0.0-rc.1` in `apollo-tooling` and `apollo-debug-server`

martinbonnin: This version is the first one to have everything in the `com.apollographql.apollo`  package name. 

The previous one would duplicate `apollo-api` in some cases because of this. 

> Gradle should resolve apollo-runtime to 4.0.0-rc.2, right?

Correct. If no version is set by the user then AKGP uses the same version as itself but that doesn't help about `apollo-debug-server` pulling apollo3 dependencies.

martinbonnin (approved):

PR #6072, opened 2024-07-23
Allow changing the websocket url in interceptors

martinbonnin:
Honor `request.httpHeaders` in `WebSocketNetworkTransport` as well.

If a `request` requires a different set of headers, a new `WebSocket` is created and the old one times out. 

Also took this opportunity to change the millis `Long` arguments to `Duration`

See https://github.com/apollographql/apollo-kotlin/issues/6063#issuecomment-2241101473


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketNetworkTransport.kt
```diff
@@ -129,29 +127,6 @@ class WebSocketNetworkTransport private constructor(
...
-
-    /**
-     * Add a [HttpHeader] to the handshake query.
-     */
-    @Deprecated("use addHttpHeader instead", ReplaceWith("addHttpHeader(name, value)"))
-    @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
-    fun addHeader(name: String, value: String) = apply {
-      this.httpHeaders = this.httpHeaders.orEmpty() + HttpHeader(name, value)
-    }
-
```
martinbonnin, line 154: Another bit of precedence bites the dust. We used to have 

* ApolloClient
* ApolloRequest
* NetworkTransport

NetworkTransport is now gone for websockets. We should probably do the same for HTTP

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketNetworkTransport.kt
```diff
@@ -187,17 +161,16 @@ class WebSocketNetworkTransport private constructor(
      *
      * Default: -1
```
BoD, line 162: Default is null (sentence above also mentions -1)

BoD (approved):
🚀

PR #6074, opened 2024-07-23
Deprecate HttpNetworkTransport.httpHeaders()

martinbonnin:
I don't think this brings much value? [Keeps precedence a bit contained](https://youtu.be/YZstpc2939s?t=1888)


BoD (approved):
👍

PR #6075, opened 2024-07-24
[runtime] Remove ensureUuid()

martinbonnin:
`WebSocketNetworkTransport` already has logic to renew the uuid. It's working as long as there's a single `Flow` instance so memoize it. Makes it even simpler to implement retry interceptors. 

See https://github.com/apollographql/apollo-kotlin/pull/6069#discussion_r1687964151


BoD (approved):
👍

PR #6091, opened 2024-08-02
Allow to shutdown SchemaDownloader

martinbonnin:
This is important for tools like CLI that otherwise stay stuck until the threadpool terminates


BoD (approved):
👍

PR #6092, opened 2024-08-02
[Introspection] Remove warning

martinbonnin:
Fixes 

```
Apollo: schema is missing 'description', double check your introspection query
```

when downloading a schema from the fullstack tutorial (which uses a GraphQL versions pre-October2021 😅 )


BoD (approved):
🤯

PR #6093, opened 2024-08-02
Codegen: avoid enum value clashing with the getter `field`

martinbonnin:
Reproducer:

```graphql
enum Foo {
  field
}
```

Because `field` is the [backing field](https://kotlinlang.org/docs/properties.html#backing-fields) in the context of the getter, we need to scope the identifier better.


BoD (approved):
🚀

jadar:
Thanks for this!

PR #6100, opened 2024-08-06
[📚 docs] Add a truth table for error handling

martinbonnin:


docs/source/essentials/errors.mdx
```diff
@@ -137,3 +142,16 @@ For an example, it is possible for a person to not have a starship:
 ```
 
 In that case, `starship` is a true null and not an error.
+
+## Truth table
+
+| exception | data    | errors  | meaning                                                                     |
+|-----------|---------|---------|-----------------------------------------------------------------------------|
+| `null`    | `null`  | `null`  | Non-compliant server, if `data` is null, there should be at least one error |
```
BoD, line 150: Nit: I'd use "Something" or "Not null" to make it a bit more legible (maybe it's just me, feel free to ignore!)

BoD (approved):

docs/source/essentials/errors.mdx
martinbonnin: Yup doesn't look good in preview. I'll add "not null" (or is it "non null" 🤔?)

Screenshot 2024-08-06 at 16 19 40

martinbonnin: I went with `present`

Screenshot 2024-08-06 at 16 19 40

PR #6103, opened 2024-08-07
[experimental WebSockets] connectionParams -> connectionPayload

martinbonnin:
For symmetry with the current websocket implementation. Also it makes more sense.


BoD (approved):
👍

PR #6104, opened 2024-08-07
Add 'OfflineException' as a cause when using failFastIfOffline

### LINKED ISSUE #6101: Add cause to ApolloNetworkException thrown when offline
### Use case

We'd like to handle the offline exception thrown in `RetryOnErrorInterceptor` in a type-safe way.

Discussed on Slack: https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1722933522055469

martinbonnin:
`OfflineException` is an `IOException`, not an `ApolloException` so that users can still check for `ApolloNetworkException` to filter some exceptions.

Closes #6101


BoD (approved):
👍

PR #6105, opened 2024-08-07
[WebSockets] add NetworkTransport.closeConnection()

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/5862#issuecomment-2272925736


BoD (approved):
👍

PR #6116, opened 2024-08-13
Escape properties in input builder function body

ebrattli:

PR #6117, opened 2024-08-13
[⚙️ compiler] Fix a few additional instances of %L used instead of %N

martinbonnin:
See #6116

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/helpers/NamedType.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/helpers/NamedType.kt
@@ -71,10 +71,10 @@ internal fun NamedType.toSetterFunSpec(context: KotlinContext): FunSpec {
   val body = CodeBlock.builder()
   val parameterType: IrType
   if (type.optional) {
-    body.add("this.%L·=·%T(%L)\n", propertyName, KotlinSymbols.Present, propertyName)
+    body.add("this.%N·=·%T(%N)\n", propertyName, KotlinSymbols.Present, propertyName)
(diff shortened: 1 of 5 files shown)


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/helpers/NamedType.kt
```diff
@@ -129,7 +129,7 @@ private fun List<NamedType>.toBuildFunSpec(context: KotlinContext, returnedClass
               .apply {
                 forEach {
                   val propertyName = context.layout.propertyName(it.graphQlName)
-                  add("%L·=·%L", propertyName, propertyName)
+                  add("%N·=·%N", propertyName, propertyName)
```
martinbonnin, line 132: Those are for data builders

BoD (approved):
👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/operations/CompiledSelectionsBuilder.kt
```diff
@@ -125,7 +125,7 @@ internal class CompiledSelectionsBuilder(
   private fun IrArgument.codeBlock(): CodeBlock {
     val argumentBuilder = CodeBlock.builder()
     argumentBuilder.add(
-        "%T(%T.%L)",
+        "%T(%T.%N)",
         KotlinSymbols.CompiledArgument,
         context.resolver.resolveArgumentDefinition(definitionId),
         definitionPropertyName,
```
martinbonnin, line 131: Same here, not really needed because there is already `__` in front of `definitionPropertyName` but still feels more correct

PR #6119, opened 2024-08-13
[runtime] Change NetworkMonitor to expose a StateFlow directly

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/6115

The previous interface was based on what `ApolloClient` needs but it prevented using `NetworkMonitor` for other use cases. This PR allows to use `NetworkMonitor` in a more generic way. 

This is obviously 200% breaking. If anyone implements their own `NetworkMonitor`, they'll need to update but hopefully it's not too hard (and the symbol is `@ApolloExperimental` in all cases).


BoD (approved):
🎉

PR #6136, opened 2024-08-29
Provide a more descriptive error message when a resolution of a specific `ResolverKey` fails

agrosner:
This likely affects multimodule projects, but when a symbol fails to resolve in a project, we've found that it is a result of engineers forgetting to add an `isADependencyOf` backlink to the project that performs an operation on the schema. The best we can do is warn them in the logs and hope it fixes it for them.


martinbonnin (approved):
Thanks!

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/java/JavaResolver.kt
```diff
@@ -96,7 +96,8 @@ internal class JavaResolver(
     val result = resolve(ResolverKey(kind, id))
 
     check(result != null) {
-      "Cannot resolve $kind($id)"
+        "Cannot resolve $kind($id)" +
```
agrosner, line 99: whoops. one more change here.

PR #6137, opened 2024-09-03
Fix cascade SQL cache remove loops forever on cyclic references

### LINKED ISSUE #6133: Cascade cache remove loops forever on cyclic references
### Version

4.0.0

### Summary

Our schema:

image

Screenshot 2024-08-26 at 14 50 36

When deleting the Record object using its CacheKey with cascade = true. It seems to loop forever.

Here's the remove method: https://github.com/apollographql/apollo-kotlin/blob/e969ade1ae58513a0f11b74d250cef094d8bb1e0/libraries/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/api/internal/OptimisticCache.kt#L46

Just looking at the code, there is a recursive call which probably calls remove between those two objects indefinitely. The cascade remove should probably gather all unique references first and then delete them.

cvb941:
Should fix #6133


martinbonnin (approved):
Thanks!

PR #6145, opened 2024-09-11
Allow compiling the project with Java22

martinbonnin:
By default, GR8 uses the current Java runtime and the R8 version we use cannot deal with 22. By passing Java 8 here, we hide 22 bytecode from R8. We still cannot bump the R8 version due to https://issuetracker.google.com/u/1/issues/365784991 but this buys us some time.


BoD (approved):
🕙

PR #6149, opened 2024-09-16
[gradle-plugin] Use `registerJavaGeneratingTask`

martinbonnin:
Go back to `registerJavaGeneratingTask` as in 3.x instead of `kotlin.srcDir()`

Some background:

1. We initially used `kotlin.srcDir()` and switched to `registerJavaGeneratingTask` in 2022 ([commit](https://github.com/apollographql/apollo-kotlin/pull/4486)) because it works better with lint (it marks the sources as generated)
1. But this had 2 drawbacks:
  a. `registerJavaGeneratingTask` wasn't working from `afterEvaluate {}`
  b. there was an issue with incremental compilation and tests ([issuetracker/268218176](https://issuetracker.google.com/u/1/issues/268218176))
1. So we reverted to `kotlin.srcDir()` ([commit](https://github.com/apollographql/apollo-kotlin/pull/4674))
1. [issuetracker/268218176](https://issuetracker.google.com/u/1/issues/268218176) has an explanation (we were adding to test variants something that was already accessible)
1. we don't need `afterEvaluate {}` anymore because there is no default service anymore 
1. besides the lint issue, `kotlin.srcDir()` doesn't seem to carry task dependencies for AGP source jar tasks ([issuetracker/366293471](https://issuetracker.google.com/u/1/issues/366293471))
1. So we're back to 1., with hopefully 2 fixes issues since 2022 (`afterEvaluate` and test variants)

Also took this opportunity to remove some reflexion that is not needed anymore now that we require Gradle 8.0.

All in all, I wish `kotlin.srcDir()` would just work as it's more consistent with how generated sources are added in non-Android cases but for the time being, I think we're good with `registerJavaGeneratingTask()` 🤞 

Fixes https://github.com/apollographql/apollo-kotlin/issues/6146

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
@@ -871,8 +871,12 @@ interface Service {
     fun connectToAndroidSourceSet(name: String)
 
     /**
-     * Connects the generated sources to all the Android variants
-     * Throws if the Android plugin is not applied
+     * Connects the generated sources to the main Android variants.
+     * Note: despite the name, this method does not connect to the test variants as the main
+     * classpath is accessible from the tests and adding the sources twice causes issues.
+     * See https://issuetracker.google.com/u/1/issues/268218176
+     *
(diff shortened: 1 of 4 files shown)


BoD (approved):
👍

BoD (approved):
👍

PR #6164, opened 2024-09-26
[apollo-ast] Add GQLDocument.validate(SchemaValidationOptions) to allow adding external schemas.

martinbonnin:
Also add support for kotlin_labs/v0.4

```kotlin
val cacheControlSchema = ForeignSchema("cache", "v0.1", definitions)

schemaString.toGQLDocument().validateAsSchema(
    SchemaValidationOptions(
        addKotlinLabsDefinitions = false,
        foreignSchemas = listOf(cacheControlSchema)
    )
).getOrThrow()
```

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
@@ -1,5 +1,6 @@
 package com.apollographql.apollo.ast.internal
 
+import com.apollographql.apollo.annotations.ApolloExperimental
 import com.apollographql.apollo.annotations.ApolloInternal
 import com.apollographql.apollo.ast.ConflictResolution
 import com.apollographql.apollo.ast.DirectiveRedefinition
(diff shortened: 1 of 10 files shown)


BoD (approved):
🚀

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
```diff
@@ -303,83 +358,105 @@ private fun List<GQLSchemaExtension>.getForeignSchemas(
...
+          imports.add(
+              Import(
+                  foreignSchema = foreignSchema,
+                  renamedDefinitions = definitions,
                   newNames = renames,
-                  directivesToStrip = definitions.filterIsInstance<GQLDirectiveDefinition>().map { it.name }
               )
           )
+        } else {
+          issues.add(OtherValidationIssue("Apollo: unknown foreign schema '$foreignName/$foreignVersion'", gqlDirective.sourceLocation))
```
BoD, line 398: Hey actually, maybe this should be allowed? In my case, the cache gradle plugin can add the cache foreign definitions 👍 - but users will still also apply the regular apollo gradle plugin, which won't accept the `extend schema @link(url: "https://specs.apollo.dev/cache/v0.1" import: ...)` with this check.

martinbonnin: Yikes. Good point. I still think this is something we want to validate. For an example, someone asking for `nullability/v0.1` support should get an error: this is not supported anymore. 
Yet another argument for a plugin API. Schema validation should be done only once. But plugins could contribute foreign schemas. I'll open a separate PR

PR #6170, opened 2024-10-01
Add support for application/graphql-response+json

martinbonnin:
The new content-type allows us to have more guarantees when a response is a valid GraphQL response vs potential error from an intermediate node. 

See https://graphql.github.io/graphql-over-http/draft/#sec-application-graphql-response-json
See https://github.com/apollographql/apollo-kotlin/issues/6169


BoD (approved):
👍

PR #6184, opened 2024-10-03
Add note about bidirectional requiring `alwaysGenerateTypesMatching.set(emptyList())`

martinbonnin:
This isn't needed with `isADependencyOf()` for REASONs. We could probably fix this. Will do together with https://github.com/apollographql/apollo-kotlin/issues/6180

### DIFF
--- a/docs/source/advanced/multi-modules.mdx
+++ b/docs/source/advanced/multi-modules.mdx
@@ -97,6 +97,21 @@ apollo {
 }
 ```
 
+Also disable generating types in the schema module:
+
+```kotlin
+// schema/build.gradle.kts
+apollo {
+  service("service") {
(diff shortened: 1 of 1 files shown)


BoD:
Should I update [the sample](https://github.com/apollographql/apollo-kotlin-samples/blob/main/multi-modules-and-services/graphqlSchema/build.gradle.kts)?

BoD (approved):

martinbonnin:
> Should I update [the sample](https://github.com/apollographql/apollo-kotlin-samples/blob/main/multi-modules-and-services/graphqlSchema/build.gradle.kts)?

Probably worth removing the `bidirectional` flag for now until project isolation works.

PR #6188, opened 2024-10-07
Unbreak benchmarks

martinbonnin:
Turns out `libs.apollo.normalizedcache.sqlite` does not pull `libs.apollo.normalizedcache` transitively


BoD (approved):
👍

PR #6195, opened 2024-10-10
[codegen] Use `layout` in data builder names

martinbonnin:
`layout` properly escapes type names in case of name clash.

See https://github.com/apollographql/apollo-kotlin/issues/5537


BoD (approved):
👍

PR #6197, opened 2024-10-10
Websockets: Guard against synchronous `onError()` calls

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/5862#issuecomment-2404607206


BoD (approved):
👍

BoD (approved):

PR #6215, opened 2024-10-23
Update kotlinpoet to 2.0.0

martinbonnin:
See https://github.com/square/kotlinpoet/releases/tag/2.0.0

It bumps Kotlin to 2.0.10 from 2.0.0 which should be fine and is otherwise backward compatible.


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/helpers/KDoc.kt
```diff
@@ -18,31 +18,31 @@ internal fun TypeSpec.Builder.maybeAddDescription(description: String?): TypeSpe
...
+  return addKdoc("%L", description.replace(' ', '♢'))
 }
 
 internal fun FunSpec.Builder.maybeAddDescription(description: String?): FunSpec.Builder {
   if (description.isNullOrBlank()) {
     return this
   }
 
-  return addKdoc("%L", description)
+  return addKdoc("%L", description.replace(' ', '♢'))
```
martinbonnin, line 45: This is the only place where I added possibly wrapping spaces to make KDoc comments easier to read.

BoD (approved):
🚀

PR #6218, opened 2024-10-24
Remove mockserver as a dependency of apollo-testing-support

martinbonnin:
`apollo-mockserver` is resolved locally but is only published as a maven tombstone and cannot be resolved in external projects.

Many thanks @jvanderwee for catching this.


BoD (approved):
✅

PR #6222, opened 2024-10-28
Remove Maven relocations for mockserver and ktor-engine

martinbonnin:
Maven relocations do not come with warnings (see [here](https://github.com/gradle/gradle/issues/26436) and [here](https://issuetracker.google.com/u/2/issues/375959945)) so we might as well remove the bandaid all together and call it a day. 

`4.1.0` is breaking because of Ktor3 already, might as well put that behind.


BoD (approved):
👍

PR #6223, opened 2024-10-28
[compiler] Make apollo-compiler aware of federation

martinbonnin:
Fixes https://github.com/apollographql/apollo-kotlin/issues/6221

I'm not 100% sold on this solution yet. If other directives appear, we can have the problem again. But maybe it's good that we have this problem? Clients should use introspection results. But sometimes introspection is disabled. And [directives are not exposed in introspection](https://github.com/graphql/graphql-spec/issues/300)...

tldr; this should at least fix the above issue while keeping our future options relatively open.


BoD (approved):
Yeah it's a pragmatic tradeoff 👍

martinbonnin:
Closing in favor of https://github.com/apollographql/apollo-kotlin-ffs

PR #6232, opened 2024-11-04
[tests] minor fix in engine tests to fail on more exceptions

martinbonnin:
Not sure why this was caught. Probably some debugging leftovers.


BoD (approved):
👍

PR #6244, opened 2024-11-06
[compiler] depend on kotlinpoet-jvm and not kotlinpoet

martinbonnin:
This is a workaround for https://github.com/gradle/gradle/issues/31158. 
FWIW, I believe something spooky is happening around [here](https://github.com/gradle/gradle/blob/0490d220bbfa1a4a590cf1a5c0113d6a7d12d70d/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/ivyservice/resolveengine/graph/builder/NodeState.java#L684) but I'm not 100% sure what. 

In all cases, it shouldn't hurt.


BoD (approved):
🐘

PR #6248, opened 2024-11-08
[compiler] Add Enum.KNOWN__ as an intermediary interface

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/6243

## Generated code

At a high level, this change introduces an intermediate `KNOWN__` interface, symmetrical with `UNKNOWN__`:

```kotlin
public sealed interface Color {
  public val rawValue: String

  public object BLUEBERRY : KNOWN__ {
    override val rawValue: String = "BLUEBERRY"
  }

  public object CHERRY : KNOWN__ {
    override val rawValue: String = "CHERRY"
  }

  public object CANDY : KNOWN__ {
    override val rawValue: String = "CANDY"
  }

  /**
   * An enum value that is known at build time.
   */
  @Suppress("ClassName")
  public sealed interface KNOWN__ : Color {
    override val rawValue: String
  }

  /**
   * An enum value that isn't known at build time.
   */
  @Suppress("ClassName")
  public class UNKNOWN__ @ApolloPrivateEnumConstructor constructor(
    override val rawValue: String,
  ) : Color 
}
```

## Usage

This allows unwrapping an unknown value to a known one to provide a "default":

```kotlin
  /**
   * Turns a maybe unknown color value into a known one
   */
  private fun Color.orDefault(): Color.KNOWN__ = when (this) {
    is Color.UNKNOWN__ -> Color.CANDY
    is Color.KNOWN__ -> this
  }
```

This way, callers don't have to switch on the unknown case:

```kotlin
    when (color.orDefault()) {
      Color.BLUEBERRY -> TODO()
      Color.CANDY -> TODO()
      Color.CHERRY -> TODO()
      // no need to check for unknown here
    }
```

## Compatibility

I contemplated adding another option but ultimately decided to reuse `sealedClassesAsEnumMatching`. 

This is a source-compatible change but a binary breaking change because of the changing of `UNKNOWN__` from interface to class.

See https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.4:

```
If C is not an interface, interface method resolution throws an IncompatibleClassChangeError. 
```

You are impacted if you are shipping a library that exposes a generated `SomeEnum.UNKNOWN__` in its public API.

I'm hoping that this pattern is not too widespread. If it is, and you're reading this, past me apologize in advance 😬 .

### DIFF
--- a/libraries/apollo-annotations/api/apollo-annotations.api
+++ b/libraries/apollo-annotations/api/apollo-annotations.api
@@ -8,6 +8,9 @@ public abstract interface annotation class com/apollographql/apollo/annotations/
 public abstract interface annotation class com/apollographql/apollo/annotations/ApolloInternal : java/lang/annotation/Annotation {
 }
 
+public abstract interface annotation class com/apollographql/apollo/annotations/ApolloPrivateEnumConstructor : java/lang/annotation/Annotation {
+}
+
 public abstract interface annotation class com/apollographql/apollo/annotations/ApolloRequiresOptIn : java/lang/annotation/Annotation {
 }
 
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/Identifiers.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/Identifiers.kt
@@ -40,9 +40,7 @@ internal object Identifier {
   const val copy = "copy"
   const val Data = "Data"
 
-  const val cacheKeyForObject = "cacheKeyForObject"
   const val field = "field"
-  const val __map = "__map"
   const val __path = "__path"
   const val __fields = "__fields"
 
@@ -60,11 +58,16 @@ internal object Identifier {
   const val knownValues = "knownValues"
   const val knownEntries = "knownEntries"
 
-  // extra underscores at the end to prevent potential name clashes
+  /**
+   * UNKNOWN__ and KNOWN__ should probably have been __UNKNOWN because GraphQL reserves the leading __ but it's too late now.
+   *
+   * All in all it's not too bad because typing 'U', 'N', ... is usually more intuitive and in the very unlikely event that
+   * there is a name clash, it can always be resolved with `@targetName`
+   */
   const val UNKNOWN__ = "UNKNOWN__"
+  const val KNOWN__ = "KNOWN__"
   const val rawValue = "rawValue"
   const val types = "types"
-  const val testResolver = "testResolver"
   const val block = "block"
   const val resolver = "resolver"
   const val newBuilder = "newBuilder"
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/schema/EnumAsSealedInterfaceBuilder.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/schema/EnumAsSealedInterfaceBuilder.kt
@@ -1,6 +1,10 @@
 package com.apollographql.apollo.compiler.codegen.kotlin.schema
 
 import com.apollographql.apollo.compiler.codegen.Identifier
+import com.apollographql.apollo.compiler.codegen.Identifier.KNOWN__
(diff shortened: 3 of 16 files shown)


libraries/apollo-annotations/api/apollo-annotations.api
```diff
@@ -8,6 +8,9 @@ public abstract interface annotation class com/apollographql/apollo/annotations/
 public abstract interface annotation class com/apollographql/apollo/annotations/ApolloInternal : java/lang/annotation/Annotation {
 }
 
+public abstract interface annotation class com/apollographql/apollo/annotations/ApolloPrivateEnumConstructor : java/lang/annotation/Annotation {
+}
+
```
martinbonnin, line 13: This introduces the Opt-in marker that we removed in https://github.com/apollographql/apollo-kotlin/commit/d160e982a8afe8e0e7a58df427cf9f83c4eee2ee.

The underlying reason is that Kotlin doesn't support static factory method (companion obejcts cannot access private constructors of nested classes). 

The previous trick of extracting a public interface doesn't work with sealed classes because the compiler insists on having `UNKNOWN__Impl` be part of the exhaustive `when` statements and it's private in the file. Maybe a future version of the compiler could fix that but as of today I haven't found a way.

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/Identifiers.kt
```diff
@@ -60,11 +58,16 @@ internal object Identifier {
   const val knownValues = "knownValues"
   const val knownEntries = "knownEntries"
 
-  // extra underscores at the end to prevent potential name clashes
+  /**
+   * UNKNOWN__ and KNOWN__ should probably have been __UNKNOWN because GraphQL reserves the leading __ but it's too late now.
+   *
+   * All in all it's not too bad because typing 'U', 'N', ... is usually more intuitive and in the very unlikely event that
+   * there is a name clash, it can always be resolved with `@targetName`
```
BoD, line 65: 👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/schema/EnumAsSealedInterfaceBuilder.kt
```diff
@@ -91,71 +100,99 @@ internal class EnumAsSealedBuilder(
...
   }
 
   private fun IrEnum.safeValueOfFunSpec(): FunSpec {
-    return FunSpec.builder(Identifier.safeValueOf)
+    return FunSpec.builder(safeValueOf)
         .addKdoc(
-            "Returns the [%T] that represents the specified [rawValue].\n" +
-                "Note: unknown values of [rawValue] will return [UNKNOWN__]. You may want to update your schema instead of calling this function directly.\n",
+            """
+            $safeValueOf returns an instance of [%T] representing [$rawValue].
```
BoD, line 164:
```suggestion
            Returns an instance of [%T] representing [$rawValue].
```

Nitpick, I'd avoid repeating the fun nam in its doc

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -623,14 +623,16 @@ interface Service {
   val debugDir: DirectoryProperty
 
   /**
-   * A list of [Regex] patterns for GraphQL enums that should be generated as Kotlin sealed classes instead of the default Kotlin enums.
+   * A list of [Regex] patterns for GraphQL enums that should be generated as a Kotlin sealed interface.
    *
-   * Use this if you want your client to have access to the rawValue of the enum. This can be useful if new GraphQL enums are added but
-   * the client was compiled against an older schema that doesn't have knowledge of the new enums.
+   * This provides several benefits over the default of mapping GraphQL enum to Kotlin enums:
```
BoD, line 628:
```suggestion
   * This provides several benefits over the default of mapping GraphQL enums to Kotlin enums:
```

BoD (approved):
🚀

PR #6250, opened 2024-11-08
Add licensee to the repo

martinbonnin:
Also fix a couple of opportunistic warnings

### DIFF
--- a/build-logic/src/main/kotlin/api.kt
+++ b/build-logic/src/main/kotlin/api.kt
@@ -1,3 +1,5 @@
+import app.cash.licensee.LicenseeExtension
+import app.cash.licensee.UnusedAction
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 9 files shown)


build-logic/src/main/kotlin/api.kt
```diff
@@ -75,6 +77,22 @@ fun Project.apolloLibrary(
...
+    allow("Apache-2.0")
+    allow("MIT")
+    allow("CC0-1.0")
+    allow("MIT-0")
+
+    allowUrl("https://raw.githubusercontent.com/apollographql/apollo-kotlin-execution/main/LICENSE")
+    allowUrl("https://raw.githubusercontent.com/apollographql/apollo-kotlin-mockserver/main/LICENSE")
+    allowUrl("https://raw.githubusercontent.com/apollographql/apollo-kotlin/main/LICENSE")
+    allowUrl("https://asm.ow2.io/license.html")
+    allowUrl("https://spdx.org/licenses/MIT.txt")
```
BoD, line 94: 👀 surprised that this is needed when `MIT` is allowed (not a problem at all, just wondering)

BoD (approved):
Cool! 👍

PR #6252, opened 2024-11-11
Compiler plugins: only call `onSchema()` once in multi-module scenarios

### LINKED ISSUE #6251: Compiler plugin `onSchema()` is not ergonomic in multi-module setups
### Use case

We are trying to use the compiler plugin `onSchema()` method to generate some custom code using kotlin-poet. Previously, we did this using a custom gradle task. We have a multi-module setup. 

The `onSchema` method is a much better fit for our use case: we no longer need to "hook into" apollo internals to convert the file to a schema. However, the downside is that the `onSchema` method is called multiple times: once for each module. 

This means that our codegen ends up running multiple times, creating duplicate files. Ideally, we would want to run this once only, for the schema module, and not for the feature modules. This is currently not very easy to do. Our workaround was to pass an argument to the plugin, e.g. `isSchemaModule`, and use that to determine whether to run the codegen or not.  

Perhaps the `onSchema` method of the `SchemaListener` can provide some more information about the module being operated on? Is there a clear way to identify schema vs feature modules?

martinbonnin:
Fixes #6251


BoD (approved):
👍

PR #6281, opened 2024-11-25
Close ApolloInterceptor if they implement Closeable

BoD:
Related to https://github.com/apollographql/apollo-kotlin-normalized-cache-incubating/pull/67.


BoD:
After consideration we don't think this is useful.

Issue #6276, opened 2024-11-22
Using WebSocketNetworkTransport throws `IllegalArgumentException: Flow has more than one element`


akashamin:
### Version

3.8.5

### Summary

We are currently working on a migration from Apollo2 to Apollo3 and see a bunch of `IllegalArgumentException: Flow has more than one element` thrown when receiving updates through webSocket. 

ApolloClient builder code below

```kotlin
ApolloClient.Builder().apply {
    serverUrl(baseUrl)
    httpEngine(DefaultHttpEngine(newOkHttpClient))
    dispatcher(defaultDispatcher())
    normalizedCache(
        normalizedCacheFactory = cacheFactory,
        cacheResolver = resolver,
        cacheKeyGenerator = generator,
    )
    webSocketUrl?.let { url ->
        subscriptionNetworkTransport(WebSocketNetworkTransport.Builder()
            .serverUrl(url)
            .okHttpClient(newOkHttpClient)
            .protocol(
                SubscriptionWsProtocol.Factory(
                    connectionPayload = {
                        authorizationHeaders()
                    }
                )
            )
            .reopenWhen { throwable, attempt ->
                if (throwable is DisableSubscriptionException) {
                    reconnectSignalChannel.receive()
                }
                true
            }
            .build())
    }

    autoPersistedQueries(enableByDefault = enablePersistedQueries)

    customTypeAdapters?.forEach { (scalarType, customTypeAdapter) ->
        addCustomTypeAdapter(scalarType, customTypeAdapter)
    }
}.build()
```

I found an old [issue](https://github.com/apollographql/apollo-kotlin/issues/4617) related to the same exception, when I remove `autoPersistedQueries` I don't see the exception, not sure if this is a configuration issue on our end or a bug in the library. 

### Logs

```
java.lang.IllegalArgumentException: Flow has more than one element
    at kotlinx.coroutines.flow.FlowKt__ReduceKt$single$2.emit(Reduce.kt:54)
    at com.apollographql.apollo3.network.ws.WebSocketNetworkTransport$execute$$inlined$filterNot$1$2.emit(Emitters.kt:223)
    at com.apollographql.apollo3.network.ws.WebSocketNetworkTransport$execute$$inlined$map$1$2.emit(Emitters.kt:223)
    at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:11)
[...]

PR #6283, opened 2024-11-26
Fix using PQs with subscriptions

martinbonnin:
See #6276

* Make it possible to use PQs with subscriptions.
* Make `release-3.x` compile now that jcenter is dead.
* Fix https://github.com/apollographql/apollo-kotlin/pull/5894#pullrequestreview-2117767392 (subscriptions may do a partial read)

See test [here](https://github.com/martinbonnin/reproducers/tree/da6a1ee2454b2affda8ba174c6360e85bb5119a1/apollo-6276)

### DIFF
--- a/libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
+++ b/libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
@@ -56,12 +56,7 @@ actual class MockServer actual constructor(override val mockServerHandler: MockS
   }
 
   override suspend fun url() = url ?: suspendCoroutine { cont ->
-    val server = requireNotNull(this.server) { "Server is not listening, please call listen() before calling address()" }
-
     server.once(Event.LISTENING) {
-      val server = requireNotNull(this.server) {
-        "close() was called during a call to address()"
-      }
       val address = server.address().unsafeCast<AddressInfo>()
       url = "http://localhost:${address.port}/"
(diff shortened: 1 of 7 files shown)


martinbonnin:
Sadly the release-3.x CI has flakes. The integration tests "run on my machine" though so I'm tempted to merge anyways.

libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
```diff
@@ -48,18 +49,28 @@ actual class MockServer actual constructor(override val mockServerHandler: MockS
...
+  }
+
+  init {
+    server.listen()
+  }
 
   override suspend fun url() = url ?: suspendCoroutine { cont ->
-    url = "http://localhost:${server.address().unsafeCast<AddressInfo>().port}/"
-    server.on("listening") { _ ->
+    val server = requireNotNull(this.server) { "Server is not listening, please call listen() before calling address()" }
```
BoD, line 59: Can `this.server` ever be null?

BoD (approved):
👍

libraries/apollo-mockserver/src/jsMain/kotlin/com/apollographql/apollo3/mockserver/MockServer.kt
martinbonnin: You're right, it can't, this is a bad copy/paste from v4, I'll remove it

PR #6284, opened 2024-11-26
Allow multiple @link schema extensions

BoD:
Adding multiple `@link` extensions was failing because the definition of `@link` wasn't known, so couldn't check if it's repeatable [here](https://github.com/apollographql/apollo-kotlin/blob/7668a302eb5c62ca1d3e423ccfefa1142acee1f4/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/ExtensionsMerger.kt#L265).

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
@@ -65,7 +65,7 @@ class SchemaValidationOptions(
 
 internal fun validateSchema(definitions: List<GQLDefinition>, options: SchemaValidationOptions): GQLResult<Schema> {
   val issues = mutableListOf<Issue>()
-  val builtinDefinitions = builtinDefinitions()
+  val builtinDefinitions = builtinDefinitions() + linkDefinitions()
 
   // If the builtin definitions are already in the schema, keep them
   var allDefinitions = combineDefinitions(definitions, builtinDefinitions, ConflictResolution.TakeLeft)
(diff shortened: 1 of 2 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
```diff
@@ -326,7 +326,7 @@ private fun List<GQLSchemaExtension>.getImports(
          * Validate `@link` using a very minimal schema.
          * This ensure we can safely cast the arguments below
          */
-        val minimalSchema = builtinDefinitions + linkDefinitions()
+        val minimalSchema = builtinDefinitions
```
martinbonnin, line 329: Can we have `getImports` return its own foreign schema for `@link`? `@link` is a very special case and I'd rather keep everything contained in `getImports()`

martinbonnin: Something along these lines:

```
      if (gqlDirective.name == "link") {
        val linkForeignSchema = ForeignSchema("link", "v1.0", """
              directive @link(
                url: String!,  
                as: String,
                import: [Import],
                for: Purpose)
                repeatable on SCHEMA

              scalar Import

              enum Purpose {
                SECURITY
                EXECUTION
              }

            """.trimIndent().toGQLDocument().definitions
        )
        imports.add(
            Import(linkForeignSchema, linkForeignSchema.definitions, mapOf("link" to "link"))
        )
```

martinbonnin (approved):

PR #6286, opened 2024-11-27
Remove CloseableBackgroundDispatcher

martinbonnin:
`Dispatchers.Default.limitedParallelism(1)` is a better replacement that doesn't need lifecycle management


BoD (approved):
🧹

PR #6287, opened 2024-11-27
Deprecate all intermediate `Builder` functions

martinbonnin:
Deprecate things like

```kotlin
ApolloClient.Builder()
    .httpEngine(...)
    .webSocketTimeoutMillis(...)
    // etc...
    .build()
```

Those are properties of `HttpNetworkTransport` and `WebSocketNetworkTransport` respectively and what we gain in the "simple" cases is outweight by the complexity induced (more docs, impossible states, bad error messages (https://github.com/apollographql/apollo-kotlin/issues/6253), etc...).

I am now convinced that [precedence rules are clumsy](https://www.youtube.com/watch?v=YZstpc2939s), and this PR removes a lot of the precedence. 

The code above can be replaced by the (longer but also much more consistent and less footgunny) version:

```kotlin
ApolloClient.Builder()
    .networkTransport(HttpNetworkTransport.Builder().httpEngine(engine).build())
    .subscriptionNetworkTransport(WebSocketNetworkTransport.Builder().timeoutMillis(timeoutMillis).build())
    // etc...
    .build()
```

I just kept `serverUrl()` because it's used everywhere in the docs.

### DIFF
--- a/libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/OkHttpExtensions.kt
+++ b/libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/OkHttpExtensions.kt
@@ -1,6 +1,9 @@
+@file:Suppress("DEPRECATION")
+
 package com.apollographql.apollo.network
 
 import com.apollographql.apollo.ApolloClient
+import com.apollographql.apollo.annotations.ApolloDeprecatedSince
 import com.apollographql.apollo.api.http.HttpHeader
 import com.apollographql.apollo.network.http.DefaultHttpEngine
 import com.apollographql.apollo.network.http.HttpNetworkTransport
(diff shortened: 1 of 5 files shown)


libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/ws/OkHttpWebSocketEngine.kt
```diff
@@ -13,10 +13,17 @@ import okhttp3.WebSocket
 import okhttp3.WebSocketListener
 import okio.ByteString
 
+/**
+ * See https://github.com/square/okhttp/pull/8248 for why we use a lambda here
+ */
 actual class DefaultWebSocketEngine(
-    private val webSocketFactory: WebSocket.Factory,
+    webSocketFactory: () -> WebSocket.Factory,
```
martinbonnin, line 20: New constructor. We need that to avoid initializing from the main thread

libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/OkHttpExtensions.kt
```diff
@@ -16,6 +19,18 @@ import okhttp3.OkHttpClient
...
+    ReplaceWith(
+        "networkTransport(HttpNetworkTransport.Builder().httpEngine(DefaultHttpEngine(okHttpClient)).build())" +
+            ".subscriptionNetworkTransport(WebSocketNetworkTransport.Builder().webSocketEngine(DefaultWebSocketEngine(okHttpClient)).build())",
+        "com.apollographql.apollo.network.http.HttpNetworkTransport",
+        "com.apollographql.apollo.network.http.DefaultHttpEngine",
+        "com.apollographql.apollo.network.ws.DefaultWebSocketEngine",
+        "com.apollographql.apollo.network.ws.WebSocketNetworkTransport"
+    )
+)
+@ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_1_1)
```
BoD, line 33: I worry a bit about this one in particular as it's probably one the most often used, and the replacement expression is a bit of mouthful... But I agree about the tradeoff, it's for the better.

BoD (approved):
More explicit > less explicit 👍

libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/OkHttpExtensions.kt
martinbonnin: That's a fair point. We don't want to become the `capitalized()` of GraphQL... Will think more about this

martinbonnin:
This has been open far too long. I'll get back to it at some point. In the meantime, closing it.

PR #6291, opened 2024-11-28
Bump Kotlin to 2.1.0

BoD:
Note: we keep the stdlib fixed to 2.0.0


build-logic/build.gradle.kts
```diff
@@ -61,7 +61,9 @@ java {
 }
 
 tasks.withType<JavaCompile>().configureEach {
-  options.release.set(17)
+  // Do not use options.release - see https://issuetracker.google.com/issues/278800528
+  sourceCompatibility = "17"
+  targetCompatibility = "17"
```
martinbonnin, line 66: IIRC, this is only for Android compile tasks. Java compile tasks should still use `options.release.set(17)`

gradle/libraries.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 # The version we use by default in tests and benchmarks
 # Can be as high as IJ supports
-android-plugin = "8.2.2"
+android-plugin = "8.7.0"
```
martinbonnin, line 4: Is this required to build with Kotlin 2.1.0? If yes, is there an AGP <-> KGP compatibility matrix somewhere?

build-logic/build.gradle.kts
martinbonnin: FWIW, this is how I do it in librarian: https://github.com/GradleUp/librarian/blob/2be199a3b4f3d3cac8f7560d101a138d64ae8f96/librarian-gradle-plugin/src/main/kotlin/com/gradleup/librarian/gradle/compatibility.kt#L50

This is definitely not tested against 2.1.0 so take with large grains of salt

gradle/libraries.toml
```diff
@@ -133,8 +133,11 @@ kotlin-stdlib-wasm-js = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib-
 kotlin-test = { group = "org.jetbrains.kotlin", name = "kotlin-test" } # the Kotlin plugin resolves the version
 kotlin-test-js = { group = "org.jetbrains.kotlin", name = "kotlin-test-js" } # the Kotlin plugin resolves the version
 kotlin-test-junit = { group = "org.jetbrains.kotlin", name = "kotlin-test-junit" } # the Kotlin plugin resolves the version
+kotlin-test-stdlib = { group = "org.jetbrains.kotlin", name = "kotlin-test", version.ref = "kotlin-stdlib" } # use the same version as the stdlib
+kotlin-test-junit-stdlib = { group = "org.jetbrains.kotlin", name = "kotlin-test-junit", version.ref = "kotlin-stdlib" } # use the same version as the stdlib
```
martinbonnin, line 137: This is https://youtrack.jetbrains.com/issue/KT-58999/ coming back at us. Ideally we don't need all of that because `apiVersion`. Actually I wonder if we could implement that ourselved 🤔 . I'll try to open a PR

libraries/apollo-debug-server/src/commonMain/kotlin/com/apollographql/apollo/debugserver/internal/graphql/GraphQL.kt
```diff
@@ -49,12 +49,13 @@ internal class GraphQL(
  * The root query
  */
 @GraphQLQuery
-internal class Query(private val apolloClients: AtomicReference<Map<ApolloClient, String>>) {
+// Prefixing private fields with _ to work around https://github.com/google/ksp/issues/2135
```
martinbonnin, line 52: The CI left quite a trail in https://github.com/google/ksp/issues/2135 😅

gradle/libraries.toml
martinbonnin: 👍 I'm all for more granular releases

Re: build-time AGP, I'm team "update to the latest version that's working in IJ" but I can never remember which one it is 🙃

gradle/libraries.toml
martinbonnin: https://github.com/apollographql/apollo-kotlin/pull/6293

gradle/libraries.toml
BoD: Ah there is something [here](https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#version-compatibility) actually. It says AGP 8.2.2 should be compatible with KGP 2.1.0 👍

PR #6295, opened 2024-12-02
Add KDoc for ApolloInterceptor

martinbonnin:


BoD (approved):
👍

Should we also add a note in [the guide](https://www.apollographql.com/docs/kotlin/advanced/interceptors-http)?

martinbonnin:
> Should we also add a note in [the guide](https://www.apollographql.com/docs/kotlin/advanced/interceptors-http)?

I wouldn't necessarily. [There's some value in the guide being small.](https://publicobject.com/2024/10/25/it-needs-documentation/)

PR #6308, opened 2024-12-09
Compatibility with Gradle 9

martinbonnin:
See https://docs.gradle.org/8.11/userguide/upgrading_version_8.html#deprecate_get_dependency_project


BoD (approved):
👍

PR #6309, opened 2024-12-09
[gradle-plugin] Fix classloader caching. 

martinbonnin:
So many things went wrong there 🙈

Many thanks @scana for catching this 🙏


BoD (approved):
🐘

PR #6325, opened 2024-12-13
Add `ApolloExtension.useGradleVariants`

martinbonnin:
This adds the apollo metadata as variants in the "java" software component instead of requiring a separate publication. This should fix publishing modules with apollo dependencies.

This is wildly experiment and only for JVM targets due to https://youtrack.jetbrains.com/issue/KT-58830/

If using this, you can remove the apollo publication:

```kotlin
apollo {
  // add this (experimental)
  useGradleVariants.set(true)
}

// remove this
publishing {
  publications {
    create<MavenPublication>("apollo") {
      from(components["apollo"])
      artifactId = "foo-apollo"
    }
  }
}
```


BoD (approved):
👍

PR #6329, opened 2024-12-17
[gradle-plugin] expose the outgoing variants

martinbonnin:
Instead of trying to guess what SoftwareComponent to add the metadata to, just expose them to the user so they can decide what to do with them:

```kotlin
  service("service") {
    outgoingVariantsConnection {
      addToSoftwareComponent("java")
    }
  }
```

Note that Android requires `afterEvaluate {}`:

```kotlin
  service("service") {
    outgoingVariantsConnection {
      afterEvaluate {
        addToSoftwareComponent("release")
      }
    }
  }
```


BoD (approved):
⚙️

PR #6331, opened 2024-12-17
Implement ＠defer "June 2023" response format

BoD:
Implements the latest (as of 2024-12-17) ＠defer incremental response format as seen in [examples](https://github.com/graphql/defer-stream-wg/discussions/69) and [spec edits](https://github.com/graphql/graphql-spec/pull/1124).

This shouldn't be merged as-is, as at least the end-to-end tests with the Router will fail.

There may be a way to have both the "old" (`20220824`) and latest formats working but not sure this would be useful.


martinbonnin:
Apollo Client 4 introduces different [handlers](https://github.com/apollographql/apollo-client/blob/e1590744210962dd66a46aa21016a22225817996/docs/source/migrating/apollo-client-4-migration.mdx#enable-incremental-delivery-defer) for different incremental formats. We could do the same while the format iterates.

.github/workflows/defer-integration-tests.yml
```diff
@@ -42,3 +42,31 @@ jobs:
...
+        run: |
+          npm install --legacy-peer-deps
+          npx patch-package
+          APOLLO_PORT=4000 npm start &
+
+      - name: Setup Java
+        uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 #v4.2.1
+        with:
+          distribution: 'temurin'
+          java-version: 17
```
martinbonnin, line 63: Any thoughts about removing this? I'm trying to keep our yaml files clean (see https://github.com/apollographql/apollo-kotlin/pull/6705#discussion_r2375537148)

BoD: No objection either 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/internal/incremental/Defer20220824IncrementalResultsMerger.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.apollographql.apollo.api.IncrementalResultIdentifier
+import com.apollographql.apollo.api.IncrementalResultIdentifiers
+import okio.BufferedSource
+
+/**
+ * Merger for the [com.apollographql.apollo.network.http.HttpNetworkTransport.IncrementalDeliveryProtocol.Defer20220824] protocol format.
+ */
+@ApolloInternal
+@Suppress("UNCHECKED_CAST")
+class Defer20220824IncrementalResultsMerger : IncrementalResultsMerger {
```
martinbonnin, line 13: Can this be `internal` ?

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/HttpNetworkTransport.kt
```diff
@@ -392,4 +406,44 @@ private constructor(
...
+
+  /**
+   * The protocol to use for incremental delivery (`@defer` and `@stream`).
+   */
+  sealed interface IncrementalDeliveryProtocol {
+    @ApolloInternal
+    val acceptHeader: String
+
+    @ApolloInternal
+    fun newIncrementalResultsMerger(): IncrementalResultsMerger
```
martinbonnin, line 418: What about a sealed marker interface to avoid exposing the interface functions? I also think it needs to be @ApolloExperimental?

```kotlin
@ApolloExperimental
sealed interface IncrementalResultsMerger
```

I don't think we want to allow users to implement their own `IncrementalResultsMerger` at this point?

BoD: I went this route: 
- `IncrementalDeliveryProtocol` is a public enum
- `IncrementalDeliveryProtocolImpl` are internal `IncrementalResultsMerger` sealed interfaces

martinbonnin: Enum is even better 👍

libraries/apollo-api/api/apollo-api.api
```diff
@@ -713,6 +713,12 @@ public final class com/apollographql/apollo/api/ImmutableMapBuilder {
 	public final fun put (Ljava/lang/Object;Ljava/lang/Object;)Lcom/apollographql/apollo/api/ImmutableMapBuilder;
 }
 
+public final class com/apollographql/apollo/api/IncrementalResultIdentifierKt {
+	public static final fun isPending (Ljava/util/Set;)Z
+	public static final fun nonPending (Ljava/util/Set;)Ljava/util/Set;
+	public static final fun pending (Ljava/util/Set;)Ljava/util/Set;
+}
```
martinbonnin, line 720: Can those be removed from public API? `isPending` can be internal I think, `nonPending` can be moved to `apollo-runtime` and `pending` is only used in tests.

libraries/apollo-api/api/apollo-api.api
```diff
@@ -938,9 +946,6 @@ public final class com/apollographql/apollo/api/http/DefaultHttpRequestComposer$
 	public final fun buildPostBody (Lcom/apollographql/apollo/api/Operation;Lcom/apollographql/apollo/api/CustomScalarAdapters;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)Lcom/apollographql/apollo/api/http/HttpBody;
 	public final fun buildPostBody (Lcom/apollographql/apollo/api/Operation;Lcom/apollographql/apollo/api/CustomScalarAdapters;ZZLjava/lang/String;)Lcom/apollographql/apollo/api/http/HttpBody;
 	public final fun composePayload (Lcom/apollographql/apollo/api/ApolloRequest;)Ljava/util/Map;
-	public final fun getHEADER_ACCEPT_NAME ()Ljava/lang/String;
-	public final fun getHEADER_ACCEPT_VALUE_DEFER ()Ljava/lang/String;
-	public final fun getHEADER_ACCEPT_VALUE_MULTIPART ()Ljava/lang/String;
```
martinbonnin, line 943: We need a deprecation cycle before removing these.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -32,17 +33,23 @@ import okio.buffer
...
 class DefaultHttpRequestComposer(
     private val serverUrl: String,
+    private val acceptHeaderQueriesAndMutations: String = HEADER_ACCEPT_VALUE_QUERIES_AND_MUTATIONS,
 ) : HttpRequestComposer {
 
+  constructor(serverUrl: String) : this(
+      serverUrl = serverUrl,
+      acceptHeaderQueriesAndMutations = HEADER_ACCEPT_VALUE_QUERIES_AND_MUTATIONS,
+  )
+
```
martinbonnin, line 43: Instead of adding another constructor, can we handle all the HTTP headers in the request ([here](https://github.com/apollographql/apollo-kotlin/blob/a4d459cdde8770c0230a348c4574ccb56ed413ba/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/HttpNetworkTransport.kt#L58))?

```kotlin
  override fun <D : Operation.Data> execute(
      request: ApolloRequest<D>,
  ): Flow<ApolloResponse<D>> {
    val customScalarAdapters = request.executionContext[CustomScalarAdapters]!!

    val request = if (request.httpHeaders.orEmpty().none { it.name.lowercase() == "accept" }) {
      val accept = if (request.operation is Subscription<*>) {
        "multipart/mixed;subscriptionSpec=1.0, application/graphql-response+json, application/json"
      } else {
        incrementalDeliveryProtocolImpl.acceptHeader
      }
      request.newBuilder().addHttpHeader("accept", accept).build()
    } else {
      request
    }
    val httpRequest = httpRequestComposer.compose(request)

    return execute(request, httpRequest, customScalarAdapters)
  }
```

At the end of the day, I'd like to consolidate all the headers there and probably also remove `HttpNetworkTransport.Builder.addHttpHeader()` (but that's another story)

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/HttpNetworkTransport.kt
```diff
@@ -218,17 +222,16 @@ private constructor(
               else -> null
             }
           } else {
```
martinbonnin, line 224: Is there a way to test from the response whether it supports the incremental delivey protocol? Is there something in `response.headers["content-type"]` that signals what protocol was used by the server?

libraries/apollo-runtime/src/commonTest/kotlin/test/defer/GraphQL17Alpha9IncrementalResultsMergerTest.kt
```diff
@@ -0,0 +1,2511 @@
...
+    incrementalResultsMerger.merge(payload4.buffer())
+    assertEquals(jsonToMap(mergedPayloads_1_2_3_4), incrementalResultsMerger.merged)
+    assertEquals(
+        setOf(),
+        incrementalResultsMerger.incrementalResultIdentifiers.nonPending()
+    )
+  }
+
+  /**
+   * Example J from https://github.com/graphql/defer-stream-wg/discussions/69 (Jul 18 2025 version)
```
martinbonnin, line 2348: Nice!

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketNetworkTransport.kt
```diff
@@ -195,14 +194,13 @@ class WebSocketNetworkTransport private constructor(
...
 
-private object DefaultSubscriptionParserFactory: SubscriptionParserFactory {
+private object DefaultSubscriptionParserFactory : SubscriptionParserFactory {
   override fun <D : Operation.Data> createParser(request: ApolloRequest<D>): SubscriptionParser<D> {
     return DefaultSubscriptionParser(request)
   }
 }
 
 private class DefaultSubscriptionParser<D : Operation.Data>(private val request: ApolloRequest<D>) : SubscriptionParser<D> {
-  private var deferredJsonMerger: DeferredJsonMerger = DeferredJsonMerger()
```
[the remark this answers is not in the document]
BoD: You’re right, I was thinking subscriptions (unsupported) but WS is not just subs. I’ve reverted.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Executables.kt
```diff
@@ -71,8 +71,8 @@ fun <D : Executable.Data> Executable<D>.parseData(
     jsonReader: JsonReader,
     customScalarAdapters: CustomScalarAdapters = CustomScalarAdapters.Empty,
     falseVariables: Set<String>? = null,
-    deferredFragmentIds: Set<DeferredFragmentIdentifier>? = null,
-    errors: List<Error>? = null
+    deferredFragmentIds: IncrementalResultIdentifiers? = null,
```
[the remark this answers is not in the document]
martinbonnin: LGTM

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -39,14 +40,20 @@ class DefaultHttpRequestComposer(
...
       }
+      if (get("accept") == null) {
+        add(
+            HttpHeader("accept",
+                if (apolloRequest.operation is Subscription<*>) {
+                  "multipart/mixed;subscriptionSpec=1.0, application/graphql-response+json, application/json"
+                } else {
+                  "multipart/mixed;deferSpec=20220824, application/graphql-response+json, application/json"
+                }
+            )
```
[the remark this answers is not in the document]
martinbonnin: What about `application/graphql-response+json, application/json` as a fallback? This is the most standard `accept` header. This way `DefaultHttpRequestComposer` is isolated from the incremental delivery concerns. That's a behavior change but probably very small blast radius?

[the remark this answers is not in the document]
BoD: Works for me 👍

libraries/apollo-runtime/api/android/apollo-runtime.api
```diff
@@ -240,6 +229,38 @@ public final class com/apollographql/apollo/internal/MultipartReader$Part : java
...
+	public fun getHasNext ()Z
+	public fun getIncrementalResultIdentifiers ()Ljava/util/Set;
+	public fun getMerged ()Ljava/util/Map;
+	public fun isEmptyResponse ()Z
+	public fun merge (Ljava/util/Map;)Ljava/util/Map;
+	public fun merge (Lokio/BufferedSource;)Ljava/util/Map;
+	public fun reset ()V
+}
+
+public final class com/apollographql/apollo/internal/incremental/GraphQL17Alpha9IncrementalResultsMerger : com/apollographql/apollo/internal/incremental/IncrementalResultsMerger {
```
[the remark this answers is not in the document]
BoD: In the end, I opted for this naming in the API:
- `DraftInitial`: our initial implem, still the default
- `Draft0_1`: what will be on https://specs.apollo.dev/incremental/v0.1/ very soon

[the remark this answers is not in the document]
BoD: Yes I like it! But then it would make sense to have specs.apollo.dev/incremental/v0.0 to match it.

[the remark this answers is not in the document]
martinbonnin: > it would make sense to have specs.apollo.dev/incremental/v0.0 to match it.

Yea, would be nice of course but if we don't have time for this, I think it's still fine. Besides me, everyone finds 0.0 to be a weird version name. So I think most people would understand if 0.0 is "unreleased initial version"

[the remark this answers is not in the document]
martinbonnin: If we don't have it, add a KDoc comment: "0.0 is the `@defer` protocol used as of 20220824. It has not been specified and is kept for backward compatibility reasons", something like this.

.github/workflows/defer-with-router-tests.yml
```diff
@@ -24,9 +24,23 @@ jobs:
       - run: |
           ./router --supergraph tests/defer/router/simple-supergraph.graphqls &
 
-      - uses: gradle/actions/setup-gradle@dbbdc275be76ac10734476cc723d82dfe7ec6eda #v3.4.2
```
martinbonnin: 👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -38,24 +39,30 @@ import okio.buffer
...
-      if (apolloRequest.operation is Subscription<*>) {
-        add(HttpHeader(HEADER_ACCEPT_NAME, HEADER_ACCEPT_VALUE_MULTIPART))
-      } else {
-        add(HttpHeader(HEADER_ACCEPT_NAME, HEADER_ACCEPT_VALUE_DEFER))
-      }
       if (apolloRequest.httpHeaders != null) {
         addAll(apolloRequest.httpHeaders)
       }
+      if (get("accept") == null) {
+        add(
```
martinbonnin:
```suggestion
        /** 
          * This is for backward compatibility reasons only. 
          * We should encorage users to set the accept headers before calling DefaultHttpRequestComposer         
          */
        add(
```

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/BooleanExpression.kt
```diff
@@ -74,18 +75,29 @@ fun BooleanExpression<BTerm>.evaluate(
...
     // By default, parse all deferred fragments - this is the case when parsing from the normalized cache.
     return true
   }
-  return deferredFragmentIdentifiers.contains(DeferredFragmentIdentifier(path, label))
+  val identifier = DeferredFragmentIdentifier(path, label)
+  return if (deferredFragmentIdentifiers.isPending()) {
+    // Modern (v0.1 draft) protocol: parse fragments that are _not_ pending
+    !deferredFragmentIdentifiers.contains(identifier)
+  } else {
+    // Legacy (v0.0 draft) protocol: parse fragments that have been merged
```
martinbonnin:
```suggestion
    // v0.1: parse fragments that have been merged
```

martinbonnin (commented):
A few comments but this looks good! Almost there!

martinbonnin (approved):
🎉

PR #6343, opened 2025-01-09
Added exclude rule with dependsOn

NaveenPokhriyal:
Erlier with version 3 we have control over depencies which to include or exclude for apollometadata however it's not the case with newer version of apollo that's why adding support for excluding dependencies from the metadata.
ex : Apollo V3 
```
apolloMetadata(libs.apollo.commomodels.metadata.get().toString()){
        exclude group: 'com.x.y.z', module: 'main-api-schema'
 }
```

Apollo V4 : with this implementation
```
dependsOn(
            dependencyNotation = libs.apollo.commomodels.metadata.get().toString(),
            excludes = listOf(
                DefaultExcludeRule(group = "com.x.y.z", module = "main-api-schema"),
                DefaultExcludeRule(group = "com.example", module = "another-module")
            )
        )
```

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
@@ -12,12 +12,14 @@ import com.apollographql.apollo.compiler.PackageNameGenerator
 import org.gradle.api.Action
 import org.gradle.api.Task
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.ExcludeRule
 import org.gradle.api.component.SoftwareComponent
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFile
 import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.internal.artifacts.DefaultExcludeRule
 import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
@@ -775,8 +777,12 @@ interface Service {
   /**
    * Adds a dependency for the codegen. Use this when some types/fragments are generated
    * in upstream modules.
+   *
+   * @param excludeRules List of [DefaultExcludeRule] to specify which dependencies should be excluded.
+   * Each rule can define group and/or module to exclude.
+   * Default is an empty list, meaning no exclusions.
    */
-  fun dependsOn(dependencyNotation: Any)
+  fun dependsOn(dependencyNotation: Any, excludeRules: List<DefaultExcludeRule> = emptyList())
 
   /**
    * Counterpoint of [dependsOn]. [isADependencyOf] allows a schema module to discover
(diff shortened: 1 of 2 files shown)


martinbonnin:
Thanks for opening this 🙏 

That's a fair point and something that was indeed lost in v4. I'm a bit wary of adding too many options there since the dependency resolution APIs in Gradle are probably more than just exclude, we might want to revert to a single configuration per service that is extended by all the other ones.

Out of curiosity, what's the use case? I didn't expect dependency conflicts to happen there.

NaveenPokhriyal:
> Out of curiosity, what's the use case? I didn't expect dependency conflicts to happen there.

So we have a common module which basicaly inherit a schema module as an dependency so whenever we are publishing the common module metadata it also including metadata info of parenet schema module.
as i have seen this is default behaviour of Mavenpublication, 

> void from([SoftwareComponent](https://docs.gradle.org/current/javadoc/org/gradle/api/component/SoftwareComponent.html) component)
> Provides the software component that should be published.
> 
> * Any artifacts declared by the component will be included in the publication.
> * The dependencies declared by the component will be included in the published meta-data.

https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPublication.html

so to resolve this issue earlier i used to exclude parent schema module from the artifact but with V4 i can't do it anymore.

NaveenPokhriyal:
> common/build.gradle.kts
> 
> ```kotlin
> dependencies {
>   implementation(project(":schema"))
> }
> 
> apollo {
>   service("service") {
>     generateApolloMetadata.set(true)
>   }
> }
> ```
> 
> schema/build.gradle.kts
> 
> ```kotlin
> apollo {
>   service("service") {
>     generateApolloMetadata.set(true)
>   }
> }
> ```
> 
Yes we have similar setup but common does not dependsOn the schema.

martinbonnin:
> Yes we have similar setup but common does not dependsOn the schema.

So then there is no need to exclude it or is there? 

FWIW, I think you have a good idea and allowing to customize dependency resolution in general is a good idea but doing it in a future-proof way might take a bit of design time and I'm trying to understand if there's a way you could have your setup working without this.

NaveenPokhriyal:
> So then there is no need to exclude it or is there?
> 
Yes that issue is there cause of that dependency. but i have a local workaround for this.
where i have created a filtered configuration without this dependency by modifying AdhocComponentWithVariants

martinbonnin:
@NaveenPokhriyal I'm going to close this in favor of https://github.com/apollographql/apollo-kotlin/pull/6408. #6408 adds 2 new user-exposed configurations (per service) that you can use in your `dependencies {}` block with all the tools from Gradle dependency management: 


*  schema/build.gradle.kts

```kotlin
apollo {
  service("service") {
    packageName.set("com.example")]
    schemaFiles.from("src/main/graphql/schema.graphqls")
    generateApolloMetatata.set(true)
    alwaysGenerateTypesMatching.set(emptyList())
  }
}

// This needs to be **after** the apollo {} block
dependencies {
  // The name of the configuration is "apollo${serviceName.capitalized()}UsedCoordinates"
  add("apolloServiceUsedCoordinates", project(":feature"))
}
```

*  feature/build.gradle.kts

```kotlin
apollo {
  service("service") {
    packageName.set("com.example")]
    generateApolloMetatata.set(true)
    alwaysGenerateTypesMatching.set(emptyList())
  }
}

dependencies {
  // The name of the configuration is "apollo${serviceName.capitalized()}"
  add("apolloService", project(":schema"))
}
```

I think it'll allow you to add exclude rules as you see fit. Let us know if it doesn't. In all cases, thank you very much for opening this discussion 🙏

NaveenPokhriyal:
Hey @martinbonnin , thanks for the update sure this could help with the missing configurations. 👍

Issue #6332, opened 2024-12-17
Use Kotlin's Inline value classes for custom scalars


x-sheep:
### Use case

[Inline value classes](https://kotlinlang.org/docs/inline-classes.html#members) are a thin wrapper over other types, which can potentially be erased entirely by the compiler.

```kotlin
value class ID(val value: String)
```

The above is safer to use in Client code, since it's not possible to assign a `String` to an `ID` or vice versa. It also discourages (but doesn't prevent) string manipulation by the client, which is useful for scalars that are documented to be opaque.

### Describe the solution you'd like

The various shortcuts like `mapScalarToKotlinLong`, `mapScalarToKotlinString` etc. could be expanded to take an additional argument `asValueClass: Boolean` which will generate a value class instead of a typealias.

I'm not sure if it's preferable to generate another Adapter for this new type, or if its usage should be applied inside every existing object adapter:
 ```kotlin
val id = ID(value = StringAdapter.fromJson(_, _)

val json = StringAdapter.toJson(_, _, id.value)
```
I think avoiding a new adapter would allow the compiler to inline the type in more cases and increase performance.

**Comment by BoD** [comment]
Thanks for the excellent suggestion!

The way it works currently doesn't involve typealiases, when using e.g. `mapScalarToKotlinLong`, the generated model will use a `Long` directly.

Currently, you can uses value classes for scalars with adapters looking like this:

```kotlin
val IDAdapter = object : Adapter<ID> {
  override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): ID {
    return ID(reader.nextString()!!)
  }

  override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: ID) {
    writer.value(value.value)
  }
}
```

and configuring it like:

```kotlin
mapScalar(graphQLName = "ID", targetName = "com.example.ID", expression = "com.example.IDAdapter")
```

To make things less manual, the library could do 2 things:

- A/ generate the scalar value class and make the generated code box/unbox it before using the existing String/Long/etc. adapter
- B/ let the user define their scalar value class (like now) and make the generated code box/unbox it before using the existing String/Long/etc. adapter

### A/ Generate scalar

Configuration:

```kotlin
mapScalarToKotlinLong("Length", asValueClass = true)
```

Would generate your scalar:

```kotlin
package package com.example

@JvmInline
value class Length(val value: Long)
```

Generated model adapter:

```kotlin
[...]

PR #6352, opened 2025-01-15
Add inline classes support for scalars

BoD:
This adds 2 things:

1. Scalar mappings can indicate the target is an inline class, and the name of the wrapped property.
When doing so, the generated adapter calls the primitive adapter on the wrapped property to avoid boxing/unboxing.
To do this, new versions of `mapScalarToKotlinXyz()` are added that take the inline class target name and property name.

2. Ability to generate inline classes for scalars.
Done by applying a new `@inlineClass` directive, e.g. `extend scalar Length @inlineClass(coercion: Long)`
This will generate:
```kotlin
package com.example.xyz.type.scalar

@JvmInline
public value class Length(
  public val `value`: Long,
)
``` 
When doing so, the mapping is also added (no need to call `mapScalarToXyz()`).

Related to #6332.


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -135,41 +135,104 @@ interface Service {
...
    * Map the given GraphQL scalar to [kotlin.Any] and use the builtin adapter
    */
   fun mapScalarToKotlinAny(graphQLName: String)
 
+  /**
+   * Map the given GraphQL scalar to a [kotlin.Any] wrapped in an inline class and use the builtin adapter
+   */
+  @ApolloExperimental
+  fun mapScalarToKotlinAny(graphQLName: String, inlineClass: String, inlineClassProperty: String)
+
```
martinbonnin, line 235: Can you remove the overloads? 

* They sound a bit weird to me as it's not mapping to String/Boolean/etc... but to a generated inline class that as a String/Boolean/etc... property
* They create a secondary source of truth: what happens if someone uses that + `@inlineClass` ? Who takes precedence? [Precedence is bad](https://youtu.be/YZstpc2939s?t=1965).
* Let's preserve option value? We can always add them later if needed?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/definitions.kt
```diff
@@ -174,9 +174,22 @@ on OBJECT
...
+Since: 4.1.1
+""${'"'}
+enum InlineClassCoercion { String, Boolean, Int, Long, Float, Double, Any }
+
+""${'"'}
+Generate an inline class for the given scalar type. The wrapped type is determined by the `coercion` argument.
+Since: 4.1.1
+""${'"'}
+directive @inlineClass(coercion: InlineClassCoercion!) on SCALAR
+"""
```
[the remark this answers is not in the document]
martinbonnin: Also reminder to add the matching definition on https://github.com/apollographql/specs if we agree on this

libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
martinbonnin: Note that the 2nd point is also a more general problem of `@inlineClass`. It may conflict with the Gradle mapping but at least there's a clear limit:

* user writes the target class => Gradle
* codegen writes the target class => `@inlineClass`

BoD: Yeah actually I agree, they're a bit much and not particularly clear. I was thinking of the case where you have your own inline class, but for that you can call `mapScalar` directly.

BoD:
#### Generated Kotlin

```kotlin
public object InlineClassQuery_ResponseAdapter {
  public object Data : Adapter<InlineClassQuery.Data> {
    public val RESPONSE_NAMES: List<String> = listOf("nonNullableLength")

    override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): InlineClassQuery.Data {
      var _nonNullableLength: Length? = null

      while (true) {
        when (reader.selectName(RESPONSE_NAMES)) {
          0 -> _nonNullableLength = custom.scalars.type.scalar.Length(com.apollographql.apollo.api.LongAdapter.fromJson(reader, customScalarAdapters))
          else -> break
        }
      }

      return InlineClassQuery.Data(
        nonNullableLength = _nonNullableLength ?: missingField(reader, "nonNullableLength")
      )
    }

    override fun toJson(
      writer: JsonWriter,
      `value`: InlineClassQuery.Data,
    ) {
      writer.name("nonNullableLength")
      com.apollographql.apollo.api.LongAdapter.toJson(writer, customScalarAdapters, value.nonNullableLength.value)
    }
  }
}
```

#### Decompiled Java equivalent
```java
  @NotNull
  public InlineClassQuery.Data fromJson(@NotNull JsonReader reader, @NotNull CustomScalarAdapters customScalarAdapters) {
     Intrinsics.checkNotNullParameter(reader, "reader");
     Intrinsics.checkNotNullParameter(customScalarAdapters, "customScalarAdapters");
     Length _nonNullableLength = null;

     while(true) {
        int var4 = reader.selectName(RESPONSE_NAMES);
        if (var4 != 0) {
           InlineClassQuery.Data var10000 = new InlineClassQuery.Data;
           if (_nonNullableLength != null) {
              var10000.<init>(_nonNullableLength.unbox-impl(), (DefaultConstructorMarker)null);
              return var10000;
           } else {
              Assertions.missingField(reader, "nonNullableLength");
              throw new KotlinNothingValueException();
           }
        }

        _nonNullableLength = Length.box-impl(Length.constructor-impl(((Number)Adapters.LongAdapter.fromJson(reader, customScalarAdapters)).longValue()));
     }
  }

  public void toJson(@NotNull JsonWriter writer, @NotNull CustomScalarAdapters customScalarAdapters, @NotNull InlineClassQuery.Data value) {
     Intrinsics.checkNotNullParameter(writer, "writer");
     Intrinsics.checkNotNullParameter(customScalarAdapters, "customScalarAdapters");
     Intrinsics.checkNotNullParameter(value, "value");
     writer.name("nonNullableLength");
     Adapters.LongAdapter.toJson(writer, customScalarAdapters, value.getNonNullableLength-7cMOgQw());
  }
```

- In `fromJson`, `_nonNullableLength` is a nullable `Length` and boxing/unboxing happens :(  I wonder if we could avoid it, while keeping the `missingField` behavior...

- In `toJson`, `value.getNonNullableLength-7cMOgQw()` returns a `long`: no boxing 👍

martinbonnin:
> In fromJson, _nonNullableLength is a nullable Length and boxing/unboxing happens :( I wonder if we could avoid it, while keeping the missingField behavior...

I guess we could do it by using a separate boolean. No idea if it's worth the added complexity without additional benchmarks though.

BoD:
> I guess we could do it by using a separate boolean.

Yeah thought of using booleans... honestly I'm not convinced it's worth it.

I think this is still useful even with this un/boxing?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/definitions.kt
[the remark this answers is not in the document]
martinbonnin: Also reminder to add support in the IDE plugin

Screenshot 2025-01-15 at 19 52 03

martinbonnin:
Superseded by https://github.com/apollographql/apollo-kotlin/pull/6404

Issue #6327, opened 2024-12-16
Introspection Query Parse Error with Apollo GraphQL Schema Download


kibitheM:
### Question

Hello There,

I'm encountering an issue while attempting to download a schema from my GraphQL endpoint using Apollo. I am working on integrating Apollo GraphQL with a Supabase project. Here's the configuration I'm using in my `build.gradle.kts` file:

```kotlin
apollo {
    service("service") {
        packageName.set("com.ngumi.app")
        srcDir("src/main/graphql")

        introspection {
            endpointUrl.set("https://peclvhlyxzwmegdkhtvj.supabase.co/graphql/v1
")
            headers.put("apikey", "<my-api-key>")
        }
         codegenModels.set("operationBased")
    }
}


When I run the following Gradle task:

./gradlew downloadApolloSchema

Bash
It fails with the following error:

Execution failed for task ':app:downloadApolloSchema'.
> java.lang.Exception: Introspection response from [https://<your-supabase-endpoint>](https://peclvhlyxzwmegdkhtvj.supabase.co/graphql/v1).supabase.co/graphql/v1 can not be parsed
Use code [with caution](https://support.google.com/legal/answer/13505487).
Debugging Steps Taken
To investigate further, I directly fetched the introspection query response from the server using Invoke-WebRequest in PowerShell. Here's the response I received:

{
    "errors": [
        {
            "message": "query parse error: Parse error at 1:787\nUnexpected `}[Punctuator]`\nExpected `end of input`\n"
        }
    ]
}


I also used the Apollo sandbox to run the introspection query, and it was successful and returned a full schema.

Additional Context
The GraphQL endpoint is functional, as I've successfully executed other queries through Postman and the Apollo sandbox.

The parse error suggests that there is an unexpected } character at the end of the introspection query being sent (Based on the error provided by my supabse endpoint).

Running the Gradle task with --info and --debug did not provide additional clarity beyond pointing to the parse error.

Questions
Is it possible that the introspection query generated by Apollo is malformed, or it is not compatible with the version of the GraphQL server that my Supabase endpoint is using? If so, where can I adjust it?

Are there any known issues with Apollo's handling of introspection queries on Supabase-hosted GraphQL endpoints?

Is there a way to customize or manually supply the introspection query used by Apollo? Or an alternative approach to downloading a schema when introspection is not possible?

I appreciate any help or guidance the team can provide! Please let me know if additional information is needed.

Thank you in advance for your support!

[...]

Issue #6127, opened 2024-08-19
[Android] Apollo Introspection JsonDecodingException with Hasura server


javierpe:
### Version

4.0.0

### Summary

Hi, Im trying download schema from Hasura GraphQL but I get an exception.

My Apollo Gradle config:

`apollo {
    service("service") {
        packageName.set("com.survey.graphql")
        generateKotlinModels.set(true)
        generateSourcesDuringGradleSync.set(true)
        generatedSchemaName.set("schema")
        introspection {
            schemaFile.set(file("src/main/graphql/schema.graphql"))
            endpointUrl.set("http://mgsurveys24.mgroup.com.mx:1337/v1/graphql")
            headers.put("x-hasura-admin-secret", "---")
        }
    }
}`

With terminal and graphqurl works fine:
graphqurl http://mgsurveys24.mgroup.com.mx:1337/v1/graphql -H 'X-Hasura-Admin-Secret: ---' --introspect > schema.graphql

### Steps to reproduce the behavior
./gradlew :network:impl:downloadServiceApolloSchemaFromIntrospection

### Logs

```
Caused by: com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 611385: Expected valid boolean literal prefix, but had 'null' at path: $.data.__schema.directives[0].isRepeatable
JSON input: .....ue":null}],"isRepeatable":null},{"name":"skip","description".....
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(SourceFile:1)
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(SourceFile:2)
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.StringJsonLexer.fail(SourceFile:601)
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.StringJsonLexer.fail$default(SourceFile:1)
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.StringJsonLexer.consumeBoolean(SourceFile:15)
        at com.apollographql.apollo.relocated.kotlinx.serialization.json.internal.StringJsonLexer.consumeBooleanLenient$kotlinx$serialization$json$internal$AbstractJsonLexer(SourceFile:11)
[...]

PR #6360, opened 2025-01-21
Add a fail-safe mode to disable 2-step introspection and use minimal introspection query

BoD:
Should unblock users using problematic servers (e.g. #6327, #6127).


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloDownloadSchemaTask.kt
```diff
@@ -71,6 +71,11 @@ abstract class ApolloDownloadSchemaTask : DefaultTask() {
   @get:Option(option = "insecure", description = "if set to true, TLS/SSL certificates will not be checked when downloading")
   abstract val insecure: Property<Boolean>
 
+  @get:Optional
+  @get:Input
+  @get:Option(option = "failSafeIntrospection", description = "if set to true a minimal introspection query is used and no pre-introspection query")
+  abstract val failSafeIntrospection: Property<Boolean>
```
martinbonnin, line 77: What about falling back to minimal introspection query if everything else fails? 
One less parameter to maintain?

BoD: Good point! Simpler 👍

PR #6376, opened 2025-02-04
Use ephemeral NSURLSessionConfiguration by default

jvanderwee:
`NSURLCache` is not read from as `NSURLRequestReloadIgnoringCacheData` is set as the request policy [here](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-runtime/src/appleMain/kotlin/com/apollographql/apollo/network/http/DefaultHttpEngine.apple.kt#L102). Therefore it makes sense to use `ephemeralSessionConfiguration` to prevent writing to `NSURLCache`.

Related to https://github.com/apollographql/apollo-kotlin/issues/6375.


martinbonnin (approved):

BoD (approved):
Thanks! 👍

PR #6386, opened 2025-02-19
Update nullability.mdx

dhritzkiv:
Fix up `levels` syntax for `\@catch(to:` directive example


dhritzkiv:
I think `level` was from v0.1 of the nullability spec, whereas `levels` replaced it in v0.2. I'm not entirely sure, however.

BoD (approved):
Good catch!

PR #6388, opened 2025-02-19
Simplify semanticEquals

martinbonnin:
Instead of implementing custom logic, compare the SDL representations. It's probably less efficient but we're computing the SDL representation in most cases for the error message anyways. This makes `semanticEquals()` work on any `GQLNode`


BoD (approved):
👍

PR #6389, opened 2025-02-19
Remove `withBuiltInDefinitions()` and add flag to dump scalar definitions in SDL

martinbonnin:
Turns out removing the scalar creates a lot of complexity in the code. It's just easier to have a SDL schema matching 1:1 with the introspection. This is somewhat discouraged by the spec but I think this is just that the spec is mostly written with backend frameworks in mind and didn't consider clients wanting to validate operations, etc...


libraries/apollo-ast/src/jvmTest/kotlin/com/apollographql/apollo/graphql/ast/test/SDLWriterTest.kt
```diff
@@ -18,15 +19,6 @@ class SDLWriterTest {
...
   }
 
-  @Test
-  fun typeRedefinitionInspectionIsIgnored() {
-    val sdlSchema = File("${CWD}/test-fixtures/sdl/type_redefinitions.graphqls")
-
-    checkExpected(sdlSchema) {
-      it.parseAsGQLDocument().getOrThrow().withBuiltinDefinitions().toSDL("    ")
-    }
-  }
```
martinbonnin, line 28: I'm not 100% sure what this was testing. Removed it. Might add new ones later on.

BoD (approved):
👍

PR #6396, opened 2025-02-27
Refactor schema validation

martinbonnin:
**Terminology**:
* API schema: the server schema as seen from introspection (can be either JSON or SDL)
* Server definition: a definition present in the API schema.
  * Examples: `@include`, `@oneOf`, `__Schema` 
* Client definition: a definition that is added to the API schema by the client, using either `@link` or by concatenating new definitions to the schema.
* Linked definition: a definition is added to a schema using `@link`. 
  * Examples: `@targetName`, `@typePolicy` 
* Checked definitions: a definition that is recognized by Apollo Kotlin and required to be a certain shape. Apollo Kotlin checks those definitions to avoid crashes in the compiler and/or surprising behaviours. This is typically the case for server definitions that and are not specified yet and/or may vary depending on the version of the spec used.
  * Examples: `@oneOf`, `@semanticNonNull`, `@targetName` 

The PR adds 2 changes:
* allow duplicate linked definitions as long as they come from the same foreign schema.
* adds `@defer` and `kotlin_labs` to the checked definitions.

Besides that, it should behave like previously. This is mostly a refactoring that makes it easier to evolve that code. In particular, it keeps two issues we currently have:

* we're adding way too much built-in definitions by default, those should be limited to the built-in scalars.
* if a checked definition is also a server definition, we may have a name clash. It may be desired (`@semanticNonNull` for an example) or it may be annoying (`@catch` maybe used differently by some other tools)

How to fix this (if we want to fix it?) is left to the next breaking version.


BoD (approved):
👍

BoD:
Would it make sense to have this terminology in the glossary [here](https://github.com/apollographql/apollo-kotlin/blob/main/design-docs/Glossary.md)?

martinbonnin:
> Would it make sense to have this terminology in the glossary [here](https://github.com/apollographql/apollo-kotlin/blob/main/design-docs/Glossary.md)?

Always! Will add it 👍

PR #6399, opened 2025-02-27
Make logLevel property internal to not take it into account when caching

### LINKED ISSUE #6398: `ApolloTaskWithClasspath` is not cacheable if different Gradle log level options are used
### Version

4.0.0

### Summary

I know 4.0.0 is not the most recent release, but I think the issue is still there.

Here, Apollo plugin will construct options with the first enabled Gradle logger level: https://github.com/apollographql/apollo-kotlin/blob/a13fafd252c0d730eb3a5f742af1f2e35fe33f1d/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultApolloExtension.kt#L478

Eventually here, this log level is a task input: https://github.com/apollographql/apollo-kotlin/blob/a13fafd252c0d730eb3a5f742af1f2e35fe33f1d/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloTaskWithClasspath.kt#L28-L29

I may be wrong but I don't think log level affects the task outputs, and as such shouldn't be an input to the task? The problem with current implementation is that `./gradlew <apolloGeneratingTask>` and `./gradlew <apolloGeneratingTask> --quiet` will not reuse the cache because task inputs have changed. 

_No response_

### Logs

```
(Your logs here)
```

lwasyl:
Fixes #6398

### DIFF
--- a/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloTaskWithClasspath.kt
+++ b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloTaskWithClasspath.kt
@@ -25,7 +25,7 @@ abstract class ApolloTaskWithClasspath: DefaultTask() {
   @get:Input
   abstract val arguments: MapProperty<String, Any?>
 
-  @get:Input
+  @get:Internal
   abstract val logLevel: Property<LogLevel>
 
   @Inject


martinbonnin (approved):
Thanks!

scana:
Hey @martinbonnin is there any chance this could get released as a patch for `4.1.1` instead of in `4.2.0`?

Was talking to @erichaagdev from Gradle and we've thought of asking for it while investigating some build caching issues.

martinbonnin:
Hi @scana 👋 Is the concern of a timing concern? [`4.2.0` is to be released in a matter of minutes](https://github.com/apollographql/apollo-kotlin/actions/runs/14706576323). Would that work for you?

scana:
Oh wow - that totally works! Thank you @martinbonnin!

martinbonnin:
All the credit goes to @BoD who just fired up the release. It's online now: https://repo1.maven.org/maven2/com/apollographql/apollo/apollo-api/4.2.0/

PR #6406, opened 2025-03-05
Keep the @RequiresOptIn annotation in the Gradle plugin

martinbonnin:
If we do not keep it, we get those warnings when trying to use it:

```
> Configure project :runtime
w: file:///Users/martinbonnin/git/apollo-kotlin/tests/runtime/build.gradle.kts:22:5: Annotation 
'com.apollographql.apollo.annotations.ApolloExperimental' is not an opt-in requirement marker, 
therefore its usage in @OptIn is ignored
```


BoD (approved):
👍

PR #6408, opened 2025-03-05
[gradle-plugin] expose regular configurations for multi-module projects

martinbonnin:
This allows to configure dependency resolution using the full Gradle APIs. See https://github.com/apollographql/apollo-kotlin/pull/6343.

*  schema/build.gradle.kts

```kotlin
apollo {
  service("service") {
    packageName.set("com.example")]
    schemaFiles.from("src/main/graphql/schema.graphqls")
    generateApolloMetatata.set(true)
    alwaysGenerateTypesMatching.set(emptyList())
  }
}

dependencies {
  add("apolloServiceUsedCoordinates", project(":feature"))
}
```

*  feature/build.gradle.kts

```kotlin
apollo {
  service("service") {
    packageName.set("com.example")]
    generateApolloMetatata.set(true)
    alwaysGenerateTypesMatching.set(emptyList())
  }
}

dependencies {
  add("apolloService", project(":schema"))
}
```

One drawback is this breaks the IJ tooling model [here](https://github.com/apollographql/apollo-kotlin/blob/e111a4fb61967e6a0a4ecbff5a35a057bec1a34b/libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/DefaultApolloExtension.kt#L65) because the dependencies are not known until the configurations are resolved.

### DIFF
--- a/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/entrypoints.kt
+++ b/libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/entrypoints.kt
@@ -186,6 +186,9 @@ internal fun Consumer<String>.toLogger(): ApolloCompiler.Logger {
 @ApolloInternal
 fun Iterable<File>.findCodegenSchemaFile(): File {
   return firstOrNull {
+    /*
+     * TODO v5: simplify this and add a schema { } block to the Gradle configuration
+     */
     it.length() > 0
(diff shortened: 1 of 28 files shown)


BoD (approved):
👍

For reference, applying this on the multi module sample is [here](https://github.com/apollographql/apollo-kotlin-samples/compare/use-apollo-configurations?expand=1).

Issue #5872, opened 2024-05-01
Introduce a custom `ApolloResponse.toString()` 


rohandhruva:
### Use case

Since `ApolloResponse` changed to being a standard `class` (I think it used to be a `data class`?), the `toString()` now is not very useful because it prints the default hashcode / instance code based `toString`. 

Can we introduce a custom `toString` for `ApolloResponse`? While printing all of `data` does not make sense, I'm trying to think what might be useful -- perhaps the cache key?

martinbonnin:
Agreed. This is also recommended by the [Kotlin API guidelines](https://kotlinlang.org/docs/jvm-api-guidelines-debuggability.html#always-provide-a-tostring-method). While I wouldn't put `toString()` on each and every class, `ApolloResponse` is central enough that it deserves one.

martinbonnin:
Available in `4.2.0`

PR #6409, opened 2025-03-05
Add toString() method to ApolloResponse

rohandhruva:
First pass at adding a custom `toString()` to `ApolloResponse`.
Addresses #5872.

Open to suggestions on how this can be more useful, particularly the `data` portion of it.

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ApolloResponse.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ApolloResponse.kt
@@ -133,8 +133,7 @@ private constructor(
         "operationName=${operation.name()}, " +
         "data=${if (data == null) "null" else "${operation.name()}.Data"}, " +
         "errors=${errors?.size ?: "null"}, " +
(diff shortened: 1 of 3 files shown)


BoD (approved):
Thanks! 👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/ApolloResponse.kt
```diff
@@ -128,6 +128,16 @@ private constructor(
         .isLast(isLast)
   }
 
+  override fun toString(): String {
+    return "ApolloResponse(" +
+        "operationName=${operation.name()}, " +
+        "data=${if (data == null) "null" else "${operation.name()}.Data"}, " +
+        "errors=${errors?.size ?: "null"}, " +
+        "exception=${if (exception == null) "null" else exception::class.simpleName ?: "true"}, " +
+        "isLast=$isLast" +
```
martinbonnin, line 137: Can you remove `isLast` here? I'm not happy with this field and will probably deprecate it in the mid term. I don't want to advertise it too much.

martinbonnin (approved):
👍

PR #6416, opened 2025-03-07
Track `@ApolloExperimental` symbols in public API

martinbonnin:
Follow up from https://github.com/apollographql/apollo-kotlin/pull/6407: we also use `@ApolloExperimental` symbol accross module boundaries so we need to track those as well. 

In a perfect world, we need a specific `@OptIn` symbol for each module that is added to that specific module (`@ExperimentalApolloRuntimeApi`, ...). I'm too lazy to do it now and also don't want to introduce too much annotation noise so this is left for a future PR.

Changes to `@ApolloExperimental` apis will need to be evaluated manually whether they are safe to do (because not used by other modules) or require additional care.


BoD (approved):
👍

PR #6427, opened 2025-03-17
Inline classes, fix using inline classes in input positions

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/6332#issuecomment-2728670680


BoD (approved):
👍

Issue #6436, opened 2025-03-27
Could not read normalized cache - com/android/tools/idea/adblib/ddmlibcompatibility/UtilsK


tfcporciuncula:
### Version

4.1.1 (including Android Studio plugin)

### Summary

When I tap "Pull from device" and select the db file, this pops up:

![Image](https://github.com/user-attachments/assets/6d872e9c-d6ae-458b-8ca7-cbc7600e8cc0)

I tried both from a real device (Asus Zenfone 10 running Android 15) and from an emulator (API 34), and I'm on Android Studio Meerkat 2024.3.1.

### Steps to reproduce the behavior

See above.

### Logs

```
2025-03-27 09:52:07,966 [ 106451]   INFO - #com.android.ddmlib - Device 'Pixel_3a_API_34 [emulator-5554]' jdwp process '4221' is now known as pkg='com.wave.field.dbg' (proc='com.wave.field.dbg')
2025-03-27 09:52:13,592 [ 112077]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil - Command took 42.93 ms to execute: cat /proc/net/unix | grep apollo_debug_ | cat
2025-03-27 09:52:13,592 [ 112077]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil -   Output line 1 (out of 1): 
2025-03-27 09:52:13,624 [ 112109]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil - Command took 29.15 ms to execute: run-as com.wave.field.dbg ls -1 /data/data/com.wave.field.dbg/databases
2025-03-27 09:52:13,624 [ 112109]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil -   Output line 1 (out of 14): ppollo_wave_app
2025-03-27 09:52:13,624 [ 112109]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil -   Output line 2 (out of 14): apollo_wave_app-journal
2025-03-27 09:52:17,094 [ 115579]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil - Command took 22.12 ms to execute: touch /data/local/tmp/apollo_wave_app_+221761110040-tmp3750866015625969384.db
2025-03-27 09:52:17,095 [ 115580]   INFO - #com.android.tools.idea.adb.AdbShellCommandsUtil -   Output line 1 (out of 1): 
[...]

PR #6437, opened 2025-03-27
[IJ Plugin] Fix pulling normalized cache for recent AS

BoD:
Fix for #6436

There were 2 issues:
- `CoroutineDispatcher.limitedParallelism` referenced from `WebSocketNetworkTransport`, which we don't need
- `ddmlibcompatibility.toDeviceSelector` apparently no longer exists in AS, but we also don't need it

### DIFF
--- a/intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/apollodebugserver/ApolloDebugClient.kt
+++ b/intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/apollodebugserver/ApolloDebugClient.kt
@@ -3,12 +3,18 @@ package com.apollographql.ijplugin.apollodebugserver
 import com.android.ddmlib.IDevice
 import com.android.tools.idea.adb.AdbShellCommandsUtil
 import com.apollographql.apollo.ApolloClient
+import com.apollographql.apollo.api.ApolloRequest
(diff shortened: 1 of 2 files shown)


intellij-plugin/src/main/kotlin/com/apollographql/ijplugin/apollodebugserver/ApolloDebugClient.kt
```diff
@@ -65,6 +71,13 @@ class ApolloDebugClient(
       .addHttpInterceptor(LoggingInterceptor { line ->
         logd("ApolloDebugClient HTTP - $line")
       })
+      // Use a dummy NetworkTransport for subscriptions (which are not used), because the default one
+      // references CoroutineDispatcher.limitedParallelism which doesn't exist in certain IDE versions.
+      // (See https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#coroutinesLibraries)
```
martinbonnin, line 76: TIL. Is there a Youtrack to remove this limitation? We're able to workaround for now but it's not great moving forward.

BoD: I didn't find one, so I opened https://youtrack.jetbrains.com/issue/IJPL-182123

PR #6440, opened 2025-03-31
Implement defaultValues coercion

martinbonnin:
Closes https://github.com/apollographql/apollo-kotlin/issues/6433

Technically a breaking change, probably worthy of Apollo Kotlin 5, to be discussed.


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
```diff
@@ -759,6 +791,143 @@ private fun ValidationScope.validateInputObjects() {
...
+  if (state.visitedTypes.contains(inputObjectTypeDefinition.name)) {
+    return
+  }
+  state.visitedTypes.add(inputObjectTypeDefinition.name)
+
+  state.fieldPathIndexByTypeName[inputObjectTypeDefinition.name] = state.fieldPath.size
+
+  inputObjectTypeDefinition.inputFields.forEach {
+    val type = it.type
+    if (type is GQLNonNullType && type.type is GQLNamedType) {
```
BoD, line 817: Do I understand correctly that
```graphql
input A{
  x: A!
}
```
and
```graphql
input A{
  x: A = {}
}
```
are invalid, but
```graphql
input A{
  x: A
}
```
isn't?

martinbonnin: We understand the same thing, which I believe is the correct thing ^^

martinbonnin: Interestingly, lists also makes things valid:

```graphql
# this is valid because the list breaks the cycle
input A{
  x: [A!]!
}
```

BoD: Gotcha 👍 
I haven't followed the whole conversation on the WG ticket but I guess "there's nothing wrong until there is" 😅.

I would vote for it to be invalid in GraphQL 2 ;)

BoD (approved):

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
martinbonnin: For lists you mean? I think this one could stay in GraphQL 2 actually. There's no need to restrict too much:

```graphql 
# this is valid
{
  field(arg: { x: [{ x: [] }] })
}
```
Not that it's super useful but I don't think it hurts too much...

BoD: Actually I couldn't see a case where it can be useful - but yes it's pretty useful, e.g. `input Block { type: Type!, children: [Block]! = []}` and
```
{type: Page, children: [
  {type: Title}
  {type: Section, children: [
    {type: Paragraph}
  ]}
]}
```

Maybe it's not obvious at first sight, or maybe it's just me :D But yeah no reason to be against it actually.

PR #6466, opened 2025-04-14
Use npm instead of yarn

martinbonnin:
See https://kotlinlang.slack.com/archives/C0B8L3U69/p1740063113894839

Yarn doesn't seem to handle duplicates


martinbonnin:
Merging this and see how it goes

PR #6482, opened 2025-04-24
[4.x] Revert behaviour change

martinbonnin:
Revert part of https://github.com/apollographql/apollo-kotlin/commit/31009bc5db464b336e7c1a762acf55e6aeb1086a so that the behaviour of interceptors is not changed.


BoD (approved):
👍

PR #6485, opened 2025-04-28
Allow to generate Data Builders outside the main source set

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/5257

Apologies for the huge PR, this required a lot of changes to properly be able to separate the concerns.

To keep the diff to a non-crazy size, the test fixtures are update in a separate PR there: https://github.com/apollographql/apollo-kotlin/pull/6486. I'll merge it before merging this one.

From a user point of view, the changes are:

* All `buildFoo()` functions are now in separate files in a separate package:

```kotlin
// Replace
import com.example.type.buildCat

// With
import com.example.builder.buildCat
```

* The `Data {}` root function to build `Data` instances is now an extension function:

```kotlin
// Add
import com.example.builder.Data
```

* The `FakeResolver` is not passed automatically. If you are relying on fake values, you need to pass an instance of `FakeResolver` explicitely. This is to make it work better with https://github.com/apollographql/apollo-kotlin-faker

```kotlin
// Replace
val data = GetFooQuery.Data {}

// With 
val data = GetFooQuery.Data(fakeResolver) {}
```

* Similarly, if you are using runtime scalar adapters, they need to be passed explicitely

```kotlin
// Replace
val data = GetFooQuery.Data {}

// With 
val data = GetFooQuery.Data(customScalarAdapters = customScalarAdapters) {}
```

Not that it's possible to register all your scalar adapters at build time using the 3 parameters `mapScalar(graphQLName: String, targetName: String, expression: String)`.

* `FakeResolver` works exclusively in the Json domain now. There is a `FakeResolverContext.adaptToJson()` helper function to help with this:

```kotlin
// Replace
override fun resolveLeaf(context: FakeResolverContext): Any {
  return if (context.mergedField.type.rawType().name == "Date") {
    LocalDate.of(2025, Month.APRIL, 28)
  } else {
    super.resolveLeaf(context)
  }
}

// With
override fun resolveLeaf(context: FakeResolverContext): Any {
  return if (context.mergedField.type.rawType().name == "Date") {
    context.adaptToJson(LocalDate.of(2025, Month.APRIL, 28))
  } else {
    super.resolveLeaf(context)
  }
}
```


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/ir/IrOperationsBuilder.kt
```diff
@@ -601,8 +522,14 @@ internal class IrOperationsBuilder(
...
       usedCoordinates.putType(first.type.rawType().name)
 
-      // Track argument usage
+      /**
+       * Track field and arguments usage.
+       * The field usage is only useful for data builders.
+       */
       fields.map { it.parentType }.distinct().forEach { parentType ->
+        if (schema.generateDataBuilders) {
+          usedCoordinates.putField(parentType, first.name)
```
martinbonnin, line 531: This sadly needs to stay here as we need field usage information for data builders (or we'd have to generate data builders for the complete schema which sounds like a waste).

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/kotlin/helpers/data-builders.kt
```diff
@@ -1,249 +1,115 @@
...
-      .returns(mapClassName)
-      .addModifiers(KModifier.OVERRIDE)
-      .addCode(
-          CodeBlock.builder()
-              .addStatement("return %T(${Identifier.__fields})", mapClassName)
-              .build()
-      )
-      .build()
-}
+//package com.apollographql.apollo.compiler.codegen.kotlin.helpers
```
BoD, line 1: Can be removed?

martinbonnin: Definitely 👍

BoD (approved):
🎉

PR #6491, opened 2025-04-28
Changelog for v4.2.0

BoD:


CHANGELOG.md
```diff
@@ -3,8 +3,67 @@ Change Log
...
+It is now possible to configure the adapters to use with [scalars](https://www.apollographql.com/docs/kotlin/essentials/custom-scalars) with schema extensions:
+```graphql
+extend scalar Date @map(to: "kotlinx.datetime.Instant", with: "com.apollographql.adapters.InstantAdapter")
+```
+
+If a scalar can be represented as a built-in Kotlin type (String, Boolean, Int, Long, Float, Double), you can use `@mapTo`:
+```graphql
+extend scalar Length @mapTo(builtIn: Long)
+```
+This will also generate an inline class wrapper for the scalar (use `inline: false` to use the built-in type directly instead).
```
martinbonnin, line 21:
```suggestion
## Generate custom scalars as inline classes with `@map` and `@mapTo` (#6404)

It is now possible to configure the adapters to use with [scalars](https://www.apollographql.com/docs/kotlin/essentials/custom-scalars) by using the [`@map` directive](https://specs.apollo.dev/kotlin_labs/v0.5/#@map):

```graphql
extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.5/", import: ["@mapTo", "@map"])

extend scalar Date @map(to: "kotlinx.datetime.Instant", with: "com.apollographql.adapters.InstantAdapter")
```

If the `to` parameter is an [inline value class](https://kotlinlang.org/docs/inline-classes.html), use the `inlineProperty` parameter to have the generated code box/unbox the value. In that case, `with` needs to point to an adapter of the underlying type:

```graphql
extend scalar Timestamp @map(to: "com.example.Timestamp", with: "com.apollographql.apollo.api.LongAdapter", inlineProperty: "ts")
```

For the common cases where the scalar can be represented as a built-in Kotlin type (String, Boolean, Int, Long, Float, Double), you can use `@mapTo` and the compiler infers the adapter and generates an inline value class automatically:

```graphql
extend scalar Length @mapTo(builtIn: Long)
```
```

Comment (martinbonnin) [reply CHANGELOG.md]: I think I messed up the formatting and only part of the suggestion was applied?

Comment (BoD) [reply CHANGELOG.md]: Woops! Yes it looks like so. Let me fix that.

File: CHANGELOG.md [hunk 2]
```diff
@@ -3,8 +3,63 @@ Change Log
...
+
+Downloading or converting an SDL schema from introspection now includes scalar definitions.
+This is required for clients to get a [full view of the schema](https://github.com/graphql/graphql-wg/blob/main/rfcs/FullSchemas.md).
+
+## Contributors 💜
+Many thanks to @bobbysothebys, @jvanderwee, @dhritzkiv, @lwasyl and @rohandhruva for all the contributions and help in this release 💜 
+
+## 👷‍♂️ All changes
+
+* [intellij-plugin] Fix MemoryCache package name (6383)
```
martinbonnin, line 29: I think PR numbers need a `#` in front of them to become clickable in the versions page

CHANGELOG.md
```diff
@@ -3,8 +3,63 @@ Change Log
...
+```graphql
+extend schema @link(url: "https://specs.apollo.dev/kotlin_labs/v0.5/", import: ["@mapTo", "@map"])
+
+extend scalar Date @map(to: "kotlinx.datetime.Instant", with: "com.apollographql.adapters.InstantAdapter")
+
+## Scalar definitions in schemas downloaded from introspection (#6389)
+
+Downloading or converting an SDL schema from introspection now includes scalar definitions.
+This is required for clients to get a [full view of the schema](https://github.com/graphql/graphql-wg/blob/main/rfcs/FullSchemas.md).
+
```
martinbonnin, line 36:
```suggestion

## Support for `@disableErrorPropagation`

Apollo Kotlin now automatically adds [`@disableErrorPropagation`](https://github.com/graphql/nullability-wg/discussions/85) if your server supports it and you opted in `@catchByDefault` in your schema extensions.

See https://github.com/graphql/graphql-js/pull/4348 and https://github.com/graphql-java/graphql-java/pull/3772

```

CHANGELOG.md
BoD: Woops good catch 👍

martinbonnin (approved):

PR #6493, opened 2025-04-28
Adding support for linuxX64

### LINKED ISSUE #3483: Add Linux target
Apollo Android v3 doesn't currently target Linux.

ToxicBakery:
Resolves #3483

- Updated apollo-mpp-utils to leverage kotlin 2.1.20 clock implementation for epoch seconds
- Updated apollo-runtime to support linuxX64
- Added Ktor client (Curl) based http engine for linuxX64
- Added Kotlinx io okio for convenience functions for mapping kotlinx-io to okio


libraries/apollo-runtime/src/linuxMain/kotlin/com/apollographql/apollo/network/http/DefaultHttpEngine.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  private val client = HttpClient(Curl)
+
+  override suspend fun execute(request: HttpRequest): HttpResponse {
+    val response = client.request(urlString = request.url) {
+      method = when (request.method) {
+        HttpMethod.Get -> io.ktor.http.HttpMethod.Get
+        HttpMethod.Post -> io.ktor.http.HttpMethod.Post
+      }
+      setBody(request.body)
```
ToxicBakery, line 30: Pretty sure this is wrong, looking into what is required to map this.

libraries/apollo-mpp-utils/build.gradle.kts
```diff
@@ -1,10 +1,12 @@
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
+
 plugins {
   id("org.jetbrains.kotlin.multiplatform")
 }
 
 apolloLibrary(
     namespace = "com.apollographql.apollo.mpp",
-    withLinux = false
+    kotlinCompilerOptions = KotlinCompilerOptions(version = KotlinVersion.KOTLIN_2_1)
```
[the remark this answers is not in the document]
martinbonnin: Can you use posix directly? That will save users a bit of time before they need to upgrade to K2 and will make this easier to backport to the v4 branch if needed

```kotlin
  memScoped {
    val timespec = alloc<timespec>()
    clock_gettime(CLOCK_REALTIME, timespec.ptr)
    return timespec.tv_sec * 1000 + (timespec.tv_nsec + 500_000) / 1_000_000
  }
```

libraries/apollo-testing-support-internal/src/linuxMain/kotlin/com/apollographql/apollo/testing/currentThreadId.linux.kt
```diff
@@ -0,0 +1,5 @@
+package com.apollographql.apollo.testing
+
+actual fun currentThreadId(): String {
+  return "linux"
```
martinbonnin, line 4:
```suggestion
  return pthread_self().toString()

```

libraries/apollo-runtime/build.gradle.kts
```diff
@@ -89,6 +88,13 @@ kotlin {
       }
     }
 
+    findByName("linuxMain")?.apply {
+      dependencies {
+        implementation(libs.kotlinx.io.okio)
+        implementation(libs.ktor.client.curl)
```
martinbonnin, line 94: We usually avoid ktor in the main artifacts (ktor support is provided in a separate artifact [there](https://github.com/apollographql/apollo-kotlin-ktor-support)). But as long as it doesn't leak to other platforms and it's allowing to unblock linux, LGTM

libraries/apollo-mpp-utils/build.gradle.kts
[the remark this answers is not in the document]
ToxicBakery: Makes sense, will do

libraries/apollo-runtime/build.gradle.kts
ToxicBakery: Understood. This was a question to @BoD at the start of this in the issue thread if ktor would be a reasonable solution for the platform now. If there is another alternative I'm open to doing some research to evaluate it.

martinbonnin: Sorry I forgot the initial context. The alternative is calling the [libcurl C API](https://curl.se/libcurl/c/) but this is probably going to be more involved so agree with @BoD that ktor is a reasonable solution for now.

libraries/apollo-testing-support-internal/src/linuxMain/kotlin/com/apollographql/apollo/testing/currentThreadId.linux.kt
martinbonnin: Right, JS is always single threaded so the thread id is always the same there but it's not the case on linux.

libraries/apollo-runtime/src/linuxMain/kotlin/com/apollographql/apollo/network/http/DefaultHttpEngine.kt
BoD: Maybe what's in the Ktor adapter [here](https://github.com/apollographql/apollo-kotlin-ktor-support/blob/main/apollo-engine-ktor/src/commonMain/kotlin/com/apollographql/ktor/http/KtorHttpEngine.kt) can be copy/pasted. This would also leave out `kotlinx.io.okio` (not really an issue, just a tendency to avoid a dependency in general).

martinbonnin (approved):
One mini comment, looks good to me otherwise, thanks for diving into this!

ToxicBakery:
@martinbonnin Thank you, is there anything I need to do for the new failing tests? The Linux runtime does not have websocket support in this implementation so those tests are failing. I didn't see an obvious way to prevent/skip those tests.

martinbonnin:
@ToxicBakery I think it's the same situation for wasm. If you calll `.gradlew build` some tests are failing but the CI calls [`ciTestsNoGradle`](https://github.com/apollographql/apollo-kotlin/blob/a3c11be76a79551394ee25de653bff01c60720bd/build.gradle.kts#L66) that excludes a bunch of tests, including the Wasm and now the linux ones.

It's not perfect but it's been working so far!

PR #6494, opened 2025-04-29
Remove PackageNameGenerator and OperationOutputGenerator

martinbonnin:
`PackageNameGenerator` and `OperationOutputGenerator` required running in the main Gradle buildscript and duplicated a lot of codepaths in the Gradle plugin.

Moving forward:

* `PackageNameGenerator` can be replaced with `ApolloCompilerPlugin.layout()` (but probably shouldn't, I would recomment using flat package names for simplicity. If you require custom package names, please let us know)
* `OperationOutputGenerator` can be replaced with `ApolloCompilerPlugin.pluginIds()`


BoD (approved):
🧹

PR #6496, opened 2025-04-29
Update deprecations for v5

martinbonnin:
* `DeprecationLevel.WARNING` are turned into `DeprecationLevel.ERROR`
  * Except for a couple perpetual soft deprecations such as `SubscriptionWsProtocol`
* `DeprecationLevel.ERROR` are removed
* Galaxy and RxJava support artifacts are removed


libraries/apollo-gradle-plugin-external/src/main/kotlin/com/apollographql/apollo/gradle/internal/ModelNames.kt
```diff
@@ -30,11 +30,6 @@ internal object ModelNames {
   fun checkApolloVersions() = "checkApolloVersions"
   fun convertApolloSchema() = "convertApolloSchema"
 
-  // Configuration names
-  @Deprecated("Unused. Use dependsOn() instead.")
-  @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
-  fun metadataConfiguration() = "apolloMetadata"
```
martinbonnin, line 36: This one was mostly an internal API. I'm not going through the `ERROR` level

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloCall.kt
```diff
@@ -147,34 +147,6 @@ class ApolloCall<D : Operation.Data> internal constructor(
...
 
-  /**
-   * A version of [execute] that restores 3.x behaviour:
-   * - throw on fetch errors.
-   * - make `CacheFirst`, `NetworkFirst` and `CacheAndNetwork` policies ignore fetch errors.
-   * - throw ApolloComposite exception if needed.
-   */
-  @Deprecated("Use toFlow() and handle ApolloResponse.exception instead")
-  @ApolloDeprecatedSince(ApolloDeprecatedSince.Version.v4_0_0)
-  fun toFlowV3(): Flow<ApolloResponse<D>> {
```
martinbonnin, line 158: The V3 helpers are also not going through `ERROR` because I think it's relatively clear that they were a temporary solution,

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/HttpEngine.kt
```diff
@@ -27,18 +25,12 @@ interface HttpEngine : Closeable {
   @Throws(ApolloNetworkException::class, CancellationException::class)
   suspend fun execute(request: HttpRequest): HttpResponse
 
-  @ApolloDeprecatedSince(v4_0_0)
-  @Deprecated("Use close", ReplaceWith("close()"), level = DeprecationLevel.ERROR)
-  fun dispose() {
-  }
```
martinbonnin, line 33: This one is an interesting one because even though is is marked as `ERROR`, nothing prevents users to actually implement the `dispose()` function. All they would see is a warning that the override is not deprecated while the base method is.

Not sure if there is a good solution and I don't really want to keep this forever...

libraries/apollo-annotations/src/commonMain/kotlin/com/apollographql/apollo/annotations/ApolloDeprecatedSince.kt
```diff
@@ -33,7 +33,7 @@ annotation class ApolloDeprecatedSince(val version: Version) {
     v4_0_0,
     v4_0_1,
     v4_0_2,
-    v4_1_2,
+    v4_1_2, // All symbols above are ERRORs in 5.0.0 except a few ones that are perpetual soft deprecations like SubscriptionWsProtocol (https://youtrack.jetbrains.com/issue/KT-54106)
```
BoD, line 36: 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/HttpEngine.kt
BoD: Yeah it would be nice if implementing a fun deprecated with `ERROR` level was an error?

BoD (approved):
🧹

PR #6498, opened 2025-04-29
[gradle] Do not walk the file system to build the project graph

martinbonnin:
This is slow and invalidates the configuration cache all the time.

### DIFF
--- a/tests/settings.gradle.kts
+++ b/tests/settings.gradle.kts
@@ -10,21 +10,85 @@ plugins {
 
 rootProject.name = "apollo-tests"
 
-// Include all tests
-rootProject.projectDir
-    .listFiles()!!
-    .filter { it.isDirectory }
-    .flatMap {
-      it.walk()
-    }
-    .filter { it.isDirectory && File(it, "build.gradle.kts").exists() }
-    .forEach {
-      // Do no create intermediate projects as they will fail for apolloTestAgreggation
-      // See https://stackoverflow.com/questions/21015353/gradle-intermediate-dir-of-multiproject-not-subproject
-      val project = it.relativeTo(rootProject.projectDir).path.replace(File.separatorChar, '-')
-      include(project)
-      project(":$project").projectDir = it
-    }
+listOf(
+    "ast-benchmark",
+    "browser-tests",
+    "cache-variables-arguments",
+    "catch",
+    "compiler-plugins/add-field",
+    "compiler-plugins/app",
+    "compiler-plugins/capitalize-enum-values",
+    "compiler-plugins/custom-flatten",
+    "compiler-plugins/default-null-values",
+    "compiler-plugins/getters-and-setters",
+    "compiler-plugins/prefix-names",
+    "compiler-plugins/schema-codegen",
+    "compiler-plugins/typename-interface",
+    "data-builders-java",
+    "data-builders-kotlin",
+    "defer",
+    "deprecated-requires-opt-in",
+    "enums",
+    "escaping",
+    "filesystem-sensitivity",
+    "generated-methods",
+    "gzip",
+    "http-cache",
+    "http-headers",
+    "include-skip-operation-based",
+    "input",
+    "integration-tests",
+    "intellij-plugin-test-project",
+    "ios-test",
+    "java-nullability",
+    "js",
+    "jsexport",
+    "jvmoverloads",
+    "kdoc",
+    "kotlin-codegen",
+    "model-builders-java",
+    "models-operation-based",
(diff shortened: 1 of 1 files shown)


BoD (approved):
👍

martinbonnin:
> excluding some directories like src, build, etc.

That's probably the most important thing. Walking `build` in our case means that the contents where very likely to change from one build to the other and all `build` becomes an input for the configuration.

I considered doing the same thing you do but ultimately, listing everything is a one time thing vs we have projects in subfolders so maintaining the filesystem scanning logic would also have _some_  cost. We are [still scanning the filesystem for the main build](https://github.com/apollographql/apollo-kotlin/blob/main/settings.gradle.kts#L18) and it's working fine so both do work at the end of the day (just don't scan `build`!)

SimonMarquis:
> That's probably the most important thing. Walking build in our case means that the contents where very likely to change from one build to the other and all build becomes an input for the configuration

But why would `build` be part of the configuration cache input? It is not "walked through" with Gradle APIs from what I can read in your code. The only issue is the fact that indeed, build directories are huge in size and in number of files, so it takes time to crawl them, especially when there is nothing to use inside :)

> We are [still scanning the filesystem for the main build](https://github.com/apollographql/apollo-kotlin/blob/main/settings.gradle.kts#L18)

👍 By the way, it is only a matter of returning `SKIP_SUBTREE` from `preVisitDirectory`, and you will not crawl unnecessary file trees.

SimonMarquis:
Nevermind, your code in the main `settings.gradle.kts` is only listing files in `/libraries` so it is probably not taking more than a few ms.

SimonMarquis:
But in reality it probably is not. Nothing in the configuration cache reports mentions it. But most importantly, the configuration cache is always reused (as written in the logs on each gradle invocation: `Reusing configuration cache.`) even when modifying files that were queried. 
I think Gradle only add files that are either directly used as inputs/outputs of tasks, and maybe if they are used through a "Gradle" type like `Project.projectDirectory`, `ProjectLayout`, etc.

SimonMarquis:
Our implementation was slightly different, and relies on `java.nio.file.FileVisitor`.
You can try the same scenario as above, and you'll get a configuration cache hit every time.

- https://github.com/martinbonnin/test-cc/pull/1

Not sure if this is an intended behavior or not 🤔

SimonMarquis:
One issue that I can spot with our implementation (although it never happens in practice) is that if we were actually correctly using the configuration cache, and we added a new gradle module, without referencing it in any other module (this is what is very unlikely), Gradle would never detect it as it would reuse the previous configuration cache entry.

martinbonnin:
I believe this is a bug. 

Gradle has no way to know what you're going to do with the filesystem information. If you're not using it to alter the task graph then there is no bug but it still feels conceptually wrong. I filed a bug there: https://github.com/gradle/gradle/issues/33317

SimonMarquis:
Noooo 😨 I don't want to go back to a 1000+ lines long file to list our modules 🙈 (I don't think there are any alternatives)
Anyways, thanks a lot for taking the time to look at it and opening the bug report 🙏

martinbonnin:
I'd say you can probably use `File.listFiles()` and implement the `walk()` yourself, [`listFile`](https://github.com/gradle/gradle/blob/0ce655bf466a99edf278d6ffa94e1fc56236a925/subprojects/core/src/main/java/org/gradle/internal/classpath/declarations/FileInterceptorsDeclaration.java#L38) is instrumented so it will track the contents of the directories correctly I think...

SimonMarquis:
Yeah but I just want to find Gradle projects to "include", and not track the content of what I'm walking through. I'm not sure to understand how `listFiles` or implementing a custom `walk()` would change the fact that it is either a bug or working as intended.

martinbonnin:
My suspicion is that it's actually 2 different problems:

* Kotlin `walk()` uses `listFiles()` under the hood but that bytecode is probably not instrumented because it is loaded in a very early classloader that is not in an instance of `InstrumentingClassLoader`. This is a wild guess but that wouldn't be completely unexpected given how `kotlin-stdlib` is a specal dependency.
* Java `walkFileTree()` is just not implemented at all from what I can see in the code.

If you call `listFiles()` manually, instrumentation should kick in and monitor the content of the directories (not the files) that you call `listFiles()` on.

Again, I'm not 100% sure so let's wait to see what the Gradle team says but if you're trying the manual `listFiles()` approach, I'm 99% sure it will invalidate your CC if you remove a file in one of the walked directories.

PR #6503, opened 2025-04-30
Add key fields to selections even when they're already selected with an alias

BoD:


martinbonnin (commented):
Could we avoid duplicating the field? I think we have everything for that in `CacheKeyGeneratorContext.field.selections`? 

Fragments make it a bit more complicated (because we don't have the full schema at runtime so a given field may be key in one fragment but not in the other and we probaby wouldn't know which one...).

martinbonnin:
Another option is to codegen the key responseNames in `CompiledField`... We could make it so that we only codegen them if there are aliases but it's still one extra pointer for every field.

BoD:
> Could we avoid duplicating the field? I think we have everything for that in CacheKeyGeneratorContext.field.selections?

Ah yes we could look for the field's name, and fallback to its alias(es).

But I actually don't follow your caveat about fragments, do you have an example?

PR #6507, opened 2025-05-02
Remove defer Helix tests

BoD:
These tests have been `@Ignored` for a while.

Recently we've seen sporadic build fails about using `require()` around these tests. It may be related to #6466 but I don't really understand the root cause, and it's not 100% reproducible.

I've briefly tried to resurrect those tests and use `graphql-js` `17.0.0-alpha.8`, but that failed with `Ensure that there is only one instance of \\\"graphql\\\" in the node_modules\\ndirectory`... Given that the latest release of Helix is from 2022, maybe I shouldn't expect it to work with a version that recent.

We still have integration tests with the router so I don't feel too bad giving up on these.


martinbonnin (approved):
🧹

PR #6514, opened 2025-05-06
Remove error message when trying to register a compiler plugin directly

martinbonnin:
It all needs to go through providers now


BoD (approved):
🧹

PR #6515, opened 2025-05-07
Add key fields of possible types of interfaces and fragments

BoD:
See https://github.com/apollographql/apollo-kotlin-normalized-cache/issues/141


libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/internal/ApolloDocumentTransform.kt
```diff
@@ -100,20 +122,45 @@ private fun List<GQLSelection>.addRequiredFields(
 
   val fieldNames = parentFields + selectionSet.filterIsInstance<GQLField>().map { it.responseName() }.toSet()
 
+  // Interfaces and unions: add key fields of all possible types in inline fragments
+  val parentTypeDefinition = schema.typeDefinition(parentType)
+  val possibleTypes = if (parentTypeDefinition is GQLInterfaceTypeDefinition || parentTypeDefinition is GQLUnionTypeDefinition) {
+    schema.possibleTypes(parentTypeDefinition)
+  } else {
+    emptySet()
+  } - parentAddedFragmentTypeConditions
```
martinbonnin, line 131: Looks like `parentAddedFragmentTypeConditions` is to add some more fragments in nested fragments? 
I'm struggling to find a use case:

```graphql
{
  mammal {
    ... on Animal {
      // do we ever introduce new fragments here?
    }
  }
}

```

BoD: It's to avoid adding fragments when they are already added in the parent.

```graphql
{
  animal {
    ... on Mammal {
      # no need to add key fields fragments here because they're already added above
    }

  # key fields fragments added here
  }
}

```

martinbonnin: Would `isRoot` work? 

No need to track a set of type conditions. The fragments should always be added directly under the field (or fragment definition). Every nested fragment is only restricting the possible types so it's not going to require new synthetic fragments

BoD: Totally! `isRoot` is much simpler 👍

martinbonnin (approved):

PR #6522, opened 2025-05-10
Add CacheAndNetwork fetch policy to HttpCache

### LINKED ISSUE #4235: Support `CacheAndNetwork` Option in the `HttpCache`
**Use case**
To replicate the convenient functionality of using the normalized cache, the `HttpCache` should _also_ support this option. Currently, it only supports `CacheFirst`, or `NetworkFirst`, which means we only ever receive one response, if its in the cache.

Currently, what this means, is if, we use `HttpCache` without a timeout specified, the cache never expires. It never hits the network again after it receives the initial response, unless I set an explicit timeout. I'd certainly like to cache the previous output and still reach out for the latest - this way the UI can reactively update if the data changes in any meaningful way. 

A workaround is to send two queries - 1 for initial cache, then a second one with cache disabled on that request.

Ideally, using `watch()` on our queries, we could receive multiple results. 1 for cache (if it exists), then 1 for the network (if we have a network connection).

Add the `CacheAndNetwork` option to the `HttpFetchPolicy`. Optionally, merge the two enums so they are both equally supported in each cache type.

2307vivek:
Added support for `CacheAndNetwork` fetch policy to `HttpCache`, replicating the behavior available in the normalized cache. Using this fetch policy will always emit two values, initially the cached http response (if available), and then the network response when online. 
In case of a cache miss or a network error a non-null `.exception` will be emitted for both the cases.

Closes #4235.

### DIFF
--- a/docs/source/caching/http-cache.mdx
+++ b/docs/source/caching/http-cache.mdx
@@ -64,6 +64,9 @@ val response = apolloClient.query(query)
 
   // Don't use the cache
   .httpFetchPolicy(HttpFetchPolicy.NetworkOnly)
+
+  // Try the cache, then also try the network
+  .httpFetchPolicy(HttpFetchPolicy.CacheAndNetwork)
   
   // Finally, execute your query
   .execute()
@@ -71,6 +74,8 @@ val response = apolloClient.query(query)
 Note: mutations and subscriptions don't go through the cache.
 
 If the query is present in cache, it will be used to return `response.data`. If not, a `HttpCacheMissException` will be thrown.
+
+`CacheAndNetwork` will always emit two values, you should use `.toFlow()` when using that policy.
 
 You can also set an expiration time either globally or for specific queries. The entries will automatically be removed from the cache after the expiration time:
 
--- a/libraries/apollo-http-cache/api/apollo-http-cache.api
+++ b/libraries/apollo-http-cache/api/apollo-http-cache.api
@@ -5,6 +5,11 @@ public abstract interface class com/apollographql/apollo/cache/http/ApolloHttpCa
 	public abstract fun write (Lcom/apollographql/apollo/api/http/HttpResponse;Ljava/lang/String;)Lcom/apollographql/apollo/api/http/HttpResponse;
 }
 
+public final class com/apollographql/apollo/cache/http/CacheAndNetworkApolloInterceptor : com/apollographql/apollo/interceptor/ApolloInterceptor {
+	public fun <init> ()V
+	public fun intercept (Lcom/apollographql/apollo/api/ApolloRequest;Lcom/apollographql/apollo/interceptor/ApolloInterceptorChain;)Lkotlinx/coroutines/flow/Flow;
+}
+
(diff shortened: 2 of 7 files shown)


BoD:
Thanks a lot for looking into this @2307vivek!

We’re actually now wondering if this is worth a new API given that a simple extension can be used, e.g.:

```kotlin
fun <D : Query.Data> ApolloClient.queryCacheAndNetwork(query: Query<D>): Flow<ApolloResponse<D>> {
  return flow {
    emitAll(query(query).httpFetchPolicy(HttpFetchPolicy.CacheOnly).toFlow())
    emitAll(query(query).httpFetchPolicy(HttpFetchPolicy.NetworkOnly).toFlow())
  }
}
```

Another reticence is that using `CacheAndNetwork` will crash if used with `.execute()` which can be surprising - we have the same issue with the Normalized Cache version, which we regret nowadays.

Moreover, we’ve are actually [wondering](https://github.com/apollographql/apollo-kotlin/issues/6302) about the future of the HTTP cache itself.

Does that make sense? I know the initial issue was marked with “PR welcome”, apologies about that!

2307vivek:
Thanks for replying @BoD!

I though about the extension function way, but decided to make an proper `enum `with documentation to enforce consistency with other fetch policies, moreover a library user expects a Http `CacheAndNetwork `policy given the fact that it is a useful fetch policy and the normalized cache already has one.

> `CacheAndNetwork ` will crash if used with `.execute()` which can be surprising.

I don't understand why is it a problem, isn't it an expected behavior? The docs clearly says that `.execute()` will return a single response, and it will throw if more than one success response is returned. The docs also says to use `.toFlow()` with `CachAndNetwork`. Am I missing something?

I am not sure about the future of Http cache, it's true that most of the things that it does can be done with persistent cache, let the community decide. My take is that the Http cache is a simplest way to implement client side cache, especially given the fact that normalized cache has some cpu overhead normalizing and de-normalizing large responses.

Let me know your thoughts, and I am also OK with closing this pr if a decision is made otherwise.

BoD:
>I don't understand why is it a problem, isn't it an expected behavior? The docs clearly says that .execute() will return a single response, and it will throw if more than one success response is returned. The docs also says to use .toFlow() with CachAndNetwork. Am I missing something?

No that’s exactly right! It’s just that one of the enum values doesn’t behave like the others -  not a huge deal of course, just a small API inconsistency that we’re not fans of in hindsight.

>I am not sure about the future of Http cache, it's true that most of the things that it does can be done with persistent cache, let the community decide. My take is that the Http cache is a simplest way to implement client side cache, especially given the fact that normalized cache has some cpu overhead normalizing and de-normalizing large responses.

Thanks for the feedback! Agreed about an HTTP cache being a simple solution - it’s more about the relevance of the Apollo implementation, given that there is already one built-in in OkHttp. GraphQL uses POST by default, which makes the responses not cacheable - but this changes when using persisted queries (which we recommend), making the Apollo one kind of redundant.


All in all, this PR improves the consistency of the 2 caches, but we have doubts about this API, and the functionality can easily be addressed with an extension.

2307vivek:
Alright, it makes sense now. I am closing this pr.

BoD:
Thanks again for looking into this!

PR #6524, opened 2025-05-13
Switch the gradle plugin to gratatouille

martinbonnin:
Opoening as draft as it's still missing:

* Classloader cache
* A version of gratatouille on Maven Central
* Migration guide

The main change is that the generated folder location changed and that `com.apollographql.apollo.compiler.*` is not available anymore from the build script classpath anymore.


intellij-plugin/build.gradle.kts
```diff
@@ -171,11 +171,7 @@ dependencies {
     zipSigner()
   }
 
-  // Coroutines must be excluded to avoid a conflict with the version bundled with the IDE
-  // See https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#coroutinesLibraries
-  implementation(project(":apollo-gradle-plugin-external")) {
-    exclude(group = "org.jetbrains.kotlinx", module = "kotlinx-coroutines-core")
-  }
+  implementation(project(":apollo-gradle-plugin"))
```
BoD, line 174: No longer needed because it's relocated?

libraries/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/internal/ApolloBuildService.kt
```diff
@@ -0,0 +1,8 @@
+package com.apollographql.apollo.gradle.internal
+
+import org.gradle.api.file.FileCollection
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+import java.net.URI
+import java.net.URLClassLoader
```
BoD, line 7: File can be removed?

BoD (approved):
🚀  🐀  🧑‍🍳

intellij-plugin/build.gradle.kts
martinbonnin: The exclusion is not needed anymore because `apollo-gradle-plugin` doesn't depends on `kotlinx-coroutines` anymore. 

`apollo-gradle-plugin-tasks` does but it's a completely separate tree.

BoD: Ah right! Cool 👍

PR #6528, opened 2025-05-16
fix the batch size not respected issue

mengdd:
Fix the issue:
https://github.com/apollographql/apollo-kotlin/issues/6527

Fix: If the batch size is reaching limit, send the requests immediately in the current `withLock` block.


BoD:
Thanks a lot for the detailed bug report, and fix!

I've just added a test that fails on the main branch and works here.

BoD (approved):

tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt
```diff
@@ -90,7 +99,7 @@ class QueryBatchingTest {
 
     // Only one request must have been sent
     assertFails {
-      mockServer.awaitRequest()
+      mockServer.awaitRequest(3.seconds)
```
BoD, line 102: (Not related, but these tests took 30s each time)

PR #6537, opened 2025-05-26
feature: Enhanced Client Awareness (v5)

calvincestari:
This adds support for the Enhanced Client Awareness request extensions; similar to https://github.com/apollographql/apollo-ios-dev/pull/638 in Apollo iOS.

**Note:** This PR is against `main` which makes it part of a future v5 release.


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -50,19 +50,20 @@ class DefaultHttpRequestComposer(
     }
 
     val sendApqExtensions = apolloRequest.sendApqExtensions ?: false
+    val sendEnhancedClientAwarenessExtensions = true
```
calvincestari, line 53: This obviously needs to get it from somewhere; I _think_ we'd discussed setting the value at client creation?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -147,22 +149,33 @@ class DefaultHttpRequestComposer(
...
         query: String?,
     ): Map<String, Upload> {
       return composePostParams(
-          writer, operation, customScalarAdapters, query, apqExtensionsWriter(operation.id(), sendApqExtensions)
+          writer, operation, customScalarAdapters, query, extensionsWriter(operation.id(), sendApqExtensions, sendEnhancedClientAwarenessExtensions)
       )
     }
 
-    private fun apqExtensionsWriter(id: String, sendApqExtensions: Boolean): JsonWriter.() -> Unit {
+    private fun extensionsWriter(apqId: String, sendApqExtensions: Boolean, sendEnhancedClientAwarenessExtensions: Boolean): JsonWriter.() -> Unit {
```
calvincestari, line 160: Is there a preferred formatting style to stay within a column limit for these long function definitions?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -283,10 +314,11 @@ class DefaultHttpRequestComposer(
         customScalarAdapters: CustomScalarAdapters,
         autoPersistQueries: Boolean,
         sendDocument: Boolean,
+        sendEnhancedClientAwarenessExtensions: Boolean,
```
calvincestari, line 317: This appears to build without error but with this function being public (I think that's the default visibility when not specified?), is it going to break user code without a default on the param; do users typically call this function directly?

calvincestari (commented):
@martinbonnin, @BoD - is this PR on the right track so far?

calvincestari:
I'm not sure what the `:apollo-api:jvmApiCheck` is that's failing. We'll have to talk tomorrow..

calvincestari:
I cancelled the `tests-integration` check, it looked like it was taking far too long.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -147,22 +149,33 @@ class DefaultHttpRequestComposer(
...
+    private fun extensionsWriter(apqId: String, sendApqExtensions: Boolean, sendEnhancedClientAwarenessExtensions: Boolean): JsonWriter.() -> Unit {
+      if (!sendApqExtensions && !sendEnhancedClientAwarenessExtensions) return { }
+
       return {
-        if (sendApqExtensions) {
-          name("extensions")
-          writeObject {
+        name("extensions")
+        writeObject {
+          if (sendApqExtensions) {
```
martinbonnin, line 166: Can you make it so that we don't send an empty object if there are neither apqs nor client awareness?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -147,22 +149,33 @@ class DefaultHttpRequestComposer(
...
-              name("sha256Hash").value(id)
+              name("sha256Hash").value(apqId)
+            }
+          }
+
+          if (sendEnhancedClientAwarenessExtensions) {
+            name("clientLibrary")
+            writeObject {
+              name("name").value("apollo-kotlin")
+              name("version").value(0)
```
martinbonnin, line 178: Sounds like we need to send a version here?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -210,8 +224,24 @@ class DefaultHttpRequestComposer(
             }
           }
         }
+
+        if (sendEnhancedClientAwareness) {
+          val extensions = buildJsonString {
```
BoD, line 229: I think it should just be `writeObject {` here?

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
martinbonnin: This will most likely break the ABI compatibility indeed. How default parameter values work is that they generate a synthetic function that takes an extra int parameter, that extra int being a bitmask of what values are explicitely passed. See https://medium.com/@yi.lu_55329/is-kotlin-default-arguments-a-solution-of-binary-backward-compatibility-374d2c9214d4 for a more in depth explanation.

You can see the binary compatibility validator failed [here](https://github.com/apollographql/apollo-kotlin/actions/runs/15263435025/job/42924993454?pr=6537#step:5:308).

I recommend to make it an extra new method and deprecate the old one. This is a bit painful but it's the only way to preserve backward compatibility.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
martinbonnin: No preferrence from my side. We have the formatting rules in source control [here](https://github.com/apollographql/apollo-kotlin/tree/main/.idea/codeStyles). If you "reformat all" (code -> reformat file...) and it stays like this, I'm down with this.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
martinbonnin: Right. This is both one of the most boring and one of the most complicated part of the codebase. 

I would recommend adding a property to `ApolloClient.Builder` like `failFastIfOnline` [here](https://github.com/apollographql/apollo-kotlin/blob/4b6dbc6750672faab897656a3651a655b99038fe/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt#L391). The other option is to be able to set it individually per request in `MutableExecutionOptions` but I don't think it's worth the added complexity.

martinbonnin (commented):
Nice 🚀

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
calvincestari: Avoiding an empty object should be handled by the `if` above no? That's an early exit for neither of the extensions being set and by here we know we're sending an object so these `if` statements are to only include the sub-objects that we need.

tests/integration-tests/src/commonTest/kotlin/test/BodyExtensionsTest.kt
```diff
@@ -71,4 +71,60 @@ class BodyExtensionsTest {
...
+    mockServer.close()
+  }
+
+  @Suppress("UNCHECKED_CAST")
+  @Test
+  fun enhancedClientAwarenessExtensionsExcludedWhenDisabled() = runTest {
+    val mockServer = MockServer()
+
+    val apolloClient = ApolloClient.Builder()
+        .sendEnhancedClientAwareness(false)
```
calvincestari, line 109: @martinbonnin / @BoD  - this does not appear to be working as intended but I'm unsure if I'm just not constructing things correctly.

calvincestari:
@martinbonnin / @BoD - I think this is now ready for another review.

tests/integration-tests/src/commonTest/kotlin/test/BodyExtensionsTest.kt
BoD: I think it's because the value needs to be passed over from the client to the request inside `ApolloClient.apolloResponses()`

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -147,22 +150,37 @@ class DefaultHttpRequestComposer(
...
             name("persistedQuery")
             writeObject {
               name("version").value(1)
-              name("sha256Hash").value(id)
+              name("sha256Hash").value(apqId)
+            }
+          }
+
+          if (sendEnhancedClientAwarenessExtensions) {
+            name("clientLibrary")
```
martinbonnin, line 180: Do we have a reference that we could link as a comment here for future readers that would like more context for the feature?

tests/integration-tests/src/commonTest/kotlin/test/HttpRequestComposerTest.kt
```diff
@@ -35,6 +35,13 @@ class HttpRequestComposerTest {
     checkTestFixture(bodyText, "allPlanets.json")
   }
 
+  @Test
+  fun requestBuilderSetsEnhancedClientAwarenessExtensionsTrueByDefault() {
```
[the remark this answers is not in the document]
martinbonnin: I'm fine leaving it out.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -210,8 +229,22 @@ class DefaultHttpRequestComposer(
...
+        if (sendEnhancedClientAwareness) {
+          writeObject {
+            name("clientLibrary")
+            writeObject {
+              name("name").value("apollo-kotlin")
+              name("version").value(0)
+            }
+          }
+        }
+      }
```
martinbonnin, line 242: You'll need to put all the extension in a single object here:

```kotlin
      val extensions = buildJsonString {
        writeObject {
          if (autoPersistQueries) {
            name("persistedQuery")
            writeObject {
              name("version").value(1)
              name("sha256Hash").value(operation.id())
            }
          }
          if (sendEnhancedClientAwareness) {
            name("clientLibrary")
            writeObject {
              name("name").value("apollo-kotlin")
              name("version").value(0)
            }
          }
        }
      }
```

calvincestari:
OK, I believe this is now all done. 🎉 

I think the only thing left to discuss about it is whether we want v5 or v4 + v5?

calvincestari:
> I've noticed that the tests-integration step appears stuck. I looked at the log and could not see any error so I'm not sure what the delay was - is this something that you've seen before?

OK, I see what the issue is; there are some test failures but the logs do not make that obvious. Looks like the test fixtures need to be updated or this feature disabled in those tests. Here's an example:
```
test.LoggingInterceptorTest.levelBody[macosArm64] FAILED
    kotlin.AssertionError: Expected <post http://0.0.0.0/
    accept: multipart/mixed;deferspec=20220824, application/graphql-response+json, application/json
    [end of headers]
    {"operationname":"heroname","variables":{},"query":"query heroname { hero { name } }"}

    >, actual <post http://0.0.0.0/
    accept: multipart/mixed;deferspec=20220824, application/graphql-response+json, application/json
    [end of headers]
    {"operationname":"heroname","variables":{},"query":"query heroname { hero { name } }","extensions":{"clientlibrary":{"name":"apollo-kotlin","version":"5.0.0-snapshot"}}}
```

calvincestari:
@martinbonnin - I think in this case it might be better to turn off the feature because of the version string; the tests will all need to be updated each time a new version is released.

calvincestari:
Oh wait..that was your commit updating them.

martinbonnin:
I'd like to avoid those tests doing too much configuration. The drawback is that now the expected string is computed based on the current version instead of before it was a plain string literal but I think this is OKay.

calvincestari:
Thanks for helping fix it. ❤️

tests/integration-tests/testFixtures/expectedOperationsPartBodyMultiple.json
```diff
@@ -1 +1 @@
-{"operationName":"MultipleUpload","variables":{"files":[null,null]},"query":"mutation MultipleUpload($files: [Upload!]!) { multipleUpload(files: $files) { id path filename mimetype } }"}
\ No newline at end of file
+{"operationName":"MultipleUpload","variables":{"files":[null,null]},"query":"mutation MultipleUpload($files: [Upload!]!) { multipleUpload(files: $files) { id path filename mimetype } }","extensions":{"clientLibrary":{"name":"apollo-kotlin","version":"5.0.0-SNAPSHOT"}}}
```
BoD, line 1: I updated the `release.main.kts` script to take care of this

BoD (approved):
🙏

martinbonnin (approved):

scripts/release.main.kts
```diff
@@ -243,6 +244,18 @@ fun setVersionInIntelliJPlugin(version: String) {
   }
 }
 
+fun setVersionInFixtures(nextSnapshot: String) {
```
calvincestari, line 247: 🎉

PR #6538, opened 2025-05-27
Fix losing response headers when using batch request

gnehz972:
While using the [new Normalized Cache](https://github.com/apollographql/apollo-kotlin/issues/6533) in the [Server-controlled way](https://apollographql.github.io/apollo-kotlin-normalized-cache/cache-control.html#server-controlled), the batched requests are unable to follow the `cache-control` header from the server side and always return the cached data.

This is caused by losing response headers when splitting the batched response

Add the headers to the split response to fix this.


BoD (approved):
Thanks!

PR #6539, opened 2025-05-27
Apollo API: generate the version

martinbonnin:
Adds `com.apollographql.apollo.api.apolloApiVersion` that can be used for client awareness.


BoD (approved):
👍

PR #6541, opened 2025-05-30
Skip required fields if compiler plugin is present

martinbonnin:
The cache compiler plugin is going to add them anyways


martinbonnin:
Differences between main and 4.x:

* `main` has the separated data builders
* `main` has `SchemaDocumentTransform`
* `main` removes `OperationOutputGenerator` and uses `OperationIdsGenerator` instead
* `main` removes all `ApolloCompilerPlugin` methods except for `beforeCompilationStep` and `operationIds`

BoD (approved):
⚙️

PR #6542, opened 2025-05-30
Add more documentation to `@typePolicy`

martinbonnin:
Will do the matching change in https://specs.apollo.dev/kotlin_labs/v0.5/#@typePolicy once we agree on the wording.

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/definitions.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/definitions.kt
@@ -35,7 +35,41 @@ internal val kotlinLabsDefinitions_0_3 = """
   ""${'"'}
   directive @typePolicy(
       ""${'"'}
-      a selection set containing fields used to compute the cache key of an object. Order is important.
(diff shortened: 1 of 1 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/definitions.kt
```diff
@@ -35,7 +35,41 @@ internal val kotlinLabsDefinitions_0_3 = """
   ""${'"'}
   directive @typePolicy(
       ""${'"'}
-      a selection set containing fields used to compute the cache key of an object. Order is important.
+      A selection set containing fields used to compute the cache key of an object. 
+      Nested selection sets are currently not supported. Order is important. 
+      
+      Key fields can be defined on interfaces. In that case, the key fields apply to all sub-types. Sub-types are not allowed to define their own key fields.
```
BoD, line 41: Nit
```suggestion
      Key fields can be defined on interfaces. In that case, the key fields apply to all sub-types and sub-types are not allowed to define their own key fields.
```

BoD (approved):
👍

PR #6544, opened 2025-05-30
[4.x] Add key fields to selections even when they're already selected with an alias (#6503)

martinbonnin:
Backport of https://github.com/apollographql/apollo-kotlin/pull/6503


BoD (approved):
👍

PR #6547, opened 2025-05-30
Keep compiler plugin in sync with v4

martinbonnin:
Keep compiler plugins in sync with https://github.com/apollographql/apollo-kotlin/pull/6546

* `OperationsTransform` -> `ExecutableDocumentTransform`
* Made `beforeCompilationStep()` default for legacy users
* Only pass `outputDirectory` in `CodgeGenerator`
* Fix sorting of transforms


BoD (approved):
👍

PR #6551, opened 2025-06-04
[4.x] Add back a test + tweak a function name + tweak API stability

martinbonnin:
4.x still has the `checkKeyFields()` function so keep the test around for now


BoD (approved):
👍

PR #6562, opened 2025-06-11
[gradle-plugin] Simplify task wiring

martinbonnin:
`registerJavaGeneratingTask()` works for Kotlin sources as well, no need to distinguish the cases. Fixes `connectToAndroidSourceSet` that broke with https://github.com/apollographql/apollo-kotlin/pull/6442.

As a workaround until this is release, you can set up explicit dependencies:

```kotlin
apollo {
    service("service") {
        packageName.set("com.example")
        outputDirConnection {
            tasks.withType(KotlinCompile::class.java).configureEach {
                dependsOn(task)
            }
            connectToAndroidSourceSet("main")
        }
    }
}
```


BoD (approved):
👍

PR #6611, opened 2025-07-04
Introduce `Service.generateApolloEnums` to generate enums with a `__Known` interface

martinbonnin:
This is a follow up from https://github.com/apollographql/apollo-kotlin/pull/6248 with the suggestions from @Emplexx


libraries/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -510,17 +510,27 @@ interface Service {
...
 
   /**
-   * A list of [Regex] patterns for GraphQL enums that should be generated as Kotlin sealed classes instead of the default Kotlin enums.
+   * A list of [Regex] patterns for GraphQL enums that should be generated as a Kotlin sealed interface.
    *
-   * Use this if you want your client to have access to the rawValue of the enum. This can be useful if new GraphQL enums are added but
-   * the client was compiled against an older schema that doesn't have knowledge of the new enums.
+   * This provides several benefits over the default of mapping GraphQL enums to Kotlin enums:
+   * - the client can access the string value of unknown values (enum values added on the server after the client has been compiled).
+   * - it introduces an intermediate `KNOWN__` type that does not contain the unknown value for the cases where you want to map all unknown values to a known one.
```
BoD, line 517: This is actually for `generateApolloEnums`, not for `sealedClassesForEnumsMatching`, right?

martinbonnin: Ah yes, good catch, I forgot to revert that part 👍

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/codegen/Identifiers.kt
```diff
@@ -63,11 +61,17 @@ internal object Identifier {
   const val knownValues = "knownValues"
   const val knownEntries = "knownEntries"
 
-  // extra underscores at the end to prevent potential name clashes
+  /**
+   * UNKNOWN__should probably have been __UNKNOWN because GraphQL reserves the leading __ but it's too late now.
```
BoD, line 65:
```suggestion
   * UNKNOWN__ should probably have been __UNKNOWN because GraphQL reserves the leading __ but it's too late now.
```

BoD (approved):
👍

PR #6616, opened 2025-07-15
Added default Accept header to introspection query

pedromfmachado:


BoD (approved):
Thanks!

pedromfmachado:
Hey @BoD, it seems the failing test also fails in the `main` branch. Is there something we should do about it or should it be fixed in a different PR?

BoD:
Actually I just opened #6618 to fix it 😅. We also have an issue with the docs job - I'll have a look at it.

In any case your fix looks good and not related to the error, I'll merge it now.

PR #6618, opened 2025-07-16
Fix JvmFileUploadTest

BoD:
Small regression introduced [here](https://github.com/apollographql/apollo-kotlin/pull/6581/files?w=1#diff-731c79854e920476a065378ba36ca1d7b52695e41cbcfe1436b8f71b648caa54R1) (not sure why we didn't catch it before).


martinbonnin:
Oh gosh, thanks for fixing this. What about we remove enhanced client awareness in this test? As of now it will require manual intervention for releases where the version length changes.

PR #6622, opened 2025-07-17
Introduce Service.pluginsArguments and relax the check for multiple plugins

martinbonnin:


libraries/apollo-gradle-plugin-tasks/src/main/kotlin/com/apollographql/apollo/gradle/task/apolloGenerateCodegenSchema.kt
```diff
@@ -11,7 +11,7 @@ import gratatouille.tasks.GTask
 @GTask
 internal fun apolloGenerateCodegenSchema(
     logger: GLogger,
-    arguments: Map<String, GAny?>,
+    arguments: Map<String, GAny>,
```
martinbonnin, line 14: It's not 100% clear whether Gradle `MapProperty` supports nullable values. To remove any ambiguity, I made this non-nullable.

martinbonnin:
> missed it, all good

I was adding it while you were reviewing, sorry forgot to add this earlier 😅

PR #6643, opened 2025-07-28
Centralize adding the kotlin-test library

martinbonnin:
For `kotlin-test`, we always want the same version as kotlinc or else we end up with errors such as https://youtrack.jetbrains.com/issue/KT-79627/WasmJS-NullPointerException-in-GenerateWasmTests.makeTestFunctionDeclaratorlambda4


BoD (approved):
👍

PR #6657, opened 2025-08-06
Make sure that the introspection endpoint property value is checked at execution time

francescocervone:
Fixes https://github.com/apollographql/apollo-kotlin/issues/6656

Honestly I didn't find a better place to move the existing check.
There is already [a check performed by the SchemaDownloader](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo/tooling/SchemaDownloader.kt#L92). 
It seemed redundant adding another one, but let me know what you think @martinbonnin.


martinbonnin (approved):
LGTM! Thanks for adding integration tests 🙏

PR #6686, opened 2025-08-27
Make `Optional.Absent` a `data object`

JakeWharton:
This produces a better `toString()` for logging purposes.

### DIFF
--- a/libraries/apollo-api/api/apollo-api.api
+++ b/libraries/apollo-api/api/apollo-api.api
@@ -831,6 +831,9 @@ public abstract class com/apollographql/apollo/api/Optional {
 
 public final class com/apollographql/apollo/api/Optional$Absent : com/apollographql/apollo/api/Optional {
 	public static final field INSTANCE Lcom/apollographql/apollo/api/Optional$Absent;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
 }
 
 public final class com/apollographql/apollo/api/Optional$Companion {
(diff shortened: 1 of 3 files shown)


martinbonnin (approved):

martinbonnin:
Thanks! I'm curious what's your take on implementing the `toString()` for debugging purposes. We have quite inconsistent `toString()` accross the codebase. 

It's nicer where you're in your debugger but it's also additional bytecode that is not always easy to tree shake? Where do you draw the line?

JakeWharton:
In this case, the `toString()` method only loads a constant string which matches the class name (meaning we get it for free), so we only pay for the cost of the instructions which are few. I actually didn't realize Kotlin generated `hashCode` and `equals` since it's not really needed. Probably an artifact of reusing the codepath for `data class`es but just with zero properties.

If you want to optimize for size, I would focus on the generated types first since they scale at a much higher rate than manually-written types like this one. Currently they're all marked with `data` so they're carrying the `componentN` functions, the `copy` function, as well as `toString()`/`equals`/`hashCode()` functions which are not the most optimal.

I've spent a lot of time [optimizing](https://jakewharton.com/the-economics-of-generated-code/) [generated](https://jakewharton.com/optimizing-bytecode-by-manipulating-source-code/) [code](https://github.com/drewhamilton/Poko/issues/204), and I'm just getting started using this library, so I can keep an eye out for places to reduce stuff as well.

martinbonnin:
There is [generateMethods](https://apollographql.github.io/apollo-kotlin/kdoc/older/4.2.0/apollo-gradle-plugin-external/com.apollographql.apollo.gradle.api/-service/generate-methods.html) to fine tune what gets generated in the models but we never dared change the default. Especially since we haven't had that much complaints about this, it just stayed like this.

martinbonnin:
The hard thing is how to measure. I'm down to adding `toString()` everywhere but I have no clue if this is ultimately going to slow down the loading on older samsung devices or if I'm just overthinking this.

martinbonnin:
Merging this. We can always add more `toString()` later.

JakeWharton:
Nothing else jumped out at me for lacking a `toString()`. I don't see many other `object`s which are in the public API or used in a way that I would expect to be logged ever. I am only using a small fraction of the library thus far, however.

PR #6701, opened 2025-09-19
Update gratatouille so that we can have an `apolloTasks` configuration that hold the classpath for the tasks

martinbonnin:


BoD (approved):
🚀

PR #6703, opened 2025-09-22
AGP9 support

martinbonnin:


libraries/apollo-gradle-plugin/src/main/kotlin/com/apollographql/apollo/gradle/api/Service.kt
```diff
@@ -808,54 +807,76 @@ interface Service {
...
+
+    /**
+     * Connects the generated sources to all the TestComponents.
+     */
+    fun connectToAndroidTestComponents(kotlin: Boolean)
+
+    /**
+     * Connects the generated sources to the given Android variant.
+     */
+    @Deprecated("use connectToAndroidComponent() instead", ReplaceWith("connectToAndroidComponent(variant)"))
```
BoD, line 847: Nit: I guess the replacement won't work because it needs a value for `kotlin`. Maybe it should default to true?

martinbonnin: https://github.com/apollographql/apollo-kotlin/pull/6703/commits/f86dd33ce288bad9e5e43504ebd49126028af2b3

PR #6705, opened 2025-09-22
Dependency updates

martinbonnin:
A bunch of updates from looking at our `libraries.toml` before 5.0 goes out. Most of our `libraries.toml` is for tests/benchmarks/build-logic, there shouldn't be anything breaking in terms of what we publish.


.github/workflows/benchmarks.yml
```diff
@@ -12,7 +12,7 @@ env:
 
 jobs:
   benchmarks:
-    runs-on: macos-14
+    runs-on: macos-15
```
martinbonnin, line 15: I didn't check all the workflows, I'll keep an eye on them in the coming days, please let me know if you see anything strange

docs/source/index.mdx
```diff
@@ -150,8 +150,8 @@ Some platforms have specific runtime requirements:
 At build time, it requires:
 
 * Gradle 8.0+
-* Kotlin 1.9+ for JVM projects
-* Kotlin 2.0+ for native, JS, and Wasm projects
+* Kotlin 2.0+ for JVM projects
```
martinbonnin, line 153: This is because okio pulls kotlin-stdlib 2.1. It's technically not a binary breaking change as the runtime stdlib will be compatible.

PR #6718, opened 2025-10-01
Do not check already checked fragments in checkCapitalizedFields

BoD:
Avoids a StackOverflow when there's a cycle. Cycles are checked before, but this check still runs after.


martinbonnin:
Excellent, thanks!

PR #6724, opened 2025-10-02
Enable K/N incremental compilation

martinbonnin:
See https://kotlinlang.org/docs/native-improving-compilation-time.html


BoD (approved):
🚀

PR #6729, opened 2025-10-03
[compiler] Allow empty deprecationReasons

martinbonnin:
Some users are doing this:

```graphql
type Query {
  foo: Int @deprecated(reason: "")
}
```

The motivation is questionable but nevertheless, [`empty` != `absent`](https://external-preview.redd.it/kRt0jItRkbMkJLfQRXpAc_oj-9O9nlVMsYCMWZmgedE.jpg?auto=webp&s=3e1170f8c4935f53d18a76c098dba919e0c646cc) and we should be consistent here.


BoD (approved):
👍

PR #6739, opened 2025-10-14
Add HTTP caching using OkHttp cacheUrlOverride

martinbonnin:
Deprecate all of `apollo-http-cache`. Everything is now in `DefaultHttpRequestComposer` and a few lines of code 🧹 `apollo-http-cache` will be removed in v6.

Because it bumps okhttp and okio, we now have Kotlin 2.2 transitive dependencies, which means KGP 2.1 is required to build Apollo projects. 

I have kept the default `languageVersion` to `2.1` because we can but it's not very useful TBH. 

`apollo-gradle-plugin` and `apollo-annotations`, which do not use `okio`, now target 2.0 (1.9 didn't work because of a klib resolver issue). This theorically bumps the Gradle requirement to 8.3 (embedded Kotlin 2.1) but tests are working with 8.0 so I'm assuming we're good there.

See also https://github.com/square/okhttp/issues/9032
Closes https://github.com/apollographql/apollo-kotlin/issues/6302

### DIFF
--- a/docs/source/migration/5.0.mdx
+++ b/docs/source/migration/5.0.mdx
@@ -139,7 +139,7 @@ val apolloClient = ApolloClient.Builder()
     .build()
 ```
 
-The OkHttp cache uses the standard `Cache-Control` HTTP header. Compared Apollo Kotlin 4 and `HttpFetchPolicy` the semantics are different but you can map most of the concepts:
+The OkHttp cache uses the standard `Cache-Control` HTTP header. Compared to Apollo Kotlin 4 and `HttpFetchPolicy` the semantics are different but you can map most of the concepts:
 
 **CacheFirst**:
 ```kotlin


docs/source/caching/http-cache.mdx
```diff
@@ -7,97 +7,78 @@ title: HTTP cache
...
-  // Finally, execute your query
-  .execute()
+  val successResponse = MockResponse.Builder()
+      .body(FooQuery.successResponse)
+      .addHeader("cache-control", "max-age=100")
+      .build()
+  mockServer.enqueue(successResponse)
+
+  apolloClient.query(FooQuery()).execute() // cache miss
+  apolloClient.query(FooQuery()).execute() // cache hit
```
BoD: Not sure if something is missing here?

docs/source/caching/http-cache.mdx
```diff
@@ -7,97 +7,78 @@ title: HTTP cache
...
-android {
-    compileOptions {
-        // Flag to enable support for the new language APIs
-        coreLibraryDesugaringEnabled = true
-    }
-}
+  val apolloClient = ApolloClient.Builder()
+        .serverUrl(url)
+        .okHttpClient(apolloClient) // highlight-line
+        .addInterceptor(CacheUrlOverrideInterceptor("https://example.com/graphql"))
```
BoD: Should this be `CacheUrlOverrideInterceptor(url)`?

```suggestion
        .addInterceptor(CacheUrlOverrideInterceptor(url)) 
```

Actually, if I understand correctly, this could be any URL, it doesn't have to match the server's URL, right? Which could be a bit puzzling :) But I guess it makes more sense if it is the same in the doc.

martinbonnin: Moving everything to `DefaultHttpRequestComposer` removes this all together since `serverUrl` is also going to be used for cache url override. 
There's always the possibility that our query parameters (`operationName`, `variablesHash`, `operationId`) clash with the user ones. If that happens, they will have to define their own `HttpRequestComposer`

docs/source/migration/5.0.mdx
```diff
@@ -96,3 +96,83 @@ query GetUser {
...
+                      .cache(directory = "http_cache2", maxSize = 10_000_000)
+                      .build()
+                }
+            )
+            .build()
+    )
+    .build()
+```
+
+The OkHttp cache uses the standard `Cache-Control` HTTP header. Compared Apollo Kotlin 4 and `HttpFetchPolicy` the semantics are different but you can map most of the concepts:
```
BoD:
```suggestion
The OkHttp cache uses the standard `Cache-Control` HTTP header. Compared to Apollo Kotlin 4 and `HttpFetchPolicy` the semantics are different but you can map most of the concepts:
```

BoD (approved):
This is great! 🧹

libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/http/DefaultHttpEngine.jvm.kt
```diff
@@ -23,38 +29,41 @@ import java.util.concurrent.TimeUnit
...
-fun DefaultHttpEngine(httpCallFactory: () -> Call.Factory): HttpEngine = JvmHttpEngine(httpCallFactory)
+actual fun DefaultHttpEngine(timeoutMillis: Long): HttpEngine = DefaultHttpEngine {
+  defaultOkHttpClientBuilder
+      .connectTimeout(timeoutMillis, TimeUnit.MILLISECONDS)
+      .readTimeout(timeoutMillis, TimeUnit.MILLISECONDS)
+      .build()
+}
 
-fun DefaultHttpEngine(okHttpClient: OkHttpClient): HttpEngine = JvmHttpEngine(okHttpClient)
+fun DefaultHttpEngine(connectTimeoutMillis: Long, readTimeoutMillis: Long): HttpEngine = DefaultHttpEngine {
```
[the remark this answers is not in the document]
martinbonnin: I didn't deprecated it because it's not calling `OkHttpClient()` from the main thread and it doesn't cost much to keep it.

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/http/DefaultHttpRequestComposer.kt
```diff
@@ -68,14 +77,24 @@ class DefaultHttpRequestComposer(
...
+          }
+          if (enablePostCaching && operation is Query<*>) {
+            val cacheParameters = mutableMapOf<String, String>()
+
+            val variables = operation.variables(customScalarAdapters)
+            if (variables.valueMap.isNotEmpty()) {
+              cacheParameters.put("variablesHash", variables.valueMap.sha256())
             }
+            cacheParameters.put("operationName", operation.name())
+            cacheParameters.put("operationId", operation.id())
```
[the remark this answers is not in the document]
BoD: 👍 Yeah it's nice for debugging

PR #6746, opened 2025-10-17
Ignore unknown keys when deserializing options

BoD:
The IJ plugin calls these methods - adding new keys shouldn't break it.


martinbonnin:
When does that happen? 
In theory, this is called to parse the intermediate artifacts of the compiler. If every intermediate is built using the same compiler version, it should all be consistent. Or is there another case there?

BoD:
> When does that happen?

The IJ plugin also reads these files to know which options are used, for telemetry.

BoD:
> Do you need to link with apollo-compiler for anything else? 

I'm still using `EntryPoints` and all the symbols it needs (at runtime it's the version of the project, not the linked version). 

But for reading the files yeah I can duplicate the logic and be lenient, that sounds safer 👍

PR #6758, opened 2025-10-31
Add `ApolloRequest.Builder.url(String)`

martinbonnin:
The url goes in `ExecutionOptions` and overrides whatever is set in `serverUrl()`, `webSocketServerUrl()`. 

It can be used in an `ApolloInterceptor` to change the value of the url dynamically for WebSockets (see https://github.com/apollographql/apollo-kotlin/issues/5862) but also in general for any request.

```kotlin
      ApolloClient.Builder()
          .addInterceptor(object : ApolloInterceptor {
            override fun <D : Operation.Data> intercept(
                request: ApolloRequest<D>,
            ): Flow<ApolloResponse<D>> {
              val url = if (request.operation is Subscription) {
                val token = getValidToken()
                "https://example.com/subscription?token=$token"
              } else {
                "https://example.com/graphql"
              }

              return chain.proceed(request.newBuilder().url(url).build())
            }
          })
          .build()
```

### DIFF
--- a/tests/integration-tests/src/commonTest/kotlin/test/FetchPolicyTest.kt
+++ b/tests/integration-tests/src/commonTest/kotlin/test/FetchPolicyTest.kt
@@ -3,7 +3,6 @@
 
 package test
 
-import app.cash.turbine.test
 import com.apollographql.apollo.ApolloClient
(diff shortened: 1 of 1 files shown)


BoD (approved):
🚀

PR #6764, opened 2025-11-03
Introduce `RetryStrategy`

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/5862#issuecomment-3242853663

In addition to the new API, this:

* stabilizes the retry API. It's been around for more than 1 year now. Might as well consider it stable. I'm still not 100% certain we reached the end of the road. There are a lot of heuristics down there and finding the good default that is also flexible is not easy but for the sake of moving forward, let's commit to this and go through deprecation cycles if we ever need to change.
* embraces the subscription vs query differences. I think most subscriptions want to be retried when it's not always the case for queries. The default retry behaviour is now to retry forever for subscriptions and give up after 7 seconds for queries.

### DIFF
--- a/docs/source/advanced/network-connectivity.mdx
+++ b/docs/source/advanced/network-connectivity.mdx
@@ -53,7 +53,7 @@ When a `NetworkMonitor` is configured, `retryOnError` uses `NetworkMonitor.waitF
 
 ### Customizing the retry algorithm
 
-You can customize the retry algorithm further by using a `RetryStrategy`. A very basic `RetryStrategy` that uses exponential backoff and gives up after 4 tries looks like this:
+To customize the retry algorithm, use a `RetryStrategy`. The following example shows a `RetryStrategy` that uses exponential backoff and gives up after four tries: 
 
 ```kotlin
 val apolloClient = ApolloClient.Builder()


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/interceptor/RetryOnErrorInterceptor.kt
```diff
@@ -61,39 +72,47 @@ private class DefaultRetryOnErrorInterceptorImpl(private val networkMonitor: Net
...
   }
 }
 
 private fun ApolloException.isRecoverable(): Boolean {
   /**
    * TODO: refine this. Some networks errors are probably not recoverable (SSL errors probably, maybe others?)
    */
-  return this is ApolloNetworkException
+  return when (this) {
+    is ApolloNetworkException,
```
BoD: IIRC `OfflineApolloException` (which is a `ApolloNetworkException`) should be excluded here, otherwise we retry when we should fail fast if offline

docs/source/advanced/network-connectivity.mdx
```diff
@@ -57,35 +53,20 @@ When a `NetworkMonitor` is configured, `retryOnError` uses `NetworkMonitor.waitF
...
+      if (exception == null) {
+        return@RetryOnErrorInterceptor false
       }
-    }
-  }
-}
+
+      delay(2.0.pow(state.attempt).seconds)
+      state.attempt++
+      return@RetryOnErrorInterceptor false
```
BoD:
```suggestion
      return@RetryOnErrorInterceptor state.attempt < 4
```

martinbonnin: Good catch!

BoD (approved):

Issue #6749, opened 2025-10-21
Exclude `@typePolicy` and `@fieldPolicy` from the implicitly imported `kotlin_labs` definitions when we the cache plugin is present


BoD:
We want to move `@typePolicy` and `@fieldPolicy` to the `cache` spec, but removing is breaking so currently they still exist both in `cache` and `kotlin_labs`. Since the `kotlin_labs` directives are imported implicitly we get an error when importing the `cache` ones.

A workaround is to "un-import" the `kotlin_labs` like this:

```graphql
    extend schema
    @link(url: "https://specs.apollo.dev/kotlin_labs/v0.5")
    @link(url: "https://specs.apollo.dev/cache/v0.3", import: ["@typePolicy", "@fieldPolicy"])
```

but this is not a great dev experience.

Instead, we can detect that the cache plugin is used, and exclude these directives from being implicitly imported.

PR #6766, opened 2025-11-03
Exclude auto-imported cache related directives when the Apollo cache plugin is present

BoD:
Fix for #6749


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/SchemaValidationScope.kt
```diff
@@ -54,14 +58,32 @@ import com.apollographql.apollo.ast.withBuiltinDefinitions
  */
 @ApolloExperimental
 class SchemaValidationOptions(
-    val addKotlinLabsDefinitions: Boolean,
+    val addKotlinLabsDefinitions: AddKotlinLabsDefinitions,
```
martinbonnin: Probably worth avoiding the breaking change here

martinbonnin:
> I think that's on purpose because we want the newer definitions to be explicitly imported.

Right. Probably means we need (korlin_labs:0.3 - cache) somewhere. Unless there is a difference in `kotlin_labs:0.3:@typePolicy` vs `cache:0.3:@typePolicy`?

BoD:
No, they are the same.

So we can auto link `kotlin_labs/0.3-minus-cache` if the cache plugin is present, else auto link `kotlin_labs/0.3` 👍

PR #6775, opened 2025-11-05
Restore JsonReader state if a field throws in-flight

martinbonnin:
The `JsonReader` is stateful and error-aware parsing may throw on any value. If that's the case, catch the exception and keep the `JsonReader` state in sync.


martinbonnin:
Merging quickly so we can test snapshots

PR #6803, opened 2025-12-05
Add support for directives on directive definitions

BoD:
This is based on [this draft spec change](https://github.com/graphql/graphql-spec/pull/1206).

### DIFF
--- a/libraries/apollo-execution/src/commonMain/kotlin/com/apollographql/apollo/execution/internal/introspection.kt
+++ b/libraries/apollo-execution/src/commonMain/kotlin/com/apollographql/apollo/execution/internal/introspection.kt
@@ -263,6 +263,12 @@ internal fun introspectionResolvers(schema: Schema): Map<String, Resolver> {
       "description" to Resolver {
         it.parentObject.cast<GQLDirectiveDefinition>().description
       },
+      "isDeprecated" to Resolver {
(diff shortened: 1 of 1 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
```diff
@@ -737,6 +738,7 @@ internal class Parser(
     expectToken<Token.At>()
     val name = parseName()
     val arguments = parseArgumentDefinitions()
+    val directives = parseDirectives(const = true)
```
martinbonnin: Can you put this in `ParserOptions`?  I'd like all new additions to the spec to be opt-in.

martinbonnin (changes requested):
LGTM except this should be opt-in until this is in the spec. 

Also do you mind updating the execution part of it? ([here](https://github.com/apollographql/apollo-kotlin/blob/d6e1222c0c4009fd01275f5893c62d06858336cd/libraries/apollo-execution/src/commonMain/kotlin/com/apollographql/apollo/execution/internal/introspection.kt#L259)) Or I can do it later if you prefer.

martinbonnin (approved):

PR #6820, opened 2025-12-24
Gradle Plugin: Avoid eager configuration of Gradle objects

simonlebras:
- Use `register` instead of `create` for configurations to support lazy evaluation.
- Wrap `Configuration` objects in `project.files()` to prevent eager resolution.
- Update `Configurations` data class and internal methods to use `NamedDomainObjectProvider<Configuration>`.

### DIFF
--- a/libraries/apollo-gradle-plugin/lint-baseline.xml
+++ b/libraries/apollo-gradle-plugin/lint-baseline.xml
@@ -30,7 +30,7 @@
         errorLine2="                                                                         ~~~~~~">
         <location
             file="build/gtask/gratatouilleUnzipPluginSources/outputDirectory/com/apollographql/apollo/gradle/task/ApolloDownloadSchemaTask.kt"
-            line="55"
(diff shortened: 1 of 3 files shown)


martinbonnin:
Thanks for opening this. Looks good to me except this:

> Wrap Configuration objects in project.files() to prevent eager resolution.

I haven't been able to confirm the current code is resolving the configuration eagerly. I believe it might be a false positive in the lint rules. Do you have any evidence of eager configuration there?

simonlebras:
Replacing `create` with `register` change the types from `Configuration` to `NamedDomainObjectProvider<Configuration>`, so that Configurations are not created until they are actually needed.
I used `project.files` to create a "live" FileCollection that will only resolve the configuration when the files are actually needed and preserve laziness. I also needed from an API point of view, because NamedDomainObjectProvider is a Provider, and not FileCollection.

martinbonnin (approved):

PR #6823, opened 2026-01-05
Bump librarian and update badge url

martinbonnin:
New librarian version fixes empty version list in `maven-metadata.xml`


BoD (approved):
👍

PR #6824, opened 2026-01-05
Add pagination benchmarks

BoD:
Add a benchmark that writes 2 pages then reads the cache. Also adds the same with pagination disabled so we can compare.


martinbonnin:
What do the results look like?

BoD:
Getting this on my Galaxy S8:

```
   8,403,713   ns       13974 allocs    noPaginationMemory
   9,755,863   ns       16692 allocs    paginationMemory

   25,976,903   ns       15605 allocs     noPaginationSql
   26,348,941   ns       18645 allocs     paginationSql

   24,937,326   ns       14788 allocs     noPaginationMemoryThenSql
   26,595,691   ns       17497 allocs     paginationMemoryThenSql
```

PR #6834, opened 2026-01-12
Add `ApolloCall.extensions()`

martinbonnin:
Add `ApolloCall.extensions()` to make it easier to set request extensions. See https://github.com/apollographql/apollo-kotlin/issues/5627

This also refactors `DefaultHttpRequestComposer` and has a couple behavior changes:

1. `variables: {}` is now skipped if the variables are empty.
1. `buildPostBody()` will fail if the query document doesn't match the operation document.
1. The order of the request parameter changed. They are now sent in the [same order as in the spec](https://graphql.github.io/graphql-over-http/draft/#sec-Request-Parameters): `query` goes first, then `operationName`, `variables` (if any) and finally `extensions`.


BoD (approved):
🚀

PR #6858, opened 2026-01-28
Add support for service capabilities

martinbonnin:
See https://github.com/graphql/graphql-spec/pull/1208

The parser has good tests. Introspection/execution still need a bit more tests but since this is RFC1, I expect things to change a bit and I don't want to spend too much time updating every test.

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
@@ -449,10 +449,6 @@ internal class Parser(
       parseServiceCapability()
     }.orEmpty()
 
-    if (directives.isEmpty() && capabilities.isEmpty()) {
-      unexpected()
-    }
-
     return GQLServiceExtension(
         sourceLocation = sourceLocation(start),
         directives = directives,


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
```diff
@@ -369,6 +403,23 @@ internal class Parser(
     )
   }
 
+  private fun parseServiceDefinition(start: Token): GQLServiceDefinition {
+    val description = parseDescription()
+
+    expectKeyword("service")
+    val directives = parseDirectives(const = true)
+    val capabilities = parseNonEmptyList<Token.LeftBrace, Token.RightBrace, GQLCapability> {
```
BoD:
```suggestion
    val capabilities = parseList<Token.LeftBrace, Token.RightBrace, GQLCapability> {
```
I think?

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
```diff
@@ -389,6 +440,26 @@ internal class Parser(
...
+  private fun parseServiceExtension(): GQLServiceExtension {
+    val start = token
+    expectKeyword("extend")
+    expectKeyword("service")
+    val directives = parseDirectives(const = true)
+    val capabilities = parseNonEmptyListOrNull<Token.LeftBrace, Token.RightBrace, GQLCapability> {
+      parseServiceCapability()
+    }.orEmpty()
+
+    if (directives.isEmpty() && capabilities.isEmpty()) {
```
BoD: Same here, this makes sense but doesn't match the current spec.

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/introspection/introspection_writer.kt
```diff
@@ -28,19 +30,21 @@ import com.apollographql.apollo.ast.findOneOf
...
- * The Json Model matching the `__Schema` [GraphQL type](https://spec.graphql.org/draft/#sec-The-__Schema-Type)
+ * The JSON Model matching the `__Schema` [GraphQL type](https://spec.graphql.org/draft/#sec-The-__Schema-Type)
  *
  * It matches the draft version of the spec. Some fields did not exist in earlier versions
  * like `__InputValue.isDeprecated` for an example.
  *
- * The models match with the spec types except for Type that is used recursively, and therefore
- * we have [WTypeFull], [WTypeRef] and [WTypeRoot]
+ * TODO: This is potentially an issue because it means adding new features changes the output of this function.
+ * For an example, it doesn't support service capabilities as of now to avoid confusing other external tools.
```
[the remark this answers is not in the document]
BoD: Agreed this should probably be deprecated, I can't think of a use-case for it. Ah except maybe to generate fixtures for our own tests? Meh.

libraries/apollo-tooling/src/main/resources/base-introspection.graphql
```diff
@@ -24,6 +24,7 @@ query IntrospectionQuery {
       # deprecationReason - introduced in https://github.com/graphql/graphql-spec/pull/907
     }
   }
+  # __service { } - experimental
```
BoD:
```suggestion
  # __service { } - experimental, introduced in https://github.com/graphql/graphql-spec/pull/1208
```

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/Parser.kt
martinbonnin: Ah, yes, I remember that discussion. There are multiple issues of people wanting to have empty objects/selection sets and IIRC, @benjie used capabilities as a place to toy with empty blocks. I'll update that code so that it matches the current PR.

PR #6874, opened 2026-02-14
Use public AGP API for version detection instead of internal class

abeggsnf:
## Summary
Replace reflection access to internal `com.android.Version` class with `AndroidComponentsExtension.pluginVersion` public API. The internal class was causing `ClassNotFoundException` in some AGP 8.x versions.

## Changes
- Access `pluginVersion.major` directly via reflection on public API
- Remove `AgpVersion.kt` as it's no longer needed  
- Maintain compatibility with AGP 8.0+ and 9.0+

## Testing
Tested with AGP 8.0.0, 8.13.0, and 9.0.0-alpha10. All Android project tests pass.

## Motivation
The `com.android.Version` class is internal to AGP and was removed or relocated in some AGP 8.x versions, causing `ClassNotFoundException`. The `AndroidComponentsExtension.pluginVersion` property is the public API way to access the AGP version and has been available since AGP 7.0+.

### DIFF
--- a/libraries/apollo-gradle-plugin/src/agp-8/kotlin/com/apollographql/apollo/gradle/Agp8.kt
+++ b/libraries/apollo-gradle-plugin/src/agp-8/kotlin/com/apollographql/apollo/gradle/Agp8.kt
@@ -2,6 +2,7 @@
 
 package com.apollographql.apollo.gradle
 
+import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.gradle.AppExtension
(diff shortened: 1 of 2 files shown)


martinbonnin:
@abeggsnf no problem! The original issue is that we used `Thread.currentThread().classLoader` and Gradle seems to change threads sometimes during configuration (especially during `afterEvaluate`), which means the `Version` class wasn't accessible.

This should all be fixed in the latest SNAPSHOTs.

PR #6922, opened 2026-04-13
CacheBatchReader: merge data in the presence of include directives

### LINKED ISSUE #6901: Fetch failures with `@include` directive
### Version

5.0.0-alpha.5-20260208

### Summary

I'm fetching the same object in two fragments, but it one fragment that entire object is guarded by an `@include` directive. With this setup, when I do fetch the object in the second place (with `@include(if: true)`) I see parse failures — looks like fields from the _first_ fragment are not requested, but they're still parsed.

This is with `codegenModels = "responseBased"`

Schema:
```graphql
type Query {
  item: Item
}

type Item {
  id: ID!
  details: Details
}

type Details {
  id: ID!
  title: String!
  summary: String
  tags: [String!]
}
```

Operation:
```graphql
query GetItem($includeExtras: Boolean!) {
  item {
    ...CoreFields
    ...ExtraFields
  }
}

fragment CoreFields on Item {
  id
  details {
    id
    title
    summary
  }
}

fragment ExtraFields on Item {
  details @include(if: $includeExtras) {
    id
    tags
  }
}
```

What I observe is:
 - when `includeExtras = true`
 - the API call fails during parsing with `title == null` error

The generated adapter seems to merge fields from both `CoreFields` and `ExtraFields` fragment and parse them in one function, but then pass both fragments there separately? Either way the parsing fails despite the response being correct. The issue seems to go away if the `@include` directive is placed on individual fields rather than entire `details` object

### Logs

_No response_

martinbonnin:
Fixes #6901

### DIFF
--- a/libraries/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/api/internal/CacheBatchReader.kt
+++ b/libraries/apollo-normalized-cache-api/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/api/internal/CacheBatchReader.kt
@@ -125,7 +125,23 @@ internal class CacheBatchReader(
           it.responseName to value
         }.toMap()
 
-        data[pendingReference.path] = map
+        if (data.contains(pendingReference.path)) {
+          /**
+           * In the presence of include directives, we might already have read some data.
+           * In that case, we need to merge them.
+           *
+           * Note1: if a key is present in both the existing and the new data, we expect the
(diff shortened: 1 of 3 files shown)


BoD (approved):
Does this need to be backported to the new cache?

martinbonnin:
Most probably yes but I didn't check the implementation there.

BoD:
>  Most probably yes but I didn't check the implementation there.

I had a look, looks like yes :). I'll open a PR.

PR #6933, opened 2026-04-25
Remove "built-in" from the caching docs since they are in an external repo now

martinbonnin:
No strong opinion but this felt a bit weird


BoD (approved):
👍

PR #6939, opened 2026-04-30
Optimize the field merging algorithm.

### LINKED ISSUE #6938: Compiler performance regression in fields_merging.kt
### Version

5.0.0-rc.0

### Summary

The codegen duration for our largest apollo module increased more than 2x after updating to v5. Based on LLM analysis, the source of the added latency appears to be in fields_merging.kt where it suggests a few improvements, in order from highest to lowest impact:
- **Visited-set caching:**
Added visitedShapeSets and visitedParentSets to prevent re-validating the same field sets in the recursive sameResponseShapeByName and sameForCommonParentsByName phases. This avoids exponential re-expansion of the same fragment/field combinations. This is consistent with the caching approach in the [article](https://tech.new-work.se/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01) referenced at the top of the file.
- **mergeSubSelections result caching:**
Cached the results of mergeSubSelections so both validation phases share the same collectFields results instead of independently re-expanding fragments.
- **Fragment argument fast path:**
Skips the expensive substituteVariables + pushFragment/popFragment cycle when a fragment spread has no arguments (which is the case for all fragments in this project).

I attached a patch file with the code changes it suggested (they seem plausible, but I'm unfamiliar with the code so I can't vouch for correctness):
[fields_merging_perf_fix.patch](https://github.com/user-attachments/files/27177776/fields_merging_perf_fix.patch)

_No response_

### Logs

```
(Your logs here)
```

martinbonnin:
Fixes #6938

### DIFF
--- a/libraries/apollo-ast/test-fixtures/validation/executable/graphql-js/field_merging/field_merging.graphql
+++ b/libraries/apollo-ast/test-fixtures/validation/executable/graphql-js/field_merging/field_merging.graphql
@@ -1,0 +1,4389 @@
+fragment Fragment0164 on Type1575 {
+  field1840
+  field1841
+  field1842
+  field1843 {
+    field0020
+  }
+}
+
+fragment Fragment0208 on Type0253 {
+  field0047 {
+    ...Fragment0010
+  }
+  field0501 {
+    ...Fragment0004
+  }
+  field0283 {
+    ...Fragment0004
+  }
+}
+
+fragment Fragment0158 on Type0874 {
+  field0023
+  field2184
+  field2183
+  field2182
+}
+
+fragment Fragment0308 on Type0093 {
+  field0174
+  field0545
+  field0109 {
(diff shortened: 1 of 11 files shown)


BoD (approved):
🚀

libraries/apollo-ast/test-fixtures/validation/executable/graphql-js/field_merging/field_merging.graphql
```diff
@@ -0,0 +1,4391 @@
+# See https://github.com/apollographql/apollo-kotlin/issues/6938
+
+fragment Fragment0164 on Type1575 {
```
martinbonnin, line 3: @apramana, since this is anonymized, I assumed it was OK to add to the test suite, let us know otherwise and I'll remove it.

PR #6941, opened 2026-05-05
Add KDoc for experimental AST features

martinbonnin:
Add KDoc to experimental features.


BoD (approved):
👍

martinbonnin:
@BoD I changed my mind and kept the `@ApolloExperimental` annotation unchanged. Users will have to opt-in even if not using the feature but simpler is better than complex.

PR #6949, opened 2026-05-13
Workaround introspection for servers that do not support isOneOf despite advertizing it

martinbonnin:
See https://github.com/apollographql/apollo-kotlin/issues/6945


BoD (approved):
👍

PR #6958, opened 2026-05-29
Docs: add compatibility matrix

martinbonnin:


BoD (approved):
👍 

We probably need the same on the cache repo.

PR #6961, opened 2026-06-03
Fix duplicate headers in Batching Interceptor

barnhill:
Addresses https://github.com/apollographql/apollo-kotlin/issues/6960

Version
4.4.2

Summary
When using httpBatching and auto persisted queries together on the same Apollo call, the headers are being duplicated by BatchingHttpInterceptor (from what I can tell) when the APQ replay happens due to it adding headers without preventing duplicates here:

```
return chain.proceed(request.newBuilder().addHeaders(headers = request.headers.filter { it.name != ExecutionOptions.CAN_BE_BATCHED }).build())
```
Steps to reproduce the behavior
configure the apollo client with:

```
.httpBatching(enableByDefault = false)
.autoPersistedQueries(httpMethodForHashedQueries = HttpMethod.Post)
```

and make any call.

### DIFF
--- a/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/BatchingHttpInterceptor.kt
+++ b/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/BatchingHttpInterceptor.kt
@@ -93,7 +93,7 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
 
     if (!canBeBatched) {
       // Remove the CAN_BE_BATCHED header and forward directly
-      return chain.proceed(request.newBuilder().addHeaders(headers = request.headers.filter { it.name != ExecutionOptions.CAN_BE_BATCHED }).build())
+      return chain.proceed(request.newBuilder().headers(request.headers.filter { it.name != ExecutionOptions.CAN_BE_BATCHED }).build())
     }
 
     // Keep the chain for later


libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
```diff
@@ -991,7 +991,7 @@ private constructor(
         enableByDefault: Boolean = true,
     ) = apply {
       _httpInterceptors.removeAll { it is BatchingHttpInterceptor }
-      addHttpInterceptor(BatchingHttpInterceptor(batchIntervalMillis, maxBatchSize))
+      addHttpInterceptor(BatchingHttpInterceptor(batchIntervalMillis, maxBatchSize, false, enableByDefault))
       canBeBatched(enableByDefault)
```
martinbonnin, line 995: Thanks for opening this! This line should already take care of enabling batching by default. Did you find it's not the case?

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/network/http/BatchingHttpInterceptor.kt
```diff
@@ -88,12 +90,12 @@ class BatchingHttpInterceptor @JvmOverloads constructor(
...
   override suspend fun intercept(request: HttpRequest, chain: HttpInterceptorChain): HttpResponse {
-    // Batching is enabled by default, unless explicitly disabled
-    val canBeBatched = request.headers.valueOf(ExecutionOptions.CAN_BE_BATCHED)?.toBoolean() ?: true
+    // Batching is enabled if no header exists only if the default is to enable it by default
+    val canBeBatched = request.headers.valueOf(ExecutionOptions.CAN_BE_BATCHED)?.toBoolean() ?: enableByDefault
 
     if (!canBeBatched) {
       // Remove the CAN_BE_BATCHED header and forward directly
-      return chain.proceed(request.newBuilder().addHeaders(headers = request.headers.filter { it.name != ExecutionOptions.CAN_BE_BATCHED }).build())
+      return chain.proceed(request.newBuilder().headers(request.headers.filter { it.name != ExecutionOptions.CAN_BE_BATCHED }).build())
```
martinbonnin, line 96: 👍

libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt
barnhill: Correct.  It defaults to batching on for everything even if you set `enableByDefault` to false.  If the intention is to have it available but not utilized by every request unless opted in by calling `canBeBatched` on each call then it would need to default to the same value as the `enableByDefault` flag unless Im missing something.

martinbonnin: This is surprising, we have a test for this [here](github.com/apollographql/apollo-kotlin/blob/5a9edd48d9524fe18080a2d0d3161e56a2a24927/tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt#L161):

```kotlin
    mockServer.enqueueString("""{"data":{"launch":{"id":"83"}}}""")
    mockServer.enqueueString("""{"data":{"launch":{"id":"84"}}}""")
        .serverUrl(mockServer.url())
        .httpBatching(batchIntervalMillis = 10, maxBatchSize = 10, enableByDefault = false)
        .build()

    val result1 = async {
      apolloClient.query(GetLaunchQuery()).execute()
    }
    val result2 = async {
      delay(50)
      apolloClient.query(GetLaunch2Query()).execute()
    }
    assertEquals("83", result1.await().dataOrThrow().launch?.id)
    assertEquals("84", result2.await().dataOrThrow().launch?.id)

    // 2 requests are made
    mockServer.awaitRequest()
    mockServer.awaitRequest()
```

(That snippet has a duplicate `canBeBatched(false)` removed but I think it's still relevant).

What am I missing?

barnhill: Basically if the header doesnt exist (CAN_BE_BATCHED) it will fall back to the default inside BatchingHttpInterceptor which will fall back to true.  But the user has specified that the default should be false, to allow other queries to opt in if they want.  This would opt everyone in by default since the canBeBatched would be set to true, but the consumer has specified false as the default when calling `.httpBatching()`.  This default value never gets propagated to the interceptor except by the way of the header but this is the case where the header doesnt exist, in which case default is set to true in all cases.

I hope that makes sense, been a long day :). Super glad to work with ya and if Im missing something, point me in the right direction.  Thanks for your time and hope you have a great day!!

barnhill: Let me know if you would like me to remove that change and can address it separately if it helps.

martinbonnin: > if the header doesnt exist (CAN_BE_BATCHED) 

I'm trying to understand how come the header would not exist. It is set [here](https://github.com/apollographql/apollo-kotlin/blob/d7b7aed250a0c0ebfa50ab0628d2529429eebfcc/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt#L296):

```kotlin
      val canBeBatched = canBeBatched ?: apolloClient.canBeBatched
      if (canBeBatched != null) {
        // Because batching is handled at the HTTP level, move the information to HTTP headers
        // canBeBatched(apolloRequest.canBeBatched)
        addHttpHeader(ExecutionOptions.CAN_BE_BATCHED, canBeBatched.toString())
      }

```

This should happen way before `BatchingHttpInterceptor` and the `canBeBatched` variable should be set from [`httpBatching()`](https://github.com/apollographql/apollo-kotlin/blob/d7b7aed250a0c0ebfa50ab0628d2529429eebfcc/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt#L995) but maybe there's an interaction with other interceptors that disables this. 

Could you add an integration test next to [the existing one](https://github.com/apollographql/apollo-kotlin/blob/5a9edd48d9524fe18080a2d0d3161e56a2a24927/tests/integration-tests/src/commonTest/kotlin/test/batching/QueryBatchingTest.kt#L161) ?  I'd like to get to the bottom of this. (you'll need to open the `tests` directory in your IDE to work in integration tests)

barnhill: On vacation today so it might be a day or so.  Let me separate these two issues so I don't hold up the main culprit that we know is an issue.  I'll log a separate issue and MR so it's easier to manage and continue the investigation on there if that's ok with you?

martinbonnin: Sounds great, thanks!

barnhill: @martinbonnin got just the fix we discussed isolated in this PR now.

martinbonnin:
Thanks!

PR #6963, opened 2026-06-05
Add `ApolloClient` support for `onError`

martinbonnin:
Previously only execution supported it. Moves `OnError` to `apollo-api`


BoD (approved):
🚀

PR #6970, opened 2026-06-22
Enable error propagation by default

martinbonnin:
The directive is unconditionally added on the GraphQL Java side so it feels dangerous to add it without opt-in here.


BoD (approved):

martinbonnin:
Closing as non-actionable.

PR #6975, opened 2026-06-29
Expose services as a `NamedDomainObjectCollection`

alvr:
## Motivation

Up until now, Apollo services were kept internally as a `MutableList<DefaultService>` without any public API for looking them up or configuring them post-creation. This made it pretty much impossible to tweak a service defined in one module from another.

Now, with the new `apollo.services.named("service").configure {}` API, downstream modules and convention plugins can finally configure an already-registered service by name.

## Changes

- **New `ApolloServicesContainer` interface**: this exposes `val services: NamedDomainObjectCollection<Service>` on `ApolloExtension`.
- **`Service` extends Gradle's `Named`**: this replaces the standalone `name` property. It lets you look up services by name through `NamedDomainObjectCollection`.
- **`ApolloExtension` extends `ApolloServicesContainer`**: it delegates to a `DefaultApolloServicesContainer` using `by` delegation.
- **Backed by `DomainObjectContainer`**: we've swapped out `MutableList<DefaultService>` for Gradle's managed `DomainObjectContainer`, which gives us handy methods like `named()`, `matching()`, and other usual collection operations without extra work.
- **Read-only exposed collection**: trying to directly mutate it (like `add`, `remove`, `clear`, etc.) will throw an `UnsupportedOperationException`, with a message guiding users to `apollo.service("name") { }`.
- **Better duplicate-name error**: duplicate names are caught earlier (right in `service()`) and the error message recommends `apollo.services.named("name").configure {}` as the fix.

## Breaking changes

1. **Services are now sorted alphabetically, not by insertion order.** Gradle's `NamedDomainObjectContainer` holds elements in a `SortedSet` keyed by `name`. If you loop through `apollo.services` or depend on `first()`/`last()` ordering, expect a shift: it’ll be `[alpha, beta]` instead of `[beta, alpha]` if `beta` was added first. We updated the `GradleToolingTests` to reflect this (the service assertions were reordered).

2. **`Service.name` moved to `Named.getName()`.** It’s still source-compatible (the property is still `name`), but the API dump (`apollo-gradle-plugin.api`) now shows `getName()` being inherited from `org.gradle.api.Named`, not originally declared on `Service`.

3. **`services` is now a public API on `ApolloExtension`.** Before, the backing list was a `private val` and couldn’t be accessed from build scripts. Now, `apollo.services` is available for everyone and returns a read-only collection. If someone had a local script variable named `services`, it might just overshadow this new property.

---

This is basically a proposal to allow configuration outside of the main `service` registry. In one of my projects, I have the main configuration in a [convention plugin](https://github.com/alvr/katana/blob/77b19d396128add297112c0bf0fddb5667e8995a/build-logic/katana-convention/src/main/kotlin/dev/alvr/katana/buildlogic/mp/data/KatanaMultiplatformDataRemotePlugin.kt#L49-L76). To avoid having to configure all the modules in the same plugin, I prefer to configure them in their own module; otherwise, there would be an endless number of conditions for each module. You can also be able to use typesafe project accessors if enabled.

`common/media/data/gradle.build.kts`
```kotlin
apollo.services.named("service").configure {
    generateAsInternal = false
}
```

`feature/lists/data/gradle.build.kts`
```kotlin
apollo.services.named("service").configure {
    dependsOn(projects.common.media.data)
}
```

### DIFF
--- a/libraries/apollo-gradle-plugin/api/apollo-gradle-plugin.api
+++ b/libraries/apollo-gradle-plugin/api/apollo-gradle-plugin.api
@@ -25,7 +25,7 @@ public final class com/apollographql/apollo/gradle/api/ApolloDependencies {
 	public final fun getRuntime ()Lorg/gradle/api/artifacts/Dependency;
 }
 
-public abstract interface class com/apollographql/apollo/gradle/api/ApolloExtension {
+public abstract interface class com/apollographql/apollo/gradle/api/ApolloExtension : com/apollographql/apollo/gradle/api/ApolloServicesContainer {
 	public abstract fun createAllAndroidVariantServices (Ljava/lang/String;Ljava/lang/String;Lorg/gradle/api/Action;)V
 	public abstract fun createAllKotlinSourceSetServices (Ljava/lang/String;Ljava/lang/String;Lorg/gradle/api/Action;)V
 	public abstract fun getDeps ()Lcom/apollographql/apollo/gradle/api/ApolloDependencies;
@@ -101,6 +101,10 @@ public abstract interface class com/apollographql/apollo/gradle/api/ApolloGradle
 	public abstract fun getWarnOnDeprecatedUsages ()Ljava/lang/Boolean;
 }
 
+public abstract interface class com/apollographql/apollo/gradle/api/ApolloServicesContainer {
+	public abstract fun getServices ()Lorg/gradle/api/NamedDomainObjectCollection;
+}
+
 public abstract interface class com/apollographql/apollo/gradle/api/CompilerPlugin {
 	public abstract fun argument (Ljava/lang/String;Ljava/lang/Object;)V
 }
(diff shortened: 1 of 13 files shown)


martinbonnin:
Hi @alvr 👋 

I know Gradle encourages us to use `NamedDomainObjectCollection` but it's something I'd like to avoid for 2 reasons:

* Some of the `Service` APIs have side effects. Specifically, `outputDirConnection {}` and the likes wire tasks together and calling them several times will have unexpected consequences.
* More generally, I have been burnt several times with different plugins modifying properties in different places. It makes it really hard to track who's changing a property. I don't want to encourage global mutable state in build logic.

Instead, you can define your convention plugin extension to account for things you'd like to tweak:

```kotlin
// Define an extension for your convention plugin
abstract class KatanaMultiplatformDataRemoteExtension {
    abstract val packageName: Property<String>
    abstract val generateAsInternal: Property<Boolean>
    // ...
}

internal class KatanaMultiplatformDataRemotePlugin : Plugin<Project> {
    override fun apply(target: Project) =
        with(target) {
            val extension = extensions.create("dataRemote", KatanaMultiplatformDataRemoteExtension::class.java)

            apply(plugin = "katana.multiplatform.core")
            apply(plugin = "com.apollographql.apollo")

            with(extensions) {
                configure<KotlinMultiplatformExtension> { configureMultiplatform() }
                configure<ApolloExtension> { configureApollo(extension) }
            }
        }

// ...

    private fun ApolloExtension.configureApollo(extension: KatanaMultiplatformDataRemoteExtension) {
        service("anilist") {
            decapitalizeFields = true
            // Use the extension to configure the apollo plugin
            generateAsInternal = extension.generateAsInternal.orElse(true)
            generateMethods = listOf("equalsHashCode")
            packageName = extension.packageName.orElse(project.fullPackageName)

// ...

```

It's more verbose but it makes it clear what is the separation of concerns.

PR #6978, opened 2026-07-03
Make it possible to use the cache 1.0.4 with v5 on K/N

martinbonnin:
Fix K/N ABI
On native, making a function parameter nullable is a binary breaking change.
To fix that, we introduce back the non-nullable version as hidden.


BoD (approved):
🚀

PR #6984, opened 2026-07-13
Fix CatchToResult toJson

### LINKED ISSUE #6983: @catch(to: RESULT): serializing a FieldResult.Failure corrupts the JSON writer
### Version

5.0.1

### Summary

`CatchToResultAdapter.toJson` writes nothing at all for a `FieldResult.Failure`:

https://github.com/apollographql/apollo-kotlin/blob/b87c1a8b1b574ca7b3dc67bd7f74bee68c2f258f/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt#L437-L442

By the time the field adapter runs, the containing object adapter has already called `writer.name(fieldName)`. Writing nothing leaves the writer with a dangling pending name; serializing any subsequent field in the same object then throws:

```
java.lang.IllegalStateException: Check failed.
    at com.apollographql.apollo.api.json.MapJsonWriter.name(MapJsonWriter.kt:121)
```

If the `Failure` happens to be the last field serialized in its object, `endObject()` tolerates the pending name and no exception is thrown, but the composed output then silently omits the errored field instead of representing it as `null` (which is what the server sent). So the output is wrong in every case:
```kotlin
// query UserResult { user @catch(to: RESULT) { name } }  - the Failure is the last (only) field
UserResultQuery().adapter().toJson(writer, CustomScalarAdapters.Empty, dataWithFailure)
writer.root() // {} - no "user" key at all
```

For the normalized cache, this omission means the stored record is simply missing the field, causing cache misses on later reads instead of replaying the stored error.

```graphqls
# schema.graphqls
type Query {
  user: User
  product: Product
}

type User {
  name: String
}

type Product {
  price: String
}

# extra.graphqls
extend schema @link(url: "https://specs.apollo.dev/nullability/v0.4", import: ["@catch", "CatchTo", "@catchByDefault"])
extend schema @catchByDefault(to: NULL)
```

```graphql
query UserResultAndProduct {
  user @catch(to: RESULT) {
    name
  }
  product {
    price
  }
}
```

```kotlin
val data = UserResultAndProductQuery.Data(
    user = FieldResult.Failure(
        ApolloGraphQLException(Error.Builder("cannot resolve user").path(listOf("user")).build())
    ),
    product = UserResultAndProductQuery.Product(price = "10"),
)

// throws java.lang.IllegalStateException: Check failed. (MapJsonWriter.name)
UserResultAndProductQuery().adapter().toJson(MapJsonWriter(), CustomScalarAdapters.Empty, data)
```

### Logs

_No response_

dhritzkiv:
Fixes #6983 
  
`CatchToResultAdapter.toJson` writes nothing for a `FieldResult.Failure` (`else -> Unit`), leaving the `JsonWriter` with a
dangling field name. Composing any later sibling throws `IllegalStateException: Check failed.` from `MapJsonWriter.name`,
and a `Failure` in the last position is silently omitted from the composed output. This surfaces through the normalized
cache, which composes every response before writing it (full analysis and stack trace in the issue).

The fix writes the `null` the server sent for the errored field, mirroring `CatchToNullAdapter`.

### Tests

The existing catch tests only exercise parsing, and no operation had a RESULT-caught field followed by a sibling, so this
adds a `UserResultAndProduct` operation to `tests/catch` and three regression tests to `CatchResultTest`:

- `composeFailureFollowedBySibling` — threw `IllegalStateException` before the fix
- `composeFailureAsLastField` — composed `{}` instead of `{user=null}` before the fix
- `composeRoundTripsParsedFailure` — parse → compose symmetry for the existing `userNameError` fixture

All three fail on `main` without the fix; `:catch:test` and `:catch-responseBased:test` pass with it.

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
@@ -437,7 +437,8 @@ private class CatchToResultAdapter<T>(private val wrappedAdapter: Adapter<T>) : 
   override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: FieldResult<T>) {
     when (value) {
       is FieldResult.Success -> wrappedAdapter.toJson(writer, customScalarAdapters, value.getOrThrow())
-      else -> Unit // ignore errors
+      // write the null the server sent for the errored field, so that the enclosing object stays well-formed
+      else -> writer.nullValue()
     }
   }
 }
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
@@ -437,7 +437,10 @@ private class CatchToResultAdapter<T>(private val wrappedAdapter: Adapter<T>) : 
   override fun toJson(writer: JsonWriter, customScalarAdapters: CustomScalarAdapters, value: FieldResult<T>) {
     when (value) {
       is FieldResult.Success -> wrappedAdapter.toJson(writer, customScalarAdapters, value.getOrThrow())
-      // write the null the server sent for the errored field, so that the enclosing object stays well-formed
+      /**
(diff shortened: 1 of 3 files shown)


martinbonnin:
Good catch. Thanks for contributing this 💙

PR #6994, opened 2026-08-14
perf(compiler): index the type hierarchy and drop O(n^2) work from the schema checks

AlexanderGH:
Measured on a 3.9 MB schema (4350 type definitions, 347 interfaces and unions) with 139 operations. `buildIrOperations` goes from 425 ms to 226 ms. Most of what it spends is schema-wide and identical in every module using the schema, so the win scales with the number of modules.

`Schema`: index the type hierarchy once instead of scanning every type definition on every call.

- `possibleTypes` scanned all the type definitions to find the implementations of an interface, and did it again at every level of the hierarchy. It now walks an interface name -> direct implementations map. Iterating all the abstract types: 108 ms -> 4 ms.
- `implementedTypes` and `superTypes` scanned all the type definitions to find the unions a type belongs to. They now use a member name -> union names map. Iterating all the object types: 2270 ms -> 10 ms.
- `rootTypeNameOrNullFor` re-filtered and re-associated all the definitions on every call, even though the root operation types are already resolved in the constructor. 10k calls: 446 ms -> 0 ms.

Both indexes are built lazily and keep schema declaration order, so the returned sets iterate exactly as before and the generated sources are unchanged.

`checkApolloTargetNameClashes`: copied the accumulated key set on every one of the type definitions, and keyed its intermediate maps by `GQLTypeDefinition`, which hashes a whole type subtree per lookup. It now keeps a list and passes the live key set to `uniqueName`, which only reads it: 148 ms -> 1 ms.

`SchemaSerializer`: encoded the schema as a Json *string* containing Json, so the SDL was escaped twice and the nested document buffered into a string. Encoding a `JsonElement` inline instead takes the codegen schema artifact from 4,456,148 to 4,076,650 bytes and its deserialization from 40 ms to 35 ms. Every module reads that artifact at least twice.

`toCodegenMetadata`: read the whole file twice, once for the version check.

Also remove `ValidationContext.mergeSubSelectionsCache`, which is never read or written.

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
@@ -54,6 +54,43 @@ class Schema internal constructor(
 
   val subscriptionTypeDefinition: GQLTypeDefinition? = rootOperationTypeDefinition("subscription", definitions)
 
+  /**
+   * Maps an interface name to the object and interface types implementing it directly, in schema declaration order.
+   *
+   * Without this, [possibleTypes] has to scan every type definition of the schema on every call, and again for
+   * every level of the interface hierarchy.
+   */
+  private val directSubtypes: Map<String, List<GQLTypeDefinition>> by lazy {
+    val result = mutableMapOf<String, MutableList<GQLTypeDefinition>>()
+    typeDefinitions.values.forEach { typeDefinition ->
+      val implementsInterfaces = when (typeDefinition) {
+        is GQLObjectTypeDefinition -> typeDefinition.implementsInterfaces
+        is GQLInterfaceTypeDefinition -> typeDefinition.implementsInterfaces
+        else -> return@forEach
+      }
+      implementsInterfaces.forEach {
+        result.getOrPut(it) { mutableListOf() }.add(typeDefinition)
+      }
+    }
+    result
+  }
+
+  /**
+   * Maps a type name to the names of the unions it is a member of, in schema declaration order.
+   *
+   * Without this, [implementedTypes] and [superTypes] have to scan every type definition of the schema on every call.
+   */
+  private val unionsByMember: Map<String, List<String>> by lazy {
+    val result = mutableMapOf<String, MutableList<String>>()
+    typeDefinitions.values.forEach { typeDefinition ->
+      if (typeDefinition !is GQLUnionTypeDefinition) return@forEach
+      typeDefinition.memberTypes.forEach {
+        result.getOrPut(it.name) { mutableListOf() }.add(typeDefinition.name)
+      }
+    }
+    result
+  }
+
   fun toGQLDocument(): GQLDocument = GQLDocument(
       definitions = definitions,
       sourceLocation = null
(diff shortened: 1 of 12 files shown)


libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/internal/fields_merging.kt
```diff
@@ -57,7 +57,6 @@ private data class ValidationContext(
   val visitedShapeSets = mutableSetOf<Set<FieldAndType>>()
   val visitedParentSets = mutableSetOf<Set<FieldAndType>>()
   val variableValuesInScope = mutableListOf<Map<String, GQLValue>>()
-  val mergeSubSelectionsCache = mutableMapOf<Set<FieldAndType>, Map<String, MutableSet<FieldAndType>>>()
```
martinbonnin, line 60: 👍

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
```diff
@@ -69,7 +106,16 @@ class Schema internal constructor(
...
-    return rootOperationTypeDefinition(operationType, definitions)?.name
+    /*
+     * The root operation types are resolved once in the constructor. Do not resolve them again here, that would
+     * mean filtering all the definitions on every call.
+     */
+    return when (operationType) {
+      "query" -> queryTypeDefinition.name
+      "mutation" -> mutationTypeDefinition?.name
+      "subscription" -> subscriptionTypeDefinition?.name
+      else -> rootOperationTypeDefinition(operationType, definitions)?.name
```
martinbonnin:
```suggestion
      else -> error("Unknown operation type: '$operationType'")
```

libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
```diff
@@ -92,12 +138,8 @@ class Schema internal constructor(
   fun possibleTypes(typeDefinition: GQLTypeDefinition): Set<String> {
     return when (typeDefinition) {
       is GQLUnionTypeDefinition -> typeDefinition.memberTypes.map { it.name }.toSet()
-      is GQLInterfaceTypeDefinition -> typeDefinitions.values.filter {
-        it is GQLObjectTypeDefinition && it.implementsInterfaces.contains(typeDefinition.name)
-            || it is GQLInterfaceTypeDefinition && it.implementsInterfaces.contains(typeDefinition.name)
-      }.flatMap {
+      is GQLInterfaceTypeDefinition -> directSubtypes[typeDefinition.name].orEmpty().flatMap {
         // Recurse until we reach the concrete types
-        // This could certainly be improved
```
martinbonnin, line 100: 🎉

libraries/apollo-compiler/src/main/kotlin/com/apollographql/apollo/compiler/internal/serialization.kt
```diff
@@ -29,18 +29,25 @@ import kotlinx.serialization.json.intOrNull
...
-    val map = Json.parseToJsonElement(decoder.decodeString()).toAny()
+    val map = decoder.decodeSerializableValue(JsonElement.serializer()).toAny()
 
     @Suppress("UNCHECKED_CAST")
     return Schema.fromMap(map as Map<String, Any>)
   }
 
   override fun serialize(encoder: Encoder, value: Schema) {
-    encoder.encodeString(value.toMap().toJsonElement().toString())
+    encoder.encodeSerializableValue(JsonElement.serializer(), value.toMap().toJsonElement())
```
martinbonnin, line 50: That's a breaking change in the format of the metadata. If anyone publishes their Apollo metadata, they will have to consume them with a matching version. I don't think this is too much of a problem. Just something worth highlighting in the changelog.

martinbonnin (approved):
Looks great. Just one minor comment.

I added a benchmark [there](https://github.com/apollographql/apollo-kotlin/pull/6999).

On my machine, we go from ~19ops/s to ~46ops/s for `buildIrOperations()` on 100 queries using the GitHub schema 🥳

PR #6995, opened 2026-08-14
perf(ast): skip the @defer walk and index field definitions

AlexanderGH:
- `@defer` walk skipped — ExecutableValidationScope computes documentHasDeferDirective once per document with a cheap non-expanding scan of the operations and fragment definitions (hasDeferDirective in defer_labels.kt), and only then calls validateDeferLabels, which inlines every fragment spread per operation. Also hoisted the label Regex to a top-level val.
- Field-definition index — Schema holds a lazy Map<typeName, Map<fieldName, GQLFieldDefinition>> (meta fields included); definitionFromScope delegates to it. A GQLTypeDefinition that isn't identical to the schema's own definition falls back to the old rebuild-and-scan, so the public API keeps working for synthetic definitions.
- usedFragments memoized per fragment name, with the map owned by IrOperationsBuilder so it spans all operations in a module.

Measured on a schema (3.9 MB, 4350 types) with 139 operations:

```
┌───────────────────────────────────────┬─────────────┬────────────┬────────────┐
│                metric                 │  upstream   │  PR 6994   │ This PR    │
├───────────────────────────────────────┼─────────────┼────────────┼────────────┤
│ buildIrOperations                     │ 450 ms      │ 226 ms     │ 69 ms      │
├───────────────────────────────────────┼─────────────┼────────────┼────────────┤
│ toCodegenSchema (per module)          │ 53 ms       │ 48 ms      │ 48 ms      │
├───────────────────────────────────────┼─────────────┼────────────┼────────────┤
│ possibleTypes over all abstract types │ 7 ms        │ 0 ms       │ 0 ms       │
├───────────────────────────────────────┼─────────────┼────────────┼────────────┤
│ serialized IR size                    │ 2,955,473 B │ same       │ same       │
└───────────────────────────────────────┴─────────────┴────────────┴────────────┘
```

The IR serializes byte-for-byte identically, which is the strongest signal that codegen output is unchanged. Green: :apollo-ast:jvmTest (incl. the 144 ExecutableValidationTest fixtures — duplicate_path_and_labels_spread covers `@defer` living only inside fragment definitions, exactly the case the early-out could have broken), :apollo-compiler:test (247 golden CodegenTest cases), apiCheck on both modules, and tests/defer:jvmTest via the composite build. Added FieldDefinitionTest (5 tests) for the field index, including the foreign-type-definition fallback.

### DIFF
--- a/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
+++ b/libraries/apollo-ast/src/commonMain/kotlin/com/apollographql/apollo/ast/Schema.kt
@@ -91,6 +91,33 @@ class Schema internal constructor(
     result
   }
 
+  /**
+   * Maps a type name to its field definitions by name, including the meta fields (see [GQLTypeDefinition.fieldDefinitions]).
+   *
+   * Without this, resolving a field definition rebuilds the field definitions of the parent type and scans them
+   * linearly, and that happens for every field of every selection set.
+   */
+  private val fieldDefinitionsByName: Map<String, Map<String, GQLFieldDefinition>> by lazy {
+    typeDefinitions.mapValues { (_, typeDefinition) ->
+      buildMap {
+        typeDefinition.fieldDefinitions(this@Schema).forEach {
+          // Keep the first one, like the linear scan this replaces: an invalid schema may define a field twice
+          if (!containsKey(it.name)) {
+            put(it.name, it)
+          }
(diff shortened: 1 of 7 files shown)


martinbonnin (approved):
Note: the long term fix for `@defer` validation is to implement some kind of visitor pattern so we don't pay for the traversal price several times. Problem for another day.

PR #6996, opened 2026-08-14
perf(api): stop using exceptions to guess JSON number types

AlexanderGH:
# What changed

The dominant cost in response parsing was readAny()'s guessNumber(), which called nextInt() → nextLong() → nextDouble() in sequence and kept the first that didn't throw. Every Long paid for one JVM exception, every fractional Double for two. For projects with queries that use fragments heavily, nearly every object goes through obj(buffered = true) → readAny(), so this ran on every number in every response.

A single generic rewrite would not have been behavior-preserving — MapJsonReader holds JsonNumber values whose semantics are String-based (JsonNumber("1.0") narrows to Double today, but to Int under a naive double-ladder). So the fix dispatches on the concrete reader (each knows its own number representation and can narrow without throwing) and leaves third-party JsonReader implementations on the original try-chain.

Secondary allocation removals: ErrorAwareAdapter skips getPath() when the response has no errors; parseData() inlines the null check instead of allocating a NullableAdapter and shares the adapters map instead of copying it via newBuilder(); falseVariables() returns emptySet() when an operation has no boolean variables; responseAdapterFor() does one map lookup and a when on the class name instead of a listOf(...) per branch test; ApolloClient/HttpNetworkTransport build their interceptor chain lists once instead of per request.

# Numbers

392,777-byte response (40 rows × 30 entities, epoch-millis Longs + fractional Doubles), median of 11 rounds:

```
┌─────────────────┬──────────┬──────────┬──────┐
│                 │ baseline │   new    │      │
├─────────────────┼──────────┼──────────┼──────┤
│ parseResponse   │ 6310 µs  │ 1981 µs  │ 3.2x │
├─────────────────┼──────────┼──────────┼──────┤
│ readAny         │ 4075 µs  │ 1472 µs  │ 2.8x │
├─────────────────┼──────────┼──────────┼──────┤
│ allocated/parse │ 9165 KiB │ 3073 KiB │ 3.0x │
└─────────────────┴──────────┴──────────┴──────┘
```

An Int-only-numbers control with the same response shape gives 1970 → 1896 µs and 3030 → 2993 KiB, which attributes the wall-clock win entirely to removing the exceptions; the other changes are allocation reductions too small to see at this granularity.

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
@@ -409,7 +409,8 @@ class ObjectAdapter<T>(
 
 private class ErrorAwareAdapter<T>(private val wrappedAdapter: Adapter<T>) : Adapter<T> {
   override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): T {
-    if (reader.peek() == JsonReader.Token.NULL) {
+    // reader.getPath() allocates, only compute it if there are errors that could match
+    if (!customScalarAdapters.errors.isNullOrEmpty() && reader.peek() == JsonReader.Token.NULL) {
       val error = customScalarAdapters.firstErrorStartingWith(reader.getPath())
       if (error != null) {
         reader.skipValue()
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/json/MapJsonReader.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/json/MapJsonReader.kt
@@ -4,7 +4,9 @@ import com.apollographql.apollo.api.json.BufferedSourceJsonReader.Companion.INIT
 import com.apollographql.apollo.api.json.MapJsonReader.Companion.buffer
 import com.apollographql.apollo.api.json.internal.toDoubleExact
 import com.apollographql.apollo.api.json.internal.toIntExact
+import com.apollographql.apollo.api.json.internal.toIntExactOrNull
 import com.apollographql.apollo.api.json.internal.toLongExact
+import com.apollographql.apollo.api.json.internal.toLongExactOrNull
 import com.apollographql.apollo.exception.JsonDataException
 import kotlin.jvm.JvmOverloads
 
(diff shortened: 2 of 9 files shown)


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/Adapters.kt
```diff
@@ -409,7 +409,8 @@ class ObjectAdapter<T>(
 
 private class ErrorAwareAdapter<T>(private val wrappedAdapter: Adapter<T>) : Adapter<T> {
   override fun fromJson(reader: JsonReader, customScalarAdapters: CustomScalarAdapters): T {
-    if (reader.peek() == JsonReader.Token.NULL) {
+    // reader.getPath() allocates, only compute it if there are errors that could match
+    if (!customScalarAdapters.errors.isNullOrEmpty() && reader.peek() == JsonReader.Token.NULL) {
```
martinbonnin, line 413: 👍

libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/json/MapJsonReader.kt
```diff
@@ -319,6 +321,35 @@ constructor(
...
+   *
+   * This is the same as calling [nextInt], [nextLong], [nextDouble] and [nextNumber] in turn and
+   * keeping the first one that doesn't throw, but without paying for the exceptions. Only call this
+   * when [peek] is [JsonReader.Token.NUMBER] or [JsonReader.Token.LONG].
+   */
+  internal fun nextGuessedNumber(): Any {
+    val result = when (val value = peekedData) {
+      is Int -> value
+      is Long -> value.toIntExactOrNull() ?: value
+      is Double -> value.toIntExactOrNull() ?: value.toLongExactOrNull() ?: value
```
martinbonnin:
```suggestion
      is Double -> value.toIntExactOrNull() ?: value.toLongExactOrNull() ?: value
      is Int -> value
      is Long -> value.toIntExactOrNull() ?: value
```

Workaround for https://youtrack.jetbrains.com/issue/KT-18177.
THe JS tests are currently failing because `1e12` is stored as a JS `number` so both `isDouble` and `isInt` is true.

martinbonnin (approved):
Thanks!

PR #6998, opened 2026-08-14
Unbreak the dashboards

martinbonnin:
The search results changed and weren't pointing to the benchmarks dashboard anymore but to https://github.com/apollographql/apollo-kotlin/pull/5608

Hardcode the benchmarks issueId instead


BoD (approved):
🚀

PR #7001, opened 2026-08-17
perf(api): Test a fragment's type condition against a set, not a list

AlexanderGH:
Collecting the fields to read or write for an object walks the operation's selections and, at every fragment, asks whether the object's `__typename` is one of that fragment's possible types. `CompiledFragment.possibleTypes` is a `List`, so the question is answered by a linear scan comparing strings one at a time.

The cost is paid per object rather than per fragment. A list of a hundred objects with an interface selection runs the same scan a hundred times, and an abstract type with many implementations makes each scan longer, so the work grows with the product of the two -- over strings that were fixed when the operation was compiled.

`CompiledFragment` now derives a set from `possibleTypes` and exposes `isPossibleType(typename)`, which the field collectors of the normalized cache and the fake resolver call instead of testing the list. The set is built on first use, so a selection whose type condition is never tested against a concrete type does not pay for it, and it is shared across every object of a list because generated selections are static.

`possibleTypes` keeps its `List` type and its place in the public API, and generated code is unchanged: the new method is purely additive.

In a profile of a large scrolling feed in a production app, these scans were roughly 3% of all CPU spent in the GraphQL client, and a noticeable part of its time in `String.equals`.

### DIFF
--- a/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CompiledGraphQL.kt
+++ b/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CompiledGraphQL.kt
@@ -218,31 +218,29 @@ class CompiledField internal constructor(
  */
 class CompiledFragment internal constructor(
     val typeCondition: String,
-    val possibleTypes: List<String>,
+    val possibleTypesSet: Set<String>,
     val condition: List<CompiledCondition>,
     val selections: List<CompiledSelection>,
 ) : CompiledSelection() {
-  /**
-   * [possibleTypes] as a set, so that [isPossibleType] does not scan it.
-   *
-   * Built on first use: a selection whose type condition is never tested against a concrete type
-   * does not pay for it. [LazyThreadSafetyMode.PUBLICATION] because building it is idempotent and
-   * generated selections are shared, so the cost of contending for a lock outweighs the cost of
-   * occasionally building the set twice.
-   */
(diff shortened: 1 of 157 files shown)


libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo/api/CompiledGraphQL.kt
```diff
@@ -214,6 +222,25 @@ class CompiledFragment internal constructor(
...
+   *
+   * Built on first use: a selection whose type condition is never tested against a concrete type
+   * does not pay for it. [LazyThreadSafetyMode.PUBLICATION] because building it is idempotent and
+   * generated selections are shared, so the cost of contending for a lock outweighs the cost of
+   * occasionally building the set twice.
+   */
+  private val possibleTypesSet: Set<String> by lazy(LazyThreadSafetyMode.PUBLICATION) {
+    possibleTypes.toSet()
+  }
+
```
martinbonnin: Can we invert the logic here and avoid creating the list in the first place? Something like this: 

```kotlin
class CompiledFragment internal constructor(
    val typeCondition: String,
    val possibleTypesSet: Set<String>,
    val condition: List<CompiledCondition>,
    val selections: List<CompiledSelection>,
) : CompiledSelection() {
  @Deprecated("Use possibleTypesSet instead")
  val possibleTypes: List<String> by lazy(LazyThreadSafetyMode.PUBLICATION) {
    @Suppress("DEPRECATION")
    possibleTypes.toList()
  }
```

Issue #6078, opened 2024-07-25
Cleanups


bignimbus:
This is a list of changes that we don't want to forget about.  They are likely to be backwards-incompatible so most of these will need to wait for a future major release. This list is mostly a braindump and is missing a lot of context. **This is not about new features but rather a list of small cleanups to do**.

Continued from https://github.com/apollographql/apollo-kotlin/issues/4171

If you see anything that doesn't make any sense, it's probably because it doesn't. Feel free to comment and we'll dive into it.

### Potential Breaking Changes/cleanups

* [ ] Enforce opt-in the `kotlin_labs` linking
* [ ] [Rename `Foo.type` to `Foo.__type` to avoid nameclashes](https://github.com/apollographql/apollo-kotlin/pull/4144#issuecomment-1141109867) (or just leave it like this because we can detect the case. See https://github.com/apollographql/apollo-kotlin/issues/4293)
* [ ] Rewrite the buffered adapters to not use JsonReader anymore
* [ ] Cleanup the `JsonReader` API and distinguish between the buffer/non buffered case. 
* [ ] Generate the customScalarAdapters?
* [ ] Make the field merging validation more strict (see https://github.com/apollographql/apollo-kotlin/pull/4342#pullrequestreview-1083903524)
* [ ] Use scalar instead of customScalar https://github.com/apollographql/apollo-kotlin/issues/3815
* [ ] Refactor `shapes()` 
* [ ] Move httpInfo and others outside of executionContext
* [ ] Improve the Java ergonomics of `HttpKt`
* [ ] (To be considered): pass the variant name to the lambda for `createAllAndroidVariantServices` (see [this comment](https://github.com/apollographql/apollo-kotlin/issues/4490#issuecomment-1303256930)) (or deprecate the function?)
* [ ] [Rename `WsProtocol.Factory` to `WsProtocol.Builder`](https://github.com/apollographql/apollo-kotlin/pull/4589/files#r1050794657)
* [ ] Remove IdlingResource. See [this comment](https://github.com/JakeWharton/okhttp-idling-resource/issues/23#issuecomment-1353225191) from OkHttp for instance, and [this article](https://medium.com/androiddevelopers/alternatives-to-idling-resources-in-compose-tests-8ae71f9fc473) about how to avoid them with Compose.
* [ ] Remove circular dependencies between Gradle projects
* [ ] extract HttpEngine and WebSocketEngine to a common module
* [ ] Remove Gradle defaultService

[...]

PR #7018, opened 2026-09-05
Allow custom JVM HTTP engines without OkHttp

gregjotau:
HTTP-only JVM consumers cannot currently exclude OkHttp even after supplying a custom `HttpEngine`: `ApolloClient` eagerly creates its unused default WebSocket engine, which links `okhttp3.WebSocket.Factory` and fails with `NoClassDefFoundError` during construction.

Defer the default JVM WebSocket engine until its first connection. Closing an unused engine does not initialize it, creation and closure are synchronized, and existing public APIs and the default OkHttp dependency remain unchanged. Document the dependency exclusion for consumers providing their own engines.

This helps ReAI's Kotlin/Spring Boot backend retain Apollo's generated parsing and runtime while reusing its existing HTTP stack for ten Shopify queries/mutations, with no subscriptions. It makes the roughly 848 KiB OkHttp JVM JAR removable from that distribution. Related to the engine extraction item in #6078; this is a focused fix rather than the full module split.

Validation:
- A separate `jvmTestWithoutOkHttp` task, included in `check`, runs query/mutation execution and client closure with OkHttp JARs removed from the test JVM. Confirmed failure with the original factory and success with this change.
- Lifecycle tests cover deferred initialization, repeated use, closure before use, repeated closure, and rejection after closure.
- All 66 JVM runtime tests, the isolated no-OkHttp test, and `:apollo-runtime:apiCheck` pass.

### DIFF
--- a/libraries/apollo-runtime/build.gradle.kts
+++ b/libraries/apollo-runtime/build.gradle.kts
@@ -97,3 +97,16 @@ kotlin {
     }
   }
 }
+
+val jvmTest = tasks.named<Test>("jvmTest")
+val jvmTestWithoutOkHttp = tasks.register<Test>("jvmTestWithoutOkHttp") {
+  testClassesDirs = jvmTest.get().testClassesDirs
+  classpath = jvmTest.get().classpath.filter { !it.name.startsWith("okhttp") }
+  filter.includeTestsMatching("NoOkHttpTest")
+}
+jvmTest.configure {
+  filter.excludeTestsMatching("NoOkHttpTest")
+}
+tasks.named("check") {
+  dependsOn(jvmTestWithoutOkHttp)
+}
--- a/libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketEngine.jvm.kt
+++ b/libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketEngine.jvm.kt
@@ -113,7 +113,32 @@ internal class JvmWebSocket(
 }
 
 
-actual fun WebSocketEngine(): WebSocketEngine = JvmWebSocketEngine { defaultOkHttpClientBuilder.build() }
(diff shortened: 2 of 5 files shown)


libraries/apollo-runtime/build.gradle.kts
```diff
@@ -97,3 +97,16 @@ kotlin {
     }
   }
 }
+
+val jvmTest = tasks.named<Test>("jvmTest")
+val jvmTestWithoutOkHttp = tasks.register<Test>("jvmTestWithoutOkHttp") {
+  testClassesDirs = jvmTest.get().testClassesDirs
+  classpath = jvmTest.get().classpath.filter { !it.name.startsWith("okhttp") }
+  filter.includeTestsMatching("NoOkHttpTest")
+}
```
martinbonnin, line 106: Nice tests!

libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/websocket/WebSocketEngine.jvm.kt
```diff
@@ -113,7 +113,32 @@ internal class JvmWebSocket(
 }
 
 
-actual fun WebSocketEngine(): WebSocketEngine = JvmWebSocketEngine { defaultOkHttpClientBuilder.build() }
+actual fun WebSocketEngine(): WebSocketEngine = LazyWebSocketEngine {
+  // Loading JvmWebSocketEngine links OkHttp classes, even though its client is lazy.
+  JvmWebSocketEngine { defaultOkHttpClientBuilder.build() }
+}
+
+internal class LazyWebSocketEngine(private val factory: () -> WebSocketEngine) : WebSocketEngine {
```
martinbonnin, line 121: We probably need to wrap the whole engine. Just changing the constructor signature should do the job. Follow up PR there: https://github.com/apollographql/apollo-kotlin/pull/7020

martinbonnin (commented):
Thanks! I can't push to your branch since it looks like it's an org-owned branch. I made a follow up PR there: https://github.com/apollographql/apollo-kotlin/pull/7020

PR #7027, opened 2026-09-15
Fix @catch handling of malformed response data

### LINKED ISSUE #7014: `fragment`ed queries silently nullify the closest nullable parent in hierarchy instead of failing the whole query when parsing fails
### Version

`5.1.0`

### Summary

Given this schema:

```graphql
type Query {
  user: User
}

type User {
  id: Int!
}
```

and this response:

```json
{
  "data": {
    "user": {
      "id": null
    }
  }
}
```

these two queries produce different `ApolloResponse`s.

**Query without a fragment:**

```graphql
query GetUserStandard {
  user {
    id
  }
}
```

Result:

```kotlin
response.data == null
response.exception is JsonDataException
```

**Query with a fragment:**

```graphql
query GetUserFragmented {
  user {
    ...UserFragment
  }
}

fragment UserFragment on User {
  id
}
```

Result:

```kotlin
response.data == GetUserFragmentedQuery.Data(user = null)
response.exception == null
```

### Expected Behavior

Both queries should produce the same result for the same response.

### Actual Behavior

The direct query returns a parsing exception. The query with the fragment returns successful data with `user = null` and no exception.

I created a repo to reproduce the issue
- Check out https://github.com/mecoFarid/apollo-7014
- Run tests

AvikMakwana:
Fixes #7014

Malformed response data could be swallowed by `@catch` adapters because they caught all `ApolloException`s, including parsing errors such as `JsonDataException`.

This narrows the adapters to catch `ApolloGraphQLException`, allowing malformed response data to propagate as a parsing failure while preserving GraphQL error handling.

Adds regression coverage for fragmented queries and both `NULL` and `RESULT` catch behavior.


martinbonnin (approved):
Thank you!

PR #7028, opened 2026-09-16
Default to allowing empty selection sets

martinbonnin:
The proposal is RFC3 now. Make `allowEmptySelectionSets` default to true and deprecate it. This also mirrors graphql-js: https://github.com/graphql/graphql-js/pull/4852


BoD (approved):
👍

